Issue #104, opened 2024-09-12
Make `MemoryResource` core and concrete MR implementations public


leofang:
In beta 1 (#87) part of MR is added but mostly for placeholder purposes. We would like to have the core RMM Python refactored into `cuda.py`.

leofang:
Moved #634 to the backlog. The remaining is done. Closing.

PR #446, opened 2025-02-12
Mempool memory resource - IPC

keenan-simpson:
This is ready for at least a first look

Add Mempool class as the first public MemoryResource implementation. It supports IPC.

Add IPC buffer which is an internal buffer implementation used to represent a buffer descriptor which is used to export and import buffers across process boundaries. It implements __reduce__ which is expected by common multiprocessing libraries. It's use is seen in the tests added to this review.


cuda_core/cuda/core/experimental/_memory.py
```diff
@@ -238,6 +277,414 @@ def device_id(self) -> int:
...
+    system = platform.system()
+    if system == "Linux":
+        return driver.CUmemAllocationHandleType.CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR
+    elif system == "Windows":
+        return driver.CUmemAllocationHandleType.CU_MEM_HANDLE_TYPE_WIN32
+    else:
+        raise RuntimeError(f"Unsupported platform: {system}")
+
+
+class Mempool(MemoryResource):
```
jrhemstad: question: What is the intention with this `Mempool` type? Is this just a placeholder until we can wrap the C++ resource implementations from CCCL?

keenan-simpson: The current state of the review lacks a solution for the integration of the Mempool memory resource with the device API. 

The goal of the review is to provide an avenue for users to create and use IPC-able allocations. Currently device is attached to _DefaultAsyncMempool by default, but the user can set the device.mr with the mr setter. Whichever memory resource backs device is responsible for device.allocate() -> buffer. I am working on fitting this together with that semantic, such that it is easy to use. edit: Its possible that the best way to do it, is not attach it to the device and let people use it as-is.

I will take some time to familiarize myself with the CCCL resource implementations as I'm not up to speed with CCCL yet.

leofang: Let me put a hard stop here and note that integration with `cuda::memory_resource` and RMM Python will not start until later (after GTC / beta 3). I think there are some design discussions we'll need to go over first before merging (ex: this class is independent of `_DefaultAsyncMempool` is concerning). Whatever we do between now and later, the new features need to be future compatible/extensible.

leofang: Another note is that `MemoryResource` generally is not required to be bindable with a `Device` object, for example there could be a multi-device or collective `MemoryResource` (an internal team is working on this). So
>  is not attach it to the device and let people use it as-is. 

this is the right way to go. 

Perhaps what's missing from the original `MemoryResource` design is a `.is_single_device` (name TBD) attribute that we should ask `MemoryResource` providers to implement, and check it during `Device.memory_resource = mr` assignment...

jrhemstad: > Whatever we do between now and later, the new features need to be future compatible/extensible.

How are we going to achieve this when the implementation in this PR is incompatible with the `cuda::memory_resource` concept defined in `cuda::mr`? Simple example is `allocate/deallocate` not taking an alignment in the parameter list. 

I'm concerned that we're moving forward implementing a thing called a "memory resource" that actually has no connection to the canonical memory resource interface and implementations we're defining in CCCL.

leofang: > Simple example is `allocate/deallocate` not taking an alignment in the parameter list.

I am not sure what all of this is about? Please do not panic... 🙂 
1. please feel free to create an issue to track any gap, as always
2. there was no single question asked on the current `MemoryResource`'s (placeholder) design, with regard to lack of `alignment` argument, during all past online/offline design reviews
3. it was made clear in various internal/external occasions that all **private** `MemoryResource` APIs existing in `cuda.core` is a placeholder just enough for us to make progress while waiting for `cuda::mr`/RMM integration (ex: https://github.com/NVIDIA/cuda-python/issues/104)
4. If we look at RMM Python ([link](https://github.com/rapidsai/rmm/blob/88eaaaf4cfffdc4c4cdab805829e55be65707ee0/python/rmm/rmm/pylibrmm/memory_resource.pyx#L85)) or any other Python allocators (ex: PyTorch, Numba, CuPy, ...), pretty much everyone takes no `alignment` argument (and likely makes internal assumptions about alignment). It means in the grand scheme of things this is not the most urgent task that needs our full attention. 

If this is about the PR description stating "this is ready," as noted this is a miscommunication and it is **not ready** to merge yet, as we did not have enough time to go through all the design decisions. This is ready in the sense of exploration efforts are made to guide design discussions.

jrhemstad: > a placeholder just enough for us to make progress while waiting for cuda::mr/RMM integration

Okay, makes sense. 

To be clear, I explicitly asked for clarity on this point above:
> Is this just a placeholder until we can wrap the C++ resource implementations from CCCL?

In hindsight, I think a simple "yes" would have sufficed 😄.

leofang: yes

leofang: Sorry I got myself confused...

cuda_core/cuda/core/experimental/_memory.py
```diff
@@ -19,6 +19,21 @@
...
+        import win32security
+
+        _PYWIN32_AVAILABLE = True
+    except ImportError:
+        import warnings
+
+        warnings.warn(
+            "pywin32 module not found. For better IPC support on Windows, " "install it with: pip install pywin32",
+            stacklevel=2,
+        )
```
kkraus14: How hard would it be to recreate the functionality we need from here to avoid this optional dependency?

leofang: I think we'll need to discuss our next move first and then clean this up based on the decision. Keenan struggled for quite a long time and it turns out the IPC support for mempools is not enabled on Windows at all. There is a number of doc and sample bugs we'll create later (the memory team created one).

keenan-simpson: I've created 2 bugs to address the documentation issues.

yeah the review is in limbo right now. I'm not sure if we should push the change for IPC mempools and have it only support linux for now, or if we should shelve it and add a cuMem and VMM API IPC memory_resource in the meantime. 

All of the Windows specific changes should be treated as draft. That optional dependency is from when I was debugging windows errors and operating under the assumption that I was improperly creating the windows security descriptor for the mempool handle. In hindsight it was a (temporary) lack of suport in driver layer as Leo said.

ksimpson-work:
I am tasked with adding the IPC mempool support for windows at the driver level so I think we should roll with this implementation and have it pick up windows once that change is integrated. Users who need IPC on windows for older CTKs can implement their own MR or use the cuMemCreate & VMM API bindings. WDYT

cuda_core/cuda/core/experimental/_memory.py
leofang: To circle back, adding `alignment` is now tracked in #634.

leofang:
IPC support for Linux landed (https://github.com/NVIDIA/cuda-python/pull/930) 🎉 We'll revisit (and close) this PR when there's a viable path emerging for Windows counterpart (and preferably not through VMM #968) .

PR #455, opened 2025-02-19
Draft for graph APIs

vzhurba01:


cuda_core/cuda/core/experimental/_graph.py
```diff
@@ -0,0 +1,786 @@
...
+                handle_return(driver.cuGraphDestroy(self.graph))
+            self.graph = None
+            self.conditional_graph = None
+
+    __slots__ = ("__weakref__", "_mnff", "_building_ended")
+
+    def __init__(self):
+        raise NotImplementedError(
+            "directly creating a Graph object can be ambiguous. Please either "
+            "call Device.create_graph_builder() or stream.creating_graph_builder()"
```
leofang:
```suggestion
            "call Device.create_graph_builder() or stream.create_graph_builder()"
```

cuda_core/cuda/core/experimental/_graph.py
```diff
@@ -0,0 +1,786 @@
...
+            if options.batch_mem_op_node_params:
+                flags |= driver.CUgraphDebugDot_flags.CU_GRAPH_DEBUG_DOT_FLAGS_BATCH_MEM_OP_NODE_PARAMS
+            if options.extra_topo_info:
+                flags |= driver.CUgraphDebugDot_flags.CU_GRAPH_DEBUG_DOT_FLAGS_EXTRA_TOPO_INFO
+            if options.conditional_node_params:
+                flags |= driver.CUgraphDebugDot_flags.CU_GRAPH_DEBUG_DOT_FLAGS_CONDITIONAL_NODE_PARAMS
+
+        handle_return(driver.cuGraphDebugDotPrint(self._mnff.graph, path, flags))
+
+    def split(self, count) -> Tuple[GraphBuilder, ...]:
```
leofang:
```suggestion
    def split(self, count: int) -> tuple[GraphBuilder, ...]:
```

cuda_core/cuda/core/experimental/_graph.py
```diff
@@ -0,0 +1,786 @@
...
+        *graph_builders : :obj:`~_graph.GraphBuilder`
+            The graph builders to join.
+
+        Returns
+        -------
+        graph_builder : :obj:`~_graph.GraphBuilder`
+            The newly joined graph builder.
+
+        """
+        if not all(isinstance(builder, GraphBuilder) for builder in graph_builders):
```
leofang: nit:
```suggestion
        if any(not isinstance(builder, GraphBuilder) for builder in graph_builders):
```

cuda_core/cuda/core/experimental/_graph.py
```diff
@@ -0,0 +1,786 @@
...
+            if not builder.is_join_required:
+                root_idx = i
+                break
+
+        # Join all onto the root builder
+        for idx, builder in enumerate(graph_builders):
+            if idx == root_idx:
+                continue
+            graph_builders[root_idx].stream.wait(builder.stream)
+            builder.close()
```
leofang: nit:
```suggestion
        # Join all onto the root builder
        root_bdr = graph_builders[root_idx]
        for idx, builder in enumerate(graph_builders):
            if idx == root_idx:
                continue
            root_bdr.stream.wait(builder.stream)
            builder.close()
```

cuda_core/cuda/core/experimental/_graph.py
```diff
@@ -0,0 +1,786 @@
...
+
+        return graph_builders[root_idx]
```
leofang: nit:
```suggestion
        return root_bdr
```

cuda_core/cuda/core/experimental/_graph.py
```diff
@@ -0,0 +1,786 @@
...
+
+        status, _, graph, _, _ = handle_return(driver.cuStreamGetCaptureInfo(self._mnff.stream.handle))
+        if status != driver.CUstreamCaptureStatus.CU_STREAM_CAPTURE_STATUS_ACTIVE:
+            raise RuntimeError("Cannot create a conditional handle when graph is not being built")
+
+        return int(
+            handle_return(
+                driver.cuGraphConditionalHandleCreate(graph, self._get_conditional_context(), default_value, flags)
+            )
+        )
```
leofang: Q: Do we have to erase the type of the handle? Can we instead keep the driver type around for as long as it can?

cuda_core/cuda/core/experimental/_graph.py
```diff
@@ -0,0 +1,786 @@
...
+            graph_builders[root_idx].stream.wait(builder.stream)
+            builder.close()
+
+        return graph_builders[root_idx]
+
+    def __cuda_stream__(self) -> Tuple[int, int]:
+        """Return an instance of a __cuda_stream__ protocol."""
+        return self.stream.__cuda_stream__()
+
+    def _get_conditional_context(self):
```
leofang: No need to address in this PR: Our stream already keeps a reference to the context from which it was created. We can just use it and not call any CUDA API.

cuda_core/cuda/core/experimental/_graph.py
```diff
@@ -0,0 +1,786 @@
...
+
+    global _py_major_minor, _driver_ver
+    # binding availability depends on cuda-python version
+    _py_major_minor = get_binding_version()
+    _driver_ver = handle_return(driver.cuDriverGetVersion())
+    _inited = True
+
+
+@dataclass
+class DebugPrintOptions:
```
leofang: I prefer to call this `GraphDebugPrintOptions` to make sure this is for CUDA graphs and not just something generic.

cuda_core/cuda/core/experimental/_graph.py
```diff
@@ -0,0 +1,786 @@
...
+    handles: bool = False
+    mem_alloc_node_params: bool = False
+    mem_free_node_params: bool = False
+    batch_mem_op_node_params: bool = False
+    extra_topo_info: bool = False
+    conditional_node_params: bool = False
+
+
+@dataclass
+class CompleteOptions:
```
leofang: ditto I prefer to call this `GraphCompleteOptions` or `GraphFinalizeOptions` to make sure this is for CUDA graphs and not just something generic.

cuda_core/cuda/core/experimental/_graph.py
```diff
@@ -0,0 +1,786 @@
...
+        """
+        if _driver_ver < 12080:
+            raise RuntimeError(f"Driver version {_driver_ver} does not support conditional if-else")
+        if _py_major_minor < (12, 8):
+            raise RuntimeError(f"Binding version {_py_major_minor} does not support conditional if-else")
+        node_params = driver.CUgraphNodeParams()
+        node_params.type = driver.CUgraphNodeType.CU_GRAPH_NODE_TYPE_CONDITIONAL
+        node_params.conditional.handle = handle
+        node_params.conditional.type = driver.CUgraphConditionalNodeType.CU_GRAPH_COND_TYPE_IF
+        node_params.conditional.size = 2
```
leofang, line 579: Nice 👍

cuda_core/cuda/core/experimental/_graph.py
```diff
@@ -0,0 +1,786 @@
...
+        raise ValueError("Child graph has not finished building.")
+
+    if not parent_graph.is_building:
+        raise ValueError("Parent graph is being built.")
+
+    status, _, graph_out, dependencies_out, num_dependencies_out = handle_return(
+        driver.cuStreamGetCaptureInfo(parent_graph.stream.handle)
+    )
+    if status != driver.CUstreamCaptureStatus.CU_STREAM_CAPTURE_STATUS_ACTIVE:
+        raise ValueError("Parent graph is not in an active capture state")
```
leofang: Q: Isn't this already checked in `if not parent_graph.is_building:` above?

cuda_core/tests/test_graph.py
```diff
@@ -0,0 +1,743 @@
...
+    launch_stream.sync()
+    assert arr[0] == 1
+
+    # Continue capturing to extend the graph
+    with pytest.raises(RuntimeError, match="^Cannot resume building after building has ended."):
+        gb.begin_building()
+
+    # Close the memroy resource now because the garbage collected might
+    # de-allocate it during the next graph builder process
+    b.close()
```
leofang: Between the two steps, can we re-run the graph for say 3 more times, and check `arr` is incremented correctly?

cuda_core/tests/test_graph.py
```diff
@@ -0,0 +1,743 @@
...
+        b.close()
+        pytest.skip("Driver does not support conditional handle")
+    launch(gb, LaunchConfig(grid=1, block=1), set_handle, handle, condition_value)
+
+    # # Add Node B (if condition)
+    gb_if = gb.if_cond(handle).begin_building()
+    launch(gb_if, LaunchConfig(grid=1, block=1), add_one, arr.ctypes.data)
+    gb_if_0, gb_if_1 = gb_if.split(2)
+    launch(gb_if_0, LaunchConfig(grid=1, block=1), add_one, arr.ctypes.data)
+    launch(gb_if_1, LaunchConfig(grid=1, block=1), add_one, arr.ctypes.data + 4)
```
leofang: Let's not compute pointer offset. This could be slower but safer.
```suggestion
    launch(gb_if_1, LaunchConfig(grid=1, block=1), add_one, arr[1:].ctypes.data)
```

cuda_core/tests/test_graph.py
```diff
@@ -0,0 +1,743 @@
...
+        gb.end_building()
+        b.close()
+        pytest.skip("Driver does not support conditional if-else")
+
+    ## IF nodes
+    gb_if = gb_if.begin_building()
+    launch(gb_if, LaunchConfig(grid=1, block=1), add_one, arr.ctypes.data)
+    gb_if_0, gb_if_1 = gb_if.split(2)
+    launch(gb_if_0, LaunchConfig(grid=1, block=1), add_one, arr.ctypes.data)
+    launch(gb_if_1, LaunchConfig(grid=1, block=1), add_one, arr.ctypes.data + 4)
```
leofang: ditto

cuda_core/tests/test_graph.py
```diff
@@ -0,0 +1,743 @@
...
+    launch(gb_case[0], LaunchConfig(grid=1, block=1), add_one, arr.ctypes.data)
+    launch(gb_case[0], LaunchConfig(grid=1, block=1), add_one, arr.ctypes.data)
+    gb_case[0].end_building()
+
+    ## Case 1
+    gb_case[1].begin_building()
+    launch(gb_case[1], LaunchConfig(grid=1, block=1), add_one, arr.ctypes.data + 4)
+    gb_case_1_left, gb_case_1_right = gb_case[1].split(2)
+    launch(gb_case_1_left, LaunchConfig(grid=1, block=1), add_one, arr.ctypes.data + 4)
+    launch(gb_case_1_right, LaunchConfig(grid=1, block=1), add_one, arr.ctypes.data + 8)
```
leofang: ditto (no pointer offset)

cuda_core/cuda/core/experimental/_graph.py
```diff
@@ -0,0 +1,786 @@
...
+        Parameters
+        ----------
+        stream : :obj:`~_stream.Stream`
+            The stream in which to launch the graph
+
+        """
+        handle_return(driver.cuGraphLaunch(self._mnff.graph, stream.handle))
+
+
+def launch_graph(parent_graph: GraphBuilder, child_graph: GraphBuilder):
```
leofang: I thought `launch_graph()` was meant to be a standalone `launch()` function but for graphs. I dunno how I got that impression (most likely I misunderstood it entirely during offline discussion...), but now that I've reviewed everything and have a clear mental model, I wonder if we can just rename this function to a `GraphBuilder.add_child()` method? Seems more logical to me to treat the first argument as `self` 🤔 It's also less confusing with the `launch()` function. WDYT?

leofang (commented):
LGTM overall, thanks a lot @vzhurba01 for solid works in design/implementation/testing! I left a few hopefully quick questions mainly on naming of a few things.

cuda_core/cuda/core/experimental/_graph.py
vzhurba01: Done

cuda_core/cuda/core/experimental/_graph.py
vzhurba01: done

cuda_core/cuda/core/experimental/_graph.py
vzhurba01: Done

cuda_core/cuda/core/experimental/_graph.py
vzhurba01: Done

cuda_core/cuda/core/experimental/_graph.py
vzhurba01: done

cuda_core/tests/test_graph.py
vzhurba01: Done

cuda_core/tests/test_graph.py
vzhurba01: Done

cuda_core/tests/test_graph.py
vzhurba01: Done

cuda_core/cuda/core/experimental/_graph.py
vzhurba01: Yeah, I've now removed it. Giving errors with "active capture state" doesn't make sense either since we've never talked about what a "capture" is.

cuda_core/cuda/core/experimental/_graph.py
vzhurba01: Done

cuda_core/cuda/core/experimental/_graph.py
vzhurba01: Done

cuda_core/cuda/core/experimental/_graph.py
vzhurba01: IIUC `cuda.core` tries to avoid directly giving objects from `cuda.bindings`, and would instead try to wrap them a `cuda.core` object. The header uses it as a simple int (`typedef cuuint64_t CUgraphConditionalHandle;`), so it felt a bit heavy handed to me to wrap it, but maybe it's not a big deal doing so (or even just returning the `cuda.bindings` extension class directly).

leofang: I guess it's because it can be taken as a kernel argument and we need its address. Let me know if the reasoning is correct. I can work on a patch to address this in the kernel arg handler (move `int(handle)` to the internal implementation of `launch()`).

vzhurba01: Yeah. The conditional methods in there file too would need to be updated since right now they're set to `handle: int`.

cuda_core/tests/test_graph.py
vzhurba01: Done

cuda_core/cuda/core/experimental/_graph.py
vzhurba01: Yeah I like the `GraphBuilder.add_child()` a lot more.

leofang: Changed in commit 668a8ce9.

cuda_core/cuda/core/experimental/_graph.py
leofang: Changed in commit 812bcbe7. However, I am reconsidering what you said above regarding `CUgraphConditionalHandle` being a simple typedef... Let me sleep over this.

cuda_core/cuda/core/experimental/_graph.py
leofang: Changed in commit d9c3526f.

cuda_core/cuda/core/experimental/_graph.py
leofang: I am marking this thread as resolved. This is my thought process.
-  At the level that we operate, we need something explicit (let's call it "token" for now) that we can pass around -- in particular, between host/device. It does not seem possible to hide it entirely.
- For users who write a CUDA kernel expecting `CUgraphConditionalHandle`, they want to know how to cast the token to it (ex: `token.kernel_arg()`).
- We can have a `class Token:` or even `cdef class Token:` holding a `CUgraphConditionalHandle` internally, so `token = Token()`, and provide a type caster for kernel launch purposes. It should be a typed class in case `CUgraphConditionalHandle` has the underlying C type changed in the future (say from `uint64_t` to a POD struct).  
- With all things considered, our `cdef class CUgraphConditionalHandle` in the Python layer already meets everything that we require for this hypothetical `Token` class. So instead of calling it `Token` we accept using it as-is.

cuda_core/cuda/core/experimental/_graph.py
```diff
@@ -661,6 +654,45 @@ def close(self):
...
+            raise NotImplementedError(
+                f"Launching child graphs is not implemented for versions older than CUDA 12."
+                f"Found driver version is {_driver_ver} and binding version is {_py_major_minor}"
+            )
+
+        if not child_graph._building_ended:
+            raise ValueError("Child graph has not finished building.")
+
+        if not self.is_building:
+            raise ValueError("Parent graph is being built.")
```
vzhurba01: Nit: "Parent graph is not being built."?

leofang:
```suggestion
            raise ValueError("Parent graph is not being built.")
```

leofang:
Thanks, Vlad! Since the CI was green and the last change is string-only, let me admin-merge it.

PR #457, opened 2025-02-20
[DO NOT MERGE] check if MSVC pre installed in the VM

leofang:


leofang:
Answer is no. So we can't run Cython tests...

Issue #439, opened 2025-02-07
Querying current device is slow compared to CuPy


shwina:
Getting the current device using `cuda.core` is quite a bit slower than CuPy:

```python
In [1]: import cupy as cp

In [2]: %timeit cp.cuda.Device()
69 ns ± 0.496 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [3]: from cuda.core.experimental import Device

In [4]: %timeit Device()
795 ns ± 0.273 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

Ultimately, my goal is to get the compute capability of the current device, and this is even slower:

```python
In [5]: %timeit cp.cuda.Device().compute_capability
89.1 ns ± 0.413 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [6]: %timeit Device().compute_capability
2.64 μs ± 122 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

Are there tricks (e.g., caching) CuPy is employing here that `cuda.core` can use as well? Alternately, is there another way for me to use `cuda.core` or `cuda.bindings` to get this information quickly? Note that for my use case, I'm not super concerned about the _first_ call to `Device()`, but I do want _subsequent_ calls to be trivially inexpensive if the current device hasn't changed.

---

Using the low-level cuda.bindings is also not quite as fast:

```python
In [11]: def get_cc():
    ...:     dev = runtime.cudaGetDevice()[1]
    ...:     return driver.cuDeviceComputeCapability(dev)
    ...:

In [12]: get_cc()
Out[12]: (<CUresult.CUDA_SUCCESS: 0>, 7, 5)

In [13]: %timeit get_cc()
597 ns ± 0.494 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

leofang:
We'll have to cache CC on a per-`Device` object level to bring this down to O(10) ns level.
```python
In [32]: def get_cc(dev):
    ...:    if dev in data:
    ...:        return data[dev]
    ...:    data[dev] = (driver.cuDeviceGetAttribute(driver.CUdevice_attribute.CU_DEVICE_ATTRIBUTE_COMPUTE_CAPABILITY_MAJOR, dev)[1],
[...]

PR #459, opened 2025-02-21
Improve perf of accessing `dev.compute_capability`

leofang:
Part of #439.

Before this PR: 
```python
In [4]: %timeit dev.compute_capability
1.87 μs ± 2.34 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

With this PR: 
```python
In [7]: %timeit dev.compute_capability
97.2 ns ± 0.087 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```
which I consider good enough for a pure Python implementation. Compared to the CuPy counterpart (which is Cython-based and returning a string instead of namedtuple):
```python
In [12]: %timeit dev.compute_capability
41.6 ns ± 0.0415 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

Note that the perf improvement of retrieving a `Device()` instance is out of scope of this PR and pending investigation (https://github.com/NVIDIA/cuda-python/issues/439#issuecomment-2673234572).

As part of this PR, I also removed a silly lock in the `Device` constructor. The data being protected is already placed in the thread-local storage, so it makes no sense to add another lock.

### DIFF
--- a/cuda_core/cuda/core/experimental/_program.py
+++ b/cuda_core/cuda/core/experimental/_program.py
@@ -5,6 +5,7 @@
 import weakref
 from dataclasses import dataclass
 from typing import List, Optional, Tuple, Union
+from warnings import warn
 
 from cuda.core.experimental._device import Device
 from cuda.core.experimental._linker import Linker, LinkerOptions
@@ -12,6 +13,7 @@ from cuda.core.experimental._module import ObjectCode
 from cuda.core.experimental._utils import (
     _handle_boolean_option,
     check_or_create_options,
+    driver,
     handle_return,
     is_nested_sequence,
     is_sequence,
@@ -378,6 +380,7 @@ class Program:
                 raise TypeError("c++ Program expects code argument to be a string")
             # TODO: support pre-loaded headers & include names
             # TODO: allow tuples once NVIDIA/cuda-python#72 is resolved
+
             self._mnff.handle = handle_return(nvrtc.nvrtcCreateProgram(code.encode(), b"", 0, [], []))
             self._backend = "nvrtc"
             self._linker = None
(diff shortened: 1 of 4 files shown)


rwgk:
I looked here while working on the release notes. (I looked at the code before while working on #439 a couple days ago, but didn't realize then that the code is so new.)

From the PR description:

> As part of this PR, I also removed a silly lock in the Device constructor. The data being protected is already placed in the thread-local storage, so it makes no sense to add another lock.

I'm almost certain that there can be a race now:

https://chatgpt.com/share/67d4bc2f-6c7c-8008-9d86-425fe77a3ed9

leofang:
@rwgk how about reverting 95777c478ca02d2c94f5e56aced9254872038fbf?

leofang:
I don't get it. This is thread local storage, not normal Python object, why do we need any lock?

rwgk:
I can try a fwd fix later tonight. Should be relatively easy.

Multiple threads can get to the for loop.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

leofang:
Let us not rush into a fix and check CPython threading.local internal first. I believe the impl already has a lock so that Python level access is guaranteed thread safe.

rwgk:
I only have a minute right now, hoping for a piece of information that will help me when I have a block of time later:

* I have to read up on `threading.local`.
* But if I'm guessing correctly, a difference before/after this PR is that `_tls.devices` is now recomputed for each thread, while it was computed only once before. Does that sound right?
* Assuming my guess is correct, is that what you wanted, or just something you accepted?

leofang:
Yes, each thread has its own `_tls` and by definition (thread local storage) and by CPython implementation accessing `_tls` will not race. `_tls` should not be considered as normal Python module-level object.

rwgk:
See #520: no race, but it recomputes `_tls.devices` for each new thread.

leofang:
Right, this is the consequence of storing data in thread-local storage. It was already the case before this PR, and this is why having a lock makes no sense. Each thread always has its own copy of `_tls.devices`.

rwgk:
Do we want that behavior? Or would it be better if `devices` was computed only once per process?

leofang:
Yes. For example, `cudaGetDevice`/`cudaSetDevice` are already accessing per-thread-level information:
https://developer.nvidia.com/blog/cuda-pro-tip-always-set-current-device-avoid-multithreading-bugs/

rwgk:
Wow, thanks! I need to read up. I completely misinterpreted what the loop is supposed to do.

leofang:
No worries! Always good to have extra pairs of eyes 🙂

PR #460, opened 2025-02-21
Switch to use CUDA driver APIs in `Device` constructor

leofang:
Before this PR:
```python
In [3]: %timeit Device()
660 ns ± 2.01 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [4]: %timeit Device(0)
644 ns ± 2.05 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```
With this PR:
```python
In [3]: %timeit Device()
396 ns ± 1.78 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [4]: %timeit Device(0)
165 ns ± 0.983 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```
(Bindings are built from the main branch.)


cuda_core/cuda/core/experimental/_device.py
```diff
@@ -957,34 +957,42 @@ def __new__(cls, device_id=None):
 
         # important: creating a Device instance does not initialize the GPU!
         if device_id is None:
-            device_id = handle_return(runtime.cudaGetDevice())
-            assert_type(device_id, int)
+            err, dev = driver.cuCtxGetDevice()
+            if err == 0:
+                device_id = int(dev)
+            else:
+                ctx = handle_return(driver.cuCtxGetCurrent())
```
keenan-simpson, line 980: what's going on here? Is there some requirement from cudart which requires CtxGetCurrent() to be called before the device can be queried?

cuda_core/cuda/core/experimental/_device.py
```diff
@@ -957,34 +957,42 @@ def __new__(cls, device_id=None):
...
+                assert int(ctx) == 0
+                device_id = 0  # cudart behavior
+            assert isinstance(device_id, int), f"{device_id=}"
```
rwgk: Is this intentionally not using the
```
             assert_type(device_id, int)
```
helper?

cuda_core/cuda/core/experimental/_device.py
rwgk: Could it be helpful to raise a more specific error here?

It might be helpful to add a comment right after the else:

```python
            else:
                # Emulate cudart behavior
                err, ctx = driver.cuCtxGetCurrent()
                if err != 0:
                    raise <Informative Error, what we really want is the current device (not primarily current context)>
```

leofang: This handles the case that no context is set to current. The logic is as follows:
- if a context is set to current, we can easily get the device ID associated with the current context
- if no context is set to current (which can happen right after `cuInit(0)` and before anything else is called), we confirm it is the case by checking `ctx` pointer is zero (`err` will always succeed), and then pick device 0

cuda_core/cuda/core/experimental/_device.py
leofang: I think this PR predates the helper, I'll add it

leofang: Unfortunately `assert_type` adds 50 ns on my machine. Since we're in a hotpot now (#439, #658) and by construct we always have an `int` here, I removed `assert instance(...)` which saves about 15 ns.

leofang:
This is ready.

cuda_core/cuda/core/experimental/_device.py
```diff
@@ -957,26 +957,34 @@ def __new__(cls, device_id=None):
...
+                device_id = int(dev)
+            else:
+                ctx = handle_return(driver.cuCtxGetCurrent())
+                assert int(ctx) == 0
+                device_id = 0  # cudart behavior
         else:
-            total = handle_return(runtime.cudaGetDeviceCount())
-            assert_type(device_id, int)
-            if not (0 <= device_id < total):
+            total = handle_return(driver.cuDeviceGetCount())
```
rwgk: Assuming that the happy path is common, this `driver.cuDeviceGetCount()` call seems redundant.

Also assuming it's not actually worth the cycles checking for `isinstance`, we could replace the `else` block here with:

```python
        elif device_id < 0:
            raise ValueError(f"device_id must be >= 0, got {device_id!r}")
```

Then below (new line 998) we could do this:

```python
        try:
            return devices[device_id]
        except IndexError:
            raise ValueError(f"device_id must be within [0, {len(devices)}), got {device_id!r}")
```

WDYT?

rwgk (approved):

cuda_core/cuda/core/experimental/_device.py
```diff
@@ -957,26 +957,34 @@ def __new__(cls, device_id=None):
 
         # important: creating a Device instance does not initialize the GPU!
         if device_id is None:
-            device_id = handle_return(runtime.cudaGetDevice())
-            assert_type(device_id, int)
+            err, dev = driver.cuCtxGetDevice()
+            if err == 0:
+                device_id = int(dev)
+            else:
+                ctx = handle_return(driver.cuCtxGetCurrent())
```
kkraus14: Is there a specific error code or set of error codes we should be handling here? If the above `driver.cuCtxGetDevice()` call returns an error that we don't expect we should probably raise it as an exception instead of it propagating to the `driver.cuCtxGetCurrent()` call?

cuda_core/cuda/core/experimental/_device.py
```diff
@@ -957,26 +957,34 @@ def __new__(cls, device_id=None):
...
                 raise ValueError(f"device_id must be within [0, {total}), got {device_id}")
 
         # ensure Device is singleton
-        if not hasattr(_tls, "devices"):
-            total = handle_return(runtime.cudaGetDeviceCount())
-            _tls.devices = []
+        try:
+            devices = _tls.devices
+        except AttributeError:
+            total = handle_return(driver.cuDeviceGetCount())
```
kkraus14, line 992: I think we can reuse total that was already calculated above?

cuda_core/cuda/core/experimental/_device.py
```diff
@@ -957,26 +957,34 @@ def __new__(cls, device_id=None):
...
+            devices = _tls.devices = []
             for dev_id in range(total):
```
kkraus14, line 994: If someone tries to create a Device with a specific ID, why do we need to initialize all of the devices at that point? Instead of calling `driver.cuDeviceGetCount()` to get the number of devices to initialize all of `_tls.devices`, could we use `driver.cuDeviceGet` and lazily populate `_tls.devices` as devices are created?

rwgk: We had discussions about this code back in March. @leofang [wrote here](https://github.com/NVIDIA/cuda-python/pull/459#issuecomment-2726176716):

> Each thread always has its own copy of `_tls.devices`.

I'm still unclear though TBH: Is that why we cannot lazily populate?

cuda_core/cuda/core/experimental/_device.py
leofang: I think this is an excellent idea, done in commit 6b245ff17. It also optimized the case with an explicit device ID by almost 4x, see the updated PR description.

cuda_core/cuda/core/experimental/_device.py
leofang: Each thread has a private storage of an N-list (N is the number of devices seen in the process), and currently the list is initialized at once instead of lazily. Since this is a one-time only initialization we're discussing here, I prefer to pay this cost upfront, to reduce the possibility of perf jittering in a multithreading code. Given the ongoing perf investigation (#677) we might end up dropping all of these later. So let's keep it simple for now.

cuda_core/cuda/core/experimental/_device.py
leofang: After applying Ralf's suggestion (https://github.com/NVIDIA/cuda-python/pull/460#discussion_r2112905363), this is moot now. We only call `cuDeviceGetCount` once to create the TLS.

cuda_core/cuda/core/experimental/_device.py
leofang: Invalid context (201) is now checked in commit https://github.com/NVIDIA/cuda-python/commit/6b245ff1784004816c32429ccaf6517218e03ace

cuda_core/cuda/core/experimental/_device.py
```diff
@@ -968,26 +969,34 @@ def __new__(cls, device_id=None):
...
-            assert_type(device_id, int)
-        else:
-            total = handle_return(runtime.cudaGetDeviceCount())
-            assert_type(device_id, int)
-            if not (0 <= device_id < total):
-                raise ValueError(f"device_id must be within [0, {total}), got {device_id}")
+            err, dev = driver.cuCtxGetDevice()
+            if err == 0:
+                device_id = int(dev)
+            elif err == 201:  # CUDA_ERROR_INVALID_CONTEXT
```
kkraus14: Can we use the enum value from `driver` here instead of 201?

cuda_core/cuda/core/experimental/_device.py
```diff
@@ -959,7 +960,7 @@ class Device:
 
     __slots__ = ("_id", "_mr", "_has_inited", "_properties")
 
-    def __new__(cls, device_id=None):
+    def __new__(cls, device_id: int = None):
```
kkraus14: I think this should be `Optional[int]`?

cuda_core/cuda/core/experimental/_device.py
```diff
@@ -968,26 +969,34 @@ def __new__(cls, device_id=None):
...
         if device_id is None:
-            device_id = handle_return(runtime.cudaGetDevice())
-            assert_type(device_id, int)
-        else:
-            total = handle_return(runtime.cudaGetDeviceCount())
-            assert_type(device_id, int)
-            if not (0 <= device_id < total):
-                raise ValueError(f"device_id must be within [0, {total}), got {device_id}")
+            err, dev = driver.cuCtxGetDevice()
+            if err == 0:
```
kkraus14: +1 would be good to use `driver.CUResult.CUDA_SUCCESS`?

leofang:
Blocked by #687.

cuda_core/cuda/core/experimental/_device.py
leofang: Good idea, I dunno what I was thinking, done in commit 708fd70f

cuda_core/cuda/core/experimental/_device.py
leofang: fixed in commit 708fd70f

cuda_core/cuda/core/experimental/_device.py
leofang: done in commit 708fd70f

leofang:
Thanks, Ralf/Keith!

PR #461, opened 2025-02-21
WIP: CI: Also test WDDM mode in Windows GPU runners

leofang:


leofang:
We need a vGPU driver for this...

PR #435, opened 2025-02-04
Expose `ObjectCode` as public API + prune unnecessary input arguments

ksimpson-work:
close #194 
close #448 
close #450

PR #463, opened 2025-02-22
Various handle-related changes and improvements

### LINKED ISSUE #370: Retain the underlying object types when retrieving the `.handle` attribute
Today we always return a Python `int` and erase the type information, e.g.
https://github.com/NVIDIA/cuda-python/blob/61ef2249a1a49684a4995206b2b9983fce211e35/cuda_core/cuda/core/experimental/_stream.py#L140-L143
However, it could be possible that users want to retain the underlying `cuda.binding` type so that it can be consumed in, say, Cython:
```cython
from cuda.core.experimental import Device

from cuda.bindings.runtime cimport cudaStream_t

s = Device().create_stream()

cdef cudaStream_t s_cy = s.handle
```
We should keep `.handle`'s type, ~~and add `.ptr` that returns the type-erased pointer address as Python int~~.

### LINKED ISSUE #433: revise the handling of backend and handle properties in Program
Program is either backed by nvrtc (cuda-bindings), or the Linker object (object oriented binding abstraction).

Come up with, and implement a scheme for the Program.handle and Program.backend slots whose return values are currently ill defined and vary depending on the Program's backend.

see thread https://github.com/NVIDIA/cuda-python/pull/317#discussion_r1939927881

leofang:
Close #370. Close #433.

- Ensure `Program`/`Linker` have consistent `.backend` and `.handle`
   - This allows us to add e.g. `CXXProgram`, `PTXProgram`, ... in the future without breaking changes (https://github.com/NVIDIA/cuda-python/pull/435#issuecomment-2666959958) 
- Add `get_cuda_native_handle()` to fetch the underlying CUDA C object from the corresponding `cuda.bindings` Python-layer object
   - Recall that for all generated bindings there are 3 layers: Python, Cython, and internal. This refers to the Python layer. The Cython layer provides the C type information (i.e. think of it as a CUDA header)
   - Steps:
      1. Take the Python wrapper object from `<some_cuda_core_object>.handle`
      2. Pass it to `get_cuda_native_handle()`, and the underlying C object is returned (no refcount or lifetime management)
   - Since `cuda.core` is not yet cythonized, this is currently the best solution for Python/Cython/C++ interoperability. Once we lower to Cython we can also apply a similar treatment to all `cuda.core` objects.
      - This includes implementing ref counting via `shared_ptr` or other C++ machinery to keep track of lifetime across C++ and Python. We are not there yet.
      - Additional work on `cuda.bindings` is also needed (#464)
   - For reviewers of this part: Please focus on these two files (the rest of Cython-related files are direct copy/paste from our existing `cuda.bindings` infrastructure):
      - `tests/cython/test_get_cuda_native_handle.pyx` 
      - `cuda/core/experimental/include/utility.hpp`
- some minor doc build fixes


cuda_core/cuda/core/experimental/include/utility.hpp
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. ALL RIGHTS RESERVED.
+//
+// SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
+
+#pragma once
+
+#include <type_traits>
+
+// In cuda.bindings 12.8, the private member name was renamed from "_ptr" to "_pvt_ptr".
```
leofang, line 9: @vzhurba01 do you recall what prompted this change when updating to 12.8?

cuda_core/cuda/core/experimental/include/utility.hpp
```diff
@@ -0,0 +1,23 @@
...
+// We want to have the C++ layer supporting all past 12.x versions, so some tricks are needed.
+// Since there's no std::has_member<T, member_name> so we use SFINAE to create the same effect.
+
+template <typename T,
+          std::enable_if_t<std::is_pointer_v<decltype(std::remove_pointer_t<T>::_pvt_ptr)>, bool> = true>
```
rwgk: Super tiny suggestion:

`if` at the beginning and `bool> = true>` at the end always throws me off a bit. `int> = 0>` is more neutral.

.github/workflows/test-wheel-linux.yml
```diff
@@ -221,6 +230,21 @@ jobs:
...
+          # with the test requirements.
+          if [[ "${{ inputs.local-ctk }}" == 1 && "SKIP_CUDA_CORE_CYTHON_TEST" == 0 ]]; then
+            pip install cython
+            if [[ "${{ inputs.host-platform }}" == linux* ]]; then
+              bash tests/cython/build_tests.sh
+            elif [[ "${{ inputs.host-platform }}" == win* ]]; then
+              # TODO: enable this once win-64 runners are up
+              exit 1
+            fi
+            pytest -rxXs tests/cython
```
rwgk: Here and for the existing pytest calls: Add `-v`? (I often find that very useful.)

cuda_core/cuda/core/experimental/_event.py
```diff
@@ -1,13 +1,18 @@
...
+
 import weakref
 from dataclasses import dataclass
-from typing import Optional
+from typing import TYPE_CHECKING, Optional
 
 from cuda.core.experimental._utils import CUDAError, check_or_create_options, driver, handle_return
 
+if TYPE_CHECKING:
+    import cuda.bindings
```
rwgk, line 14: Would it be a disadvantage to have this import unconditionally?

It'll probably be less surprising unconditionally. (I found myself surprised by such details when debugging complex failures.)

cuda_core/cuda/core/experimental/_linker.py
```diff
@@ -323,6 +328,10 @@ def _exception_manager(self):
         raise e
 
 
+nvJitLinkHandleT = int
+LinkerHandleT = Union[nvJitLinkHandleT, "cuda.bindings.driver.CUlinkState"]
```
rwgk, line 332: (With unconditional import this wouldn't have to be a string here.)

cuda_core/cuda/core/experimental/_program.py
```diff
@@ -489,15 +497,15 @@ def compile(self, target_type, name_expressions=(), logs=None):
 
             return ObjectCode._init(data, target_type, symbol_mapping=symbol_mapping)
 
-        if self._backend == "linker":
+        if self._backend in ("nvJitLink", "driver"):
             return self._linker.link(target_type)
```
rwgk, line 502: It's a question for the existing code, too:

Is it OK if `None` is returned here?

If yes, I'd make that explicit (`return None`), maybe with a short comment to explain why that's OK. Maybe also mention under `Returns` in the docstring above.

cuda_core/cuda/core/experimental/include/utility.hpp
rwgk: We're using a private member? Do we own the code with the private member?

rwgk (approved):
I only have a few minor comments. I didn't see anything troubling, except the `raise` vs `pass` question maybe.

rwgk:
> I only have a few minor comments. I didn't see anything troubling, except the `raise` vs `pass` question maybe.

Oh wait, did the `except ImportError: raise` disappear in the meantime? Please disregard.

cuda_core/cuda/core/experimental/include/utility.hpp
leofang: Yes we own everything, they are autogen'd, for example for `CUstream`:
https://github.com/NVIDIA/cuda-python/blob/02a41789641dec6ab5fa4c07ed12ad8600aa5be4/cuda_bindings/cuda/bindings/driver.pxd.in#L185-L197
hence I feel SFINAE over all possibilities is OK. I think if we accidentally change the codegen in the future, the added Cython test should be able to catch it.

cuda_core/tests/cython/test_cython.py
```diff
@@ -0,0 +1,39 @@
...
+        # wrap the callable in a plain Python function
+        # and set the result as an attribute of this module.
+        mod = importlib.import_module(mod)
+        for name in dir(mod):
+            item = getattr(mod, name)
+            if callable(item) and name.startswith("test_"):
+                item = py_func(item)
+                setattr(sys.modules[__name__], name, item)
+    except ImportError:
+        raise
```
rwgk, line 39: Here it is, I accidentally lost my original comment.

Did you mean to write `pass` here?

Also, it's safer to make the try-except scope as narrow as possible, e.g. (untested):

```
    try:
        # For each callable in `mod` with name `test_*`,
        # wrap the callable in a plain Python function
        # and set the result as an attribute of this module.
        mod = importlib.import_module(mod)
    except ImportError:
        pass
    else:
        for name in dir(mod):
            item = getattr(mod, name)
            if callable(item) and name.startswith("test_"):
                item = py_func(item)
                setattr(sys.modules[__name__], name, item)
```

If there is any tooling that might import this file for analysis: Moving the `for` loop into a function and using `if __name__ == "__main__":` could guard against surprises.

rwgk (approved):

cuda_core/cuda/core/experimental/_program.py
leofang: No, the 2 if-branches should cover all possibilities, and `compile()` cannot return None. Are you thinking about adding an `assert False` below?

rwgk: Maybe

```
        assert self._backend in ("nvJitLink", "driver")
        return self._linker.link(target_type)
```

?

I'd definitely do *something* to make it obvious that we cannot possibly get a `None` return. (Refactoring accidents happen.)

It could alternative be
```
        raise RuntimeError("This code path is meant to be unreachable.")
```

cuda_core/tests/cython/test_cython.py
leofang: This file is copied/pasted from cuda-bindings. I think Vlad (or @shwina? 🙂) was the original author:
https://shwina.github.io/cython-testing/
> Did you mean to write `pass` here?

I suspect the answer was yes (based on Ashwin's blog post) but perhaps it caused some issues, so it was changed to `raise`? I can see that this is a cheap way to ensure the listed modules must exist and get tested. But then the whole try-except can be removed (or refactored). Perhaps @vzhurba01 can confirm the intention?

In any case, since this is a copy/paste let's change both files in a separate PR.

cuda_core/cuda/core/experimental/include/utility.hpp
leofang: Out of curiosity, how is `int` more natural? With the level of complexity here I feel everything is unnatural 😂 

FWIW the original version I had was:
```c++
typename = std::enable_if_t<std::is_pointer_v<decltype(std::remove_pointer_t<T>::_pvt_ptr)>>>
```
which to me is actually more natural (no need to use a dummy `int`/`bool` type, no need to do a silly comparison, etc). But unfortunately it does not work with SFINAE when I added the 2nd definition... the compiler complained about re-definition 😢

.github/workflows/test-wheel-linux.yml
```diff
@@ -221,6 +230,21 @@ jobs:
           pushd ./cuda_core
           pip install -r "tests/requirements-cu${TEST_CUDA_MAJOR}.txt"
           pytest -rxXs tests/
+
+          # It is a bit convoluted to run the Cython tests against CTK wheels,
+          # so let's just skip them. Also, currently our CI always installs the
+          # latest bindings (from either major version). This is not compatible
+          # with the test requirements.
+          if [[ "${{ inputs.local-ctk }}" == 1 && "${SKIP_CUDA_CORE_CYTHON_TEST}" == 0 ]]; then
+            pip install cython
```
leofang: Every time I am stumbled by this bug when using Cython as a run-time dependency (outside of a build system): Since starting Python 3.12 distutils is removed, we need to add setuptools explicitly. Cython for some reason faced some pushback... (https://github.com/cython/cython/issues/5751, https://github.com/cython/cython/pull/5754)
```suggestion
            pip install cython setuptools
```

cuda_core/cuda/core/experimental/include/utility.hpp
rwgk: > here I feel everything is unnatural 😂

No doubt! 😂

I wrote "neutral".

Our options, and what's triggered in my mind:

* `bool> = true>` — "What's true here?"
* `bool> = false>` — "What's false here?"
* `int> = 0> — "Aha, it's just a dummy value we need to plug in here to make the compiler happy."

.github/workflows/test-wheel-linux.yml
leofang: fixed in bc9595b

cuda_core/cuda/core/experimental/_program.py
leofang: applied in commit 43654c4

cuda_core/cuda/core/experimental/include/utility.hpp
leofang: updated in commit 37a6850

cuda_core/cuda/core/experimental/_event.py
leofang: Indeed. Now we have two good reasons to avoid doing this, see https://github.com/NVIDIA/cuda-python/issues/468. Since we already use `TYPE_CHECKING` in other places of `cuda.core`, let's do a floor sweep in a separate PR.

cuda_core/cuda/core/experimental/_linker.py
leofang: (see https://github.com/NVIDIA/cuda-python/pull/463#discussion_r1967738849)

cuda_core/tests/cython/build_tests.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+SCRIPTPATH=$(dirname $(realpath "$0"))
+CPLUS_INCLUDE_PATH=$SCRIPTPATH/../../cuda/core/experimental/include:$CUDA_HOME/include:$CPLUS_INCLUDE_PATH cythonize -3 -i $(dirname "$0")/test_*.pyx
```
leofang, line 4: Note: This approach left a bug uncaught (fixed in commit 37a6850): Ideally, we want to have a cmdline functionality like PyBind11 that returns the header path, something like `python -m cuda.core --include-path`. Here, we use the header from the local clone of this repo (hence using `$SCRIPTPATH`) instead of the installed cuda.core package, meaning we do not check if the header is actually packaged (and it was not before the fix). However, I think the said functionality needs to be done in a separate PR...

leofang: Example that I added to cuQuantum: https://docs.nvidia.com/cuda/cuquantum/latest/python/api/cuquantum.html#command-line-support

rwgk (dismissed):

cuda_core/cuda/core/experimental/_program.py
```diff
@@ -489,15 +497,15 @@ def compile(self, target_type, name_expressions=(), logs=None):
...
 
     @property
-    def backend(self):
-        """Return the backend type string associated with this program."""
+    def backend(self) -> str:
+        """Return this Program instance's underlying backend."""
         return self._backend
 
     @property
-    def handle(self):
```
keenan-simpson, line 501: could we add a little note about what the handle represents depending on the backend?

leofang: good idea, done in commit c96f6de

.github/workflows/test-wheel-linux.yml
leofang: I have slight preference over not cluttering the CI logs.

rwgk: In cases where we cannot easily reproduce locally, this can make the difference between day and night. (I've been there several times in my career.)

Extra log output is easy to ignore.

And often impossible to guess.

leofang: OK, added in commit e4cde77

rwgk (approved):

PR #471, opened 2025-02-25
Improvements to some cuda bindings example

oleksandr-pavlyk:
Modified `cuda_bindings/examples/0_Introduction/vectorAddDrv_test.py` to use `randn(N)` instead of `randn(size)`. Also renamed `size` to `nbytes` to readability.

Modified `clock_nvrtc_test.py` to improve readability.

### DIFF
--- a/cuda_bindings/examples/0_Introduction/clock_nvrtc_test.py
+++ b/cuda_bindings/examples/0_Introduction/clock_nvrtc_test.py
@@ -58,6 +58,10 @@ NUM_BLOCKS = 64
 NUM_THREADS = 256
 
 
+def elems_to_bytes(nelems, dt):
+    return nelems * np.dtype(dt).itemsize
+
+
 def main():
     print("CUDA Clock sample")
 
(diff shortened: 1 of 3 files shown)


ksimpson-work (dismissed):
lgtm. added vlad as reviewer. Thanks!

oleksandr-pavlyk:
`simpleCubmapTexture_test.py` just like it's original in cuda-samples does not pass the test when `num_layers` is other than 1. If this is always the case, the code can be further simplified, since in that case we always have `h_data_ref = -h_data`.

ksimpson-work (approved):

cuda_bindings/examples/0_Introduction/clock_nvrtc_test.py
```diff
@@ -58,6 +58,10 @@
 NUM_THREADS = 256
 
 
+def elems_to_bytes(nelems, dt):
+    return nelems * np.dtype(dt).itemsize
```
leofang, line 62: @seberg this has been an eternal mystery to me 😛 Can the NumPy types (not dtypes) carry an `itemsize` class attribute so that we can just do `dt.itemsize` without wrapping it with `np.dtype` first?

leofang (approved):

cuda_bindings/examples/0_Introduction/clock_nvrtc_test.py
seberg: The scalar type can't have it e.g. for strings/structs and it mixes the two concepts a bit more.  But mostly I think just nobody ever really thought about whether just allowing it may be convenient.
(maybe the more annoying thing is that there is no shorter spelling for the singleton/default `np.dtype(np.float32)` instance.)

Issue #189, opened 2024-10-25
Implement Context.__init__() in cuda.core.experimental._context


keenan-simpson:
implement the constructor for the Context object

leofang:
Right now our `Context` object is really just a non-functional wrapper holding the `CUcontext` pointer. I'd like to see what @vzhurba01 concludes for green contexts (#112), before committing to full CUDA context support. There might be a chance (best case scenario) that we simply don't need it.

leofang:
Push this out since it's not critical.

leofang:
This is technically a "won't do" because we will offer `Device.create_context()` instead (#1976). Closing.

PR #474, opened 2025-02-26
Device properties example

samaid:
This example aims to get familiar with cuda.core device and device properties concepts. The script iterates across all devices in the system and prints their device properties.
[log.txt](https://github.com/user-attachments/files/18994556/log.txt)


cuda_core/examples/device_properties.py
```diff
@@ -0,0 +1,169 @@
+# Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. ALL RIGHTS RESERVED.
+#
+# SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
+
+from cuda.core.experimental import Device, system
+
+
+def _bool_str(value: bool) -> str:
```
rwgk: `_yes_or_no` (or just `_yes_no` works, too)

makes this self-explanatory (in the call sites).

cuda_core/examples/device_properties.py
```diff
@@ -0,0 +1,169 @@
...
+    return "YES" if value else "NO"
+
+
+def _bytes_to_mbytes(value):
```
rwgk:
```
def _bytes_to_mbytes(value):
    return f"{value / (1024 ** 2):.2f}MB"

def _bytes_to_gbytes(value):
    return f"{value / (1024 ** 3):.2f}GB"
```

These were generated by ChatGPT: https://chatgpt.com/share/67bfde2c-7b20-8008-a099-302d0ed4dc60

cuda_core/examples/device_properties.py
```diff
@@ -0,0 +1,169 @@
...
+          {_bool_str(properties.read_only_host_register_supported)}")
+    print(f"- Amount of shared memory per block reserved by CUDA driver: \
+          {properties.reserved_shared_memory_per_block}B")
+    print(f"- Support for sparse CUDA arrays and sparse CUDA mipmapped arrays: \
+          {_bool_str(properties.sparse_cuda_array_supported)}")
+    print(f"- Using TCC driver: {_bool_str(properties.tcc_driver)}")
+    print(f"- Constant memory available: {properties.total_constant_memory}B")
+    print(f"- Support for unified address space with host: {_bool_str(properties.unified_addressing)}")
+    print(f"- Support for virtual memory management: {_bool_str(properties.virtual_memory_management_supported)}")
+
```
rwgk: Optional, but I'd lean towards adding a `print()` here, so that the output doesn't run directly into the command line prompt.

cuda_core/examples/device_properties.py
```diff
@@ -0,0 +1,169 @@
...
+ndev = system.num_devices
+print(f"Number of GPUs: {ndev}")
+
+for device_id in range(ndev):
+    device = Device(device_id)
+    print(f"DEVICE {device.name} (id={device_id})")
+
+    device.set_current()
+    ctx = device.context
```
rwgk: This seems to be unused. (Hm, not sure why ruff doesn't catch this.)

cuda_core/examples/device_properties.py
```diff
@@ -0,0 +1,169 @@
...
+          {properties.reserved_shared_memory_per_block}B")
+    print(f"- Support for sparse CUDA arrays and sparse CUDA mipmapped arrays: \
+          {_bool_str(properties.sparse_cuda_array_supported)}")
+    print(f"- Using TCC driver: {_bool_str(properties.tcc_driver)}")
+    print(f"- Constant memory available: {properties.total_constant_memory}B")
+    print(f"- Support for unified address space with host: {_bool_str(properties.unified_addressing)}")
+    print(f"- Support for virtual memory management: {_bool_str(properties.virtual_memory_management_supported)}")
+
+
+ndev = system.num_devices
```
rwgk: Could you please indent everything from here to the end?

```python
def show_device_properties():
    ... your code here ...


if __name__ == "__main__":
    assert len(sys.args) == 1, "no command-line arguments expected"
    show_device_properties()
```

The main difference is that tooling can import this without running anything other than the imports.

rwgk (commented):
What do you think about renaming to show_device_properties.py?

For quicker aha when people see the filename.

cuda_core/examples/device_properties.py
rwgk: > Hm, not sure why ruff doesn't catch this.

Oh! Because it's currently a module-level variable (which could in theory be used from elsewhere). Another reason to move this code into a function. I think ruff will then flag this.

cuda_core/examples/device_properties.py
samaid: Yep, will change both filename and function to `_yes_no`

cuda_core/examples/device_properties.py
samaid: No, I am an old schooler. I used Windows calculator to calculate magic numbers

samaid: I agree, it's more readable but my guts protest to use expensive power and division ops

cuda_core/examples/device_properties.py
samaid: Good point

cuda_core/examples/device_properties.py
samaid: I have initiated a separate discussion on `set_current()`. I wanted to add some meaningful example for it but it appears that context management is not yet fully implemented. Or am I wrong? I am happy to extend example with context stuff

cuda_core/examples/device_properties.py
samaid: Yes, will do. Thank you for great comments, Ralf @rwgk!

cuda_core/examples/device_properties.py
rwgk: I upvoted your suggestion (yesterday).

> it appears that context management is not yet fully implemented.

I had to look this up myself: https://github.com/NVIDIA/cuda-python/issues/189

What do you think about adding a comment (very rough, just to give the idea):

```
# Extend example to show device context information after #189 is resolved.
```

cuda_core/examples/device_properties.py
rwgk: (Today I started experimenting with claude.ai)

https://claude.ai/share/5830e427-4b3b-44f5-8bef-c47d10ebe01f

I'm slightly modifying what it suggests:

```python
BYTES_TO_GBYTES = 1 / (1024 ** 3)

def _bytes_to_gbytes(value):
    return f"{value * BYTES_TO_GBYTES:.2f}GB"
```

Making our examples readable seems valuable.

cuda_core/examples/device_properties.py
samaid: I double checked `log.txt`. The last few lines in the output will be:
```
...
- Support for unified address space with host: YES
- Support for virtual memory management: YES
*****************************************************

```
Two blank lines are printed before command line appears. See [line 169](https://github.com/NVIDIA/cuda-python/pull/474/files#diff-c3b856710c1de922c5df3763707ea750343f929e1e4cf66805a373f47f0cea64R169)

samaid:
@rwgk Your suggestions have been addressed

rwgk (approved):
Looks great to me!

We trigger CI testing via /ok to test comments. I think it works for anyone, but I just did that right now to get it going.

leofang:
> We trigger CI testing via /ok to test comments. I think it works for anyone,

It works for all GitHub IDs linked with a valid NVIDIA email address, so Sergey you should be able to trigger it yourself!

leofang:
It was a transient network issue, I reran the failed job and all is good now. Merging, thanks Sergey & Ralf!

PR #475, opened 2025-02-26
Add Linker example

### LINKED ISSUE #324: Add a `Linker` example
For major features we should ensure we offer code samples to teach people how to use them. We already have samples covering `Program`. We should add at least one sample to cover `Linker`.

vzhurba01:
close #324

### DIFF
--- a/cuda_core/examples/jit_lto_fractal.py
+++ b/cuda_core/examples/jit_lto_fractal.py
@@ -1,0 +1,274 @@
+# Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. ALL RIGHTS RESERVED.
+#
+# SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
+
+# ################################################################################
+#
+# This demo aims to illustrate a couple takeaways:
+#
+#   1. How to use the JIT LTO feature provided by the Linker class to link multiple objects together
+#   2. That linking allows for libraries to modify workflows dynamically at runtime
+#
+# This demo mimics a relationship between a library and a user. The user's sole responsability is to
+# provide device code that generates art. Where as the library is responsible for all steps involved in
+# setting up the device, launch configurations and arguments as well as linking the provided device code.
+#
+# Two algorithms are implemented:
+#   1. A Mandelbrot set
+#   2. A Julia set
+#
+# The user can choose which algorithm to use at runtime and generate the resulting image.
+#
+# ################################################################################
+
+import argparse
+import sys
+
+import cupy as cp
+
+from cuda.core.experimental import Device, LaunchConfig, Linker, LinkerOptions, Program, ProgramOptions, launch
+
+
+# ################################################################################
+#
+# This Mocklibrary is responsible for all steps involved launching the device code.
+#
+# The user is responsible for providing the device code that will be linked into the library's workflow.
+# The provided device code must contain a function with the signature `void generate_art(float* Data)`
+class MockLibrary:
+    def __init__(self):
+        # For this mock library, the main workflow is intentially kept simple by limiting itself to only calling the
+        # externally defined generate_art function. More involved libraries have the option of applying pre and post
+        # processing steps before calling user-defined device code. Conversely, these responsabilities can be reversed
+        # such that the library owns the bulk of the workflow while allowing users to provide customized pre/post
+        # processing steps.
+        code_main = r"""
+        extern __device__ void generate_art(float* Data);
+
+        extern "C"
+        __global__
+        void main_workflow(float* Data) {
+            // Preprocessing steps can be called here
+            // ...
+
+            // Call the user-defined device code
+            generate_art(Data);
+
+            // Postprocessing steps can be called here
+            // ...
+        }
+        """
+
+        # Most of the launch configurations can be preemptively done before the user provides their device code
+        # Therefore lets compile our main workflow device code now, and link the remaining pieces at a later time
+        self.arch = "".join(f"{i}" for i in Device().compute_capability)
+        self.program_options = ProgramOptions(std="c++11", arch=f"sm_{self.arch}", relocatable_device_code=True)
+        self.main_object_code = Program(code_main, "c++", options=self.program_options).compile("ptx")
+
+        # Setup device state
+        self.dev = Device()
+        self.dev.set_current()
+        self.stream = self.dev.create_stream()
+
+        # Setup buffer to store our results
+        self.width = 1024
+        self.height = 512
+        self.buffer = cp.empty(self.width * self.height * 4, dtype=cp.float32)
+
+        # Setup the launch configuration such that each thread will be generating one pixel, and subdivide
+        # the problem into 16x16 chunks.
+        self.grid = (self.width / 16, self.height / 16, 1)
+        self.block = (16, 16, 1)
+        self.config = LaunchConfig(grid=self.grid, block=self.block, stream=self.stream)
+
+    def link_and_run_code(self, user_code):
+        # First, user-defined code is compiled into a PTX object code
+        user_object_code = Program(user_code, "c++", options=self.program_options).compile("ptx")
+
+        # Then a Linker is created to link the main object code with the user-defined code
+        linker_options = LinkerOptions(arch=f"sm_{self.arch}")
+        linker = Linker(self.main_object_code, user_object_code, options=linker_options)
+        linked_code = linker.link("cubin")
+
+        # Now we're ready to retrieve the main device function and execute our library's workflow
+        kernel = linked_code.get_kernel("main_workflow")
+        launch(kernel, self.config, self.buffer.data.ptr)
+        self.stream.sync()
+
+        # Return the result buffer as a CUPY array
+        return self.buffer.get().astype(float).reshape(self.height, self.width, 4)
+
+
+# Now lets proceed with code from the user's perspective!
+#
+# ################################################################################
+
+# Simple implementation of Mandelbrot set from Wikipedia
+# http://en.wikipedia.org/wiki/Mandelbrot_set
+#
+# Note that this kernel is meant to be a simple, straight-forward
+# implementation, and so may not represent optimized GPU code.
+code_mandelbrot = r"""
+__device__
+void generate_art(float* Data) {
+    // Which pixel am I?
+    unsigned DataX = blockIdx.x * blockDim.x + threadIdx.x;
+    unsigned DataY = blockIdx.y * blockDim.y + threadIdx.y;
+    unsigned Width = gridDim.x * blockDim.x;
+    unsigned Height = gridDim.y * blockDim.y;
+
+    float R, G, B, A;
+
+    // Scale coordinates to (-2.5, 1) and (-1, 1)
+
+    float NormX = (float)DataX / (float)Width;
+    NormX *= 3.5f;
+    NormX -= 2.5f;
+
+    float NormY = (float)DataY / (float)Height;
+    NormY *= 2.0f;
+    NormY -= 1.0f;
+
+    float X0 = NormX;
+    float Y0 = NormY;
+
+    float X = 0.0f;
+    float Y = 0.0f;
+
+    unsigned Iter = 0;
+    unsigned MaxIter = 1000;
+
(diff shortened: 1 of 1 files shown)


vzhurba01:
Here's what `python examples/jit_lto_fractal.py -t all -d` displays:
hit_lto_fractal_image

cuda_core/examples/jit_lto_fractal.py
```diff
@@ -0,0 +1,274 @@
...
+
+            // Postprocessing steps can be called here
+            // ...
+        }
+        """
+
+        # Most of the launch configurations can be preemptively done before the user provides their device code
+        # Therefore lets compile our main workflow device code now, and link the remaining pieces at a later time
+        self.arch = "".join(f"{i}" for i in Device().compute_capability)
+        self.program_options = ProgramOptions(std="c++11", arch=f"sm_{self.arch}", relocatable_device_code=True)
```
keenan-simpson: this is nitty but arch is the current device arch by default. Also does the sample work without specifying c++11? I'm just thinking of ways to make it as clean and easy looking as possible for the demo

ksimpson-work (commented):
I left one little nit, but otherwise it looks good to me. Fractal art engages my monkey brain

cuda_core/examples/jit_lto_fractal.py
```diff
@@ -0,0 +1,274 @@
...
+if args.display:
+    fig = plt.figure()
+    for i, (image, title) in enumerate(result_to_display):
+        axs = fig.add_subplot(len(result_to_display), 1, i + 1)
+        axs.imshow(image)
+        axs.set_title(title)
+        axs.axis("off")
+    plt.show()
+
+print("done!")
```
keenan-simpson: could we add a programatic entry point so the testsuite can run this?

rwgk: +1

Coincidentally the same suggestion here](https://github.com/NVIDIA/cuda-python/pull/474#discussion_r1972801172).

Which made me realize that [that's the only way that we can get ruff to warn us about unused variables](https://github.com/NVIDIA/cuda-python/pull/474#discussion_r1972815147).

leofang:
Thanks, Vlad! I have a local patch that uses LTO-IR instead. I will push but I'd like to come up with a way of showing both approaches (linking all PTXs vs linking 1 PTX with LTO-IRs).

cuda_core/examples/jit_lto_fractal.py
leofang: done

leofang:
CI is green!

cuda_core/examples/jit_lto_fractal.py
```diff
@@ -0,0 +1,298 @@
...
+# ################################################################################
+#
+# This demo aims to illustrate a couple takeaways:
+#
+#   1. How to use the JIT LTO feature provided by the Linker class to link multiple objects together
+#   2. That linking allows for libraries to modify workflows dynamically at runtime
+#
+# This demo mimics a relationship between a library and a user. The user's sole responsability is to
+# provide device code that generates art. Where as the library is responsible for all steps involved in
+# setting up the device, launch configurations and arguments as well as linking the provided device code.
```
rwgk: responsability -> responsibility

... generates some art (?) (without the some it reads odd to me)

Where as -> whereas?

... arguments, as well ... (extra comma)

cuda_core/examples/jit_lto_fractal.py
```diff
@@ -0,0 +1,298 @@
...
+import sys
+
+import cupy as cp
+
+from cuda.core.experimental import Device, LaunchConfig, Linker, LinkerOptions, Program, ProgramOptions, launch
+
+
+# ################################################################################
+#
+# This Mocklibrary is responsible for all steps involved launching the device code.
```
rwgk, line 34: What do you think about

LinkAndLaunch

as the name for this class?

It's not really a library.

"mock library" is usually something like this:

https://docs.python.org/3/library/unittest.mock.html

cuda_core/examples/jit_lto_fractal.py
```diff
@@ -0,0 +1,298 @@
...
+#
+# The user is responsible for providing the device code that will be linked into the library's workflow.
+# The provided device code must contain a function with the signature `void generate_art(float* Data)`
+class MockLibrary:
+    def __init__(self):
+        # For this mock library, the main workflow is intentially kept simple by limiting itself to only calling the
+        # externally defined generate_art function. More involved libraries have the option of applying pre and post
+        # processing steps before calling user-defined device code. Conversely, these responsabilities can be reversed
+        # such that the library owns the bulk of the workflow while allowing users to provide customized pre/post
+        # processing steps.
```
rwgk: intentially -> intentionally

responsabilities -> responsibilities

cuda_core/examples/jit_lto_fractal.py
```diff
@@ -0,0 +1,298 @@
...
+
+        # Setup device state
+        self.dev = Device()
+        self.dev.set_current()
+        self.stream = self.dev.create_stream()
+
+        # Setup buffer to store our results
+        self.width = 1024
+        self.height = 512
+        self.buffer = cp.empty(self.width * self.height * 4, dtype=cp.float32)
```
rwgk, line 75: Might be nice to add a hint (comment) what the `4` is for. (R, G, B, A appears only much further down.)

cuda_core/examples/jit_lto_fractal.py
```diff
@@ -0,0 +1,298 @@
...
+
+        # Setup the launch configuration such that each thread will be generating one pixel, and subdivide
+        # the problem into 16x16 chunks.
+        self.grid = (self.width / 16, self.height / 16, 1)
```
rwgk: self.grid will contain two floats and one integer.

What if self.width or self.height are not multiples of 16? — Since this is an example, it'd be ideal to show best practices. This code is likely to get copied.

cuda_core/examples/jit_lto_fractal.py
```diff
@@ -0,0 +1,298 @@
...
+
+    def link(self, user_code, target_type):
+        if target_type == "ltoir":
+            program_options = ProgramOptions(std="c++11", arch=f"sm_{self.arch}", link_time_optimization=True)
+            linker_options = LinkerOptions(arch=f"sm_{self.arch}", link_time_optimization=True)
+        elif target_type == "ptx":
+            program_options = self.program_options
+            linker_options = LinkerOptions(arch=f"sm_{self.arch}")
+        else:
+            raise AssertionError
```
rwgk:
```
raise AssertionError(f"Invalid {target_type=}")
```

cuda_core/examples/jit_lto_fractal.py
```diff
@@ -0,0 +1,298 @@
...
+        linked_code = linker.link("cubin")
+
+        # Now we're ready to retrieve the main device function and execute our library's workflow
+        return linked_code.get_kernel("main_workflow")
+
+    def run(self, kernel):
+        launch(kernel, self.config, self.buffer.data.ptr)
+        self.stream.sync()
+
+        # Return the result buffer as a CuPy array
```
rwgk:
```
# Return the result as a numpy array (host array).
```
?

cuda_core/examples/jit_lto_fractal.py
```diff
@@ -0,0 +1,298 @@
...
+
+# Now lets proceed with code from the user's perspective!
+#
+# ################################################################################
+
+# Simple implementation of Mandelbrot set from Wikipedia
+# http://en.wikipedia.org/wiki/Mandelbrot_set
+#
+# Note that this kernel is meant to be a simple, straight-forward
+# implementation, and so may not represent optimized GPU code.
```
rwgk:
```
# implementation. No attempt is made to optimize this GPU code.
```
?

cuda_core/examples/jit_lto_fractal.py
```diff
@@ -0,0 +1,298 @@
...
+
+    unsigned ColorG = Iter % 50;
+    unsigned ColorB = Iter % 25;
+
+    R = 0.0f;
+    G = (float)ColorG / 50.0f;
+    B = (float)ColorB / 25.0f;
+    A = 1.0f;
+
+    Data[DataY*Width*4+DataX*4+0] = R;
```
rwgk: This would look more readable with:
```
unsigned i = DataY*Width*4+DataX*4;
Data[i + 0] = R;
...
```

cuda_core/examples/jit_lto_fractal.py
```diff
@@ -0,0 +1,298 @@
...
+    Data[DataY*Width*4+DataX*4+2] = B;
+    Data[DataY*Width*4+DataX*4+3] = A;
+}
+"""
+
+# Simple implementation of Julia set from Wikipedia
+# http://en.wikipedia.org/wiki/Julia_set
+#
+# Note that this kernel is meant to be a simple, straight-forward
+# implementation, and so may not represent optimized GPU code.
```
rwgk: like suggested above

cuda_core/examples/jit_lto_fractal.py
```diff
@@ -0,0 +1,298 @@
...
+
+    unsigned ColorG = Iter % 50;
+    unsigned ColorB = Iter % 25;
+
+    R = 0.0f;
+    G = (float)ColorG / 50.0f;
+    B = (float)ColorB / 25.0f;
+    A = 1.0f;
+
+    Data[DataY*Width*4+DataX*4+0] = R;
```
rwgk: like suggested above

cuda_core/examples/jit_lto_fractal.py
```diff
@@ -0,0 +1,298 @@
...
+    Data[DataY*Width*4+DataX*4+1] = G;
+    Data[DataY*Width*4+DataX*4+2] = B;
+    Data[DataY*Width*4+DataX*4+3] = A;
+}
+"""
+
+
+if __name__ == "__main__":
+    # Parse command line arguments
```
rwgk, line 237: Best to move this to a function: you also get unused variable warnings. (Otherwise they will be at module scope.)

cuda_core/examples/jit_lto_fractal.py
vzhurba01: Added explanation for this buffer

cuda_core/examples/jit_lto_fractal.py
vzhurba01: Updated to have all of them as float.

I want to keep it fixed to 16 for simplicity. This is an algorithmic detail that distracts from the purpose of this file, that is how to link and why someone would link.

cuda_core/examples/jit_lto_fractal.py
vzhurba01: The intent is to demonstrate how libraries can link user input to customize their own workflows. While this is a not a library, it's the next closest thing while keeping the example self contained. Maybe replacing the word "mock" with "sample" or "example" could work? But this feels even farther from the truth, this is just a class that's trying to mock what a library could look like.

I can extend to comments further to document this nuance, but I find that that just adds more noise to the sample and makes it less clear.

rwgk: ExampleLinkAndLaunch

would sound good to me. But I'll go with what you decide.

Looks great to me otherwise, except I see there are still two easy unresolved comments.

cuda_core/examples/jit_lto_fractal.py
leofang:
```suggestion
            raise AssertionError(f"Invalid {target_type=}")
```

cuda_core/examples/jit_lto_fractal.py
leofang:
```suggestion
        # Return the result as a NumPy array (on host).
```

rwgk (approved):

leofang:
Thanks, @vzhurba01 @rwgk!

leofang:
Note: the bot cleanup came too late because I noticed the CPU runners were not accessible when merging to main:
https://github.com/NVIDIA/cuda-python/actions/runs/13606001690/attempts/1
and I had to manually trigger a re-run for failed jobs.

PR #481, opened 2025-03-03
Add event timing

leofang:
When working on #475 (68c93652eb4037af9c5755745668800566413689) I noticed we surprisingly haven't implemented event-based timing...  Looks like an oversight to me because we did spend a lot of time discussing the default of the `enabling_timing` option and how to retrieve the timing result. This PR adds the implementation, which is also aligned with the `cuda::experimental::timed_event` treatment: 
https://github.com/NVIDIA/cccl/blob/096596b3580d6cb2042fff692e8162d2a0ea356b/cudax/include/cuda/experimental/__event/timed_event.cuh#L90

### DIFF
--- a/cuda_core/cuda/core/experimental/_event.py
+++ b/cuda_core/cuda/core/experimental/_event.py
@@ -117,7 +117,12 @@ class Event:
 
     def __sub__(self, other):
         # return self - other (in milliseconds)
-        timing = handle_return(driver.cuEventElapsedTime(other.handle, self.handle))
+        try:
+            timing = handle_return(driver.cuEventElapsedTime(other.handle, self.handle))
+        except CUDAError as e:
+            raise RuntimeError(
+                "Timing capability must be enabled in order to subtract two Events; timing is disabled by default."
+            ) from e
         return timing
 
     @property
--- a/cuda_core/cuda/core/experimental/_event.py
+++ b/cuda_core/cuda/core/experimental/_event.py
@@ -48,7 +48,8 @@ class Event:
 
     Events can be used to monitor device's progress, query completion
     of work up to event's record, help establish dependencies
-    between GPU work submissions, and record the elapsed time on GPU:
+    between GPU work submissions, and record the elapsed time (in milliseconds)
+    on GPU:
 
     .. code-block:: python
 
@@ -58,14 +59,14 @@ class Event:
         # ... run some GPU works ...
         e2 = s.record(options={"enable_timing": True})
         e2.sync()
-        print(f"time = {e2 - e1}")
+        print(f"time = {e2 - e1} milliseconds")
 
         # Or, if events are already created:
         s.record(e1)
         # ... run some more GPU works ...
         s.record(e2)
         e2.sync()
-        print(f"time = {e2 - e1}")
+        print(f"time = {e2 - e1} milliseconds")
 
     Directly creating an :obj:`~_event.Event` is not supported due to ambiguity,
     and they should instead be created through a :obj:`~_stream.Stream` object.
@@ -120,7 +121,7 @@ class Event:
         return NotImplemented
 
     def __sub__(self, other):
-        # return self - other
+        # return self - other (in milliseconds)
         timing = handle_return(driver.cuEventElapsedTime(other.handle, self.handle))
         return timing
 
--- a/cuda_core/cuda/core/experimental/_event.py
+++ b/cuda_core/cuda/core/experimental/_event.py
@@ -47,8 +47,25 @@ class Event:
     the last recorded stream.
 
     Events can be used to monitor device's progress, query completion
-    of work up to event's record, and help establish dependencies
-    between GPU work submissions.
+    of work up to event's record, help establish dependencies
+    between GPU work submissions, and record the elapsed time on GPU:
+
+    .. code-block:: python
+
+        # To create events and record the timing:
+        s = Device(0).create_stream()
+        e1 = s.record(options={"enable_timing": True})
+        # ... run some GPU works ...
+        e2 = s.record(options={"enable_timing": True})
+        e2.sync()
+        print(f"time = {e2 - e1}")
+
+        # Or, if events are already created:
+        s.record(e1)
+        # ... run some more GPU works ...
+        s.record(e2)
+        e2.sync()
+        print(f"time = {e2 - e1}")
 
     Directly creating an :obj:`~_event.Event` is not supported due to ambiguity,
     and they should instead be created through a :obj:`~_stream.Stream` object.
@@ -95,6 +112,17 @@ class Event:
     def close(self):
         """Destroy the event."""
         self._mnff.close()
+
+    def __isub__(self, other):
+        return NotImplemented
+
+    def __rsub__(self, other):
+        return NotImplemented
+
+    def __sub__(self, other):
+        # return self - other
+        timing = handle_return(driver.cuEventElapsedTime(other.handle, self.handle))
+        return timing
 
     @property
     def is_timing_disabled(self) -> bool:
--- a/cuda_core/cuda/core/experimental/_event.py
+++ b/cuda_core/cuda/core/experimental/_event.py
@@ -54,16 +54,11 @@ class Event:
     .. code-block:: python
 
         # To create events and record the timing:
-        s = Device(0).create_stream()
-        e1 = s.record(options={"enable_timing": True})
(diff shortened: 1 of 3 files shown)


carterbox (commented):
I like the semantics of this implementation. It wouldn't have thought that you could directly subtract two `Event`s to get the difference between them in time, but I like it.

One thing that may be missing is documentation of what happens if timing is *not* enabled. What is the expected behavior? I suspect the answer is that the difference is zero, but raising an error because we tried to do timing without enabling timing seems more pythonic.

cuda_core/cuda/core/experimental/_event.py
```diff
@@ -47,8 +47,25 @@ class Event:
     the last recorded stream.
 
     Events can be used to monitor device's progress, query completion
-    of work up to event's record, and help establish dependencies
-    between GPU work submissions.
+    of work up to event's record, help establish dependencies
+    between GPU work submissions, and record the elapsed time on GPU:
```
carterbox: The text needs to name the expected units for the elapsed time. milliseconds, I believe.

carterbox (changes requested):

cuda_core/cuda/core/experimental/_event.py
leofang: Yes indeed, like this?
```suggestion
    between GPU work submissions, and record the elapsed time (in milliseconds)
    on GPU:
```

cuda_core/cuda/core/experimental/_event.py
```diff
@@ -47,8 +47,25 @@ class Event:
...
+
+    .. code-block:: python
+
+        # To create events and record the timing:
+        s = Device(0).create_stream()
+        e1 = s.record(options={"enable_timing": True})
+        # ... run some GPU works ...
+        e2 = s.record(options={"enable_timing": True})
+        e2.sync()
+        print(f"time = {e2 - e1}")
```
rwgk: `f"time = {e2 - e1} milliseconds"`
?
Also below.

cuda_core/cuda/core/experimental/_event.py
```diff
@@ -96,6 +113,17 @@ def close(self):
...
         self._mnff.close()
 
+    def __isub__(self, other):
+        return NotImplemented
+
+    def __rsub__(self, other):
+        return NotImplemented
+
+    def __sub__(self, other):
+        # return self - other
```
rwgk: `# return self - other (in milliseconds)`

?

cuda_core/tests/test_event.py
```diff
@@ -15,8 +17,17 @@
...
+    e1 = stream.record(options=options)
+    time.sleep(n_seconds)
+    e2 = stream.record(options=options)
+    for e in (e1, e2):
+        assert e.is_timing_disabled == (not enable_timing if enable_timing is not None else True)
+    if enable_timing:
+        e2.sync()
+        elapsed_time = e2 - e1
+        assert isinstance(elapsed_time, float)
+        assert n_seconds * 1000 <= elapsed_time < n_seconds * 1000 + 2  # tolerance 2 ms
```
rwgk: When I was reading this, for a moment I thought, what's that (`n_seconds`), nanoseconds?

`n_seconds` -> `delay_seconds`

`elapsed_time` -> `elapsed_time_ms`

Is the `tolerance` big enough? — I've battled a lot of flakes in my life because of tight tolerances. Unless it could do harm:

```
... + 100  # very generous tolerance to avoid flaky tests
```

I figure we don't want to stress-test `cudaEventElapsedTime`, we just want to be sure we're in the right ballpark.

cuda_core/tests/test_event.py
```diff
@@ -15,8 +17,17 @@
...
     options = EventOptions(enable_timing=enable_timing)
     stream = Device().create_stream()
-    event = stream.record(options=options)
-    assert event.is_timing_disabled == (not enable_timing if enable_timing is not None else True)
+    n_seconds = 0.5
+    e1 = stream.record(options=options)
+    time.sleep(n_seconds)
+    e2 = stream.record(options=options)
+    for e in (e1, e2):
+        assert e.is_timing_disabled == (not enable_timing if enable_timing is not None else True)
```
rwgk: This is a brain teaser!

How about

```
    assert e.is_timing_disabled == (True if enable_timing is None else not enable_timing)
```

rwgk (approved):

leofang:
> One thing that may be missing is documentation of what happens if timing is _not_ enabled. What is the expected behavior? I suspect the answer is that the difference is zero, but raising an error because we tried to do timing without enabling timing seems more pythonic.

Currently we do raise an exception since the driver API would return an error if the events were created without timing enabled:
```
cuda.core.experimental._utils.CUDAError: CUDA_ERROR_INVALID_HANDLE: invalid resource handle
```
So the question is if we want to do an exception chaining and raise, say, `RuntimeError` with a clearer err msg. @rwgk thoughts?

cuda_core/tests/test_event.py
rwgk: AFK but quick, I think I'm missing a `not`

rwgk: Ok edit from phone worked 😅

should be correct now

or

not bool(enable_timing)

maybe, although that's less readable

cuda_core/tests/test_event.py
leofang: `cudaEventElapsedTime`'s claimed accuracy is 1 ms, and event-based timing usually does far better than that, so I think a 2-ms tolerance should be met easily.

cuda_core/tests/test_event.py
leofang: fixed in commit 8aca468

cuda_core/tests/test_event.py
leofang: fixed in commit 8aca468

cuda_core/cuda/core/experimental/_event.py
leofang: fixed in commit 8aca468

cuda_core/cuda/core/experimental/_event.py
leofang: fixed in commit 8aca468

cuda_core/cuda/core/experimental/_event.py
leofang: fixed in commit 8aca468

leofang:
> Currently we do raise an exception since the driver API would return an error if the events were created without timing enabled:

We ensure this exception is raised in commit 8aca468. The discussion for an optional exception chaining is still open.

carterbox:
If it's possible to catch and reraise the error with a message like "Timing must be enabled in order to compute a diff between Events. Events have timing disabled by default." That would be better than "invalid resource handle".

cuda_core/cuda/core/experimental/_event.py
```diff
@@ -96,6 +114,17 @@ def close(self):
...
 
+    def __isub__(self, other):
+        return NotImplemented
+
+    def __rsub__(self, other):
+        return NotImplemented
+
+    def __sub__(self, other):
+        # return self - other (in milliseconds)
+        timing = handle_return(driver.cuEventElapsedTime(other.handle, self.handle))
```
rwgk: @carterbox wrote:
> If it's possible to catch and reraise the error with a message like "Timing must be enabled in order subtract Events. Events have timing disabled by default." That would be better than "invalid resource handle".

I believe this could be pretty simple (with less runtime overhead, too):

Roughly, untested:
```
err, timing = driver.cuEventElapsedTime(other.handle, self.handle)
if err == driver.CUresult.CUDA_SUCCESS:
    return timing
if err == driver.CUresult.CUDA_ERROR_INVALID_HANDLE:
    raise RuntimeError("Timing must be enabled in order subtract Events. Events have timing disabled by default.")
handle_return(err)
raise RuntimeError("This code path is meant to be unreachable.")
```

rwgk (dismissed):

cuda_core/cuda/core/experimental/_event.py
leofang: updated in commit 268eee5 based on the discussion on Tuesday

leofang:
This is ready.

rwgk (approved):
Seeing the code now, I like the exception chaining approach better than the idea I posted before.

leofang:
Thanks, Ralf/Daniel!

PR #458, opened 2025-02-21
Clearer error messages (cuda.core)

rwgk:
This is the main PR to address issue #198

Cleanup of error messages is considered complete.

Test coverage is expanded, but still not complete. Many changes were exercised only interactively, by temporarily changing `if cond` to `if not cond` and visually inspecting the raised exceptions.

For completeness: These will be handled in separate PRs: 5dc2331b96d51eaef7a703ad434add801f3f1bda, d3df80d6d4db1062f3dd969e2a99b3dc08837f60

Issue #198, opened 2024-10-30
Make error messages more clear and actionable


leofang:
In the current prototype we often just raise the exception without any error message. It was meant to be only as a placeholder that we need to improve/iterate quickly. Error messages should be short but clear, informing users what's wrong (and potential fixes if possible).

leofang:
Ralf to oversee the clear error msg process, need to know bw constraint by beta 3

ksimpson-work:
Context:

cuda.core obviously wraps the bindings which are generated via either the cuda-python binding generator, or using cybind. Currently nvjitlink is the only library which has bindings generated by cybind, the rest of the bindings (cuda driver, cuda runtime, nvrtc, culink etc.) are all from the cuda-python binding generator. I will refer to these as cuda-python-bindings

The relevance of this is that there is a difference in how each of these bindings handle errors:

nvjitlink + other cybind bindings which come in the future -> **raise exceptions at the cython layer**
all the rest (cuda-python-bindings) -> **return a tuple (error_code, result), no exception raised on error**

We have _utils.handle_return() which we use to parse the error_code from any call to the cuda-python-bindings. When it is not success we **raise a generic exception at the cuda.core layer**

In order to improve the error message from certain types of errors (specific error codes, or specific library calls) we implement a context manager to manage exceptions in _linker.py. In this case we know that it is likely helpful to provide the error log from the linker along with the original exception.

My idea to improve the error messages module wide, would be to give every class an implementation of that contextmanager (either re write, or pass a callback to some shared context manager in utils). This exception manager can switch case on the error code pertaining to any raised exception, and provide a better message. It would have the ability to fallback to the generic message if there was nothing meaningful to add, or if we wanted to go lightweight at first. This way we can handle the exceptions raised by nvjitlink, as well as the exceptions raised by _utils.handle_return across all files the same way. This gives you the freedom to design something slick and maintainable, so that it will be easy to add a custom handler for a specific call or exception type in the future (since we don't need to handle every single special case in this first review).

Long story short:

look at _linker.py and port the _exception_manager over to all the classes so that they can all easily customize and standardize exception handling. OR come at it from a different angle entirely.

Let me know if you have any questions. The changes I had were actually not in an isolated branch, and to clean them up would take awhile. Happy to do that if it would be helpful but let me know.

PR #484, opened 2025-03-03
Apply `__new__` approach to disabling `__init__`

rwgk:
Contributes to #198

Make disabling `__init__` uniform across cuda.core:

* Use `@classmethod` as the only constructor: Best for strict enforcement.

* Uniform `raise RuntimeError()` with a standardized message form.

The starting point for this change was extracted from PR #458:

* https://github.com/NVIDIA/cuda-python/pull/458#issuecomment-2692862620

* https://github.com/NVIDIA/cuda-python/pull/458#issuecomment-2692875231

### DIFF
--- a/cuda_core/cuda/core/experimental/_stream.py
+++ b/cuda_core/cuda/core/experimental/_stream.py
@@ -74,7 +74,7 @@ class Stream:
     def __new__(self, *args, **kwargs):
         raise RuntimeError(
             "Stream objects cannot be instantiated directly. "
-            "Please use Device APIs (create_stream) or other Stream APIs (from_handle, wait)."
+            "Please use Device APIs (create_stream) or other Stream APIs (from_handle)."
         )
 
     __slots__ = ("__weakref__", "_mnff", "_nonblocking", "_priority", "_device_id", "_ctx_handle")
--- a/cuda_core/cuda/core/experimental/_stream.py
+++ b/cuda_core/cuda/core/experimental/_stream.py
@@ -71,18 +71,37 @@ class Stream:
                 self.owner = None
             self.handle = None
 
-    __slots__ = ("__weakref__", "_mnff", "_nonblocking", "_priority", "_device_id", "_ctx_handle")
-
-    def __init__(self):
-        raise NotImplementedError(
-            "directly creating a Stream object can be ambiguous. Please either "
-            "call Device.create_stream() or, if a stream pointer is already "
-            "available from somewhere else, Stream.from_handle()"
+    def __new__(self, *args, **kwargs):
+        raise RuntimeError(
+            "Stream objects cannot be instantiated directly. "
+            "Please use Device APIs (create_stream) or other Stream APIs (from_handle, wait)."
         )
 
-    @staticmethod
-    def _init(obj=None, *, options: Optional[StreamOptions] = None):
-        self = Stream.__new__(Stream)
+    __slots__ = ("__weakref__", "_mnff", "_nonblocking", "_priority", "_device_id", "_ctx_handle")
+
+    @classmethod
+    def _legacy_default(cls):
+        self = super().__new__(cls)
+        self._mnff = Stream._MembersNeededForFinalize(self, driver.CUstream(driver.CU_STREAM_LEGACY), None, True)
+        self._nonblocking = None  # delayed
+        self._priority = None  # delayed
+        self._device_id = None  # delayed
+        self._ctx_handle = None  # delayed
+        return self
+
+    @classmethod
+    def _per_thread_default(cls):
+        self = super().__new__(cls)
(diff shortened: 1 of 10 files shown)


leofang:
Looks fine to me! Merge?

rwgk:
I forgot that I wanted to add a couple more small tests. Doing that now.

rwgk:
@leofang Test coverage is now systematically complete, covering "front doors" and "back doors"; see comments added with commit eda324f80864e6b1036441149276ec8c8834fb06.

Testing that the back doors are locked is done through APIs that are unavoidably (for all practical purposes) visible through public APIs.

Recently I experienced pushback when testing private functionality under the CCCL repo, so I asked ChatGPT about it:

* https://chatgpt.com/share/67c5dcca-088c-8008-9e4a-bfc778f01204

The lists of pros and cons are complex. I'm convinced it would be unproductive to be to pure one way or another, case-by-case judgement is important. That's how I decided to add tests that ensure the back doors are locked, and remain so.

cuda_core/cuda/core/experimental/_stream.py
```diff
@@ -71,18 +71,37 @@ def close(self):
...
-            "available from somewhere else, Stream.from_handle()"
-        )
+    @classmethod
+    def _legacy_default(cls):
+        self = super().__new__(cls)
+        self._mnff = Stream._MembersNeededForFinalize(self, driver.CUstream(driver.CU_STREAM_LEGACY), None, True)
+        self._nonblocking = None  # delayed
+        self._priority = None  # delayed
+        self._device_id = None  # delayed
+        self._ctx_handle = None  # delayed
```
leofang, line 89: IIUC this is actually a bug fix too! I wonder why it wasn't caught before...

cuda_core/cuda/core/experimental/_stream.py
```diff
@@ -71,18 +71,37 @@ def close(self):
...
-    def _init(obj=None, *, options: Optional[StreamOptions] = None):
-        self = Stream.__new__(Stream)
+    @classmethod
+    def _per_thread_default(cls):
+        self = super().__new__(cls)
+        self._mnff = Stream._MembersNeededForFinalize(self, driver.CUstream(driver.CU_STREAM_PER_THREAD), None, True)
+        self._nonblocking = None  # delayed
+        self._priority = None  # delayed
+        self._device_id = None  # delayed
+        self._ctx_handle = None  # delayed
```
leofang, line 99: ditto

cuda_core/cuda/core/experimental/_stream.py
```diff
@@ -71,18 +71,37 @@ def close(self):
                 self.owner = None
             self.handle = None
 
+    def __new__(self, *args, **kwargs):
+        raise RuntimeError(
+            "Stream objects cannot be instantiated directly. "
+            "Please use Device APIs (create_stream) or other Stream APIs (from_handle, wait)."
```
leofang: `wait()` is not a constructor:
```suggestion
            "Please use Device APIs (create_stream) or other Stream APIs (from_handle)."
```

leofang (approved):
LGTM other than fixing one error message.

cuda_core/cuda/core/experimental/_stream.py
rwgk: Double-checking before I remove it (and also for me to learn):

I saw this code in `Stream.wait`:

```
            if not isinstance(event_or_stream, Stream):                         
                try:                                                            
                    stream = Stream._init(event_or_stream)                      
                except Exception as e:                                          
                    raise ValueError("only an Event, Stream, or object supporting __cuda_stream__ can be waited") from e
```

IIUC this calls `__cuda_stream__` to create a `Stream` object.

Is the following too twisted? — Someone may have the idea to call `Stream(object_with_dunder_cuda_stream)`, because they don't expect the automagic conversion?

leofang: > Is the following too twisted? — Someone may have the idea to call `Stream(object_with_dunder_cuda_stream)`, because they don't expect the automagic conversion?

Right now, if one wants to create a stream from a protocol-supported object, the intended way is to do `s = Device().create_stream(object_with_dunder_cuda_stream)`. Not sure if this is what you are asking?

rwgk: Yes, thanks for catching this!

`wait` removed in commit 201d3fc17e42262c5aed965e7257edd6992f90c4

Your review went to "stale", could you please re-approve?

leofang (approved):

leofang:
Since the previous run was green and the most recent change is comment-only, I suggest we admin-merge this without another CI run. @rwgk ready?

rwgk:
> Since the previous run was green and the most recent change is comment-only, I suggest we admin-merge this without another CI run. @rwgk ready?

Yes, please! :-)

PR #492, opened 2025-03-06
add public handle to object code

### LINKED ISSUE #490: Provide accessor for `ObjectCode._handle`
Currently if we have a lazily loaded `ObjectCode`, we could get its underlying `CUlibrary` via `_handle` attribute. We should provide a public accessor like `handle`, just like other objects in `cuda.core`. And it would be good UX if the accessor has a precondition with `_lazy_load_module`.

ksimpson-work:
Weirdly I am not able to tag Michael or add him as reviewer. 

close #490

### DIFF
--- a/cuda_core/cuda/core/experimental/_module.py
+++ b/cuda_core/cuda/core/experimental/_module.py
@@ -341,3 +341,9 @@ class ObjectCode:
     def code(self) -> CodeTypeT:
         """Return the underlying code object."""
         return self._module
+
+    @property
+    @precondition(_lazy_load_module)
+    def handle(self):
+        """Return the underlying handle object."""
+        return self._handle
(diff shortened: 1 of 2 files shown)


isVoid (approved):
Many thanks for the quick response. An optimization tips, with this new fixture, you can actually simplify a few others by:

```python
@pytest.fixture(scope="function")
def get_saxpy_object_code(init_cuda):
...

@pytest.fixture(scope="function")
def get_saxpy_kernel_ptx(get_saxpy_object_code):
    mod = get_saxpy_object_code
    return mod._module, mod

@pytest.fixture(scope="function")
def get_saxpy_kernel(get_saxpy_object_code):
    mod = get_saxpy_object_code
    return mod.get_kernel("saxpy<float>"), mod
```

leofang (approved):

PR #493, opened 2025-03-06
Ensure `cuda` is a namespace package & Improve warnings

### LINKED ISSUE #162: Turn `DeprecationWarning` to `UserWarning`

### LINKED ISSUE #477: Remove `cuda/__init__.py` to avoid confusing various tools
It seems hatchling (according to @rwgk from https://github.com/NVIDIA/cccl/pull/3201#issue-2751245974) and pip (#476) are confused by the conflict between our intention of `cuda` being a namespace package and `cuda/__init__.py` still exists today.

leofang:
Close #477. Close #162.


rwgk (approved):
This certainly looks great to me!

I'm wondering though, what exactly was the reason for adding these files? How much trouble is it to leave some hints in the PR description?

leofang:
> what exactly was the reason for adding these files? How much trouble is it to leave some hints in the PR description?

As part of the epic #493, we need to remove `cuda/__init__.py` to ensure `cuda` is a [namespace package](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages). The existence of this file has been a source of multiple traps and bugs.

We still kept it around because it used to provide `cuda.__version__`. Since we have raised a deprecation warning for a few releases, and it is a minor thing, we now remove it entirely.

vzhurba01 (dismissed):

cuda_bindings/cuda/__init__.py
```diff
@@ -1,14 +0,0 @@
-def __getattr__(name):
-    if name == "__version__":
-        import warnings
-
-        warnings.warn(
-            "accessing cuda.__version__ is deprecated, " "please switch to use cuda.bindings.__version__ instead",
-            DeprecationWarning,
```
leofang, line 7: @vzhurba01 I did a few quick searches on my cell and noticed that some projects still use this attribute. Maybe for consistency we should just change this to `UserWarning` instead of removing it entirely in 12.9.0? We can remove it in say 12.9.1 instead. WDYT?

kkraus14: Introducing a breaking change like this is unideal in general in a minor / patch release. Can we defer removing it until the next major release?

kkraus14: If we just don't want the `DeprecationWarning` getting swallowed by default we should use `FutureWarning` instead of a generic `UserWarning`

leofang: The major problem is we want `cuda` to be a namespace package, but the existence of `cuda/__init__.py` breaks a bunch of tools. We need to get rid of it sooner than later to not block any of our projects. Let me see if a `.pth` based hack can serve the same purpose.

leofang: > Let me see if a .pth based hack can serve the same purpose.

Commit [2537375](https://github.com/NVIDIA/cuda-python/pull/493/commits/25373752904f4fd8c0d64c8713dad773b970f356) seems to do the trick!

> we should use FutureWarning instead of a generic UserWarning

Done in commit [cf28b44](https://github.com/NVIDIA/cuda-python/pull/493/commits/cf28b44d4a488012d8b2cd8dcba840e160c3f8fd)

vzhurba01 (approved):

kkraus14 (approved):

PR #497, opened 2025-03-06
cuda.bindings: Fix segfault when converting `char*` `NULL` to `bytes`

rwgk:
The segfault was discovered while working on PR #458.

Full test coverage for changed code.

### DIFF
--- a/cuda_bindings/tests/test_cuda.py
+++ b/cuda_bindings/tests/test_cuda.py
@@ -977,3 +977,15 @@ def test_all_CUresult_codes(max_code=1000):
     # The number will increase over time as new enums are added and support for
     # old CTKs is dropped, but it is not critical that this number is updated.
     assert num_good >= 76  # CTK 11.0.3_450.51.06
+
+
+def test_cuKernelGetName_failure():
+    err, name = cuda.cuKernelGetName(0)
+    assert err == cuda.CUresult.CUDA_ERROR_INVALID_VALUE
+    assert name is None
+
+
+def test_cuFuncGetName_failure():
+    err, name = cuda.cuFuncGetName(0)
+    assert err == cuda.CUresult.CUDA_ERROR_INVALID_VALUE
+    assert name is None
--- a/cuda_bindings/tests/test_cuda.py
+++ b/cuda_bindings/tests/test_cuda.py
@@ -948,3 +948,29 @@ def test_conditional():
 def test_CUmemDecompressParams_st():
     desc = cuda.CUmemDecompressParams_st()
     assert int(desc.dstActBytes) == 0
+
+
+def test_all_CUresult_codes(max_code=1000):
+    num_good = 0
+    for code in range(max_code + 1):
+        try:
+            error = cuda.CUresult(code)
+        except ValueError:
+            pass  # cython-generated enum does not exist for this code
+        else:
+            err_name, name = cuda.cuGetErrorName(error)
+            if err_name == cuda.CUresult.CUDA_SUCCESS:
+                assert name
+                err_desc, desc = cuda.cuGetErrorString(error)
+                assert err_desc == cuda.CUresult.CUDA_SUCCESS
+                assert desc
+                num_good += 1
+            else:
+                # cython-generated enum exists but is not known to an older driver
+                # (example: cuda-bindings built with CTK 12.8, driver from CTK 12.0)
+                assert name is None
+                assert err_name == cuda.CUresult.CUDA_ERROR_INVALID_VALUE
+    # Super-simple smoke check: Do we have at least some "good" codes?
+    # The number will increase over time as new enums are added and support for
+    # old CTKs is dropped, but it is not critical that this number is updated.
+    assert num_good >= 76  # CTK 11.0.3_450.51.06
--- a/cuda_bindings/tests/test_cuda.py
+++ b/cuda_bindings/tests/test_cuda.py
@@ -950,9 +950,12 @@ def test_CUmemDecompressParams_st():
     assert int(desc.dstActBytes) == 0
 
 
-def test_all_CUresult_codes(max_code=1000):
+def test_all_CUresult_codes():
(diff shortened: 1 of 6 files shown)


rwgk:
@leofang The two test failures are these flakes:

```
>           assert delay_seconds * 1000 <= elapsed_time_ms < delay_seconds * 1000 + 2  # tolerance 2 ms
E           assert 503.1598205566406 < ((0.5 * 1000) + 2)
```

```
>           assert delay_seconds * 1000 <= elapsed_time_ms < delay_seconds * 1000 + 2  # tolerance 2 ms
E           assert 505.0583190917969 < ((0.5 * 1000) + 2)
```

vzhurba01:
Changes LGTM at the moment but I see that this is still a draft

rwgk:
> Changes LGTM at the moment

Thanks!

> but I see that this is still a draft

I'm looking into adding more tests, to ideally cover all changes.

And we need to do something about the flakes, even though they are unrelated.

rwgk:
Woah, looks like I'm good at generating segfaults ...

```
Fatal Python error: Segmentation fault

Current thread 0x0000ffff9f213020 (most recent call first):
  File "/__w/cuda-python/cuda-python/cuda_bindings/tests/test_nvrtc.py", line 35 in test_nvrtcGetLoweredName_failure
...
  File "/opt/hostedtoolcache/Python/3.9.21/arm64/lib/python3.9/site-packages/_pytest/config/__init__.py", line 201 in console_main
  File "/opt/hostedtoolcache/Python/3.9.21/arm64/bin/pytest", line 8 in <module>
/__w/_temp/6c5a43e8-eaeb-4c41-8067-c06dcfabc91e.sh: line 12:  4545 Segmentation fault      (core dumped) pytest -rxXs -v tests/
tests/test_nvrtc.py::test_nvrtcGetLoweredName_failure 
```

@leofang should I just remove that new test and leave that for another PR? I don't think I'll get to the bottom of the newly discovered segfault today. Tomorrow pretty sure.

leofang:
Just isolate out the cuda.core changes from those for bindings. Let's merge core-related things first; bindings can wait.

rwgk:
The cuda.core change is to resolve the flakes. (Without it I'd (maybe) have to hit rerun a few times to deflake.)

I just backed out the test that triggers the segfaults: commit a39720b5c8186d966666d0ecbc05e6bab5fb8f45

leofang:
> The cuda.core change is to resolve the flakes

Ah, I meant we can separate it out and merge it first. Mainly if we want to auto-backport the fix, the cuda-11 branch does not have any cuda.core code, so the apto-backport would fail. I am curious if it could work without any cuda-core changes involved.

rwgk:
I removed the `test_timings()` change from this PR. Could you please admin-merge this, too? The tests passed before.

cuda_bindings/tests/test_cuda.py
```diff
@@ -948,3 +948,44 @@ def test_conditional():
 def test_CUmemDecompressParams_st():
     desc = cuda.CUmemDecompressParams_st()
     assert int(desc.dstActBytes) == 0
+
+
+def test_all_CUresult_codes(max_code=1000):
```
leofang: btw I think this should work: the number of `IntEnum` can be queried via `len()`
```suggestion
def test_all_CUresult_codes(max_code=len(cuda.CUresult)):
```

leofang (dismissed):
LGTM. It'd be nice if Vlad can take a final look.

cuda_bindings/tests/test_cuda.py
rwgk: I want to intentionally run past the end of the enums. Stress test; simpler, too.

rwgk: To be sure I understand correctly, I tried this:
```
print(f"\n{len(cuda.CUresult)=}")
```

That produces:
```
len(cuda.CUresult)=100
```

That's the number of enum values. My original intent was to brute-force loop over all integers up to
```
CUDA_ERROR_UNKNOWN = 999
```
and then one past that. It's the simplest loop we can write, and it exercises that the `cuda.CUresult` enum type works robustly, too.

I'm assuming that `CUDA_ERROR_*` will be filled in in the future, but only with values < 999.

leofang: Ah, ok, so I confused myself (and you, sorry!), sounds like what we need is `range(int(max(cuda.CUresult))+1)` to go beyond the max value 999?

kkraus14 (dismissed):
lgtm

cuda_bindings/tests/test_cuda.py
rwgk: Done in commit 781446943cb22130be26dea22d76aef06f0b7ac8.

leofang:
@rwgk could you update the commit message and avoid `@someone`? Last time you did that I ended up receiving a ton of notifications whenever someone rebased a branch that contains this commit lol

rwgk:
> @rwgk could you update the commit message and avoid `@someone`? Last time you did that I ended up receiving a ton of notifications whenever someone rebased a branch that contains this commit lol

Oh ... sorry, done.

(I did this a lot in the pybind11 repo, wanting to give proper credit. I didn't realize this can lead to spammy notifications.)

leofang (approved):

rwgk:
@leofang The tests passed previously with the exact same code. The only change was to the commit message. I.e. admin merge would be ideal.

leofang:
ok

rwgk:
> Backport failed because this pull request contains merge commits. You can either backport this pull request manually, or configure the action to skip merge commits.

Should I take care of that? (Happy to.)

leofang:
> Should I take care of that? (Happy to.)

Yes, please. My bad in updating your branch, sorry...

rwgk:
I'll backport this together with https://github.com/NVIDIA/cuda-python/pull/499, after that is merged.

PR #499, opened 2025-03-07
cuda.bindings: Add `test_nvrtcGetLoweredName_failure`

rwgk:
This test was first added while working on PR #497, but moved to this separate PR, so that we could focus on the Segmentation Fault in isolation.

It turns out that only CTK 12.0 does not handle invalid `prog` arguments gracefully, newer CTKs do. The approach taken here is to simply skip the new test for CTK < 12.1. This will be useful as a stress-test for future CTK releases.

See https://github.com/NVIDIA/cuda-python/pull/499#issuecomment-2707349350 for more details.

### DIFF
--- a/.github/workflows/test-wheel-linux.yml
+++ b/.github/workflows/test-wheel-linux.yml
@@ -194,7 +194,7 @@ jobs:
 
           pushd ./cuda_bindings
           pip install -r requirements.txt
-          pytest -ra -s -v tests/
+          pytest -rxXs -v tests/
 
           # It is a bit convoluted to run the Cython tests against CTK wheels,
           # so let's just skip them.
@@ -205,7 +205,7 @@ jobs:
               # TODO: enable this once win-64 runners are up 
               exit 1
             fi  
-            pytest -ra -s -v tests/cython
+            pytest -rxXs -v tests/cython
           fi
           popd
 
@@ -229,7 +229,7 @@ jobs:
 
           pushd ./cuda_core
           pip install -r "tests/requirements-cu${TEST_CUDA_MAJOR}.txt"
-          pytest -ra -s -v tests/
+          pytest -rxXs -v tests/
 
           # It is a bit convoluted to run the Cython tests against CTK wheels,
           # so let's just skip them. Also, currently our CI always installs the
@@ -243,7 +243,7 @@ jobs:
               # TODO: enable this once win-64 runners are up
               exit 1
             fi
-            pytest -ra -s -v tests/cython
+            pytest -rxXs -v tests/cython
           fi
           popd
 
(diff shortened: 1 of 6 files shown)


rwgk:
All jobs with Segmentation faults are running with CUDA 12.0.1:

After unpacking logs_35364236774.zip (https://github.com/NVIDIA/cuda-python/actions/runs/13725931628/)

```
$ grep 'Segmentation fault' *.txt | cut -d: -f1 | uniq
23_Test__linux-64__Python_3.10__CUDA_12.0.1__Runner_default__local_CTK____test.txt
27_Test__linux-64__Python_3.11__CUDA_12.0.1__Runner_default__local_CTK____test.txt
32_Test__linux-64__Python_3.12__CUDA_12.0.1__Runner_default__local_CTK____test.txt
38_Test__linux-64__Python_3.13__CUDA_12.0.1__Runner_default__local_CTK____test.txt
44_Test__linux-64__Python_3.9__CUDA_12.0.1__Runner_default__local_CTK____test.txt
47_Test__linux-aarch64__Python_3.10__CUDA_12.0.1__Runner_default__local_CTK____test.txt
53_Test__linux-aarch64__Python_3.11__CUDA_12.0.1__Runner_default__local_CTK____test.txt
58_Test__linux-aarch64__Python_3.12__CUDA_12.0.1__Runner_default__local_CTK____test.txt
63_Test__linux-aarch64__Python_3.13__CUDA_12.0.1__Runner_default__local_CTK____test.txt
67_Test__linux-aarch64__Python_3.9__CUDA_12.0.1__Runner_default__local_CTK____test.txt
```

vzhurba01:
While debugging this there are two things that stood out:

1. The segfault happens after we've called into the Cython layer. In that layer all the data is already packaged and the only action taken is the call into the NVRTC library.
2. When looking at the header differences between between 12.8 (passing) and 12.0 (failing) the API signature and types are identical... but the documentation has minor differences.

>https://github.com/NVIDIA/cuda-python/blob/v12.0.0/cuda/nvrtc.pyx.in#L784
>https://github.com/NVIDIA/cuda-python/blob/v12.8.0/cuda_bindings/cuda/bindings/nvrtc.pyx.in#L808

In 12.8.0 the API `nvrtcAddNameExpression()` states that it can return `NVRTC_ERROR_INVALID_PROGRAM`, but this line is missing in 12.0.0. This difference reflects what we're seeing with `nvrtc.nvrtcGetLoweredName()` where the error check has been fixed but documentation does not reflect it.

From offline discussion, having us skip this test on 12.0 would allow us to continue stress testing new releases.

vzhurba01 (approved):

PR #505, opened 2025-03-10
Add optional mypy check

shwina:


rwgk:
I developed the feeling that mypy is losing momentum. So I asked like this:

> We want to add Python type checking to pre-commit and .github/workflows. mypy is one option. Are there other options we should consider?

* https://chatgpt.com/share/67cf1139-82b0-8008-b41c-848d9047e207

Picking out one point from the response:

> Pyright is often used for editor-level feedback since it's fast and lightweight, but it works well for CI too.

I believe supporting "editor-level feedback" is the single most valuable motivation for investing time into typing. — Other than that, in my experience, typing is only a moderate win overall. (ruff gives us a lot of protections already.)

I see from the CI output that we have a sizable cleanup job. Do we want to target mypy, or pyright? There is probably a big overlap, but I expect that we'll spend a significant fraction of time dealing with idiosyncrasies of one or the other.

shwina:
> I developed the feeling that mypy is losing momentum. So I asked like this

Are there examples of projects using pyright for enforcing type checks in CI? I looked at some existing "big" Python projects to see what they were using:

- [PyTorch uses MyPy](https://github.com/pytorch/pytorch/blob/main/mypy.ini)
- [NumPy uses MyPy](https://github.com/numpy/numpy/blob/main/.github/workflows/mypy.yml)
- [scikit-learn uses MyPy](https://github.com/scikit-learn/scikit-learn/blob/5cce87176a530d2abea45b5a7e5a4d837c481749/.pre-commit-config.yaml#L21)

kkraus14:
Would it make sense to integrate this with `pre-commit` and then make the CI job call just that task in pre-commit for now? Would maybe be nice to just make this part of the developer workflow to start getting fixes in?

shwina:
Here's how the pre-commit results get reported:

https://results.pre-commit.ci/run/github/381173759/1741612446.WIQqJeAMSMy0Q8RPtY3MbA

If we made the `mypy` check "unconditionally" pass, I couldn't figure out if/whether we could actually inspect its output from that result page.

shwina:
But yes, ultimately I would like for type checking to be part of the pre-commit step (not as a standalone check)

leofang:
> Are there examples of projects using pyright for enforcing type checks in CI? I looked at some existing "big" Python projects to see what they were using:

I think mypy is really unfriendly to array libraries like NumPy and it's being discussed to move away to pyright or basedpyright. @seberg was involved in long discussions like https://github.com/numpy/numpy/issues/27957 and https://github.com/numpy/numpy/issues/28076. But it is unclear to me why we wouldn't not like mypy in cuda.core. I guess we need to go over the list of all failures and decide.

leofang:
> If we made the `mypy` check "unconditionally" pass, I couldn't figure out if/whether we could actually inspect its output from that result page.

Yeah my limited understanding about pre-commit.ci is the same. We can't make certain pre-commit steps optional. So moving this check to pre-commit in the future SGTM.

cuda_bindings/cuda/py.typed
leofang, line 1: Almost all cuda.bindings code are auto-generated, and I'd hate to make codegen harder. Let's focus on cuda-core for now (i.e. add `py.typed` to `cuda_core/cuda`).

rwgk:
> But it is unclear to me why we wouldn't not like mypy in cuda.core.

What I had in mind is optimizing for what the user community cares about the most. I'm assuming that's type hints in code editors. I could be wrong.

kkraus14:
> Here's how the pre-commit results get reported:
> 
> https://results.pre-commit.ci/run/github/381173759/1741612446.WIQqJeAMSMy0Q8RPtY3MbA
> 
> If we made the `mypy` check "unconditionally" pass, I couldn't figure out if/whether we could actually inspect its output from that result page.

We could exclude the mypy task from pre-commit in the CI run for now and then have a separate optional job that runs only the mypy task? This way developers using pre-commit locally get mypy as part of their workflow unless they opt out of it, but it remains optional to get PRs through until we hit a later point.

rwgk:
> Should we even use mypy for type checking?

I don't have strong opinions here. It's definitely a good start.

> I definitely think we should eventually invoke mypy via pre-commit.

Assuming we have pre-commit type checking, and we've made the code compatible (no errors), would we want to keep the type checking CI job?

shwina:
> Assuming we have pre-commit type checking, and we've made the code compatible (no errors), would we want to keep the type checking CI job?

Nope, at that point it goes away. The point of this PR is just to have _somewhere_ that the results of type checking are visible

rwgk:
TBH, just thinking about what I'd do:
* The cuda_bindings/pyproject.toml change you figured out looks very useful.
* The CI job, why not, but I'd probably not look at the CI logs.
* Assuming I'd decide to jump on cleaning up, I'd figure out the .pre-commit-config.yaml addition we need, then roll up my sleeves working through the pre-commit errors until I got them all.

rwgk (dismissed):
LGTM, mainly for the cuda_bindings/pyproject.toml change.

shwina:
OK - thanks. I opened this PR so that it's mergeable. I'd defer to you or @leofang if/whether to merge this - no objections either way from me

cuda_bindings/cuda/py.typed
shwina: Whoops - missed this. Just fixed.

cuda_core/pyproject.toml
leofang, line 1: what does this `overrides` section do?

leofang (approved):

cuda_bindings/cuda/py.typed
leofang: thx Ashwin!

kkraus14:
> 1. How do I configure pre-commit.ci to skip a specific pre-commit check?

You'd use the `skip` option as described here: https://pre-commit.ci/#configuration-skip. I'm not clear on if its possible to add another pre-commit.ci job that skips everything but `mypy` and isn't required to pass.

> 2. How do I make it so that a specific pre-commit check is "optional"? I guess I could wrap the `mypy` invocation in a script that always succeeds, and have `pre-commit` call that script. If we do this, (1) becomes a non-issue.

It would be optional in the sense that it wouldn't be required to pass in CI to merge PRs, but would be opt-out for developers where they'd need to add an environment variable of `SKIP=mypy` to avoid it from running when they run pre-commit locally.

leofang:
btw let's hold off merging until code freeze is lifted (we're still wrapping up v0.2.0). 

Another question is: If we add `py.typed` now, but don't fix the typing errors, would it impact downstream projects running mypy? Do they have to also waive the failures?

rwgk:
Cross-referencing #527 — I very quick experiment, trying to use Cursor to add typing annotations.

cpcloud:
What's the status of this PR? Can we make a decision and move forward on it (or close it)?

cpcloud:
There are conflicts that need to be resolved. What's the status of this PR?

shwina:
> There are conflicts that need to be resolved. What's the status of this PR?

I haven't looked at it in a very long time. At this point, for me to re-familiarize myself, get set up with a dev environment, and rebase would likely take longer than just implementing from scratch. I'd recommend closing and opening a new PR. Please feel free to take what bits are useful from here.

PR #507, opened 2025-03-11
clean up device initialization in test

ksimpson-work:
I suspect the issue was related to the fact that we had a dangling device initialization. In any case we should test against our actual guarantee which is that Device() is current device when device_ordinal is unsupplied


rwgk (approved):

leofang:
I think we miss an `import`.

rwgk:
I only have a minute right now. Just quick, I saw this in one of the failing tests:

```
______________________________ test_device_alloc _______________________________

deinit_cuda = None

    def test_device_alloc(deinit_cuda):
        device = Device()
        device.set_current()
        buffer = device.allocate(1024)
        device.sync()
        assert buffer.handle != 0
        assert buffer.size == 1024
>       assert buffer.device_id == cuda.cudaGetDevice()

tests/test_device.py:53: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

name = 'cudaGetDevice'

    def __getattr__(name):
        if name == "__version__":
            import warnings
            warnings.warn("accessing cuda.__version__ is deprecated, "
                          "please switch to use cuda.bindings.__version__ instead",
                          DeprecationWarning, stacklevel=2)
            from . import bindings
            return bindings.__version__
    
>       raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
E       AttributeError: module 'cuda' has no attribute 'cudaGetDevice'

/opt/hostedtoolcache/Python/3.9.21/x64/lib/python3.9/site-packages/cuda/__init__.py:10: AttributeError
```

I see this elsewhere:

```
handle_return(runtime.cudaGetDevice())
```

We have the `runtime` and `handle_return` imports already in test_device.py

Maybe @leofang saw something else?

ksimpson-work:
@rwgk You're totally right, I didn't test locally before running the CI since it was so small but I missed some obvious issues

cuda_core/tests/test_device.py
```diff
@@ -37,18 +37,20 @@ def test_device_set_current(deinit_cuda):
...
-def test_device_alloc(init_cuda):
+def test_device_alloc(deinit_cuda):
     device = Device()
+    device.set_current()
     buffer = device.allocate(1024)
     device.sync()
     assert buffer.handle != 0
     assert buffer.size == 1024
-    assert buffer.device_id == 0
+    assert buffer.device_id == handle_return(runtime.cudaGetDevice())
```
rwgk: Looking at this with a fresh eye, I'm wondering:
* We're in cuda.core but we're reaching back into a cuda.bindings API (`runtime`). Do we have to?
* Could/should we do this instead? (does it matter?)

```Python
    assert buffer.device_id == int(device)
```

rwgk (approved):

cuda_core/tests/test_device.py
keenan-simpson: Yeah that also works. I'll add it, and also a new test for the device id against our guarantee

rwgk (approved):
Reapproving, assuming tests pass.

The new test looks great.

PR #508, opened 2025-03-11
Fix `test_timing` flakiness under Windows

rwgk:
Observed:

```
> assert delay_ms <= elapsed_time_ms < delay_ms + generous_tolerance
E assert 500.0 <= 490.03173828125
test_event.py:41: AssertionError
```

The `generous_tolerance` also needs to be applied when checking the lower bound.

### DIFF
--- a/cuda_core/tests/test_event.py
+++ b/cuda_core/tests/test_event.py
@@ -6,6 +6,7 @@
 # this software and related documentation outside the terms of the EULA
 # is strictly prohibited.
 
+import os
 import time
 
 import pytest
@@ -37,8 +38,13 @@ def test_timing(init_cuda, enable_timing):
         # We only want to exercise the __sub__ method, this test is not meant
         # to stress-test the CUDA driver or time.sleep().
         delay_ms = delay_seconds * 1000
-        generous_tolerance = 20
-        assert delay_ms <= elapsed_time_ms < delay_ms + generous_tolerance
+        if os.name == "nt":  # noqa: SIM108
+            # Windows timer resolution is typically limited to 15.6 ms by default.
+            generous_tolerance = 100
+        else:
+            # Most modern Linux kernels have a default timer resolution of 1 ms.
+            generous_tolerance = 20
+        assert delay_ms - generous_tolerance <= elapsed_time_ms < delay_ms + generous_tolerance
     else:
         with pytest.raises(RuntimeError) as e:
             elapsed_time_ms = e2 - e1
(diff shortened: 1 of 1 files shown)


cuda_core/tests/test_event.py
```diff
@@ -37,8 +38,13 @@ def test_timing(init_cuda, enable_timing):
         # We only want to exercise the __sub__ method, this test is not meant
         # to stress-test the CUDA driver or time.sleep().
         delay_ms = delay_seconds * 1000
-        generous_tolerance = 20
-        assert delay_ms <= elapsed_time_ms < delay_ms + generous_tolerance
+        if os.name == "nt":  # noqa: SIM108
+            # Windows timer resolution is typically limited to 15.6 ms by default.
```
leofang: As discussed offline this seems only applicable to Python 3.10 and below; starting 3.11 the built-in `sleep` has a much better resolution.

leofang (dismissed):

cuda_core/tests/test_event.py
leofang: This could also explain why we did not catch it in the CI, as we only test PY312 on Windows due to budget constraints...

rwgk: I changed the comment as discussed offline.

leofang:
Since the CI was green and the last commit is comment-only, go ahead and admin-merge it.

cuda_core/tests/test_event.py
```diff
@@ -37,8 +38,13 @@ def test_timing(init_cuda, enable_timing):
         # We only want to exercise the __sub__ method, this test is not meant
         # to stress-test the CUDA driver or time.sleep().
         delay_ms = delay_seconds * 1000
-        generous_tolerance = 20
-        assert delay_ms <= elapsed_time_ms < delay_ms + generous_tolerance
+        if os.name == "nt":  # noqa: SIM108
```
keenan-simpson, line 41: what do we gain by giving different tolerances to each OS? could we set both to 100?

rwgk: Yes, but this is a little stricter; I figured Leo believes that's better.

In the long run, assuming that everything is stable (cuda.core released, and this usually works), a tighter tolerance could uncover new bugs elsewhere. So we'd contribute to the greater good.

keenan-simpson: I didn't even notice that it was already merged ha. Looks good to me, I was mostly just curious.

PR #541, opened 2025-03-31
Add section Preparing kernel arguments

### LINKED ISSUE #127: Document best practices in preparing arguments for `cuLaunchKernel`
@gigony complained the following example is unclear compared to CuPy's RawKernel, and I agree.
https://github.com/NVIDIA/cuda-python/blob/e1e332564c48db556212d59262a149b1a63285e8/docs_src/source/overview.md?plain=1#L208-L223
Specifically, how to pass typed pointers and scalars is really unclear.

This also raises the question: It seems NumPy has to be a run-time dependency of cuda-python (which is completely fine, since it is impossible that users have GPU projects installed to an environment where NumPy does not exist, we just need to ensure this intent is expressed as a package dependency).

For comparison, here's `cupy.RawKernel`'s documentation
https://docs.cupy.dev/en/stable/user_guide/kernel.html#kernel-arguments

vzhurba01:
Close #127


cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,236 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+{
+    *pi = i;
+    *pf = f;
+    ps->value = s.value;
+}
+"""
+```
+
+The first step is to create array objects with types corresponding to your kernel arguments.
+NumPy's user guide for [Kernel arguments](https://docs.cupy.dev/en/stable/user_guide/kernel.html#kernel-arguments) describes how primitive Numpy types correspond to kernel types,
```
kkraus14: Should this be cupy's user guide? In general we shouldn't need to link off to cupy documentation here, especially if we eventually want cupy to be depend on cuda-python in the future.

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,236 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+    kernel,
+    1, 1, 1,  # grid dim
+    1, 1, 1,  # block dim
+    0, stream,  # shared mem and stream
+    kernelParams=kernelParams,
+    extra=0,
+))
+```
+
+### Using ctypes
```
kkraus14, line 443: It would be good to have an example of using Python buffer protocol (https://docs.python.org/3/c-api/buffer.html) compatible objects as well, though I'm not sure how we'd specify the struct type in that situation.

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,236 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+    kernelParams=(kernelValues, kernelTypes),
+    extra=0,
+))
+```
+
+### CUDA objects
+
+Certain CUDA kernels use native CUDA types as their parameters such as `cudaTextureObject_t`. These types require special handling since they're neither a primative ctype nor a custom user type. Since `cuda.bindings` exposes each of them as Python clases, they they each implement `getPtr()` and `__int__()`. These two callables used to support the NumPy and ctypes approach. The difference between each call is further described under [Tips and Tricks](https://nvidia.github.io/cuda-python/cuda-bindings/latest/tips_and_tricks.html#).
+
+For this example, lets use the `transformKernel` from [examples/0_Introduction/simpleCubemapTexture_test.py](https://github.com/NVIDIA/cuda-python/blob/73fa3c035c1111fa8e82b8ba24ed0e33eb65e6c8/cuda_bindings/examples/0_Introduction/simpleCubemapTexture_test.py):
```
kkraus14: Should this be a relative link instead of a permalink?

cuda_bindings/docs/source/overview.md
vzhurba01: Changed to cupy. Would that future dependency be specific to cuda.core? I'd imagine that cuda.core will have a more natural argument prep work than what cuda.bindings gives today.

cuda_bindings/docs/source/overview.md
vzhurba01: Changed. Yeah that likely makes more sense.

cuda_bindings/docs/source/overview.md
vzhurba01: > The launch API supports [Buffer Protocol](https://docs.python.org/3/c-api/buffer.html) objects, therefore we can pass the array object directly.

I tried to give this example by leveraging Numpy's array objects since they already have the protocol. Were you thinking about having an example of a custom user object that supports a Python buffer protocol as an example?

kkraus14: I don't think we need to address this in this PR to be clear. Just mindful that we may want to make numpy an optional dependency in the future and if that's the case it would be nice to have an example of using buffer protocol without numpy (i.e. using the standard `array` library or something like that).

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,236 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+with a [ctypes](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.ctypes.html#numpy.ndarray.ctypes) data attribute that returns the underlying `void*` pointer value.
+
+By having the final array object contain all pointers, we fulfill the contiguous array requirement:
+
+```python
+kernelParams = np.array([arg.ctypes.data for arg in kernelValues], dtype=np.uint64)
 ```
 
-## Future of CUDA Python
+The launch API supports [Buffer Protocol](https://docs.python.org/3/c-api/buffer.html) objects, therefore we can pass the array object directly.
```
kkraus14, line 430: In this case would it actually use the buffer protocol or would it use dlpack? It's a bit confusing to use numpy as the example for buffer protocol because of this.

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,236 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+    kernelParams=(kernelValues, kernelTypes),
+    extra=0,
+))
+```
+
+### CUDA objects
+
+Certain CUDA kernels use native CUDA types as their parameters such as `cudaTextureObject_t`. These types require special handling since they're neither a primative ctype nor a custom user type. Since `cuda.bindings` exposes each of them as Python clases, they they each implement `getPtr()` and `__int__()`. These two callables used to support the NumPy and ctypes approach. The difference between each call is further described under [Tips and Tricks](https://nvidia.github.io/cuda-python/cuda-bindings/latest/tips_and_tricks.html#).
+
+For this example, lets use the `transformKernel` from [examples/0_Introduction/simpleCubemapTexture_test.py](https://github.com/NVIDIA/cuda-python/blob/main/cuda_bindings/examples/0_Introduction/simpleCubemapTexture_test.py):
```
kkraus14, line 513: The link is still an absolute path here where if we change the repo structure in the future this link could 404. Is there any way we can do some kind of relative path linking here or something similar so that we get the correct version of this example each time the docs are published?

cuda_bindings/docs/source/overview.md
kkraus14: > Would that future dependency be specific to cuda.core?

Not necessarily, but regardless I don't think we should be reliant on linking off to CuPy's documentation for our documentation in `cuda.bindings` or `cuda.core`. We should probably capture the information a user needs for the kernel arguments directly into the `cuda.bindings` docs.

cuda_bindings/docs/source/overview.md
vzhurba01: Sort of, but I don't think it's better. What I can do is:
```
For this example, lets use the `transformKernel` from [examples/0_Introduction/simpleCubemapTexture_test.py](../../examples/0_Introduction/simpleCubemapTexture_test.py):
```

What this does is it downloads a copy of the example at build time and the link them redirects users to:
`/docs/build/html/_downloads/54802b8ecbf7ce3c15adbca71c4048dd/simpleCubemapTexture_test.py`. This looks unrendered and is equivalent to a perma link.

cuda_bindings/docs/source/overview.md
vzhurba01: >https://github.com/NVIDIA/cuda-python/blob/main/cuda_bindings/cuda/bindings/_lib/utils.pyx.in#L58

Buffer protocol since we don't support dlpack.

cuda_bindings/docs/source/overview.md
vzhurba01: I see my confusion. We have a soft reliance on NumPy (as the alternative to ctypes), but I incorrectly treated CuPy and NumPy as one and the same.

I've now added the table for correspondence between NumPy types and kernel types.

cuda_bindings/docs/source/overview.md
kkraus14: Yea, that's definitely not great. Maybe it makes sense to just pull out the kernel from the example into the docs here to avoid having to cross reference?

kkraus14 (approved):
Approving because only remaining comment is related to the linking to the example which we can handle in a follow up if desired

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+
+def main():
+    ...
+    d_data = checkCudaErrors(cudart.cudaMalloc(size))
+    width = 64
+    tex = checkCudaErrors(cudart.cudaCreateTextureObject(texRes, texDescr, None))
+    ...
+```
+
+For NumPy, we can convert these CUDA types by leveraging the `__int__()` call to fetch the adress of the underlining `cudaTextureObject_t` C object and wrapping it in a NumPy object array of type `np.uint64`:
```
vzhurba01:
```suggestion
For NumPy, we can convert these CUDA types by leveraging the `__int__()` call to fetch the address of the underlying `cudaTextureObject_t` C object and wrapping it in a NumPy object array of type `np.uint64`:
```

cuda_bindings/docs/source/overview.md
vzhurba01: The content of the kernel doesn't actually matter for this doc, only the signature, hence why I rewrote it bellow with the shortened form:
```
simpleCubemapTexture = """\
extern "C"
__global__ void transformKernel(float *g_odata, int width, cudaTextureObject_t tex)
{
    ...
}
"""
```
The kernel is too large in my opinion to be shared here.

kkraus14: Resolving this, I think this is more than good enough at this point.

cuda_bindings/docs/source/overview.md
```diff
@@ -193,20 +189,23 @@ checkCudaErrors(driver.cuMemcpyHtoDAsync(
...
 dY = np.array([int(dYclass)], dtype=np.uint64)
 dOut = np.array([int(dOutclass)], dtype=np.uint64)
+```
 
+The launch API `cuLaunchKernel` also expects a pointer input for the argument list
+but this time it's of type `void**`. What this means is that our argument list needs to
+be a contiguous array of `void*` elements, where each element is the pointer to a kernel
+argument. Since we already prepared each of our arguments into a `np.array` type, the
+construction of our final contiguous array is done by retrieving the `XX.ctypes.data`
+of each kernel argument.
```
leofang:
```suggestion
be a contiguous array of `void*` elements, where each element is the pointer to a kernel
argument on either host or device. Since we already prepared each of our arguments into a `np.array` type, the
construction of our final contiguous array is done by retrieving the `XX.ctypes.data`
of each kernel argument.
```

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+```python
+kernel_string = """\
+struct testStruct {
+    int value;
+};
+
+extern "C" __global__
+void testkernel(int i, int *pi,
+                float f, float *pf,
+                struct testStruct s, struct testStruct *ps)
```
leofang: nit
```suggestion
typedef struct {
    int value;
} testStruct;

extern "C" __global__
void testkernel(int i, int *pi,
                float f, float *pf,
                testStruct s, testStruct *ps)
```

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+:header-rows: 1
+
+* - NumPy type
+  - Corresponding kernel types
+  - itemsize (bytes)
+* - bool
+  - bool
+  - 1
+* - int8
+  - char, signed char
```
leofang: this may be trivial but it is still good to have when eyeballing the table
```suggestion
  - char, signed char, int8_t
```

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+  - 1
+* - int16
+  - short, signed short
```
leofang: ditto
```suggestion
  - short, signed short, int16_t
```

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+  - 2
+* - int32
+  - int, signed int
```
leofang: ditto
```suggestion
  - int, signed int, int32_t
```

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+  - 4
+* - int64
+  - long long, signed long long
```
leofang:
```suggestion
  - long long, signed long long, int64_t
```

cuda_bindings/docs/source/overview.md
leofang: (ps. I am glad we do not list `long` here...)

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+  - 8
+* - uint8
+  - unsigned char
```
leofang:
```suggestion
  - unsigned char, uint8_t
```

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+  - 1
+* - uint16
+  - unsigned short
```
leofang:
```suggestion
  - unsigned short, uint16_t
```

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+  - 2
+* - uint32
+  - unsigned int
```
leofang:
```suggestion
  - unsigned int, uint32_t
```

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+  - 4
+* - uint64
+  - unsigned long long
```
leofang:
```suggestion
  - unsigned long long, uint64_t
```

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+  - double2, cuDoubleComplex, complex&lt;double&gt;
+  - 16
+```
+
+Furthermore, custom NumPy types can be used to support both platform-dependent types and user-defined structures as kernel arguments.
+
+This example uses the following types:
+* `int` is `np.uint32`
+* `float` is `np.float32`
+* `int*` and `float*` is `np.uint64`
```
leofang: nit: NumPy suggested to use this for pointers
```suggestion
* `int*` and `float*` is `np.intp`
```

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+* `testStruct` is a custom user type `np.dtype([("value", np.int32)])`
```
leofang: To be safe when the struct is more complex, we want to set `align=True`
```suggestion
* `testStruct` is a custom user type `np.dtype([("value", np.int32)], align=True)`
```
which would give us the same layout as a C compiler would produce.

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+
+Note how both of the pointers are `np.uint64` since the pointers values are always a representation of an address space.
```
leofang: ditto
```suggestion
Note how both of the pointers are `np.intp` since the pointers values are always a representation of an address space.
```

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+
+Putting it all together:
+```python
+# Define a custom type
+testStruct = np.dtype([("value", np.int32)])
```
leofang:
```suggestion
testStruct = np.dtype([("value", np.int32)], align=True)
```

leofang (commented):
Half way done, sorry for my slowness

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+    ctypes.c_int,
+    ctypes.c_void_p,
+    ctypes.c_float,
+    ctypes.c_void_p,
+    None,
+    ctypes.c_void_p,
+)
+```
+
+Values that are assigned a `None` type has a special meaning:
```
leofang:
```suggestion
Values that are set to `None` have a special meaning:
```

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+
+1. The value supports a callable `getPtr` that returns the pointer address of the underlining C object address (e.g. all CUDA C types that are exposed to Python as Python classes)
+2. The value is an instance of `ctypes.Structure`
+3. The value is an `Enum`
+
+In each of these three cases, the API call will fetch the underlining pointer value and construct a contiguous array with other kernel parameters.
```
leofang: nit
```suggestion
In all three cases, the API call will fetch the underlying pointer value and construct a contiguous array with other kernel parameters.
```

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+
+Putting it all together:
+```python
+# Define a custom type
+testStruct = np.dtype([("value", np.int32)], align=True)
+
+# Allocate host memory
+pInt_host = checkCudaErrors(cudart.cudaHostAlloc(np.dtype(np.int32).itemsize, cudart.cudaHostAllocMapped))
+pFloat_host = checkCudaErrors(cudart.cudaHostAlloc(np.dtype(np.float32).itemsize, cudart.cudaHostAllocMapped))
+pStruct_host = checkCudaErrors(cudart.cudaHostAlloc(testStruct.itemsize, cudart.cudaHostAllocMapped))
```
leofang: I suggest that we either switch to just use device memory (which is the most common situation for kernels taking pointers as arguments), or make it clear that we're allocating pinned/page-locked (host) memory here that are accessible by GPU kernels.

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+                float f, float *pf,
+                struct testStruct s, struct testStruct *ps)
+{
+    *pi = i;
+    *pf = f;
+    ps->value = s.value;
+}
+"""
+```
+
```
leofang: nit: maybe we just omit it and say we use the same example from the previous section?

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+The ctypes approach treats the `kernelParams` argument as a pair of two tuples: `kernel_values` and `kernel_types`.
+
+* `kernel_values` contain Python values to be used as an input to your kernel
+* `kernel_types` contain the data types that your kernel_values should be converted into
+
+The ctypes [Fundamental data types](https://docs.python.org/3/library/ctypes.html#fundamental-data-types) documentation describes the compatibility between different Python types and C types.
```
leofang: nit
```suggestion
The ctypes [fundamental data types](https://docs.python.org/3/library/ctypes.html#fundamental-data-types) documentation describes the compatibility between different Python types and C types.
```

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+
+```python
+# Define a custom type
+class testStruct(ctypes.Structure):
+    _fields_ = [("value", ctypes.c_int)]
+
+# Allocate host memory
+pInt_host = checkCudaErrors(cudart.cudaHostAlloc(ctypes.sizeof(ctypes.c_int), cudart.cudaHostAllocMapped))
+pFloat_host = checkCudaErrors(cudart.cudaHostAlloc(ctypes.sizeof(ctypes.c_float), cudart.cudaHostAllocMapped))
+pStruct_host = checkCudaErrors(cudart.cudaHostAlloc(ctypes.sizeof(testStruct), cudart.cudaHostAllocMapped))
```
leofang: ditto

leofang (commented):
This is great improvement, thanks Vlad! I've completed my review.

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+                testStruct s, testStruct *ps)
+{
+    *pi = i;
+    *pf = f;
+    ps->value = s.value;
+}
+"""
+```
+
+The first step is to create array objects with types corresponding to your kernel arguments. Primative Numpy types have the following corresponding kernel types:
```
rwgk: typo: **Primitive** NumPy types

NumPy is usually written with a capital P

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+
+Furthermore, custom NumPy types can be used to support both platform-dependent types and user-defined structures as kernel arguments.
+
+This example uses the following types:
+* `int` is `np.uint32`
+* `float` is `np.float32`
+* `int*`, `float*` and `testStruct*` is `np.intp`
+* `testStruct` is a custom user type `np.dtype([("value", np.int32)], align=True)`
+
+Note how both of the pointers are `np.intp` since the pointers values are always a representation of an address space.
```
rwgk: "all three" (instead of "both")?

(maybe I'm confused what "both" is referring to?)

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+    1, 1, 1,  # block dim
+    0, stream,  # shared mem and stream
+    kernelParams=(kernelValues, kernelTypes),
+    extra=0,
+))
+```
+
+### CUDA objects
+
+Certain CUDA kernels use native CUDA types as their parameters such as `cudaTextureObject_t`. These types require special handling since they're neither a primative ctype nor a custom user type. Since `cuda.bindings` exposes each of them as Python clases, they they each implement `getPtr()` and `__int__()`. These two callables used to support the NumPy and ctypes approach. The difference between each call is further described under [Tips and Tricks](https://nvidia.github.io/cuda-python/cuda-bindings/latest/tips_and_tricks.html#).
```
rwgk: Python **classes** (missing s)

there are also two "they" right after

cuda_bindings/docs/source/overview.md
rwgk: +1

I diffed them, they are subtly different, and as a non-expert I'm wondering if that's somehow important (I guess not).

```diff
 kernel_string = """\
-typedef struct {
+struct testStruct {
     int value;
-} testStruct;
+};
 
 extern "C" __global__
 void testkernel(int i, int *pi,
                 float f, float *pf,
-                testStruct s, testStruct *ps)
+                struct testStruct s, struct testStruct *ps)
 {
```

rwgk (commented):
I spotted some nits.

cuda_bindings/docs/source/overview.md
leofang: Yeah this is another reason we should keep one source of truth. This diff was due to my previous suggestion: https://github.com/NVIDIA/cuda-python/pull/541#discussion_r2026003300

cuda_bindings/docs/source/overview.md
leofang:
```suggestion
The first step is to create array objects with types corresponding to your kernel arguments. Primitive NumPy types have the following corresponding kernel types:
```

cuda_bindings/docs/source/overview.md
leofang:
```suggestion
Certain CUDA kernels use native CUDA types as their parameters such as `cudaTextureObject_t`. These types require special handling since they're neither a primitive ctype nor a custom user type. Since `cuda.bindings` exposes each of them as Python classes, they each implement `getPtr()` and `__int__()`. These two callables used to support the NumPy and ctypes approach. The difference between each call is further described under [Tips and Tricks](https://nvidia.github.io/cuda-python/cuda-bindings/latest/tips_and_tricks.html#).
```

cuda_bindings/docs/source/overview.md
```diff
@@ -291,22 +290,286 @@ nsys profile -s none -t cuda --stats=true <executable>
...
+  - double2, cuDoubleComplex, complex&lt;double&gt;
+  - 16
+```
+
+Furthermore, custom NumPy types can be used to support both platform-dependent types and user-defined structures as kernel arguments.
+
+This example uses the following types:
+* `int` is `np.uint32`
+* `float` is `np.float32`
+* `int*`, `float*` and `testStruct*` is `np.intp`
```
leofang:
```suggestion
* `int*`, `float*` and `testStruct*` are `np.intp`
```

cuda_bindings/docs/source/overview.md
leofang:
```suggestion
Note how all three pointers are `np.intp` since the pointer values are always a representation of an address space.
```

cuda_bindings/docs/source/overview.md
vzhurba01: Done. Switched to device since being on host is irrelevant for this example.

cuda_bindings/docs/source/overview.md
vzhurba01: Removed and said we use same example as previous section.

leofang (approved):

leofang:
Thanks, Vlad and Keith/Ralf!

PR #545, opened 2025-04-02
Import cudart symbols properly

leofang:
Currently I hand-modify this, but we should be able to autogen it.


vzhurba01:
I can apply the fix in autogen and push a commit to this PR.

vzhurba01:
I'll need https://github.com/NVIDIA/cuda-python/pull/544 to be merged first for this to be a clean commit though.

leofang:
This is done as part of #597. Thanks, @vzhurba01!

PR #554, opened 2025-04-10
os.remove on Windows could fail if files are still in use

### LINKED ISSUE #535: `os.remove` on Windows could fail if files are still in use
This code
https://github.com/NVIDIA/cuda-python/blob/1c6f3bca5458bd4575ac77f7d77891a494b42faf/cuda_core/tests/conftest.py#L63-L66
could fail on Windows if the module is still loaded (and therefore the module file is in use):
> On Windows, attempting to remove a file that is in use causes an exception to be raised

https://docs.python.org/3/library/os.html#os.remove

As per suggestion by @rwgk we should move cleanup code from test shutdown to test setup time.

kshitiz305:
## Description

Closes #535

## Checklist

### DIFF
--- a/cuda_core/tests/conftest.py
+++ b/cuda_core/tests/conftest.py
@@ -57,10 +57,19 @@ sys.path.append(os.getcwd())
 
 @pytest.fixture(scope="session", autouse=True)
 def clean_up_cffi_files():
-    yield
+    # Clean up files from previous runs before the session starts
     files = glob.glob(os.path.join(os.getcwd(), "_cpu_obj*"))
     for f in files:
         try:  # noqa: SIM105
             os.remove(f)
         except FileNotFoundError:
             pass  # noqa: SIM105
+        except PermissionError:
+            # On Windows, removing a file in use raises PermissionError
+            # We can ignore this during setup, as the goal is just to
+            # clean up leftovers from *previous* runs. If a file is
+            # still in use now, it's likely from an ongoing process
+            # unrelated to the test session about to start.
+            pass
+    yield
+    # No cleanup needed after yield anymore


cuda_core/tests/conftest.py
```diff
@@ -57,10 +57,19 @@ def deinit_cuda():
...
             pass  # noqa: SIM105
+        except PermissionError:
+            # On Windows, removing a file in use raises PermissionError
+            # We can ignore this during setup, as the goal is just to
+            # clean up leftovers from *previous* runs. If a file is
+            # still in use now, it's likely from an ongoing process
+            # unrelated to the test session about to start.
+            pass
+    yield
+    # No cleanup needed after yield anymore
```
rwgk, line 75: That's a great move.

What do you think about adding something like this?

```
import logging

logger = logging.getLogger(__name__)

...

        except PermissionError as e:
            logger.warning(f"Could not remove file {f!r}: {e}")
```

That way, if a file is in use unexpectedly, we’ll at least have breadcrumbs.

We should probably also log the `FileNotFoundError`.

cuda_core/tests/conftest.py
```diff
@@ -57,10 +57,19 @@ def deinit_cuda():
 
 @pytest.fixture(scope="session", autouse=True)
 def clean_up_cffi_files():
-    yield
+    # Clean up files from previous runs before the session starts
     files = glob.glob(os.path.join(os.getcwd(), "_cpu_obj*"))
```
jakirkham, line 61: Because of issues like the Windows cleanup issue, would highly recommend using [pytest's builtin temporary directory management fixtures]( https://docs.pytest.org/en/stable/how-to/tmp_path.html )

This will handle all the directory creation and cleanup for us with less fuss

rwgk: +1 — if we have the free bandwidth

jakirkham: Simpler, more reliable cleanup of the test directory will lead to bandwidth savings in the long term ;)

leofang:
Hi @kshitiz305, a quick update: We've recently opened up for external contributions to `cuda.core` after changing its license to Apache 2.0 (#583). If you're still interested in pushing this PR through the finish line, feel free to reopen it! 🙂

cuda_core/tests/conftest.py
leofang: I've explained this in a past group meeting, repeating here for wider visibility. pytest fixtures like `tmp_path` etc unfortunately do not work. The temp files are generated by CFFI inside a code sample ([here](https://github.com/NVIDIA/cuda-python/blob/c21613bbcb5f59067105f586115771160219a642/cuda_core/examples/strided_memory_view.py#L65-L76)), and we have no way to influence where it places the modules (without making the code sample overly complicated) I am afraid. I'd be happy to be proven wrong!

PR #393, opened 2025-01-14
Minor CI improvements 

leofang:
- exclude redundant test jobs (close #392) 
- ensure build jobs pass
   - we could not guarantee this before #271 was fixed; now that it can be guaranteed, let's enforce it to avoid waiting in the self-runner queue while some builds still fail (due to network connection issues)
- enable Windows parallel builds
   - again fixing #271 allows this
- remove unused id-token permission
- ~~factor out the `gh` installation step~~
   - ~~this means `gh` is installed unconditionally instead of as needed, but it is small & quick so I don't worry too much~~
- ensure folders are empty

Issue #328, opened 2024-12-29
CI: Centralize all runner usage in a json file


leofang:
Currently our choices of CI runners (either GH- or self- hosted) are hard-coded in the workflows' `runs-on:` field. It might be better to consolidate them in a single json file for easier update and audit.

leofang:
assigned to Marcus for triaging/prioritizing

cryos:
This was done in #889

Issue #280, opened 2024-12-09
CI: Test conda-based workflows


leofang:
1. Build conda packages using the public recipe (xref: https://github.com/NVIDIA/cuda-python/issues/273)
2. Test by creating a conda environment and installing the build artifact and its dependencies

leofang:
@jakirkham suggested to use Miniforge GHA: https://github.com/NVIDIA/cuda-python/pull/289#discussion_r1883278334, good to know it exists! 🙏

rparolin:
Closing as done. Conda packages are built and tested in a conda environment by `.github/workflows/ci-pixi-source-test.yml` (#2185, foundation #1226) — PR-tier smoke plus a nightly GPU cron.

Bullet 1's "public recipe" is satisfied in spirit rather than literally: this repo builds via `pixi-build-python`, while the public recipe lives in `conda-forge/cuda-python-feedstock` and is built by conda-forge CI (alignment RFC #273 closed completed). If we want the feedstock recipe exercised from this repo's CI, that is a new and narrower issue rather than a reason to keep this one open.

PR #555, opened 2025-04-10
Refactor the CI matrix

### LINKED ISSUE #576: CI: Refactor the CI matrix
Take a pass at refactoring the CI matrix used for PRs with the goal of adding some grouping to related builds/tests and making the test matrix more configurable to enable CI and nightly test matrix options. Ideally this enables partial matrix expansion on some elements and having a wide variety of coverage while keeping test times under control.

The driving idea for grouping is to have build/test architectures at the top-level, and then group variations on other things such as Python version, local versus wheels CTK, driver versions and GPUs below that.

- Build linux-64
  - py3.9
  - py3.12
  - py3.13
- Build linux-aarch64
  - py3.9
  - py3.12
  - py3.13
- Build win-64
  - py3.9
  - py3.12
- Test linux-64
  - py3.9, 11.8.0, local, l4
  - py3.9, 12.8.0, wheels, l4
  - py3.13, 12.8.0, local, H100
- Test linux-aarach64
  - ...

cryos:
## Description

Refactor the CI matrix - closes #576.

Refactor the way the CI matrix is composed to be layered, add proxy cache to improve Python package download times, offer a concept of pull request and nightly matrix for the tests. 

## Checklist


.github/workflows/build.yml
leofang, line 1: I haven't checked the diff of this PR carefully but this seems to be on the right direction! I've been wanting to factor out the build parts from `build-and-test.yml` to `build-wheel.yml` (the test parts were already done and are now in `test-wheel.yml`).

cryos: Thanks Leo, just working through test matrix proposals now. If you want to discuss this in more detail I am more than happy to, just trying a few things out and scoping out what fits well right now along with trying to offer better grouping/labeling.

.github/workflows/build.yml
```diff
@@ -0,0 +1,188 @@
...
+        continue-on-error: true
+        # Skip the cache on Windows nodes outside of our org.
+        if: ${{ inputs.host-platform != 'win-64' }}
+
+      - name: Set up Python
+        if: ${{ startsWith(inputs.host-platform, 'linux') }}
+        id: setup-python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
```
kkraus14: Does this need to use the python version matrix?

.github/workflows/build.yml
```diff
@@ -0,0 +1,188 @@
...
+          if-no-files-found: error
+
+      # We only need/want a single pure python wheel, pick linux-64 index 0.
+      - name: Build and check cuda-python wheel
+        if: ${{ strategy.job-index == 0 && inputs.host-platform == 'linux-64' }}
+        run: |
+          pushd cuda_python
+          pip wheel -v --no-deps .
+          twine check *.whl
+          popd
```
kkraus14: Is there a way to do this with `cibuildwheel` just to be consistent?

.github/workflows/test-wheel-linux.yml
```diff
@@ -3,32 +3,137 @@ name: "CI: Test wheels"
...
+          special_runner=""
+          if [[ "${ARCH}" == "amd64" ]]; then
+            special_runner="- { ARCH: ${ARCH}, PY_VER: '3.13', CUDA_VER: '12.8.0', LOCAL_CTK: '1', GPU: 'H100', DRIVER: 'latest' }"
+          fi
+
+          # Please keep the matrices sorted in ascending order by the following:
+          #
+          #     [PY_VER, CUDA_VER, LOCAL_CTK, GPU, DRIVER]
+          #
+          export MATRICES="
```
kkraus14, line 63: We don't really have much interaction between Python version and driver version, where we don't need the cross product of them, just coverage of each. Similarly, we don't really have much interaction between Python version and CTK install type, where we don't need the cross product of them, just coverage of each.

I think the coverage matrix we want to hit is `CUDA_VER` x `LOCAL_CTK` x `DRIVER` which would give us 12x jobs per `ARCH` today that we can spread the `PY_VER` to have coverage. When we move to 13.x, we'll then have 4 `CUDA_VER` we want to test as opposed to 3 today, which would give us 24x jobs per ARCH which I believe can still be handled reasonably easily.

versions.json
```diff
@@ -0,0 +1,7 @@
+{
```
kkraus14: What is this file needed for?

.github/workflows/build.yml
cryos: It was not, and does not seem to need to. Good call out as I had meant to ask and forgotten - @leofang https://github.com/NVIDIA/cuda-python/blob/main/.github/workflows/build-and-test.yml#L56-L61 is where I got this from and it was said the matrix version there, but then set up 3.12 everytime.

.github/workflows/build.yml
cryos: This was a straight up copy of the existing approach, I can look into it, but had focused more on porting.

.github/workflows/test-wheel-linux.yml
cryos: This is a great point - I wanted to get opinions. This is a literal copy of what was happening but I thought it would be great to cut it down a fair bit. It sounds like you are onboard. I figured old and new driver to have some coverage, some local versus wheel. I would gladly reduce it, I was also thinking that nightly might do more, and PRs could do less.

versions.json
cryos: This is read in at the start to share the CUDA build version, it was an initial stab at offering a central spot to write that down once, I can back it out, move it, etc. I had seen a desire to have a JSON file store some of these things rather than having it in several CI workflows. I think it might be useful but agree it is a tiny file with a single version in it right now.

.github/workflows/build.yml
cryos: Checking it looks like they do not support pure python, e.g. https://github.com/pypa/cibuildwheel/issues/1633 among other issues I came across.

.github/workflows/test-wheel-linux.yml
cryos: Not sure I hit exactly what you were thinking, but I got it to 12 per architecture aside from the special one for `amd64` and have a mixture to cover different elements of the matrix. This could be tweaked up or down for sure or mixed with some other variations (part of why I wanted to structure it this way).

kkraus14: I was thinking to make sure we have coverage on each combination of `CUDA_VER: {'11.8.0', '12.0.1', '12.8.0'}` x `LOCAL_CTK: {'0', '1'}` x `DRIVER: {'earliest', 'latest'}`

Your update looks like the `DRIVER: 'earliest'` is only covered with `LOCAL_CTK: '0'` and `CUDA_VER='11.8.0'` currently. 

Unless I'm misunderstanding and we need to cover something like a third driver version as well where it's the earliest 12.x supported version for example.

cryos: Ah, there was no existing support for changing the driver - it wasn't clear what we wanted to cover there and I figured adding something. I proved  `earliest` does not work with CUDA `12.8.0` at all in an earlier run, so that can't be added. I get what you mean, cover as much of `CUDA_VER`, `LOCAL_CTK` and `DRIVER` as possible spread across Python versions as you see little to no interaction there. I can tweak but I won't add `earliest` back with CUDA `12.8.0` as it fails at runtime. Additionally CUDA `12.0.1` does not work for `LOCAL_CTK` set to `0`.

versions.json
kkraus14: Given this is currently only used for CI, it would be nice to put it in a folder away from the root of the directory then.

.github/workflows/test-wheel-linux.yml
kkraus14: > I proved `earliest` does not work with CUDA `12.8.0` at all in an earlier run, so that can't be added.

Yea, this is expected since `earliest` is a CUDA 11 driver that does not support CUDA 12. Ideally we would have a test that uses the CUDA 12.0 driver with our CUDA 12.8 build.

Doesn't need to be covered in this PR, but we should probably support something like `cuda11_earliest`, `cuda11_latest`, `cuda12_earliest`, `cuda12_latest` for the driver versions.

> Additionally CUDA 12.0.1 does not work for LOCAL_CTK set to 0.

This is a bit surprising. Any idea why?

.github/workflows/build-and-test.yml
```diff
@@ -1,320 +0,0 @@
...
-          - "12.8.0"
-          - "12.0.1"
-          - "11.8.0"
-        local-ctk:
-          - 1  # use mini CTK
-          - 0  # use CTK wheels
-        runner:
-          - default
-        exclude:
-          # To test this combo would require nontrivial installation steps.
```
cryos, line 236: @kkraus14 this is the exclusion - I think a practical matter of packaging and something I preserved for no 12.0.1 with maybe not having wheels that far back, or being in flux?

.github/workflows/test-wheel-linux.yml
cryos: > > I proved `earliest` does not work with CUDA `12.8.0` at all in an earlier run, so that can't be added.
> 
> Yea, this is expected since `earliest` is a CUDA 11 driver that does not support CUDA 12. Ideally we would have a test that uses the CUDA 12.0 driver with our CUDA 12.8 build.
> 
> Doesn't need to be covered in this PR, but we should probably support something like `cuda11_earliest`, `cuda11_latest`, `cuda12_earliest`, `cuda12_latest` for the driver versions.

Yes - we need support on the runner side, this is something that is on the list for future additions to the runners.
> 
> > Additionally CUDA 12.0.1 does not work for LOCAL_CTK set to 0.
> 
> This is a bit surprising. Any idea why?

This [line here](https://github.com/NVIDIA/cuda-python/pull/555/files#diff-bc668a2c9f2299cef15b222055b4b4d5311646caec2e7610e540cee18ae9b948L232) is where it was excluded in the original matrix.

versions.json
cryos: Sounds good - I nearly did that, just create a `ci` folder, or preferences on the path?

.github/workflows/build.yml
leofang: @cryos Back then there was a nasty setup-python bug (see the comment right above the lines you linked to) that made the build steps fail. That was why the old build pipeline needed to lock to PY312. But when actually building the wheels (using the cibuildwheel action), we honor the Python version set in the matrix. 
https://github.com/NVIDIA/cuda-python/blob/c71ed39d9fee8454a286a870b6b0a7cae44a07be/.github/workflows/build-and-test.yml#L69-L77
I haven't checked the latest new build pipeline carefully -- will try to get to it tomorrow.

.github/workflows/test-wheel-linux.yml
leofang: > > > Additionally CUDA 12.0.1 does not work for LOCAL_CTK set to 0.
> > 
> > 
> > This is a bit surprising. Any idea why?
> 
> This [line here](https://github.com/NVIDIA/cuda-python/pull/555/files#diff-bc668a2c9f2299cef15b222055b4b4d5311646caec2e7610e540cee18ae9b948L232) is where it was excluded in the original matrix.

Reason: We have no way to express "I want to install the CUDA wheels associated with a certain CTK major.minor(.patch)" (https://github.com/NVIDIA/cuda-python/pull/393#discussion_r1917266401), until Emma's metapackage lands (soon, @rwgk has been testing them internally).

versions.json
leofang: Whoever taking over the CI development/maintenance gets to make the call :)

leofang: xref: https://github.com/NVIDIA/cuda-python/issues/328, it'd be nice (not for this PR, but let's keep this in mind) in the future to encode the CI matrix/runner/etc in a json file instead of embedding this information in the workflow code, if possible (sometimes it is surprisingly hard, e.g. per-arch `matrix:` that we discussed before).

.github/workflows/build-and-test.yml
leofang: https://github.com/NVIDIA/cuda-python/pull/555#discussion_r2061982698

.github/workflows/ci.yml
```diff
@@ -0,0 +1,119 @@
+name: "CI"
```
leofang, line 7: Important: As noted in the original file, this name is referred to in the test stages to query the run ID, e.g. here: https://github.com/NVIDIA/cuda-python/blob/2aca3064514855fb0d9766880faf7ab623bdccc9/.github/workflows/test-wheel-linux.yml#L129
and here:
https://github.com/NVIDIA/cuda-python/blob/2aca3064514855fb0d9766880faf7ab623bdccc9/.github/workflows/test-wheel-windows.yml#L113
so either we make a global change, or we keep the original name intact:
```suggestion
# Note: This name is referred to in the test job, so make sure any changes are sync'd up!
name: "CI: Build and test"
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,119 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}
+  cancel-in-progress: true
```
leofang, line 11: btw I don't think it is in-scope for this PR, just something that came to my mind. 

This PR is in fact v3 of the CI. When I refactored to v2, one thing I noticed is that it has less concurrency compared to v1 (and I think v3 inherits from v2). In v1, when a build job finishes, the dependent test job starts immediately (and potentially overlaps with other build jobs). But in v2, all build jobs have to finish, synchronize, and then all test jobs start to queue up. I think this is because of mismatching number of jobs between build and test in v2, and of the use of `needs:` to declare job dependency. If you have any idea for restoring the v1-like behavior, we should discuss/implement in a future PR.

.github/workflows/build.yml
leofang: Let's rename this to `build-wheel.yml`. In the future we should also add `build-conda.yml` (#280).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,119 @@
...
+      matrix:
+        host-platform:
+          - linux-64
+          - linux-aarch64
+          - win-64
+    name: Build ${{ matrix.host-platform }}, CUDA ${{ needs.ci-vars.outputs.CUDA_BUILD_VER }}
+    if: ${{ github.repository_owner == 'nvidia' }}
+    secrets: inherit
+    uses:
+      ./.github/workflows/build.yml
```
leofang: ditto
```suggestion
      ./.github/workflows/build-wheel.yml
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,188 @@
...
+      - name: Checkout ${{ github.event.repository.name }}
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Setup proxy cache
+        uses: nv-gha-runners/setup-proxy-cache@main
+        continue-on-error: true
+        # Skip the cache on Windows nodes outside of our org.
+        if: ${{ inputs.host-platform != 'win-64' }}
```
leofang: ❤️

.github/workflows/build.yml
leofang:
```suggestion
          # WAR: setup-python is not relocatable...
          # see https://github.com/actions/setup-python/issues/871
          python-version: "3.12"
```

.github/workflows/build.yml
leofang: Yeah in addition to cibuildwheel support the bigger problem was that we only need one single pure Python wheel to support all platforms. We could in theory build this in parallel and block redundant builds during the upload stage (similar to conda upload's handling of packages with the same build string). However, upload-artifact is not sophisticated enough to allow this
https://github.com/NVIDIA/cuda-python/blob/2aca3064514855fb0d9766880faf7ab623bdccc9/.github/workflows/build-and-test.yml#L173-L174

.github/workflows/test-wheel-linux.yml
```diff
@@ -3,32 +3,126 @@ name: "CI: Test wheels"
...
-        type: string
-        required: true
-      local-ctk:
-        type: string
-        required: true
-      runner:
+      matrix_filter:
         type: string
-        required: true
+        default: "."
```
leofang, line 21: Q: I don't see `matrix_filter` currently passed from the caller, I assume this is reserved for future extension? What's our plan for using this?

.github/workflows/test-wheel-linux.yml
```diff
@@ -3,32 +3,126 @@ name: "CI: Test wheels"
...
+
+          MATRIX="$(
+            yq -n -o json 'env(TEST_MATRIX)' | \
+            jq -c '${{ inputs.matrix_filter }} | if (. | length) > 0 then {include: .} else "Error: Empty matrix\n" | halt_error(1) end'
+          )"
+
+          echo "MATRIX=${MATRIX}" | tee --append "${GITHUB_OUTPUT}"
+
   test:
+    name: py${{ matrix.PY_VER }}, ${{ matrix.CUDA_VER }}, ${{ (matrix.LOCAL_CTK == '1' && 'local') || 'wheels' }}, GPU ${{ matrix.GPU }}
```
leofang, line 124: :+1:

.github/workflows/test-wheel-linux.yml
```diff
@@ -3,32 +3,126 @@ name: "CI: Test wheels"
...
+      - name: Validate Test Type
+        run: |
+          if [[ "$BUILD_TYPE" != "pull-request" ]] && [[ "$BUILD_TYPE" != "nightly" ]] && [[ "$BUILD_TYPE" != "branch" ]]; then
+              echo "Invalid build type! Must be one of 'nightly', 'pull-request', or 'branch'."
+              exit 1
+          fi
+      - name: Compute Python Test Matrix
+        id: compute-matrix
+        run: |
+          set -eo pipefail
```
leofang: nit: we could move this to either the job default 
```yaml
    ...
    defaults:
      run:
        shell: bash --noprofile --norc -xeuo pipefail {0}
    steps:
      - name: Validate Test Type
        ...
```
or the workflow default (note that the other job below also set this).

.github/workflows/test-wheel-linux.yml
leofang: To confirm my understanding: The `nightly` CIs will cover the cartesian product of `PY_VER`, `CUDA_VER`, and `LOCAL_CTK`, except for the above case (CUDA 12.0.1 wheels). 

When we enable the `nightly` CI in the next PR, we should also spread it out across more GPU variants (ex: RTX 2080, T4, V100, ..., which we do have in nv-gha-runners).

.github/workflows/test-wheel-windows.yml
```diff
@@ -3,30 +3,81 @@ name: "CI: Test wheels"
...
+      - name: Validate Test Type
+        run: |
+          if [[ "$BUILD_TYPE" != "pull-request" ]] && [[ "$BUILD_TYPE" != "nightly" ]] && [[ "$BUILD_TYPE" != "branch" ]]; then
+              echo "Invalid build type! Must be one of 'nightly', 'pull-request', or 'branch'."
+              exit 1
+          fi
+      - name: Compute Python Test Matrix
+        id: compute-matrix
+        run: |
+          set -eo pipefail
```
leofang: ditto

.github/workflows/test-wheel-windows.yml
```diff
@@ -3,30 +3,81 @@ name: "CI: Test wheels"
...
+            pull-request:
+              - { ARCH: ${ARCH}, PY_VER: '3.12',  CUDA_VER: '11.8.0', LOCAL_CTK: '0' }
+              - { ARCH: ${ARCH}, PY_VER: '3.12',  CUDA_VER: '11.8.0', LOCAL_CTK: '1' }
+              - { ARCH: ${ARCH}, PY_VER: '3.12',  CUDA_VER: '12.8.0', LOCAL_CTK: '0' }
+              - { ARCH: ${ARCH}, PY_VER: '3.12',  CUDA_VER: '12.8.0', LOCAL_CTK: '1' }
+            nightly:
+              - { ARCH: ${ARCH}, PY_VER: '3.12',  CUDA_VER: '11.8.0', LOCAL_CTK: '0' }
+              - { ARCH: ${ARCH}, PY_VER: '3.12',  CUDA_VER: '11.8.0', LOCAL_CTK: '1' }
+              - { ARCH: ${ARCH}, PY_VER: '3.12',  CUDA_VER: '12.8.0', LOCAL_CTK: '0' }
+              - { ARCH: ${ARCH}, PY_VER: '3.12',  CUDA_VER: '12.8.0', LOCAL_CTK: '1' }
```
leofang: I hope we can do the same for Windows too and cover the full cartesian product `PY_VER` x `CUDA_VER` x `LOCAL_CTK` in `nightly`.

.github/workflows/test-wheel-windows.yml
```diff
@@ -3,30 +3,81 @@ name: "CI: Test wheels"
...
+              exit 1
+          fi
+      - name: Compute Python Test Matrix
+        id: compute-matrix
+        run: |
+          set -eo pipefail
+
+          # Please keep the matrices sorted in ascending order by the following:
+          #
+          #     [PY_VER, CUDA_VER, LOCAL_CTK]
```
leofang, line 46: Note to self: This is a smaller matrix because currently we don't get to choose GPU on Windows.

However, we do get to choose the driver version on Windows ([it is installed at job start time](https://github.com/NVIDIA/cuda-python/blob/2aca3064514855fb0d9766880faf7ab623bdccc9/.github/workflows/install_gpu_driver.ps1#L8)). @cryos when we set up `nightly` perhaps we could make it configurable?

leofang (changes requested):
Thanks a lot, Marcus! LGTM overall, we're heading toward the right direction! I left some comments but only one of them is important to address before merging (CI workflow name).

.github/workflows/ci.yml
cryos: Yes I was looking at this too, and wondering about decoupling the different platforms. I think it would be somewhat manual, but possibly worth it. I can look at this in a follow up PR, it is a shame to wait for the very slowest platform.

.github/workflows/test-wheel-linux.yml
cryos: I want to add some test time pruning on demand, the `.` avoids any filtering and so is a good default to use. I was thinking of adding some support for using labels to test a minimal set in a PR for example instead of editing the workflow files directly.

cryos:
@leofang the one thing you wanted me to change I cannot - see [here](https://github.com/NVIDIA/cuda-python/actions/runs/14841769037/job/41667214065?pr=555#step:10:208). Effectively the old run ID has to be run on the 11.8.0 branch. I think use the shorter name here to retain uniqueness. If we backport this to the 11.8.0 branch then we update the run ID find to use the new name would be my take.

.github/workflows/ci.yml
cryos: They have to be different to be unique, if this is backported to 11.8.x then we can change the corresponding code to use the shorter name once a run on 11.8.x exists.

leofang:
(btw I'll be able to review/merge this PR after we push out `cuda.bindings` 12.9.0 today.)

.github/workflows/ci.yml
leofang: Alternatively we can just backport this line? 😄 Note that the CI in the 11.8.x branch is sufficiently diverged...

cryos: Good point - that would likely be easier!

.github/workflows/build-wheel.yml
kkraus14, line 1: Need SPDX License + Copyright on this file

.github/workflows/build-wheel.yml
```diff
@@ -0,0 +1,190 @@
+on:
+  workflow_call:
+    inputs:
+      host-platform:
+        required: true
+        type: string
+      cuda-version:
+        type: string
+        default: "12.8.0"
```
kkraus14: Do we need a default value here? Could we just make it required and need to be explicitly specified when this workflow is invoked?

.github/workflows/build-wheel.yml
```diff
@@ -0,0 +1,190 @@
...
+        id: setup-python
+        uses: actions/setup-python@v5
+        with:
+          # WAR: setup-python is not relocatable...
+          # see https://github.com/actions/setup-python/issues/871
+          python-version: "3.12"
+
+      - name: Set up MSVC
+        if: ${{ startsWith(inputs.host-platform, 'win') }}
+        uses: ilammy/msvc-dev-cmd@v1
```
kkraus14, line 60: Can we win this to a commit instead of a tag?

.github/workflows/build-wheel.yml
```diff
@@ -0,0 +1,190 @@
...
+          echo "CUDA_BINDINGS_ARTIFACT_NAME=${CUDA_BINDINGS_ARTIFACT_BASENAME}-${{ github.sha }}" >> $GITHUB_ENV
+          echo "CUDA_BINDINGS_ARTIFACTS_DIR=$(realpath "$REPO_DIR/cuda_bindings/dist")" >> $GITHUB_ENV
+          echo "CIBW_BUILD=${CIBW_BUILD}" >> $GITHUB_ENV
+      
+      - name: Dump environment
+        run: |
+          env
+      
+      - name: Build cuda.core wheel
+        uses: pypa/cibuildwheel@v2.22.0
```
kkraus14: +1 pin to commit instead of tag

.github/workflows/build-wheel.yml
```diff
@@ -0,0 +1,190 @@
...
+
+      - name: Set up mini CTK
+        uses: ./.github/actions/fetch_ctk
+        continue-on-error: false
+        with:
+          host-platform: ${{ inputs.host-platform }}
+          cuda-version: ${{ inputs.cuda-version }}
+
+      - name: Build cuda.bindings wheel
+        uses: pypa/cibuildwheel@v2.22.0
```
kkraus14: +1 pin to commit instead of tag

.github/workflows/ci.yml
kkraus14, line 1: Need SPDX + Copyright

ci/versions.json
kkraus14, line 1: Need SPDX License and Copyright for this file

cryos: I don't think I have seen JSON files with SPDX License and Copyright - examples I could crib from? JSON doesn't support comments.

.github/workflows/build-wheel.yml
cryos: Agreed - made it required.

.github/workflows/build-wheel.yml
cryos: Done

.github/workflows/build-wheel.yml
cryos: Done.

.github/workflows/build-wheel.yml
cryos: Done.

leofang (approved):
LGTM! Please hold off merging until we push out 12.9.0 wheels later today 🤞

kkraus14 (dismissed):

leofang (approved):

leofang:
FYI, we've done GitHub & PyPI releases, so feel free to merge anytime!

leofang:
Many thanks, Marcus! Nice refactoring!!

PR #560, opened 2025-04-15
Add CodeQL and Bandit Static Analysis Scans

### LINKED ISSUE #534: CI: Set up a static code analysis tool
We need to run a CI job that executes static analysis tools, starting with CodeQL and increasing the variety of tools on an as-needed basis:
https://docs.github.com/en/code-security/code-scanning/introduction-to-code-scanning/about-code-scanning-with-codeql

kkraus14:
## Description

Resolves #534

Adds scans using both CodeQL and Bandit. Could use some discussion on what level of reporting we wish to have here and when we want to error. I have updated the repo settings to alert on any Security alert severity level and set the Standard alert severity level to "Errors and warnings" as a starting point.

## Checklist

### DIFF
--- a/.github/workflows/bandit.yml
+++ b/.github/workflows/bandit.yml
@@ -5,6 +5,9 @@ on:
     branches:
       - "pull-request/[0-9]+"
       - "main"
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}
+  cancel-in-progress: true
 
 jobs:
   analyze:
--- a/.github/workflows/bandit.yml
+++ b/.github/workflows/bandit.yml
@@ -1,0 +1,16 @@
+name: "Static Analysis: Bandit Scan"
+
+on:
+  push:
+    branches:
+      - "pull-request/[0-9]+"
+      - "main"
+
+jobs:
+  analyze:
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+    steps:
+      - name: Perform Bandit Analysis
+        uses: PyCQA/bandit-action@8a1b30610f61f3f792fe7556e888c9d7dffa52de  # v1.0.0
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -5,6 +5,9 @@ on:
     branches:
       - "pull-request/[0-9]+"
       - "main"
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}
+  cancel-in-progress: true
 
 jobs:
   analyze:
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,7 +1,10 @@
 name: "Static Analysis: CodeQL Scan"
 
-on: workflow_dispatch
-
+on:
+  push:
+    branches:
+      - "pull-request/[0-9]+"
+      - "main"
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}
   cancel-in-progress: true
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,10 +1,7 @@
 name: "Static Analysis: CodeQL Scan"
 
-on:
-  push:
-    branches:
-      - "pull-request/[0-9]+"
-      - "main"
+on: workflow_dispatch
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}
   cancel-in-progress: true
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,36 @@
+name: "Static Analysis: CodeQL Scan"
+
+on:
+  push:
+    branches:
+      - "pull-request/[0-9]+"
+      - "main"
+
+jobs:
+  analyze:
+    name: Analyze (${{ matrix.language }})
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - language: python
+          build-mode: none
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v3
+      with:
+        languages: ${{ matrix.language }}
+        build-mode: ${{ matrix.build-mode }}
+        queries: security-extended
+
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@v3
+      with:
+        category: "/language:${{matrix.language}}"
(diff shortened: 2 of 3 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,36 @@
...
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - language: python
+          build-mode: none
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
```
kkraus14, line 27: Note I didn't pin these to specific commits like I did for Bandit above because I figured that these are GitHub managed where we'd be more secure allowing to float forward to get fixes than pinning to a specific commit. I figure if GitHub gets compromised we're in trouble regardless 😄.

kkraus14:
/ok to test [b8d0441](https://github.com/NVIDIA/cuda-python/pull/560/commits/b8d04414f92de78e780e1c920422eb00f1fe7ec2)

.github/workflows/codeql.yml
leofang: Either way is fine. I dunno if we can set up a dependentbot, if we can then it supports either pinning.

.github/workflows/bandit.yml
```diff
@@ -0,0 +1,16 @@
+name: "Static Analysis: Bandit Scan"
+
+on:
+  push:
+    branches:
+      - "pull-request/[0-9]+"
+      - "main"
+
```
cryos, line 11:
```suggestion
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}
  cancel-in-progress: true

```

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,36 @@
+name: "Static Analysis: CodeQL Scan"
+
+on:
+  push:
+    branches:
+      - "pull-request/[0-9]+"
+      - "main"
+
```
cryos, line 11:
```suggestion
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}
  cancel-in-progress: true

```

cryos (approved):
This looks good to me, I added a proposal for concurrency groups that are pretty simple and should work well, choosing something that should work well for both merges to main and PRs.

cryos:
This needs the addition of bandit to the whitelist as noted in slack.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,39 @@
...
+jobs:
+  analyze:
+    name: Analyze (${{ matrix.language }})
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+
+    strategy:
+      fail-fast: false
```
kkraus14, line 20: Should we do `fail-fast: false` for the bandit scan as well?

.github/workflows/codeql.yml
kkraus14: We probably should set up dependabot or something similar in general

.github/workflows/codeql.yml
cryos: There is no matrix to fail fast on is there?

cryos:
Bandit should be allowed to run now if you want to retry it.

.github/workflows/codeql.yml
kkraus14: 🤦 I misinterpreted the docs to be all workflows instead of job matrix within a workflow. Resolving this.

kkraus14:
@leofang do you want me to add bandit / codeql to pre-commit before we merge this?

kkraus14:
Do not merge. Needs an internal discussion before moving forward.

leofang:
> do you want me to add bandit / codeql to pre-commit before we merge this?

I think it is fine to do it in a separate PR, so we only need to resolve the internal discussion before merging.

.github/workflows/bandit.yml
```diff
@@ -0,0 +1,19 @@
...
+  cancel-in-progress: true
+
+jobs:
+  analyze:
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+    steps:
+      - name: Perform Bandit Analysis
+        uses: PyCQA/bandit-action@8a1b30610f61f3f792fe7556e888c9d7dffa52de  # v1.0.0
```
kkraus14, line 19: This action is here: https://github.com/PyCQA/bandit-action/blob/main/action.yml

This doesn't pin the version of Bandit

.pre-commit-config.yaml
```diff
@@ -11,11 +11,15 @@ ci:
...
-    rev: v0.11.4
+    rev: 971923581912ef60a6b70dbf0c3e9a39563c9d47  #v0.11.4
     hooks:
       - id: ruff
         args: [--fix, --show-fixes]
       - id: ruff-format
+  - repo: https://github.com/PyCQA/bandit
+    rev: 8ff25e07e487f143571cc305e56dd0253c60bc7b  #v1.8.3
+    hooks:
+      - id: bandit
```
kkraus14, line 22: This does pin the version of Bandit to v1.8.3 which could cause mismatches between this and the GitHub workflow.

kkraus14:
An issue here:
- The GitHub action workflow for bandit runs `pip install bandit[sarif]` and doesn't give us a way to pin the version of bandit. (https://github.com/PyCQA/bandit-action/blob/8a1b30610f61f3f792fe7556e888c9d7dffa52de/action.yml#L80)
- The pre-commit hook does allow us to pin the version of bandit, which I've done to the current release.
- Running bandit via pre-commit.ci doesn't allow us to grab and upload the SARIF file into the GitHub code security tooling.

I've also temporarily moved the CodeQL action to be manually triggered only until our internal discussion is completed.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,36 @@
+name: "Static Analysis: CodeQL Scan"
+
+on: workflow_dispatch
```
leofang: > I've also temporarily moved the CodeQL action to be manually triggered only until our internal discussion is completed.

As discussed internally, we can re-enable `on: push:` now!

.pre-commit-config.yaml
```diff
@@ -6,16 +6,20 @@ ci:
     autoupdate_branch: ''
     autoupdate_commit_msg: '[pre-commit.ci] pre-commit autoupdate'
     autoupdate_schedule: quarterly
-    skip: []
+    skip: [bandit]
```
leofang, line 9: Q: I assume this is needed because without it the pre-commit check would currently fail?

leofang (approved):

.pre-commit-config.yaml
kkraus14: No, it's because we are running the check via the Bandit GitHub Action separately which then can upload the output SARIF file into GitHub code scanning (https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/uploading-a-sarif-file-to-github). So I figured there's no need to run it twice unnecessarily.

That being said we still have the issue I commented above about the action not allowing us to control the version of bandit at present as it just runs a `pip install bandit[sarif]` command.

kkraus14:
Merging for now and will create issues for following up on Bandit version pinning for the Action and CodeQL pre-commit hook.

leofang:
Thanks, Keith!

.pre-commit-config.yaml
leofang: Asking in upstream here: https://github.com/PyCQA/bandit-action/issues/13

PR #592, opened 2025-04-30
Update _launcher.py to fix vector_add.py

markwhen:
## Description

closes 

## Checklist

### DIFF
--- a/cuda_core/cuda/core/experimental/_launcher.py
+++ b/cuda_core/cuda/core/experimental/_launcher.py
@@ -146,6 +146,6 @@ def launch(stream, config, kernel, *kernel_args):
         # TODO: check if config has any unsupported attrs
         handle_return(
             driver.cuLaunchKernel(
-                int(kernel._handle), *config.grid, *config.block, config.shmem_size, stream._handle, args_ptr, 0
+                int(kernel._handle), *config.grid, *config.block, config.shmem_size, stream._ctx_handle, args_ptr, 0
             )
         )


cuda_core/cuda/core/experimental/_launcher.py
```diff
@@ -146,6 +146,6 @@ def launch(stream, config, kernel, *kernel_args):
         # TODO: check if config has any unsupported attrs
         handle_return(
             driver.cuLaunchKernel(
-                int(kernel._handle), *config.grid, *config.block, config.shmem_size, stream._handle, args_ptr, 0
+                int(kernel._handle), *config.grid, *config.block, config.shmem_size, stream._ctx_handle, args_ptr, 0
```
leofang, line 149: We need a stream here, not context:
```suggestion
                int(kernel._handle), *config.grid, *config.block, config.shmem_size, stream.handle, args_ptr, 0
```

leofang (changes requested):
Thanks for catching this bug @markwhen. The fix is not quite right, though, see below. Also, I am curious if you are using driver <11.8 by any chance? This can be confirmed via `nvidia-smi`.

@rwgk this seems like an overlook from our earlier refactoring. I suspect our CI did not catch it because we don't have runners with old 11.x driver that could trigger this path.

leofang:
Actually I am trying this example on a machine with A100, driver 520.169 (11.8), CTK 11.8, and cuda-bindings 11.8.7. I hit no issue. @markwhen what exactly are we trying to fix here?

leofang:
Closing due to lack of response. Please feel free to open an issue if you still encounter problems. Thanks.

PR #600, opened 2025-05-02
Ensure `fetch_ctk` action can be used everywhere

### LINKED ISSUE #588: CI: Windows local CTK installation at test time shows a warning
xref: https://github.com/Jimver/cuda-toolkit/issues/390

### LINKED ISSUE #599: CI: Enable our homegrown `fetch_ctk` action on Windows?
We took the shortcut and used a 3rd-party action on the Windows test runners due to some technical reasons (the build runners already use `fetch_ctk` just fine)
https://github.com/NVIDIA/cuda-python/blob/bd770e14e0fb960b64862743eafc88394fb6f14a/.github/workflows/test-wheel-windows.yml#L161-L165
It works very well most of the time, except when updating to the latest CTK release as evidenced in #597:
https://github.com/NVIDIA/cuda-python/actions/runs/14795896391/job/41543758103?pr=597#step:15:27
Because of this and other reasons (ex: #588), we probably should continue finding a way to enable `fetch_ctk` universally.

leofang:
## Description

closes #599 
closes #588

## Checklist

### DIFF
--- a/.github/actions/fetch_ctk/action.yml
+++ b/.github/actions/fetch_ctk/action.yml
@@ -19,11 +19,14 @@ inputs:
     type: string
     default: "cuda_nvcc,cuda_cudart,cuda_nvrtc,cuda_profiler_api,cuda_cccl,cuda_sanitizer_api,libnvjitlink"
 
+defaults:
+  run:
+    shell: bash ${{ (inputs.host-platform != 'win-64' && '--noprofile --norc') || '' }} -xeuo pipefail {0}
+
 runs:
   using: composite
   steps:
     - name: Set up CTK cache variable
-      shell: bash --noprofile --norc -xeuo pipefail {0}
       run: |
         HASH=$(echo -n "${{ inputs.cuda-components }}" | sha256sum | awk '{print $1}')
         echo "CTK_CACHE_KEY=mini-ctk-${{ inputs.cuda-version }}-${{ inputs.host-platform }}-$HASH" >> $GITHUB_ENV
@@ -48,7 +51,6 @@ runs:
 
     - name: Get CUDA components
       if: ${{ steps.ctk-get-cache.outputs.cache-hit != 'true' }}
-      shell: bash --noprofile --norc -xeuo pipefail {0}
       run: |
         CUDA_PATH="./cuda_toolkit"
         mkdir $CUDA_PATH
@@ -126,7 +128,6 @@ runs:
 
     - name: Restore CTK cache
       if: ${{ steps.ctk-get-cache.outputs.cache-hit == 'true' }}
-      shell: bash --noprofile --norc -xeuo pipefail {0}
       run: |
         ls -l
         CUDA_PATH="./cuda_toolkit"
@@ -137,7 +138,6 @@ runs:
         fi
 
     - name: Set output environment variables
-      shell: bash --noprofile --norc -xeuo pipefail {0}
       run: |
         CUDA_PATH=$(realpath "./cuda_toolkit")
         echo "CUDA_PATH=${CUDA_PATH}" >> $GITHUB_ENV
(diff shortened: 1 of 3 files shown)


.github/actions/fetch_ctk/action.yml
```diff
@@ -23,7 +23,7 @@ runs:
   using: composite
   steps:
     - name: Set up CTK cache variable
-      shell: bash --noprofile --norc -xeuo pipefail {0}
+      shell: bash -xeuo pipefail {0}
```
leofang: For some reason the Git Bash that I installed really hates `--noprofile --norc` -- It becomes a no-op entirely (nothing happens when a job step launches it). However, this is not the case on ~~GH-hosted~~ **self-** hosted Windows (build) runners. I have no clue why it is the case, but if the CI is all green I don't plan to dig further.

leofang: It turns out that this weirdness is a symptom that I used the wrong bash executable. It should be `bin\bash.exe` (which runs in the same terminal) instead of the root `git-bash.exe` (which opens a new window when tested locally). See also https://www.timschaeps.be/post/adding-git-bash-to-windows-terminal/. Commit a73564c shows that by fixing this we not only recover all `fetch_ctk` verbose output (critical for debugging!) but also can revert this hack. Now the custom Git Bash installation behaves like the one installed in the GH-hosted Windows images:
https://github.com/actions/runner-images/blob/5676b8b82520e84040b0e57569be0c0db7d92c4c/images/windows/scripts/build/Configure-Shell.ps1#L16

leofang: > critical for debugging!

oh and this also fixes the silent error suppression -- if `fetch_ctk` fails, now it propagates correctly (PowerShell can check this via `$?`
https://github.com/NVIDIA/cuda-python/actions/runs/14808554816/job/41580483605

leofang:
OK this PR is ready/safe to merge. I confirm `fetch_ctk` can handle all cache hit/miss paths properly on all supported GHA runners:
- Run with existing cached mini CTKs: https://github.com/NVIDIA/cuda-python/actions/runs/14816796095?pr=600
- Run with mini CTKs purged (so rebuilding the cache): https://github.com/NVIDIA/cuda-python/actions/runs/14817034449?pr=600

.github/actions/fetch_ctk/action.yml
```diff
@@ -146,8 +148,17 @@ runs:
...
+          CUDA_PATH=$(realpath "./cuda_toolkit")
+          echo "LD_LIBRARY_PATH=${LD_LIBRARY_PATH:-}:${CUDA_PATH}/lib:${CUDA_PATH}/nvvm/lib64" >> $GITHUB_ENV
+        elif [[ "${{ inputs.host-platform }}" == win* ]]; then
+          function normpath() {
+            echo "$(echo $(cygpath -w $1) | sed 's/\\/\\\\/g')"
+          }
+          CUDA_PATH=$(normpath $(realpath "./cuda_toolkit"))
+          echo "$(normpath ${CUDA_PATH}/bin)" >> $GITHUB_PATH
+          echo "$(normpath $CUDA_PATH/nvvm/bin)" >> $GITHUB_PATH
+        fi
```
rwgk, line 162: Might be useful to guard against accidents (especially accidents elsewhere):
```
else
    echo "::error::Unsupported host-platform: ${{ inputs.host-platform }}" >&2
    exit 1
```

rwgk (approved):

.github/actions/fetch_ctk/action.yml
leofang: This is a good suggestion we can revisit later, I have a few things in mind regarding the `fetch_ctk` action. 

Currently we're protected by `runs-on:` in the caller workflows: 
https://github.com/NVIDIA/cuda-python/blob/9579678a6c2f561790127f225975f4818eaad1ee/.github/workflows/build-and-test.yml#L43-L45
If `host-platform` is not in the 3 choices that we currently support, the CI workflow wouldn't even run.

PR #578, opened 2025-04-25
2025-05-01 version of `cuda.bindings.path_finder`

rwgk:
## Description
Major milestone for the work tracked under #451

This PR introduces only two public APIs:

* `cuda.bindings.path_finder.SUPPORTED_LIBNAMES` (currently `('nvJitLink', 'nvrtc', 'nvvm')`)
* `cuda.bindings.path_finder.load_nvidia_dynamic_library(libname: str) -> LoadedDL`

With:
```Python
@dataclass                                                                      
class LoadedDL:                                                                 
    handle: int                                                                 
    abs_path: Optional[str]                                                     
    was_already_loaded_from_elsewhere: bool                                     
```

However, the implementations were actually thoroughly tested (under #558) for all

```
SUPPORTED_LIBNAMES + PARTIALLY_SUPPORTED_LIBNAMES
```

enumerated under `cuda.bindings._path_finder.supported_libs` (note that this module is private).

To make this PR easier to review, the [changes to the `nvJitLink`, `nvrtc`, and `nvvm` bindings are NOT included](https://github.com/NVIDIA/cuda-python/pull/578/commits/17478da1e9677c44f117c817feeac9c236cfab4f) in this PR. Those changes were also already tested under #558. They will be merged into `main` with two follow-on PRs (one for the `nvrtc` bindings, one for `nvJitLink` and `nvvm`).

Thorough testing of all `SUPPORTED_LIBNAMES + PARTIALLY_SUPPORTED_LIBNAMES` requires changes to the GitHub Actions configs, to set up suitable CTK installations. This will also be handled separately in follow-on PRs.
________

Suggested order for reviewing files:

* `cuda/bindings/_path_finder/supported_libs.py`
* `cuda/bindings/_path_finder/load_nvidia_dynamic_library.py`
* `cuda/bindings/_path_finder/load_dl_common.py`
* `cuda/bindings/_path_finder/load_dl_linux.py`
* `cuda/bindings/_path_finder/load_dl_windows.py`
* `tests/test_path_finder.py`
* `cuda/bindings/_path_finder/find_nvidia_dynamic_library.py`
* everything else

Discussion points:

* Copyright notice for `cuda/bindings/_path_finder/cuda_paths.py` (the original file under numba-cuda does not have one)
* Documentation for the new public APIs
* Documentation for maintaining `SUPPORTED_LIBNAMES + PARTIALLY_SUPPORTED_LIBNAMES` as new CTK versions are released

PR #604, opened 2025-05-04
WIP: path_finder Search Priority v2

rwgk:
Continuation of #578

Completely replace `cuda_paths.py` to achieve the desired Search Priority (see updated README.md).

Further testing is under #610


cuda_bindings/cuda/bindings/_path_finder/README.md
```diff
@@ -24,52 +24,27 @@ strategy for locating NVIDIA shared libraries:
...
-     - Distribution-specific packages (RPM/DEB)
-       EXCEPT Debian's `nvidia-cuda-toolkit`
-
-5. **OS Default Mechanisms**
+3. **OS default mechanisms / Conda environments**
    - Falls back to native loader:
      - `dlopen()` on Linux
      - `LoadLibraryW()` on Windows
+   - CTK installations with system config updates are expected to be discovered:
+     - Linux: Via `/etc/ld.so.conf.d/*cuda*.conf`
```
kkraus14, line 35: Does this get set for both package manager and runfile installations?

cuda_bindings/cuda/bindings/_path_finder/find_nvidia_dynamic_library.py
```diff
@@ -61,88 +59,114 @@ def _find_dll_using_nvidia_bin_dirs(libname, lib_searched_for, error_messages, a
...
-    for so_name in candidate_so_names:
-        if os.path.isfile(so_name):
-            return so_name
-        error_messages.append(f"No such file: {so_name}")
-    for so_dirname in candidate_so_dirs:
-        attachments.append(f'  listdir("{so_dirname}"):')
-        if not os.path.isdir(so_dirname):
-            attachments.append("    DIRECTORY DOES NOT EXIST")
+    if IS_WINDOWS:
+        if libname == "nvvm":  # noqa: SIM108
```
kkraus14: Should we just appease the linter instead of noqa'ing these?

cuda_bindings/cuda/bindings/_path_finder/load_nvidia_dynamic_library.py
```diff
@@ -38,6 +59,15 @@ def _load_nvidia_dynamic_library_no_cache(libname: str) -> LoadedDL:
         loaded = load_with_system_search(libname, found.lib_searched_for)
         if loaded is not None:
             return loaded
+        if libname == "nvvm":
+            # Use cudart as anchor point (libcudart.so.12 is only ~720K, cudart64_12.dll ~560K).
```
kkraus14: `cudart` is probably not the best choice here as it's by default statically linked by `nvcc` and is most commonly statically linked.

cuda_bindings/cuda/bindings/_path_finder/load_nvidia_dynamic_library.py
```diff
@@ -38,6 +59,15 @@ def _load_nvidia_dynamic_library_no_cache(libname: str) -> LoadedDL:
...
+            loaded_cudart = check_if_already_loaded_from_elsewhere("cudart")
+            if loaded_cudart is not None:
+                found.retry_with_other_abs_path(loaded_cudart.abs_path)
+            else:
+                cudart_abs_path = _load_other_in_subprocess("cudart", found.error_messages)
```
kkraus14: I forget the exact rules but CUDA Contexts are not fork safe where if someone has created a CUDA Context before potentially hitting this code under the covers it is problematic.

If we absolutely need to use another process because there's no way for us to resolve the path without possibly loading unwanted symbols, then we need to make sure that we spawn a new process instead of forking a child process.

cuda_bindings/cuda/bindings/_path_finder/README.md
leofang: Runfiles actually touch `/etc/ld.so.conf.d` too (KITMAKER-351), so yes.

cuda_bindings/cuda/bindings/_path_finder/load_dl_common.py
```diff
@@ -38,3 +40,21 @@ def load_dependencies(libname: str, load_func: Callable[[str], LoadedDL]) -> Non
...
+
+
+def load_in_subprocess(python_code, timeout=30):
+    # This is to avoid loading libraries into the parent process.
+    return subprocess.run(  # nosec B603
+        [sys.executable, "-c", python_code],
+        capture_output=True,
+        encoding="utf-8",
+        timeout=timeout,  # Ensure this does not hang for an excessive amount of time.
+    )
```
leofang: I am juggling between chats so just leaving a drive-by comment for now. We need to pass `env=os.environ` too to ensure the new process is seeing the same environment (`PATH`, `LD_LIBRARY_PATH`, `CONDA_PREFIX`, etc).

rwgk: Done: commit 1f728c0cb4a5a99788e230edb739cc29a12ded7a

Thanks for catching this!

cuda_bindings/cuda/bindings/_path_finder/load_nvidia_dynamic_library.py
rwgk: Ah, I **mistakenly** believed that `subprocess` does not use fork, but I just confirmed that it does.

I need to change this.

cuda_bindings/cuda/bindings/_path_finder/find_nvidia_dynamic_library.py
leofang: It seems the if-else blocks are used so that we can have a multi-line list with inline comments

cuda_python/docs/source/release/12.9.0-notes.rst
leofang, line 1: These diffs seem like rebasing errors?

cuda_bindings/cuda/bindings/_path_finder/run_python_code_safely.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def __call__(self):
+        # Capture stdout/stderr
+        old_stdout = sys.stdout
+        old_stderr = sys.stderr
+        sys.stdout = StringIO()
+        sys.stderr = StringIO()
+
+        try:
+            exec(self.python_code, {"__name__": "__main__"})  # nosec B102
```
leofang: I feel I am being the pedantic old man on the team keeping flagging the usage of `exec`, and sorry for feeling obligated to raise this, but why do we need a worker class like this, instead of keeping it simple and just passing the function object to `multiprocessing.Process()` (and let it pickle it under the hood)? Then we don't need to execute arbitrary Python strings using `exec`.

rwgk:
Tracking the development history of up to commit 9b474bc80b079f5977712970d63439310a9f3319:

1. [Perplexity_run_python_code_safely_original_version.pdf](https://github.com/user-attachments/files/20048745/1_Perplexity_run_python_code_safely_original_version.pdf)
2. [ChatGPT Deep research followed by feeding back Gemini 2.5 findings](https://chatgpt.com/share/681914ce-f274-8008-9e9f-4538716b4ed7)
3. [Claude_3.7_Sonnet_2025-05-05+131336.pdf](https://github.com/user-attachments/files/20048764/3_Claude_3.7_Sonnet_2025-05-05%2B131336.pdf) — This didn't lead to any changes.
4. [Gemini_2.5_2025-05-05+132203.pdf](https://github.com/user-attachments/files/20048767/4_Gemini_2.5_2025-05-05%2B132203.pdf) — Discovered a "classical race condition"

cuda_bindings/cuda/bindings/_path_finder/find_nvidia_dynamic_library.py
rwgk: Done: commit ab00a872181d0a84cd20bb0459e2fe20b3dbcc37

(It came out looking better than I was anticipating.)

cuda_bindings/cuda/bindings/_path_finder/load_nvidia_dynamic_library.py
rwgk: New solution: run_python_code_safely.py, see https://github.com/NVIDIA/cuda-python/pull/604#issuecomment-2852339361

cuda_bindings/cuda/bindings/_path_finder/run_python_code_safely.py
rwgk: We could require a callable (one of the AIs actually suggested that), but then this wouldn't be a drop-in replacement for `subprocess.run` anymore. I think it could take a few hours to fully and cleanly pivot in that direction. Could we leave this for later? The context is so narrowly defined, and so special, I don't think this particular `exec` is an actual problem.

cuda_python/docs/source/release/12.9.0-notes.rst
rwgk: Yes, sorry I forgot to `git push` the `path_finder_dev` branch before. The diff is gone now.

leofang: But I still see all the doc changes from #611 in the diff?

cuda_bindings/cuda/bindings/_path_finder/load_nvidia_dynamic_library.py
```diff
@@ -21,6 +27,21 @@
     )
 
 
+def _load_other_in_subprocess(libname, error_messages):
+    code = f"""\
+from cuda.bindings._path_finder.load_nvidia_dynamic_library import load_nvidia_dynamic_library
+import json
+import sys
+loaded = load_nvidia_dynamic_library({libname!r})
+sys.stdout.write(json.dumps(loaded.abs_path, ensure_ascii=True))
```
leofang: Q: Are there concerns dumping the path directly to stdout without json? I suppose we can handle unicode paths just fine?

cuda_bindings/cuda/bindings/_path_finder/load_nvidia_dynamic_library.py
```diff
@@ -21,6 +27,21 @@
     )
 
 
+def _load_other_in_subprocess(libname, error_messages):
+    code = f"""\
```
leofang: I think we can turn this code into a Python function and pass it to multiprocessing.

cuda_bindings/cuda/bindings/_path_finder/run_python_code_safely.py
leofang: I think we still have time to handle this properly today, and have the PR merged and release cut first thing tomorrow morning (today is too late and we still have CI issues). I am not sure why we want to keep it as a drop-in replacement for `subprocess.run` when we know it isn't the way to go?

cuda_bindings/cuda/bindings/_path_finder/README.md
```diff
@@ -24,52 +24,27 @@ strategy for locating NVIDIA shared libraries:
      The absolute path of the already loaded library will be returned, along
      with the handle to the library.
 
-1. **Python Package Ecosystem**
-   - Scans `sys.path` to find libraries installed via NVIDIA Python wheels.
+1. **Environment variables**
+   - Relies on `CUDA_HOME` or `CUDA_PATH` environment variables if set
+     (in that order).
```
leofang: Capturing the internal discussion here: During the meeting last week, Satya’s comment reminded me that if we honor `CUDA_HOME` first, we actually block our own internal users, with a typical workflow like this:
1. Set `CUDA_HOME` pointing to the latest public CTK release from the container or an internal pre-release build
2. Set `LD_LIBRARY_PATH` to the local builds of math libs (ex: cuBLAS, cuTENSOR, etc) that are in development, so that these target components have a higher loading precedence.

I think `CUDA_HOME`/`CUDA_PATH` should be the last resort, not the first.

cuda_bindings/cuda/bindings/_path_finder/load_nvidia_dynamic_library.py
rwgk: Changed to `nppc` as `ANCHOR_LIBNAME`: commit 2a039d2a7ec44e1d77b68b7fcc610897fcb4bd4f

That library is smaller than `cudart` on Windows, but more than double the size of `cudart` on Linux. But `nppc` will only be imported once per process, and only if `nvvm` is requested.

cuda_bindings/cuda/bindings/_path_finder/README.md
rwgk: My 2cents:

A "regular" CUDA user outside NVIDIA will likely be surprised if `CUDA_HOME` is ignored. For example, assume there is some old installation under `/usr/local/cuda`, but the user doesn't have root, installs a modern CTK under `$HOME/cuda_modern` and sets `CUDA_HOME` to that directory. They will be very confused to find that `/usr/local/cuda` is used regardless.

I think we need to cater to both situations.

I'm assuming regular CUDA users outnumber NVIDIAns by 100:1 (or more).

The current implementation caters to regular CUDA users.

We could add another environment variable to cater to NVIDIAns, e.g. `CUDA_PYTHON_CUDA_HOME_PRIORITY`, which, if defined, must have one of two values: `first`, `last`

A helpful error is raised if we find any other value.

We could support additional values in the future, as the need arises.

leofang: I find this example a bit confusing:
> A "regular" CUDA user outside NVIDIA will likely be surprised if `CUDA_HOME` is ignored. For example, assume there is some old installation under `/usr/local/cuda`, but the user doesn't have root, installs a modern CTK under `$HOME/cuda_modern` and sets `CUDA_HOME` to that directory. They will be very confused to find that `/usr/local/cuda` is used regardless.

There is a reason modern users prefer pip/conda over system installation, because in such cases (no root access) users should install (relocatable) packages to the user space and make them discoverable:
- If a user installs a full CTK to the user space, the standard way is to set up `PATH` and `LD_LIBRARY_PATH`. The runfile installer hints about this, for example. 
  - This is my biggest worry. This is the standard practice to make executables/shared libraries discoverable, but we are ignoring them if `CUDA_HOME` is (mistakenly) set.
- If a user has CTK components installed via pip or conda, while mistakenly setting `CUDA_HOME` pointing to somewhere else,  the pip/conda installation is ignored, which cab be equally confusing if not more.

I think our design should encourage (general, the 100) users to prefer pip or conda packages over a bulky installation, while keeping a door open for internal (the 1) development. Having `CUDA_HOME` searched the last meets this requirement without the need of introducing extra env vars, but it's not the case if it is the other way around.

cuda_python/docs/source/release/12.9.0-notes.rst
leofang: It's now gone lol... probably there's some cache lifetime glitches with my browser

leofang:
It is pretty awesome that the search priority update also allows us to drop the old Numba code entirely!

cuda_bindings/cuda/bindings/_path_finder/README.md
rwgk: I was (almost) done with commit f978e67ece5503b814f0b4c5984fc87310343a6f before seeing your message, so I committed what I had, with minimal interactive testing.

Next I removed the `ANCHOR_LIBNAME` feature and made it so that `CUDA_HOME` is unconditionally referenced **last**  in the follow-on commit 782fcf6f0ba07ad48f1649907f8e28de7c18c7d6.

That commit demonstrates that it is pretty easy now to shuffle the search priority around.

Due to the removed `ANCHOR_LIBNAME` feature, we can no longer remove the nvvm-related `LD_LIBRARY_PATH`, `PATH` manipulations under our `.github` directory, therefore I added them back with commit ad9e994df879c5a4e68f6ac39554bcb1c6663b01 **under PR #610**, where I'm testing with the `path_finder` integration into the nvJitLink, nvrtc, and nvvm bindings.

I tested commit 782fcf6f0ba07ad48f1649907f8e28de7c18c7d6 only [under PR #610, with this run](https://github.com/NVIDIA/cuda-python/actions/runs/14851110820).

That testing is a superset of testing here.

One test is failing under 610: https://github.com/NVIDIA/cuda-python/actions/runs/14851110820/job/41695273335

Maybe because I still need to update supported_libs.py for the 12.9.0 release. Working on that.

leofang: Looks like you fixed it in commit a5763272? 🎉

cuda_bindings/cuda/bindings/_path_finder/load_nvidia_dynamic_library.py
leofang: This is moot now that we no longer use `_load_anchor_in_subprocess`

cuda_bindings/cuda/bindings/_path_finder/load_nvidia_dynamic_library.py
leofang: This is moot now that we no longer use `_load_anchor_in_subprocess`

cuda_bindings/cuda/bindings/_path_finder/load_nvidia_dynamic_library.py
```diff
@@ -21,6 +27,21 @@
...
+loaded = load_nvidia_dynamic_library({libname!r})
+sys.stdout.write(json.dumps(loaded.abs_path, ensure_ascii=True))
+"""
+    result = load_in_subprocess(code)
+    if result.returncode == 0:
+        return json.loads(result.stdout)
+    error_messages.extend(build_subprocess_failed_for_libname_message(libname, result).splitlines())
+    return None
+
+
```
leofang: Removing dead code
```suggestion
```

cuda_bindings/cuda/bindings/_path_finder/run_python_code_safely.py
leofang, line 1: I think this whole file can be removed now!

cuda_bindings/cuda/bindings/_path_finder/run_python_code_safely.py
leofang: This is moot now that we no longer use `_load_anchor_in_subprocess`

cuda_bindings/cuda/bindings/_path_finder/load_nvidia_dynamic_library.py
```diff
@@ -2,12 +2,18 @@
...
+import json
 
 from cuda.bindings._path_finder.find_nvidia_dynamic_library import _find_nvidia_dynamic_library
-from cuda.bindings._path_finder.load_dl_common import LoadedDL, load_dependencies
+from cuda.bindings._path_finder.load_dl_common import (
+    LoadedDL,
+    build_subprocess_failed_for_libname_message,
+    load_dependencies,
+    load_in_subprocess,
+)
```
leofang: restore to remove dead code
```suggestion
from cuda.bindings._path_finder.load_dl_common import LoadedDL, load_dependencies
```

cuda_bindings/cuda/bindings/_path_finder/load_dl_common.py
```diff
@@ -38,3 +35,16 @@ def load_dependencies(libname: str, load_func: Callable[[str], LoadedDL]) -> Non
...
+    # This is to avoid loading libraries into the parent process.
+    return run_python_code_safely(python_code, timeout=timeout)
+
+
+def build_subprocess_failed_for_libname_message(libname, result):
+    return (
+        f"Subprocess failed for {libname=!r} with exit code {result.returncode}\n"
+        f"--- stdout-from-subprocess ---\n{result.stdout}<end-of-stdout-from-subprocess>\n"
+        f"--- stderr-from-subprocess ---\n{result.stderr}<end-of-stderr-from-subprocess>\n"
+    )
```
leofang: These are not needed now
```suggestion
```

leofang (commented):
Looking good! Several previous discussions on `exec` and subprocess/multiprocessing are now moot by removing these unused code entirely.

cuda_bindings/cuda/bindings/_path_finder/load_nvidia_dynamic_library.py
leofang: This is now moot since the anchor is no longer in use (782fcf6).

cuda_bindings/cuda/bindings/_path_finder/run_python_code_safely.py
rwgk: run_python_code_safely.py provides best-possible isolation because it does not use fork. It was moved to tests/ (commit 5fb2d1fec4932b7b86b4111d87ba73ce8a3c1f71).

cuda_bindings/cuda/bindings/_path_finder/load_nvidia_dynamic_library.py
rwgk: This code was removed with commit 5fb2d1fec4932b7b86b4111d87ba73ce8a3c1f71.

I wanted to play it safe.

cuda_bindings/cuda/bindings/_path_finder/load_nvidia_dynamic_library.py
rwgk: This code was removed with commit 5fb2d1fec4932b7b86b4111d87ba73ce8a3c1f71.

The run_python_code_safe still exists, but only in the tests/ directory. We can replace `exec` with passing a callable separately.

cuda_bindings/cuda/bindings/_path_finder/load_nvidia_dynamic_library.py
rwgk: Done with commit 5fb2d1fec4932b7b86b4111d87ba73ce8a3c1f71.

cuda_bindings/cuda/bindings/_path_finder/run_python_code_safely.py
rwgk: It's still needed for test_path_finder_load.py, and was moved to tests/ with commit 5fb2d1fec4932b7b86b4111d87ba73ce8a3c1f71.

It's better than `subprocess.run` because it does not use fork, i.e. has the best-possible isolation.

cuda_bindings/cuda/bindings/_path_finder/load_nvidia_dynamic_library.py
rwgk: Done with commit 5fb2d1fec4932b7b86b4111d87ba73ce8a3c1f71.

cuda_bindings/cuda/bindings/_path_finder/load_dl_common.py
rwgk: Done with commit 5fb2d1fec4932b7b86b4111d87ba73ce8a3c1f71:

`load_in_subprocess()` was removed completely.

`build_subprocess_failed_for_libname_message()` was moved back into test_path_finder_load.py

rwgk:
This PR has base branch `path_finder_dev`. — I'm keeping it around because it's easiest to look at for a final review.

**PR #613 has base branch `main`, but against the same branch.**

Interesting/surprising: I triggered the CI under #613, but it shows up here, too.

leofang (approved):

cuda_bindings/cuda/bindings/_path_finder/README.md
```diff
@@ -24,52 +24,27 @@ strategy for locating NVIDIA shared libraries:
...
-
-5. **OS Default Mechanisms**
+2. **OS default mechanisms / Conda environments**
    - Falls back to native loader:
      - `dlopen()` on Linux
      - `LoadLibraryW()` on Windows
+   - CTK installations with system config updates are expected to be discovered:
+     - Linux: Via `/etc/ld.so.conf.d/*cuda*.conf`
+     - Windows: Via `C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\vX.Y\bin` on system `PATH`
+   - Conda installations are expected to be discovered:
```
kkraus14, line 37: nitpick: would be good to move this higher than CTK installations as it would be found via `RPATH` / `PATH` earlier than `LD_LIBRARY_PATH` / `ld.so.conf.d`

cuda_bindings/tests/run_python_code_safely.py
kkraus14, line 1: Need SPDX license + Copyright

cuda_bindings/tests/run_python_code_safely.py
```diff
@@ -0,0 +1,86 @@
...
+            sys.stdout = old_stdout
+            sys.stderr = old_stderr
+            try:  # noqa: SIM105
+                self.result_queue.put((returncode, stdout, stderr))
+            except Exception:  # nosec B110
+                # If the queue is broken (e.g., parent gone), best effort logging
+                pass
+
+
+def run_python_code_safely(python_code, *, timeout=None):
```
kkraus14, line 42: Can we rename this function? Something like `run_py_spawned_subprocess` or something like that?

cuda_bindings/cuda/bindings/_path_finder/README.md
leofang: The `CUDA_*` env vars are now checked the last.

cuda_bindings/tests/run_python_code_safely.py
leofang: This is done in commit c6f2205 from the other PR.

cuda_bindings/cuda/bindings/_path_finder/README.md
rwgk: Done under #621

cuda_bindings/tests/run_python_code_safely.py
rwgk: Done under #621

Issue #454, opened 2025-02-18
Stop repeating input/output type info in docstrings


leofang:
I think Sphinx is able to figure out the types and keep them in the document by inspecting the typing info, which going forward should be our source of truth (https://github.com/NVIDIA/cuda-python/issues/312#issue-2746694448). We should not repeat ourselves.

leofang:
Found another relevant issue: https://github.com/sphinx-doc/sphinx/issues/7077, and someone pointed to the `sphinx-autodoc-typehint` plugin: https://github.com/sphinx-doc/sphinx/issues/7077#issuecomment-1571885203

PR #612, opened 2025-05-06
Implement Kernel.num_arguments, and Kernel.arguments_info

### LINKED ISSUE #568: Expose `cuda.core.Kernel.attributes.num_arguments`
This is a handy approach as suggested by @galv since the driver does not expose any queryable APIs for this purpose:
> Hello pipul, there is not a direct API, but you can do this https://github.com/pytorch/pytorch/pull/137318/files#diff-d7302d133bb5e0890fc94de9aeea4d9d442555a3b40772c9db10edb5cf36a35cR424-R429
>
> In other words:
> 
> ```
> for(size_t i = 0; cudaFuncGetParamInfo(func, i, &param_offset, &param_size) != CUDA_ERROR_INVALID_VALUE; i++) {}
> // i now contains the total number of arguments to your kernel.
> ```
>
> I'm sure you can translate that to python.

_Originally posted by @galv in https://github.com/NVIDIA/cuda-python/discussions/543#discussioncomment-12740542_

oleksandr-pavlyk:
## Description

closes #568 

This PR implements `Kernel.num_arguments` and `Kernel.argument_info` properties:
  - `Kernel.num_arguments` returns the number of arguments in the kernel instance
  - `Kernel.argument_info` returns a list of tuples `(offset, size)` which describes layout of the struct containing all kernel arguments,

## Checklist

### DIFF
--- a/cuda_core/cuda/core/experimental/_module.py
+++ b/cuda_core/cuda/core/experimental/_module.py
@@ -232,6 +232,8 @@ class Kernel:
                 p_info = Kernel.ParamInfo(offset=result[1], size=result[2])
                 param_info_data.append(p_info)
             arg_pos = arg_pos + 1
+        if result[0] != driver.CUresult.CUDA_ERROR_INVALID_VALUE:
+            handle_return(result)
         return arg_pos, param_info_data
 
     @property
--- a/cuda_core/cuda/core/experimental/_module.py
+++ b/cuda_core/cuda/core/experimental/_module.py
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from collections import namedtuple
-from typing import NamedTuple, Optional, Union
+from typing import Optional, Union
 from warnings import warn
 
 from cuda.core.experimental._utils.clear_error_support import (
@@ -184,6 +184,9 @@ class KernelAttributes:
         )
 
 
+ParamInfo = namedtuple("ParamInfo", ["offset", "size"])
+
+
 class Kernel:
     """Represent a compiled kernel that had been loaded onto the device.
 
@@ -196,7 +199,6 @@ class Kernel:
     """
 
     __slots__ = ("_handle", "_module", "_attributes")
-    ParamInfo = namedtuple("ParamInfo", ["offset", "size"])
 
     def __new__(self, *args, **kwargs):
         raise RuntimeError("Kernel objects cannot be instantiated directly. Please use ObjectCode APIs.")
@@ -218,7 +220,7 @@ class Kernel:
             self._attributes = KernelAttributes._init(self._handle)
         return self._attributes
 
-    def _get_arguments_info(self, param_info=False) -> tuple[int, list[NamedTuple]]:
+    def _get_arguments_info(self, param_info=False) -> tuple[int, list[ParamInfo]]:
         attr_impl = self.attributes
         if attr_impl._backend_version != "new":
             raise NotImplementedError("New backend is required")
@@ -229,7 +231,7 @@ class Kernel:
             if result[0] != driver.CUresult.CUDA_SUCCESS:
                 break
             if param_info:
-                p_info = Kernel.ParamInfo(offset=result[1], size=result[2])
+                p_info = ParamInfo(offset=result[1], size=result[2])
                 param_info_data.append(p_info)
             arg_pos = arg_pos + 1
         if result[0] != driver.CUresult.CUDA_ERROR_INVALID_VALUE:
@@ -243,8 +245,8 @@ class Kernel:
         return num_args
 
     @property
-    def arguments_info(self) -> list[NamedTuple]:
-        """list[NamedTuple[int, int]]: (offset, size) for each argument of this function"""
+    def arguments_info(self) -> list[ParamInfo]:
+        """list[ParamInfo]: (offset, size) for each argument of this function"""
         _, param_info = self._get_arguments_info(param_info=True)
         return param_info
 
--- a/cuda_core/cuda/core/experimental/_module.py
+++ b/cuda_core/cuda/core/experimental/_module.py
@@ -2,7 +2,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-from typing import Optional, Union
+from collections import namedtuple
+from typing import NamedTuple, Optional, Union
 from warnings import warn
 
 from cuda.core.experimental._utils.clear_error_support import (
@@ -216,18 +217,20 @@ class Kernel:
             self._attributes = KernelAttributes._init(self._handle)
         return self._attributes
 
-    def _get_arguments_info(self, param_info=False) -> tuple[int, list[tuple[int, int]]]:
+    def _get_arguments_info(self, param_info=False) -> tuple[int, list[NamedTuple]]:
         attr_impl = self.attributes
         if attr_impl._backend_version != "new":
             raise NotImplementedError("New backend is required")
         arg_pos = 0
+        if param_info:
+            ParamInfo = namedtuple("ParamInfo", ["offset", "size"])
         param_info_data = []
         while True:
             result = attr_impl._loader["paraminfo"](self._handle, arg_pos)
             if result[0] != driver.CUresult.CUDA_SUCCESS:
                 break
             if param_info:
-                param_info_data.append(result[1:3])
+                param_info_data.append(ParamInfo._make(result[1:3]))
             arg_pos = arg_pos + 1
         return arg_pos, param_info_data
 
@@ -238,8 +241,8 @@ class Kernel:
         return num_args
 
     @property
-    def arguments_info(self) -> list[tuple[int, int]]:
-        """list[tuple[int, int]]: (offset, size) for each argument of this function"""
+    def arguments_info(self) -> list[NamedTuple]:
+        """list[NamedTuple[int, int]]: (offset, size) for each argument of this function"""
         _, param_info = self._get_arguments_info(param_info=True)
         return param_info
 
--- a/cuda_core/cuda/core/experimental/_module.py
+++ b/cuda_core/cuda/core/experimental/_module.py
@@ -43,6 +43,7 @@ def _lazy_init():
             "data": driver.cuLibraryLoadData,
             "kernel": driver.cuLibraryGetKernel,
             "attribute": driver.cuKernelGetAttribute,
+            "paraminfo": driver.cuKernelGetParamInfo,
         }
         _kernel_ctypes = (driver.CUfunction, driver.CUkernel)
     else:
@@ -214,6 +215,36 @@ class Kernel:
         if self._attributes is None:
             self._attributes = KernelAttributes._init(self._handle)
         return self._attributes
+
+    @property
+    def num_arguments(self) -> int:
+        """int : The number of arguments of this function"""
+        attr_impl = self.attributes
+        if attr_impl._backend_version != "new":
+            raise NotImplementedError("New backend is required")
+        arg_pos = 0
+        while True:
+            result = attr_impl._loader["paraminfo"](self._handle, arg_pos)
+            if result[0] != driver.CUresult.CUDA_SUCCESS:
+                break
+            arg_pos = arg_pos + 1
+        return arg_pos
+
+    @property
+    def arguments_info(self) -> list[tuple[int, int]]:
+        """tuple[tuple[int, int]]: (offset, size) for each argument of this function"""
+        attr_impl = self.attributes
+        if attr_impl._backend_version != "new":
+            raise NotImplementedError("New backend is required")
+        arg_pos = 0
+        param_info = []
+        while True:
+            result = attr_impl._loader["paraminfo"](self._handle, arg_pos)
+            if result[0] != driver.CUresult.CUDA_SUCCESS:
+                break
+            param_info.append((result[1], result[2]))
+            arg_pos = arg_pos + 1
+        return param_info
 
     # TODO: implement from_handle()
 
--- a/cuda_core/cuda/core/experimental/_module.py
+++ b/cuda_core/cuda/core/experimental/_module.py
@@ -196,6 +196,7 @@ class Kernel:
     """
 
     __slots__ = ("_handle", "_module", "_attributes")
+    ParamInfo = namedtuple("ParamInfo", ["offset", "size"])
 
     def __new__(self, *args, **kwargs):
         raise RuntimeError("Kernel objects cannot be instantiated directly. Please use ObjectCode APIs.")
@@ -222,15 +223,14 @@ class Kernel:
         if attr_impl._backend_version != "new":
             raise NotImplementedError("New backend is required")
         arg_pos = 0
-        if param_info:
-            ParamInfo = namedtuple("ParamInfo", ["offset", "size"])
         param_info_data = []
         while True:
             result = attr_impl._loader["paraminfo"](self._handle, arg_pos)
             if result[0] != driver.CUresult.CUDA_SUCCESS:
                 break
             if param_info:
-                param_info_data.append(ParamInfo._make(result[1:3]))
+                p_info = Kernel.ParamInfo(offset=result[1], size=result[2])
+                param_info_data.append(p_info)
             arg_pos = arg_pos + 1
         return arg_pos, param_info_data
 
--- a/cuda_core/cuda/core/experimental/_module.py
+++ b/cuda_core/cuda/core/experimental/_module.py
@@ -216,34 +216,31 @@ class Kernel:
             self._attributes = KernelAttributes._init(self._handle)
         return self._attributes
 
+    def _get_arguments_info(self, param_info=False) -> tuple[int, list[tuple[int, int]]]:
+        attr_impl = self.attributes
+        if attr_impl._backend_version != "new":
+            raise NotImplementedError("New backend is required")
+        arg_pos = 0
+        param_info_data = []
+        while True:
+            result = attr_impl._loader["paraminfo"](self._handle, arg_pos)
+            if result[0] != driver.CUresult.CUDA_SUCCESS:
+                break
+            if param_info:
+                param_info_data.append(result[1:3])
+            arg_pos = arg_pos + 1
+        return arg_pos, param_info_data
+
     @property
     def num_arguments(self) -> int:
         """int : The number of arguments of this function"""
-        attr_impl = self.attributes
-        if attr_impl._backend_version != "new":
-            raise NotImplementedError("New backend is required")
-        arg_pos = 0
-        while True:
-            result = attr_impl._loader["paraminfo"](self._handle, arg_pos)
-            if result[0] != driver.CUresult.CUDA_SUCCESS:
-                break
-            arg_pos = arg_pos + 1
-        return arg_pos
+        num_args, _ = self._get_arguments_info()
+        return num_args
 
     @property
     def arguments_info(self) -> list[tuple[int, int]]:
-        """tuple[tuple[int, int]]: (offset, size) for each argument of this function"""
-        attr_impl = self.attributes
-        if attr_impl._backend_version != "new":
-            raise NotImplementedError("New backend is required")
-        arg_pos = 0
-        param_info = []
-        while True:
-            result = attr_impl._loader["paraminfo"](self._handle, arg_pos)
-            if result[0] != driver.CUresult.CUDA_SUCCESS:
-                break
-            param_info.append((result[1], result[2]))
-            arg_pos = arg_pos + 1
+        """list[tuple[int, int]]: (offset, size) for each argument of this function"""
+        _, param_info = self._get_arguments_info(param_info=True)
         return param_info
 
     # TODO: implement from_handle()
--- a/cuda_core/tests/conftest.py
+++ b/cuda_core/tests/conftest.py
@@ -31,29 +31,45 @@ def init_cuda():
     device = Device()
     device.set_current()
     yield
-    _device_unset_current()
+    _ = _device_unset_current()
 
 
-def _device_unset_current():
+def _device_unset_current() -> bool:
+    """Pop current CUDA context.
+
+    Returns True if context was popped, False it the stack was empty.
+    """
     ctx = handle_return(driver.cuCtxGetCurrent())
     if int(ctx) == 0:
         # no active context, do nothing
-        return
+        return False
     handle_return(driver.cuCtxPopCurrent())
     if hasattr(_device._tls, "devices"):
         del _device._tls.devices
+    return True
 
 
 @pytest.fixture(scope="function")
 def deinit_cuda():
     # TODO: rename this to e.g. deinit_context
     yield
-    _device_unset_current()
+    _ = _device_unset_current()
 
 
 @pytest.fixture(scope="function")
-def deinit_context_function():
-    return _device_unset_current
+def deinit_all_contexts_function():
+    def pop_all_contexts():
+        max_iters = 256
+        for _ in range(max_iters):
+            if _device_unset_current():
+                # context was popped, continue until stack is empty
+                continue
+            # no active context, we are ready
+            break
+        else:
+            raise RuntimeError(f"Number of iterations popping current CUDA contexts, exceded {max_iters}")
+
+    return pop_all_contexts
 
 
 # samples relying on cffi could fail as the modules cannot be imported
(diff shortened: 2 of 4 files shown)


cuda_core/cuda/core/experimental/_module.py
```diff
@@ -215,6 +216,36 @@ def attributes(self) -> KernelAttributes:
             self._attributes = KernelAttributes._init(self._handle)
         return self._attributes
 
+    @property
+    def num_arguments(self) -> int:
+        """int : The number of arguments of this function"""
```
leofang: nit: convert the docstring to NumPy style (follow the `ObjectCode` class below)

cuda_core/cuda/core/experimental/_module.py
```diff
@@ -215,6 +216,36 @@ def attributes(self) -> KernelAttributes:
...
+        while True:
+            result = attr_impl._loader["paraminfo"](self._handle, arg_pos)
+            if result[0] != driver.CUresult.CUDA_SUCCESS:
+                break
+            arg_pos = arg_pos + 1
+        return arg_pos
+
+    @property
+    def arguments_info(self) -> list[tuple[int, int]]:
+        """tuple[tuple[int, int]]: (offset, size) for each argument of this function"""
```
leofang: ditto

leofang (commented):
Thanks, Sasha! Looks great except for the docstrings, see my comment below.

We also need to add a feature entry to `cuda_core/docs/source/release/0.3.0-notes.rst`.

cuda_core/cuda/core/experimental/_module.py
```diff
@@ -215,6 +216,36 @@ def attributes(self) -> KernelAttributes:
...
+        arg_pos = 0
+        while True:
+            result = attr_impl._loader["paraminfo"](self._handle, arg_pos)
+            if result[0] != driver.CUresult.CUDA_SUCCESS:
+                break
+            arg_pos = arg_pos + 1
+        return arg_pos
+
+    @property
+    def arguments_info(self) -> list[tuple[int, int]]:
```
rwgk: The two implementations are very similar.

Did you consider using a private helper function?

```python
    def _arguments_helper(self, param_info=None):
        ...
        if param_info is not None:
            param_info.append((results[1], result[2]))
        ...
```

Then:

```python
    @property
    def num_arguments(self) -> int:
        return self._arguments_helper()

    @property
    def arguments_info(self) -> list[tuple[int, int]]:
        param_info = []
        self._arguments_helper(param_info)
        return param_info
```

cuda_core/cuda/core/experimental/_module.py
oleksandr-pavlyk: Docstring style follows that of methods of `KernelAttributes`:

```python
In [1]: import cuda.core.experimental as cc

In [2]: o = cc.Program("__global__ void foo0(void) { }", code_type="c++").compile("cubin", name_expressions=("foo0",))

In [3]: k = o.get_kernel("foo0")

In [4]: k.num_arguments.__doc__  # access docs of `int`
Out[4]: "int([x]) -> integer\nint(x, base=10) -> integer\n\nConvert a number or string to an integer, or return 0 if no arguments\nare given.  If x is a number, return x.__int__().  For floating-point\nnumbers, this truncates towards zero.\n\nIf x is not a number or if base is given, then x must be a string,\nbytes, or bytearray instance representing an integer literal in the\ngiven base.  The literal can be preceded by '+' or '-' and be surrounded\nby whitespace.  The base defaults to 10.  Valid bases are 0 and 2-36.\nBase 0 means to interpret the base from the string as an integer literal.\n>>> int('0b100', base=0)\n4"

In [5]: type(k).num_arguments.__doc__  # access docs of the property
Out[5]: 'int : The number of arguments of this function'

In [6]: k.attributes.const_size_bytes.__doc__   # access docs of attribute method
Out[6]: 'int : The size in bytes of user-allocated constant memory required by this function.\n        This attribute is read-only.'
```

leofang: It seems you're right and we have a discrepancy within the codebase. This format is correct as per the NumPy style which we follow: https://www.sphinx-doc.org/en/master/usage/extensions/example_numpy.html. We'll need to ensure consistency in a separate PR.

That said, I think both Google/NumPy style guides are a bit outdated now that Sphinx can parse and render type annotations just fine. I'd prefer us to not duplicate it in the docstring if possible (#454).

cuda_core/cuda/core/experimental/_module.py
oleksandr-pavlyk: Good idea, thanks @rwgk . I have pushed the change.

cuda_core/cuda/core/experimental/_module.py
oleksandr-pavlyk: Ok, left as is for now, only fixed docstring for `arguments_info` to specify `list[tuple[int, int]]`.

cuda_core/cuda/core/experimental/_module.py
```diff
@@ -215,6 +216,33 @@ def attributes(self) -> KernelAttributes:
...
+        return arg_pos, param_info_data
+
+    @property
+    def num_arguments(self) -> int:
+        """int : The number of arguments of this function"""
+        num_args, _ = self._get_arguments_info()
+        return num_args
+
+    @property
+    def arguments_info(self) -> list[tuple[int, int]]:
```
rwgk:
```

@dataclass
class ParamInfo:  # maybe this is not the best name
    offset: int
    size: int
```

could make the code in the call sites more readable. — My experience passing anonymous tuples around isn't great: easily misunderstood, prone to accidents in refactoring work.

rwgk (dismissed):
LGTM although I'd use a `dataclass` for better ergonomics.

cuda_core/cuda/core/experimental/_module.py
leofang: How about `namedtuple`? 
https://github.com/NVIDIA/cuda-python/blob/ee6b92ef4f204abfa9a2a586693c915777631517/cuda_core/cuda/core/experimental/_utils/cuda_utils.py#L30
btw IIRC we currently use `dataclass` for input options and `namedtuple` for return types, but it is not a must-obey rule, and also I am a bit hesitant about introducing a type for this simple thing

rwgk: > btw IIRC we currently use `dataclass` for input options and `namedtuple` for return types, but it is not a must-obey rule,

`namedtuple` is great, too

> and also I am a bit hesitant about introducing a type for this simple thing

usually it backfires — maybe not for one particular thing, but in aggregate almost certainly

oleksandr-pavlyk: Change to use `namedtuple` has been pushed.

oleksandr-pavlyk:
I fixed pre-commit, and added a line to `"0.3.0-notes.rst"`.

cuda_core/cuda/core/experimental/_module.py
```diff
@@ -215,6 +217,35 @@ def attributes(self) -> KernelAttributes:
             self._attributes = KernelAttributes._init(self._handle)
         return self._attributes
 
+    def _get_arguments_info(self, param_info=False) -> tuple[int, list[NamedTuple]]:
+        attr_impl = self.attributes
+        if attr_impl._backend_version != "new":
+            raise NotImplementedError("New backend is required")
+        arg_pos = 0
+        if param_info:
+            ParamInfo = namedtuple("ParamInfo", ["offset", "size"])
```
rwgk: Suggestion to move this to class scope, I think then it reads really well:
```
Kernel.ParamInfo
```

cuda_core/cuda/core/experimental/_module.py
```diff
@@ -215,6 +217,35 @@ def attributes(self) -> KernelAttributes:
...
+        param_info_data = []
+        while True:
+            result = attr_impl._loader["paraminfo"](self._handle, arg_pos)
+            if result[0] != driver.CUresult.CUDA_SUCCESS:
+                break
+            if param_info:
+                param_info_data.append(ParamInfo._make(result[1:3]))
```
rwgk:
```
ParamInfo(*result[1:3])
```
is more modern.

rwgk (approved):
Two minor suggestions, please see here for background:

https://chatgpt.com/share/681d1a68-97d0-8008-bad5-40ae287f7d55

(When I wrote the first prompt I had "module scope" in mind, but I think class scope is better.)

cuda_core/cuda/core/experimental/_module.py
oleksandr-pavlyk: Don, it does make it cleaner. pushed

cuda_core/cuda/core/experimental/_module.py
oleksandr-pavlyk: changed

oleksandr-pavlyk:
This perplexes me:

```
(dev-cuda-python) opavlyk@ee09c48-lcedt:~/repos/cuda-python/cuda_core$ ipython
Python 3.12.10 | packaged by conda-forge | (main, Apr 10 2025, 22:21:13) [GCC 13.3.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 9.2.0 -- An enhanced Interactive Python. Type '?' for help.
Tip: Use `object?` to see the help on `object`, `object??` to view its source

In [1]: import cuda.core.experimental as cc

In [2]: o1 = cc.Program("__global__ void bar(double *p, int n, double x) { *p = n * x; }", code_type="c++").compile("cubin", name_expressions=("bar",))

In [3]: k1 = o1.get_kernel("bar")

In [4]: k1.arguments_info
Out[4]: []

In [5]: k1.num_arguments
Out[5]: 0
```

rwgk (dismissed):
This looks perfect to me. Caveat: I cannot explain the ipython observation.

leofang:
> This perplexes me

Let me look into it tmr

rwgk:
> The behavior is as expected if a device is initialized and set current manually.

Would it make sense to raise an exception in the new properties if the device isn't initialized?

oleksandr-pavlyk:
> > The behavior is as expected if a device is initialized and set current manually.
> 
> Would it make sense to raise an exception in the new properties if the device isn't initialized?

Yes, I just pushed the change to do that. With it a nice error results if no initialization was performed.

<details>
<summary>Open to see the error reported now</summary>

```
Python 3.12.10 | packaged by conda-forge | (main, Apr 10 2025, 22:21:13) [GCC 13.3.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 9.2.0 -- An enhanced Interactive Python. Type '?' for help.
Tip: You can find how to type a Unicode symbol by back-completing it, eg `\Ⅷ<tab>` will expand to `\ROMAN NUMERAL EIGHT`.

In [1]: import cuda.core.experimental as cc

In [2]: o1 = cc.Program("__global__ void bar(double *p, int n, double x) { *p = n * x; }", code_type="c++").compile("cubin", name_expressions=("bar",))

In [3]: k1 = o1.get_kernel("bar")

In [4]: k1.num_arguments, k1.arguments_info
---------------------------------------------------------------------------
CUDAError                                 Traceback (most recent call last)
Cell In[4], line 1
----> 1 k1.num_arguments, k1.arguments_info

File ~/repos/cuda-python/cuda_core/cuda/core/experimental/_module.py:242, in Kernel.num_arguments(self)
    239 @property
    240 def num_arguments(self) -> int:
    241     """int : The number of arguments of this function"""
--> 242     num_args, _ = self._get_arguments_info()
    243     return num_args

File ~/repos/cuda-python/cuda_core/cuda/core/experimental/_module.py:236, in Kernel._get_arguments_info(self, param_info)
    234     arg_pos = arg_pos + 1
    235 if result[0] != driver.CUresult.CUDA_ERROR_INVALID_VALUE:
--> 236     handle_return(result)
    237 return arg_pos, param_info_data

File ~/repos/cuda-python/cuda_core/cuda/core/experimental/_utils/cuda_utils.py:105, in handle_return(result, handle)
    104 def handle_return(result, handle=None):
--> 105     _check_error(result[0], handle=handle)
    106     if len(result) == 1:
    107         return

File ~/repos/cuda-python/cuda_core/cuda/core/experimental/_utils/cuda_utils.py:87, in _check_error(error, handle)
     85 def _check_error(error, handle=None):
     86     if isinstance(error, driver.CUresult):
---> 87         _check_driver_error(error)
     88     elif isinstance(error, runtime.cudaError_t):
     89         _check_runtime_error(error)

File ~/repos/cuda-python/cuda_core/cuda/core/experimental/_utils/cuda_utils.py:60, in _check_driver_error(error)
     58 expl = DRIVER_CU_RESULT_EXPLANATIONS.get(int(error))
     59 if expl is not None:
---> 60     raise CUDAError(f"{name}: {expl}")
     61 desc_err, desc = driver.cuGetErrorString(error)
     62 if desc_err != driver.CUresult.CUDA_SUCCESS:

CUDAError: CUDA_ERROR_INVALID_CONTEXT: This most frequently indicates that there is no context bound to the current thread. This can also be returned if the context passed to an API call is not a valid handle (such as a context that has had cuCtxDestroy() invoked on it). This can also be returned if a user mixes different API versions (i.e. 3010 context with 3020 API calls). See cuCtxGetApiVersion() for more details. This can also be returned if the green context passed to an API call was not converted to a CUcontext using cuCtxFromGreenCtx API.

In [5]: quit
```
</details>

rwgk (dismissed):
Ah! Funny that we all overlooked that.

Is it difficult to cover this with a unit test? (I don't know how tricky it is to set up the test so that no device is initialized.)

oleksandr-pavlyk:
> Is it difficult to cover this with a unit test?

Should not be. We just need to not use `cuda_init` fixture in the test.

leofang (approved):

leofang:
- Need to skip the tests when the sanitizer is in use: `@skipif_testing_with_compute_sanitizer`
- Need to skip when `cuda.bindings` is 11.x

oleksandr-pavlyk:
@leofang I pushed a change to add skips you suggested to implement.

cuda_core/tests/test_module.py
```diff
@@ -6,15 +6,22 @@
 # this software and related documentation outside the terms of the EULA
 # is strictly prohibited.
 
+try:
+    from cuda.bindings import driver
+except ImportError:
+    from cuda import cuda as driver
```
leofang: FYI this can be done with
```suggestion
from cuda.core.experimental._utils.cuda_utils import driver
```
(We might want to group all cuda.core imports to the block below.)

cuda_core/tests/test_module.py
```diff
@@ -29,6 +36,14 @@
 """
 
 
+@pytest.fixture(scope="module")
```
leofang, line 35: nit: perhaps this can be scoped to session?

cuda_core/tests/test_module.py
```diff
@@ -29,6 +36,14 @@
...
+def cuda_version():
+    # binding availability depends on cuda-python version
+    _py_major_ver, _ = cuda_utils.get_binding_version()
+    _driver_ver = cuda_utils.handle_return(driver.cuDriverGetVersion())
+    return _py_major_ver, _driver_ver
```
leofang: I think this check can be more specific
```suggestion
def is_using_old_backend():
    # binding availability depends on cuda-python version
    _py_major_ver, _ = cuda_utils.get_binding_version()
    _driver_ver = cuda_utils.handle_return(driver.cuDriverGetVersion())
    return _py_major_ver < 12 or _driver_ver < 12000
```
because both checks matters in our CI (whereas currently only the driver version is used). For example, I believe our CI  test this combo: `cuda-core` + `cuda-bindings` v11 + latest driver v12.

cuda_core/tests/test_module.py
```diff
@@ -162,3 +177,77 @@ def test_object_code_load_cubin_from_file(get_saxpy_kernel, tmp_path):
 def test_object_code_handle(get_saxpy_object_code):
     mod = get_saxpy_object_code
     assert mod.handle is not None
+
+
+@skipif_testing_with_compute_sanitizer
+def test_saxpy_arguments(get_saxpy_kernel, cuda_version):
+    _, dr_ver = cuda_version
+    if dr_ver < 12:
```
leofang:
```suggestion
def test_saxpy_arguments(get_saxpy_kernel, is_using_old_backend):
    if is_using_old_backend:
```

cuda_core/tests/test_module.py
```diff
@@ -162,3 +177,77 @@ def test_object_code_load_cubin_from_file(get_saxpy_kernel, tmp_path):
...
+    expected_sizes = tuple(m.size for m in members)
+    assert all(actual == expected for actual, expected in zip(sizes, expected_sizes))
+
+
+@skipif_testing_with_compute_sanitizer
+@pytest.mark.parametrize("nargs", [0, 1, 2, 3, 16])
+@pytest.mark.parametrize("c_type_name,c_type", [("int", ctypes.c_int), ("short", ctypes.c_short)], ids=["int", "short"])
+def test_num_arguments(init_cuda, nargs, c_type_name, c_type, cuda_version):
+    _, dr_ver = cuda_version
+    if dr_ver < 12:
```
leofang:
```suggestion
def test_num_arguments(init_cuda, nargs, c_type_name, c_type, is_using_old_backend):
    if is_using_old_backend:
```

cuda_core/tests/test_module.py
```diff
@@ -162,3 +177,77 @@ def test_object_code_load_cubin_from_file(get_saxpy_kernel, tmp_path):
...
+
+    arg_info = krn.arguments_info
+    assert all([actual.offset == expected.offset for actual, expected in zip(arg_info, members)])
+    assert all([actual.size == expected.size for actual, expected in zip(arg_info, members)])
+
+
+@skipif_testing_with_compute_sanitizer
+def check_num_args_error_handling(deinit_cuda, cuda_version):
+    _, dr_ver = cuda_version
+    if dr_ver < 12:
```
leofang:
```suggestion
def check_num_args_error_handling(deinit_cuda, is_using_old_backend):
    if is_using_old_backend:
```

cuda_core/cuda/core/experimental/_module.py
```diff
@@ -215,6 +218,36 @@ def attributes(self) -> KernelAttributes:
...
+        return arg_pos, param_info_data
+
+    @property
+    def num_arguments(self) -> int:
+        """int : The number of arguments of this function"""
+        num_args, _ = self._get_arguments_info()
+        return num_args
+
+    @property
+    def arguments_info(self) -> list[NamedTuple]:
```
leofang: Q: Shouldn't the return type be `list[ParamInfo]`? (same for the docstring below)

leofang (approved):

cuda_core/tests/test_module.py
```diff
@@ -29,6 +32,15 @@
 """
 
 
+@pytest.fixture(scope="module")
+def cuda12_prerequisite_check():
+    # binding availability depends on cuda-python version
+    # and version of underlying CUDA toolkit
+    _py_major_ver, _ = get_binding_version()
```
rwgk, line 39: Very minor, but

`binding_major_ver`?

(`py` here throws me off a bit)

Also, I'm surprised to see local variables prefixed with an underscore. (If we get to that level of using underscores, they almost lose their meaning?)

rwgk (approved):

cuda_core/tests/test_module.py
oleksandr-pavlyk: I took it from https://github.com/NVIDIA/cuda-python/blob/main/cuda_core/tests/test_device.py#L26-L31

Let me defer renaming of these variables to a different PR.

cuda_core/cuda/core/experimental/_module.py
oleksandr-pavlyk: Interestingly using `list[ParamInfo]` refers to `cuda.core.experimental._module.ParamInfo`, whereas the `ParamInfo` was declared in the class scope. 

Using `list[Kernel.ParamInfo]` runs into errors, since `Kernel` class is not defined yet. So I made a change to use `list[ParamInfo]` and moved definition of `ParamInfo` to the module scope.

cuda_core/tests/test_module.py
rwgk: Sounds good. — Coincidentally I only saw a few minutes ago that we have similar existing code already.

leofang (approved):

oleksandr-pavlyk:
Test that `num_arguments` should raise if CUDA has not been initialized failed:

```
_________________________ test_num_args_error_handling _________________________

deinit_cuda = None, cuda12_prerequisite_check = True

    @skipif_testing_with_compute_sanitizer
    def test_num_args_error_handling(deinit_cuda, cuda12_prerequisite_check):
        if not cuda12_prerequisite_check:
            pytest.skip("Test requires CUDA 12")
        src = "__global__ void foo(int a) { }"
        prog = Program(src, code_type="c++")
        mod = prog.compile(
            "cubin",
            name_expressions=("foo",),
        )
        krn = mod.get_kernel("foo")
>       with pytest.raises(CUDAError):
E       Failed: DID NOT RAISE <class 'cuda.core.experimental._utils.cuda_utils.CUDAError'>

tests/test_module.py:246: Failed
```

Did I misunderstand the intent of `deinit_cuda` fixture?

leofang:
I feel `deinit_cuda` might be a confusing name. The purpose of it is to clean up the per-thread device instances + ensure no CUDA context is set to current _after_ the test finishes. Did you mean to do this _before_ the test starts?

oleksandr-pavlyk:
I meant it to ensure that cuda is not initialized at the start if this test. Kind of like opposite of `cuda_init` fixture

leofang:
Yes feel free to implement any fixture you need. No need to refactor the existing fixtures if you don't want to sweat in this PR -- but then we should create an issue to track it (to the very least, we should rename `deinit_cuda` to something like `deinit_device_at_teardown`).

To make sure we understand "initialized" -- You don't mean `driver.cuInit(0)`, do you? Note that it is a once-per-process call, effectively as if `std::call_once()` is used (I dunno how it's implemented exactly in the driver), so any subsequent calls are cheap no-ops.

cuda_core/tests/test_module.py
```diff
@@ -162,3 +174,85 @@ def test_object_code_load_cubin_from_file(get_saxpy_kernel, tmp_path):
...
+        pytest.skip("Test requires CUDA 12")
+    src = "__global__ void foo(int a) { }"
+    prog = Program(src, code_type="c++")
+    mod = prog.compile(
+        "cubin",
+        name_expressions=("foo",),
+    )
+    krn = mod.get_kernel("foo")
+    # Unset current context using function from conftest
+    while True:
```
rwgk: Question: Could this `while` loop be moved into `deinit_context_function`?

Suggestion (outline only):

```
    for _ in range(100):
        ... code you have here ...
    else:
        raise RuntimeError("Unexpected number of loop iterations.")
```

This is to prevent a stuck CI in case there is a bug elsewhere. (Been there several times.)

cuda_core/cuda/core/experimental/_module.py
```diff
@@ -215,6 +220,36 @@ def attributes(self) -> KernelAttributes:
             self._attributes = KernelAttributes._init(self._handle)
         return self._attributes
 
+    def _get_arguments_info(self, param_info=False) -> tuple[int, list[ParamInfo]]:
+        attr_impl = self.attributes
+        if attr_impl._backend_version != "new":
+            raise NotImplementedError("New backend is required")
+        arg_pos = 0
+        param_info_data = []
+        while True:
```
rwgk, line 229: Maybe (sorry I overlooked this before):

```python
        for arg_pos in itertools.count():
```

Then you don't need `arg_pos = 0` above and `arg_pos = arg_pos + 1` below.

cuda_core/tests/conftest.py
```diff
@@ -31,24 +31,45 @@ def init_cuda():
...
+@pytest.fixture(scope="function")
+def deinit_all_contexts_function():
+    def pop_all_contexts():
+        max_iters = 256
+        for _ in range(max_iters):
+            if _device_unset_current():
+                # context was popped, continue until stack is empty
+                continue
+            # no active context, we are ready
+            break
```
rwgk, line 68: Maybe shorter (replace 5 lines with 2):

```python
            if not _device_unset_current():
                break
```

rwgk (approved):
Just a couple cosmetic suggestions. Optional.

PR #618, opened 2025-05-06
Implement device and context properties for Event

### LINKED ISSUE #491: `Event` should provide an inverse lookup to its origin `Device`/`Context`
Follow-up of #487. Like CUDA streams, events today are also tied to a CUDA context, so it makes sense to offer the reverse link when an `Event` is created. This is particularly useful for bookkeeping, since unlike CUDA streams there is no driver/runtime API for querying the associated context.

NaderAlAwar:
## Description

closes #491

## Checklist
- [X] New or existing tests cover these changes.

### DIFF
--- a/cuda_core/cuda/core/experimental/_event.py
+++ b/cuda_core/cuda/core/experimental/_event.py
@@ -9,7 +9,6 @@ from dataclasses import dataclass
 from typing import TYPE_CHECKING, Optional
 
 from cuda.core.experimental._context import Context
-from cuda.core.experimental._device import Device
 from cuda.core.experimental._utils.cuda_utils import (
     CUDAError,
     check_or_create_options,
@@ -22,6 +21,7 @@ from cuda.core.experimental._utils.cuda_utils import (
 
 if TYPE_CHECKING:
     import cuda.bindings
+    from cuda.core.experimental._device import Device
 
 
 @dataclass
@@ -96,7 +96,7 @@ class Event:
     __slots__ = ("__weakref__", "_mnff", "_timing_disabled", "_busy_waited")
 
     @classmethod
-    def _init(cls, options: Optional[EventOptions] = None):
+    def _init(cls, device_id: int, ctx_handle=None, options: Optional[EventOptions] = None):
         self = super().__new__(cls)
         self._mnff = Event._MembersNeededForFinalize(self, None)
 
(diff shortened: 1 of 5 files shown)


cuda_core/cuda/core/experimental/_event.py
```diff
@@ -111,6 +113,8 @@ def _init(cls, options: Optional[EventOptions] = None):
         if options.support_ipc:
             raise NotImplementedError("WIP: https://github.com/NVIDIA/cuda-python/issues/103")
         self._mnff.handle = handle_return(driver.cuEventCreate(flags))
+        self._device_id = int(handle_return(driver.cuCtxGetDevice()))
+        self._ctx_handle = handle_return(driver.cuStreamGetCtx(self._mnff.handle))
```
leofang: Let's avoid driver calls if possible to reduce overhead. The same info can be passed in from within the two public constructors (I think!) `Device.create_event` and `Stream.record`.

cuda_core/cuda/core/experimental/_event.py
```diff
@@ -198,3 +202,22 @@ def handle(self) -> cuda.bindings.driver.CUevent:
             handle, call ``int(Event.handle)``.
         """
         return self._mnff.handle
+
+    @property
+    def device(self) -> Device:
```
leofang, line 207: it seems we have a circular import here, I seemed to solve it once in the past, let me look it up what I did...

cuda_core/tests/test_event.py
```diff
@@ -174,3 +174,15 @@ def test_error_timing_incomplete():
     arr[0] = 1
     event3.sync()
     event3 - event1  # this should work
+
+
+def test_event_device(init_cuda):
+    event = Device().create_event(options=EventOptions())
+    device = event.device
+    assert isinstance(device, Device)
```
rwgk: The docstring ...
```
Return the :obj:`~_device.Device` singleton associated with this event.
```
mentions "singleton". → Could this be

```
device = Device()
event  = device.create_event(options=EventOptions())
assert event.device is device
```
?

leofang:
LGTM! Let's add a release note entry to 0.3.0

leofang:
Thanks, Nader! Since the CI was green, let me admin-merge.

PR #620, opened 2025-05-07
CI: Reference workflow file name for old branch

cryos:
## Description

Switch to use the workflow file name as the name is getting missed in the merged code in main. Verified the behavior locally too. This is a follow up from #555 addressing an issue seen after merging.

## Checklist

### DIFF
--- a/.github/workflows/test-wheel-linux.yml
+++ b/.github/workflows/test-wheel-linux.yml
@@ -234,7 +234,7 @@ jobs:
 
           OLD_BRANCH=$(cat .github/BACKPORT_BRANCH)
           OLD_BASENAME="cuda-bindings-python${PYTHON_VERSION_FORMATTED}-cuda*-${{ inputs.host-platform }}*"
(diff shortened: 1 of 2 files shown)


cryos:
@leofang see [failures after merge](https://github.com/NVIDIA/cuda-python/actions/runs/14883762212) where the workflow name is not found. In general I have used workflow file names and they have been reliable.

leofang (approved):
Thanks, Marcus! Yes, I noticed the main failed too. So instead of backporting the CI name change, we just fix it by referring to the workflow filename. (I did not know this works!) The only catch is that when the current main branch becomes the backport branch in the future, we'll need to change it to `ci.yaml`.

leofang:
One question: Why didn't we observe this failure in #555?

cryos:
> One question: Why didn't we observe this failure in #555?

Not clear and that is frustrating! I guess we hit a corner case, I will merge this and see if it is more reliable.

cryos:
> Thanks, Marcus! Yes, I noticed the main failed too. So instead of backporting the CI name change, we just fix it by referring to the workflow filename. (I did not know this works!) The only catch is that when the current main branch becomes the backport branch in the future, we'll need to change it to `ci.yaml`.

Yes, and I think I will have factored out some of this common code into helper scripts better by that point too.

PR #623, opened 2025-05-10
Remove nvvm-related `LD_LIBRARY_PATH`, `PATH` manipulations in fetch_ctk action.

rwgk:
## Description
Currently we do not need any `LD_LIBRARY_PATH` or `PATH` manipulations for `nvvm` in any of our testing environments.

Possible explanation (I have NOT verified this): Either `nvvm` is installed via a Python wheel, or `CUDA_HOME` is set.


leofang (approved):
Nice, thanks Ralf!

Issue #274, opened 2024-12-06
Building using versions of cuda-python with the new layout breaks runtime compatibility with older versions


vyasr:
RAPIDS started rolling out new pinnings to use cuda-python 11.8.5/12.6.2 at build-time. Once we did so, we started observing some cascading issues like [this one](https://github.com/rapidsai/cuml/pull/6149#issuecomment-2521302938). Specifically, once [rmm](https://github.com/rapidsai/rmm/) was built using the new-layout cuda-python versions, we started seeing the following error in downstream repositories that were still pinned to older versions of cuda-python due to #215 and #226:
```
  File "device_buffer.pyx", line 1, in init rmm.pylibrmm.device_buffer
ModuleNotFoundError: No module named 'cuda.bindings'
```

The last frame in the traceback always points to the initialization of the `rmm.pylibrmm.device_buffer` module. This indicates to me that parts of cuda-python that are cimported into this module are embedding the cuda.bindings namespace into the module initialization in a way that is likely defeating the trampoline modules that were added to cuda-python for backwards compatibility, thus making the rmm modules compiled against new-layout cuda-python incompatible with runtime usage of old-layout cuda-python.

My guess is that some of the same issues that are causing us to have to manually do the `__pyx_capi__` definition are causing this. Cython simply isn't designed for mismatching layouts in this way, and my guess is that some of the objects that it defines internally in the cuda-bindings module are being copied directly over to the legacy cuda modules even though that isn't what was intended, resulting in those modules having internal objects that still specify the new-layout module names and then break consumers in the mixed build/runtime version case.

I'm not sure this will be fixable without further interactions with Cython internals like the `__pyx_capi__` change. It may not be worthwhile, and at this point it might make sense for cuda-python to simply advocate a clean break.

leofang:
Spoke to Vyas while I was on site. We did not have a clear vision whether this is even fixable or just something we have to document/keep track, and be sure to not introduce another ABI break in a minor release. (So far, we've had major breaking changes twice: 11.7.1 & 12.6.2). Now that our Cython expert @shwina is back, I'd like to have one final confirmation before we declare this as a "won't fix"...

vyasr:
At a high level my suspicion is that we are explicitly doing things that Cython is not designed to support by creating trampolines that do not reflect the original layout, and as such I suspect that we can only expect limited success in trying to make each of these cases work.

shwina:
[...]

PR #640, opened 2025-05-16
Always build and run Cython tests + other CI improvements 

### LINKED ISSUE #466: CI: Enable Cython tests on Windows runners
Currently this is blocked by MSVC not pre-installed on the VM image that we use (#457). Based on the past experience (#267), it takes way too long to just install MSVC as a job step (it took [about 15-20 mins](https://github.com/NVIDIA/cuda-python/actions/runs/12185874395/job/33993087004#step:4:1)). We need it pre-installed in the image.

### LINKED ISSUE #525: CI: Ensure all actions are pinned by commit instead of by version/branch

### LINKED ISSUE #639: Build Cython tests as part of build step as opposed to in test steps
Currently we build Cython tests via bash scripts at test time:
* https://github.com/NVIDIA/cuda-python/blob/71762af687ef697279e1065c71d5806a78f60bc2/.github/workflows/test-wheel-linux.yml#L335
* https://github.com/NVIDIA/cuda-python/blob/71762af687ef697279e1065c71d5806a78f60bc2/.github/workflows/test-wheel-linux.yml#L373

We should build these tests as part of a build process and retrieve the resulting artifacts to use at test time.

### LINKED ISSUE #644: CI: Set up a dependentbot to update GHA versions
xref: https://github.com/NVIDIA/cuda-python/pull/560#discussion_r2044831359

leofang:
## Description
- close #525
- close #466 
- close #639
- close #644
- tentatively pin at cython<3.1.0 (#642)

Previously, the Cython tests are run on
- Latest CUDA 11.x (11.8.0) and 12.x (12.9.0)
- Linux only 

With this PR, the Cython tests are run on
- Latest CUDA 12.x (12.9.0)
- Linux and Windows

We can no longer run against the previous major version because of the way of building the Cython tests is changed. It is now built against the build-time CTK, which is always the latest major.minor. But I think this is OK, because the same Cython tests are already run in each major version's branch ([example from 11.x](https://github.com/NVIDIA/cuda-python/blob/e290d52b779daff9961c3c18e30633a0052a31ea/.github/workflows/build-and-test.yml#L317)).

There is still one open task relevant to Cython tests (https://github.com/NVIDIA/cuda-python/issues/274#issuecomment-2824637343), but the requirement there is likely incompatible with either the old or new approach here and always needs a separate job/workflow, so I consider it out of scope.

## Checklist

### DIFF
--- a/.github/workflows/build-wheel.yml
+++ b/.github/workflows/build-wheel.yml
@@ -219,5 +219,5 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: ${{ env.CUDA_BINDINGS_ARTIFACT_NAME }}-tests
-          path: ${{ env.CUDA_BINDINGS_CYTHON_TESTS_DIR }}/test_*.${{ env.PY_EXT_SUFFIX }}
+          path: ${{ env.CUDA_BINDINGS_CYTHON_TESTS_DIR }}/test_*${{ env.PY_EXT_SUFFIX }}
           if-no-files-found: error
--- a/.github/workflows/build-wheel.yml
+++ b/.github/workflows/build-wheel.yml
@@ -48,12 +48,12 @@ jobs:
 
       - name: Set up Python
         id: setup-python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065  # v5.6.0
         with:
           python-version: ${{ matrix.python-version }}
         env:
           # we use self-hosted runners on which setup-python behaves weirdly...
-          AGENT_TOOLSDIRECTORY: "../RUNNER_TEMP"
+          AGENT_TOOLSDIRECTORY: "${{ env.RUNNER_TEMP }}"
 
       - name: Set up MSVC
         if: ${{ startsWith(inputs.host-platform, 'win') }}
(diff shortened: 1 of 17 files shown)


.github/workflows/build-wheel.yml
```diff
@@ -187,8 +188,52 @@ jobs:
...
+        with:
+          name: ${{ env.CUDA_BINDINGS_ARTIFACT_NAME }}-tests
+          path: ${{ env.CUDA_BINDINGS_CYTHON_TESTS_DIR }}/test_*${{ env.PY_EXT_SUFFIX }}
+          if-no-files-found: error
+
+      - name: Build cuda.core Cython tests
+        run: |
+          pip install $(ls ${{ env.CUDA_CORE_ARTIFACTS_DIR }}/*.whl)[test]
+          pushd ${{ env.CUDA_CORE_CYTHON_TESTS_DIR }}
+          bash build_tests.sh
```
kkraus14, line 231: It would be nice to have these built by a more standard mechanism and avoid needing to run a separate bash script. Any reason we wouldn't want these built when invoking `pip install`?

Doesn't need to be solved in this PR regardless.

kkraus14 (approved):

.github/workflows/build-wheel.yml
cryos: I had the same thought, I have some local changes (started on build) I will rebase once this is merged.

leofang: Because I don't want to include the built tests in the wheel. If there is a way to achieve this with the build frontend/backend, it'd certainly simplify `build-wheel.yml` here, but I couldn't find any.

PR #645, opened 2025-05-19
Test out a simple env-vars script for build

### LINKED ISSUE #603: CI: Explore consolidating Linux/Windows test workflows?
#600 opens up the door toward a single (Bash-based) implementation of the test workflow, which was previously not possible due to lack of Git Bash on the GH-hosted Windows GPU runner. It was the main reason that we had to re-implement `test-wheel-windows.yml` in Powershell in the first place (https://github.com/NVIDIA/cuda-python/pull/444#issue-2846097497). Note that our build workflow is already cross-platform (and Bash-based, since all of our build runners have Bash pre-installed -- trivial on Linux, and thanks to GH-hosted runner images on Windows).

There might still be a few kinks, so a naive unification likely won't work (ex: container vs VM image, gcc vs msvc, ...). But I think most job steps are enough identical that we could probably extract out a reusable workflow, e.g. `test-wheel-common.yml`, for both OS to call.

cryos:
## Description

Unify CI code to share common code between different workflows, closes #603.

## Checklist

### DIFF
--- a/ci/tools/env-vars
+++ b/ci/tools/env-vars
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+
+# Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. ALL RIGHTS RESERVED.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# A utility script to set up the GitHub environment variables for the CI.
+
+set -euo pipefail
+
+# Check if the script was called with exactly 1 argument
+if [[ ${#} -ne 1 ]]; then
+  echo "Error: This script requires exactly 1 argument (the build mode). You provided ${#}"
+  echo "Usage: ${0} build_mode[build or test]"
+  exit 1
+fi
+
+PYTHON_VERSION_FORMATTED=$(echo '${PY_VER}' | tr -d '.')
+if [[ "${HOST_PLATFORM}" == linux* ]]; then
+  CIBW_BUILD="cp${PYTHON_VERSION_FORMATTED}-manylinux*"
+  REPO_DIR=$(pwd)
+elif [[ "${HOST_PLATFORM}" == win* ]]; then
+  CIBW_BUILD="cp${PYTHON_VERSION_FORMATTED}-win_amd64"
+  PWD=$(pwd)
+  REPO_DIR=$(cygpath -w ${PWD})
+fi
+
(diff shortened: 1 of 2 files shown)


leofang:
LGTM so far. I see the assumption here is that the env vars are set in the CI and we just run the scripts, so the scripts are not meant to be run outside of the CI. This is how we avoid param checks (which would add a ton of boilerplates). Let me know whenever you're happy to merge!

leofang:
Up to you -- so far I've been able to follow just fine 🙂

ci/tools/env-vars
```diff
@@ -0,0 +1,77 @@
...
+    if [[ ${BUILD_CUDA_MINOR} != ${TEST_CUDA_MINOR} ]]; then
+      SKIP_CYTHON_TEST=1
+    else
+      SKIP_CYTHON_TEST=0
+    fi
+  fi
+  # We don't test compute-sanitizer on CTK<12 because backporting fixes is too much effort
+  # We only test compute-sanitizer on python 3.12 arbitrarily; we don't need to use sanitizer on the entire matrix
+  # Only local ctk installs have compute-sanitizer; there is no wheel for it
+  if [[ "${PY_VER}" == "3.12" && "${CUDA_VER}" != "11.8.0" && "${LOCAL_CTK}" == 1 && "${HOST_PLATFORM}" == linux* ]]; then
```
leofang, line 62: Thanks, good catch, this will fix #647.

ci/tools/env-vars
```diff
@@ -0,0 +1,77 @@
...
+    echo "LATEST_CUDA_VERSION=$(bash .github/workflows/guess_latest.sh)" >> $GITHUB_ENV
+    SETUP_SANITIZER=1
+  else
+    SETUP_SANITIZER=0
+    echo "SANITIZER_CMD=" >> $GITHUB_ENV
```
leofang, line 67: nit: no need to set it at this stage?

leofang (approved):
LGTM!

ci/tools/env-vars
cryos: You still want this for workflows that do not call `setup-sanitizer` as the variable would be unbound. It is nice not to force all workflows to call `setup-sanitizer`. We can remove it if we have all workflows call the setup, which thinking about it we could just do from the start of `run-tests` as it relies on the sanitizer stuff being ready there.

PR #648, opened 2025-05-20
Feature/occupancy

### LINKED ISSUE #504: Cover occupancy calculator APIs
We have them in the bindings: 
- driver: https://nvidia.github.io/cuda-python/cuda-bindings/latest/module/driver.html#occupancy
- runtime: https://nvidia.github.io/cuda-python/cuda-bindings/latest/module/runtime.html#occupancy

and this issue is about cuda.core exposure.

oleksandr-pavlyk:
## Description

closes #504 

## Checklist

### DIFF
--- a/cuda_core/cuda/core/experimental/_module.py
+++ b/cuda_core/cuda/core/experimental/_module.py
@@ -228,14 +228,14 @@ class KernelOccupancy:
         """ "int: The maximum cluster size that can be launched for this kernel and launch configuration"""
         drv_cfg = _to_native_launch_config(config)
         if stream is not None:
-            drv_cfg.hStream = stream._handle
+            drv_cfg.hStream = stream.handle
         return handle_return(driver.cuOccupancyMaxPotentialClusterSize(self._handle, drv_cfg))
 
     def max_active_clusters(self, config: LaunchConfig, stream: Optional[Stream] = None) -> int:
         """ "int: The maximum number of clusters that could co-exist on the target device"""
         drv_cfg = _to_native_launch_config(config)
         if stream is not None:
-            drv_cfg.hStream = stream._handle
+            drv_cfg.hStream = stream.handle
         return handle_return(driver.cuOccupancyMaxActiveClusters(self._handle, drv_cfg))
 
 
--- a/cuda_core/cuda/core/experimental/_module.py
+++ b/cuda_core/cuda/core/experimental/_module.py
@@ -205,7 +205,30 @@ class KernelOccupancy:
         return self
 
     def max_active_blocks_per_multiprocessor(self, block_size: int, dynamic_shared_memory_size: int) -> int:
-        """int : Occupancy of the kernel"""
+        """Occupancy of the kernel.
+
+        Returns the maximum number of active blocks per multiprocessor for this kernel.
+
+        Parameters
+        ----------
+            block_size: int
+                Block size parameter used to launch this kernel.
+            dynamic_shared_memory_size: int
+                The amount of dynamic shared memory in bytes needed by block.
+                Use `0` if block does not need shared memory.
+
+        Returns
+        -------
+        int
+            The maximum number of active blocks per multiprocessor.
+
+        Note
+        ----
+            The fraction of the product of maximum number of active blocks per multiprocessor
+            and the block size to the maximum number of threads per multiprocessor is known as
+            theoretical multiprocessor utilization (occupancy).
+
+        """
         return handle_return(
             driver.cuOccupancyMaxActiveBlocksPerMultiprocessor(self._handle, block_size, dynamic_shared_memory_size)
         )
@@ -262,20 +285,65 @@ class KernelOccupancy:
         return MaxPotentialBlockSizeOccupancyResult(min_grid_size=min_grid_size, max_block_size=max_block_size)
 
     def available_dynamic_shared_memory_per_block(self, num_blocks_per_multiprocessor: int, block_size: int) -> int:
-        """int: Dynamic shared memory available per block for given launch configuration."""
+        """Dynamic shared memory available per block for given launch configuration.
+
+        The amount of dynamic shared memory per block, in bytes, for given kernel launch configuration.
+
+        Parameter
+        ---------
+            num_blocks_per_multiprocessor: int
+                Number of blocks to be concurrently executing on a multiprocessor.
+            block_size: int
+                Block size parameter used to launch this kernel.
+
+        Returns
+        -------
+        int
+            Dynamic shared memory available per block for given launch configuration.
+        """
         return handle_return(
             driver.cuOccupancyAvailableDynamicSMemPerBlock(self._handle, num_blocks_per_multiprocessor, block_size)
         )
 
     def max_potential_cluster_size(self, config: LaunchConfig, stream: Optional[Stream] = None) -> int:
-        """ "int: The maximum cluster size that can be launched for this kernel and launch configuration"""
+        """Maximum potential cluster size.
+
+        The maximum potential cluster size for this kernel and given launch configuration.
+
+        Parameters
+        ----------
+            config: :obj:`~_launch_config.LaunchConfig`
+                Kernel launch configuration. Cluster dimensions in the configuration are ignored.
+            stream: :obj:`~Stream`, optional
+                The stream on which this kernel is to be launched.
+
+        Returns
+        -------
+        int
+            The maximum cluster size that can be launched for this kernel and launch configuration.
+        """
         drv_cfg = _to_native_launch_config(config)
         if stream is not None:
             drv_cfg.hStream = stream.handle
         return handle_return(driver.cuOccupancyMaxPotentialClusterSize(self._handle, drv_cfg))
 
     def max_active_clusters(self, config: LaunchConfig, stream: Optional[Stream] = None) -> int:
-        """ "int: The maximum number of clusters that could co-exist on the target device"""
+        """Maximum number of active clusters on the target device.
+
+        The maximum number of clusters that could concurrently execute on the target device.
+
+        Parameters
+        ----------
+            config: :obj:`~_launch_config.LaunchConfig`
+                Kernel launch configuration.
+            stream: :obj:`~Stream`, optional
+                The stream on which this kernel is to be launched.
+
+        Returns
+        -------
+        int
+            The maximum number of clusters that could co-exist on the target device.
+        """
         drv_cfg = _to_native_launch_config(config)
         if stream is not None:
             drv_cfg.hStream = stream.handle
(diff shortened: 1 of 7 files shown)


oleksandr-pavlyk:
Outstanding issues:

 - [ ] Write tests
 - [ ] Should we support variable dynamic shared memory size dependent on block size in `max_potential_block_size`?
 - [ ] In method names use fully spelled `shared_memory_size`, should `shmem_size` be used instead?
 - [ ] For cluster-related occupancy queries, driver API takes `CUlaunchConfig` struct which maps to pair `LaunchConfig` data class instance and `Stream` class instance, because the data class does not contain the stream information. Is this design decision acceptable?

oleksandr-pavlyk:
In case we do not want `numba` as test dependency (even optional), we could consider compiling the following `b2dsize.c` library:

```c
// gcc -fshared -fPIC b2dsize.c -o b2dsize.so
#include <stddef.h>

size_t dynamic_shared_memory_needed(int blockSize) {
  return (blockSize <= 32) ? (size_t)0 : (size_t)((blockSize - 1)  / 32) * ((size_t)1024);
}
```

Then

```
import ctypes

lib = ctypes.cdll.LoadLibrary("./b2dsize.so")
cfunc = lib.dynamic_shared_memory_needed
fn_ptr = ctypes.cast(cfunc, ctypes.c_void_p).value

dynamic_smem_needed_fn = driver.CUoccupancyB2DSize(_ptr = fn_ptr)
```

This would require compiler being available at test time, which is easy to arrange in conda, at least.

We could build a fixture that would build such a library, and skip the test if building step fails due to absent compiler.

leofang:
cc @dongxiao92 @pentschev @bandokihiro for vis

kkraus14:
> * document that Python callbacks are uncharted territory (due to GIL contention)

To expand on this and capture offline discussion... The concern here is that we have two global locks in play, 1 from the Python Global Interpreter Lock, and 1 from the CUDA driver. We risk running into the following situation:
* Thread 1: Releases the GIL and acquires the CUDA driver lock when calling `cuOccupancyMaxPotentialBlockSize`, gets blocked in waiting to reacquire the GIL for the callback function
* Thread 2: Potentially calls a CUDA API without releasing the GIL, gets blocked in waiting to acquire the CUDA driver lock which holding the GIL

This would lead to a deadlock and we've seen this behavior in the past, i.e. https://github.com/numba/numba/pull/4581

cuda_core/cuda/core/experimental/_module.py
```diff
@@ -184,6 +186,102 @@ def cluster_scheduling_policy_preference(self, device_id: int = None) -> int:
...
+
+    @classmethod
+    def _init(cls, handle):
+        self = super().__new__(cls)
+        self._handle = handle
+
+        return self
+
+    def max_active_blocks_per_multiprocessor(self, block_size: int, dynamic_shared_memory_size: int) -> int:
+        """int : Occupancy of the kernel"""
```
leofang: Would it be possible to expand the docstrings in the methods here (same for all methods below)? These are methods not properties, so I believe we should have short descriptions for the methods, input arguments, and the return value(s).

leofang: (`max_potential_block_size` below already did this)

cuda_core/cuda/core/experimental/_module.py
```diff
@@ -250,6 +349,13 @@ def arguments_info(self) -> list[ParamInfo]:
         _, param_info = self._get_arguments_info(param_info=True)
         return param_info
 
+    @property
+    def occupancy(self) -> KernelOccupancy:
+        """Get the read-only attributes of this kernel."""
```
leofang: nit: we already have `.attributes` above, let's disambiguate it
```suggestion
        """Get the occupancy information of this kernel."""
```

cuda_core/tests/test_module.py
```diff
@@ -248,6 +258,118 @@ def test_num_args_error_handling(deinit_all_contexts_function, cuda12_prerequisi
...
+    kernel, _ = get_saxpy_kernel
+
+    def shared_memory_needed(block_size: numba.intc) -> numba.size_t:
+        "Size of dynamic shared memory needed by kernel of this block size"
+        return 1024 * (block_size // 32)
+
+    b2dsize_sig = numba.size_t(numba.intc)
+    dsmem_needed_cfunc = numba.cfunc(b2dsize_sig)(shared_memory_needed)
+    fn_ptr = ctypes.cast(dsmem_needed_cfunc.ctypes, ctypes.c_void_p).value
+    b2dsize_fn = driver.CUoccupancyB2DSize(_ptr=fn_ptr)
```
leofang, line 318: Just for my own curiosity... Would this be a valid test alternative?
```python
>>> import ctypes
>>> PyB2DSizeFunc = ctypes.CFUNCTYPE(ctypes.c_size_t, ctypes.c_int)
>>> def py_b2d_size(block_size):
...     return 0
... 
>>> 
>>> b2d_size = PyB2DSizeFunc(py_b2d_size)
>>> fn_ptr = ctypes.cast(b2d_size, ctypes.c_void_p).value
>>> b2dsize_fn = driver.CUoccupancyB2DSize(_ptr=fn_ptr)
```
I dunno how it's possible, but the docs of `ctypes.CFUNCTYPE` says the GIL is released when calling this function 🤔

cuda_core/tests/test_module.py
```diff
@@ -248,6 +258,118 @@ def test_num_args_error_handling(deinit_all_contexts_function, cuda12_prerequisi
         _ = krn.num_arguments
 
 
+@pytest.mark.parametrize("block_size", [32, 64, 96, 120, 128, 256])
+@pytest.mark.parametrize("smem_size_per_block", [0, 32, 4096])
+def test_saxpy_occupancy_max_active_block_per_multiprocessor(get_saxpy_kernel, block_size, smem_size_per_block):
```
leofang: Q: Do we need to mention `saxpy` in every test name? Does it matter what kernel we use in tests?

leofang (commented):
Thanks, Sasha! LGTM overall, most comments below are doc-related. 

For example, we need to add `_launch_config.LaunchConfig`, `_module.KernelOccupancy`, etc, to `cuda_core/docs/source/api_private.rst` to get them rendered and cross-ref'd.

cuda_core/tests/test_module.py
kkraus14: My understanding is that it requires the GIL to make the function call, but then releases the GIL as part of the call, which means that this is still problematic?

oleksandr-pavlyk: `CFUNCTYPE` can be constructed with function pointer, or with Python callable:

```
In [1]: import ctypes

In [2]: factory = ctypes.CFUNCTYPE(ctypes.c_size_t, ctypes.c_int)

In [3]: factory(Ellipsis)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[4], line 1
----> 1 factory(Ellipsis)

TypeError: argument must be callable or integer function address
```

If it is a function callable, I suspect the GIL is likely never released.

If it is a pointer, `ctypes` takes care of unboxing, releases GIL and makes the C-function call, then reacquires GIL and boxes the result. 

Anyhow, the example works:

```
Python 3.12.10 | packaged by conda-forge | (main, Apr 10 2025, 22:21:13) [GCC 13.3.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 9.2.0 -- An enhanced Interactive Python. Type '?' for help.
Tip: You can find how to type a Unicode symbol by back-completing it, eg `\Ⅷ<tab>` will expand to `\ROMAN NUMERAL EIGHT`.

In [1]: import ctypes

In [2]: import cuda.core.experimental as cc

In [3]: import cuda.core.experimental._utils.cuda_utils as cuda_utils

In [4]: cc.Device(0).set_current()

In [5]: o1 = cc.Program("__global__ void bar(double *p, int n, double x) { *p = n * x; }", code_type="c++").compile("cubin", name_expressions=("bar",))

In [6]: k1 = o1.get_kernel("bar")

In [7]: PyB2DSizeFunc = ctypes.CFUNCTYPE(ctypes.c_size_t, ctypes.c_int)

In [8]: def py_b2d_size(block_size):
   ...:     return 128 * (block_size // 32)
   ...: 

In [9]: b2d_size = PyB2DSizeFunc(py_b2d_size)

In [10]: fn_ptr = ctypes.cast(b2d_size, ctypes.c_void_p).value

In [11]: b2d_arg = cuda_utils.driver.CUoccupancyB2DSize(_ptr=fn_ptr)

In [12]: k1.occupancy.max_potential_block_size(b2d_arg, 0)
Out[12]: MaxPotential(min_grid_size=168, max_block_size=768)
```

but since it likely holds GIL, I am not sure we should include it

cuda_core/tests/test_module.py
oleksandr-pavlyk: No, it does not matter. I will shorten test names.

cuda_core/tests/test_module.py
leofang: Yeah sorry I dropped the ball here. I dug into it a bit, I guess you're right and this is what the doc is trying to say based on [the `qsort` callback example](https://docs.python.org/3/library/ctypes.html#callback-functions):
- we pass a `ctypes.CFUNCTYPE` callback to a C function (ex: `qsort` or `cuOccupancyMaxPotentialBlockSizeWithFlags`)
- the GIL is released when the C function is run
- when the callback is invoked
   - if the callback is created from a functor, the GIL is not acquired
   - if the callback is created from a Python function, the GIL is acquired 

Based on our offline discussion ([which Keith kindly summarized above](https://github.com/NVIDIA/cuda-python/pull/648#issuecomment-2922919836)), we don't want to show the last example.

leofang: btw as part of the investigation I ran into this very nice blog post:
https://astr.cc/blog/what-is-libffi-anyway/
It does not answer our question on GIL but it does dive deeply into how the `CFUNCTYPE` magic works

cuda_core/cuda/core/experimental/_module.py
```diff
@@ -184,6 +186,170 @@ def cluster_scheduling_policy_preference(self, device_id: int = None) -> int:
...
+            )
+        return MaxPotentialBlockSizeOccupancyResult(min_grid_size=min_grid_size, max_block_size=max_block_size)
+
+    def available_dynamic_shared_memory_per_block(self, num_blocks_per_multiprocessor: int, block_size: int) -> int:
+        """Dynamic shared memory available per block for given launch configuration.
+
+        The amount of dynamic shared memory per block, in bytes, for given kernel launch configuration.
+
+        Parameter
+        ---------
```
leofang: It seems this confused Sphinx:
https://nvidia.github.io/cuda-python/pr-preview/pr-648/cuda-core/latest/generated/cuda.core.experimental._module.KernelOccupancy.html#cuda.core.experimental._module.KernelOccupancy
```suggestion
        Parameters
        ----------
```

leofang (approved):
LGTM, thanks Sasha! I made a doc-only fix. The CI was green so let me admin-merge to save some resources.

PR #669, opened 2025-06-01
Add tests to cover scalar handling in `launch()` + Fix fp16 bug

### LINKED ISSUE #260: Add tests to cover `cuda.core.experimental.launch()`
We support
- Python scalars
- NumPy scalars
- ctypes scalars

These should all be tested in `test_launcher.py`.

leofang:
## Description

closes #260.

The test coverage revealed a bug in the fp16 scalar handling. The existing handling is obviously wrong and it's me to blame, but I honestly can't recall why I believed the code was correct 😞 

## Checklist

### DIFF
--- a/cuda_core/cuda/core/experimental/_kernel_arg_handler.pyx
+++ b/cuda_core/cuda/core/experimental/_kernel_arg_handler.pyx
@@ -22,6 +22,33 @@ ctypedef cpp_complex.complex[float] cpp_single_complex
 ctypedef cpp_complex.complex[double] cpp_double_complex
 
 
+# We need an identifier for fp16 for copying scalars on the host. This is a minimal
+# implementation borrowed from cuda_fp16.h.
+cdef extern from *:
+    """
+    #if __cplusplus >= 201103L
+    #define __CUDA_ALIGN__(n) alignas(n)    /* C++11 kindly gives us a keyword for this */
+    #else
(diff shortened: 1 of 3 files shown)


cuda_core/cuda/core/experimental/_kernel_arg_handler.pyx
```diff
@@ -85,6 +113,8 @@ cdef inline int prepare_arg(
         (<supported_type*>ptr)[0] = cpp_complex.complex[float](arg.real, arg.imag)
     elif supported_type is cpp_double_complex:
         (<supported_type*>ptr)[0] = cpp_complex.complex[double](arg.real, arg.imag)
+    elif supported_type is __half_raw:
+        (<supported_type*>ptr).x = <int16_t>(arg.view(numpy_int16))
     else:
         (<supported_type*>ptr)[0] = <supported_type>(arg)
```
leofang, line 119: Here is the bug: When `arg` is a `np.float16` scalar, the old code would treat it as `int16_t` (due to lack of standard C++ identifier for fp16 before C++23), and the scalar would be `static_cast` to `int16_t`, which triggered non-trivial conversion operators. The new code ensures that there is no conversion and the bytes are `reinterpret_cast`'d, and in order to hit this new path we need a unique type identifier, which is `__half_raw`.

kkraus14 (approved):

Issue #496, opened 2025-03-06
RFC: Add support for `launch_attr` in `LaunchConfig` ctor


realarnavgoel:
Today, `LaunchConfig` only supports `cuLaunchKernel` driver API to launch kernels on a single GPU. When extending to broader usecases where there is a need for inter-SM synchronization or multi-GPU synchronization, one would need to use `cuLaunchCooperativeKernel` to launch kernels safely in a deadlock-free manner. To support this, one could extend `LaunchConfig(..., launch_attr=None)` with an optional `launch_attr` that could set equivalent cuda-python data-type for [`CUlaunchAttribute`](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1g6f6565b334be6bb3134868e10bbdd331).

**Background:**
This issue came out of discussion: https://github.com/NVIDIA/numba-cuda/issues/128#issuecomment-2702689412 where existing implementation of cuda driver bindings in `numba-cuda` uses `cuLaunchCooperativeKernel` or `cuLaunchKernel` based on the existence of `grid.sync()` in the kernel and in the effort to migrate it to `cuda.core`, one would need to provide the capability to select launch kernel API variant at runtime based on the `LaunchConfig`.

leofang:
To make it 100% clear: An internal dispatching to `cuLaunchCooperativeKernel()` is not necessary. `cuLaunchKernelEx()` is guaranteed to work if the attribute is set.

realarnavgoel:
> It seems we just need to add `LaunchConfig.cooperative_launch: bool = False`, and map it to `CU_LAUNCH_ATTRIBUTE_COOPERATIVE` when preparing the launch config for `cuLaunchKernelEx`. What would be the right way to test this capability? Write a dummy kernel that only calls `grid.sync()` and launch it with/without setting the boolean to True?

I think its more than just that using the appropriate launch attribute to avoid deadlocks/hangs. To make sure that the concerns are separated by design choices, I am separating the discussion into 2 options

[...]

PR #676, opened 2025-06-03
Support cooperative launch

leofang:
## Description

Part of #496. 

~~Merge after #648.~~

## Checklist


cuda_core/tests/test_launcher.py
```diff
@@ -152,3 +152,48 @@ def test_launch_scalar_argument(python_type, cpp_type, init_value):
...
+    ker = prog.compile("cubin").get_kernel("test_grid_sync")
+
+    # # Launch without setting cooperative_launch
+    # # Commented out as this seems to be a sticky error...
+    # config = LaunchConfig(grid=1, block=1)
+    # launch(s, config, ker)
+    # from cuda.core.experimental._utils.cuda_utils import CUDAError
+    # with pytest.raises(CUDAError) as e:
+    #     s.sync()
+    # assert "CUDA_ERROR_LAUNCH_FAILED" in str(e)
```
leofang, line 189: xref: https://github.com/NVIDIA/cuda-python/issues/496#issuecomment-2936273691

cuda_core/cuda/core/experimental/_launch_config.py
```diff
@@ -58,11 +58,15 @@ class LaunchConfig:
...
+    cooperative_launch: Optional[bool] = False
 
     def __post_init__(self):
         _lazy_init()
         self.grid = cast_to_3_tuple("LaunchConfig.grid", self.grid)
         self.block = cast_to_3_tuple("LaunchConfig.block", self.block)
+        # FIXME: Calling Device() strictly speaking is not quite right; we should instead
+        # look up the device from stream. We probably need to defer the checks related to
+        # device compute capability or attributes.
         # thread block clusters are supported starting H100
```
kkraus14, line 70: Instead of doing all of these checks in Python, if we just set the options as passed and try the launch, does CUDA error gracefully where we can post process the error code to make it meaningful to the user? Doing all of these checks in Python feels expensive and will hurt kernel launch latency.

cuda_core/cuda/core/experimental/_launcher.py
```diff
@@ -86,3 +89,17 @@ def launch(stream, config, kernel, *kernel_args):
...
+    dev = stream.device
+    num_sm = dev.properties.multiprocessor_count
+    max_grid_size = (
+        kernel.occupancy.max_active_blocks_per_multiprocessor(_reduce_tuple(config.block), config.shmem_size) * num_sm
+    )
+    if _reduce_tuple(config.grid) > max_grid_size:
+        # For now let's try not to be smart and adjust the grid size behind users' back.
+        # We explicitly ask users to adjust.
+        raise ValueError(
+            "The specified grid size ({} * {} * {}) exceeds the limit ({}).".format(*config.grid, max_grid_size)
```
kkraus14: nitpick: use f-string instead of `str.format`?

cuda_core/tests/test_launcher.py
```diff
@@ -152,3 +152,48 @@ def test_launch_scalar_argument(python_type, cpp_type, init_value):
 
     # Check result
     assert arr[0] == init_value, f"Expected {init_value}, got {arr[0]}"
+
+
+@pytest.mark.skipif(os.environ.get("CUDA_PATH") is None, reason="need cg header")
```
kkraus14: Requiring this environment variable feels a bit yucky. Is there any way we can handle this more nicely? Maybe something to revisit once we have `path_finder` support for headers?

cuda_core/cuda/core/experimental/_launch_config.py
leofang: I can try but let me ping @pciolkosz here in case he knows off top of his head if launching a thread block cluster kernel on a device that does not support it is also a sticky error 🙂 It is certainly the case for cooperative launch, Piotr confirmed my observation.

cuda_core/tests/test_launcher.py
leofang: Yes I do plan to have this re-written once path finder is up. We use this pattern in a few places. Let's do another PR that refactors this skip to a fixture first, and once path finder is up we change the fixture implementation.

cuda_core/cuda/core/experimental/_launch_config.py
pciolkosz: Launching a cluster on a GPU that does not support it is not a sticky error.
In general anything CUDA can verify ahead of the kernel execution, like configuration issues are not going to be a sticky error.
Missing the cooperative launch option is sort of an exception, because we don't have a way to tell that somewhere in your kernel you do `grid.sync()`. So its only observable during kernel execution and it is a sticky error, because its a trap in device code

leofang: Oh OK when you said it traps I thought it traps on host, did not realize it's on device. Couldn't the compiler emit a flag or something to the kernel image so that the driver can check prior to launch? I think there are already other metadata encoded by the compiler (ex: kernel argument sizes) so seems like an easy solution to me... ?

pciolkosz: It might be possible the compiler could detect access to the env-reg we are accessing to get the barrier pointer and emit elf attribute to mark the kernel, but I don't think we ever requested that. I think better solution is what we did with the launch function in `cudax` where the kernel configuration can be passed into the kernel. I plan to add cooperative groups integration where you would pass the configuration to the `grid_group` API and it would check its type if the cooperative launch option was added to it. This way you will get compilation error instead of runtime error if you forget the cooperative launch option

leofang: Right, that sounds like a good plan in C++, but not in Python 😛  I am curious how much work is needed to get the ELF attribute emitted. Let's discuss offline. 

I'll create an issue to track the needed changes for better thread block cluster checks. It's out of scope for this PR and it's not the most common code path that users would hit, anyway.

cuda_core/cuda/core/experimental/_launcher.py
leofang: done in commit 138f0d1

cuda_core/tests/test_launcher.py
leofang: a placeholder fixture is added in commit 167136e

cuda_core/cuda/core/experimental/_utils/cuda_utils.py
```diff
@@ -48,6 +48,10 @@ def cast_to_3_tuple(label, cfg):
     return cfg + (1,) * (3 - len(cfg))
 
 
+def _reduce_3_tuple(t: tuple):
+    return t[0] * t[1] * t[2]
```
kkraus14, line 52: Can we just use `math.prod` instead of this? Do we need to handle tuples larger than this length?

leofang:
```python
In [4]: %timeit _reduce_3_tuple( (7, 2, 5) )
48 ns ± 0.328 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [5]: %timeit prod( (7, 2, 5) )
62.8 ns ± 0.0386 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```
For now we always handle 3-tuples. If there are extra needs we should revisit.

cuda_core/cuda/core/experimental/_launch_config.py
leofang: > I'll create an issue to track the needed changes for better thread block cluster checks.

#685

leofang:
Thanks, Keith/Piotr!

PR #684, opened 2025-06-06
Add cufile bindings

### LINKED ISSUE #585: Add cuFile bindings

sourabgupta3:
## Description

closes #585 

## Checklist


leofang:
@sourabgupta3 there are some CI works I'll need to address later, but please proceed to test it locally. It builds on Linux now.

cuda_bindings/pyproject.toml
```diff
@@ -35,6 +35,7 @@ all = [
     "nvidia-cuda-nvcc-cu12",
     "nvidia-cuda-nvrtc-cu12",
     "nvidia-nvjitlink-cu12>=12.3",
+    "nvidia-cufile-cu12",
```
leofang:
```suggestion
    "nvidia-cufile-cu12; sys_platform == 'linux'",
```

cuda_bindings/tests/test_cufile.py
```diff
@@ -0,0 +1,1983 @@
...
+import tempfile
+from contextlib import suppress
+
+import pytest
+
+import cuda.bindings.driver as cuda
+from cuda.bindings import cufile
+
+# from cuda.bindings.cycufile import CUfileDescr_t, CUfileFileHandleType
+
```
leofang: nit: clean up
```suggestion
```

cuda_bindings/tests/test_cufile.py
```diff
@@ -0,0 +1,1983 @@
...
+
+def cufileLibraryAvailable():
+    """Check if cuFile library is available on the system."""
+    try:
+        # Try to initialize cuFile driver
+        cufile.driver_open()
+        cufile.driver_close()
```
leofang: Could you leave a comment explaining why we need to do this as part of the check? Do they have any side effect? For example, if we were only to check if a cuFile library exists and loadable, we could just call `cufile.get_version()` as done below or use the function pointer inspection trick discussed offline.

cuda_bindings/tests/test_cufile.py
```diff
@@ -0,0 +1,1983 @@
...
+
+        # Check cuFile library version
+        try:
+            # Get cuFile library version
+            version = cufile.get_version()
+            print(f"cuFile library version: {version}")
+
+            # Check if version is 1.14.1 or higher (1140)
+            if version < 1140:
```
leofang: As discussed offline, we need a comment here. Also, v1.14.1 is `1140`?!

cuda_bindings/tests/test_cufile.py
```diff
@@ -0,0 +1,1983 @@
...
+    """Test file handle registration with cuFile."""
+    # Initialize CUDA
+    (err,) = cuda.cuInit(0)
+    assert err == cuda.CUresult.CUDA_SUCCESS
+
+    err, device = cuda.cuDeviceGet(0)
+    assert err == cuda.CUresult.CUDA_SUCCESS
+
+    err, ctx = cuda.cuCtxCreate(0, device)
+    assert err == cuda.CUresult.CUDA_SUCCESS
```
leofang: This applies to everything below too:
1. What's the reason that we need to use a custom context instead of the primary context here?
2. If we must use a custom contest, we need to make sure it's destroyed at the end of the test, otherwise the side effect carries over to other tests

cuda_bindings/tests/test_cufile.py
```diff
@@ -0,0 +1,1983 @@
...
+
+    # Write test data using POSIX write
+    test_data = b"Test data for cuFile - POSIX write"
+    bytes_written = os.write(fd, test_data)
+
+    # Sync to ensure data is on disk
+    os.fsync(fd)
+
+    # Close and reopen with O_DIRECT for cuFile operations
+    os.close(fd)
```
leofang, line 141: Let's use pytest's `tmp_path` fixture to keep the code clean and let it handle cleanup. The same applies to everything below. I think the only place we need `os.open`/`os.close` is when we need to access the file descriptor directly. (It doesn't seem that `pathlib` allows us to open a file with flags, get a file object, and then retrieve its `fileno`.)
```suggestion
    # Create test file
    file_path = tmp_path / "test_handle_register.bin"

    # Write test data and close
    test_data = b"Test data for cuFile - POSIX write"
    bytes_written = file_path.write_bytes(fd, test_data)
```

cuda_bindings/tests/test_cufile.py
```diff
@@ -0,0 +1,1983 @@
...
+        # Register the handle
+        handle = cufile.handle_register(descr.ptr)
+
+        # Deregister the handle
+        cufile.handle_deregister(handle)
+
+    finally:
+        os.close(fd)
+        with suppress(OSError):
+            os.unlink(file_path)
```
leofang, line 163: ditto

cuda_bindings/tests/test_cufile.py
```diff
@@ -0,0 +1,1983 @@
...
+
+        # Perform async write
+        cufile.write_async(
+            int(handle),
+            int(buf_ptr),
+            ctypes.addressof(size_p),
+            ctypes.addressof(file_offset_p),
+            ctypes.addressof(buf_ptr_offset_p),
+            ctypes.addressof(bytes_written_p),
+            int(stream),
```
leofang, line 799: The design of internal codegen is to enable call-site casting with little user intervention. The use of ctypes raises a red flag to me. I think we might have transformed the signature wrong. Let me know which arguments are considered output of this function, and I can patch them for you. I think once we fix everything ctypes usage should be minimal

leofang (changes requested):
Left some quick comments. I haven't gone through the whole test suite yet.

cuda_bindings/tests/test_cufile.py
sourabgupta3: @leofang We can't use tmpfs to test cufile APIs, that's why I didn't use /tmp.

cuda_bindings/tests/test_cufile.py
sourabgupta3: read_async and write_async are APIs that expect pointer variables which will be filled up by the API in future. I am not sure if there is any other way to do it python right now.

cuda_bindings/tests/test_cufile.py
leofang: Oh! Interesting, what's the reason that `/tmp` can't be used? (I know very little about filesystems, so I am curious 🙂)

sourabgupta3: GPUDirectStorage was mainly for P2P(peer to peer) devices and tmpfs isn't exactly on a separate peer device, so it wasn't in our support matrix. But in a larger effort to support more kinds of filesystem regardless of whether they support P2P or not we are adding support for tmpfs starting 13.0.

cuda_bindings/tests/test_cufile.py
sourabgupta3: Fixed as per your suggestion.

cuda_bindings/tests/test_cufile.py
sourabgupta3: Cleaned up now.

cuda_bindings/tests/test_cufile.py
sourabgupta3: Instead of skipping all the tests based on library version, now only tests that are skipped are the one's that need 1.14 library

cuda_bindings/tests/test_cufile.py
sourabgupta3: Thanks! Using primary context now.

sourabgupta3:
/ok to test  [fbb4cdb](https://github.com/NVIDIA/cuda-python/pull/684/commits/fbb4cdb70a339656f8bc6e3bd24885c7b2bf0d4a)

leofang:
The CI was not run because the commit cannot be hyperlinked.  Must be a raw text. (GitHub does render commit texts to be clickable, so it's hard by eye-balling it, but it does make a difference to the copy-pr-bot...)

cuda_bindings/tests/test_cufile.py
leofang: Discussed offline. Sourab explained to me that the pointer arguments may not be populated by the time the API returns, so we can't make the signatures simpler otherwise we destroy the async/deferred capability.

cuda_bindings/tests/test_cufile.py
leofang: Thanks!

cuda_bindings/cuda/bindings/cufile.pyx
```diff
@@ -0,0 +1,1288 @@
...
+cdef int check_status(ReturnT status) except 1 nogil:
+    if ReturnT is CUfileError_t:
+        if status.err != 0 or status.cu_err != 0:
+            with gil:
+                raise cuFileError(status.err, status.cu_err)
+    elif ReturnT is ssize_t:
+        if status == -1:
+            # note: this assumes cuFile already properly resets errno in each API
+            with gil:
+                raise cuFileError(errno.errno)
```
leofang, line 988: @sourabgupta3 One question that came back to my mind. Did I miss a minus sign here (for `errno`)? The in-header docstring is not very clear about this.

sourabgupta3: No, this looks correct. The APIs will return -1 if the return type is ssize_t and errno will be set accodingly.

cuda_bindings/docs/source/conf.py
```diff
@@ -103,6 +104,7 @@
     "numpy": ("https://numpy.org/doc/stable/", None),
     "nvvm": ("https://docs.nvidia.com/cuda/libnvvm-api/", None),
     "nvjitlink": ("https://docs.nvidia.com/cuda/nvjitlink/", None),
+    "cufile": ("https://docs.nvidia.com/gpudirect-storage/api-reference-guide/", None),
```
leofang, line 107: As discussed offline, this currently does not work, but it's not a blocker for merging.

leofang (dismissed):
CI is all green!

cuda_bindings/cuda/bindings/cufile.pyx
```diff
@@ -0,0 +1,1288 @@
...
+
+cpdef int get_version() except? 0:
+    cdef int version
+    with nogil:
+        status = cuFileGetVersion(&version)
+    check_status(status)
+    return version
+
+
+cpdef get_parameter_size_t(int param, intptr_t value):
```
leofang: Thinking about this more, we probably could generate a more user friendly signature? 
```suggestion
cpdef size_t get_parameter_size_t(int param):
```
which would return a Python int. WDYT? (Same for the other two getter APIs.)

sourabgupta3: I agree.

leofang: done in commit [a3530ae](https://github.com/NVIDIA/cuda-python/pull/684/commits/a3530ae7ccac13db4b49e41b3b3b6c428daba891)

leofang:
/ok to test [01c8de9](https://github.com/NVIDIA/cuda-python/pull/684/commits/01c8de9240933eeeef8bcd159ffe32a7581e8d79)

leofang (dismissed):

cuda_bindings/docs/source/module/cufile.rst
```diff
@@ -0,0 +1,76 @@
...
+.. module:: cuda.bindings.cufile
+
+cufile
+======
+
+The ``cuda.bindings.cufile`` Python module wraps the
+`cuFile C APIs <https://docs.nvidia.com/gpudirect-storage/api-reference-guide/index.html>`_.
+Supported on Linux only.
+
+  * Currently using this module requires NumPy to be present. Any recent NumPy 1.x or 2.x should work.
```
leofang: I think this renders incorrectly. Let me apply this patch and admin-merge.
 ```suggestion
Currently using this module requires NumPy to be present. Any recent NumPy 1.x or 2.x should work.
```

leofang:
Since the CI was green and we only apply a doc change, let me admin-merge. Many thanks @sourabgupta3 for pushing this through the finish line with us!

PR #686, opened 2025-06-06
Add PTX helpers

### LINKED ISSUE #653: Add a helper function to map PTX ISA versions to the corresponding CTK versions
We want to bite the bullet and encode this table as a helper function in `cuda.bindings`:
https://docs.nvidia.com/cuda/parallel-thread-execution/#release-notes-ptx-release-history

Ex: PTX ISA 8.8 -> CTK 12.9

leofang:
## Description

Close #653.

## Checklist

### DIFF
--- a/cuda_bindings/cuda/bindings/utils/__init__.py
+++ b/cuda_bindings/cuda/bindings/utils/__init__.py
@@ -1,0 +1,131 @@
+# Copyright 2025 NVIDIA Corporation.  All rights reserved.
+# SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
+
+import re
+
+
+# Mapping based on the official PTX ISA <-> CUDA Release table
+# https://docs.nvidia.com/cuda/parallel-thread-execution/#release-notes-ptx-release-history
+_ptx_to_cuda = {
+    "1.0": (1, 0),
+    "1.1": (1, 1),
+    "1.2": (2, 0),
+    "1.3": (2, 1),
+    "1.4": (2, 2),
+    "2.0": (3, 0),
+    "2.1": (3, 1),
+    "2.2": (3, 2),
+    "2.3": (4, 0),
+    "3.0": (4, 1),
+    "3.1": (5, 0),
+    "3.2": (5, 5),
+    "4.0": (6, 0),
+    "4.1": (6, 5),
+    "4.2": (7, 0),
+    "4.3": (7, 5),
+    "5.0": (8, 0),
+    "6.0": (9, 0),
+    "6.1": (9, 1),
+    "6.2": (9, 2),
+    "6.3": (10, 0),
+    "6.4": (10, 1),
+    "6.5": (10, 2),
+    "7.0": (11, 0),
+    "7.1": (11, 1),
+    "7.2": (11, 2),
+    "7.3": (11, 3),
+    "7.4": (11, 4),
+    "7.5": (11, 5),
+    "7.6": (11, 6),
+    "7.7": (11, 7),
+    "7.8": (11, 8),
+    "8.0": (12, 0),
+    "8.1": (12, 1),
+    "8.2": (12, 2),
+    "8.3": (12, 3),
+    "8.4": (12, 4),
+    "8.5": (12, 5),
+    "8.6": (12, 7),
+    "8.7": (12, 8),
+    "8.8": (12, 9),
+}
+
+
+def get_minimal_required_driver_ver_from_ptx_ver(ptx_version: str) -> int:
+    """
+    Maps PTX ISA version to the minimal CUDA driver version.
+
+    Parameters
+    ----------
+    ptx_version : str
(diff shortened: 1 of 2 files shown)


cuda_bindings/cuda/bindings/utils/__init__.py
```diff
@@ -0,0 +1,131 @@
...
+    "8.3": (12, 3),
+    "8.4": (12, 4),
+    "8.5": (12, 5),
+    "8.6": (12, 7),
+    "8.7": (12, 8),
+    "8.8": (12, 9),
+}
+
+
+def get_minimal_required_driver_ver_from_ptx_ver(ptx_version: str) -> int:
```
kkraus14: Instead of `driver_ver` should we call this `cuda_ver`? Since someone could be using `nvjitlink` instead?

cuda_bindings/cuda/bindings/utils/__init__.py
```diff
@@ -0,0 +1,131 @@
...
+    try:
+        major, minor = _ptx_to_cuda[ptx_version]
+        return 1000 * major + 10 * minor
+    except KeyError:
+        raise ValueError(f"Unknown or unsupported PTX ISA version: {ptx_version}")
+
+
+
+# Regex pattern to match .version directive and capture the version number
+_ptx_ver_pattern = re.compile(r'\.version\s+([0-9]+\.[0-9]+)')
```
kkraus14: Maybe a future note that this will slow down import time which some people care about and that we could potentially lazily initialize this instead if needed?

cuda_bindings/cuda/bindings/utils/__init__.py
```diff
@@ -0,0 +1,131 @@
...
+    ...
+    ... .visible .entry test_kernel()
+    ... {
+    ...     ret;
+    ... }
+    ... '''
+    >>> get_ptx_ver(ptx)
+    '8.8'
+    """
+    m = _ptx_ver_pattern.search(ptx)
```
kkraus14: Based on https://docs.nvidia.com/cuda/parallel-thread-execution/#source-format, the first line is always going to be `.version`, so we'd probably be better off avoiding the regular expression and always just look at the beginning of the string?

cuda_bindings/cuda/bindings/utils/__init__.py
leofang: My experience is that regex compilation is quick, so I wouldn't worry too much about it. I did add a comment in commit b732b9d7.

cuda_bindings/cuda/bindings/utils/__init__.py
leofang: updated in commit b732b9d7

cuda_bindings/cuda/bindings/utils/__init__.py
leofang: The first line is generally a comment 😛 
```
//
// Generated by NVIDIA NVVM Compiler
//
// Compiler Build ID: CL-36037853
// Cuda compilation tools, release 12.9, V12.9.86
// Based on NVVM 7.0.1
//

.version 8.8
.target sm_75
.address_size 64

	// .globl	empty_kernel

.visible .entry empty_kernel()
{



	ret;

}
```
And if the PTX is hand-written instead of generated by a compiler, the formatting might be off (ex: extra spaces here and there). Using regex should the most robust approach without making any assumption or implementing any custom string processing code.

leofang:
/ok to test [3b60964](https://github.com/NVIDIA/cuda-python/pull/686/commits/3b60964c1fe5ec06790f21fde15056a16335abbf)

leofang:
cc @gmarkall for vis (context is https://github.com/NVIDIA/numba-cuda/issues/262)

cuda_bindings/cuda/bindings/utils/_ptx_utils.py
```diff
@@ -0,0 +1,131 @@
...
+    try:
+        major, minor = _ptx_to_cuda[ptx_version]
+        return 1000 * major + 10 * minor
+    except KeyError:
+        raise ValueError(f"Unknown or unsupported PTX ISA version: {ptx_version}") from None
+
+
+# Regex pattern to match .version directive and capture the version number
+# TODO: if import speed is a concern, consider lazy-initializing it.
+_ptx_ver_pattern = re.compile(r"\.version\s+([0-9]+\.[0-9]+)")
```
kkraus14, line 90: I just ran a quick test on my (WSL) environment and this takes ~100ns, so I can't imagine us needing to move it to lazy-initializing. Was quite surprised how fast this was 😄

kkraus14 (approved):

leofang:
Thanks, Keith!

cuda_bindings/cuda/bindings/utils/_ptx_utils.py
leofang: I did notice `re.compile` is not easy to benchmark, because it has an internal cache too.

Issue #685, opened 2025-06-06
Rely more on the driver for input checks to reduce latency


leofang:
> Instead of doing all of these checks in Python, if we just set the options as passed and try the launch, does CUDA error gracefully where we can post process the error code to make it meaningful to the user? Doing all of these checks in Python feels expensive and will hurt kernel launch latency.

_Originally posted by @kkraus14 in https://github.com/NVIDIA/cuda-python/pull/676#discussion_r2126787942_

PR #687, opened 2025-06-06
Make compute-sanitizer not report API errors as errors

leofang:
## Description

We rely on driver/runtime to report API errors to us for decision making and input checking (https://github.com/NVIDIA/cuda-python/issues/685). If the compute-sanitizer keeps treating API errors as real errors, we end up skipping the majority of the test suite. Before the sanitizer can suppress errors on a per-API basis, we turn `--report-api-errors` off globally.

## Checklist


leofang:
cc @carterbox for vis

kkraus14 (approved):

leofang:
Since the CI was green and the last commit was largely innocent, let me admin-merge and test it in the other PR. Thanks, Keith!

PR #693, opened 2025-06-07
Make a few memory management objects public 

### LINKED ISSUE #596: Expose pinned memory resource as public API
We've meant to do this but I noticed we haven't had any issue to track this need. This is to avoid users from having to use the private API like this
https://github.com/NVIDIA/cuda-python/blob/bd770e14e0fb960b64862743eafc88394fb6f14a/cuda_core/tests/test_event.py#L155-L156

However, unlike the device MR for which we make it accessible through the `Device` instance (`dev.memory_resource`), we probably need to think about the way(s) to expose it. I think starting with a free-standing MR is probably fine.

leofang:
## Description

closes #596.

- `Buffer` and `MemoryResource` are already public (ex: returned by `Device.memory_resource.allocate()`). We just expose them under the `cuda.core` namespace for documentation purposes (#601)
- The names `DeviceMemoryResource` and `LegacyPinnedMemoryResource` follow the cccl-rt/cudax counterparts. 

## Checklist

### DIFF
--- a/cuda_core/cuda/core/experimental/__init__.py
+++ b/cuda_core/cuda/core/experimental/__init__.py
@@ -8,6 +8,7 @@ from cuda.core.experimental._event import Event, EventOptions
 from cuda.core.experimental._launch_config import LaunchConfig
 from cuda.core.experimental._launcher import launch
 from cuda.core.experimental._linker import Linker, LinkerOptions
+from cuda.core.experimental._memory import Buffer, DeviceMemoryResource, LegacyPinnedMemoryResource, MemoryResource
 from cuda.core.experimental._module import ObjectCode
 from cuda.core.experimental._program import Program, ProgramOptions
 from cuda.core.experimental._stream import Stream, StreamOptions
--- a/cuda_core/cuda/core/experimental/_memory.py
+++ b/cuda_core/cuda/core/experimental/_memory.py
@@ -338,3 +338,7 @@ class _SynchronousMemoryResource(MemoryResource):
     @property
     def device_id(self) -> int:
         return self._dev_id
+
+
+DeviceMemoryResource = _DefaultAsyncMempool
+LegacyPinnedMemoryResource = _DefaultPinnedMemorySource
--- a/cuda_core/docs/source/api.rst
+++ b/cuda_core/docs/source/api.rst
@@ -25,6 +25,11 @@ CUDA runtime
    StreamOptions
    LaunchConfig
 
+   Buffer
+   MemoryResource
+   DeviceMemoryResource
+   LegacyPinnedMemoryResource
+
 
 CUDA compilation toolchain
 --------------------------
--- a/cuda_core/docs/source/api_private.rst
+++ b/cuda_core/docs/source/api_private.rst
@@ -15,7 +15,6 @@ CUDA runtime
 .. autosummary::
    :toctree: generated/
 
-   _memory.Buffer
    _stream.Stream
    _event.Event
    _device.DeviceProperties
--- a/cuda_core/tests/test_event.py
+++ b/cuda_core/tests/test_event.py
@@ -10,8 +10,15 @@ import pytest
 from conftest import skipif_need_cuda_headers
 
 import cuda.core.experimental
-from cuda.core.experimental import Device, EventOptions, LaunchConfig, Program, ProgramOptions, launch
-from cuda.core.experimental._memory import _DefaultPinnedMemorySource
+from cuda.core.experimental import (
+    Device,
+    EventOptions,
+    LaunchConfig,
+    LegacyPinnedMemoryResource,
+    Program,
+    ProgramOptions,
+    launch,
+)
 
 
 def test_event_init_disabled():
@@ -143,7 +150,7 @@ __global__ void wait(int* val) {
     mod = prog.compile(target_type="cubin")
     ker = mod.get_kernel("wait")
 
-    mr = _DefaultPinnedMemorySource()
+    mr = LegacyPinnedMemoryResource()
     b = mr.allocate(4)
     arr = np.from_dlpack(b).view(np.int32)
     arr[0] = 0
--- a/cuda_core/tests/test_launcher.py
+++ b/cuda_core/tests/test_launcher.py
@@ -9,8 +9,7 @@ import numpy as np
 import pytest
 from conftest import skipif_need_cuda_headers
 
-from cuda.core.experimental import Device, LaunchConfig, Program, ProgramOptions, launch
-from cuda.core.experimental._memory import _DefaultPinnedMemorySource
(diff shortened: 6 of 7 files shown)


cuda_core/cuda/core/experimental/_device.py
```diff
@@ -1206,7 +1206,7 @@ def create_context(self, options: ContextOptions = None) -> Context:
         raise NotImplementedError("WIP: https://github.com/NVIDIA/cuda-python/issues/189")
 
     @precondition(_check_context_initialized)
-    def create_stream(self, obj=None, options: StreamOptions = None) -> Stream:
+    def create_stream(self, obj: Optional[IsStreamT] = None, options: StreamOptions = None) -> Stream:
```
kkraus14, line 1210: I think we need to update the docstring typing for this as well?

cuda_core/cuda/core/experimental/_launcher.py
```diff
@@ -34,7 +35,7 @@ def _lazy_init():
     _inited = True
 
 
-def launch(stream, config, kernel, *kernel_args):
+def launch(stream: Union[Stream, IsStreamT], config: LaunchConfig, kernel: Kernel, *kernel_args):
```
kkraus14, line 38: It's a bit inconsistent where only this API supports `__cuda_stream__` protocol, but many other APIs only work with explicit `Stream` type objects. We should probably be consistent.

Maybe push this to a follow up PR?

cuda_core/cuda/core/experimental/_stream.py
```diff
@@ -43,6 +43,47 @@ class StreamOptions:
...
+        warnings.simplefilter("once", DeprecationWarning)
+        warnings.warn(
+            "Implementing __cuda_stream__ as an attribute is deprecated; it must be implemented as a method",
+            stacklevel=3,
+            category=DeprecationWarning,
+        )
+
+    try:
+        len_info = len(info)
+    except Exception as e:
```
kkraus14: Is there any way we can more tightly scope the exception we catch here?

cuda_core/docs/source/contribute.rst
```diff
@@ -1,5 +1,7 @@
 .. SPDX-License-Identifier: Apache-2.0
 
+.. _contributer_guide:
```
kkraus14:
```suggestion
.. _contributor_guide:
```

cuda_core/docs/source/interoperability.rst
```diff
@@ -52,7 +52,8 @@ Then such objects can be understood by ``cuda.core`` anywhere a stream-like obje
 is needed.
```
kkraus14: I think we should discuss this. This will introduce non-negligible overhead, similar to DLPack, where we might be better off having a user explicitly wrap their stream objects in a `cuda.core.Stream` and reuse that `Stream` object as needed.

cuda_core/docs/source/release/0.3.0-notes.rst
```diff
@@ -7,33 +7,51 @@
 
 Released on MM DD, 2025
 
+
 Highlights
 ----------
 
-- Starting this release ``cuda.core`` is licensed under Apache 2.0.
+- Starting this release ``cuda.core`` is licensed under Apache 2.0. The biggest implication of this change is that we are open to external contribution now! Please kindly follow the :ref:`Contributor Guide <contributer_guide>` for detailed instructions.
```
kkraus14:
```suggestion
- Starting this release ``cuda.core`` is licensed under Apache 2.0. The biggest implication of this change is that we are open to external contribution now! Please kindly follow the :ref:`Contributor Guide <contributor_guide>` for detailed instructions.
```

cuda_core/cuda/core/experimental/_launcher.py
leofang: We already have public examples for [PyTorch](https://github.com/NVIDIA/cuda-python/blob/1cb8a6fa79b4c6be95634e852d303fe11883ec5d/cuda_core/examples/pytorch_example.py#L43) and [CuPy](https://github.com/NVIDIA/cuda-python/blob/1cb8a6fa79b4c6be95634e852d303fe11883ec5d/cuda_core/examples/simple_multi_gpu_example.py#L72) showcasing the use of this protocol. With this PR we made it slightly faster in favor of our own `Stream` (through try-except). Perhaps we should instead encourage users to use our native objects, and state that using the protocol could add a slight overhead?

cuda_core/cuda/core/experimental/_stream.py
leofang: This was from the old code refactored to here. Narrowed it down to `TypeError` in commit f65b644a.

cuda_core/cuda/core/experimental/_device.py
leofang: fixed in 7f34922

cuda_core/docs/source/interoperability.rst
leofang: Yes right now only `launch()` and `Device.create_stream()` support the protocol, due to the exact concern of adding overheads. Overhead aside, I think essentially we are assessing which interoperability story we want to tell: 
- We give people a convenient way for them to convert their types to ours, and only accept our types in our APIs
- We accept everyone's types in our APIs

Both stories need the protocol so it's not a waste. Just how to use the protocol is different.

kkraus14: Yes, we need the protocol regardless. My point was that we document here that an object that supports `__cuda_stream__` protocol can be used anywhere `cuda.core` expects a stream-like object, which isn't currently true. We should do one of following things here:
1. Update the docs to make it clear that a user is required to wrap their `__cuda_stream__` protocol supporting object in a `cuda.core.Stream` before passing it to other `cuda.core` APIs.
2. Update the `cuda.core` implementation to support `IsStreamT` objects everywhere Streams are an API argument.
3. Update the docs to make it clear which APIs explicitly allow `IsStreamT` objects vs required explicit `Stream` objects.
    * I'm -1 on this because it is confusing and non-intuitive to users
    
Based on the current state of implementation, I think we're the closest to 1 (minus the `launch()` API) where we should probably make the docs reflect that for the time being?

leofang: I'll work on 1 now.

leofang: @kkraus14 could you check out commit bbc1c655? I try to rephrase to hint that we want people to wrap their streams, instead of accepting them as-is. We probably have to defer the change of `launch()` to a later time. Vlad's graph support (nicely) took advantage of `launch()` supporting `__cuda_stream__`. We can look into it during graph phase 2.

leofang:
CI is green now.

kkraus14 (approved):

PR #709, opened 2025-06-13
WIP: Cythonize away some perf hot spots

leofang:
## Description

Less aggressive version of #677. PR description to follow later. Still WIP.

Preliminary data:
- cuda.core main branch
```python
In [5]: %timeit e = dev.create_event()
4.65 μs ± 18.4 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```
```python
In [5]: %timeit s = dev.create_stream()
7.7 μs ± 9.98 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```
- this PR
```python
In [6]: %timeit e = dev.create_event()
1.11 μs ± 6.91 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```
```python
In [4]: %timeit s = dev.create_stream()
4.12 μs ± 14 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```
- cupy
```python
In [8]: %timeit e = cp.cuda.Event(disable_timing=True)
749 ns ± 5.45 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```
```python
In [14]: %timeit s = cp.cuda.Stream(non_blocking=True)
3.8 μs ± 8.54 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

## Checklist

### DIFF
--- a/cuda_core/cuda/core/experimental/_context.pyx
+++ b/cuda_core/cuda/core/experimental/_context.pyx
@@ -16,7 +16,7 @@ class ContextOptions:
 cdef class Context:
 
     cdef:
-        object _handle
+        readonly object _handle
         int _device_id
 
     def __init__(self, *args, **kwargs):
--- a/cuda_core/cuda/core/experimental/_context.pyx
+++ b/cuda_core/cuda/core/experimental/_context.pyx
@@ -13,16 +13,21 @@ class ContextOptions:
     pass  # TODO
 
 
-class Context:
-    __slots__ = ("_handle", "_id")
+cdef class Context:
 
-    def __new__(self, *args, **kwargs):
+    cdef:
+        object _handle
+        int _device_id
+
+    def __init__(self, *args, **kwargs):
         raise RuntimeError("Context objects cannot be instantiated directly. Please use Device or Stream APIs.")
 
     @classmethod
-    def _from_ctx(cls, obj, dev_id):
-        assert_type(obj, driver.CUcontext)
-        ctx = super().__new__(cls)
-        ctx._handle = obj
-        ctx._id = dev_id
+    def _from_ctx(cls, handle: driver.CUcontext, int device_id):
+        cdef Context ctx = Context.__new__(Context)
+        ctx._handle = handle
+        ctx._device_id = device_id
         return ctx
+
+    def __eq__(self, other):
+        return int(self._handle) == int(other._handle)
--- a/cuda_core/cuda/core/experimental/_event.pyx
+++ b/cuda_core/cuda/core/experimental/_event.pyx
@@ -88,19 +88,19 @@ cdef class Event:
         raise RuntimeError("Event objects cannot be instantiated directly. Please use Stream APIs (record).")
 
     @classmethod
-    def _init(cls, device_id: int, ctx_handle: Context, opts=None):
+    def _init(cls, device_id: int, ctx_handle: Context, options: Optional[EventOptions] = None):
         cdef Event self = Event.__new__(Event)
-        cdef EventOptions options = check_or_create_options(EventOptions, opts, "Event options")
+        cdef EventOptions opts = check_or_create_options(EventOptions, options, "Event options")
         flags = 0x0
         self._timing_disabled = False
         self._busy_waited = False
-        if not options.enable_timing:
+        if not opts.enable_timing:
             flags |= driver.CUevent_flags.CU_EVENT_DISABLE_TIMING
             self._timing_disabled = True
-        if options.busy_waited_sync:
+        if opts.busy_waited_sync:
             flags |= driver.CUevent_flags.CU_EVENT_BLOCKING_SYNC
             self._busy_waited = True
-        if options.support_ipc:
+        if opts.support_ipc:
             raise NotImplementedError("WIP: https://github.com/NVIDIA/cuda-python/issues/103")
         err, self._handle = driver.cuEventCreate(flags)
         raise_if_driver_error(err)
--- a/cuda_core/cuda/core/experimental/_event.pyx
+++ b/cuda_core/cuda/core/experimental/_event.pyx
@@ -4,14 +4,12 @@
 
 from __future__ import annotations
 
-import weakref
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Optional
 
 from cuda.core.experimental._context import Context
 from cuda.core.experimental._utils.cuda_utils import (
     CUDAError,
-    check_or_create_options,
     driver,
     handle_return,
 )
@@ -25,7 +23,7 @@ if TYPE_CHECKING:
 
 
 @dataclass
-class EventOptions:
+cdef class EventOptions:
     """Customizable :obj:`~_event.Event` options.
 
     Attributes
@@ -49,7 +47,27 @@ class EventOptions:
     support_ipc: Optional[bool] = False
 
 
-class Event:
+cdef inline EventOptions check_or_create_options(options, str options_description):
+    """
+    Create the specified options dataclass from a dictionary of options or None.
+    """
+    cdef EventOptions opts
+    if options is None:
+        opts = EventOptions()
+    elif isinstance(options, dict):
+        opts = EventOptions(**options)
+    elif not isinstance(options, EventOptions):
+        raise TypeError(
+            f"The {options_description} must be provided as an object "
+            f"of type {EventOptions.__name__} or as a dict with valid {options_description}. "
+            f"The provided object is '{options}'."
+        )
+
+    return opts
+
+
+
+cdef class Event:
     """Represent a record at a specific point of execution within a CUDA stream.
 
     Applications can asynchronously record events at any point in
@@ -77,30 +95,20 @@ class Event:
     and they should instead be created through a :obj:`~_stream.Stream` object.
 
     """
+    cdef:
+        object _handle
+        bint _timing_disabled
+        bint _busy_waited
+        int _device_id
+        object _ctx_handle
 
-    class _MembersNeededForFinalize:
-        __slots__ = ("handle",)
-
-        def __init__(self, event_obj, handle):
-            self.handle = handle
-            weakref.finalize(event_obj, self.close)
-
-        def close(self):
-            if self.handle is not None:
-                handle_return(driver.cuEventDestroy(self.handle))
-                self.handle = None
-
-    def __new__(self, *args, **kwargs):
+    def __init__(self, *args, **kwargs):
         raise RuntimeError("Event objects cannot be instantiated directly. Please use Stream APIs (record).")
 
-    __slots__ = ("__weakref__", "_mnff", "_timing_disabled", "_busy_waited", "_device_id", "_ctx_handle")
-
     @classmethod
-    def _init(cls, device_id: int, ctx_handle: Context, options: Optional[EventOptions] = None):
-        self = super().__new__(cls)
-        self._mnff = Event._MembersNeededForFinalize(self, None)
-
-        options = check_or_create_options(EventOptions, options, "Event options")
+    def _init(cls, device_id: int, ctx_handle: Context, opts=None):
+        cdef Event self = Event.__new__(Event)
+        cdef EventOptions options = check_or_create_options(opts, "Event options")
         flags = 0x0
         self._timing_disabled = False
         self._busy_waited = False
@@ -112,14 +120,22 @@ class Event:
             self._busy_waited = True
         if options.support_ipc:
             raise NotImplementedError("WIP: https://github.com/NVIDIA/cuda-python/issues/103")
-        self._mnff.handle = handle_return(driver.cuEventCreate(flags))
+        _, self._handle = driver.cuEventCreate(flags)
         self._device_id = device_id
         self._ctx_handle = ctx_handle
         return self
 
+    cdef _close(self):
+        if self._handle is not None:
+            _ = driver.cuEventDestroy(self._handle)
+            self._handle = None
+
     def close(self):
         """Destroy the event."""
-        self._mnff.close()
+        self._close()
+
+    def __dealloc__(self):
+        self._close()
 
     def __isub__(self, other):
         return NotImplemented
@@ -129,7 +145,7 @@ class Event:
 
     def __sub__(self, other):
         # return self - other (in milliseconds)
-        err, timing = driver.cuEventElapsedTime(other.handle, self.handle)
+        err, timing = driver.cuEventElapsedTime(other.handle, self._handle)
         try:
             raise_if_driver_error(err)
             return timing
@@ -180,12 +196,12 @@ class Event:
         has been completed.
 
         """
-        handle_return(driver.cuEventSynchronize(self._mnff.handle))
+        handle_return(driver.cuEventSynchronize(self._handle))
 
     @property
     def is_done(self) -> bool:
         """Return True if all captured works have been completed, otherwise False."""
-        (result,) = driver.cuEventQuery(self._mnff.handle)
+        (result,) = driver.cuEventQuery(self._handle)
         if result == driver.CUresult.CUDA_SUCCESS:
             return True
         if result == driver.CUresult.CUDA_ERROR_NOT_READY:
@@ -201,7 +217,7 @@ class Event:
             This handle is a Python object. To get the memory address of the underlying C
             handle, call ``int(Event.handle)``.
         """
-        return self._mnff.handle
+        return self._handle
 
     @property
     def device(self) -> Device:
--- a/cuda_core/cuda/core/experimental/_event.pyx
+++ b/cuda_core/cuda/core/experimental/_event.pyx
@@ -88,7 +88,7 @@ cdef class Event:
         raise RuntimeError("Event objects cannot be instantiated directly. Please use Stream APIs (record).")
 
     @classmethod
-    def _init(cls, device_id: int, ctx_handle: Context, options: Optional[EventOptions] = None):
+    def _init(cls, device_id: int, ctx_handle: Context, options=None):
         cdef Event self = Event.__new__(Event)
         cdef EventOptions opts = check_or_create_options(EventOptions, options, "Event options")
         flags = 0x0
--- a/cuda_core/cuda/core/experimental/_event.pyx
+++ b/cuda_core/cuda/core/experimental/_event.pyx
@@ -102,7 +102,8 @@ cdef class Event:
             self._busy_waited = True
         if options.support_ipc:
             raise NotImplementedError("WIP: https://github.com/NVIDIA/cuda-python/issues/103")
-        _, self._handle = driver.cuEventCreate(flags)
+        err, self._handle = driver.cuEventCreate(flags)
+        raise_if_driver_error(err)
         self._device_id = device_id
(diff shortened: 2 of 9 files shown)


leofang:
I'll need some help in looking into the weird CI failures...

leofang:
PR description updated. Thanks to @oleksandr-pavlyk, the CI failure was identified and fixed (the refactoring introduced a call to `cuStreamGetCtx`, which could not be called during stream capturing with 12.2 driver). This is ready to review.

cuda_core/cuda/core/experimental/_context.pyx
```diff
@@ -13,16 +13,21 @@ class ContextOptions:
     pass  # TODO
 
 
-class Context:
-    __slots__ = ("_handle", "_id")
+cdef class Context:
 
-    def __new__(self, *args, **kwargs):
+    cdef:
+        readonly object _handle
```
kkraus14, line 19: Can we make this a `driver.CUcontext` class instead of an object?

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -77,49 +77,46 @@ class Event:
...
-            weakref.finalize(event_obj, self.close)
-
-        def close(self):
-            if self.handle is not None:
-                handle_return(driver.cuEventDestroy(self.handle))
-                self.handle = None
-
-    def __new__(self, *args, **kwargs):
+    cdef:
+        object _handle
```
kkraus14, line 81: +1 can we make this a `driver.CUevent`?

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -77,49 +77,46 @@ class Event:
...
+        bint _timing_disabled
+        bint _busy_waited
+        int _device_id
+        object _ctx_handle
```
kkraus14, line 85: +1 `driver.CUcontext`?

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -77,49 +77,46 @@ class Event:
...
+    cpdef close(self):
         """Destroy the event."""
-        self._mnff.close()
+        if self._handle is not None:
+            err, = driver.cuEventDestroy(self._handle)
+            raise_if_driver_error(err)
+            self._handle = None
+
+    def __del__(self):
+        self.close()
```
kkraus14, line 119: Do we have any concerns at interpreter shutdown or otherwise with using `__del__` instead of a weakref finalizer? I know we're trying to avoid the overhead of the weakref finalizer.

is `self._handle` guaranteed to either be the value assigned to it or `None` in the case of shutdown?

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -103,35 +105,27 @@ class Stream:
...
-            self.builtin = builtin
-            weakref.finalize(stream_obj, self.close)
+    cdef:
+        object _handle
+        object _owner
+        object _builtin
+        object _nonblocking
+        object _priority
+        object _device_id
+        object _ctx_handle
```
kkraus14, line 115: Should we type these more strongly?

cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
```diff
@@ -52,26 +52,29 @@ def _reduce_3_tuple(t: tuple):
...
-    if desc_err != driver.CUresult.CUDA_SUCCESS:
+    if desc_err != _DRIVER_SUCCESS:
         raise CUDAError(f"{name}")
     desc = desc.decode()
     raise CUDAError(f"{name}: {desc}")
 
 
-def _check_runtime_error(error):
+cpdef inline int _check_runtime_error(error) except?-1:
     if error == runtime.cudaError_t.cudaSuccess:
```
kkraus14: Might be worth pulling this out similar to what you did for `_DRIVER_SUCCESS`?

cuda_core/tests/test_cuda_utils.py
```diff
@@ -44,7 +44,7 @@ def test_check_driver_error():
     num_unexpected = 0
     for error in driver.CUresult:
         if error == driver.CUresult.CUDA_SUCCESS:
-            assert cuda_utils._check_driver_error(error) is None
+            assert cuda_utils._check_driver_error(error) == 0
```
kkraus14, line 47: nitpick, use the driver enum value instead of 0

cuda_core/cuda/core/experimental/_context.pyx
leofang: Same for everything below, no unless we switch to cimport from cuda.bindings and use its Cython layer instead of the Python layer, since Python types can't be used to declare a cdef type. The whole point of this PR is to defer this move to a later time while still getting good perf (see the PR description).

cuda_core/cuda/core/experimental/_stream.pyx
kkraus14: Outside of the driver types, I think we could use `bint` for the flags here?

cuda_core/cuda/core/experimental/_event.pyx
leofang: My understanding is that since Cython 3 `__del__` for cdef classes is now supported and correctly mapped to `tp_finalize`. I haven't heard any issue since it was released in 3.0alpha10, so I assume it is now handled correctly compared to what we saw in the old days with 0.29.x.

Technically we should not use `__dealloc__` here because despite `.close()` is marked as `cpdef` and Cython should just access the C method, the attributes that we're touching are Python objects. These are the only two options that we have.

cuda_core/tests/test_cuda_utils.py
leofang: No, this is Python int, not enum 🙂 The int is coming from our `cpdef` wrapper.

cuda_core/cuda/core/experimental/_stream.pyx
leofang: When I discussed with Sasha, we agreed that since all objects that we're touching here are Python objects, typing them as `bint` or `int` introduces unnecessary boxing/unboxing.

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -4,28 +4,28 @@
...
-)
 
 if TYPE_CHECKING:
     import cuda.bindings
     from cuda.core.experimental._device import Device
 
 
 @dataclass
-class EventOptions:
+cdef class EventOptions:
```
oleksandr-pavlyk, line 28: Note to self: It would be interesting to compare generated C++ code between using `@dataclass` and using `@cython.dataclasses.dataclass`.

Since the latter does not require `import dataclasses`, it might result in leaner code.

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -77,49 +77,46 @@ class Event:
...
             raise NotImplementedError("WIP: https://github.com/NVIDIA/cuda-python/issues/103")
-        self._mnff.handle = handle_return(driver.cuEventCreate(flags))
+        err, self._handle = driver.cuEventCreate(flags)
+        raise_if_driver_error(err)
         self._device_id = device_id
         self._ctx_handle = ctx_handle
         return self
 
-    def close(self):
+    cpdef close(self):
```
oleksandr-pavlyk, line 111: Note to self: check if this method `cdef` calling signature is used in this file.

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -77,49 +77,46 @@ class Event:
...
         """Destroy the event."""
-        self._mnff.close()
+        if self._handle is not None:
+            err, = driver.cuEventDestroy(self._handle)
+            raise_if_driver_error(err)
+            self._handle = None
```
oleksandr-pavlyk: Perhaps swap these lines so `_handle` is unset even if error is encountered?

cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
leofang: Good catch, done in commit cc6339ee

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -103,35 +105,27 @@ class Stream:
...
-        __slots__ = ("handle", "owner", "builtin")
-
-        def __init__(self, stream_obj, handle, owner, builtin):
-            self.handle = handle
-            self.owner = owner
-            self.builtin = builtin
-            weakref.finalize(stream_obj, self.close)
+    cdef:
+        object _handle
+        object _owner
```
oleksandr-pavlyk, line 110: Spacing on this line renders differently from spacing of adjacent lines on GH. Not sure why, but should check it out.

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -140,66 +134,76 @@ def _legacy_default(cls):
...
-        nonblocking = options.nonblocking
-        priority = options.priority
+        cdef StreamOptions opts = check_or_create_options(StreamOptions, options, "Stream options")
+        nonblocking = opts.nonblocking
+        priority = opts.priority
 
         flags = driver.CUstream_flags.CU_STREAM_NON_BLOCKING if nonblocking else driver.CUstream_flags.CU_STREAM_DEFAULT
-
-        high, low = handle_return(runtime.cudaDeviceGetStreamPriorityRange())
+        err, high, low = driver.cuCtxGetStreamPriorityRange()
```
oleksandr-pavlyk, line 171: Note to myself: check spacing

cuda_core/cuda/core/experimental/_event.pyx
leofang: Yeah would be interesting to confirm Cython's claim that both are equivalent
https://cython.readthedocs.io/en/latest/src/userguide/extension_types.html#dataclass-extension-types

cuda_core/cuda/core/experimental/_stream.pyx
oleksandr-pavlyk: checks out

cuda_core/cuda/core/experimental/_stream.pyx
oleksandr-pavlyk: checked.

cuda_core/cuda/core/experimental/_event.pyx
oleksandr-pavlyk: I am going to push the change swapping these two line tomorrow.

cuda_core/cuda/core/experimental/_event.pyx
oleksandr-pavlyk: Tried this. The only minor difference to the generated code is that with `import dataclasses` there is an unused global symbol generated, absent if `from cython.dataclasses cimport dataclasses` is used:

```
(dev-cuda-python) opavlyk@ee09c48-lcedt:~/repos/cuda-python$ grep -r "__pyx_k_dataclasses" _stream_ref.cpp 
static const char __pyx_k_dataclasses[] = "dataclasses";
  {__pyx_k_dataclasses, sizeof(__pyx_k_dataclasses), 0, 1, 1}, /* PyObject cname: __pyx_n_u_dataclasses */
```

```
(dev-cuda-python) opavlyk@ee09c48-lcedt:~/repos/cuda-python$ grep -r "__pyx_k_dataclasses" _stream_change.cpp 
```

Otherwise codes appear as very similar.

cuda_core/cuda/core/experimental/_event.pyx
oleksandr-pavlyk: This method appears to only be used from `__del__` special method. As such, I'd recommend to change `cpdef` to `def`.

cuda_core/tests/test_cuda_utils.py
kkraus14: Woops, mistakenly though this was returning a `CUresult`, but this is a return code from our own function (using return code instead of potentially throwing for performance reasons)

cuda_core/cuda/core/experimental/_event.pyx
leofang: Right, this was marked as `cpdef` on purpose, so that `__del__` can use the C version of it for (slightly) better perf.

oleksandr-pavlyk: I am not sure `__del__` is calling the cdef function since `self` is not typed in the definition, but I will have a quick look into generated `.cpp` tomorrow morning.

leofang: Oh I think I checked that.
```c
/* "cuda/core/experimental/_event.pyx":118
 *             self._handle = None
 * 
 *     def __del__(self):             # <<<<<<<<<<<<<<
 *         self.close()
 * 
 */

/* Python wrapper */
static void __pyx_pw_4cuda_4core_12experimental_6_event_5Event_7__del__(PyObject *__pyx_v_self); /*proto*/
static void __pyx_pw_4cuda_4core_12experimental_6_event_5Event_7__del__(PyObject *__pyx_v_self) {
  CYTHON_UNUSED PyObject *const *__pyx_kwvalues;
  __Pyx_RefNannyDeclarations
  __Pyx_RefNannySetupContext("__del__ (wrapper)", 0);
  __pyx_kwvalues = __Pyx_KwValues_VARARGS(__pyx_args, __pyx_nargs);
  __pyx_pf_4cuda_4core_12experimental_6_event_5Event_6__del__(((struct __pyx_obj_4cuda_4core_12experimental_6_event_Event *)__pyx_v_self));

  /* function exit code */
  __Pyx_RefNannyFinishContext();
}

static void __pyx_pf_4cuda_4core_12experimental_6_event_5Event_6__del__(struct __pyx_obj_4cuda_4core_12experimental_6_event_Event *__pyx_v_self) {
  __Pyx_RefNannyDeclarations
  PyObject *__pyx_t_1 = NULL;
  int __pyx_lineno = 0;
  const char *__pyx_filename = NULL;
  int __pyx_clineno = 0;
  __Pyx_RefNannySetupContext("__del__", 1);

  /* "cuda/core/experimental/_event.pyx":119
 * 
 *     def __del__(self):
 *         self.close()             # <<<<<<<<<<<<<<
 * 
 *     def __isub__(self, other):
 */
  __pyx_t_1 = ((struct __pyx_vtabstruct_4cuda_4core_12experimental_6_event_Event *)__pyx_v_self->__pyx_vtab)->close(__pyx_v_self, 0); if (unlikely(!__pyx_t_1)) __PYX_ERR(1, 119, __pyx_L1_error)
  __Pyx_GOTREF(__pyx_t_1);
```
`__del__` calls the method through a vtable. It should be possible to verify that this is the C version of `close()`:
```c
  __pyx_vtabptr_4cuda_4core_12experimental_6_event_Event = &__pyx_vtable_4cuda_4core_12experimental_6_event_Event;
  __pyx_vtable_4cuda_4core_12experimental_6_event_Event.close = (PyObject *(*)(struct __pyx_obj_4cuda_4core_12experimental_6_event_Event *, int __pyx_skip_dispatch))__pyx_f_4cuda_4core_12experimental_6_event_5Event_close;
```
Another place where `__pyx_f_4cuda_4core_12experimental_6_event_5Event_close` is used; I think this is its Python counterpart:
```c
static PyObject *__pyx_pf_4cuda_4core_12experimental_6_event_5Event_4close(struct __pyx_obj_4cuda_4core_12experimental_6_event_Event *__pyx_v_self) {
  PyObject *__pyx_r = NULL;
  __Pyx_RefNannyDeclarations
  PyObject *__pyx_t_1 = NULL;
  int __pyx_lineno = 0;
  const char *__pyx_filename = NULL;
  int __pyx_clineno = 0;
  __Pyx_RefNannySetupContext("close", 1);
  __Pyx_XDECREF(__pyx_r);
  __pyx_t_1 = __pyx_f_4cuda_4core_12experimental_6_event_5Event_close(__pyx_v_self, 1); if (unlikely(!__pyx_t_1)) __PYX_ERR(1, 111, __pyx_L1_error)
  __Pyx_GOTREF(__pyx_t_1);
  __pyx_r = __pyx_t_1;
  __pyx_t_1 = 0;
  goto __pyx_L0;

  /* function exit code */
  __pyx_L1_error:;
  __Pyx_XDECREF(__pyx_t_1);
  __Pyx_AddTraceback("cuda.core.experimental._event.Event.close", __pyx_clineno, __pyx_lineno, __pyx_filename);
  __pyx_r = NULL;
  __pyx_L0:;
  __Pyx_XGIVEREF(__pyx_r);
  __Pyx_RefNannyFinishContext();
  return __pyx_r;
}
```

oleksandr-pavlyk: Live and learn! Thanks @leofang

cuda_core/cuda/core/experimental/_event.pyx
oleksandr-pavlyk: pushed

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -77,49 +77,46 @@ class Event:
...
 
-    __slots__ = ("__weakref__", "_mnff", "_timing_disabled", "_busy_waited", "_device_id", "_ctx_handle")
-
     @classmethod
-    def _init(cls, device_id: int, ctx_handle: Context, options: Optional[EventOptions] = None):
-        self = super().__new__(cls)
-        self._mnff = Event._MembersNeededForFinalize(self, None)
-
-        options = check_or_create_options(EventOptions, options, "Event options")
+    def _init(cls, device_id: int, ctx_handle: Context, options=None):
```
oleksandr-pavlyk, line 91: Since `Event` contains native class members, perhaps adding `__cinit__` to initialize them is appropriate. Something like

```
    def __cinit__(self):
        self._timing_disabled = False
        self._busy_waited = False
        self._device_id = -1
```

I also think it would be safe to set `object` class members to `None`.

This would ensure that `Event.__new__(Event)` would return an initialized struct.

leofang: I think Cython sets everything to None for us, but it'd be good to verify this indeed
> Cython additionally takes responsibility of setting all object attributes to None, 

https://cython.readthedocs.io/en/latest/src/userguide/special_methods.html#initialisation-methods-cinit-and-init

leofang:
CI is green

cuda_core/cuda/core/experimental/_event.pyx
oleksandr-pavlyk: Ok, let's leave `object` members out. Should I push adding `Event.__cinit__` ?

kkraus14 (approved):

cuda_core/cuda/core/experimental/_event.pyx
leofang: I think the same section says all members are zero/null initialized?

oleksandr-pavlyk: Yes, but is it appropriate to zero initialize `_device_id`? Perhaps it does not matter much.

Issue #564, opened 2025-04-22
[FEA]: Expose `get_cuda_native_handle` to Python


leofang:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

cuda.bindings

### Is your feature request related to a problem? Please describe.

In #463 we added experimental C++ APIs `get_cuda_native_handle` for C/C++/Cython to access the C handles wrapped by cuda.bindings types/objects. This issue is to track the need of exposing such APIs to Python to replace the current usage of `__int__()`, something along this line:
```python
from cuda.bindings.utils import get_cuda_native_handle
...

err, s = driver.cuStreamCreate()
s_ptr = get_cuda_native_handle(s)
assert isinstance(s_ptr, int)
```
We may also consider moving the C++ counterparts from `cuda.core` to `cuda.bindings`.

### Describe the solution you'd like

n/a

### Describe alternatives you've considered

n/a

### Additional context

n/a

PR #717, opened 2025-06-18
Ensure correct handling of buffers allocated with `LegacyPinnedMemoryResource.allocate` as kernel parameters

### LINKED ISSUE #715: [BUG]: Cannot pass pinned memory buffer to kernels
### Is this a duplicate?

### Type of Bug

Runtime Error

### Component

cuda.core

### Describe the bug

Creating a pinned memory buffer via `LegacyPinnedMemoryResource` and passing it to a kernel fails with:

```python
>   self.data_addresses[i] = <void*><intptr_t>(arg.handle.getPtr())
E   AttributeError: 'int' object has no attribute 'getPtr'

cuda/core/experimental/_kernel_arg_handler.pyx:215: AttributeError
```

### How to Reproduce

It's not minimal, but here's a failing script:

```python
import cupy as cp
import numpy as np
from cuda.core.experimental import (
    Device, LaunchConfig, Program, ProgramOptions, launch,
    DeviceMemoryResource, LegacyPinnedMemoryResource, Buffer
)
from cuda.core.experimental._memory import MemoryResource
from cuda.core.experimental._utils.cuda_utils import handle_return

# Kernel for memory operations
code = """
extern "C"
__global__ void memory_ops(float* device_data, 
                          float* pinned_data,
                          size_t N) {
    const unsigned int tid = threadIdx.x + blockIdx.x * blockDim.x;
    if (tid < N) {
        // Access device memory
        device_data[tid] = device_data[tid] + 1.0f;
        
        // Access pinned memory (zero-copy from GPU)
        pinned_data[tid] = pinned_data[tid] * 3.0f;
    }
}
"""

dev = Device()
dev.set_current()
stream = dev.create_stream()

# Compile kernel
arch = "".join(f"{i}" for i in dev.compute_capability)
program_options = ProgramOptions(std="c++17", arch=f"sm_{arch}")
prog = Program(code, code_type="c++", options=program_options)
mod = prog.compile("cubin")
kernel = mod.get_kernel("memory_ops")

# Create different memory resources
device_mr = DeviceMemoryResource(dev.device_id)
pinned_mr = LegacyPinnedMemoryResource()

# Allocate different types of memory
size = 1024
dtype = cp.float32
element_size = dtype().itemsize
total_size = size * element_size

# 1. Device Memory (GPU-only)
device_buffer = device_mr.allocate(total_size, stream=stream)
device_array = cp.ndarray(
    size, dtype=dtype,
    memptr=cp.cuda.MemoryPointer(
        cp.cuda.UnownedMemory(int(device_buffer.handle), device_buffer.size, device_buffer), 0
    )
)

# 2. Pinned Memory (CPU memory, GPU accessible)
pinned_buffer = pinned_mr.allocate(total_size, stream=stream)
pinned_array = cp.ndarray(
    size, dtype=dtype,
    memptr=cp.cuda.MemoryPointer(
        cp.cuda.UnownedMemory(int(pinned_buffer.handle), pinned_buffer.size, pinned_buffer), 0
    )
)

# Initialize data
rng = cp.random.default_rng()
device_array[:] = rng.random(size, dtype=dtype)
pinned_array[:] = rng.random(size, dtype=dtype)

# Store original values for verification
device_original = device_array.copy()
pinned_original = pinned_array.copy()

# Sync before kernel launch
dev.sync()

# Launch kernel
block = 256
grid = (size + block - 1) // block
config = LaunchConfig(grid=grid, block=block)

launch(stream, config, kernel, 
       device_buffer, pinned_buffer, cp.uint64(size))
stream.sync()

# Verify kernel operations
assert cp.allclose(device_array, device_original + 1.0), "Device memory operation failed"
assert cp.allclose(pinned_array, pinned_original * 3.0), "Pinned memory operation failed"

# Demonstrate buffer copying operations
print("Memory buffer properties:")
print(f"Device buffer - Device accessible: {device_buffer.is_device_accessible}")
print(f"Pinned buffer - Device accessible: {pinned_buffer.is_device_accessible}")

# Assert memory properties
assert device_buffer.is_device_accessible, "Device buffer should be device accessible"
assert not device_buffer.is_host_accessible, "Device buffer should not be host accessible"
assert pinned_buffer.is_device_accessible, "Pinned buffer should be device accessible"
assert pinned_buffer.is_host_accessible, "Pinned buffer should be host accessible"

# Copy data between different memory types
print("\nCopying data between memory types...")

# Copy from device to pinned memory
device_buffer.copy_to(pinned_buffer, stream=stream)
stream.sync()

# Verify the copy operation
assert cp.allclose(pinned_array, device_array), "Device to pinned copy failed"

# Create a new device buffer and copy from pinned
new_device_buffer = device_mr.allocate(total_size, stream=stream)
new_device_array = cp.ndarray(
    size, dtype=dtype,
    memptr=cp.cuda.MemoryPointer(
        cp.cuda.UnownedMemory(int(new_device_buffer.handle), new_device_buffer.size, new_device_buffer), 0
    )
)

pinned_buffer.copy_to(new_device_buffer, stream=stream)
stream.sync()

# Verify the copy operation
assert cp.allclose(new_device_array, pinned_array), "Pinned to device copy failed"

# Demonstrate DLPack integration
print("\nDLPack device information:")
print(f"Device buffer DLPack device: {device_buffer.__dlpack_device__()}")
print(f"Pinned buffer DLPack device: {pinned_buffer.__dlpack_device__()}")

# Assert DLPack device types
from cuda.core.experimental._memory import DLDeviceType

device_dlpack = device_buffer.__dlpack_device__()
pinned_dlpack = pinned_buffer.__dlpack_device__()

assert device_dlpack[0] == DLDeviceType.kDLCUDA, "Device buffer should have CUDA device type"
assert pinned_dlpack[0] == DLDeviceType.kDLCUDAHost, "Pinned buffer should have CUDA host device type"

# Test buffer size properties
assert device_buffer.size == total_size, f"Device buffer size mismatch: expected {total_size}, got {device_buffer.size}"
assert pinned_buffer.size == total_size, f"Pinned buffer size mismatch: expected {total_size}, got {pinned_buffer.size}"
assert new_device_buffer.size == total_size, f"New device buffer size mismatch: expected {total_size}, got {new_device_buffer.size}"

# Test memory resource properties
assert device_buffer.memory_resource == device_mr, "Device buffer should use device memory resource"
assert pinned_buffer.memory_resource == pinned_mr, "Pinned buffer should use pinned memory resource"
assert new_device_buffer.memory_resource == device_mr, "New device buffer should use device memory resource"

# Clean up
device_buffer.close(stream)
pinned_buffer.close(stream)
new_device_buffer.close(stream)
stream.close()

# Verify buffers are properly closed
assert device_buffer.handle == 0, "Device buffer should be closed"
assert pinned_buffer.handle == 0, "Pinned buffer should be closed"
assert new_device_buffer.handle == 0, "New device buffer should be closed"

print("Memory management example completed!")

```

### Expected behavior

The script above should run successfully.

### Operating System

_No response_

### nvidia-smi output

_No response_

shwina:
## Description

Buffers allocated using the `LegacyPinnedMemoryResource` have , but the kernel param handler doesn't know how to handle these buffers. This PR fixes that.

Closes #715 

## Checklist

### DIFF
--- a/cuda_core/cuda/core/experimental/_kernel_arg_handler.pyx
+++ b/cuda_core/cuda/core/experimental/_kernel_arg_handler.pyx
@@ -212,7 +212,13 @@ cdef class ParamHolder:
         for i, arg in enumerate(kernel_args):
             if isinstance(arg, Buffer):
                 # we need the address of where the actual buffer address is stored
-                self.data_addresses[i] = <void*><intptr_t>(arg.handle.getPtr())
+                if isinstance(arg.handle, int):
+                    # see note below on handling int arguments
+                    prepare_arg[intptr_t](self.data, self.data_addresses, arg.handle, i)
+                    continue              
+                else:
+                    # it's a CUdeviceptr:
+                    self.data_addresses[i] = <void*><intptr_t>(arg.handle.getPtr())
                 continue
             elif isinstance(arg, int):
                 # Here's the dilemma: We want to have a fast path to pass in Python
--- a/cuda_core/examples/memory_ops.py
+++ b/cuda_core/examples/memory_ops.py
@@ -1,0 +1,163 @@
+import cupy as cp
+import numpy as np
+from cuda.core.experimental import (
+    Device, LaunchConfig, Program, ProgramOptions, launch,
+    DeviceMemoryResource, LegacyPinnedMemoryResource, Buffer
+)
+from cuda.core.experimental._memory import MemoryResource
+from cuda.core.experimental._utils.cuda_utils import handle_return
+from cuda.bindings import driver
+
+# Kernel for memory operations
+code = """
+extern "C"
+__global__ void memory_ops(float* device_data, 
+                          float* pinned_data,
+                          size_t N) {
+    const unsigned int tid = threadIdx.x + blockIdx.x * blockDim.x;
+    if (tid < N) {
+        // Access device memory
+        device_data[tid] = device_data[tid] + 1.0f;
+        
+        // Access pinned memory (zero-copy from GPU)
+        pinned_data[tid] = pinned_data[tid] * 3.0f;
+    }
+}
+"""
+
+dev = Device()
+dev.set_current()
+stream = dev.create_stream()
+
+# Compile kernel
+arch = "".join(f"{i}" for i in dev.compute_capability)
+program_options = ProgramOptions(std="c++17", arch=f"sm_{arch}")
+prog = Program(code, code_type="c++", options=program_options)
+mod = prog.compile("cubin")
+kernel = mod.get_kernel("memory_ops")
+
+# Create different memory resources
+device_mr = DeviceMemoryResource(dev.device_id)
+pinned_mr = LegacyPinnedMemoryResource()
+
+# Allocate different types of memory
+size = 1024
+dtype = cp.float32
+element_size = dtype().itemsize
+total_size = size * element_size
+
+# 1. Device Memory (GPU-only)
+device_buffer = device_mr.allocate(total_size, stream=stream)
+device_array = cp.ndarray(
+    size, dtype=dtype,
+    memptr=cp.cuda.MemoryPointer(
+        cp.cuda.UnownedMemory(int(device_buffer.handle), device_buffer.size, device_buffer), 0
+    )
+)
+
+# 2. Pinned Memory (CPU memory, GPU accessible)
+pinned_buffer = pinned_mr.allocate(total_size, stream=stream)
+pinned_array = cp.ndarray(
+    size, dtype=dtype,
+    memptr=cp.cuda.MemoryPointer(
+        cp.cuda.UnownedMemory(int(pinned_buffer.handle), pinned_buffer.size, pinned_buffer), 0
+    )
+)
+
+# Initialize data
+rng = cp.random.default_rng()
+device_array[:] = rng.random(size, dtype=dtype)
+pinned_array[:] = rng.random(size, dtype=dtype)
+
+# Store original values for verification
+device_original = device_array.copy()
+pinned_original = pinned_array.copy()
+
+# Sync before kernel launch
+dev.sync()
+
+# Launch kernel
+block = 256
+grid = (size + block - 1) // block
+config = LaunchConfig(grid=grid, block=block)
+
+launch(stream, config, kernel, 
+       device_buffer, pinned_buffer, cp.uint64(size))
+stream.sync()
+
+# Verify kernel operations
+assert cp.allclose(device_array, device_original + 1.0), "Device memory operation failed"
+assert cp.allclose(pinned_array, pinned_original * 3.0), "Pinned memory operation failed"
+
+# Demonstrate buffer copying operations
+print("Memory buffer properties:")
+print(f"Device buffer - Device accessible: {device_buffer.is_device_accessible}")
+print(f"Pinned buffer - Device accessible: {pinned_buffer.is_device_accessible}")
+
+# Assert memory properties
+assert device_buffer.is_device_accessible, "Device buffer should be device accessible"
+assert not device_buffer.is_host_accessible, "Device buffer should not be host accessible"
+assert pinned_buffer.is_device_accessible, "Pinned buffer should be device accessible"
+assert pinned_buffer.is_host_accessible, "Pinned buffer should be host accessible"
+
+# Copy data between different memory types
+print("\nCopying data between memory types...")
+
+# Copy from device to pinned memory
+device_buffer.copy_to(pinned_buffer, stream=stream)
+stream.sync()
+
+# Verify the copy operation
+assert cp.allclose(pinned_array, device_array), "Device to pinned copy failed"
+
+# Create a new device buffer and copy from pinned
+new_device_buffer = device_mr.allocate(total_size, stream=stream)
+new_device_array = cp.ndarray(
+    size, dtype=dtype,
+    memptr=cp.cuda.MemoryPointer(
+        cp.cuda.UnownedMemory(int(new_device_buffer.handle), new_device_buffer.size, new_device_buffer), 0
+    )
+)
+
+pinned_buffer.copy_to(new_device_buffer, stream=stream)
+stream.sync()
+
+# Verify the copy operation
+assert cp.allclose(new_device_array, pinned_array), "Pinned to device copy failed"
+
+# Demonstrate DLPack integration
+print("\nDLPack device information:")
+print(f"Device buffer DLPack device: {device_buffer.__dlpack_device__()}")
+print(f"Pinned buffer DLPack device: {pinned_buffer.__dlpack_device__()}")
+
+# Assert DLPack device types
+from cuda.core.experimental._memory import DLDeviceType
+
+device_dlpack = device_buffer.__dlpack_device__()
+pinned_dlpack = pinned_buffer.__dlpack_device__()
+
+assert device_dlpack[0] == DLDeviceType.kDLCUDA, "Device buffer should have CUDA device type"
+assert pinned_dlpack[0] == DLDeviceType.kDLCUDAHost, "Pinned buffer should have CUDA host device type"
+
+# Test buffer size properties
+assert device_buffer.size == total_size, f"Device buffer size mismatch: expected {total_size}, got {device_buffer.size}"
+assert pinned_buffer.size == total_size, f"Pinned buffer size mismatch: expected {total_size}, got {pinned_buffer.size}"
+assert new_device_buffer.size == total_size, f"New device buffer size mismatch: expected {total_size}, got {new_device_buffer.size}"
+
+# Test memory resource properties
+assert device_buffer.memory_resource == device_mr, "Device buffer should use device memory resource"
+assert pinned_buffer.memory_resource == pinned_mr, "Pinned buffer should use pinned memory resource"
+assert new_device_buffer.memory_resource == device_mr, "New device buffer should use device memory resource"
+
+# Clean up
+device_buffer.close(stream)
+pinned_buffer.close(stream)
+new_device_buffer.close(stream)
+stream.close()
+
+# Verify buffers are properly closed
+assert device_buffer.handle == 0, "Device buffer should be closed"
+assert pinned_buffer.handle == 0, "Pinned buffer should be closed"
+assert new_device_buffer.handle == 0, "New device buffer should be closed"
+
+print("Memory management example completed!")
--- a/cuda_core/examples/memory_ops.py
+++ b/cuda_core/examples/memory_ops.py
@@ -14,7 +14,6 @@ from cuda.core.experimental import (
     ProgramOptions,
     launch,
 )
-from cuda.core.experimental._dlpack import DLDeviceType
 
 # Kernel for memory operations
 code = """
@@ -36,6 +35,8 @@ __global__ void memory_ops(float* device_data,
 dev = Device()
 dev.set_current()
 stream = dev.create_stream()
+# tell CuPy to use our stream as the current stream:
+cp.cuda.ExternalStream(int(stream.handle)).use()
 
 # Compile kernel
 arch = "".join(f"{i}" for i in dev.compute_capability)
@@ -106,23 +107,12 @@ stream.sync()
 # Verify the copy operation
 assert cp.allclose(new_device_array, pinned_array), "Pinned to device copy failed"
 
-# Demonstrate DLPack integration
-print("\nDLPack device information:")
-print(f"Device buffer DLPack device: {device_buffer.__dlpack_device__()}")
-print(f"Pinned buffer DLPack device: {pinned_buffer.__dlpack_device__()}")
-
-# Assert DLPack device types
-device_dlpack = device_buffer.__dlpack_device__()
-pinned_dlpack = pinned_buffer.__dlpack_device__()
-
-assert device_dlpack[0] == DLDeviceType.kDLCUDA, "Device buffer should have CUDA device type"
-assert pinned_dlpack[0] == DLDeviceType.kDLCUDAHost, "Pinned buffer should have CUDA host device type"
-
 # Clean up
 device_buffer.close(stream)
 pinned_buffer.close(stream)
 new_device_buffer.close(stream)
 stream.close()
+cp.cuda.Stream.null.use()  # reset CuPy's current stream to the null stream
 
 # Verify buffers are properly closed
 assert device_buffer.handle == 0, "Device buffer should be closed"
--- a/cuda_core/examples/memory_ops.py
+++ b/cuda_core/examples/memory_ops.py
@@ -72,7 +72,7 @@ device_original = device_array.copy()
 pinned_original = pinned_array.copy()
 
 # Sync before kernel launch
-dev.sync()
+stream.sync()
 
 # Launch kernel
 block = 256
--- a/cuda_core/examples/memory_ops.py
+++ b/cuda_core/examples/memory_ops.py
@@ -2,6 +2,18 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+# ################################################################################
+#
+# This demo illustrates:
+#
+#   1. How to use different memory resources to allocate and manage memory
+#   2. How to copy data between different memory types
+#   3. How to use DLPack to interoperate with other libraries
+#
+# ################################################################################
+
+import sys
+
 import cupy as cp
 import numpy as np
 
@@ -14,6 +26,10 @@ from cuda.core.experimental import (
     ProgramOptions,
     launch,
 )
+
+if np.__version__ < "2.1.0":
+    print("This example requires NumPy 2.1.0 or later", file=sys.stderr)
+    sys.exit(0)
 
 # Kernel for memory operations
 code = """
--- a/cuda_core/examples/memory_ops.py
+++ b/cuda_core/examples/memory_ops.py
@@ -19,7 +19,6 @@ import numpy as np
 
 from cuda.core.experimental import (
     Device,
-    DeviceMemoryResource,
     LaunchConfig,
     LegacyPinnedMemoryResource,
     Program,
@@ -62,7 +61,7 @@ mod = prog.compile("cubin")
 kernel = mod.get_kernel("memory_ops")
 
 # Create different memory resources
-device_mr = DeviceMemoryResource(dev.device_id)
+device_mr = dev.memory_resource
 pinned_mr = LegacyPinnedMemoryResource()
 
 # Allocate different types of memory
--- a/cuda_core/examples/memory_ops.py
+++ b/cuda_core/examples/memory_ops.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import cupy as cp
+import numpy as np
 
 from cuda.core.experimental import (
     Device,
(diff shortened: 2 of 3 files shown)


cuda_core/cuda/core/experimental/_kernel_arg_handler.pyx
```diff
@@ -212,7 +212,13 @@ cdef class ParamHolder:
         for i, arg in enumerate(kernel_args):
             if isinstance(arg, Buffer):
                 # we need the address of where the actual buffer address is stored
-                self.data_addresses[i] = <void*><intptr_t>(arg.handle.getPtr())
+                if isinstance(arg.handle, int):
```
shwina, line 215: Can we stomach the cost of an `isinstance` check here? 

* One alternative is to use a `try..except`, where entering the `try` block is cheap, but entering the `except` block is expensive.

* Another alternative, which will eliminate the need to make any changes to the kernel arg handling logic here:
  - introduce a new type `HostPtr` which wraps an integer representing a pointer, and exposes a `getPtr()` method to get it.
  -  Expand the return type of [`Buffer.handle`](https://github.com/NVIDIA/cuda-python/blob/24fde1755ad1efa6e7d73b2236dad9681fa0b727/cuda_core/cuda/core/experimental/_memory.py#L78) to `DevicePtrT | HostPtr`
  - Change `LegacyPinnedMemoryResource` to return a buffer whose handle is a `HostPtr`.

leofang: I think `isinstance` in Cython is cheap and what you have here is good. I don't want to introduce more types than needed, partly because we want MR providers to focus on the MR properties (`is_host_accessible` etc), which is nicer for programmatic checks. I actually think that `Buffer.handle` should be of `Any` type so as to not  get in the way of the MR providers. From both CUDA and  cccl-rt  perspectives they should be all `void*`. We don't want to encode the memory space information as part of the type.

cuda_core/examples/memory_ops.py
leofang, line 1: Would you mind moving part of the code here to tests? I feel some of the assert checks are too detailed. See below.

cuda_core/examples/memory_ops.py
```diff
@@ -0,0 +1,172 @@
...
+
+# 1. Device Memory (GPU-only)
+device_buffer = device_mr.allocate(total_size, stream=stream)
+device_array = cp.ndarray(
+    size,
+    dtype=dtype,
+    memptr=cp.cuda.MemoryPointer(
+        cp.cuda.UnownedMemory(int(device_buffer.handle), device_buffer.size, device_buffer), 0
+    ),
+)
```
leofang: By design, this should just be
```suggestion
device_array = cp.from_dlpack(device_buffer).view(dtype=dtype)
```

cuda_core/examples/memory_ops.py
```diff
@@ -0,0 +1,172 @@
...
+
+# 2. Pinned Memory (CPU memory, GPU accessible)
+pinned_buffer = pinned_mr.allocate(total_size, stream=stream)
+pinned_array = cp.ndarray(
+    size,
+    dtype=dtype,
+    memptr=cp.cuda.MemoryPointer(
+        cp.cuda.UnownedMemory(int(pinned_buffer.handle), pinned_buffer.size, pinned_buffer), 0
+    ),
+)
```
leofang: Right now, `cupy.from_dlpack` does not support pinned memory yet, I left a TODO there:
https://github.com/cupy/cupy/blob/3a2c950d64ee707096bc7ca1bf0b953a08206384/cupy/_core/dlpack.pyx#L583
so depends on how complex we want to present this example we could choose:
1. using `cupy.ndarray` to wrap `pinned_buffer`: In this case, what we have here is fine
2. using `numpy.ndarray` to wrap `pinned_buffer`: In this case, we can just do the same thing as above
```suggestion
pinned_array = np.from_dlpack(pinned_buffer).view(dtype=dtype)
```
which is a nice pattern that we already use in some tests.

leofang (changes requested):
Thanks for catching the bug, Ashwin! I left some comments. 

It seems to be a miss (due to the way we implicitly test `LegacyPinnedMemoryResource` through NumPy today) that we do not test passing `Buffer` from various MRs to `launch()`. Could you add some tests to `test_launcher.py`? In a comment below, I suggest we could move part of what you have there.

cuda_core/cuda/core/experimental/_kernel_arg_handler.pyx
shwina: > I actually think that Buffer.handle should be of Any type so as to not get in the way of the MR providers.

If we did type it as `Any`, how would `_kernel_arg_handler` know how to grab the pointer from underneath the `Buffer`?

leofang: Well Python does not care about type annotations, right? 🙂

shwina: My concern wasn't so much about the type annotation, but more that the kernel handler won't know what to do with a `Buffer` whose `.handle` is any arbitrary type.

Prior to this PR it could _only_ handle the case when `.handle` is a `CUdeviceptr`, or something that has a `.getPtr()` method. 

https://github.com/NVIDIA/cuda-python/blob/24fde1755ad1efa6e7d73b2236dad9681fa0b727/cuda_core/cuda/core/experimental/_kernel_arg_handler.pyx#L213-L215

This PR adds the ability to handle `int`.

_Technically_, `.handle` is also allowed to be `None`: https://github.com/NVIDIA/cuda-python/blob/24fde1755ad1efa6e7d73b2236dad9681fa0b727/cuda_core/cuda/core/experimental/_memory.py#L22

leofang: Ahh, I see, you meant the mini dispatcher here needs to enumerate all possible types. 

Let me think about it. What you have is good and a generic treatment can follow later.

Most likely with #564 we could rewrite the dispatcher that looks like this
```cython
if isinstance(arg, Buffer):
    prepare_arg[intptr_t](self.data, self.data_addresses, get_cuda_native_handle(arg.handle), i)
```
On the MR provider side, we just need them to implement a protocol
```python 
class IsHandleT(Protocol):
    def __int__(self) -> int: ...
```
if they are not using generic cuda.bindings or Python types. (FWIW we already have [`IsStreamT`](https://nvidia.github.io/cuda-python/cuda-core/latest/generated/cuda.core.experimental._stream.IsStreamT.html#cuda.core.experimental._stream.IsStreamT.__cuda_stream__).) So maybe eventually `Buffer.handle` can be typed as
```python
DevicePointerT = Optional[Union[IsHandleT, int]] 
```

cuda_core/examples/memory_ops.py
shwina: Nice - done

cuda_core/examples/memory_ops.py
shwina: Used `np.from_dlpack`

cuda_core/examples/memory_ops.py
shwina: I added a test to `test_launcher.py` - could you please take a look and lmk if that's what you had in mind?

cuda_core/cuda/core/experimental/_kernel_arg_handler.pyx
shwina: It seems like a reasonable approach and I agree it would simplify the handling here. A couple of comments:

* Perhaps we should rename `DevicePointerT` to just `PointerT`? In the case of pinned memory for instance, it doesn't actually represent a device pointer AFAIU.
* If we use the protocol as written, then `Union[IsHandleT, int]` is equivalent to just `IsHandleT` (`int` type implements `__int__`). The protocol would also allow types like `float` or `bool`.  
  - I feel like this discussion has been had before, but it might be worth considering a protocol with a `__cuda_handle__()` method or something, rather than `__int__()`

cuda_core/examples/memory_ops.py
```diff
@@ -0,0 +1,132 @@
...
+
+        // Access pinned memory (zero-copy from GPU)
+        pinned_data[tid] = pinned_data[tid] * 3.0f;
+    }
+}
+"""
+
+dev = Device()
+dev.set_current()
+stream = dev.create_stream()
```
leofang, line 52: Technically we need this to order every work on the same stream:
```suggestion
stream = dev.create_stream()
# set our stream to current as far as CuPy is concerned 
cp.cuda.ExternalStream(int(stream.handle)).use()
```

cuda_core/examples/memory_ops.py
```diff
@@ -0,0 +1,132 @@
...
+rng = cp.random.default_rng()
+device_array[:] = rng.random(size, dtype=dtype)
+pinned_array[:] = rng.random(size, dtype=dtype).get()
+
+# Store original values for verification
+device_original = device_array.copy()
+pinned_original = pinned_array.copy()
+
+# Sync before kernel launch
+stream.sync()
```
leofang, line 91: ditto, so that the sync here has the correct semantics. We could do something more complicated but the one-liner fix is the easiest for a code sample like this.

cuda_core/examples/memory_ops.py
```diff
@@ -0,0 +1,132 @@
...
+print("\nDLPack device information:")
+print(f"Device buffer DLPack device: {device_buffer.__dlpack_device__()}")
+print(f"Pinned buffer DLPack device: {pinned_buffer.__dlpack_device__()}")
+
+# Assert DLPack device types
+device_dlpack = device_buffer.__dlpack_device__()
+pinned_dlpack = pinned_buffer.__dlpack_device__()
+
+assert device_dlpack[0] == DLDeviceType.kDLCUDA, "Device buffer should have CUDA device type"
+assert pinned_dlpack[0] == DLDeviceType.kDLCUDAHost, "Pinned buffer should have CUDA host device type"
```
leofang: I am thinking this should be moved to tests, for two reasons:
1. To show case the device type, we need to import `DLDeviceType` from a private module which is not ideal
2. End users shouldn't need to know there exists `__dlpack__` and `__dlpack_device__`, they just need to know `cuda.core.Buffer` supports DLPack and so calling `from_dlpack` on it makes sense. Only library developers need to learn the DLPack protocol.

leofang (changes requested):
Thanks, Ashwin! The added test LGTM. Left a few comments mainly for the code sample correctness.

cuda_core/examples/memory_ops.py
```diff
@@ -0,0 +1,132 @@
...
+
+# Clean up
+device_buffer.close(stream)
+pinned_buffer.close(stream)
+new_device_buffer.close(stream)
+stream.close()
```
leofang, line 129:
```suggestion
stream.close()
cp.cuda.Stream.null.use()  # restore to CuPy's default stream
```

shwina: Done

cuda_core/examples/memory_ops.py
shwina: I removed these assertions. I believe these things are tested in `test_memory.py` so perhaps there's no need to add more tests. Let me know if you think otherwise

cuda_core/examples/memory_ops.py
shwina: Ah, right - good catch.

cuda_core/examples/memory_ops.py
shwina: Agree - I went with your suggestion above.

cuda_core/examples/memory_ops.py
leofang: Oh yeah, I totally forgot about it, thanks!

leofang:
I think the `memory_ops.py` sample have 1+ bugs that cause later tests to segfault (pytest runs the sample tests first).

We should early-return in the sample if NumPy is <2.1.0, because `np.from_dlpack` had a bug. In one of the samples I did this:
https://github.com/NVIDIA/cuda-python/blob/24fde1755ad1efa6e7d73b2236dad9681fa0b727/cuda_core/examples/thread_block_cluster.py#L12-L14

shwina:
I'm also seeing this failure mode in addition to the segfault:

```python
>           assert cp.allclose(array, original * 3.0), f"{memory_resource_class.__name__} operation failed"
E           AssertionError: DeviceMemoryResource operation failed
E           assert array(False)
E            +  where array(False) = <function allclose at 0xf5ce6f977740>(array([0.30819342, 0.78738075, 0.71034026, ..., 0.3719493 , 0.7937206 ,\n       0.90285766], shape=(1024,), dtype=float32), (array([0.30819342, 0.78738075, 0.71034026, ..., 0.3719493 , 0.7937206 ,\n       0.90285766], shape=(1024,), dtype=float32) * 3.0))
E            +    where <function allclose at 0xf5ce6f977740> = <module 'cupy' from '/opt/hostedtoolcache/Python/3.13.5/arm64/lib/python3.13/site-packages/cupy/__init__.py'>.allclose

tests/test_launcher.py:296: AssertionError
```

shwina:
/ok to test https://github.com/NVIDIA/cuda-python/pull/717/commits/0b2e2077aa3b73bf940f9d7b381544de5346b34b

shwina:
OK, now I'm seeing:

```
>           array[:] = rng.random(size, dtype=dtype)
E           ValueError: assignment destination is read-only
```

This one I _can_ reproduce with older numpy.

cuda_core/examples/memory_ops.py
```diff
@@ -0,0 +1,138 @@
...
+
+# Compile kernel
+arch = "".join(f"{i}" for i in dev.compute_capability)
+program_options = ProgramOptions(std="c++17", arch=f"sm_{arch}")
+prog = Program(code, code_type="c++", options=program_options)
+mod = prog.compile("cubin")
+kernel = mod.get_kernel("memory_ops")
+
+# Create different memory resources
+device_mr = DeviceMemoryResource(dev.device_id)
```
leofang: I will test this tomorrow but I suspect this is what happened on Windows:
```suggestion
try:
    device_mr = DeviceMemoryResource(dev.device_id)
except Exception as e:
    print("This example requires a driver supporting stream-ordered memory allocator", file=sys.stderr)
    cp.cuda.Stream.null.use()  # reset
    sys.exit(0)
```

leofang: One alternative to run the same code on Windows is to switch to the per-device MR from using an independent MR, with which we hide the platform difference:
```suggestion
device_mr = dev.memory_resource
```

shwina: Hmm, this feels like maybe a potential improvement that can be made to the API. I would expect the above two to behave the same way.

Is there a reason it doesn't make sense for `DeviceMemoryResource(dev.device_id) == dev.memory_resource`?

In other words, can `DeviceMemoryResource()` instances be cached on `device_id`s for example, so that `DeviceMemoryResource(my_device) is DeviceMemoryResource(my_device)` ?

leofang: It should not be identical because it's neither how the driver works nor in users' best interest to unify.

At the driver level, users are free to create as many `cudaMempool_t` as they like, because each pool could be configured differently. So the cuda.core semantics of `dev.memory_resource` reads: "Give me the memory resource that wraps the device's current (or default, if no one touches it) mempool." However, users can create a side `DeviceMemoryResource` instance wrapping a new instance of `cudaMempool_t` (say by passing a pool config or by wrapping a foreign `cudaMempool_t` pointer) and allocate from there, instead of from `dev.memory_resource`, just like how they could do the same at the C level with raw driver APIs. In fact, #446 demonstrates such a use case: Allocating a new mempool for IPC purposes (be careful when reading the diff there, the branch has staled and deviated from the main branch for a while 😅). So if we cache and make them the same, we lose this capability entirely, and misaligns with C (raw driver) and C++ (cudax) designs.

There is one thing we do need to solve, which is `Device(0).memory_resource` from thread 0 != `Device(0).memory_resource` from thread 1 when comparing using the Python `__eq__`, despite under the hood they might wrap the same `cudaMempool_t`, due to `Device` being per-thread instances. But we can fix it in another PR.

shwina: Okay, that makes sense - thanks for clarifying!

shwina: So now we no longer have the segfault but a CUDAError when attempting to do the operation you suggested above. Am I doing something wrong?

cuda_core/tests/test_launcher.py
```diff
@@ -197,3 +206,99 @@ def test_cooperative_launch():
     config = LaunchConfig(grid=1, block=1, cooperative_launch=True)
     launch(s, config, ker)
     s.sync()
+
+
+@pytest.mark.parametrize(
+    "memory_resource_class",
+    [
+        DeviceMemoryResource,
```
leofang: We should have a pytest mark here to parametrize the MR to test against. Need to check against `Device().properties.memory_pools_supported` to decide if we use `DeviceMemoryResource` or `_SynchronousMemoryResource`, as done here:
https://github.com/NVIDIA/cuda-python/blob/24fde1755ad1efa6e7d73b2236dad9681fa0b727/cuda_core/cuda/core/experimental/_device.py#L1000-L1009, otherwise attempting to use `DeviceMemoryResource` in e.g. Windows TCC mode would lead to `CUDAError` as we see in the CI.

Previously, in the code sample `CUDAError` was uncaught, so we did not get a chance to revert the change of CuPy's global current stream, and it led to a segfault eventually. What I don't understand, though, is this test reproduces the issue in the code sample, but we hit no segfault. I wonder if it's because we happen to not execute any CuPy operations after this test failed... 🤔

leofang: Another nit regarding my suggestion: If we use `Device()` it'd initialize the device earlier than intended (before `init_cuda` fixture kicks in). Not a big deal if we want to keep the test code clean this way. Alternatively, we can defer the MR choice to the test body.

cuda_core/examples/memory_ops.py
leofang: I replied below

cuda_core/tests/test_launcher.py
shwina: > Need to check against Device().properties.memory_pools_supported to decide if we use DeviceMemoryResource or _SynchronousMemoryResource

Any chance we can turn this into a wrapper function, or just have `DeviceMemoryResource()` return the appropriate type when constructed?

leofang: Based on offline discussion I created https://github.com/NVIDIA/cuda-python/issues/726. Let's just wrap up and make the test run for now.

cuda_core/tests/test_launcher.py
```diff
@@ -197,3 +209,108 @@ def test_cooperative_launch():
...
+
+    # Create memory resource
+    if memory_resource_class == "device_memory_resource":
+        if (
+            handle_return(
+                driver.cuDeviceGetAttribute(
+                    driver.CUdevice_attribute.CU_DEVICE_ATTRIBUTE_MEMORY_POOLS_SUPPORTED, dev.device_id
+                )
+            )
+        ) == 1:
```
leofang: nit:
```suggestion
        if dev.properties.memory_pools_supported:
```

leofang (dismissed):

leofang (dismissed):

leofang:
pre-commit.ci autofix

leofang (approved):

PR #722, opened 2025-06-21
cuda_core forward compatibility changes.

rwgk:
## Description
Forward compatibility changes for `cuda_core`.

This helps us staying organized.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -32,7 +32,7 @@ jobs:
           echo "OLD_BRANCH=${OLD_BRANCH}" >> $GITHUB_ENV
 
       - name: Create backport pull requests
-        uses: korthout/backport-action@436145e922f9561fc5ea157ff406f21af2d6b363  # v3.2.0
+        uses: korthout/backport-action@0193454f0c5947491d348f33a275c119f30eb736  # v3.2.1
         with:
           copy_assignees: true
           copy_labels_pattern: true
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -31,13 +31,13 @@ jobs:
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
 
     - name: Initialize CodeQL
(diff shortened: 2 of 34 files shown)


cuda_core/cuda/core/experimental/_graph.py
```diff
@@ -476,7 +476,7 @@ def create_conditional_handle(self, default_value=None) -> driver.CUgraphConditi
             default_value = 0
             flags = 0
 
-        status, _, graph, _, _ = handle_return(driver.cuStreamGetCaptureInfo(self._mnff.stream.handle))
+        status, _, graph, *_, _ = handle_return(driver.cuStreamGetCaptureInfo(self._mnff.stream.handle))
```
leofang, line 479: Q: If we're packing anyway, maybe we pack them together?
```suggestion
        status, _, graph, *_ = handle_return(driver.cuStreamGetCaptureInfo(self._mnff.stream.handle))
```

leofang:
@rwgk let's merge #709, resolve the conflicts, rerun the CI and merge.

vzhurba01:
I took a look at the graphs side and while I'm ok with it, it does feels over-engineered.
The "tricks" used with `deps_info_update` feel bad because it's scattered through out the code (just 2 places but still... what if we end up doing more tricks to handle other forward compatibility APIs).

Thinking out loud, what if we move these weird APIs to a compatibility file/layer? At least this way we're keeping these tricks contained in a single place. Oh and by tricks I mean logic like `[None] * (len(deps_info) - 1)`, and the packing `*` which is placed for some APIs at the end while in the middle for others...

leofang:
> @rwgk let's merge #709, resolve the conflicts, rerun the CI and merge.

FYI, #709 is merged

rwgk:
> I took a look at the graphs side and while I'm ok with it, it does feels over-engineered.

I was very careful to make the unpacking consistent between all the uses, even where I could have taken a shortcut (`*_, _`), to reduce the cognitive burden. I find that far easier to read than duplicate code (what the LLMs suggested initially). The unpacking feature seems very intuitive and complete to me.

> `[None] * (len(deps_info) - 1)`

Yeah, it's a bit over the top, but there are only two, and the circumstances are special.

leofang:
Thanks, Ralf!

Issue #708, opened 2025-06-12
[FEA]: Improve toolshed/build_path_finder_dlls.py to minimize potential for accidents.


rwgk:
### Is this a duplicate?

- [ ] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

cuda.bindings

### Is your feature request related to a problem? Please describe.

Currently the output of the script requires manual cleanup. Even though that's documented, in practice oversights slip in.

### Describe the solution you'd like

The script should produce output that does not require manual cleanup.

rwgk:
> Currently the output of the script requires manual cleanup.

This was fixed by #723.

More improvements are possible (https://github.com/NVIDIA/cuda-python/pull/723#discussion_r2205110802), but that's sufficiently easy and low priority, can slip in the work opportunistically.

PR #723, opened 2025-06-24
Move `pathfinder` to cuda-python top level

### LINKED ISSUE #719: [FEA]: Move `path_finder` to cuda-python top level
### Is this a duplicate?

### Area

General cuda-python

We decided to move the `path_finder` from `cuda/bindings/path_finder` to `cuda/path_finder`.

`import cuda.path_finder` works.

_No response_

### Additional context

_No response_

rwgk:
## Description
Closes #719, #708
________

Note: The name is changed from `path_finder` to `pathfinder`. The word "pathfinder" is very common, and it avoids underscore/dash confusion between directory names and package names.
________

* Move `cuda.bindings.path_finder` → `cuda.pathfinder`, with backward compatibility: `cuda.bindings.path_finder` forwards to `cuda.pathfinder`.

    * Make `LoadedDL._handle` private (underscore prefix). It is now a platform-agnostic unsigned integer.

* Make `cuda-bindings` dependent on `cuda-pathfinder`, with a temporary backward compatibility layer. — `cuda.bindings.pathfinder` is slated to be removed after the next `cuda-bindings` release.

* Remove 32-bit DLL names from `SUPPORTED_WINDOWS_DLLS` and add runtime guard in `cuda.pathfinder.load_nvidia_dynamic_lib()`: "requires 64-bit Python".

* Add specific `DynamicLibNotFoundError` exception type to public API (this was `RuntimeError` before; now inherits from `RuntimeError`).

* Run `mypy-pathfinder` from `pre-commit` (after systematic `mypy` cleanup of library code).

* Adjust GitHub Actions jobs to the move.

* Add `nvidia_wheels_cu12` to `[project.optional-dependencies]` in `cuda_pathfinder/pyproject.toml` and ensure all libs are loaded successfully. **Treat all NVIDIA libs as "supported".**

* Mark `cuda.pathfinder` as **version 1.0.0**


cuda_pathfinder/cuda/pathfinder/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+    Args:
+        libname: The name of the library to load (e.g. "cudart", "nvvm", etc.)
+
+    Returns:
+        A LoadedDL object containing the library handle and path
+
+    Raises:
+        RuntimeError: If the library cannot be found or loaded
+    """
+    return _load_nvidia_dynamic_lib.load_lib(libname)
```
shwina: You could define this wrapper function in  the `_dynamic_libs` module and just import it here. This way your `__init__.py` here wouldn't have a ton of function definitions.

rwgk: This file is meant to be the public API in one view.

> This way your `__init__.py` here wouldn't have a ton of function definitions.

I understood exactly that was your goal: a flat list of available APIs.

Note that I moved the docstring here.

shwina: > This file is meant to be the public API in one view.

Indeed, everything we define (or import) here that is not prefixed with `_` constitutes the public API of the module `cuda.path_finder`. My above suggestion is just that we import `load_nvidia_dynamic_lib` rather than define it here.

rwgk: That would make the public API far less obvious. E.g. to see the docstring, they'd need to open a private file.

shwina: I don't think that's true. They could still do:

```python
from cuda.path_finder import load_nvidia_dynamic_lib
help(load_nvidia_dynamic_lib)
```

Same as they would do now?

rwgk: That'll only work interactively.

What I have now can be inspected in obvious ways directly in the sources, e.g. when looking at the sources on github. I can send URLs pointing to specific APIs in this __init__.py file.

Each function here will just be:

```
def function(...) -> ...:
    """docstring""
    return call_into_private_code(...)
```

That's exactly the public API, with a one-line call that's easy to ignore.

What's the point of hiding that away, especially hiding away the docstring and the type hints?

shwina: > That'll only work interactively.

The docstrings and type hints would also get picked up:

- Sphinx (which we use to generate API docs
-`pydoc`
- the developers' IDEs/lsp

Those are primarily the ways consumers of a package interact with docstrings or type hints, rather than looking directly at the source.

> What's the point of hiding that away, especially hiding away the docstring and the type hints?

It tightens the scope of `__init__.py`, whose job is:

- to include any initialization code for the module
- to import stuff from submodules that the module wants to expose
- to define `__all__` for the module if needed

As examples, we can look at the `__init__.py` from some other popular libraries:

- [requests](https://github.com/psf/requests/blob/main/src/requests/__init__.py)
- [pandas](https://github.com/pandas-dev/pandas/blob/main/pandas/__init__.py)

rwgk: > It tightens the scope of __init__.py, whose job is

The most important job: Show the public API

You didn't answer why you want to hide that away.

shwina: I'm advocating for keeping function and type _definitions_ outside of `__init__.py`. I don't think that "hides" anything from the user as we still import them here. 

The main reason I'm advocating for that is because `__init__.py` typically only _defines_ functions and types needed for module initialization, and _imports_ anything else. I think the examples I linked to above are a good demonstration of that. 

Defining functions and types beyond that serves to clutter `__init__.py`.

I would argue it makes the code base less navigable than more for people looking at the source. Do I expect the function `load_nvidia_dynamic_libs` to be defined in a file called `_dynamic_libs.py`, or a file called `__init__.py`?

If this all seems nitpicky, I apologize. While I do have a strong opinion here, I don't mind at all if another reviewer (or you, as the author of this PR) made the final call about this.

leofang: cuda.core's `__init__.py` is a good example for what Ashwin suggested. As a developer who occasionally peeks into someone else's `__init__.py`, I certainly get confused why we are defining things in there directly, as opposed to properly organizing them in respective public/private modules, and just import them. What's imported are considered public APIs (including types). Does it make sense?

rwgk:
I'm a bit lost, to be honest, and I’d appreciate some clarification so I can move this forward in the direction you want.

The goals that seemed most important to me originally (from a Slack post last Wednesday) were:
1. Dependency isolation – avoid importing unrelated logic (e.g. dynamic lib loading) when only working with headers
2. Public API should be obvious from the code – minimal need for external docs or tooling to understand what’s exposed

After Ashwin’s comment on Slack, I interpreted the request as: prioritize a flat public API, even if that means giving up on some of that isolation — especially when type hints are involved.

Then Leo wrote:

> as opposed to properly organizing them in respective public/private modules

That seems more aligned with what I had before the last commit (e.g., a file like `nvidia_dynamic_libs.py` directly under `pathfinder/`). But that seemed in conflict with Ashwin’s preference for flatness.

I’m happy to implement what you both want here — I just need a bit more clarity.

If we go back to what I proposed last week:

```python
from cuda.pathfinder import nvidia_dynamic_libs, nvidia_static_libs, nvidia_headers

nvidia_dynamic_libs.load("nvrtc")
nvidia_static_libs.find("cudart")
nvidia_headers.find_file("cuda.h")
```

The file structure was:

```
cuda/pathfinder/
├── __init__.py  # empty
├── nvidia_dynamic_libs.py
├── nvidia_static_libs.py
└── nvidia_headers.py
```

I thought that structure balanced modularity and clarity, and users could still import just what they needed. But it sounds like we might instead prefer:

```python
from cuda.pathfinder import load_nvidia_dynamic_lib, find_nvidia_static_lib, find_nvidia_header_file
```

That keeps the API flat, but does give up on the isolation goal unless we resort to function-local imports or other workarounds.

Could you please help me understand: What would be the preferred file/module structure to go along with the flat API? Should I keep separate modules like `nvidia_dynamic_libs.py` and re-export in `__init__.py`, or move everything into a single file (what would be the name)?

shwina:
Given there's some urgency for getting this PR in, @rwgk @leofang do you want to proceed with the PR as is and address the questions Ralf posed above as a follow-up?

leofang:
At this point we need to follow up offline. I think there is a gap that we unfortunately did not really cover during the weekly meeting. I suggest we don't cancel the dev meeting tomorrow and use the time to finalize it. 

This PR also needs at least @kkraus14's approval for the change of license.

rwgk:
> This PR also needs at least @kkraus14's approval for the change of license.

While I'd love to have the Apache license from the get-go, I could undo the license change in less than 10 minutes. Might be good to discuss that tomorrow, too.

shwina:
> Should I keep separate modules like nvidia_dynamic_libs.py and re-export in __init__.py, or move everything into a single file (what would be the name)?

Here’s the structure I would go with (opinionated):

```
path_finder
|-__init__.py
|-nvidia_dynamic_libs.py
|-nvidia_headers.py
|-...
```

Right now we have the following, which is also OK:

```
path_finder
|-__init__.py
|-nvidia_dynamic_libs/
  |-__init__.py
  |-find_nvidia_dl.py
  |-load_nvidia_dl.py
  |-linux_dl.py
  |-windows_dl.py
  |-supported_nvidia_libs.py
  |-...
|-nvidia_headers/
|-...
```

In the latter case, I would make sure to include an `__init__.py` in `nvidia_dynamic_libs/` and export only what’s needed from the `nvidia_dynamic_libs` subpackage. This way, consumers of the subpackage don’t need to import _its_ subpackages:

```python
# file: path_finder/nvidia_dynamic_libs/__init__.py

from .common import LoadedDL
from .load_nvidia_dl import load_nvidia_dynamic_lib
from .supported_nvidia_libs import SUPPORTED_NVIDIA_LIBNAMES
```

Now the top level `__init.py__` is simple. Note that it works regardless of whether you choose to use a flat file layout or subpackages:

```python
# file: path_finder/__init__.py

from .nvidia_dynamic_libs import LoadedDL, load_nvidia_dynamic_lib, SUPPORTED_NVIDIA_LIBNAMES
```

#### Regarding discoverability

Above all, I want us to provide a package that's easy and convenient for users to interact with. They shouldn't have to dig through the API reference or GitHub source to figure out what lives where.

Consider the initial layout:

```
cuda/pathfinder/
├── __init__.py  # empty
├── nvidia_dynamic_libs.py
├── nvidia_static_libs.py
└── nvidia_headers.py
```

The empty `__init__.py` would make it so that nothing is discoverable, i.e., doing:

```python
>>> import cuda.pathfinder
>>> cuda.pathfinder.<Tab>  # shows nothing, since the package is empty
``` 

Importing on the user's behalf in `__init__.py` solves this problem.

#### Regarding “Dependency Isolation”

> Dependency isolation – avoid importing unrelated logic (e.g. dynamic lib loading) when only working with headers

Please correct me if I’m wrong, but I’m assuming you mean that if we have:

```python
# file: path_finder/__init__.py

from .nvidia_dynamic_libs import ...
from .nvidia_headers import ...
```

My understanding is that this is breaking “dependency isolation”, as importing `path_finder` will import its sub-packages, which may be unnecessary if the user cares about functionality only from one sub-pacakge. Is that correct?

If so, I think it’s important to ask: **what specific problem(s) do we hope to solve by avoiding imports from sub-packages?**.  Perhaps we can use some time in our sync today to answer that. 

For example, are we concerned about the import time? Or do we expect to have sub-packages that import  optional dependencies (such as for Windows?). Each of these has solutions, but unless we are facing them today, I think it’s fine to cross that bridge when we get there.

cuda_bindings/cuda/bindings/path_finder.py
```diff
@@ -1,11 +1,8 @@
 # Copyright 2024-2025 NVIDIA Corporation.  All rights reserved.
-#
 # SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
 
-from cuda.bindings._path_finder.load_nvidia_dynamic_library import (
-    load_nvidia_dynamic_library as _load_nvidia_dynamic_library,
-)
-from cuda.bindings._path_finder.supported_libs import SUPPORTED_LIBNAMES as _SUPPORTED_LIBNAMES
+from cuda.pathfinder import SUPPORTED_NVIDIA_LIBNAMES as _SUPPORTED_LIBNAMES
+from cuda.pathfinder import load_nvidia_dynamic_lib as _load_nvidia_dynamic_library
```
kkraus14: Can we add some kind of deprecation warning if someone uses `cuda.bindings.path_finder` instead of `cuda.pathfinder`?

cuda_pathfinder/cuda/pathfinder/README.md
```diff
@@ -0,0 +1,51 @@
...
+centralized mechanism that ensures all libraries are found in the same way.
+
+## Maintenance Requirements
+
+These key components must be updated for new CUDA Toolkit releases:
+
+- `supported_nvidia_libs.SUPPORTED_LIBNAMES`
+- `supported_nvidia_libs.SUPPORTED_WINDOWS_DLLS`
+- `supported_nvidia_libs.SUPPORTED_LINUX_SONAMES`
+- `supported_nvidia_libs.EXPECTED_LIB_SYMBOLS`
```
kkraus14: My 2c: this information would be better suited for a `CONTRIBUTING.md` as opposed to a `README.md`

cuda_pathfinder/cuda/pathfinder/_version.py
```diff
@@ -0,0 +1,4 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+__version__ = "0.0.1"
```
kkraus14: @rwgk Have you given any thought to how we want to version this moving forward?

cuda_pathfinder/pyproject.toml
```diff
@@ -0,0 +1,58 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+[project]
+name = "cuda-pathfinder"
+description = "Pathfinder for CUDA components (mainly libs, headers)"
+authors = [{ name = "NVIDIA Corporation", email = "cuda-python-conduct@nvidia.com" }]
+license = "LicenseRef-NVIDIA-SOFTWARE-LICENSE"
```
kkraus14: This license would need to be updated as well

cuda_pathfinder/pyproject.toml
```diff
@@ -0,0 +1,58 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+[project]
+name = "cuda-pathfinder"
+description = "Pathfinder for CUDA components (mainly libs, headers)"
```
kkraus14:
```suggestion
description = "Pathfinder for CUDA components"
```

cuda_pathfinder/cuda/pathfinder/_version.py
rwgk: Short answer: Not really.

I think I need some guidance.

What I'd like to do:

Have the ability to update PyPI `cuda.pathfinder` frequently, and largely independently from `cuda_bindings` and `cuda_core`, with the strong expectation to break backward compatibility only very rarely. That makes me think it will be best to use obviously different/independent versioning. Where to start is a detail, e.g. 0.0.1 or 0.1.0, not sure.

leofang: I suggest we start with 0.1.0.

cuda_bindings/cuda/bindings/path_finder.py
leofang: Sounds like this is indeed the plan: https://github.com/NVIDIA/numba-cuda/pull/308#discussion_r2172234023

cuda_pathfinder/cuda/pathfinder/README.md
rwgk: I simply moved the list into `supported_nvidia_libs.py`: 13e08ae6ce144ac0d595801314bc3cae71f68ec8

cuda_pathfinder/cuda/pathfinder/_version.py
rwgk: Done: 50f4f8eade53d8cfd4ce2cc53fc61be7fa023e82

cuda_pathfinder/pyproject.toml
rwgk: Done: 787236c14e07e4173eced30d4066d821d6cbd016

Thanks for catching this!

cuda_pathfinder/pyproject.toml
rwgk: Done: 5b5078b8bf20a823e71065b492ebc0c69abd894f

.github/workflows/build-wheel.yml
```diff
@@ -70,6 +70,40 @@ jobs:
...
+        uses: pypa/cibuildwheel@5f22145df44122af0f5a201f93cf0207171beca7  # v3.0.0
+        env:
+          CIBW_BUILD: ${{ env.CIBW_BUILD }}
+          CIBW_ARCHS_LINUX: "native"
+          CIBW_BUILD_VERBOSITY: 1
+          CIBW_BEFORE_BUILD_WINDOWS: "pip install delvewheel"
+          CIBW_REPAIR_WHEEL_COMMAND_WINDOWS: "delvewheel repair --namespace-pkg cuda -w {dest_dir} {wheel}"
+        with:
+          package-dir: ./cuda_pathfinder/
+          output-dir: ${{ env.CUDA_PATHFINDER_ARTIFACTS_DIR }}
```
leofang: cibuildwheel does not support building pure Python (noarch) wheels (https://github.com/NVIDIA/cuda-python/pull/555#discussion_r2058918005), so this is an overkill. See how `cuda-python` is built below.

.github/workflows/build-wheel.yml
```diff
@@ -70,6 +70,40 @@ jobs:
...
+            export CHOWN=chown
+          else
+            export CHOWN="sudo chown"
+          fi
+          $CHOWN -R $(whoami) ${{ env.CUDA_PATHFINDER_ARTIFACTS_DIR }}
+          ls -lahR ${{ env.CUDA_PATHFINDER_ARTIFACTS_DIR }}
+
+      - name: Check cuda.pathfinder wheel
+        run: |
+          pip install twine
```
leofang: nit: it's fine to install `twine` early here, but then let's remove the one in line 131

.github/workflows/test-wheel-linux.yml
```diff
@@ -302,7 +316,7 @@ jobs:
       - name: Ensure cuda-python installable
         run: |
           if [[ "${{ matrix.LOCAL_CTK }}" == 1 ]]; then
-            pip install cuda_python*.whl
+            pip install cuda_pathfinder*.whl cuda_python*.whl
           else
-            pip install $(ls cuda_python*.whl)[all]
+            pip install $(ls cuda_pathfinder*.whl) $(ls cuda_python*.whl)[all]
```
leofang: These changes should move to as early as `cuda.bindings` tests. By the time we reach here, it'd be too late.

.github/workflows/test-wheel-windows.yml
```diff
@@ -268,7 +268,11 @@ jobs:
...
           if ('${{ matrix.LOCAL_CTK }}' -eq '1') {
-            pip install (Get-ChildItem -Filter cuda_python*.whl).FullName
+            $pathfinder = (Get-ChildItem -Filter cuda_pathfinder*.whl).FullName
+            $umbrella = (Get-ChildItem -Filter cuda_python*.whl).FullName
+            pip install $pathfinder $umbrella
           } else {
-            pip install "$((Get-ChildItem -Filter cuda_python*.whl).FullName)[all]"
+            $pathfinder = (Get-ChildItem -Filter cuda_pathfinder*.whl).FullName
+            $umbrella = (Get-ChildItem -Filter cuda_python*.whl).FullName
+            pip install "$pathfinder" "$umbrella[all]"
```
leofang: ditto

ci/tools/env-vars
```diff
@@ -28,6 +28,10 @@ elif [[ "${HOST_PLATFORM}" == win* ]]; then
 fi
 
 echo "${TOOLS_PATH}" >> $GITHUB_PATH
+CUDA_PATHFINDER_ARTIFACT_BASENAME="cuda-pathfinder-noarch"
+echo "CUDA_PATHFINDER_ARTIFACT_BASENAME=${CUDA_PATHFINDER_ARTIFACT_BASENAME}" >> $GITHUB_ENV
+echo "CUDA_PATHFINDER_ARTIFACT_NAME=${CUDA_PATHFINDER_ARTIFACT_BASENAME}-${SHA}" >> $GITHUB_ENV
+echo "CUDA_PATHFINDER_ARTIFACTS_DIR=$(realpath "${REPO_DIR}/cuda_pathfinder/dist")" >> $GITHUB_ENV
```
leofang: Took me awhile to digest this change and realize this: For `cuda-python` we hard-coded the name and path. This is also fine. Up to you 🙂

cuda_bindings/pyproject.toml
```diff
@@ -27,6 +27,7 @@ dynamic = [
     "readme",
 ]
 dependencies = [
+  "cuda-pathfinder",
```
leofang: as discussed in the meeting we probably need a lower bound here

cuda_pathfinder/cuda/pathfinder/_version.py
```diff
@@ -0,0 +1,4 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+__version__ = "0.1.0"
```
leofang: as discussed in the meeting we might want to make a major release

cuda_pathfinder/cuda/pathfinder/__init__.py
```diff
@@ -0,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from cuda.pathfinder.nvidia_dynamic_libs import SUPPORTED_NVIDIA_LIBNAMES, LoadedDL, load_nvidia_dynamic_lib
+
+__all__ = ["LoadedDL", "load_nvidia_dynamic_lib", "SUPPORTED_NVIDIA_LIBNAMES"]
```
leofang: could you also import/add the `__version__` string to here? see `cuda/core/__init__.py`

leofang (changes requested):
as discussed offline, leaving some comments on the CI change

cuda_pathfinder/cuda/pathfinder/nvidia_dynamic_libs.py
```diff
@@ -0,0 +1,8 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from cuda.pathfinder._dynamic_libs.load_dl_common import LoadedDL
+from cuda.pathfinder._dynamic_libs.load_nvidia_dynamic_lib import load_nvidia_dynamic_lib
+from cuda.pathfinder._dynamic_libs.supported_nvidia_libs import SUPPORTED_LIBNAMES as SUPPORTED_NVIDIA_LIBNAMES
+
+__all__ = ["LoadedDL", "load_nvidia_dynamic_lib", "SUPPORTED_NVIDIA_LIBNAMES"]
```
shwina: This whole file is unnecessary - this can go in a `_nvidia_dynamic_libs/__init__.py`.
 
If you want that module to be public for whatever reason, you can always rename the folder from `_nvidia_dynamic_libs -> nvidia_dynamic_libs`.

rwgk: Yeah ... I was thinking the same. It's the logical consequence of deciding:

A: tab completion is more important than dependency isolation
B: we don't want to make docstrings for public APIs easily discoverable

I removed this file (via `git mv nvidia_dynamic_libs.py __init__.py`): 36491c88a7639a655a9d6c153a4b36a366a85c6f

For the record: I agree slightly with `A`, I disagree strongly with `B`, but I'm going with the consensus.

.github/workflows/build-wheel.yml
rwgk: Done: commit c6187db3fd1dec0bc7d5a71f59095258ee795127

.github/workflows/build-wheel.yml
rwgk: Done: commit 346e8b9d608e155f3ce067cd5a05b4df9b1bbcaf, and then move to a separate step when I got to commit 9207966d6c23e5ba6bbc8cf17bd6ff2e66f2bc57

.github/workflows/test-wheel-linux.yml
rwgk: The state of the test environment is unclear to me tbh. I'll run the CI, hopefully from looking at the logs (and expected failure) it'll become clear to me.

ci/tools/env-vars
rwgk: Ah, we don't need a `SHA` because this is `noarch`?

Simpler seems better: commit 9207966d6c23e5ba6bbc8cf17bd6ff2e66f2bc57

cuda_bindings/pyproject.toml
rwgk: I was thinking it's best to add the lower bound as needed? It'll only become a consideration after we have a second `cuda-pathfinder` release.

cuda_pathfinder/cuda/pathfinder/_version.py
rwgk: Done: commit 3682f7db6675da7175cb694c11f968a15236d3ad

cuda_pathfinder/cuda/pathfinder/__init__.py
rwgk: Done: commit 7742c5c75d46ab4ee100380db7c68f1692aa682f

cuda_pathfinder/cuda/pathfinder/nvidia_dynamic_libs.py
shwina: > I removed this file (via git mv nvidia_dynamic_libs.py __init__.py): https://github.com/NVIDIA/cuda-python/commit/36491c88a7639a655a9d6c153a4b36a366a85c6f

To clarify, you can simplify the top-level `__init__.py` further here, by having a `_nvidia_dynamic_libs/__init__.py` file (which we don't have right now). The contents of that file should be:

```python
# file: _nvidia_dynamic_libs/__init__.py

from .load_dl_common import LoadedDL
from .load_nvidia_dynamic_lib import load_nvidia_dynamic_lib
from .supported_nvidia_libs import SUPPORTED_LIBNAMES as SUPPORTED_NVIDIA_LIBNAMES
```

Then in `pathfinder/__init__.py` you simply need:

```python
from ._dynamic_libs import LoadedDL, load_nvidia_dynamic_lib, ... etc
```

This makes it so that `pathfinder/__init__.py` doesn't have to go digging in the subpackages of `_dynamic_libs` for different pieces of functionality related to dynamic libraries. Instead, `_dynamic_libs` becomes a proper package that exports relevant symbols from its subpackages.

shwina: > B: we don't want to make docstrings for public APIs easily discoverable

I don't understand. They are definitely still discoverable?

rwgk: > easily discoverable

But let's move on. Since you prefer to make everything private (everything has an underscore prefix), we can freely nudge things around as we add more functionality.

leofang:
Alright, so the lesson here is that when there's a job dependency, we cannot let only one of the jobs build the dependency, because we have no way to declare dependency among matrix jobs; other jobs cannot wait for job 0 to complete.

I suggest to keep it simple we always build cuda-pathfinder in all jobs, but only let job 0 upload the artifact (for the test stage to use).

rwgk:
> Alright, so the lesson here is that when there's a job dependency, we cannot let only one of the jobs build the dependency, because we have no way to declare dependency among matrix jobs; other jobs cannot wait for job 0 to complete.
> 
> I suggest to keep it simple we always build cuda-pathfinder in all jobs, but only let job 0 upload the artifact (for the test stage to use).

I only saw this after the last "ok to test".

I have to drop work now for a few hours. I'll look here again later today.

Note that one job ran to completion before. All the other ones stumbled of the same issue. I have hopes that they might all work now (after commit db0ea567ce7c6d13bd9a7231a27b9a3da810cdc6).

leofang:
> Note that one job ran to completion before. All the other ones stumbled of the same issue.

yes, job 0 will succeed but not other jobs.

> I'll look here again later today.

No worries at all, I think it's about time to call it a day... 🙂

.github/workflows/build-wheel.yml
```diff
@@ -69,7 +69,43 @@ jobs:
...
           env
-      
+
+      - name: Install twine
+        run: |
+          pip install twine
+
+      # We only need/want a single pure python wheel, pick linux-64 index 0.
+      - name: Build and check cuda.pathfinder wheel
+        if: ${{ strategy.job-index == 0 && inputs.host-platform == 'linux-64' }}
```
leofang:
```suggestion
      # To keep the build workflow simple, all matrix jobs will build a wheel for later use within this workflow.
      - name: Build and check cuda.pathfinder wheel
```

.github/workflows/build-wheel.yml
```diff
@@ -69,7 +69,43 @@ jobs:
...
+        run: |
+          pushd cuda_pathfinder
+          pip wheel -v --no-deps .
+          popd
+
+      - name: List the cuda.pathfinder artifacts directory
+        if: ${{ strategy.job-index == 0 && inputs.host-platform == 'linux-64' }}
```
leofang:
```suggestion
```

.github/workflows/build-wheel.yml
```diff
@@ -69,7 +69,43 @@ jobs:
...
+        run: |
+          if [[ "${{ inputs.host-platform }}" == win* ]]; then
+            export CHOWN=chown
+          else
+            export CHOWN="sudo chown"
+          fi
+          $CHOWN -R $(whoami) cuda_pathfinder/*.whl
+          ls -lahR cuda_pathfinder
+
+      - name: Check cuda.pathfinder wheel
```
leofang, line 96:
```suggestion
      # We only need/want a single pure python wheel, pick linux-64 index 0.
      # This is what we will use for testing & releasing.
      - name: Check cuda.pathfinder wheel
```

.github/workflows/build-wheel.yml
leofang: nice thx

.github/workflows/test-wheel-linux.yml
leofang: Test failures show what I meant here:
https://github.com/NVIDIA/cuda-python/actions/runs/16038987447?pr=723

ci/tools/env-vars
leofang: Not so much about `SHA` (it is the long-format commit ID). Just we don't need a very long name to encode the combo of PY x CUDA x OS.

leofang:
> I suggest to keep it simple we always build cuda-pathfinder in all jobs, but only let job 0 upload the artifact (for the test stage to use).

done in commit 9a9f1d7, the build workflow now works

cuda_pathfinder/cuda/pathfinder/__init__.py
```diff
@@ -0,0 +1,9 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from cuda.pathfinder._dynamic_libs.load_dl_common import LoadedDL
+from cuda.pathfinder._dynamic_libs.load_nvidia_dynamic_lib import load_nvidia_dynamic_lib
+from cuda.pathfinder._dynamic_libs.supported_nvidia_libs import SUPPORTED_LIBNAMES as SUPPORTED_NVIDIA_LIBNAMES
+from cuda.pathfinder._version import __version__
+
+__all__ = ["LoadedDL", "load_nvidia_dynamic_lib", "SUPPORTED_NVIDIA_LIBNAMES", "__version__"]
```
leofang: I think technically we don't need this, this file is clean, but doesn't hurt to keep it

rwgk:
I think ruff will more or less enforce this. Without we get unused errors. I'll double check later.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

cuda_pathfinder/cuda/pathfinder/__init__.py
rwgk: This is the ruff error after removing `__all__`:

```
a1u1g-mil-0536.nvidia.com:~/forked/cuda-python $ pre-commit run --all-files
ruff.....................................................................Failed
- hook id: ruff
- exit code: 1
- files were modified by this hook

cuda_pathfinder/cuda/pathfinder/__init__.py:6:87: F401 `cuda.pathfinder._dynamic_libs.supported_nvidia_libs.SUPPORTED_LIBNAMES` imported but unused; consider removing, adding to `__all__`, or using a redundant alias
  |
4 | from cuda.pathfinder._dynamic_libs.load_dl_common import LoadedDL as LoadedDL
5 | from cuda.pathfinder._dynamic_libs.load_nvidia_dynamic_lib import load_nvidia_dynamic_lib as load_nvidia_dynamic_lib
6 | from cuda.pathfinder._dynamic_libs.supported_nvidia_libs import SUPPORTED_LIBNAMES as SUPPORTED_NVIDIA_LIBNAMES
  |                                                                                       ^^^^^^^^^^^^^^^^^^^^^^^^^ F401
7 | from cuda.pathfinder._version import __version__ as __version__
  |
  = help: Use an explicit re-export: `SUPPORTED_LIBNAMES as SUPPORTED_LIBNAMES`


Fixed 3 errors:
- cuda_pathfinder/cuda/pathfinder/__init__.py:
    3 × F401 (unused-import)

Found 4 errors (3 fixed, 1 remaining).

ruff-format..............................................................Passed
Check SPDX-License-Identifier............................................Passed
bandit...................................................................Passed
```

leofang: It is OK to mark `# noqa: F401` for those lines. Ruff can be convinced.

rwgk: Done: commit 989db081c6eb437766a34cc6810cc09fa0e3d8f6

My view of it: It's slightly shorter, but also less readable.

shwina: > My view of it: It's slightly shorter, but also less readable.

I really don't mean to sound like a broken record, but I've suggested an approach several times to make this `__init__.py` an easy-to-read, one-liner:

https://github.com/NVIDIA/cuda-python/pull/723#discussion_r2181068931

ci/tools/run-tests
```diff
@@ -9,14 +9,31 @@
...
+pushd ./cuda_pathfinder
+echo "Installing pathfinder wheel"
+pwd
+ls
+pip install $(ls *.whl)[test]
+if [[ "${test_module}" == "pathfinder" ]]; then
+  echo "Running pathfinder tests"
+  pwd
+  pytest -ra -s -v tests/
+fi
```
leofang: We need to either move pathfinder tests to after cuda-core tests, or have a separate pipeline dedicated to it. Right now, we rely on cuda-bindings/cuda-core installing the dependencies for us. If we run this test first, we don't have access to any CTK components in the wheel pipelines (local CTK ones are fine).

.github/workflows/test-wheel-linux.yml
leofang: The changes to the cuda-python step should be reverted now that you install cuda-pathfinder the first.

.github/workflows/test-wheel-windows.yml
leofang: The changes to the cuda-python step should be reverted now that you install cuda-pathfinder the first.

leofang (changes requested):

ci/tools/run-tests
rwgk: Done: commit fd18e1e2a7dbb4d102c123c300d476405425aaf2

rwgk:
@leofang Does this ring any bells?

* https://github.com/NVIDIA/cuda-python/actions/runs/16061794063/job/45329524330?pr=723

```
FAILED tests/test_load_nvidia_dynamic_lib.py::test_load_nvidia_dynamic_lib[cudart] - RuntimeError: Child process failed for libname='cudart' with exit code 1
--- stdout-from-child-process ---
<end-of-stdout-from-child-process>
--- stderr-from-child-process ---
  File "C:\a\_tool\Python\3.12.10\x64\Lib\site-packages\cuda\pathfinder\_dynamic_libs\load_dl_windows.py", line 115, in load_with_abs_path
    handle = win32api.LoadLibraryEx(found_path, 0, flags)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pywintypes.error: ([193](https://github.com/NVIDIA/cuda-python/actions/runs/16061794063/job/45329524330?pr=723#step:25:194), 'LoadLibraryEx', '%1 is not a valid Win32 application.')

The above exception was the direct cause of the following exception:

  File "C:\a\cuda-python\cuda-python\cuda_pathfinder\tests\spawned_process_runner.py", line 41, in __call__
    self.target(*self.args, **self.kwargs)
  File "C:\a\cuda-python\cuda-python\cuda_pathfinder\tests\test_load_nvidia_dynamic_lib.py", line 66, in child_process_func
    loaded_dl_fresh = load_nvidia_dynamic_lib(libname)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\a\_tool\Python\3.12.10\x64\Lib\site-packages\cuda\pathfinder\_dynamic_libs\load_nvidia_dynamic_lib.py", line 61, in load_nvidia_dynamic_lib
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\a\_tool\Python\3.12.10\x64\Lib\site-packages\cuda\pathfinder\_dynamic_libs\load_nvidia_dynamic_lib.py", line 45, in _load_lib_no_cache
    return load_with_abs_path(libname, found.abs_path)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\a\_tool\Python\3.12.10\x64\Lib\site-packages\cuda\pathfinder\_dynamic_libs\load_dl_windows.py", line 117, in load_with_abs_path
    raise RuntimeError(f"Failed to load DLL at {found_path}: {e}") from e
RuntimeError: Failed to load DLL at C:\a\_tool\Python\3.12.10\x64\Lib\site-packages\nvidia\cuda_runtime\bin\cudart32_110.dll: (193, 'LoadLibraryEx', '%1 is not a valid Win32 application.')
<end-of-stderr-from-child-process>
```

rwgk:
Assuming the last commit (bf4814261dfef4557d69527840ea8dc1e92de6b1) works as hoped, the Windows `cudart` 11.8.0 `'%1 is not a valid Win32 application.'` error may be the only error left.

leofang:
This error happens usually because the exe or dll is not compatible with the platform, which is indeed the case based on the error log 
```
RuntimeError: Failed to load DLL at C:\a\_tool\Python\3.12.10\x64\Lib\site-packages\nvidia\cuda_runtime\bin\cudart32_110.dll: (193, 'LoadLibraryEx', '%1 is not a valid Win32 application.')
```
We're trying to load the 32-bit cudart (`cudart32_110.dll`) on a 64-bit platform.

Another bug is: Both wheel and local CTK pipelines hit the same error message, meaning we also installed wheels to the local CTK pipelines. There is at lease one bug in the CI update that caused this.

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/supported_nvidia_libs.py
leofang, line 1: Bug No.1: we should remove 32-bit cudart DLLs from line 253-255.

.github/workflows/test-wheel-windows.yml
```diff
@@ -265,6 +271,14 @@ jobs:
         shell: bash --noprofile --norc -xeuo pipefail {0}
         run: run-tests core
 
+      # Run only after cuda.bindings tests, so that all wheels are installed already.
+      - name: Run cuda.pathfinder tests
```
leofang: "Bug" No.2: cuda.pathfinder tests probably should be run after cuda.bindings but before cuda.core, because cuda.core needs the cudart wheel as a test dependency (it's used by CuPy, not by cuda.core).

In the long run, we should really move cuda.pathfinder to its own test workflow. This kind of execution-order-dependent issues is not obvious.

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/supported_nvidia_libs.py
rwgk: Done: commit 135a37d263802a9053d6539508830a5921d28096

.github/workflows/test-wheel-windows.yml
rwgk: Hm, I made the change locally, but then backed it out again: After removing the 32-bit DLLs (135a37d263802a9053d6539508830a5921d28096), I'd expect the tests to work, and it tests a little bit more.

rwgk:
For easy future reference: [All tests passed](https://github.com/NVIDIA/cuda-python/actions/runs/16095582749?pr=723)

This includes installing all Nvidia wheels and loading all libs:

`pyproject.toml`:
```
nvidia_wheels_cu12 = [
    "nvidia-cublas-cu12",
    "nvidia-cuda-nvcc-cu12",
    "nvidia-cuda-nvrtc-cu12",
    "nvidia-cuda-runtime-cu12",
    "nvidia-cufft-cu12",
    "nvidia-cufile-cu12; sys_platform != 'win32'",
    "nvidia-curand-cu12",
    "nvidia-cusolver-cu12",
    "nvidia-cusparse-cu12",
    "nvidia-npp-cu12",
    "nvidia-nvfatbin-cu12",
    "nvidia-nvjitlink-cu12",
    "nvidia-nvjpeg-cu12",
]
```

rwgk:
@kkraus14 @leofang This is ready for review now. Please see the PR description for a high-level overview of all the things included in this PR.

.github/workflows/test-wheel-linux.yml
```diff
@@ -286,6 +292,11 @@ jobs:
       - name: Set up compute-sanitizer
         run: setup-sanitizer
 
+      - name: Run cuda.pathfinder tests with see_what_works
```
kkraus14, line 295: Not a blocker for getting this in, but in theory since we're just testing the viability of finding and loading libraries, we shouldn't need to run these tests with a GPU where we may need to split these tests into a separate workflow.

I think the only tricky part would be handling the driver, `libcuda`, where I'm not sure how easy it is to install and load without a GPU.

.pre-commit-config.yaml
```diff
@@ -37,5 +37,13 @@ repos:
           - --ini
           - .bandit
 
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.8.0
```
kkraus14: Please use SHA instead of tag

cuda_bindings/cuda/bindings/path_finder.py
kkraus14: I think we made the decision to keep this for now and introduce the warning in a future release.

cuda_bindings/pyproject.toml
kkraus14: It's a problem where when we release a `2.0` with breaking changes, then older releases of `cuda_bindings` would break from finding it.

cuda_pathfinder/cuda/pathfinder/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Public API for loading NVIDIA Dynamic Libs
+
+* `cuda.pathfinder.SUPPORTED_NVIDIA_LIBNAMES` (`tuple[str]`)
+
+* `cuda.pathfinder.load_nvidia_dynamic_lib(libname: str) -> LoadedDL`
+
+* `cuda.pathfinder.LoadedDL`:
+    * `handle` (platform-specific type)
+    * `abs_path` (`str`)
```
kkraus14, line 10: Should we consider returning a pathlib object here instead of a string?

cuda_pathfinder/cuda/pathfinder/README.md
```diff
@@ -0,0 +1,46 @@
+# `cuda.pathfinder` Module
+
+## Public API for loading NVIDIA Dynamic Libs
+
+* `cuda.pathfinder.SUPPORTED_NVIDIA_LIBNAMES` (`tuple[str]`)
+
+* `cuda.pathfinder.load_nvidia_dynamic_lib(libname: str) -> LoadedDL`
+
+* `cuda.pathfinder.LoadedDL`:
+    * `handle` (platform-specific type)
```
kkraus14: Any concern with this being platform-specific type and API stability? Is this effectively the handle returned from `dlopen` on Linux?

.github/workflows/test-wheel-linux.yml
rwgk: That sounds good.

I think it'll be useful to keep the `see_what_works` testing here, for very inexpensive extra test coverage, but probably testing can and should indeed be run on machines without a GPU.

Currently `libcuda` isn't covered by the pathfinder.

.pre-commit-config.yaml
rwgk: Done: commit 74ef126b5aeee3c10c6b9d591abc5f7d9fe32f26

cuda_bindings/pyproject.toml
rwgk: Changed to `~= 1.0`: commit 5e85328192be7843fa6791cc2060d2572577bb7e

The idea is that we will not make breaking changes in 1.x.y releases.

cuda_pathfinder/cuda/pathfinder/README.md
rwgk: Yes, for Linux it's a plain `int`, for Windows a `pywintypes.HANDLE`. This seems to be mainstream for each platform, and very unlikely to require changes.

I guess we could introduce a level of indirection (some wrapper type), but I figure that'll be far more trouble (confused users) than we could hope to avoid.

cuda_pathfinder/cuda/pathfinder/README.md
rwgk: My vote: No. For simplicity.

Plain `str` paths are still ubiquitous, and I'd guess always will be.

.github/workflows/test-wheel-linux.yml
rwgk: That sounds good.

I think it'll be useful to keep the `see_what_works` testing here, for very inexpensive extra test coverage, but probably testing can and should indeed be run on machines without a GPU.

Currently `libcuda` isn't covered by the pathfinder.

cuda_pathfinder/cuda/pathfinder/README.md
kkraus14: Agreed on the wrapper type. Last question since this is public API is is this `handle` needed to be a public attribute today? Do we expect usage outside of it how we use it for `cuda.bindings`?

kkraus14 (approved):
Approved, but would like to finish the conversation on whether `handle` is needed to be a public attribute

rwgk:
> Approved, but would like to finish the conversation on whether handle is needed to be a public attribute

Thanks!

I'm thinking of that as essential, with the primary use case in mind, like here:

* https://github.com/NVIDIA/cuda-python/blob/126bca0fd07f8cbf64c97f0691544d503b62c1e3/cuda_bindings/cuda/bindings/_internal/nvvm_linux.pyx#L55

It's unfortunate that there is no universal handle type, but I believe what we're using (in the checked-in code already) are de-facto standards.

Introducing our own handle type: presumed over-engineering, more likely to cause confusion than being helpful.

Hiding the handle object: presumed to cause usability hardships.

While writing this, I realized that I lost comments that I had earlier, about how to use the handle objects in cython. I'll work on adding them back here (in `load_dl_common.py`) asap:

```
if IS_WINDOWS:
    import pywintypes

    HandleType = pywintypes.HANDLE
else:
    HandleType = int
```

rwgk:
In my previous comment I wrote:

> It's unfortunate that there is no universal handle type, but I believe what we're using (in the checked-in code already) are de-facto standards.

I quizzed the LLMs a bit, and I'm surprised to learn:

* ChatGPT: `ctypes.WinDLL` (and `CDLL`) is **much more widely used** and **idiomatic** for dynamically loading and calling functions from DLLs in Python — especially in cross-platform projects like yours.

* claude.ai: `ctypes.windll` is significantly more widely used than `win32api` for dynamic library loading, ...

@kkraus14 @leofang does that match your experience?

If yes, I believe I should rework `load_dl_windows.py`, in this PR, and change the `LoadDL.handle` type to `ctypes.CDLL`. It will be much nicer to have a platform-independent type.

(I'm also much better set up now to get this done quickly, compared to my situation a couple months ago: I have easy interactive access to a Windows development environment now, and fairly comprehensive CI testing is in place already.)

kkraus14:
Yes, ctypes is probably the most common way that a shared library is loaded in Python in this way. That being said, I don't think we particularly want to encourage users to start making function calls into the loaded cuda library, so returning a `CDLL` doesn't feel like the right choice in my opinion.

My 2c: I think we should remove the `handle` attribute from being part of the public API for now. For our usage in `cuda.bindings` we can access an internal attribute in `_handle` or something similar for now and as we gauge usage we can see if / how we'd like to expose something like this.

If someone wanted a ctypes object based on `LoadDL` it would be quite straightforward: `CDLL(LoadDL.abs_path)` which I think is sufficient for now.

rwgk:
> Yes, ctypes is probably the most common way that a shared library is loaded in Python in this way. That being said, I don't think we particularly want to encourage users to start making function calls into the loaded cuda library, so returning a `CDLL` doesn't feel like the right choice in my opinion.
> 
> My 2c: I think we should remove the `handle` attribute from being part of the public API for now. For our usage in `cuda.bindings` we can access an internal attribute in `_handle` or something similar for now and as we gauge usage we can see if / how we'd like to expose something like this.
> 
> If someone wanted a ctypes object based on `LoadDL` it would be quite straightforward: `CDLL(LoadDL.abs_path)` which I think is sufficient for now.

Cool! Then we can simplify and standardize to:

```
@dataclass
class LoadedDL:
    """Represents a loaded dynamic library with minimal public API."""
    abs_path: Optional[str]
    was_already_loaded_from_elsewhere: bool
    _handle: int  # Platform-agnostic unsigned pointer value
```

With that, `win32api` and `pywintypes` will appear exclusively in `load_dl_windows.py`.

This is based on the observation that ultimately `_handle` is a pointer type on all platforms:

For Linux that's extremely obvious:

```
void *dlopen(const char *filename, int flags);
```

For Windows we have to drill down:

* https://learn.microsoft.com/en-us/previous-versions/ms915415(v=msdn.10)

```
HMODULE LoadLibraryEx(
LPCTSTR lpLibFileName, 
HANDLE hFile, 
DWORD dwFlags );
```

* https://learn.microsoft.com/en-us/windows/win32/winprog/windows-data-types

`HMODULE` — A handle to a module. This is the base address of the module in memory. `HMODULE` and `HINSTANCE` are the same in current versions of Windows, but represented different things in 16-bit Windows. This type is declared in `WinDef.h` as follows: `typedef HINSTANCE HMODULE;`

`HINSTANCE` — A handle to an instance. This is the base address of the module in memory. `HMODULE` and `HINSTANCE` are the same today, but represented different things in 16-bit Windows. This type is declared in `WinDef.h` as follows: `typedef HANDLE HINSTANCE;`

`HANDLE` — A handle to an object.This type is declared in `WinNT.h` as follows: `typedef PVOID HANDLE;`

`PVOID` — A pointer to any type. This type is declared in `WinNT.h` as follows: `typedef void *PVOID;`

With that, we just have to be careful to consistently cast from/to unsigned integers.

Linux: `CDLL._handle` is already an unsigned integer.

For Windows, claude.ai suggests:

```
def _handle_to_int(native_handle: pywintypes.HANDLE) -> int:
    """Convert Windows HANDLE to unsigned int."""
    # Windows HANDLE is already an integer representation of a pointer
    # Just ensure it's treated as unsigned
    handle_int = int(native_handle)
    if handle_int < 0:
        # Convert from signed to unsigned representation
        handle_int += 2**64  # Assuming 64-bit pointers
    return handle_int
```

rwgk:
@kkraus14 Could you please take another look?

The changes in `cuda_pathfinder` turned out to be really straightforward:

* 46fd972099e69a04b457c24b0505ebf29f447f86 — Make `LoadedDL.handle` an unsigned integer also for Windows.
* 428a2dcbaee23e041791457227ebb949c1a39c31 — Rename `LoadedDL.handle` → `LoadedDL._handle_uint`
* 40968855f6793497e8d622efe2b6842de4cfa25c — Move `_handle_uint` last in `LoadedDL`, to emphasize that this is private member.

The changes in `cuda/bindings/path_finder.py` probably look a bit hairy, but I believe in a month or two we can simply delete the entire file.

rwgk:
@kkraus14 I'm testing the mostly auto-generated `load_dl_windows.py` `win32api` → `ctypes` conversion under PR #751. Assuming it passes, we could add the changes to this PR, so that `cuda-pathfinder` does not have any external dependencies from the start.

kkraus14:
I'm quite happy with the latest set of changes in terms of the state of the public API at this point.

As far as the type of the now private handle attribute, I would defer entirely to your best judgement on how it is typed and represented based on our own consumption of it.

Will do a deeper dive implementation review when I get some spare cycles.

rwgk:
> As far as the type of the now private handle attribute, I would defer entirely to your best judgement on how it is typed and represented based on our own consumption of it.

I'm very happy with the latest state of the pathfinder code, as under PR #751.

Without your feedback, I'd probably have standardized on `ctypes.CDLL`, but standardizing on unsigned int instead seems clearly better for these reasons:

* It doesn't get more robust / future-proof than that. We're not committing to any library.
* We're not "inviting" people to make `ctypes`-based function calls into the loaded shared libraries.
* For power users, a platform-agnostic convention will be the easiest to work with.

To the last point: unsigned ints are the de-facto standard under Linux, signed ints seem to be more common under Windows. In theory, we could go with that, but then people would always have to think twice. A simple "it's always an unsigned int" gives them a firm reference point on our side, then they only need to figure out what the "other side" (e.g. `win32api`) expects, and only if they actually need to interface.

rwgk:
Testing under PR #751 passed: https://github.com/NVIDIA/cuda-python/actions/runs/16180272303?pr=751

I'll hold off moving the changes over to this PR until after your review. (I assume the split makes the review a little easier.)

rwgk:
For last-minute polishing, I worked under #751 on hardening the `ruff.lint` config (going by LLM suggestions) ... and I'm so glad I got to it, this was a pretty silly oversight:

```
ruff (legacy alias)......................................................Failed
- hook id: ruff
- exit code: 1

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_common.py:10:7: N818 Exception name `DynamicLibNotFound` should be named with an Error suffix
   |
10 | class DynamicLibNotFound(RuntimeError):
   |       ^^^^^^^^^^^^^^^^^^ N818
11 |     pass
   |

Found 1 error.
```

(The other changes in e90855e18233b69a043e413e112f9ea22f73f35b and 15c57912681f057f84ecce91a52510deae0c2da4 are just nice to haves.)

cuda_bindings/cuda/bindings/path_finder.py
```diff
@@ -1,13 +1,47 @@
...
-    load_nvidia_dynamic_library as _load_nvidia_dynamic_library,
-)
-from cuda.bindings._path_finder.supported_libs import SUPPORTED_LIBNAMES as _SUPPORTED_LIBNAMES
+from cuda.pathfinder import SUPPORTED_NVIDIA_LIBNAMES as _SUPPORTED_LIBNAMES
+from cuda.pathfinder import load_nvidia_dynamic_lib
 
 __all__ = [
     "_load_nvidia_dynamic_library",
     "_SUPPORTED_LIBNAMES",
 ]
```
kkraus14: Given this module is effectively part of the public API, is using `__all__` here sufficient to guard against the non-underscored import names and class we're introducing?

kkraus14:
This looks good to me, sans the one comment above about the visibility of things in the `cuda.bindings.path_finder` and avoiding accidentally introducing public APIs that we don't wish to support.

cuda_bindings/cuda/bindings/path_finder.py
rwgk: Not perfectly:

```
>>> import cuda.bindings.path_finder
>>> dir(cuda.bindings.path_finder)
['HandleType', 'LoadedDL', 'Optional', '_SUPPORTED_LIBNAMES', '__all__', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', '_load_nvidia_dynamic_library', 'dataclass', 'load_nvidia_dynamic_lib', 'supported_nvidia_libs']
```

The `__all__` mainly (or maybe only) affects the `import *` behavior:

```
>>> dir()
['__annotations__', '__builtins__', '__doc__', '__loader__', '__name__', '__package__', '__spec__', 'cuda']
>>> from  cuda.bindings.path_finder import *
>>> dir()
['_SUPPORTED_LIBNAMES', '__annotations__', '__builtins__', '__doc__', '__loader__', '__name__', '__package__', '__spec__', '_load_nvidia_dynamic_library', 'cuda']
```

I think it's fine though, in the interest of not increasing the complexity here, and because I want to remove this entire file as soon as possible. The game plan:

* After this PR is merged, create PyPI `cuda-pathfinder` (`v1.0.0`) asap.
* Update `cuda-cccl` to use `cuda-pathfinder`.
* Update `cuda-bindings` internals to use `cuda-pathfinder` directly.
* Ask around who else is using `cuda.bindings.path_finder` and take care of those. I figure there will only be a couple at most, given the short lifetime of `cuda.bindings.path_finder`.
* Remove this file and the `_path_finder` subdirectory.

rwgk:
@kkraus14 I merged in `main` (SPDX changes) and put #751 as-is on top here: commit c0509beebe88bcbd223c7e8fe817e6a1d4fb7266

Everything else is as before. I think this is ready for merging (pending completion of tests).

@leofang Merging is blocked by your request for changes, could you please help?

rwgk:
I just added one more commit to this PR, just one extra line: eedf658020b1befe2de88f2f68dd4efcbe3c04c8

After reading up on publishing to PyPI, I'm guessing this is all that's needed on this end to publish `cuda-pathfinder` wheels to PyPI?

cuda_bindings/cuda/bindings/path_finder.py
kkraus14: I do not think it's fine. We're following strict semver here, which means if you can publicly access something via `cuda.bindings.path_finder` that doesn't begin with an underscore that it is part of our API and we will not break it in bumping minor versions.

If this is only a temporary stopgap that we don't intend to survive until a release then that is fine.

rwgk:
I was thinking of making a release right after this is merged.

This whole file is already deprecated per the README. I guess we could just add more underscores, would that be acceptable?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

cuda_bindings/cuda/bindings/path_finder.py
rwgk: I've added commit 58ca181684671988101dcac68fa7b7d3dfc527aa, which fully sanitizes the `cuda/bindings/path_finder.py` module namespace. Retesting.

cuda_pathfinder/mypy.ini
```diff
@@ -0,0 +1,33 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+[mypy]
```
kkraus14: nitpick: can we move this to a `[tool.mypy]` section in pyproject.toml instead of having a separate ini file here?

cuda_bindings/cuda/bindings/_path_finder/README.md
```diff
@@ -1,56 +1,3 @@
...
-
-## Maintenance Requirements
-
-These key components must be updated for new CUDA Toolkit releases:
-
-- `supported_libs.SUPPORTED_LIBNAMES`
-- `supported_libs.SUPPORTED_WINDOWS_DLLS`
-- `supported_libs.SUPPORTED_LINUX_SONAMES`
-- `supported_libs.EXPECTED_LIB_SYMBOLS`
+`cuda.bindings.pathfinder` is slated to be removed after the next `cuda-bindings` release.
```
kkraus14:
```suggestion
`cuda.bindings.pathfinder` is deprecated and slated to be removed in the next `cuda-bindings` major version release.
```

cuda_bindings/cuda/bindings/path_finder.py
```diff
@@ -1,11 +1,13 @@
 # SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
-#
 # SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
 
-from cuda.bindings._path_finder.load_nvidia_dynamic_library import (
+# This file is for TEMPORARY BACKWARD COMPATIBILITY only.
+# cuda.bindings.pathfinder is slated to be removed after the next cuda-bindings release.
```
kkraus14:
```suggestion
# cuda.bindings.pathfinder is deprecated and slated to be removed in the next cuda-bindings major version release.
```

cuda_pathfinder/cuda/pathfinder/README.md
kkraus14: This is no longer a public attribute / property I believe?

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py
```diff
@@ -58,4 +61,11 @@ def load_nvidia_dynamic_library(libname: str) -> LoadedDL:
...
         RuntimeError: If the library cannot be found or loaded
     """
-    return _load_nvidia_dynamic_library_no_cache(libname)
+    pointer_size_bits = struct.calcsize("P") * 8
+    if pointer_size_bits != 64:
+        raise RuntimeError(
+            f"cuda.pathfinder.load_nvidia_dynamic_lib() requires 64-bit Python."
+            f" Currently running: {pointer_size_bits}-bit Python"
+            f" {sys.version_info.major}.{sys.version_info.minor}"
+        )
```
kkraus14, line 70: Is this something we need to check or can we catch an error elsewhere? These types of input checks tend to add overhead that we'd like to avoid.

cuda_pathfinder/tests/conftest.py
```diff
@@ -9,7 +9,7 @@ def pytest_configure(config):
     config.custom_info = []
 
 
-def pytest_terminal_summary(terminalreporter, exitstatus, config):
+def pytest_terminal_summary(terminalreporter, exitstatus, config):  # noqa: ARG001
```
kkraus14, line 12: Is this "noqa" needed or should we drop the `exitstatus` argument?

toolshed/build_pathfinder_dlls.py
```diff
@@ -7,24 +7,26 @@
...
-# The output of this script
-# requires obvious manual edits to remove duplicates and unwanted dlls.
+# The output of this script is expected to be usable as-is.
 
+import collections
 import sys
 
-LIBNAMES_IN_SCOPE_OF_CUDA_BINDINGS_PATH_FINDER = (
+# ATTENTION: Ambiguous shorter names need to appear after matching longer names
+#            (e.g. "cufft" after "cufftw")
```
kkraus14, line 16: Any way to better protect ourselves here?

kkraus14 (dismissed):

cuda_pathfinder/mypy.ini
rwgk: Done: commit b73f42133a9b023073b0a80d6f433361c8f457a5

(that's a "today I learned")

cuda_bindings/cuda/bindings/_path_finder/README.md
rwgk: Done: commit 88c4ebcfacc812c7031fe5ae52a3cdc1ef35e060

(There was also an underscore missing that slipped my attention before.)

cuda_bindings/cuda/bindings/path_finder.py
rwgk: Done: commit 88c4ebcfacc812c7031fe5ae52a3cdc1ef35e060

cuda_pathfinder/cuda/pathfinder/README.md
rwgk: Removed (it was an oversight). Thanks for catching this!

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py
rwgk: This is catching the issue "just in time" because I figured that'll be the best way to ensure it happens when it's needed, and is sure not to be a false positive.

This function is decorated with `@functools.cache`, and the runtime for actually finding and loading the library is probably several orders of magnitude (millions?) more than the `struct.calcsize("P")` call.

We could move `struct.calcsize("P")` into another `@functools.cache` helper function, but considering added complexity vs runtime gain, I figure that's on the over-engineering side of the spectrum.

Please let me know if you think this is worth more effort.

cuda_pathfinder/tests/conftest.py
rwgk: Ah, this is the result of a small battle between `mypy` and `ruff`. The LLM I was using (not sure anymore which one) suggested this is the overall best fix/workaround. I could try to dig up the conversation if you think this is worth more effort.

toolshed/build_pathfinder_dlls.py
rwgk: I didn't want to pack to much into this PR that's not really related to the move. It was more a "minimal while I'm at it" effort to achieve that the output can be used as-is. Previously I had to manually curate, which already led to one accident.

Is it OK to leave more work for later?

rwgk:
I'm not rerunning the CI:
* `pre-commit.ci` exercises the `pyproject.toml` changes.
* All other changes only affect comments and documentation.

toolshed/build_pathfinder_dlls.py
leofang: Yeah fine with me!

.github/workflows/test-wheel-linux.yml
leofang: Later the pathfinder will need to check the driver version, so the status quo is fine.

toolshed/build_pathfinder_dlls.py
```diff
@@ -39,12 +40,25 @@
...
 
+def is_suppressed_dll(libname, dll):
+    if libname == "cudart":
+        if dll.startswith("cudart32_"):
+            return True
+    elif libname == "nvrtc":
+        if dll.endswith(".alt.dll"):
+            return True
+        if dll.startswith("nvrtc-builtins"):
+            return True
```
leofang, line 55: For NVRTC wheel to work on Windows, this DLL is needed, e.g.
https://github.com/NVIDIA/nvmath-python/blob/e082fedb46a57072e63d3b8f0aae3b451eeddcbb/nvmath/_utils.py#L139

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py
```diff
@@ -58,4 +61,11 @@ def load_nvidia_dynamic_library(libname: str) -> LoadedDL:
     Raises:
         RuntimeError: If the library cannot be found or loaded
     """
-    return _load_nvidia_dynamic_library_no_cache(libname)
+    pointer_size_bits = struct.calcsize("P") * 8
+    if pointer_size_bits != 64:
```
leofang, line 65: FYI usually we can check this with `sys.maxsize > 2**32`:
https://docs.python.org/3/library/platform.html#platform.architecture

leofang (approved):
LGTM, thanks for pushing this forward Ralf!

btw I skipped reviewing commit c0509beebe88bcbd223c7e8fe817e6a1d4fb7266 from #751, unfortunately it's too much for me to review at once. In the future I'd appreciate we follow what we did in the past and have a series of smaller PRs with a focused scope.

toolshed/build_pathfinder_dlls.py
leofang: Maybe I just forget what "suppressed DLLs" means...

leofang:
btw @rwgk could you create an issue to track the need to add docs for pathfinder? We need a Sphinx-rendered docs.

cuda_pathfinder/tests/conftest.py
kkraus14: I wasn't aware this was a pytest conftest plugin hook which was the key piece of info. This is fine.

rwgk:
> btw I skipped reviewing commit [c0509be](https://github.com/NVIDIA/cuda-python/commit/c0509beebe88bcbd223c7e8fe817e6a1d4fb7266) from #751, unfortunately it's too much for me to review at once. In the future I'd appreciate we follow what we did in the past and have a series of smaller PRs with a focused scope.

Just to explain, that's my preferred work style, too, but two things led to the current situation:

* #751 removes the `pywin32` dependency (f32823083a043d7459106fa2d9c945a42a682b55), this seemed to be very valuable to have from the get-go.

* GitHub (OSS world in general) doesn't have an easy to use "stacked PRs" feature. Juggling multiple PRs is time consuming.

PR #724, opened 2025-06-24
Test against CTK 12.0 + earliest driver

leofang:
## Description

Try to reproduce nvbugs 5348750 in the CI. In theory we need r550 (12.4) driver. The "earliest" label is r535.

## Checklist


leofang:
The cuda.bindings test was not properly skipped and now shadows the real issue (`cuda.core` tests are not run) lol

cuda_bindings/tests/test_cuda.py
```diff
@@ -980,12 +980,14 @@ def test_all_CUresult_codes():
     assert num_good >= 76  # CTK 11.0.3_450.51.06
 
 
+@pytest.mark.skipif(driverVersionLessThan(12030), reason="Driver too old for cuKernelGetName")
```
kkraus14, line 983: Instead of skipping should we use the driver version check to ensure that we're throwing an exception as expected?

leofang:
@oleksandr-pavlyk @cryos As I hoped I poked a hole in our CI coverage, the error that QA reported can be reproduced in the CI now.

cuda_bindings/tests/test_cuda.py
leofang: We did get the exception as observed with the commit ef896a91. Right now I am using the same skip approach applied to other tests in the same file to get to the cuda.core test gap (main purpose of this PR). In a separate PR we could turn this into a pytest fixture that 
1. Run the test if the driver version is sufficient
2. Otherwise, ensure an exception is raised

leofang: Tracked in #728

leofang:
Sasha took over in #725. Closing.

PR #733, opened 2025-06-27
Add note about compute sanitizer for ccx.Device()

ZzEeKkAa:
## Description

Add note on using `ccx.Device()` with `compute-sanitizer`.

## Checklist
- [ ] n/a New or existing tests cover these changes.

### DIFF
--- a/cuda_core/cuda/core/experimental/_device.py
+++ b/cuda_core/cuda/core/experimental/_device.py
@@ -955,6 +955,13 @@ class Device:
     ----
     Will not initialize the GPU.
 
+    Note
+    ----
+    Internal implementation depends on the CUDA Driver API error handling, that
+    may cause ``compute-sanitizer`` to fail. In order to avoid this, you can use
+    ``--report-api-errors=no`` `compute-sanitizer argument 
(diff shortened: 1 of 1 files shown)


leofang:
Thanks, Yevhenii! I think this impacts to all cuda.core objects, not just `Device`. I am thinking we should create a new "Tips & Tricks" page like [what we did for cuda.bindings](https://nvidia.github.io/cuda-python/cuda-bindings/latest/tips_and_tricks.html), and move this note there. WDYT?

ZzEeKkAa:
Yeah, we can do that. I'm wondering how to do it the way, so if user face this problem, they will navigate to the correct place...

leofang:
What do you mean?

cpcloud (approved):
This seems fine to merge after fixing up conflicts.

Extending this to other objects/writing up tips and tricks can be done in a follow-up.

cpcloud:
@ZzEeKkAa Can you fix conflicts so we can merge?

leofang:
I suggest we close this and work on #1223. It does not make sense to add such a note to the device class.

PR #741, opened 2025-07-01
Restore option to run testing without cupy installed.

rwgk:
## Description
cuda_core is generally set up to run most tests without `cupy` installed.

PR #717 added one unconditional `cupy` import. This PR adopts the approach in `tests/test_utils.py` for the import in `tests/test_launcher.py`, and copies the `SKIPPED` message from `tests/example_tests/utils.py`.


rwgk:
Thanks @leofang

Issue #642, opened 2025-05-18
`cuda.bindings` build time on Linux becomes ~2x slower


leofang:
This is a new situation and is different from #602 (and the timings there can be used as a reference). I suspect this has to do with Cython 3.1.0, based on my eyeball-bisecting the recent build times taken in the main branch:
https://github.com/NVIDIA/cuda-python/actions/workflows/ci.yml?query=branch%3Amain++
I suspect this starts after Cython 3.1.0 was released, the first slowness observation:
- https://github.com/NVIDIA/cuda-python/actions/runs/14920621978
in contrast with the previous run:
- https://github.com/NVIDIA/cuda-python/actions/runs/14889184430

What I don't understand is Windows seems to be unaffected (and continues to be on the order of 4-5 mins)...

leofang:
Confirmed in #640 that the build time can be restored by pinning Cython<3.1.0

leofang:
I can reproduce the 2x slowdown locally. But when I tried to chase after the root cause, I probably picked the wrong module to look at. `cuda.bindings.cydriver` compilation is slower with newer Cython, but its compilation/linking times are insignificant compared to the overall build time. Anyway I attach the .cpp file generated by Cython 3.0.12 (old) and 3.1.2 (new): [cydriver.zip](https://github.com/user-attachments/files/20856694/cydriver.zip)

Below is the commands generated by setuptools with `pip install --no-build-isolation --no-deps -v .`:
- compile
```
[...]

Issue #602, opened 2025-05-04
Investigate `cuda.bindings`'s long build time on linux-aarch64


leofang:
I noticed recently that it is consistently almost 2x slower on linux-aarch64 (13-14 mins) than on linux-64 (7-8 mins). FWIW it builds blazing fast on win-64 (~4 mins), so linux-aarch64 is always the last to finish building. This can be worsen by the fact that we have fewer Arm64 runners than x86-64 runners, so if the queue is full the wait time is really really long.

leofang:
This doesn't seem to be an issue anymore, likely because we now have sccache. Closing.

PR #743, opened 2025-07-02
Fix CI build-time parallelism + support the same env var in `cuda.core`

leofang:
## Description

- Not sure if this contributes to either #602 or #642, but I noticed when reviewing #723 that the CI env var is set incorrectly, so it seems we're building with single threads
- I apply the same env var support to cuda.core, since we added more Cython modules (#709) 

## Checklist

### DIFF
--- a/cuda_core/setup.py
+++ b/cuda_core/setup.py
@@ -9,7 +9,6 @@ from Cython.Build import cythonize
 from setuptools import Extension, setup
 from setuptools.command.build_ext import build_ext as _build_ext
 
-
 nthreads = int(os.environ.get("CUDA_PYTHON_PARALLEL_LEVEL", os.cpu_count() // 2))
 
 


leofang:
pre-commit.ci autofix

leofang:
I eyeballed the build time change (against the main branch) and observed noticeable speedup!

cuda_core/setup.py
```diff
@@ -9,6 +9,9 @@
 from setuptools import Extension, setup
 from setuptools.command.build_ext import build_ext as _build_ext
 
+nthreads = int(os.environ.get("CUDA_PYTHON_PARALLEL_LEVEL", os.cpu_count() // 2))
```
rwgk, line 12: Minor drive-by comment: I'm used to seeing `nproc` for this in various context. Using that here might make this more intuitive.

leofang: The name `nthreads` was taken from cuda.bindings. The only difference here is cuda.core already has picked a default # threads.

leofang:
Let me admin-merge here. I need this in to study if we can relax Cython upper bound.

Issue #750, opened 2025-07-09
[FEA]: private constructor prevents derived classes of cythonized Stream, Event


carterbox:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

cuda.core

### Is your feature request related to a problem? Please describe.

In nvmath-python we have test the correctness of our code by checking how many times we have called `Event.sync()`. We used to do this by monkeypatching the `Event` class, so that it would update a global variable whenever called. This is no longer possible because `Event` is an immutable cython class in cuda-core 0.3.1, so it is unpatchable.

The alternative approach for us would be to create a derived class `LoggedSyncEvent` which overwrites the `sync()` method with our modified method. However, this is not possible because `Event` subverts the conventional constructor method `__init__()` and instead implements `_init()`. The problem with this custom constructor is that it will always return an `Event` even in derived classes (as opposed to `__init__()` which cython knows how to handle properly for class inheritance).  

### Describe the solution you'd like

Use the standard `__init__()` method for class constructors instead of using engineering controls to preventing users from calling the constructor directly by hiding the constructor in `_init()`. Instead  just use procedural controls by just mentioning in the documentation that calling `__init__()` is deprecated/unsupported/unstable.

For example:

```
diff --git a/cuda_core/cuda/core/experimental/_event.pyx b/cuda_core/cuda/core/experimental/_event.pyx
index 94496985..76bb7b96 100644
--- a/cuda_core/cuda/core/experimental/_event.pyx
+++ b/cuda_core/cuda/core/experimental/_event.pyx
@@ -84,12 +84,7 @@ cdef class Event:
         int _device_id
         object _ctx_handle

-    def __init__(self, *args, **kwargs):
-        raise RuntimeError("Event objects cannot be instantiated directly. Please use Stream APIs (record).")
-
-    @classmethod
-    def _init(cls, device_id: int, ctx_handle: Context, options=None):
-        cdef Event self = Event.__new__(Event)
+    def __init__(self, device_id: int, ctx_handle: Context, options=None):
         cdef EventOptions opts = check_or_create_options(EventOptions, options, "Event options")
         flags = 0x0
         self._timing_disabled = False
@@ -106,7 +101,6 @@ cdef class Event:
         raise_if_driver_error(err)
         self._device_id = device_id
[...]

PR #752, opened 2025-07-10
Construct instance of appropriate class in Event._init

shwina:
## Description

Closes https://github.com/NVIDIA/cuda-python/issues/750

closes 

## Checklist

### DIFF
--- a/cuda_core/cuda/core/experimental/_event.pyx
+++ b/cuda_core/cuda/core/experimental/_event.pyx
@@ -89,7 +89,7 @@ cdef class Event:
 
     @classmethod
     def _init(cls, device_id: int, ctx_handle: Context, options=None):
-        cdef Event self = Event.__new__(Event)
+        cdef Event self = Event.__new__(cls)
         cdef EventOptions opts = check_or_create_options(EventOptions, options, "Event options")
         flags = 0x0
         self._timing_disabled = False


cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -89,7 +89,7 @@ cdef class Event:
 
     @classmethod
     def _init(cls, device_id: int, ctx_handle: Context, options=None):
-        cdef Event self = Event.__new__(Event)
+        cdef Event self = Event.__new__(cls)
```
shwina, line 92: This ensures that a `cls` instance is produced here, which could be different from `Event` if `cls` is a subclass.

carterbox: I think this works! I was trying:
```python
cdef cls self = Event.__new__(cls)
```
and getting compile time errors.

kkraus14, line 92: Can we add a test for a derived class of Event to ensure that we don't accidentally regress on what this enables?

shwina: We can go about this in two ways:

* Write a test that invokes `MyEvent.__new__` directly and test that it produces a `MyEvent` (easy)
* Write a test that monkeypatches `Event` to `MyEvent`, and test that constructing an event via `Stream` or `Device` produces a `MyEvent` rather than an `Event` (more complex, but representative of how users can actually use subclasses). 

Ultimately, (2) calls down to (1) but in a way that's an implementation detail. Any preferences?

kkraus14: Lets take the easy path and we can always revisit in the future if we see value in doing so.

shwina:
Apologies for the delay. I added a test for `Event` subclassing and made similar changes for `Stream`.

cuda_core/cuda/core/experimental/_event.pyx
shwina: Done

leofang:
/ok to test [3e264db](https://github.com/NVIDIA/cuda-python/pull/752/commits/3e264dbe4870eb0420abaa70de991c21325998ba)

rwgk (approved):

carterbox (approved):

PR #753, opened 2025-07-11
`SPDX-FileCopyrightText` cleanup

### LINKED ISSUE #635: Update copyright text in each file to use SPDX file tags
>> If we're doing a big cleanup here, we should also move the copyright statements to use SPDX copyright statements instead. I.e. `SPDX-FileCopyrightText`.
>
> Sounds good. I could try to squeeze that in later (another weekend). I wouldn't know what exactly to do though. Could you please describe the exact requirements in an issue?

_Originally posted by @rwgk in https://github.com/NVIDIA/cuda-python/issues/625#issuecomment-2877071193_

Currently, in our files we have a comment that has a Copyright statement, i.e. https://github.com/NVIDIA/cuda-python/blob/99ffe141f7d2efb5afa2fbfb71c137848afde35c/cuda_bindings/pyproject.toml#L1. We should migrate these to use SPDX file tags instead: https://spdx.github.io/spdx-spec/v2.3/file-tags/#h2-file-tags-format

This would then look like:
```
SPDX-FileCopyrightText: Copyright 2023-2024 NVIDIA Corporation.  All rights reserved. 
```

SPDX file tags will be better supported by tooling related to checking things like licensing and copyrights.

rwgk:
## Description
Closes #635

This was done with Cursor.

I tried 03e2d89c5c68905ac3e0476437d543d3c0dda4a1 but ran into problems (and out of time). That switch is better left for a separate PR.


cuda_bindings/cuda/bindings/_internal/nvjitlink.pxd
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2024, NVIDIA CORPORATION & AFFILIATES. ALL RIGHTS RESERVED.
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
```
kkraus14, line 3: Not a blocker, but in some files there is an empty line between these and other files don't have the empty line. It would be nice to make it consistent and probably amenable to cursor fixing it.

kkraus14 (approved):

cuda_bindings/cuda/bindings/_internal/nvjitlink.pxd
rwgk: Thanks! I'll merge this and fix the empty lines separately.

Issue #754, opened 2025-07-11
[FEA]: Provide separate source distributions for individual packages


mgorny:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

General cuda-python

### Is your feature request related to a problem? Please describe.

Currently, the `cuda-bindings`, `cuda-core` and `cuda-python` packages on PyPI publish only binary wheels. Source code is available from archives attached to releases, but it includes the whole repository and therefore does not integrate well with standard PEP 517 build process (read: you can't just `pip install cuda-python-v12.9.0.tar.gz`).

### Describe the solution you'd like

While we can work around it, as downstream packagers we would find it much cleaner if you could provide [source distributions for each project separately](https://packaging.python.org/en/latest/discussions/downstream-packaging/#provide-complete-source-distributions). If you don't want to upload them to PyPI, attaching to releases would also work.

### Additional context

I understand that there are currently some cross-package dependencies that require using the monorepo archive. However, I think they could be solved cleanly with help of some symlinks that turn into included subdirectories in sdist. If you agree, I can experiment a bit and prepare a pull request if I get something neat working.

leofang:
Hi Michal, thanks for reaching out. Uploading sdists to PyPI, as you might have heard, has some issues that could trap users. Therefore,
> If you don't want to upload them to PyPI, attaching to releases would also work.

this seems like a good idea. We do plan to attach more release artifacts to GitHub Releases (#742).

mgorny:
Thanks for getting back to me. I've created #758 to fix the inability to build wheel from `cuda-python` sdist right now.

PR #758, opened 2025-07-15
cuda-python: make it possible to build wheel from sdist

mgorny:
## Description

Refactor cuda-bindings version lookup to make it possible to build cuda-python from a source distribution.  Rather than using the version file straight from `cuda_bindings` tree, use a `_version.py` symlink. This is equivalent inside the git repository, and is replaced by the actual file in the source distribution, resulting in a standalone archive.

Related to issue #754.

## Checklist

(the changes don't affect package code, so no changes required)

### DIFF
--- a/cuda_python/MANIFEST.in
+++ b/cuda_python/MANIFEST.in
@@ -1,0 +1,4 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
+
+include _version.py
--- a/cuda_python/_version.py
+++ b/cuda_python/_version.py
@@ -1,0 +1,1 @@
+../cuda_bindings/cuda/bindings/_version.py
--- a/cuda_python/setup.py
+++ b/cuda_python/setup.py
@@ -5,10 +5,10 @@
 import ast
 from setuptools import setup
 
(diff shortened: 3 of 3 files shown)


kkraus14:
@mgorny I sincerely apologize, but we need to close this pull request since we are unable to take non-NVIDIA contributions on files with the `SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE` identifier at this time per guidance from NVIDIA legal.

We appreciate your willingness to jump in and provide a fix here, where we'll do our best to prioritize landing a fix for #754 to unblock your downstream work.

mgorny:
Sorry, just to clarify: did I do something wrong, or is it currently impossible to add new files? I've only slapped the license there because that's what `pre-commit` asked for — to the best of my knowledge, that file does not qualify for copyright protection.

kkraus14:
> Sorry, just to clarify: did I do something wrong, or is it currently impossible to add new files? I've only slapped the license there because that's what `pre-commit` asked for — to the best of my knowledge, that file does not qualify for copyright protection.

You did not do anything wrong in any way. Thank you for sharing your use case and jumping in with a contribution.

As of now, the `cuda_bindings` and `cuda_python` subdirectories (for the `cuda-bindings` package and `cuda-python` metapackage respectively) are licensed under a source available license: https://github.com/NVIDIA/cuda-python/blob/main/cuda_python/LICENSE

Per the guidance of NVIDIA legal, we are unable to accept contributions from non-NVIDIA employees on the sections of the codebase with this license at this time.

PR #767, opened 2025-07-19
[pathfinder] Fix #763 (load `libnvJitLink.so.12` from conda, not `/usr/local/cuda`)

### LINKED ISSUE #763: [BUG][pathfinder]: nvJitLink loaded from system library instead of conda
### Is this a duplicate?

### Type of Bug

Silent Failure

### Component

cuda.bindings

### Describe the bug

@leofang wrote:

I noticed that on the system that I’m on, which has a system CTK 12.3 and I installed CTK 12.9 from conda, the pathfinder from either cuda.bindings 12.9.0 or cuda.pathfinder 1.0.0 would pick up nvJitLink 12.3 (the system one) instead of 12.9 (the conda one), which is not following the behavior that we documented.
I suspect that the logic in `_load_nvidia_dynamic_library_no_cache` might be wrong:
```
    # Find the library path
    found = _find_nvidia_dynamic_library(libname)
    if found.abs_path is None:
        loaded = load_with_system_search(libname, found.lib_searched_for)
```
because in `_find_nvidia_dynamic_library` we always do this on Linux:
```
    self.lib_searched_for = f"lib{libname}.so"
```
meaning we don’t search with the full soname (`libnvJitLink.so.12`), but the symlink name (`libnvJitLink.so`), which conda does not provide if we only install the libnvjitlink package and not the libnvjitlink-dev package.
Therefore, the `load_with_system_search` function behaves wrong because we fed it a wrong soname.

### How to Reproduce

I think a simple reproducer would be:
* launch a vanilla Ubuntu container
* Install miniforge and then create a new conda env with only cuda-pathfinder (from pip) and libnvjitlink (from conda-forge) installed.
* Run the pathfinder
```
from cuda import pathfinder
pathfinder.load_nvidia_dynamic_lib("nvJitLink")
```

### Expected behavior

The conda `.so` should be found.

### Operating System

_No response_

### nvidia-smi output

_No response_

rwgk:
Closes #763

This fixes the issue, but corresponding new tests are missing.

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
+++ b/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
@@ -7,6 +7,7 @@ import os
 from typing import Optional
 
 from cuda.pathfinder._dynamic_libs.load_dl_common import LoadedDL
+from cuda.pathfinder._dynamic_libs.supported_nvidia_libs import SUPPORTED_LINUX_SONAMES
 
 CDLL_MODE = os.RTLD_NOW | os.RTLD_GLOBAL
 
@@ -82,12 +83,11 @@ def check_if_already_loaded_from_elsewhere(libname: str) -> Optional[LoadedDL]:
     return None
 
 
-def load_with_system_search(libname: str, soname: str) -> Optional[LoadedDL]:
+def load_with_system_search(libname: str) -> Optional[LoadedDL]:
     """Try to load a library using system search paths.
 
     Args:
         libname: The name of the library to load
-        soname: The soname to search for
 
     Returns:
         A LoadedDL object if successful, None if the library cannot be loaded
@@ -95,14 +95,18 @@ def load_with_system_search(libname: str, soname: str) -> Optional[LoadedDL]:
     Raises:
         RuntimeError: If the library is loaded but no expected symbol is found
     """
-    try:
-        handle = ctypes.CDLL(soname, CDLL_MODE)
-        abs_path = abs_path_for_dynamic_library(libname, handle)
-        if abs_path is None:
-            raise RuntimeError(f"No expected symbol for {libname=!r}")
-        return LoadedDL(abs_path, False, handle._handle)
-    except OSError:
-        return None
+    candidate_sonames = list(SUPPORTED_LINUX_SONAMES.get(libname, ()))
+    candidate_sonames.append(f"{libname}.so")
+    for soname in candidate_sonames:
+        try:
+            handle = ctypes.CDLL(soname, CDLL_MODE)
+            abs_path = abs_path_for_dynamic_library(libname, handle)
+            if abs_path is None:
+                raise RuntimeError(f"No expected symbol for {libname=!r}")
+            return LoadedDL(abs_path, False, handle._handle)
+        except OSError:
+            pass
+    return None
 
 
 def load_with_abs_path(_libname: str, found_path: str) -> LoadedDL:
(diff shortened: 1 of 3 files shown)


leofang:
/ok to test [2107b96](https://github.com/NVIDIA/cuda-python/pull/767/commits/2107b9689720f363f4c9c030d14dbaa811d5106a)

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
```diff
@@ -82,27 +83,30 @@ def check_if_already_loaded_from_elsewhere(libname: str) -> Optional[LoadedDL]:
...
-    try:
-        handle = ctypes.CDLL(soname, CDLL_MODE)
-        abs_path = abs_path_for_dynamic_library(libname, handle)
-        if abs_path is None:
-            raise RuntimeError(f"No expected symbol for {libname=!r}")
-        return LoadedDL(abs_path, False, handle._handle)
-    except OSError:
-        return None
+    candidate_sonames = list(SUPPORTED_LINUX_SONAMES.get(libname, ()))
+    candidate_sonames.append(f"{libname}.so")
```
leofang: Following our naming convention this should be:
```suggestion
    candidate_sonames.append(f"lib{libname}.so")
```

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
```diff
@@ -82,27 +83,30 @@ def check_if_already_loaded_from_elsewhere(libname: str) -> Optional[LoadedDL]:
...
+    for soname in candidate_sonames:
+        try:
+            handle = ctypes.CDLL(soname, CDLL_MODE)
+            abs_path = abs_path_for_dynamic_library(libname, handle)
+            if abs_path is None:
+                raise RuntimeError(f"No expected symbol for {libname=!r}")
+            return LoadedDL(abs_path, False, handle._handle)
+        except OSError:
```
leofang, line 107: Shouldn't we also catch `RuntimeError` here to ensure we complete the loop over `candidate_sonames`?

leofang (changes requested):

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
rwgk: Oops, done: commit 1d5808bf2e0e0a7c635243205f6589609e65b5f5

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
rwgk: Good question — this is a rough edge I already stumbled over while working on the `unreleased` branch (I had to update `EXPECTED_LIB_SYMBOLS`).

Thinking while writing:

If the library was successfully loaded but none of the expected symbols was found, we're essentially in a "dirty state":

The fact that the library was loaded successfully is very likely to interfere with other loading attempts. For example, the `SONAME` could be the same, which means the loader will skip the actual loading but still report success. Then a RuntimeError is raised again when none of the expected symbols are found.

My tentative conclusion: we should definitely not catch the `RuntimeError`. On the contrary, we should actually raise an exception that does not inherit from `Exception` (only from `BaseException`) if we’ve successfully loaded a library but none of the expected symbols are found — so that we don’t silently slide into a dirty state.

WDYT?

I'm not sure we should do that in this PR, since it's a different, more general problem.

But if the suggested change (inherit from `BaseException`) makes sense to you, I’d like to implement it before the next `cuda-pathfinder` release (i.e., asap).

leofang: Thanks for explanation, Ralf. I agree this is an edge case that is unlikely to hit by normal users. Once a library is loaded, even if we find it is a corrupted/bad binary, it is very hard to unload and we should raise asap. The current treatment looks fine to me. Let's just test/merge.

leofang (approved):

leofang:
/ok to test [1d5808b](https://github.com/NVIDIA/cuda-python/pull/767/commits/1d5808bf2e0e0a7c635243205f6589609e65b5f5)

PR #773, opened 2025-07-22
Add `get_cuda_native_handle`

leofang:
## Description

closes #564

The performance is comparable to the existing `int()` approach (O(10) ns):
```python
In [1]: from cuda.bindings import driver

In [3]: s = driver.CUstream(1234)

In [4]: int(s)
Out[4]: 1234

In [6]: from cuda.bindings.utils import get_cuda_native_handle

In [7]: get_cuda_native_handle(s)
Out[7]: 1234

In [8]: %timeit int(s)
44.7 ns ± 1.34 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [9]: %timeit get_cuda_native_handle(s)
95.1 ns ± 1.06 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

## Checklist

- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #780, opened 2025-07-24
_cuda_bindings_redirector.py fix (made necessary by PR #773)

rwgk:
Avoid premature import of `cuda.bindings` in redirector to fix startup noise (see [here](https://github.com/NVIDIA/cuda-python/pull/773#issuecomment-3111270268)).

Previously, the `_cuda_bindings_redirector.py` file imported `cuda.bindings` at interpreter startup to ensure the `cuda` namespace was initialized. However, this triggered early loading of Cython modules that depend on `win32api`, which may not be available yet due to `.pth` file processing order.

This commit replaces `import cuda.bindings` with `import cuda`, which is sufficient to ensure the `cuda` namespace is in `sys.modules`. The version redirect via `cuda.__version__` is preserved by assigning a custom `LazyCudaModule` type to `sys.modules["cuda"]`.

As a result, `cuda.__version__` continues to work (with a deprecation warning), while avoiding startup errors when `win32api` is not yet importable.

### DIFF
--- a/cuda_bindings/site-packages/_cuda_bindings_redirector.py
+++ b/cuda_bindings/site-packages/_cuda_bindings_redirector.py
@@ -1,9 +1,30 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
 
-import os
-import site
+import sys
+from types import ModuleType
 
-_bindings_root = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
-if _bindings_root not in site.getsitepackages():
-    site.addsitedir(_bindings_root)
+
+# Make sure 'cuda' is importable as a namespace package
+import cuda
+
+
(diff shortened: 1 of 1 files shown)


rwgk:
@leofang for visibility

rwgk:
@kkraus14 @oleksandr-pavlyk @leofang 

I asked ChatGPT about `import cuda` vs `import cuda.bindings`, it's towards the end in this conversation:

* https://chatgpt.com/share/6882ae60-5a60-8008-ba99-2ed2acb65e89

Look for "Why might the original author have used import cuda.bindings"

The conclusion there:

> In short: it was probably defensive but over-eager. A simpler `import cuda` achieves the intended effect without the cascading side effects.

All tests pass, so I think this super easy fix is all we need.

rwgk:
I forgot to add: I downloaded the CI logs and searched for `win32api`, there are no matches anymore.

leofang (approved):
Thanks, Ralf. The fix looks reasonable to me. Unfortunately I can't reliably reproduce the issue, so if it fixes on your side it should be good enough, and let's merge and monitor the situation.

PR #781, opened 2025-07-25
Make `Device.set_current()` faster

### LINKED ISSUE #739: `Device.set_current()` is slow
Follow-up of #658.

With `cuda-python` 12.6.0 (this is an important case because this is a Cython-based cudart re-implementation):
```python
In [4]: %timeit cudart.cudaSetDevice(0)
355 ns ± 1.69 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```
For comparison, this is CuPy
```python
In [6]: %timeit cp.cuda.runtime.setDevice(0)
154 ns ± 1.67 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```
With `cuda-bindings` 12.9.0 (the re-implementation is replaced by the statically linked cudart)
```python
In [4]: %timeit runtime.cudaSetDevice(0)
167 ns ± 0.367 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```
and this is `cuda-core` on the main branch + `cuda-bindings` 12.9.0
```python
In [5]: %timeit dev.set_current()
1.84 μs ± 16 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```
I think we should find a way to at least revive and reuse the old re-implementation.

leofang:
## Description

closes #739 

Before this PR:
```
In [4]: %timeit dev.set_current()
1.4 μs ± 10.9 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

With this PR:
```python
In [4]: %timeit dev.set_current()
374 ns ± 1.17 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

We want to minimize the number of calls to CUDA APIs, which add up to the overheads (several hundreds of nanoseconds per call). This is achieved by combining two changes:
1. We lazily retain a reference to the primary context of each device 
   - The only scenario where this could fail is if someone calls `cudaDeviceReset()` somewhere, in which case the pointers to the primary contests would be invalidated and become dangling. However, the CUDA team strongly discourages using this API (it does not solve any real issue, e.g. restoring from sticky errors), and it is impossible to get things right especially in the Python land if it is called. Too many CUDA-related objects floating around.
2. We unconditionally set the needed primary context to current without extra checks

## Checklist


leofang:
/ok to test [92d24ce](https://github.com/NVIDIA/cuda-python/pull/781/commits/92d24ce1a78e37353ca0002937b02db80066526a)

shwina (approved):

leofang:
/ok to test [b9b3c87](https://github.com/NVIDIA/cuda-python/pull/781/commits/b9b3c873caf7429669959fb28d0cbb22c220f29b)

PR #795, opened 2025-08-04
Remove trampoline modules

leofang:
## Description

Close #161. Merge after #792.

## Checklist

- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #807, opened 2025-08-06
[Backport] Remove trampoline modules

leofang:
## Description

Manual backport of #795 

## Checklist


bdice:
Just for awareness, I unexpectedly saw these warnings the other day while trying out LLM NIMs for the first time. Not to dissuade you -- I think you can justify the change based on the length of deprecation -- but I think there are a lot of users who haven't migrated and this could have wider repercussions than anticipated.

Example of where I saw this: https://gist.github.com/bdice/1de237b48370a29ef9d5c777ac2acc5e#file-gpt-oss-20b-log-L55

leofang:
Based on an offline discussion with Bradley/Keith based on the above observation, we decided to not purge the trampolines in 12.9.x.

Issue #776, opened 2025-07-23
[FEA]: Support for finding libmathdx through pathfinder


ZzEeKkAa:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

cuda.pathfinder

### Is your feature request related to a problem? Please describe.

There are multiple ways to install libmathdx (compiled mathdx) so it can be used in numba and warp:
- conda package: https://anaconda.org/conda-forge/libmathdx
- pypi: https://pypi.org/project/nvidia-libmathdx-cu12
- pypi: https://pypi.org/project/nvidia-libmathdx-cu13
- tar.gz: https://developer.nvidia.com/cublasdx-downloads

### Describe the solution you'd like

`cuda.pathfinder.load_nvidia_dynamic_lib("mathdx")`

rwgk:
I tried this using PR #834:

```
conda create -y -n libmathdx python=3.12
conda activate libmathdx
conda install conda-forge::libmathdx
cd cuda_pathfinder
pip install .
python
Python 3.12.11 | packaged by conda-forge | (main, Jun  4 2025, 14:45:31) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from cuda.pathfinder import load_nvidia_dynamic_lib
>>> load_nvidia_dynamic_lib("mathdx")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/rgrossekunst/forked/cuda-python/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py", line 71, in load_nvidia_dynamic_lib
    return _load_lib_no_cache(libname)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rgrossekunst/forked/cuda-python/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py", line 44, in _load_lib_no_cache
    found.raise_if_abs_path_is_None()
  File "/home/rgrossekunst/forked/cuda-python/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/find_nvidia_dynamic_lib.py", line 194, in raise_if_abs_path_is_None
[...]

PR #834, opened 2025-08-13
[pathfinder] `RTLD_DI_LINKMAP`-based new implementation of `abs_path_for_dynamic_library()`

### LINKED ISSUE #833: [BUG]: KeyError in load_dl_linux.py:47 EXPECTED_LIB_SYMBOLS[libname]
### Is this a duplicate?

### Type of Bug

Runtime Error

### Component

cuda.pathfinder

### Describe the bug

Under certain circumstances code responsible for loading ctk libraries triggers key error for a library that is not part of ctk

### How to Reproduce

```bash
python -m venv .venv
pip install cuda-pathfinder
export CUDA_HOME=/usr/local/cuda12.9
export LD_LIBRARY_PATH=$CUDA_HOME/lib64:$LD_LIBRARY_PATH
wget -q https://developer.download.nvidia.com/compute/cudss/redist/libcudss/linux-x86_64/libcudss-linux-x86_64-0.5.0.16_cuda12-archive.tar.xz
tar -xvf libcudss-linux-x86_64-0.5.0.16_cuda12-archive.tar.xz
export LD_LIBRARY_PATH=$(pwd)/libcudss-linux-x86_64-0.5.0.16_cuda12-archive/lib:$LD_LIBRARY_PATH
```

### Expected behavior

Find library with respect to `LD_LIBRARY_PATH` or at least return proper runtime error that library was not found (not preferable, since with `nvidia-cudss-cu12` it loads library properly).

### Operating System

_No response_

### nvidia-smi output

_No response_

rwgk:
Closes #833

Also a step towards resolving #776 — please see comments there.

This PR eliminates `supported_nvidia_libs.EXPECTED_LIB_SYMBOLS` entirely, which is a major simplification in its own right.

Piggy-backed: Minor fix (commit d13ad8e5359e524c9024feaad93a7ca8011c74b9) and cleanup (commit e716f1cb2f2a5039a7d46d341aa09b9072ac8ff7).

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
+++ b/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
@@ -37,6 +37,7 @@ LIBDL.dlerror.restype = ctypes.c_char_p
 
 # First appeared in 2004-era glibc. Universally correct on Linux for all practical purposes.
 RTLD_DI_LINKMAP = 2
+RTLD_DI_ORIGIN = 6
 
 
 class _LinkMapLNameView(ctypes.Structure):
@@ -80,7 +81,7 @@ def _dl_last_error() -> Optional[str]:
     return msg_bytes.decode("utf-8", "surrogateescape")
 
 
-def abs_path_for_dynamic_library(libname: str, handle: ctypes.CDLL) -> str:
+def l_name_for_dynamic_library(libname: str, handle: ctypes.CDLL) -> str:
     lm_view = ctypes.POINTER(_LinkMapLNameView)()
     rc = LIBDL.dlinfo(ctypes.c_void_p(handle._handle), RTLD_DI_LINKMAP, ctypes.byref(lm_view))
     if rc != 0:
@@ -93,12 +94,31 @@ def abs_path_for_dynamic_library(libname: str, handle: ctypes.CDLL) -> str:
     if not l_name_bytes:
         raise OSError(f"dlinfo returned empty link_map->l_name for {libname=!r}")
 
-    # Won't raise, and preserves undecodable bytes round-trip
-    path = os.fsdecode(l_name_bytes)  # filesystem encoding + surrogateescape
+    path = os.fsdecode(l_name_bytes)
     if not path:
-        raise OSError(f"dlinfo returned empty path string for {libname=!r}")
+        raise OSError(f"dlinfo returned empty l_name string for {libname=!r}")
 
     return path
+
+
+def l_origin_for_dynamic_library(libname: str, handle: ctypes.CDLL) -> str:
+    l_origin_buf = ctypes.create_string_buffer(4096)
+    rc = LIBDL.dlinfo(ctypes.c_void_p(handle._handle), RTLD_DI_ORIGIN, l_origin_buf)
+    if rc != 0:
+        err = _dl_last_error()
+        raise OSError(f"dlinfo failed for {libname=!r} (rc={rc})" + (f": {err}" if err else ""))
+
+    path = os.fsdecode(l_origin_buf.value)
+    if not path:
+        raise OSError(f"dlinfo returned empty l_origin string for {libname=!r}")
+
+    return path
+
+
+def abs_path_for_dynamic_library(libname: str, handle: ctypes.CDLL) -> str:
+    l_name = l_name_for_dynamic_library(libname, handle)
+    l_origin = l_origin_for_dynamic_library(libname, handle)
+    return os.path.join(l_origin, os.path.basename(l_name))
 
 
 def get_candidate_sonames(libname: str) -> list[str]:


cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
```diff
@@ -14,48 +14,40 @@
 
 LIBDL_PATH = ctypes.util.find_library("dl") or "libdl.so.2"
 LIBDL = ctypes.CDLL(LIBDL_PATH)
-LIBDL.dladdr.argtypes = [ctypes.c_void_p, ctypes.c_void_p]
-LIBDL.dladdr.restype = ctypes.c_int
+LIBDL.dlinfo.argtypes = [ctypes.c_void_p, ctypes.c_int, ctypes.c_void_p]
+LIBDL.dlinfo.restype = ctypes.c_int
```
leofang: I am a bit nervous about us using either `dladdr` or `dlinfo` since none of them is POSIX-compliant. Since `dladdr` has been used, switching to `dlinfo` does not make it more "non-compliant", so it's still a step forward. We just have to see if it runs everywhere...

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
```diff
@@ -14,48 +14,40 @@
...
 
 
-class DlInfo(ctypes.Structure):
-    """Structure used by dladdr to return information about a loaded symbol."""
+# First appeared in 2004-era glibc. Universally correct on Linux for all practical purposes.
+RTLD_DI_LINKMAP = 2
```
leofang, line 39: Q: Can't we just use `RTLD_DI_ORIGIN` and get the path back without populating the full struct (since we just need the path)?

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
```diff
@@ -14,48 +14,40 @@
...
+    lm_ptr = ctypes.POINTER(LinkMap)()
+    rc = LIBDL.dlinfo(ctypes.c_void_p(handle._handle), RTLD_DI_LINKMAP, ctypes.byref(lm_ptr))
+    if rc == 0 and lm_ptr and lm_ptr.contents.l_name:
+        path: str = lm_ptr.contents.l_name.decode()
+        if path:
+            return path
 
-    Returns:
-        The absolute path to the library file, or None if no expected symbol is found
+    raise OSError(f"abs_path_for_dynamic_library failed for {libname=!r}")
```
leofang: One potential improvement for error handling is we subsequently call `dlerror` to get the err msg if -1 is returned by `dladdr`/`dlinfo`.

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
rwgk: > since none of them is POSIX-compliant

Do you think the ChatGPT conclusion below is reasonable?


My ChatGPT prompt:
________
This makes me want to ask:

Does CUDA run on any systems that are not Linux or Windows?

(Windows dynamic loading is handled separately. I.e. basically I'm only asking: does CUDA run on any Unix system that is not Linux?)

Do you know?
________

ChatGPT said:

Thought for 19s

Short answer: No. Today CUDA is officially supported on Linux (various distros/arches, including aarch64/Jetson) and Windows (including WSL2 Linux environments on Windows). There’s no current macOS runtime support, and no official support for other Unix-like OSes (e.g., *BSD). 

So if your scope is “Linux only,” it’s reasonable to lean on glibc-specific behavior (e.g., dlinfo)—CUDA isn’t targeting musl/Alpine or non-Linux Unix platforms.
_______

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
rwgk: I reduced the struct to just what we need (commit be2078207342031ddbff71679aeca032173490d9).

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
rwgk: Done, also commit be2078207342031ddbff71679aeca032173490d9.

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
leofang: I am unaware that ctypes allows for incomplete struct definitions. It's FFI based so I assume it needs to know the full struct layout/size for ABI-based operations. This is nerve wrecking...

kkraus14: I think because we aren't ever allocating a struct we never need to know the full size of it and only need to know the needed member offsets.

kkraus14 (approved):
LGTM

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
rwgk: Thanks @kkraus14 — yes, exactly.

That’s what I had in mind when I pared down the table.

The use of ctypes here is in line with what we do elsewhere (e.g., `cuda.cccl`). The big difference is that we don’t own `struct link_map`.

So last night (TH TZ) I started drilling down to be sure we’re on solid ground. Two questions I wanted to answer:

**Q1:** How stable has `struct link_map` been over the years?

**Q2:** Is paring down the `ctypes` definition for `struct link_map` valid for our purposes?

Side remark: ChatGPT 5 is vastly more thorough than ChatGPT 4, at the cost of longer waits (I’ve seen >2 min of "thinking" for one question; with ChatGPT 4 most responses came back almost instantly).

**Re Q1:** I started with (AI-assisted) legwork, looking at the `struct link_map` source code. This is what I found:

```
git clone https://sourceware.org/git/glibc.git
```

```
commit d66e34cd423425c348bcc83df127dd19711b0b9a
Date:   Tue May 2 06:35:55 1995 +0000
```

```
git show d66e34cd423425c348bcc83df127dd19711b0b9a:elf/link.h
```
________
```
struct link_map
  {
    /* These first few members are part of the protocol with the debugger.
       This is the same format used in SVR4.  */

    Elf32_Addr l_addr;		/* Base address shared object is loaded at.  */
    char *l_name;		/* Absolute file name object was found in.  */
    Elf32_Dyn *l_ld;		/* Dynamic section of the shared object.  */
    struct link_map *l_next, *l_prev; /* Chain of loaded objects.  */

    /* All following members are internal to the dynamic linker.
       They may change without notice.  */

    const char *l_libname;	/* Name requested (before search).  */
    Elf32_Dyn *l_info[DT_NUM];	/* Indexed pointers to dynamic section.  */
    const Elf32_Phdr *l_phdr;	/* Pointer to program header table in core.  */
    Elf32_Word l_phnum;		/* Number of program header entries.  */

    /* Symbol hash table.  */
    Elf32_Word l_nbuckets;
    const Elf32_Word *l_buckets, *l_chain;

    unsigned int l_opencount;	/* Reference count for dlopen/dlclose.  */
    enum			/* Where this object came from.  */
      {
	lt_executable,		/* The main executable program.  */
	lt_interpreter,		/* The interpreter: the dynamic linker.  */
	lt_library,		/* Library needed by main executable.  */
	lt_loaded,		/* Extra run-time loaded shared object.  */
      } l_type:2;
    unsigned int l_deps_loaded:1; /* Nonzero if DT_NEEDED items loaded.  */
    unsigned int l_relocated:1;	/* Nonzero if object's relocations done.  */
    unsigned int l_init_called:1; /* Nonzero if DT_INIT function called.  */
    unsigned int l_init_running:1; /* Nonzero while DT_INIT function runs.  */
  };
```
________

```
commit 2642002380aafb71a1d3b569b6d7ebeab3284816
Date:   Wed Jan 1 10:14:45 2025 -0800
```

```
git show 2642002380aafb71a1d3b569b6d7ebeab3284816:elf/link.h
```

```
struct link_map
  {
    /* These first few members are part of the protocol with the debugger.
       This is the same format used in SVR4.  */

    ElfW(Addr) l_addr;		/* Difference between the address in the ELF
				   file and the addresses in memory.  */
    char *l_name;		/* Absolute file name object was found in.  */
    ElfW(Dyn) *l_ld;		/* Dynamic section of the shared object.  */
    struct link_map *l_next, *l_prev; /* Chain of loaded objects.  */
  };
```
________

To make this more concrete, on my Ubuntu 24.04 workstation I get:

```
cc -E /usr/include/link.h
```

```
typedef uint64_t Elf64_Addr;

struct link_map
  {
    Elf64_Addr l_addr;
    char *l_name;
    Elf64_Dyn *l_ld;
    struct link_map *l_next, *l_prev;
  };
```
________

Conclusion: **The "first few members" of `struct link_map` have been stable for ~30 years.**

--------

**Re Q2:** Because we want to use `ctypes` (and not compile a small C helper), we need to rely on pointer arithmetic:

* `dlinfo()` gives is us a `struct link_map` pointer.

* We add `offsetof(struct link_map, l_name)` and dereference the resulting pointer.

That's exactly what the previous `ctyptes`-based `lm_ptr.contents.l_name` did. However, ChatGPT suggested an alternative that makes the pointer arithmetic explicit. That is now implemented in

commit ee5d05e908242d75aaffb3ab722af5a08f322e2e

With that, `LinkMap` is gone completely.

While I was at it, I also ensured the `.decode()`s in `load_dl_linux.py` will not raise `UnicodeDecodeError` (same commit, plus commit 97f1c360b4b0e4ac7b17359ed86e8c20181ad4d6), and I added a test to ensure the `abs_path` is actually absolute, and does in fact exist (commit 5b2012ab0558326f6c318acd91d3d5cd9c2945e1). That should make it obvious if our assumptions about `struct link_map` ever get violated.

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
```diff
@@ -39,34 +39,34 @@ def _load_libdl() -> ctypes.CDLL:
...
         raise OSError(f"dlinfo failed for {libname=!r} (rc={rc})" + (f": {err}" if err else ""))
-    if not lm_ptr:
+    if not lm_ptr.value:
         raise OSError(f"dlinfo returned NULL link_map pointer for {libname=!r}")
-    name = lm_ptr.contents.l_name
-    if not name:
+
+    # l_name is the second field, right after l_addr (both pointer-sized)
+    l_name_field_addr = lm_ptr.value + ctypes.sizeof(ctypes.c_void_p)
+    l_name_addr = ctypes.c_void_p.from_address(l_name_field_addr).value
```
kkraus14: I personally find this less intuitive than using the minimal struct definition

rwgk: I'm very ambivalent myself, but thought it'll be useful to have the alternative fully worked out here, for reference.

I'll change back to the more intuitive `LinkMap` approach, because I believe it'll be easier to maintain in the future. I'll work in comments to explain.

rwgk: Done: commit 49a54c30943743931a1ef212b2826e0122f6e422

kkraus14 (approved):

leofang:
Breaking out from [this thread](https://github.com/NVIDIA/cuda-python/pull/834#discussion_r2273751618) since it's getting a bit long and I want to leave another comment there to keep it focused.
> Q: Can't we just use `RTLD_DI_ORIGIN` and get the path back

Turns out we can. 
```python
RTLD_DI_ORIGIN = 6
mylib = ctypes.CDLL(...)

path_buf = ctypes.create_string_buffer(4096)
rc = LIBDL.dlinfo(ctypes.c_void_p(mylib._handle), RTLD_DI_ORIGIN, path_buf)
if rc != 0:
    err = _dl_last_error()
    raise OSError(f"dlinfo failed for {libname=!r} (rc={rc})" + (f": {err}" if err else ""))

print(f"origin = {path_buf.value.decode()}")
```
But for all practical use cases that the pathfinder covers, `origin` is exactly the same as `os.path.dirname(path_from_link_map)`, where `path_from_link_map` is the DSO path retrieved via querying `RTLD_DI_LINKMAP` as done in the current PR. 

The only edge case where `RTLD_DI_LINKMAP` would not work is if `mylib` loads a DSO via a relative path. In this case, only `RTLD_DI_ORIGIN` would return the correct abs path to the directory containing the DSO. But as mentioned this is a case that the pathfinder does not cover by design. 

In fact, it might be a good idea to expand the new `validate_abs_path` test to use this query? It's not a blocker, ofc.

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
leofang: >> I am unaware that ctypes allows for incomplete struct definitions. It's FFI based so I assume it needs to know the full struct layout/size for ABI-based operations. This is nerve wrecking...

Sorry, I left my comment at the airport in a hurry... Keith is correct in saying
> I think because we aren't ever allocating a struct we never need to know the full size of it and only need to know the needed member offsets.

but it only answers my concern partially. Ralf's reply also did not hit the target precisely (though it implies what I am sharing below).

With hindsight everything is obvious, but I think the key point I missed is that `dlinfo()` takes **`struct link_map**`** when querying `RTLD_DI_LINKMAP`, i.e. it's a **pointer to pointer** to struct, not just a pointer to struct. So `dlinfo()` copies an internal pointer address to the struct (which was allocated by the system, not by us as the user) for us to view the struct data, and so we never have to allocate buffer for the entire struct ourselves (which was what I was worrying about with a partial `LinkMap`), just a buffer for the pointer which is properly allocated by `ctypes.POINTER()`. The pointer arithmetic / offset consideration comes after this issue, which is not a concern to me.

Should it take a pointer to struct (`struct link_map*`) instead, the whole story would have been very different because then we're responsible for allocating a buffer of the struct size for `dlinfo` to populate. [My comment below](https://github.com/NVIDIA/cuda-python/pull/834#issuecomment-3193298409) shows one such example (with the `RTLD_DI_ORIGIN` query, which requires us to allocate 4096 bytes (=`PATH_MAX`)).

leofang (approved):
Thanks, Ralf, LGTM too!

rwgk:
Thanks @leofang , I hadn't seen `RTLD_DI_ORIGIN` before.

To be sure that we can rely on it being available, I looked up when it was added in the glibc repo:

```
commit 45e4762c2c7d04cd51213735c2b440c94cdcf28a
Date:   Sat Mar 15 23:16:22 2003 +0000
```

It's defined in `dlfcn/dlfcn.h` with this comment (which hasn't changed at all for 22+ years!):

```
    /* Treat ARG as `char *', and store there the directory name used to
       expand $ORIGIN in this shared object's dependency file names.  */
    RTLD_DI_ORIGIN = 6,
```

That makes me think it's only a "directory name", and we'd still have to get the filename from somewhere else.

> The only edge case where `RTLD_DI_LINKMAP` would not work is if `mylib` loads a DSO via a relative path. In this case, only `RTLD_DI_ORIGIN` would return the correct abs path to the directory containing the DSO. But as mentioned this is a case that the pathfinder does not cover by design.

I'll play with this later tonight, to verify that a relative path is in fact not expanded to a full absolute path. If that's true, I'll add a comment in `load_dl_linux.py` before merging this PR.

rwgk:
> That makes me think it's only a "directory name",

Confirmed.

> a relative path is in fact not expanded to a full absolute path.

Also confirmed.

Putting 1 + 1 together in commit 66c5869aeec32cb85b6dbf059d016905da048235:

`l_origin + basename(l_name) → abs_path`

@leofang as you said, currently pathfinder does not really need this, but it's easy enough to make it complete.

rwgk:
For completeness, here is how I confirmed the `dlinfo` behavior:

* https://github.com/rwgk/rwgk_config/blob/9ae0510debd5ce3b66c547328c8a99ec6d308149/bin/try_dlopen.py

```
$ cd /lib
$ try_dlopen.py x86_64-linux-gnu/libcufftMp.so.11
dlopen succeeded: 'x86_64-linux-gnu/libcufftMp.so.11'
  l_name: 'x86_64-linux-gnu/libcufftMp.so.11'
  l_origin: '/usr/lib/x86_64-linux-gnu'
  abs_path: '/usr/lib/x86_64-linux-gnu/libcufftMp.so.11'
```

```
$ cd
$ try_dlopen.py libnvJitLink.so.13
dlopen succeeded: 'libnvJitLink.so.13'
  l_name: '/usr/local/cuda/lib64/libnvJitLink.so.13'
  l_origin: '/usr/local/cuda/lib64'
  abs_path: '/usr/local/cuda/lib64/libnvJitLink.so.13'
```

leofang:
> That makes me think it's only a "directory name", and we'd still have to get the filename from somewhere else.

Yes. As mentioned earlier:

>> `origin` is exactly the same as `os.path.dirname(path_from_link_map)`, where `path_from_link_map` is the DSO path retrieved via querying `RTLD_DI_LINKMAP` as done in the current PR.

`dirname()` is the (abs) directory path. And we already have the filename.

Issue #608, opened 2025-05-04
Ensure `cuda_static`'s directory is passed to `setuptools`


leofang:
Follow-up of https://github.com/NVIDIA/cuda-python/pull/517. We need to inform `setuptools` to pass `-L` so that `cudart_static` can show up in the compiler search paths without extra hacks. The logic should eventually get `library_dirs` updated: https://github.com/NVIDIA/cuda-python/blob/bd770e14e0fb960b64862743eafc88394fb6f14a/cuda_bindings/setup.py#L285

leofang:
Depending on the installation source, we'd need either `$CUDA_PATH/lib` or `$CUDA_PATH/lib64`...

Issue #46, opened 2023-06-07
Splayed layout support


jakirkham:
Currently `cuda-python` relies on all binaries (like `nvcc`), all headers, and all libraries to live in a single directory (specified by [`$CUDA_HOME` or similar]( https://github.com/NVIDIA/cuda-python/blob/6782a648bc728cdf55e280eb20631748b8c15627/setup.py#L25-L29 )).

However there are use cases (like cross-compilation, as with conda-build) where the build tools may live in one location (and perform builds on that architecture) whereas the headers and libraries may live in a different location (and target a different architecture). In this case not everything lives in `$CUDA_HOME`.

It would be helpful to have a way of specifying where these different components come from. Here are some options:

1. Check `$NVCC` for the `nvcc` location
2. Use `$CUDA_BIN` (if specified) to get build tool directory
3. Support a list of directories in `$CUDA_HOME`
4. ?

Maybe there are other reasonable options worth considering?

robertmaynard:
Another option is to support `$CUDA_HOME` being a list of directories. So for header searching cuda-python would append `include` to each entry in `$CUDA_HOME`.

jakirkham:
cc @bdice

vyasr:
Would it make sense to standardize a tool that uses the path to nvcc to parse information out of nvcc.profile? Something that could be used by anyone, not just cuda-python?

leofang:
Sorry for lack of response. Is this still an issue? It seems we have been able to build cuda-python on conda-forge?

jakirkham:
Yep it is still an issue

We [workaround it in conda-forge]( https://github.com/conda-forge/cuda-python-feedstock/blob/8a93f7938014ec4d2f546281e94fd4053763e862/recipe/meta.yaml#L29-L32 )

Through improvements on the packaging side, we have minimized the workarounds needed ( for example: https://github.com/conda-forge/cuda-python-feedstock/pull/75 ), but we are unable to eliminate them without this fix

leofang:
@jakirkham I am sorry but I don't get it. In any case we need a way to specify where the CUDA headers are, and setting `CUDA_HOME` is the right solution that's been working fine. Are there anything specific to the splayed layout that we have not addressed (either in CUDA Python or in conda-forge)?

jakirkham:
[...]

PR #837, opened 2025-08-14
Adds path for the CUDA static library based on CUDA_HOME

Andy-Jost:
Both CUDA_HOME (or CUDA_PATH) and LIBRARY_PATH are required to be set in the environment during installation. There was already a check for CUDA_HOME in setup.py, but no check for LIBRARY_PATH. This change adds a check to make sure LIBRARY_PATH is set.

When LIBRARY_PATH is not set, the installation fails at link time and indicates that cudart_static could not be found.

### DIFF
--- a/cuda_bindings/setup.py
+++ b/cuda_bindings/setup.py
@@ -31,6 +31,7 @@ if not CUDA_HOME:
     raise RuntimeError("Environment variable CUDA_HOME or CUDA_PATH is not set")
 
 CUDA_HOME = CUDA_HOME.split(os.pathsep)
+
 if os.environ.get("PARALLEL_LEVEL") is not None:
     warn(
         "Environment variable PARALLEL_LEVEL is deprecated. Use CUDA_PYTHON_PARALLEL_LEVEL instead",
@@ -238,6 +239,8 @@ include_dirs = [
     os.path.dirname(sysconfig.get_path("include")),
 ] + include_path_list
 library_dirs = [sysconfig.get_path("platlib"), os.path.join(os.sys.prefix, "lib")]
+cudalib_subdir = r"lib\x64" if sys.platform == "win32" else "lib64"
+library_dirs.extend(os.path.join(prefix, cudalib_subdir) for prefix in CUDA_HOME)
 
 extra_compile_args = []
 extra_cythonize_kwargs = {}
--- a/cuda_bindings/setup.py
+++ b/cuda_bindings/setup.py
@@ -239,8 +239,8 @@ include_dirs = [
     os.path.dirname(sysconfig.get_path("include")),
 ] + include_path_list
 library_dirs = [sysconfig.get_path("platlib"), os.path.join(os.sys.prefix, "lib")]
-cudalib_subdir = r"lib\x64" if sys.platform == "win32" else "lib64"
-library_dirs.extend(os.path.join(prefix, cudalib_subdir) for prefix in CUDA_HOME)
+cudalib_subdirs = [r"lib\x64"] if sys.platform == "win32" else ["lib64", "lib"]
+library_dirs.extend(os.path.join(prefix, subdir) for prefix in CUDA_HOME for subdir in cudalib_subdirs)
 
 extra_compile_args = []
 extra_cythonize_kwargs = {}
--- a/.github/workflows/build-wheel.yml
+++ b/.github/workflows/build-wheel.yml
@@ -157,11 +157,9 @@ jobs:
           # CIBW mounts the host filesystem under /host
           CIBW_ENVIRONMENT_LINUX: >
             CUDA_PATH=/host/${{ env.CUDA_PATH }}
-            LIBRARY_PATH=/host/${{ env.CUDA_PATH }}/lib
             CUDA_PYTHON_PARALLEL_LEVEL=${{ env.CUDA_PYTHON_PARALLEL_LEVEL }}
           CIBW_ENVIRONMENT_WINDOWS: >
             CUDA_HOME="$(cygpath -w ${{ env.CUDA_PATH }})"
-            LIB="${CUDA_HOME}\\lib\\x64;${LIB}"
             CUDA_PYTHON_PARALLEL_LEVEL=${{ env.CUDA_PYTHON_PARALLEL_LEVEL }}
           CIBW_BEFORE_BUILD_WINDOWS: "pip install delvewheel"
           CIBW_REPAIR_WHEEL_COMMAND_WINDOWS: "delvewheel repair --namespace-pkg cuda -w {dest_dir} {wheel}"


kkraus14:
@Andy-Jost to trigger CI to run you need to comment "/ok to test". More info here: https://docs.gha-runners.nvidia.com/platform/apps/copy-pr-bot/faqs

Andy-Jost:
This breaks Windows builds because LIBRARY_PATH is specific to Linux. While I could place a guard around the check, it's probably better to close this PR without merging.

rwgk:
> This breaks Windows builds because LIBRARY_PATH is specific to Linux. While I could place a guard around the check, it's probably better to close this PR without merging.

I think it'd be lovely to complete this PR, I stumbled over this many times.

Under Windows, we need `LIB` to be set.

This is a good way to test for Windows (we use that in `cuda_pathfinder`):

```
IS_WINDOWS = sys.platform == "win32"
```

Andy-Jost:
> > This breaks Windows builds because LIBRARY_PATH is specific to Linux. While I could place a guard around the check, it's probably better to close this PR without merging.
> 
> I think it'd be lovely to complete this PR, I stumbled over this many times.
> 
> Under Windows, we need `LIB` to be set.
> 
> This is a good way to test for Windows (we use that in `cuda_pathfinder`):
> 
> ```
> IS_WINDOWS = sys.platform == "win32"
> ```

I am looking into a solution that works for Windows as well as Linux.

leofang:
Probably just update `library_dirs` that's passed to `setup()` in setup.py?

Andy-Jost:
The latest changes eliminate the need to set either LIBRARY_PATH or LIB and should resolve issue #608.

Where can I find the source for the installation instructions? I.e., https://nvidia.github.io/cuda-python/cuda-bindings/latest/install.html.

Those can be updated now.

Andy-Jost:
> Probably just update `library_dirs` that's passed to `setup()` in setup.py?

This did the trick.

cuda_bindings/setup.py
```diff
@@ -31,6 +31,7 @@
     raise RuntimeError("Environment variable CUDA_HOME or CUDA_PATH is not set")
 
 CUDA_HOME = CUDA_HOME.split(os.pathsep)
```
rwgk, line 33: Oh, surprise, I didn't notice this before ... Question for @leofang: Do we need to support multiple directories in general (e.g. in pathfinder)?

cuda_bindings/setup.py
```diff
@@ -238,6 +239,10 @@ def generate_output(infile, local):
     os.path.dirname(sysconfig.get_path("include")),
 ] + include_path_list
 library_dirs = [sysconfig.get_path("platlib"), os.path.join(os.sys.prefix, "lib")]
+if sys.platform == "win32":
+    library_dirs.extend(os.path.join(prefix, r"lib\x64") for prefix in CUDA_HOME)
+elif sys.platform == "linux":
+    library_dirs.extend(os.path.join(prefix, "lib64") for prefix in CUDA_HOME)
```
rwgk: Maybe
```
cuda_lib = r"lib\x64" if sys.platform == "win32" else "lib64"
library_dirs.extend(os.path.join(prefix, cuda_lib) for prefix in CUDA_HOME)
```
?

leofang:
@Andy-Jost the diff became very massive, could it be a rebasing issue?

cuda_bindings/setup.py
leofang: This came from [the old days](https://github.com/NVIDIA/cuda-python/issues/46#issuecomment-2190389863) when we had to deal with the way we wrote our conda recipe. By having `{{ compiler('cuda') }}` in the `build` section of the recipe, we had two places that could have CUDA headers (the build environment and the host environment). However, building any of the projects in this repo does not really need NVCC, so I've killed this usage and simplified the recipe in https://github.com/conda-forge/cuda-python-feedstock/pull/118 for this and several other reasons. This is now moot.

cuda_bindings/setup.py
Andy-Jost: Good suggestion. BTW, can I assume that the only platforms are "win32" and "linux"?

leofang:
@Andy-Jost could you resolve the conflicts before running tests?

Andy-Jost:
> @Andy-Jost could you resolve the conflicts before running tests?

OK, I'll take a look. I'm still confused about how the PR got so large.

Andy-Jost:
I think the history is clean now. Please let me know if you see any problem.

cuda_bindings/setup.py
leofang: yes

leofang:
Thanks, @Andy-Jost! The diff is clean now.

I think to test this properly in the CI, we should delete these two lines
https://github.com/NVIDIA/cuda-python/blob/b400e345b141183d2c1d54b2e69ef14400c162c7/.github/workflows/build-wheel.yml#L160
https://github.com/NVIDIA/cuda-python/blob/b400e345b141183d2c1d54b2e69ef14400c162c7/.github/workflows/build-wheel.yml#L164
Not sure if I miss any. I hope not 🤔

cuda_bindings/setup.py
```diff
@@ -238,6 +239,8 @@ def generate_output(infile, local):
     os.path.dirname(sysconfig.get_path("include")),
 ] + include_path_list
 library_dirs = [sysconfig.get_path("platlib"), os.path.join(os.sys.prefix, "lib")]
+cudalib_subdir = r"lib\x64" if sys.platform == "win32" else "lib64"
```
leofang: On the Linux side, one annoying thing is that depending on how CUDA is installed, this could be `lib` or `lib64`:
```shell
$ ls -l /usr/local/cuda-13.0/lib64/libcudart_static.a 
-rw-r--r-- 1 root root 1361104 Aug  5 09:32 /usr/local/cuda-13.0/lib64/libcudart_static.a
$ ls -l /usr/local/cuda-13.0/targets/x86_64-linux/lib/libcudart_static.a 
-rw-r--r-- 1 root root 1361104 Aug  5 09:32 /usr/local/cuda-13.0/targets/x86_64-linux/lib/libcudart_static.a
``` 

https://github.com/conda-forge/cuda-cudart-feedstock/blob/38fb08899691fa53fce33960ef37d48ab58584b0/recipe/meta.yaml#L220

https://github.com/conda-forge/cuda-cudart-feedstock/blob/38fb08899691fa53fce33960ef37d48ab58584b0/recipe/meta.yaml#L251

We probably should support both.

Andy-Jost:
I see a failure in pre-commit.ci due to "ruff format" creating diffs. I think I get the general idea, but I don't see how to get more information (e.g., see the diffs, or run the ruff format locally). I can probably fix the formatting issue by guessing, but in general how can I get the details?

leofang (approved):

leofang:
@copilot please backport this PR to the 12.9.x branch. (Since this PR touched some files in the .github folder, the backport bot does not work.)

Issue #449, opened 2025-02-14
Perf: Reduce `StridedMemoryView` construction time


leofang:
Currently it takes 3.4 - 3.45 us (depending on stream-ordering or not) to create a memory view object:
```python
In [4]: x = cp.empty((23, 4))

In [7]: %timeit s = StridedMemoryView(x, -1)
3.4 μs ± 8.7 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [8]: %timeit s = StridedMemoryView(x, 1)
3.45 μs ± 14.6 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```
which could be a bit expensive in a tight loop. We should try to reduce it down to 1 us or O(100) ns if possible.

cc @shwina for vis

leofang:
cc @NaderAlAwar for vis

leofang:
It'd be interesting to see if lazily populating the attributes of `StridedMemoryView` could help reduce the construction time. A lot of times we don't need the full information, just pointer and shape/strides.

PR #838, opened 2025-08-14
Improve #449: Improve StridedMemoryView creation time

mdboom:
Two changes:

1. Refactor the versioned/non-versioned paths to reduce the number of branches.
2. Create shape and strides tuples using Python/C API

The second change has a significant impact:

- Before: 1.23 us +- 0.06 us
- After: 1.02 us +- 0.06 us

<details>
<summary>Measured using this pyperf benchmark, based on the one in #449</summary>

```
import pyperf
import time


import cupy as cp


inner_loops = 100_000


def bench_strided_memory_view(loops, x):
    range_it = range(loops * inner_loops)
    t0 = time.perf_counter()

    for _ in range_it:
        s = StridedMemoryView(x, -1)

    return time.perf_counter() - t0


runner = pyperf.Runner()
x = cp.empty((23, 4))
runner.bench_time_func('StridedMemoryView', bench_strided_memory_view, x, inner_loops=inner_loops)
```
</details>

## Description

Improves #449

## Checklist

### DIFF
--- a/cuda_core/cuda/core/experimental/_memoryview.pyx
+++ b/cuda_core/cuda/core/experimental/_memoryview.pyx
@@ -213,7 +213,6 @@ cdef StridedMemoryView view_as_dlpack(obj, stream_ptr, view=None):
     cdef StridedMemoryView buf = StridedMemoryView() if view is None else view
     buf.ptr = <intptr_t>(dl_tensor.data)
     
-    # Construct shape and strides tuples using the Python/C API for speed
     buf.shape = cuda_utils.carray_int64_t_to_tuple(dl_tensor.shape, dl_tensor.ndim)
     if dl_tensor.strides:
         buf.strides = cuda_utils.carray_int64_t_to_tuple(dl_tensor.strides, dl_tensor.ndim)
--- a/cuda_core/cuda/core/experimental/_utils/cuda_utils.pxd
+++ b/cuda_core/cuda/core/experimental/_utils/cuda_utils.pxd
@@ -14,6 +14,7 @@ cpdef check_or_create_options(type cls, options, str options_description=*, bint
 
 
 cdef inline tuple carray_int64_t_to_tuple(libc.stdint.int64_t *ptr, int length):
+    # Construct shape and strides tuples using the Python/C API for speed
     result = cpython.PyTuple_New(length)
     for i in range(length):
         cpython.PyTuple_SET_ITEM(result, i, cpython.PyLong_FromLongLong(ptr[i]))


cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -308,8 +312,10 @@ cdef StridedMemoryView view_as_cai(obj, stream_ptr, view=None):
     buf.dtype = numpy.dtype(cai_data["typestr"])
     buf.strides = cai_data.get("strides")
     if buf.strides is not None:
-        # convert to counts
-        buf.strides = tuple(s // buf.dtype.itemsize for s in buf.strides)
+        # convert to counts, using the Python/C API for speed
+        for i in range(len(buf.strides)):
+            cpython.PyTuple_SET_ITEM(buf.strides, i,
+                cpython.PyLong_FromLong(buf.strides[i] // buf.dtype.itemsize))
```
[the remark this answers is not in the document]
mdboom: Good point, Copilot.  Fixed.

kkraus14 (approved):

kkraus14:
The changes here LGTM! To trigger CI you need to comment "/ok to test", see here for more info: https://docs.gha-runners.nvidia.com/platform/apps/copy-pr-bot/faqs/

Most people are using the `StridedMemoryView` class as a way to generically handle input arrays that support `dlpack` and/or `__cuda_array_interface__` right before kernel launches which is typically interfacing to some underlying C++ code. Currently when we initialize the class we eagerly populate python attributes for things like the `ptr`, `shape`, `strides`, `dtype`, etc. There's a non-trivial cost to this where I wonder if we should instead have C types and then lazily initialize the Python types as requested.

leofang:
> Currently when we initialize the class we eagerly populate python attributes for things like the `ptr`, `shape`, `strides`, `dtype`, etc. There's a non-trivial cost to this where I wonder if we should instead have C types and then lazily initialize the Python types as requested.

Yes, I have been thinking these attributes should be populated lazily, which is possible because we still hold a reference to the dlpack capsule. @mdboom would you like to address this in a follow-up PR?

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -184,48 +184,52 @@ cdef StridedMemoryView view_as_dlpack(obj, stream_ptr, view=None):
...
         is_readonly = False
+        used_name = DLPACK_TENSOR_USED_NAME
+    else:
+        assert False
 
     cdef StridedMemoryView buf = StridedMemoryView() if view is None else view
     buf.ptr = <intptr_t>(dl_tensor.data)
-    buf.shape = tuple(int(dl_tensor.shape[i]) for i in range(dl_tensor.ndim))
+    
+    # Construct shape and strides tuples using the Python/C API for speed
```
leofang: Q: Does this mean Cython is not generating efficient code for us? Could it be possible that I was using the generator form to construct a tuple and Cython is bad at this syntax?

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -184,48 +184,52 @@ cdef StridedMemoryView view_as_dlpack(obj, stream_ptr, view=None):
...
+    buf.shape = cpython.PyTuple_New(dl_tensor.ndim)
+    for i in range(dl_tensor.ndim):
+        cpython.PyTuple_SET_ITEM(buf.shape, i, cpython.PyLong_FromLong(dl_tensor.shape[i]))
```
leofang: I don't think we can use `PyLong_FromLong` because both shape and strides are int64_t
https://github.com/dmlc/dlpack/blob/7f393bbb86a0ddd71fde3e700fc2affa5cdce72d/include/dlpack/dlpack.h#L257-L263
we probably need to replace it with `PyLong_FromLongLong`

leofang (changes requested):

cuda_core/cuda/core/experimental/_memoryview.pyx
leofang: Another question for my understanding: Both `PyTuple_SET_ITEM` and `PyTuple_Set_Item` steal the reference of the object `o` (the Python int that we convert from DLPack shape/stride members). Doesn't it mean that we should increment its refcount before calling setitem?

mdboom:
> > Currently when we initialize the class we eagerly populate python attributes for things like the `ptr`, `shape`, `strides`, `dtype`, etc. There's a non-trivial cost to this where I wonder if we should instead have C types and then lazily initialize the Python types as requested.
> 
> Yes, I have been thinking these attributes should be populated lazily, which is possible because we still hold a reference to the dlpack capsule. @mdboom would you like to address this in a follow-up PR?

Yeah, that makes sense.  I'll do that in a separate PR and keep this just to the tuple-construction improvements.

cuda_core/cuda/core/experimental/_memoryview.pyx
mdboom: Yeah, this is pretty inefficient.  The generator causes Cython to make a separate C function which is then called through the Python iterator machinery, which is a lot less efficient than a C for loop (mainly because it makes a bunch of C function calls through C function pointers).  Secondly, since the length of the tuple isn't known in advance, it causes it to be realloc'ed at least 3 times:

- initial size of 0
- overallocate to 10
- truncate to the exact value of 2

It might be possible to write a cdef function to convert a C array pointer + known length into a tuple in this faster way, which would hide this use of the CPython API here.  I'll look at that and measure it.

cuda_core/cuda/core/experimental/_memoryview.pyx
mdboom: Good point about using `LongLong`.  I will update.

As for reference counting, we don't need to increment here because we are "giving" the reference to the int to the tuple, and we don't need to hold on to the int ourselves to do anything else with it.

leofang: I see, so either constructor would return the obj with refcount 1 (for some reason I was thinking it's 0), and the ownership of this ref is transferred to the tuple.

cuda_core/cuda/core/experimental/_memoryview.pyx
mdboom: Abstracting this out to a `cdef inline` function seems to have no measurable overhead and definitely makes this more readable, so I've updated this PR to do that.

PR #847, opened 2025-08-18
Fix #449: Delay construction of Python attributes

### LINKED ISSUE #449: Perf: Reduce `StridedMemoryView` construction time

mdboom:
## Description

For the `StridedMemoryView`, this delays construction of the Python types `shape`, `strides` and `dtype` and converts the other simpler objects to regular Cython cdefs.

This has an approximately 18% speedup on `StridedMemoryView` object creation:

  - Baseline:  Mean +- std dev: 993 ns +- 41 ns
  - Deferred Python object creation: Mean +- std dev: 821 ns +- 32 ns

This also adds tests to cover the `__cuda_array_interface__` code path.

closes 

## Checklist

### DIFF
--- a/cuda_core/cuda/core/experimental/_memoryview.pyx
+++ b/cuda_core/cuda/core/experimental/_memoryview.pyx
@@ -91,7 +91,8 @@ cdef class StridedMemoryView:
         
     # Memoized properties
     cdef tuple _shape
-    cdef object _strides
+    cdef tuple _strides
+    cdef bint _strides_init  # Has the strides tuple been init'ed?
     cdef object _dtype
 
     def __init__(self, obj=None, stream_ptr=None):
@@ -121,20 +122,24 @@ cdef class StridedMemoryView:
     @property
     def strides(self) -> Optional[tuple[int]]:
         cdef int itemsize
-        if self._strides is None and self.exporting_obj is not None:
-            if self.dl_tensor != NULL:
-                if self.dl_tensor.strides:
-                    self._strides = cuda_utils.carray_int64_t_to_tuple(
-                        self.dl_tensor.strides, 
-                        self.dl_tensor.ndim
-                    )
-            else:
-                strides = self.metadata.get("strides")
-                if strides is not None:
-                    itemsize = self.dtype.itemsize
-                    self._strides = cpython.PyTuple_New(len(strides))
-                    for i in range(len(strides)):
-                        cpython.PyTuple_SET_ITEM(self._strides, i, strides[i] // itemsize)
+        if self._strides_init is False:
+            if self.exporting_obj is not None:
+                if self.dl_tensor != NULL:
+                    if self.dl_tensor.strides:
+                        self._strides = cuda_utils.carray_int64_t_to_tuple(
+                            self.dl_tensor.strides, 
+                            self.dl_tensor.ndim
+                        )
+                else:
+                    strides = self.metadata.get("strides")
+                    if strides is not None:
+                        itemsize = self.dtype.itemsize
+                        self._strides = cpython.PyTuple_New(len(strides))
+                        for i in range(len(strides)):
+                            cpython.PyTuple_SET_ITEM(
+                                self._strides, i, strides[i] // itemsize
+                            )
+            self._strides_init = True
         return self._strides 
 
     @property


cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -51,7 +50,7 @@ cdef class StridedMemoryView:
         Pointer to the tensor buffer (as a Python `int`).
     shape : tuple
         Shape of the tensor.
-    strides : tuple
+    strides : Optional[tuple]
```
mdboom, line 53: Just updating docs to match current behavior (not a change in behavior)

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -92,9 +94,50 @@ cdef class StridedMemoryView:
...
+                return cuda_utils.carray_int64_t_to_tuple(
+                    self.dl_tensor.shape, 
+                    self.dl_tensor.ndim
+                )
+            else:
+                return self.exporting_obj.__cuda_array_interface__["shape"]
+        return ()
+
+    @property
+    def strides(self) -> Optional[tuple[int]]:
```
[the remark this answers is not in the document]
mdboom: We only need to support Python 3.9 and later, so this syntax should be fine.

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -92,9 +94,50 @@ cdef class StridedMemoryView:
             else:
                 view_as_cai(obj, stream_ptr, self)
         else:
-            # default construct
             pass
 
+    @property
+    def shape(self) -> tuple[int]:
```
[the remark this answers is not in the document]
mdboom: We only need to support Python 3.9 and later, so this syntax should be fine.

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -92,9 +94,50 @@ cdef class StridedMemoryView:
...
+                        self.dl_tensor.strides, 
+                        self.dl_tensor.ndim
+                    )
+            else:
+                strides = self.exporting_obj.__cuda_array_interface__.get("strides")
+                if strides is not None:
+                    itemsize = self.dtype.itemsize
+                    result = cpython.PyTuple_New(len(strides))
+                    for i in range(len(strides)):
+                        cpython.PyTuple_SET_ITEM(result, i, strides[i] // itemsize)
```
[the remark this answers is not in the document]
mdboom: Not correct.

cuda_core/cuda/core/experimental/_memoryview.pyx
kkraus14, line 109: Is it worth caching this and other properties?

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -70,19 +69,22 @@ cdef class StridedMemoryView:
...
-    shape: tuple = None
-    strides: tuple = None  # in counts, not bytes
-    dtype: numpy.dtype = None
-    device_id: int = None  # -1 for CPU
-    is_device_accessible: bool = None
-    readonly: bool = None
-    exporting_obj: Any = None
+    cdef readonly:
+        intptr_t ptr
+        intptr_t device_id
```
leofang:
```suggestion
        int device_id
```

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -70,19 +69,22 @@ cdef class StridedMemoryView:
     obj : Any
         Any objects that supports either DLPack (up to v1.0) or CUDA Array
         Interface (v3).
-    stream_ptr: int
+    stream_ptr: Optional[int]
```
leofang: I think we raise if `stream=None` so this is not optional

cuda_core/cuda/core/experimental/_memoryview.pyx
leofang, line 109: Should we add an internal `cdef tuple shape` member that memoizes this request so that we pay the cost only once?

kkraus14 (dismissed):

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -92,9 +94,50 @@ cdef class StridedMemoryView:
...
+    @property
+    def shape(self) -> tuple[int]:
+        if self.exporting_obj is not None:
+            if self.dl_tensor != NULL:
+                return cuda_utils.carray_int64_t_to_tuple(
+                    self.dl_tensor.shape, 
+                    self.dl_tensor.ndim
+                )
+            else:
+                return self.exporting_obj.__cuda_array_interface__["shape"]
```
leofang: Another micro optimization (probably doesn't matter much if we memoize `shape` etc): We can also cache the CAI dict just like how we hold the `dl_tensor` pointer

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -92,9 +94,50 @@ cdef class StridedMemoryView:
...
+        cdef int itemsize
+        if self.exporting_obj is not None:
+            if self.dl_tensor != NULL:
+                if self.dl_tensor.strides:
+                    return cuda_utils.carray_int64_t_to_tuple(
+                        self.dl_tensor.strides, 
+                        self.dl_tensor.ndim
+                    )
+            else:
+                strides = self.exporting_obj.__cuda_array_interface__.get("strides")
```
leofang: ditto

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -92,9 +94,50 @@ cdef class StridedMemoryView:
...
+        return None
+
+    @property
+    def dtype(self) -> Optional[numpy.dtype]:
+        if self.exporting_obj is not None:
+            if self.dl_tensor != NULL:
+                return dtype_dlpack_to_numpy(&self.dl_tensor.dtype)
+            else:
+                # TODO: this only works for built-in numeric types
+                return numpy.dtype(self.exporting_obj.__cuda_array_interface__["typestr"])
```
leofang: ditto

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -302,33 +337,29 @@ cdef StridedMemoryView view_as_cai(obj, stream_ptr, view=None):
...
-        consumer_s = <intptr_t>(stream_ptr)
-        assert producer_s > 0
-        # establish stream order
-        if producer_s != consumer_s:
-            e = handle_return(driver.cuEventCreate(
-                driver.CUevent_flags.CU_EVENT_DISABLE_TIMING))
-            handle_return(driver.cuEventRecord(e, producer_s))
-            handle_return(driver.cuStreamWaitEvent(consumer_s, e, 0))
-            handle_return(driver.cuEventDestroy(e))
+    stream_ptr = int(stream_ptr) if stream_ptr is not None else -1
```
leofang: This is a good catch... in fact at this point `stream_ptr` is guaranteed to be not None (we raise above), so maybe we just do this?
```suggestion
    stream_ptr = int(stream_ptr)
```

cuda_core/tests/test_utils.py
```diff
@@ -164,3 +165,34 @@ def _check_view(self, view, in_arr, dev):
...
+    def test_cuda_array_interface_gpu(self, in_arr, use_stream):
+        # TODO: use the device fixture?
+        dev = Device()
+        dev.set_current()
+        # This is the consumer stream
+        s = dev.create_stream() if use_stream else None
+
+        # The usual path in `StridedMemoryView` prefers the DLPack interface
+        # over __cuda_array_interface__, so we call `view_as_cai` directly
+        # here so we can test the CAI code path.
```
leofang, line 182: better test coverage 👍

kkraus14 (dismissed):

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -74,15 +73,26 @@ cdef class StridedMemoryView:
...
+    # it later.  If using CAI, this is a reference to the source 
+    # `__cuda_array_interface__` object.
+    cdef object metadata
+
+    # The tensor object if has obj has __dlpack__, otherwise must be NULL
+    cdef DLTensor *dl_tensor
+        
+    # Memoized properties
+    cdef tuple _shape
+    cdef object _strides
```
leofang: IIRC `self._strides = None` is allowed in Cython even if `_strides` is typed to tuple, so this might help readability (and perf too?)
```suggestion
    cdef tuple _strides
```

leofang (approved):
LGTM, thanks Mike!

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -92,9 +102,52 @@ cdef class StridedMemoryView:
...
+            else:
+                self._shape = self.metadata["shape"]
+        else:
+            self._shape = ()
+        return self._shape
+
+    @property
+    def strides(self) -> Optional[tuple[int]]:
+        cdef int itemsize
+        if self._strides is None and self.exporting_obj is not None:
```
leofang: I see a minor potential concern here. Unlike `_shape` where we can use `None` to indicate it's uninitialized, `_strides` cannot do this because `None` is a valid value (indicating C-contiguity) which was meant to be a fast path. Now with this check, we turn the fast path into a slow path because we keep re-running the checks (and resetting `_strides` to `None`) every time this property is requested. A dumb fix would be to add another boolean member to indicate if `_strides` is initialized. Thoughts?

mdboom: It's probably a wash because these checks are so fast, but I was thinking that.

We could also do the `__getattr__` trick where you set the attribute directly on the object's dictionary, so the overhead after the first time is 0, but I wasn't sure if that would actually be faster with Cython.

PR #907, opened 2025-08-26
[NVVM IR] NVVM IR Integration

abhilash1910:
## Description

Abstract : The cuda/bindings backend of Cuda python has NVVM support through libnvvm api . However the frontend of cuda python does not support nvvm ir as input source. Since cuda python allows users to leverage a "pythonic dsl" format for writing the host code (taking care of launch parameters etc), it makes sense to also allow NVVM IR as an alternative input to the already included list of inputs {ptx, c++, lto ir} etc.

Discussion Link: #906 

Changes made {to be made} in this PR:

- Added cuda core linkage to cuda bindings nvvm counterpart 
- Cosmetic changes in user interface to use existing nvvm backend of cuda bindings.

## Checklist

- [ TBD ] New tests needed to be added to cover these changes.
- [ TBD ] The documentation needs to be  updated with these changes.

cc @leofang

### DIFF
--- a/cuda_core/cuda/core/experimental/_module.py
+++ b/cuda_core/cuda/core/experimental/_module.py
@@ -544,6 +544,24 @@ class ObjectCode:
             them (default to no mappings).
         """
         return ObjectCode._init(module, "ltoir", name=name, symbol_mapping=symbol_mapping)
+    
+    @staticmethod
+    def from_nvvm(module: Union[bytes, str], *, name: str = "", symbol_mapping: Optional[dict] = None) -> "ObjectCode":
+        """Create an :class:`ObjectCode` instance from an existing NVVM IR.
+
+        Parameters
+        ----------
+        module : Union[bytes, str]
+            Either a bytes object containing the in-memory NVVM IR code to load, or
+            a file path string pointing to the on-disk NVVM IR file to load.
+        name : Optional[str]
+            A human-readable identifier representing this code object.
+        symbol_mapping : Optional[dict]
+            A dictionary specifying how the unmangled symbol names (as keys)
+            should be mapped to the mangled names before trying to retrieve
+            them (default to no mappings).
+        """
+        return ObjectCode._init(module, "nvvm", name=name, symbol_mapping=symbol_mapping)
 
     @staticmethod
     def from_fatbin(
--- a/cuda_core/cuda/core/experimental/_program.py
+++ b/cuda_core/cuda/core/experimental/_program.py
@@ -24,6 +24,7 @@ from cuda.core.experimental._utils.cuda_utils import (
     is_nested_sequence,
     is_sequence,
     nvrtc,
+    nvvm,
 )
 
 
@@ -370,22 +371,26 @@ class Program:
     code : Any
         String of the CUDA Runtime Compilation program.
     code_type : Any
-        String of the code type. Currently ``"ptx"`` and ``"c++"`` are supported.
+        String of the code type. Currently ``"ptx"``, ``"c++"``, and ``"nvvm"`` are supported.
     options : ProgramOptions, optional
         A ProgramOptions object to customize the compilation process.
         See :obj:`ProgramOptions` for more information.
     """
 
     class _MembersNeededForFinalize:
-        __slots__ = "handle"
+        __slots__ = "handle", "backend"
 
-        def __init__(self, program_obj, handle):
+        def __init__(self, program_obj, handle, backend="NVRTC"):
             self.handle = handle
+            self.backend = backend
             weakref.finalize(program_obj, self.close)
 
         def close(self):
             if self.handle is not None:
-                handle_return(nvrtc.nvrtcDestroyProgram(self.handle))
+                if self.backend == "NVRTC":
+                    handle_return(nvrtc.nvrtcDestroyProgram(self.handle))
+                elif self.backend == "NVVM":
+                    handle_return(nvvm.destroy_program(self.handle))
                 self.handle = None
 
     __slots__ = ("__weakref__", "_mnff", "_backend", "_linker", "_options")
@@ -402,6 +407,7 @@ class Program:
             # TODO: allow tuples once NVIDIA/cuda-python#72 is resolved
 
             self._mnff.handle = handle_return(nvrtc.nvrtcCreateProgram(code.encode(), options._name, 0, [], []))
+            self._mnff.backend = "NVRTC"
             self._backend = "NVRTC"
             self._linker = None
 
@@ -411,8 +417,21 @@ class Program:
                 ObjectCode._init(code.encode(), code_type), options=self._translate_program_options(options)
             )
             self._backend = self._linker.backend
+
+        elif code_type == "nvvm":
+            if isinstance(code, str):
+                code = code.encode('utf-8')
+            elif not isinstance(code, (bytes, bytearray)):
+                raise TypeError("NVVM IR code must be provided as str, bytes, or bytearray")
+            
+            self._mnff.handle = nvvm.create_program()
+            self._mnff.backend = "NVVM"
+            nvvm.add_module_to_program(self._mnff.handle, code, len(code), options._name.decode())
+            self._backend = "NVVM"
+            self._linker = None
+
         else:
-            supported_code_types = ("c++", "ptx")
+            supported_code_types = ("c++", "ptx", "nvvm")
             assert code_type not in supported_code_types, f"{code_type=}"
             raise RuntimeError(f"Unsupported {code_type=} ({supported_code_types=})")
 
@@ -512,6 +531,42 @@ class Program:
                     logs.write(log.decode("utf-8", errors="backslashreplace"))
 
             return ObjectCode._init(data, target_type, symbol_mapping=symbol_mapping, name=self._options.name)
+
+        elif self._backend == "NVVM":
+            if target_type != "ptx":
+                raise ValueError(f'NVVM backend only supports target_type="ptx", got "{target_type}"')
+            
+            nvvm_options = []
+            if self._options.arch is not None:
+                arch = self._options.arch
+                if arch.startswith("sm_"):
+                    arch = f"compute_{arch[3:]}"
+                nvvm_options.append(f"-arch={arch}")
+            else:
+                major, minor = Device().compute_capability
+                nvvm_options.append(f"-arch=compute_{major}{minor}")
+
+            if self._options.debug:
+                nvvm_options.append("-g")
+            if self._options.device_code_optimize is False:
+                nvvm_options.append("-opt=0")
+            elif self._options.device_code_optimize is True:
+                nvvm_options.append("-opt=3")
+
+            nvvm.compile_program(self._mnff.handle, len(nvvm_options), nvvm_options)
+            
+            size = nvvm.get_compiled_result_size(self._mnff.handle)
+            data = bytearray(size)
+            nvvm.get_compiled_result(self._mnff.handle, data)
+            
+            if logs is not None:
+                logsize = nvvm.get_program_log_size(self._mnff.handle)
+                if logsize > 1:
+                    log = bytearray(logsize)
+                    nvvm.get_program_log(self._mnff.handle, log)
+                    logs.write(log.decode("utf-8", errors="backslashreplace"))
+                    
+            return ObjectCode._init(data, target_type, name=self._options.name)
 
         supported_backends = ("nvJitLink", "driver")
         if self._backend not in supported_backends:
--- a/cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
+++ b/cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
@@ -9,11 +9,11 @@ from collections.abc import Sequence
 from typing import Callable
 
 try:
-    from cuda.bindings import driver, nvrtc, runtime
+    from cuda.bindings import driver, nvrtc, nvvm, runtime
 except ImportError:
     from cuda import cuda as driver
     from cuda import cudart as runtime
-    from cuda import nvrtc
+    from cuda import nvrtc, nvvm
 
 from cuda.core.experimental._utils.driver_cu_result_explanations import DRIVER_CU_RESULT_EXPLANATIONS
 from cuda.core.experimental._utils.runtime_cuda_error_explanations import RUNTIME_CUDA_ERROR_EXPLANATIONS
@@ -24,6 +24,10 @@ class CUDAError(Exception):
 
 
 class NVRTCError(CUDAError):
+    pass
+
+
+class NVVMError(CUDAError):
     pass
 
 
@@ -55,6 +59,7 @@ def _reduce_3_tuple(t: tuple):
 cdef object _DRIVER_SUCCESS = driver.CUresult.CUDA_SUCCESS
 cdef object _RUNTIME_SUCCESS = runtime.cudaError_t.cudaSuccess
 cdef object _NVRTC_SUCCESS = nvrtc.nvrtcResult.NVRTC_SUCCESS
+cdef object _NVVM_SUCCESS = nvvm.Result.SUCCESS
 
 
 cpdef inline int _check_driver_error(error) except?-1:
@@ -103,6 +108,22 @@ cpdef inline int _check_nvrtc_error(error, handle=None) except?-1:
     raise NVRTCError(err)
 
 
+cpdef inline int _check_nvvm_error(error, handle=None) except?-1:
+    if error == _NVVM_SUCCESS:
+        return 0
+    err = f"{error}: {nvvm.get_error_string(error)}"
+    if handle is not None:
+        try:
+            logsize = nvvm.get_program_log_size(handle)
+            if logsize > 1:
+                log = bytearray(logsize)
+                nvvm.get_program_log(handle, log)
+                err += f", compilation log:\n\n{log.decode('utf-8', errors='backslashreplace')}"
+        except:
+            pass  # Log extraction failed, but we still have the error
+    raise NVVMError(err)
+
+
 cdef inline int _check_error(error, handle=None) except?-1:
     if isinstance(error, driver.CUresult):
         return _check_driver_error(error)
@@ -110,6 +131,8 @@ cdef inline int _check_error(error, handle=None) except?-1:
         return _check_runtime_error(error)
     elif isinstance(error, nvrtc.nvrtcResult):
         return _check_nvrtc_error(error, handle=handle)
+    elif isinstance(error, nvvm.Result):
+        return _check_nvvm_error(error, handle=handle)
     else:
         raise RuntimeError(f"Unknown error type: {error}")
 


cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
```diff
@@ -103,13 +108,31 @@ cpdef inline int _check_nvrtc_error(error, handle=None) except?-1:
...
+    err = f"{error}: {nvvm.get_error_string(error)}"
+    if handle is not None:
+        try:
+            logsize = nvvm.get_program_log_size(handle)
+            if logsize > 1:
+                log = bytearray(logsize)
+                nvvm.get_program_log(handle, log)
+                err += f", compilation log:\n\n{log.decode('utf-8', errors='backslashreplace')}"
+        except:
+            pass  # Log extraction failed, but we still have the error
```
rwgk: Completely masking errors in the `try` block could become extremely confusing. How about:

```
except Exception as e:
    raise NVVMError(err) from e
```

abhilash1910: Addressed , thanks.

cuda_core/cuda/core/experimental/_module.py
```diff
@@ -544,6 +544,24 @@ def from_ltoir(module: Union[bytes, str], *, name: str = "", symbol_mapping: Opt
             them (default to no mappings).
         """
         return ObjectCode._init(module, "ltoir", name=name, symbol_mapping=symbol_mapping)
+    
+    @staticmethod
+    def from_nvvm(module: Union[bytes, str], *, name: str = "", symbol_mapping: Optional[dict] = None) -> "ObjectCode":
```
leofang: Q: How do we plan to use NVVM IR wrapped as a `ObjectCode` in this case? It's not a valid input for any linker (nvjitlink/culink).

cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
```diff
@@ -9,11 +9,11 @@ from collections.abc import Sequence
 from typing import Callable
 
 try:
-    from cuda.bindings import driver, nvrtc, runtime
+    from cuda.bindings import driver, nvrtc, nvvm, runtime
 except ImportError:
     from cuda import cuda as driver
     from cuda import cudart as runtime
-    from cuda import nvrtc
+    from cuda import nvrtc, nvvm
```
leofang: We should import nvvm separately from all other modules. The NVVM bindings was added very recently ([12.9.0](https://nvidia.github.io/cuda-python/cuda-bindings/latest/release/12.9.0-notes.html)), so it is not available for users on cuda.bindings 12.8 and below, but we don't want to force them to update unnecessarily for now. Today cuda.core works with any CUDA 12.x and 13.x.

Another note is that `from cuda import nvvm` is invalid. Always import it from `cuda.bindings` because it's where it is since it was added.

We should probably defer the import until it is actually needed, in `Program` when compiling NVVM IRs to PTX, and if NVVM is not available because of either reason
1. cuda.bindings is not new enough
2. libnvvm is not found, which can happen in Python environments (this can be checked by calling `cuda.bindings._internal.nvvm._inspect_function_pointers()`)

then we raise an error.

We can drop this defer import by CUDA 14.0 (when we drop the support for 12.x).

cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
```diff
@@ -27,6 +27,10 @@ class NVRTCError(CUDAError):
     pass
 
 
+class NVVMError(CUDAError):
```
leofang: So here's a bit of project history 😅  For the driver/runtime/nvrtc bindings, we did not raise exceptions if the C API returns a nonzero error code, and instead we return the error code as the first element of the returned tuple. So we had to cover these weird situations in cuda.core.

This is no longer the case for all other newer bindings such as nvJitLink and NVVM, which are auto-generated by another internal codegen. We already have the exception inspection done at the binding level, for example `cuda.bindings.nvvm.NVVMError` is a thing already. So we don't need to make any of the changes in this file, starting this line (you can see there's no special handling for nvJitLink in this file either).

cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
```diff
@@ -103,13 +108,31 @@ cpdef inline int _check_nvrtc_error(error, handle=None) except?-1:
     raise NVRTCError(err)
 
 
+cpdef inline int _check_nvvm_error(error, handle=None) except?-1:
```
leofang: ditto, drop

cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
```diff
@@ -103,13 +108,31 @@ cpdef inline int _check_nvrtc_error(error, handle=None) except?-1:
...
+
+
 cdef inline int _check_error(error, handle=None) except?-1:
     if isinstance(error, driver.CUresult):
         return _check_driver_error(error)
     elif isinstance(error, runtime.cudaError_t):
         return _check_runtime_error(error)
     elif isinstance(error, nvrtc.nvrtcResult):
         return _check_nvrtc_error(error, handle=handle)
+    elif isinstance(error, nvvm.Result):
```
leofang: ditto, drop

cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
abhilash1910: Addressed this change, thanks for the insight.

cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
abhilash1910: Changed the import to a standalone function in program.py where it is invoked when needed, deferring the dependency in cuda_utils by 14.0

cuda_core/cuda/core/experimental/_module.py
abhilash1910: Yes this should be "ptx" or target and has been changed as nvvm  ir as a wrapped object code is not valid for anything. This was giving compilation issues and has been fixed.

cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
leofang, line 1: nit: Let's not modify the spacing here. If we were to turn on cython-lint, this would have triggered a linter error. Functions and objects by default are expected to separate from each other with two empty lines.

cuda_core/cuda/core/experimental/_module.py
leofang: Then let's remove this method if it is not usable (and it's certainly not used by this PR anyway).

cuda_core/cuda/core/experimental/_program.py
```diff
@@ -21,11 +21,63 @@
...
+                f"NVVM bindings require cuda-bindings >= 12.9.0, but found {version[0]}.{version[1]}.x. "
+                "Please update cuda-bindings to use NVVM features."
+            )
+        
+        from cuda.bindings import nvvm
+        try:
+            from cuda.bindings._internal.nvvm import _inspect_function_pointers
+            _inspect_function_pointers()
+        except Exception as e:
+            raise ImportError(
```
leofang: At this point the nvvm bindings exist. Neither `_inspect_function_pointers` nor `_inspect_function_pointer` would raise. We check if dlsym succeeds and if the function pointer is not null.
```suggestion
        from cuda.bindings._internal.nvvm import _inspect_function_pointer
        if _inspect_function_pointer("__nvvmCreateProgram") == 0:
            raise ImportError(
```

cuda_core/cuda/core/experimental/_program.py
```diff
@@ -21,11 +21,63 @@
...
+                "NVVM library (libnvvm) is not available in this Python environment. "
+                f"Original error: {e}"
+            )
```
leofang:
```suggestion
            )
```

cuda_core/cuda/core/experimental/_program.py
```diff
@@ -370,22 +422,30 @@ class Program:
...
             if self.handle is not None:
-                handle_return(nvrtc.nvrtcDestroyProgram(self.handle))
+                if self.backend == "NVRTC":
+                    handle_return(nvrtc.nvrtcDestroyProgram(self.handle))
+                elif self.backend == "NVVM":
+                    try:
+                        nvvm = _get_nvvm_module()
+                        nvvm.destroy_program(self.handle)
+                    except ImportError as e:
+                        pass
```
leofang: I think if NVVM is not usable, `self.handle` will always be None by the time we raise ImportError at the Program construction time, in which case to deallocate the Program instance we don't need try-except here. 

If NVVM is usable, then the module is always available.
```suggestion
                    nvvm = _get_nvvm_module()
                    nvvm.destroy_program(self.handle)
```

cuda_core/cuda/core/experimental/_program.py
```diff
@@ -370,22 +422,30 @@ class Program:
...
         A ProgramOptions object to customize the compilation process.
         See :obj:`ProgramOptions` for more information.
     """
 
     class _MembersNeededForFinalize:
-        __slots__ = "handle"
+        __slots__ = "handle", "backend"
 
-        def __init__(self, program_obj, handle):
+        def __init__(self, program_obj, handle, backend="NVRTC"):
```
leofang: nit: Let's not set any default for backend 
```suggestion
        def __init__(self, program_obj, handle, backend):
```

cuda_core/cuda/core/experimental/_program.py
```diff
@@ -513,6 +588,44 @@ def compile(self, target_type, name_expressions=(), logs=None):
...
+            else:
+                major, minor = Device().compute_capability
+                nvvm_options.append(f"-arch=compute_{major}{minor}")
+
+            if self._options.debug:
+                nvvm_options.append("-g")
+            if self._options.device_code_optimize is False:
+                nvvm_options.append("-opt=0")
+            elif self._options.device_code_optimize is True:
+                nvvm_options.append("-opt=3")
```
leofang: Let's abstract this out to `_translate_program_options` (or have a separate internal method). `ProgramOptions` was meant to be a superset that encapsulates possible compiler backends' options.

cuda_core/tests/test_program.py
```diff
@@ -12,6 +12,83 @@
...
 
+def get_nvvm_ir():
+    """Generate working NVVM IR with proper version metadata"""
+    try:
+        from cuda.core.experimental._program import _get_nvvm_module
+        nvvm = _get_nvvm_module()
+        major, minor, debug_major, debug_minor = nvvm.ir_version()
+        
+        
+        nvvm_ir_template = '''target triple = "nvptx64-unknown-cuda"
```
leofang: I think this would solve the linter failure (partially)
```suggestion
        nvvm_ir_template = r'''target triple = "nvptx64-unknown-cuda"
```

cuda_core/tests/test_program.py
```diff
@@ -12,6 +12,83 @@
 
 is_culink_backend = _linker._decide_nvjitlink_or_driver()
 
+def get_nvvm_ir():
+    """Generate working NVVM IR with proper version metadata"""
+    try:
+        from cuda.core.experimental._program import _get_nvvm_module
+        nvvm = _get_nvvm_module()
+        major, minor, debug_major, debug_minor = nvvm.ir_version()
```
leofang: 1. Could you leave a comment on why we need to test the versioned IR? When will this block raise (and switch to test the unversioned one)?
2. We should make this function a pytest fixture
3. We should create a skip condition in case libNVVM is not installed

cuda_core/tests/test_program.py
```diff
@@ -12,6 +12,83 @@
...
+
+declare i32 @llvm.nvvm.read.ptx.sreg.ctaid.x() nounwind readnone
+declare i32 @llvm.nvvm.read.ptx.sreg.ntid.x() nounwind readnone
+declare i32 @llvm.nvvm.read.ptx.sreg.tid.x() nounwind readnone
+
+!nvvm.annotations = !{!0}
+!0 = !{void (i32*)* @simple, !"kernel", i32 1}
+"""
+
+nvvm_ir = get_nvvm_ir()
```
leofang: ditto, by using a fixture we won't have to create this module-level constant

cuda_core/tests/test_program.py
```diff
@@ -150,3 +227,105 @@ def test_program_close():
...
+    """Test that our deferred NVVM import works correctly"""
+    try:
+        from cuda.core.experimental._program import _get_nvvm_module
+        nvvm = _get_nvvm_module()
+        assert nvvm is not None
+    except ImportError as e:
+        pytest.skip(f"NVVM not available: {e}")
+
+
+def test_nvvm_program_creation():
```
leofang: ditto, fixtures can be passed to tests this way (pick a name and stick to it)
```suggestion
def test_nvvm_program_creation(nvvm_ir):
```

leofang: The skip condition should be used to decorate all NVVM-related tests, so that we can remove try-except in the test bodies.

cuda_core/tests/test_program.py
```diff
@@ -150,3 +227,105 @@ def test_program_close():
...
+            else:
+                ptx_text = str(ptx_code)
+            assert ".visible .entry simple(" in ptx_text
+            
+            ptx_kernel = ptx_object_code.get_kernel("simple")
+            assert isinstance(ptx_kernel, Kernel)
+            
+        except Exception as e:
+            if any(error in str(e) for error in ["ERROR_IR_VERSION_MISMATCH", "ERROR_INVALID_OPTION", "ERROR_COMPILATION"]):
+                pytest.skip(f"NVVM IR not compatible with this CUDA version: {e}")
```
leofang: Q: I'd like to know when will this skip be triggered?

cuda_core/tests/test_program.py
```diff
@@ -150,3 +227,105 @@ def test_program_close():
...
+        finally:
+            program.close()
+            
+    except ImportError as e:
+        pytest.skip(f"NVVM not available: {e}")
+
+
+@pytest.mark.parametrize("options", [
+    ProgramOptions(name="test1", arch="sm_90", device_code_optimize=False),
+    ProgramOptions(name="test2", arch="sm_90", device_code_optimize=False),
```
leofang: Q: 
1. What's the difference between these two test cases?
2. Isn't this test identical to the previous `test_nvvm_compile_to_ptx` test? Do we need both? Seems like a lot of code dup to me.

abhilash1910: yeah agree , was planning to remove this . This I initially copied to check with different Sm versions.

cuda_core/tests/test_program.py
abhilash1910: This was kept as a placeholder incase the libnvvm version mismatches cuda, but it should not happen in real conditions as the ctk bundles it  inside nvvm/lib64 . So safer to remove this .

cuda_core/cuda/core/experimental/_program.py
```diff
@@ -9,6 +9,8 @@
 from typing import TYPE_CHECKING, List, Tuple, Union
 from warnings import warn
 
+import list
+
```
leofang: `list` is a builtin type, it's ok to just use it for type annotation
```suggestion
```

cuda_core/cuda/core/experimental/_program.py
```diff
@@ -433,6 +503,27 @@ def _translate_program_options(self, options: ProgramOptions) -> LinkerOptions:
...
+        nvvm_options = []
+
+        if options.arch is not None:
+            arch = options.arch
+            if arch.startswith("sm_"):
+                arch = f"compute_{arch[3:]}"
+            nvvm_options.append(f"-arch={arch}")
+        else:
+            major, minor = Device().compute_capability
+            nvvm_options.append(f"-arch=compute_{major}{minor}")
```
leofang: IIRC we already detected current arch in `ProgramOptions` (plz double check), so this is not needed
```suggestion
```

cuda_core/cuda/core/experimental/_program.py
```diff
@@ -433,6 +503,27 @@ def _translate_program_options(self, options: ProgramOptions) -> LinkerOptions:
             ptxas_options=options.ptxas_options,
         )
 
+    def _translate_program_options_to_nvvm(self, options: ProgramOptions) -> list[str]:
```
leofang, line 530: Looking good, could you also translate
- ftz
- prec_sqrt
- prec_div
- fma

and also support `-gen-lto`? This means in addition to `target_type="ptx"` we can also support `target_type="ltoir"`

leofang (commented):
Thanks @abhilash1910! Looks very good! A few minor comments for completeness. Let me trigger the CI in the meanwhile.

leofang:
/ok to test [5e2e137](https://github.com/NVIDIA/cuda-python/pull/907/commits/5e2e137659df87a15a0b950e042a9d2c30883618)

cuda_core/cuda/core/experimental/_program.py
leofang: This one is critical bug

leofang:
/ok to test [0301a4d](https://github.com/NVIDIA/cuda-python/pull/907/commits/0301a4d670e8533ef66367e23c28f1960dda2b9f)

abhilash1910:
Fixing the CI, seems like the datalayout "\ " for nvvm ir is causing the issue. The layout should be in one line but that exceeds 120 characters limit on a single line.

leofang:
check if turning it into a raw string `r""" ... """` would make linter happy about long lines, if so we can remove the problematic line breaks

gmarkall (approved):
I've had a quick look at this and it looks good in general to me - I believe it covers all the same options that Numba-CUDA uses, for example. A couple of questions / thoughts (which shouldn't block this PR, but worth considering):

- Is there a way to add multiple modules? It's common to want to compile multiple modules together into a single output.
- If / when it is possible to add multiple modules, a test with code that uses something from libdevice is probably a good idea. It's a bitcode library, and the tests here only cover text-based IR (I think).
- It's also useful to be able to lazily add a module - for example, adding libdevice is quite time-consuming so adding it lazily saves a lot of time (especially if doing multiple compilations). An example of doing this in Numba-CUDA: https://github.com/NVIDIA/numba-cuda/blob/98a9213144be410f52f79802cfa38c4c5a0378c1/numba_cuda/numba/cuda/cudadrv/nvvm.py#L637-L640
- It might be a good idea to also verify the program before compiling it, to catch user errors that might otherwise lead to a crash.

cuda_core/tests/test_program.py
```diff
@@ -13,6 +13,106 @@
...
+declare i32 @llvm.nvvm.read.ptx.sreg.ctaid.x() nounwind readnone
+declare i32 @llvm.nvvm.read.ptx.sreg.ntid.x() nounwind readnone
+declare i32 @llvm.nvvm.read.ptx.sreg.tid.x() nounwind readnone
+
+!nvvm.annotations = !{{!0}}
+!0 = !{{void (i32*)* @simple, !"kernel", i32 1}}
+
+!nvvmir.version = !{{!1}}
+!1 = !{{i32 {major}, i32 0, i32 {debug_major}, i32 0}}
+"""
```
leofang: @abhilash1910 I was scratching my head as our other NVVM test file also has long lines but the linter does not complain at all. This is the key:
https://github.com/NVIDIA/cuda-python/blob/d64120bb6733391989174456daaf18643ea29711/cuda_bindings/tests/test_nvvm.py#L37
Let's apply this change and remove the unnecessary line breaks in the text IRs.
```suggestion
"""  # noqa: E501
```

cuda_core/tests/test_program.py
```diff
@@ -13,6 +13,106 @@
...
+  ret void
+}
+
+declare i32 @llvm.nvvm.read.ptx.sreg.ctaid.x() nounwind readnone
+declare i32 @llvm.nvvm.read.ptx.sreg.ntid.x() nounwind readnone
+declare i32 @llvm.nvvm.read.ptx.sreg.tid.x() nounwind readnone
+
+!nvvm.annotations = !{!0}
+!0 = !{void (i32*)* @simple, !"kernel", i32 1}
+"""
```
leofang: ditto
```suggestion
"""  # noqa: E501
```

leofang:
Thanks for reviews @gmarkall!
> * Is there a way to add multiple modules? It's common to want to compile multiple modules together into a single output.

Appreciate for the reminder... I had forgotten about this capability. 

This is a very nice use case we should probably support. Currently `Program` accepts a single input source because it fits the majority of use cases. Perhaps we can add an `extra_sources` argument to `ProgramOptions`? I can imagine this option being used by other backends (ex: including multiple PTX and/or objects).

An alternative is to use `Linker` instead of `Program` for this need. To me conceptually it fits better. I can imagine using the ([currently-removed-upon-my-request](https://github.com/NVIDIA/cuda-python/pull/907#discussion_r2305026940)) `ObjectCode.from_nvvm` constructor to wrap multiple NVVM IRs, and then feed them into the `Linker` and call `.link()` to compile them into a single PTX or LTO IR. Do you think this model makes better sense?

Anyway
> which shouldn't block this PR, but worth considering

I agree. We should have the next step evaluated, and as long as the current implementation is compatible with future extensions (that is, we won't have to introduce breaking changes), we can merge this PR to unblock Abhilash and address the great points you raised in another PR.

cuda_core/tests/test_program.py
abhilash1910: This is great and most easiest fix, I tried f strings and a lot of other hacks but that always increases the code lines. 
Thanks @leofang

abhilash1910:
Thanks @leofang  and @gmarkall  for the great review and suggestions

> * If / when it is possible to add multiple modules, a test with code that uses something from libdevice is probably a good idea. It's a bitcode library, and the tests here only cover text-based IR (I think).
> * It's also useful to be able to lazily add a module - for example, adding libdevice is quite time-consuming so adding it lazily saves a lot of time (especially if doing multiple compilations). An example of doing this in Numba-CUDA: https://github.com/NVIDIA/numba-cuda/blob/98a9213144be410f52f79802cfa38c4c5a0378c1/numba_cuda/numba/cuda/cudadrv/nvvm.py#L637-L640

This is something which I was thinking as well, adding libdevice lazily as a module. For libnvvm.so currently the textual representation has been deprecated for bitcode format and essentially once the toolchain upgrades , I will add in the corresponding changes from nvvm perspective. We are currently doing a couple of changes in the libnvvm api which I will add here. For the first point , agree with @leofang . and that can be extended any point of time to make it compatible with LTo or PTX in the downstream path.

> * It might be a good idea to also verify the program before compiling it, to catch user errors that might otherwise lead to a crash.

Yes this is something which I am currently thinking to verify the program before compiling. There are a couple of ways to do this: either we can use builtin ```opt -verify``` , libnvvm nullptr exception cases, or ```llvm-as``` (for both translation & compilation validation). Can take that up in a following PR . Thanks.

cuda_core/tests/test_program.py
leofang: Yeah linter errors are meant to be fixable quickly in a minute or two. If we find ourselves struggling it is better to just skip them with `# noqa`.

leofang:
/ok to test [af1008f](https://github.com/NVIDIA/cuda-python/pull/907/commits/af1008ff076accd6b40270c7e9be21e38ab897c9)

cuda_core/tests/test_program.py
```diff
@@ -150,3 +248,55 @@ def test_program_close():
...
+    program.close()
+
+
+@nvvm_available
+@pytest.mark.parametrize(
+    "options",
+    [
+        ProgramOptions(name="test1", arch="sm_90", device_code_optimize=False),
+        ProgramOptions(name="test2", arch="sm_100", device_code_optimize=False),
+        ProgramOptions(name="test3", arch="sm_110", device_code_optimize=True),
```
leofang: I think sm_110 is not supported by 12.x. Let's make it skipped based on libnvvm version. A similar approach can be found here:
https://github.com/NVIDIA/cuda-python/blob/d64120bb6733391989174456daaf18643ea29711/cuda_core/tests/test_launcher.py#L279-L285

leofang:
> * Is there a way to add multiple modules? 
> * If / when it is possible to add multiple modules, a test with code that uses something from libdevice is probably a good idea. 

Pending above discussion. Address the conclusion in a separate PR.

> * It's also useful to be able to lazily add a module

@abhilash1910 @gmarkall would it hurt to always call `lazy_add_module` instead of `add_module`? If not, let's switch to use it, in this PR.

> * It might be a good idea to also verify the program before compiling it

I believe Graham meant to call `nvvm.verify_program()`. Again, would it hurt if we always call it before compiling (other than "a little" overhead)? If not, let's do it, in this PR.

abhilash1910:
/ok to test [a85a44f](https://github.com/NVIDIA/cuda-python/pull/907/commits/a85a44f7293dfe4578b4d30a50fed6b6a5f38eaf)

cuda_core/cuda/core/experimental/_program.py
abhilash1910: Added, thanks

cuda_core/tests/test_program.py
```diff
@@ -150,3 +250,61 @@ def test_program_close():
...
+    "options",
+    [
+        ProgramOptions(name="test1", arch="sm_90", device_code_optimize=False),
+        ProgramOptions(name="test2", arch="sm_100", device_code_optimize=False),
+        pytest.param(
+            ProgramOptions(name="test3", arch="sm_110", device_code_optimize=True),
+            marks=pytest.mark.skipif(
+                12000 <= cuda_driver_version < 13000,
+                reason="Compute capability 110 not supported with CUDA 12.x",
+            ),
```
leofang: So there is a reason we filed and asked for a new API to get libNVVM **library** version (nvbugs 5312315). Unfortunately, the driver version is not a good proxy 🥲 We can use a very new driver (ex: 14.0) running with a very old libNVVM (ex: from CTK 10.2). This is a supported use case, but this test will fail.

Can we think of a way to figure out the libNVVM version? If not, I suggest we
- include a small, dummy NVVM IR that can be successfully compiled by libNVVM
- when the first time a user-provided NVVM IR is fed to `Program`, compile this dummy IR into PTX
- check if [`get_ptx_ver`](https://nvidia.github.io/cuda-python/cuda-bindings/latest/module/generated/cuda.bindings.utils.get_ptx_ver.html#cuda.bindings.utils.get_ptx_ver) and [`get_minimal_required_cuda_ver_from_ptx_ver`](https://nvidia.github.io/cuda-python/cuda-bindings/latest/module/generated/cuda.bindings.utils.get_minimal_required_cuda_ver_from_ptx_ver.html#cuda.bindings.utils.get_minimal_required_cuda_ver_from_ptx_ver) exist in `cuda.bindings` (there's a new `utils` submodule that we recently added)
- if so, feed the generated PTX to this API combo to get a version number as the proxy to the libNVVM version. Cache this version as a private global variable for future reuse. Use this version to skip tests if ver < 13000
- otherwise, leave the variable as None, and unconditionally skip the test

xref: https://github.com/NVIDIA/numba-cuda/issues/262

cuda_core/cuda/core/experimental/_program.py
leofang: I get it now why you need this. Let's change `--gpu-architecture` to `-arch` here
https://github.com/NVIDIA/cuda-python/blob/d64120bb6733391989174456daaf18643ea29711/cuda_core/cuda/core/experimental/_program.py#L231-L236
so that all backends can use this. Then the above deletion can be made.

cuda_core/tests/test_program.py
abhilash1910: Thanks , changed the details of the tests to including compiling the dummy IR with libnvvm and then checking with the generated ptx ( & versions).

leofang:
/ok to test [caf634c](https://github.com/NVIDIA/cuda-python/pull/907/commits/caf634cb95399b98f8e93ca4c472e943db452b1b)

leofang (approved):

cuda_core/cuda/core/experimental/_program.py
```diff
@@ -20,12 +20,60 @@
...
 )
 
+_nvvm_module = None
+_nvvm_import_attempted = False
+
+
+def _get_nvvm_module():
+    """
+    Handles the import of NVVM module with version and availability checks.
+    NVVM bindings were added in CUDA 12.9.0, so we need to handle cases where:
```
rwgk: Did you mean "were added in cuda-bindings 12.9.0"? (mixing this up could be quite confusing)

cuda_core/cuda/core/experimental/_program.py
```diff
@@ -20,12 +20,60 @@
...
+    1. cuda.bindings is not new enough (< 12.9.0)
+    2. libnvvm is not found in the Python environment
+
+    Returns:
+        The nvvm module if available and working
+
+    Raises:
+        ImportError: If NVVM is not available due to version or library issues
```
rwgk: I highly recommend not using `ImportError` here, but to introduce a specific exception type (which is very easy).

Consider the code you have in the test case (copied from there):

```
def _is_nvvm_available():
    """Check if NVVM is available."""
    try:
        from cuda.core.experimental._program import _get_nvvm_module

        _get_nvvm_module()
        return True
    except ImportError:
        return False
```

Now imagine that there is a genuine bug elsewhere, in any of dependencies of `cuda.bindings` (which you're importing in current line 64 below). If that happens to manifest itself as an `ImportError`, that bug will be masked completely by the `except ImportError: return False` above. This could mean hours worth of debugging and back and forth on issues until we figure out what's really going on.

Untested suggestion (the indentation is probably off):

```
class NvvmModuleNotAvailableError(RuntimeError):
    pass

...
            raise NvvmModuleNotAvailableError("NVVM module is not available (previous import attempt failed)")
...
            raise NvvmModuleNotAvailableError(
                f"NVVM bindings require cuda-bindings >= 12.9.0, but found {version[0]}.{version[1]}.x. "
                "Please update cuda-bindings to use NVVM features."
            )
...
        try:
                   from cuda.bindings import nvvm
         except ImportError as e:
                 raise NvvmModuleNotAvailableError(f"cuda.bindings.nvvm ImportError: {e}")
 ...
         if _inspect_function_pointer("__nvvmCreateProgram") == 0:
             raise NvvmModuleNotAvailableError("<new accurate message here>")
```

rwgk: Here is a safer way to try the `cuda.bindings.nvvm` import:

https://chatgpt.com/share/68c890d6-9a94-8008-a5b4-c5d63b82f4cd

leofang: @rwgk I don't quite understand, are you suggesting to make changes to the test suite or to the `Program` implementation? 

Right now we only have a few public exceptions in cuda.core, so let's keep it this way at least in Abhilash's PR, and do a global sweep later if we want more exceptions.

FWIW, this is how we keep the same exception type but update the error msg:
https://github.com/NVIDIA/cuda-python/blob/e06b2993dc446a695197ead58033c658bf6b5eeb/cuda_core/cuda/core/experimental/_linker.py#L334-L335

cuda_core/cuda/core/experimental/_program.py
```diff
@@ -229,11 +277,9 @@ def __post_init__(self):
 
         self._formatted_options = []
         if self.arch is not None:
-            self._formatted_options.append(f"--gpu-architecture={self.arch}")
+            self._formatted_options.append(f"-arch={self.arch}")
         else:
-            self._formatted_options.append(
-                "--gpu-architecture=sm_" + "".join(f"{i}" for i in Device().compute_capability)
-            )
+            self._formatted_options.append("-arch=sm_" + "".join(f"{i}" for i in Device().compute_capability))
```
leofang:
```suggestion
            self.arch = f"sm_{Device().arch}"
            self._formatted_options.append(f"-arch={self.arch}")
```

cuda_core/cuda/core/experimental/_program.py
```diff
@@ -433,6 +499,36 @@ def _translate_program_options(self, options: ProgramOptions) -> LinkerOptions:
...
+        nvvm_options = []
+
+        if options.arch is not None:
+            arch = options.arch
+            if arch.startswith("sm_"):
+                arch = f"compute_{arch[3:]}"
+            nvvm_options.append(f"-arch={arch}")
+        else:
+            major, minor = Device().compute_capability
+            nvvm_options.append(f"-arch=compute_{major}{minor}")
```
leofang:
```suggestion
        assert options.arch is not None
        arch = options.arch
        if arch.startswith("sm_"):
            arch = f"compute_{arch[3:]}"
        nvvm_options.append(f"-arch={arch}")
```

leofang (commented):
nits to complete earlier suggestions

cuda_core/cuda/core/experimental/_program.py
rwgk: I pulled out the test code only to explain the pitfalls.

The suggested changes would be here, in `_program.py`.

> Right now we only have a few public exceptions in cuda.core

I'm very worried about repurposing `ImportError` for errors that are only loosely related to actual import errors.

But maybe I need to broaden my view?

I didn't carefully consider yet: Would `NvvmModuleNotAvailableError` actually be a user-facing exception type? (I assumed so.)

leofang: Yes it'd be user-facing, because we need to let users know that we cannot support `Program(code, code_type="nvvm", ...)` if libNVVM is not usable for whatever reason.

Thinking about this more, probably for this case raising `RuntimeError` is good enough. We already raise it in a similar situation:
https://github.com/NVIDIA/cuda-python/blob/e06b2993dc446a695197ead58033c658bf6b5eeb/cuda_core/cuda/core/experimental/_program.py#L414-L417

rwgk: > probably for this case raising `RuntimeError` is good enough.

That sounds good, if you want to avoid introducing exception types.

cuda_core/cuda/core/experimental/_program.py
```diff
@@ -20,12 +20,60 @@
...
+        The nvvm module if available and working
+
+    Raises:
+        ImportError: If NVVM is not available due to version or library issues
+    """
+    global _nvvm_module, _nvvm_import_attempted
+
+    if _nvvm_import_attempted:
+        if _nvvm_module is None:
+            raise ImportError("NVVM module is not available (previous import attempt failed)")
```
leofang:
```suggestion
            raise RuntimeError("NVVM module is not available (previous import attempt failed)")
```

cuda_core/cuda/core/experimental/_program.py
leofang: > if you want to avoid introducing exception types

For now! 🙂

leofang:
```suggestion
        RuntimeError: If NVVM is not available due to version or library issues
```

cuda_core/cuda/core/experimental/_program.py
```diff
@@ -20,12 +20,60 @@
...
+        return _nvvm_module
+
+    _nvvm_import_attempted = True
+
+    try:
+        version = get_binding_version()
+        if version < (12, 9):
+            raise ImportError(
```
leofang:
```suggestion
            raise RuntimeError(
```

cuda_core/cuda/core/experimental/_program.py
```diff
@@ -20,12 +20,60 @@
...
+                f"NVVM bindings require cuda-bindings >= 12.9.0, but found {version[0]}.{version[1]}.x. "
+                "Please update cuda-bindings to use NVVM features."
+            )
+
+        from cuda.bindings import nvvm
+        from cuda.bindings._internal.nvvm import _inspect_function_pointer
+
+        if _inspect_function_pointer("__nvvmCreateProgram") == 0:
+            raise ImportError("NVVM library (libnvvm) is not available in this Python environment. ")
```
leofang:
```suggestion
            raise RuntimeError("NVVM library (libnvvm) is not available in this Python environment. ")
```

cuda_core/cuda/core/experimental/_program.py
```diff
@@ -20,12 +20,60 @@
...
+
+        _nvvm_module = nvvm
+        return _nvvm_module
+
+    except ImportError as e:
```
leofang: technically this try-except is not needed because this function is implemented in the "ask-for-permission" instead of "ask-for-forgiveness" style...
```suggestion
    except RuntimeError as e:
```

cuda_core/tests/test_program.py
```diff
@@ -9,10 +9,187 @@
...
 
 
+def _is_nvvm_available():
+    """Check if NVVM is available."""
+    try:
+        from cuda.core.experimental._program import _get_nvvm_module
+
+        _get_nvvm_module()
+        return True
+    except ImportError:
```
leofang:
```suggestion
    except RuntimeError:
```

cuda_core/cuda/core/experimental/_program.py
leofang:
```suggestion
    NVVM bindings were added in cuda-bindings 12.9.0, so we need to handle cases where:
```

leofang:
/ok to test [d5d216c](https://github.com/NVIDIA/cuda-python/pull/907/commits/d5d216c76b4a5a09a202dfecbeba8d1abe5bc5c9)

cuda_core/cuda/core/experimental/_program.py
```diff
@@ -20,12 +20,60 @@
...
+            raise RuntimeError(
+                f"NVVM bindings require cuda-bindings >= 12.9.0, but found {version[0]}.{version[1]}.x. "
+                "Please update cuda-bindings to use NVVM features."
+            )
+
+        from cuda.bindings import nvvm
+        from cuda.bindings._internal.nvvm import _inspect_function_pointer
+
+        if _inspect_function_pointer("__nvvmCreateProgram") == 0:
+            raise RuntimeError("NVVM library (libnvvm) is not available in this Python environment. ")
```
leofang, line 95: @rwgk did I make a mistake in suggesting Abhilash to use this trick instead of using the pathfinder API?

cuda_core/tests/test_program.py
```diff
@@ -9,10 +9,187 @@
...
+target triple = "nvptx64-nvidia-cuda"
+target datalayout = "e-p:64:64:64-i1:8:8-i8:8:8-i16:16:16-i32:32:32-i64:64:64-i128:128:128-f32:32:32-f64:64:64-v16:16:16-v32:32:32-v64:64:64-v128:128:128-n16:32:64"
+
+define void @dummy_kernel() {
+entry:
+  ret void
+}
+
+!nvvm.annotations = !{!0}
+!0 = !{void ()* @dummy_kernel, !"kernel", i32 1}
```
leofang: Took me a while to debug...

This IR is incomplete (it fails at `verify_program()`) because it missed the `!nvvmir.version` attribute.

cuda_core/cuda/core/experimental/_program.py
```diff
@@ -513,6 +607,31 @@ def compile(self, target_type, name_expressions=(), logs=None):
...
 
+        elif self._backend == "NVVM":
+            if target_type not in ("ptx", "ltoir"):
+                raise ValueError(f'NVVM backend only supports target_type="ptx", "ltoir", got "{target_type}"')
+
+            nvvm_options = self._translate_program_options_to_nvvm(self._options)
+            if target_type == "ltoir" and "-gen-lto" not in nvvm_options:
+                nvvm_options.append("-gen-lto")
+            nvvm = _get_nvvm_module()
+            nvvm.verify_program(self._mnff.handle, len(nvvm_options), nvvm_options)
```
leofang: Given how hard it is to debug IR issues, we should wrap `verify` and `compile` in `_exception_manager` from `_linker.py` (copy/paste it to this file and modify it for now, I'll do a refactor later), like this
```python
with _exception_manager(self):
    nvvm.verify_program(...)
    nvvm.compile_program(...)
```
In the local `_exception_manager`, make sure the program log is fetched if any exception is raised.

leofang: (We did not need `_exception_manager` for NVRTC because `handle_return` has a special path for it.)

cuda_core/tests/test_program.py
```diff
@@ -9,10 +9,187 @@
...
+
+declare i32 @llvm.nvvm.read.ptx.sreg.ctaid.x() nounwind readnone
+declare i32 @llvm.nvvm.read.ptx.sreg.ntid.x() nounwind readnone
+declare i32 @llvm.nvvm.read.ptx.sreg.tid.x() nounwind readnone
+
+!nvvm.annotations = !{{!0}}
+!0 = !{{void (i32*)* @simple, !"kernel", i32 1}}
+
+!nvvmir.version = !{{!1}}
+!1 = !{{i32 {major}, i32 0, i32 {debug_major}, i32 0}}
```
leofang: Q: For my own curiosity: `ir_version()` returns 4 integers. Why don't we use them all, instead of hard-coding the minor versions to 0?

cuda_core/tests/test_program.py
```diff
@@ -9,10 +9,187 @@
...
+  ret void
+}
+
+declare i32 @llvm.nvvm.read.ptx.sreg.ctaid.x() nounwind readnone
+declare i32 @llvm.nvvm.read.ptx.sreg.ntid.x() nounwind readnone
+declare i32 @llvm.nvvm.read.ptx.sreg.tid.x() nounwind readnone
+
+!nvvm.annotations = !{!0}
+!0 = !{void (i32*)* @simple, !"kernel", i32 1}
+"""  # noqa: E501
```
leofang: ditto, I don't think this is a valid IR at all? If this is true, I suggest we just remove this IR entirely.

leofang (changes requested):
I was wondering why the parametrized tests are skipped in the CI. It took me a while to pinpoint the issue. `_get_libnvvm_version_for_tests` fails to do the job.

cuda_core/tests/test_program.py
leofang:
```
IR version 1.0 incompatible with current version 2.0\nprecheck.ll: error: incompatible IR detected. Possible mix of compiler/IR from different releases.\x00
```

leofang:
btw plz also add a release note entry to `cuda_core/docs/source/release/0.X.Y-notes.rst` 😄

cuda_core/tests/test_program.py
abhilash1910: Yes this IR is malformed ; fixed it.

cuda_core/tests/test_program.py
abhilash1910: Yes this is modified fully.

cuda_core/cuda/core/experimental/_program.py
abhilash1910: Ack, modified. refactor can be followed up , thanks

abhilash1910:
pre-commit.ci autofix

abhilash1910:
pre-commit.ci autofix

cuda_core/cuda/core/experimental/_program.py
leofang: Sorry, but you added it to the wrong place... lol 

We need the exception manager to be added here, not in the test suite.

cuda_core/tests/test_program.py
```diff
@@ -3,16 +3,243 @@
...
 from cuda.core.experimental._module import Kernel, ObjectCode
 from cuda.core.experimental._program import Program, ProgramOptions
+from cuda.core.experimental._utils.cuda_utils import driver, handle_return
 
+cuda_driver_version = handle_return(driver.cuDriverGetVersion())
 is_culink_backend = _linker._decide_nvjitlink_or_driver()
 
 
+@contextmanager
+def _nvvm_exception_manager(nvvm, program_handle):
```
leofang: wrong place

cuda_core/tests/test_program.py
```diff
@@ -3,16 +3,243 @@
...
+    from cuda.core.experimental._utils.cuda_utils import driver, handle_return
+
+    _cuda_driver_version = handle_return(driver.cuDriverGetVersion())
+except Exception:
+    _cuda_driver_version = 0
+
+_libnvvm_version = None
+_libnvvm_version_attempted = False
+
+precheck_nvvm_ir = """target triple = "nvptx64-unknown-cuda"
```
leofang, line 43: Please don't use such a complicated IR for simple purposes. The previous IR was fine, it just missed the version attribute. I verified it locally.

cuda_core/tests/test_program.py
leofang: This is not addressed

cuda_core/docs/source/release/0.X.Y-notes.rst
```diff
@@ -28,6 +28,8 @@ New features
 
 - Added :attr:`Device.arch` property that returns the compute capability as a string (e.g., '75' for CC 7.5), providing a convenient alternative to manually concatenating the compute capability tuple.
 - CUDA 13.x testing support through new ``test-cu13`` dependency group.
+- Added NVVM IR support for cuda_core frontend, which would allow cuda_python to take NVVM IR as an input format. New tests are added inside
+ ``test_program.py``.
```
leofang:
```suggestion
- Added NVVM IR support to :class:`Program`. NVVM IR is now understood with ``code_type="nvvm"``.
```

cuda_core/tests/test_program.py
leofang: I am not sure I understand....? 🤔 Could you elaborate?

cuda_core/cuda/core/experimental/_program.py
abhilash1910: Yes will fix this.

cuda_core/cuda/core/experimental/_program.py
rwgk: I'll put something together asap (today).

Roughly, I think we can do something like this:

```
try:
    from cuda.bindings import nvvm
except ModuleNotFoundError:
    # now we know the module isn't available, without having to look at specific versions numbers
except cuda.pathfinder.DynamicLibNotFoundError:
    # now we know cuda-bindings has the nvvm bindings, but libnvvm isn't available
 ```
 
 I want try that out locally first.

File: cuda_core/cuda/core/experimental/_program.py [hunk 42]
```diff
@@ -20,12 +20,60 @@
...
     is_nested_sequence,
     is_sequence,
     nvrtc,
 )
 
+_nvvm_module = None
+_nvvm_import_attempted = False
+
+
+def _get_nvvm_module():
```
kkraus14, line 61: Should we make nvvm a required module of `cuda.core` similar to how we treat nvrtc? Beyond the version check is there a reason we need to treat it differently than nvrtc?

cuda_core/tests/test_program.py
abhilash1910: Oh yes this was an oversight, this should be taking from the nvvm ir _Version

abhilash1910:
pre-commit.ci autofix

cuda_core/cuda/core/experimental/_program.py
```diff
@@ -513,6 +634,32 @@ def compile(self, target_type, name_expressions=(), logs=None):
...
+            nvvm.get_compiled_result(self._mnff.handle, data)
+
+            if logs is not None:
+                logsize = nvvm.get_program_log_size(self._mnff.handle)
+                if logsize > 1:
+                    log = bytearray(logsize)
+                    nvvm.get_program_log(self._mnff.handle, log)
+                    logs.write(log.decode("utf-8", errors="backslashreplace"))
+
+            data_bytes = bytes(data)
```
kkraus14: This creates a deep copy of the bytearray, is this avoidable in any way? It looks like for the nvrtc backend we create `data` as bytes and pass it down through the underlying C APIs and modify it directly (maybe against the intended Python behavior?)

cuda_core/cuda/core/experimental/_program.py
leofang: This treatment is consistent (concept-wise, not implementation-wise) with that for nvJitLink. This needs to exist because NVRTC bindings have existed since day 1 of cuda-bindings, but it's not the case for nvJitLink/NVVM.

cuda_core/tests/test_program.py
```diff
@@ -9,10 +9,189 @@
...
+define void @dummy_kernel() {
+  entry:
+    ret void
+}
+
+!nvvm.annotations = !{{!0}}
+!0 = !{{void (i32*)* @simple, !"kernel", i32 1}}
+
+!nvvmir.version = !{{!1}}
+!1 = !{{i32 {major}, i32 0, i32 {debug_major}, i32 0}}
```
leofang:
```suggestion
!1 = !{{i32 {major}, i32 {minor}, i32 {debug_major}, i32 {debug_minor}}}
```

cuda_core/tests/test_program.py
```diff
@@ -9,10 +9,189 @@
...
+
+        try:
+            from cuda.bindings.utils import get_minimal_required_cuda_ver_from_ptx_ver, get_ptx_ver
+        except ImportError:
+            _libnvvm_version = None
+            return _libnvvm_version
+
+        program = nvvm.create_program()
+        try:
+            precheck_ir_bytes = precheck_nvvm_ir.encode("utf-8")
```
leofang, line 94: Please make sure the changes are tested locally.
```suggestion
            major, minor, debug_major, debug_minor = nvvm.ir_version()
            precheck_nvvm_ir = precheck_nvvm_ir.format(major=major, minor=minor, debug_major=debug_major, debug_minor=debug_minor)
            precheck_ir_bytes = precheck_nvvm_ir.encode("utf-8")
```

leofang (changes requested):

cuda_core/cuda/core/experimental/_program.py
kkraus14: Ah, that makes sense. I look forward to the future when we can require much newer versions of `cuda-bindings` 😄

cuda_core/cuda/core/experimental/_program.py
leofang: Seems like a good catch. I think `add_module_to_program` should support both `bytes` and `bytearray` (by design of the codegen):
```python
>>> from cuda.bindings import nvvm
>>> prog = nvvm.create_program()
>>> nvvm.add_module_to_program(prog, b"123", 3, "ok")
>>> nvvm.add_module_to_program(prog, bytearray(b"123"), 3, "ok")
>>> 
```
https://github.com/NVIDIA/cuda-python/blob/7a24bd80b908e42f45ba610445998af383b2bbd3/cuda_bindings/cuda/bindings/nvvm.pyx#L146

For NVRTC we had to trick the binding because the assumption was different (the signature is `char*`, so a type conversion happens at the call site): 
https://github.com/NVIDIA/cuda-python/blob/7a24bd80b908e42f45ba610445998af383b2bbd3/cuda_bindings/cuda/bindings/nvrtc.pyx.in#L231

cuda_core/cuda/core/experimental/_program.py
rwgk: This POC code should be "industrial-strength solid":

```python
import importlib

# hard dependencies
import cuda.bindings
import cuda.pathfinder


def get_nvvm_module():
    try:
        nvvm = importlib.import_module("cuda.bindings.nvvm")
    except ModuleNotFoundError as e:
        if e.name != "cuda.bindings.nvvm":
            raise  # real bug: some dependency of nvvm is missing or has issues
        print(f"cuda.bindings.nvvm module is not available.")
    else:
        try:
            nvvm.version()  # Probe that triggers loading libnvvm
        except cuda.pathfinder.DynamicLibNotFoundError as e:
            print(f"DynamicLibNotFoundError: {e}")
        else:
            print(f"Success: {nvvm!r}")


if __name__ == "__main__":
    get_nvvm_module()
```

It precisely determines if (1) the `nvvm` module is available and (2)`libnvvm` can be loaded, **without masking bugs elsewhere** (like `except ImportError` would do).

I tested it locally:

cuda-bindings 12.8.0:

```
(Cp1280) rwgk-win11.localdomain:~ $ python get_nvvm_module.py
cuda.bindings.nvvm module is not available.
```

cuda-bindings 12.9.0:

```
(Cp1290) rwgk-win11.localdomain:~ $ python get_nvvm_module.py
DynamicLibNotFoundError: Failure finding "libnvvm.so": No such file: libnvvm.so*, No such file: libnvvm.so*

(Cp1290) rwgk-win11.localdomain:~ $ CUDA_HOME=/usr/local/cuda python get_nvvm_module.py
Success: <module 'cuda.bindings.nvvm' from '/home/rgrossekunst/Cp1290/lib/python3.12/site-packages/cuda/bindings/nvvm.cpython-312-x86_64-linux-gnu.so'>
(Cp1290) rwgk-win11.localdomain:~ $
```

I'd be happy to work this into this PR after everything else has settled, or separately after this PR is merged. @leofang Please let me know.

leofang: Sounds good! Let's do it in another PR. Let me try to get this merged quickly. 

When we do it, we should:
- apply the suggestion you have above for NVVM
- apply the same approach to nvJitLink: https://github.com/NVIDIA/cuda-python/blob/7a24bd80b908e42f45ba610445998af383b2bbd3/cuda_core/cuda/core/experimental/_linker.py#L32
- ensure cuda-pathfinder is a dependency of cuda-core (today we get it indirectly through cuda-bindings)

cuda_core/cuda/core/experimental/_program.py
```diff
@@ -513,6 +634,32 @@ def compile(self, target_type, name_expressions=(), logs=None):
...
+            return ObjectCode._init(data_bytes, target_type, name=self._options.name)
```
leofang:
```suggestion
            return ObjectCode._init(data, target_type, name=self._options.name)
```

leofang: sorry this is wrong

cuda_core/cuda/core/experimental/_program.py
leofang: I looked at the wrong place. This is where it's problematic:
https://github.com/NVIDIA/cuda-python/blob/7a24bd80b908e42f45ba610445998af383b2bbd3/cuda_core/cuda/core/experimental/_module.py#L618
We can pass `bytes`/`bytearray` to any bindings, but in `ObjectCode` today we assume it's either bytes or str. Let me push a fix.

leofang: fixed in e5b5ea4c

cuda_core/cuda/core/experimental/_program.py
rwgk: Tracking this here: https://github.com/NVIDIA/cuda-python/issues/980

cuda_core/tests/test_program.py
leofang: done in 4d32276

leofang:
/ok to test [e5b5ea4](https://github.com/NVIDIA/cuda-python/pull/907/commits/e5b5ea4c7ef2d35ca13d60deb698f3a110a266dd)

leofang (approved):

rwgk (approved):
I only looked at the high-level structure; based on that, and given that we want to do #980: looks good to me.

leofang:
All CI is green except for H100, which is known to have an unusual long queue currently (see nv-gha-runners discussion). I am impatient and let me admin-merge before calling a day.

leofang:
Thanks a lot, @abhilash1910, and also @gmarkall @kkraus14 @rwgk !

abhilash1910:
Thanks a lot  @leofang , and @gmarkall @rwgk @kkraus14  for all the reviews. Will follow-up on #981.

joker-eph:
Using textual LLVM IR as an input to libNVVM is documented as deprecated, so I'm quite concerned that cuda-python is adding a new usage of this.

Another issue is that LLVM textual assembly format is more unstable than the bitcode and has no backward compatibility guarantee (contrario to the LLVM bitcode), which also likely why this was all deprecated in libNVVM.
I would think that this would be documented as, and restricted, to LLVM bitcode input only here.

Even with LLVM bitcode, there is a quite large issue of underlying compatibility with the libNVVM version: contrary to the analogy with C++ and NVRTC or PTX: the LLVM IR  isn't versioned in the same way across cuda versions.

leofang:
Hi @joker-eph:
> Using textual LLVM IR as an input to libNVVM is documented as deprecated, so I'm quite concerned that cuda-python is adding a new usage of this.

1. libNVVM has been stating that the text IR is deprecated for several years, but we have not received any notice that it'd be actually removed
2. if it is removed now, numba-cuda will break right away
3. I think both bc and text formats are already supported by `Program` through this PR, because the underlying binding does not care which format that a sequence of Python bytes contains. I did tell Abhilash offline that it’s better to get bc tested as well. I think this ball was dropped along the way.

> I would think that this would be documented as, and restricted, to LLVM bitcode input only here.

I am unable to parse this sentence 😛 Could you elaborate?

> Even with LLVM bitcode, there is a quite large issue of underlying compatibility with the libNVVM version: contrary to the analogy with C++ and NVRTC or PTX: the LLVM IR isn't versioned in the same way across cuda versions.

This is understood. See how we generate compatible IR in the test and also [this thread](https://github.com/NVIDIA/cuda-python/pull/907#discussion_r2331708460).

joker-eph:
> libNVVM has been stating that the text IR is deprecated for several years, but we have not received any notice that it'd be actually removed

I know that, I'm not sure how that addresses my comment though.

> if it is removed now, numba-cuda will break right away

Why is numba-cuda using textual IR instead of encoding to bitcode?

> I am unable to parse this sentence 😛 Could you elaborate?

Two parts to my sentence:
1) The documentation for this API should be "This is expected to use with bitcode"
2) The code could enforce that we only use bitcode.


> This is understood. See how we generate compatible IR in the test and also https://github.com/NVIDIA/cuda-python/pull/907#discussion_r2331708460.

This is understood by you maybe, what about the user that gets exposed to some unsafe APIs and that we may break in very subtle ways with future updates?
My concerns is that there is huge footgun hidden in there, and that it isn't a good API to add at all.

leofang:
> > if it is removed now, numba-cuda will break right away
> 
> Why is numba-cuda using textual IR instead of encoding to bitcode?

The short answer is that it needs to patch LLVM text IR. It'd be better if we move this conversation to either the NVIDIA/numba-cuda repo, or the internal numba dev channel, happy to continue elsewhere. It is irrelevant here.

> 1. The documentation for this API should be "This is expected to use with bitcode"
> 2. The code could enforce that we only use bitcode.

Ah ok, thanks. 1 can be added I think, with a note that the text IR is deprecated upstream. 2 is not possible as already explained earlier (we can't tell if the user provides text or bitcode IR, without leaking the magic header in the public).

> This is understood by you maybe, what about the user that gets exposed to some unsafe APIs and that we may break in very subtle ways with future updates? My concerns is that there is huge footgun hidden in there, and that it isn't a good API to add at all.

Our mission is to offer pythonic access to all CUDA components such that whatever users can do in C/C++, then can also do so without leaving Python. Unless I misunderstood what you meant, to me it sounds like the concern is "we should not make it easy to access libNVVM in Python," if so I'd wholeheartedly disagree 🙂

joker-eph:
> with a note that the text IR is deprecated upstream. 

Why "upstream"? In general I use "upstream" to refer to LLVM codebase, but are you referring to  libNVVM? This is weird to me to refer to the underlying library exposed here as "upstream": it the same product we ship and code-python should just expose it to python IMO.

> (we can't tell if the user provides text or bitcode IR, without leaking the magic header in the public).

I don't quite understand what you mean by "leaking the magic header"?  Checking if the input is bitcode seem like a trivial check to me: https://github.com/llvm/llvm-project/blob/04c01ff144a172230c053d73eb15831a4120db81/llvm/include/llvm/Bitcode/BitcodeReader.h#L244-L274


> Our mission is to offer pythonic access to all CUDA components such that whatever users can do in C/C++, then can also do so without leaving Python. Unless I misunderstood what you meant, to me it sounds like the concern is "we should not make it easy to access libNVVM in Python," if so I'd wholeheartedly disagree 🙂

You are clearly misrepresenting what I wrote: there is a difference between "exposing python access to all CUDA components" and providing direct footguns to users. This is an important part of API design is to understand these footguns and think the API to avoid them. Just the fact that you use "pythonic" shows that you're already ready to deviate from just directly binding and exposing a "raw" direct access to anything: this should be about "feature" instead.
More importantly you're again putting aside the wrinkle that this is deprecated (and this was introduced at a time where libNVVM has a **single** version of LLVM as input, the fact that it is now a moving target is entirely new).

PR #909, opened 2025-08-26
Dependabot: consolidate GitHub Actions updates to monthly

rwgk:
## Description

Update Dependabot configuration to reduce update churn.

Changes

* Switch schedule from weekly to monthly (9:00 PT).

* Limit to one open PR at a time.

* Add a group rule to bundle all minor/patch GitHub Actions updates into a single monthly PR.

This should cut down on review noise while still keeping dependencies reasonably up to date.
___

Piggy-backed: Fix existing (on `main`) pre-commit error

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,4 +7,15 @@ updates:
   - package-ecosystem: github-actions
     directory: /
     schedule:
-      interval: weekly
+      interval: "monthly"
+      time: "09:00"
+      timezone: "America/Los_Angeles"
+
+    # Keep churn down: only one open PR from this ecosystem at a time
+    open-pull-requests-limit: 1
+
+    groups:
+      actions-monthly:
+        applies-to: version-updates
+        patterns: ["*"]
+        update-types: ["minor", "patch"]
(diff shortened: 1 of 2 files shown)


rwgk:
(Intentionally no running full CI)

.github/dependabot.yml
```diff
@@ -7,4 +7,15 @@ updates:
...
+      timezone: "America/Los_Angeles"
+
+    # Keep churn down: only one open PR from this ecosystem at a time
+    open-pull-requests-limit: 1
+
+    groups:
+      actions-monthly:
+        applies-to: version-updates
+        patterns: ["*"]
+        update-types: ["minor", "patch"]
```
leofang, line 21: Q: Does this mean we disallow major updates?

rwgk: To the best of my knowledge: No

It will only not put them in one group. There will be a separate group.

Why I think that: https://chatgpt.com/share/68ae098d-5490-8008-b363-04fda01c71a4

Near the top there:

> Here’s a drop-in /.github/dependabot.yml example that (a) runs monthly, (b) groups all GitHub Actions bumps into a single PR, and **(c) keeps majors separate**

.github/dependabot.yml
```diff
@@ -7,4 +7,15 @@ updates:
   - package-ecosystem: github-actions
     directory: /
     schedule:
-      interval: weekly
+      interval: "monthly"
+      time: "09:00"
+      timezone: "America/Los_Angeles"
+
+    # Keep churn down: only one open PR from this ecosystem at a time
+    open-pull-requests-limit: 1
```
leofang, line 15: Doesn't seem like this is necessary based on the docs
- https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/optimizing-pr-creation-version-updates
- https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--

Since we group all updates into a monthly update, this should be implied. But no harm in keeping it, it seems.

leofang (approved):

leofang:
Thanks, Ralf! This is nice, the bot was a bit too noisy 😛

PR #911, opened 2025-08-27
Pre commit fixes

rwgk:
## Description
Three tiny unrelated small fixes:

* d12b9e0dbda62a0b8043592413935de75f4a18b8
* 240d637fe121be6ec564376e872ba293e567a9bc
* ed25cc72bd493aee1a42ed21ef26f09efde25c8d — Note: This undoes Dependabot PR #801. See also: PR #913

### DIFF
--- a/.github/workflows/bandit.yml
+++ b/.github/workflows/bandit.yml
@@ -20,4 +20,4 @@ jobs:
       security-events: write
     steps:
       - name: Perform Bandit Analysis
-        uses: PyCQA/bandit-action@67a458d90fa11fb1463e91e7f4c8f068b5863c7f  # v1.0.1
+        uses: PyCQA/bandit-action@8a1b30610f61f3f792fe7556e888c9d7dffa52de


.github/workflows/bandit.yml
```diff
@@ -20,4 +20,4 @@ jobs:
       security-events: write
     steps:
       - name: Perform Bandit Analysis
-        uses: PyCQA/bandit-action@67a458d90fa11fb1463e91e7f4c8f068b5863c7f  # v1.0.1
+        uses: PyCQA/bandit-action@8a1b30610f61f3f792fe7556e888c9d7dffa52de
```
kkraus14, line 23: Is this associated with a specific tag?

leofang: This came from a copy/paste from the global allowlist. At this point I am not sure it makes sense for us to dig out the corresponding version if we don't own the version bump?

kkraus14: This is the commit tied to `v1.0.0` where this would be a downgrade that also includes running on Python 3.8 which is EOL.

rwgk: At least the workflow runs with this, which seems like a strict improvement for the minute.

I want to propose to work on it more in a follow-on PR. In the meantime I figured out a simple approach to keep the workflow and pre-commit in sync (manually).

But should I add the `  # v1.0.0` comment here? (I missed that before.)

(Then you could admin merge.)

kkraus14: I guess let's add the comment for v1.0.0 here for now and then upgrade it once things are fixed at the org level

rwgk:
Ugh, I was a second too slow, disabling the auto-merge. Sorry.

kkraus14:
No worries, this should hopefully be very temporarily short-lived

PR #913, opened 2025-08-27
Sync pre-commit, workflow `bandit` versions (manually)

rwgk:
## Description

~Ensure Dependabot does not update the `PyCQA/bandit-action` version.~ (see https://github.com/NVIDIA/cuda-python/pull/913#discussion_r2319684758)

Note: Dependabot PR #801 was already undone in PR #911.

Also pin pre-commit bandit `rev` to the same version as used by the bandit workflow.

Don't skip bandit in pre-commit.ci, to ensure `cuda-python`-internal consistency.

Add `KEEP IN SYNC WITH` comments.
___

pre-commit.ci output:

```
ruff (legacy alias)......................................................Passed
ruff format..............................................................Passed
Check SPDX Headers.......................................................Passed
Prevent markdown files in docs/source directories........................Passed
check for added large files..............................................Passed
check for case conflicts.................................................Passed
check docstring is first.................................................Passed
check for merge conflicts................................................Passed
check for broken symlinks................................................Passed
check toml...............................................................Passed
check yaml...............................................................Passed
debug statements (python)................................................Passed
fix end of files.........................................................Passed
mixed line ending........................................................Passed
trim trailing whitespace.................................................Passed
rst ``code`` is two backticks............................................Passed
rst directives end with two colons.......................................Passed
rst ``inline code`` next to normal text..................................Passed
bandit...................................................................Passed
mypy-pathfinder..........................................................Passed
```
___

Relevant links for easy future reference:

* https://github.com/PyCQA/bandit-action/releases/tag/v1.0.0

* https://github.com/PyCQA/bandit/releases/tag/1.7.10


.github/dependabot.yml
```diff
@@ -19,3 +19,7 @@ updates:
         applies-to: version-updates
         patterns: ["*"]
         update-types: ["minor", "patch"]
+
+    # Prevent updates for the Bandit Action; we manage it manually according to corp policies
+    ignore:
+      - dependency-name: "PyCQA/bandit-action"
```
leofang: I have told the dependabot to skip it: https://github.com/NVIDIA/cuda-python/pull/934#issuecomment-3249694758, so I think it is safe to revert this change now.

leofang (dismissed):

.github/dependabot.yml
rwgk: Oh wow, I wasn't aware of that dependabot  feature. I'll update this PR.

leofang (approved):

PR #919, opened 2025-08-28
Search conda environment before general system search in cuda-pathfinder

### LINKED ISSUE #839: [BUG]: pathfinder `load_with_system_search` tries to load old libraries first
**This is more-or-less a sub-issue of #654**

### Is this a duplicate?

### Type of Bug

Silent Failure

### Component

cuda.pathfinder

### Describe the bug

I've been trying to work out why, when I have `cuda-nvrtc` 13.0.48 in my conda environment, pathfinder seems to always pick an old system-installed libnvrtc. First it would find one from the toolkit in `/usr/local/cuda-11.2`, then I deleted that and it picked one up from `/usr/local/cuda-12.0`.

I think the issue is that the list of NVRTC sonames to search looks like:

```python
    "nvrtc": (
        "libnvrtc.so.11.0",
        "libnvrtc.so.11.1",
        "libnvrtc.so.11.2",
        "libnvrtc.so.12",
        "libnvrtc.so.13",
    ),
```

and it will try to `dlopen()` each of these via `ctypes.CDLL` until it succeeds, then use whatever it found:

https://github.com/NVIDIA/cuda-python/blob/b87c7871d675f50fca3beab051ca383d815df8d9/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py#L101-L109

### How to Reproduce

I think these steps provide a minimal reproducer: 
 
1. Install an old CUDA toolkit (e.g. 12.0) in the usual system location, and presumably as a result of that it will show up in e.g. `ldconfig -p`.
2. Create an environment with a newer NVRTC, e.g.
   ```
   mamba create -n test-pathfinder python=3.13 cuda-python cuda-core cuda-nvrtc cuda-version=13
   ```
3. Run the following:
   ```python
   from cuda import pathfinder

   print(pathfinder.load_nvidia_dynamic_lib("nvrtc"))
   ```
   This will give
   ```python
   LoadedDL(abs_path='/usr/local/cuda-12/targets/x86_64-linux/lib/libnvrtc.so.12', was_already_loaded_from_elsewhere=False, _handle_uint=94470242743920)
   ```

I verified that the behaviour occurs because of the conditions outlined above by moving `"libnvrtc.so.13"` to the top of the list of sonames, which then makes the reproducer output

```python
LoadedDL(abs_path='/home/gmarkall/miniforge3/envs/numbadev/lib/python3.12/lib-dynload/../../libnvrtc.so.13', was_already_loaded_from_elsewhere=False, _handle_uint=95396331422112)
```

### Expected behavior

That nvrtc is loaded from the conda environment instead, giving e.g.

```python
LoadedDL(abs_path='/home/gmarkall/miniforge3/envs/numbadev/lib/python3.12/lib-dynload/../../libnvrtc.so.13', was_already_loaded_from_elsewhere=False, _handle_uint=95396331422112)
```

### Operating System

Linux

### nvidia-smi output

```
| NVIDIA-SMI 580.65.06              Driver Version: 580.65.06      CUDA Version: 13.0     |
|   0  NVIDIA RTX 3500 Ada Gene...    On  |   00000000:01:00.0 Off |                  Off |
| N/A   44C    P3            593W /   71W |      15MiB /  12282MiB |     17%      Default |

| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|    0   N/A  N/A            3035      G   /usr/lib/xorg/Xorg                        4MiB |
```

kkraus14:
## Description

Fixes #839

Adds functions for searching in conda environments for libraries and uses the functions before doing the general system search. This should resolve the problem for when you have mismatched versions between a conda environment and the underlying system CTK and pathfinder picks up the system library.

There's still a lot of TODOs with some open questions in this PR and I didn't add any testing yet, but figured I should open this to drive some conversation.

## Checklist

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
+++ b/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
@@ -173,6 +173,135 @@ def load_with_system_search(libname: str) -> Optional[LoadedDL]:
     return None
 
 
+def load_with_conda_search(libname: str) -> Optional[LoadedDL]:
+    """Try to load a library using conda search paths.
+
+    Args:
+        libname: The name of the library to load
+
+    Returns:
+        A LoadedDL object if successful, None if the library cannot be loaded
+    """
+    in_conda_build = False
+    in_conda_env = False
+    if os.getenv("CONDA_BUILD") == "1":
+        in_conda_build = True
+    elif os.getenv("CONDA_PREFIX"):
+        in_conda_env = True
+    else:
+        return None
+
+    normal_conda_lib_path = os.path.join("lib")
+    # TODO KEITH: All the libs in the targets directory are symlinked into the lib directory, do we need to search it?
+    # TODO KEITH: Should we do platform detection here to avoid extra searches? Any considerations we need to do in a
+    # cross compilation build environment?
+    nvidia_conda_target_lib_paths = [
+        os.path.join("targets", "x86_64-linux", "lib"),
+        os.path.join("targets", "sbsa-linux", "lib"),
+    ]
+    if libname == "nvvm":
+        normal_conda_lib_path = os.path.join("nvvm")
+        nvidia_conda_target_lib_paths = [
+            os.path.join("targets", "x86_64-linux", "nvvm", "lib64"),
+            os.path.join("targets", "sbsa-linux", "nvvm", "lib64"),
+        ]
+
+    for soname in get_candidate_sonames(libname):
+        if in_conda_build:
+            if prefix := os.getenv("PREFIX"):
+                for nvidia_conda_target_lib_path in nvidia_conda_target_lib_paths:
+                    prefix_target_lib_path = os.path.join(prefix, nvidia_conda_target_lib_path)
+                    if os.path.isdir(prefix_target_lib_path):
+                        soname = os.path.join(prefix_target_lib_path, soname)
+                        try:
+                            handle = _load_lib(libname, soname)
+                        except OSError:
+                            pass
+                        else:
+                            # TODO KEITH: Do we need this abs_path_for_dynamic_library call?
+                            # We're already resolving the absolute path based on the conda environment variables
+                            abs_path = abs_path_for_dynamic_library(libname, handle)
+                            if abs_path is None:
+                                raise RuntimeError(f"No expected symbol for {libname=!r}")
+                            return LoadedDL(abs_path, False, handle._handle)
+                # Only run if not found in the target lib paths
+                prefix_normal_lib_path = os.path.join(prefix, normal_conda_lib_path)
+                if os.path.isdir(prefix_normal_lib_path):
+                    soname = os.path.join(prefix_normal_lib_path, soname)
+                    try:
+                        handle = _load_lib(libname, soname)
+                    except OSError:
+                        pass
+                    else:
+                        # TODO KEITH: Do we need this abs_path_for_dynamic_library call?
+                        # We're already resolving the absolute path based on the conda environment variables
+                        abs_path = abs_path_for_dynamic_library(libname, handle)
+                        if abs_path is None:
+                            raise RuntimeError(f"No expected symbol for {libname=!r}")
+                        return LoadedDL(abs_path, False, handle._handle)
+            if build_prefix := os.getenv("BUILD_PREFIX"):
+                for nvidia_conda_target_lib_path in nvidia_conda_target_lib_paths:
+                    build_prefix_target_lib_path = os.path.join(build_prefix, nvidia_conda_target_lib_path)
+                    if os.path.isdir(build_prefix_target_lib_path):
+                        soname = os.path.join(build_prefix_target_lib_path, soname)
+                        try:
+                            handle = _load_lib(libname, soname)
+                        except OSError:
+                            pass
+                        else:
+                            # TODO KEITH: Do we need this abs_path_for_dynamic_library call?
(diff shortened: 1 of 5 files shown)


cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
```diff
@@ -173,6 +173,135 @@ def load_with_system_search(libname: str) -> Optional[LoadedDL]:
...
+                            handle = _load_lib(libname, soname)
+                        except OSError:
+                            pass
+                        else:
+                            # TODO KEITH: Do we need this abs_path_for_dynamic_library call?
+                            # We're already resolving the absolute path based on the conda environment variables
+                            abs_path = abs_path_for_dynamic_library(libname, handle)
+                            if abs_path is None:
+                                raise RuntimeError(f"No expected symbol for {libname=!r}")
+                            return LoadedDL(abs_path, False, handle._handle)
```
kkraus14, line 226: This should probably be refactored into a helper function. I copy pasted it from `load_with_system_search`

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
```diff
@@ -173,6 +173,135 @@ def load_with_system_search(libname: str) -> Optional[LoadedDL]:
...
+                for nvidia_conda_target_lib_path in nvidia_conda_target_lib_paths:
+                    prefix_target_lib_path = os.path.join(prefix, nvidia_conda_target_lib_path)
+                    if os.path.isdir(prefix_target_lib_path):
+                        soname = os.path.join(prefix_target_lib_path, soname)
+                        try:
+                            handle = _load_lib(libname, soname)
+                        except OSError:
+                            pass
+                        else:
+                            # TODO KEITH: Do we need this abs_path_for_dynamic_library call?
```
kkraus14, line 221: @rwgk was this only needed in the case we were relying on `dlopen` / windows equivalent to find the library for us?

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_windows.py
```diff
@@ -132,6 +132,64 @@ def load_with_system_search(libname: str) -> Optional[LoadedDL]:
...
+            if prefix := os.getenv("PREFIX"):
+                prefix_normal_lib_path = os.path.join(prefix, normal_conda_lib_path)
+                if os.path.isdir(prefix_normal_lib_path):
+                    dll_name = os.path.join(prefix_normal_lib_path, dll_name)
+                    handle = kernel32.LoadLibraryExW(dll_name, None, 0)
+                    if handle:
+                        # TODO KEITH: Do we need this abs_path_for_dynamic_library call?
+                        # We're already resolving the absolute path based on the conda environment variables
+                        abs_path = abs_path_for_dynamic_library(libname, handle)
+                        return LoadedDL(abs_path, False, ctypes_handle_to_unsigned_int(handle))
```
kkraus14, line 168: +1 we should probably refactor this into a helper function

cuda_pathfinder/docs/source/release/1.X.Y-notes.rst
```diff
@@ -23,3 +23,8 @@ Highlights
   - Improves stability in general and supports nvmath specifically
   - Proactive change to improve library loading consistency
   - Drops boilerplate docstrings for private functions
+
+* Add conda search path prioritization for ``load_nvidia_dynamic_lib()`` (`PR #856 <https://github.com/NVIDIA/cuda-python/pull/856>`_)
```
kkraus14, line 27:
```suggestion
* Add conda search path prioritization for ``load_nvidia_dynamic_lib()`` (`PR #919 <https://github.com/NVIDIA/cuda-python/pull/919>`_)
```

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
```diff
@@ -173,6 +173,135 @@ def load_with_system_search(libname: str) -> Optional[LoadedDL]:
...
+        libname: The name of the library to load
+
+    Returns:
+        A LoadedDL object if successful, None if the library cannot be loaded
+    """
+    in_conda_build = False
+    in_conda_env = False
+    if os.getenv("CONDA_BUILD") == "1":
+        in_conda_build = True
+    elif os.getenv("CONDA_PREFIX"):
```
rwgk, line 189: I spent a few minutes quizzing ChatGPT about the most robust way to figure out if we're in a conda environment. It produced this:
___
Typical code snippet looks like:
```
import os, sys

def in_conda_env() -> bool:
    return os.path.exists(os.path.join(sys.prefix, "conda-meta"))
```

That is generally considered the most robust test.
___

kkraus14: I had thought about this as well, but ultimately decided on using the `$CONDA_PREFIX` approach for a few reasons:
* Conda environment activation is guaranteed to set the environment variable and conda usage in general is dependent on it
* Aligned implementation between conda-build and general conda environments
* You can have a conda environment without `python`, where technically you could have a pure C++ environment and someone could want to use `cuda-pathfinder` as a utility in a C++ build pipeline where Python would potentially be coming from outside of the conda environment, in which case `sys.prefix` wouldn't properly point inside of the activated conda environment.

PR #900, opened 2025-08-25
Backport #865 to 12.9.x

mdboom:
Backport of #865 to the 12.9.x branch.

PR #836, opened 2025-08-13
[Backport] Make populating the internal symbol table thread-safe

leofang:
## Description

xref: #835

## Checklist

- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #924, opened 2025-08-28
Update with whitespace-only changes from cython-gen

mdboom:
This just commits to main the results from the latest version of the generator that are whitespace only.  I think we got in this state because some "written by hand" changes to the generated code were merged, and then the ports to the generator didn't match that exactly.  While none of these changes should have any effect, this will just help future runs of the generator apply cleanly.

### DIFF
--- a/cuda_bindings/cuda/bindings/_bindings/cydriver.pyx.in
+++ b/cuda_bindings/cuda/bindings/_bindings/cydriver.pyx.in
@@ -12,9 +12,10 @@ cimport cuda.bindings._lib.dlfcn as dlfcn
 from libc.stdint cimport intptr_t
 import os
 import sys
-import threading
 cimport cuda.bindings._bindings.loader as loader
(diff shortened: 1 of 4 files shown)


leofang (approved):

kkraus14:
Thanks @mdboom. I think we want to backport this as well?

mdboom:
> Thanks @mdboom. I think we want to backport this as well?

I don't think we have this issue on 12.9.x, since we didn't merge any hand-written changes in generated files on that branch.

leofang:
> I don't think we have this issue on 12.9.x, since we didn't merge any hand-written changes in generated files on that branch.

#836 was hand-written, but I think all whitespace changes were absorbed in the backport https://github.com/NVIDIA/cuda-python/pull/900. I confirmed rerunning the codegen for driver/runtime/nvrtc targeting 12.9.x does not generate any diff.

Issue #779, opened 2025-07-24
`Buffer` deallocation always happens on default stream


shwina:
There are two ways a buffer can be deallocated.

- Explicitly via calling Buffer.close() https://github.com/NVIDIA/cuda-python/blob/ae1cbcfb9dc631800dbeea005df8de85e3c11c02/cuda_core/cuda/core/experimental/_memory.py#L63
- Implicitly via finalizer https://github.com/NVIDIA/cuda-python/blob/ae1cbcfb9dc631800dbeea005df8de85e3c11c02/cuda_core/cuda/core/experimental/_memory.py#L43

In the implicit case, no `stream=` argument is passed to the finalizer function, which means that if a buffer gets refcounted out or gc'd, it will always be deallocated on the default stream.

Possibly we want to change this so that the `stream` passed to the `mr.allocate()` function is used for deallocation by default.

leofang:
In a number of internal discussions this issue has been brought up over and over (most recently on Sep 8 when discussing #674). We decided to align with cccl-runtime and change this to deallocate on the allocation stream, with the understanding that
1. no choice is perfect, and this choice is to serve the most common use cases (we think!) where a short-lived buffer is allocated/deallocated on the same stream
2. users are still encouraged to call `.close(stream)` (cuda.core) or `.destroy(stream)` (cccl-runtime) to preserve stream ordering explicitly 
3. this is safer when we reimplement MRs on top of those from cccl-rt

cpcloud:
@leofang Can you write down the lifetime constraints here (or link to something that spells them out) so that we can make an informed choice about the API from the POV of Python?

For example, is `lifetime(stream) >= lifetime(buffer)` (streams live at least as long as buffers)? If so, I would argue that the API should be `stream.destroy(buffer)`.

> Possibly we want to change this so that the stream passed to the `mr.allocate()` function is used for deallocation by default.

If there's a concern about which stream to dealloc a buffer on, then it seems like that dictates a clear requirement that the stream should have APIs for allocating and deallocating buffers.

```python
buf = stream.allocate()
stream.deallocate(buf)
```

I'm sketching this out, the devil is in the details here of course.

leofang:
Discussed offline, but wanna give a sketch here. 

[...]

PR #930, opened 2025-08-29
Implements IPC-enabled memory pools for Linux in DeviceMemoryResource

Andy-Jost:
Updates `cuda.core.experiment.DeviceMemoryResource` to support IPC-enabled memory pools. This change updates the DMR constructor to always create a new memory pool. The previous behavior got the current device memory pool via `cuDeviceGetMemPool`. That functionality has been moved to a static method `current`.

## Checklist

### DIFF
--- a/cuda_core/cuda/core/experimental/_memory.pyx
+++ b/cuda_core/cuda/core/experimental/_memory.pyx
@@ -11,7 +11,10 @@ from cuda.core.experimental._utils.cuda_utils cimport (
 )
 
 import abc
-from typing import Tuple, TypeVar, Union
+from typing import Tuple, TypeVar, Union, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from cuda.core.experimental._device import Device
 
 from cuda.core.experimental._dlpack import DLDeviceType, make_py_capsule
 from cuda.core.experimental._stream import Stream, default_stream
@@ -329,15 +332,17 @@ class DeviceMemoryResource(MemoryResource):
 
     Parameters
     ----------
-    device_id : int
-        Device ordinal for which a memory resource is constructed. The mempool that is
-        set to *current* on ``device_id`` is used. If no mempool is set to current yet,
-        the driver would use the *default* mempool on the device.
+    device_id : int | Device
+        Device or Device ordinal for which a memory resource is constructed.
+	The mempool that is set to *current* on ``device_id`` is used. If no
+	mempool is set to current yet, the driver would use the *default*
+	mempool on the device.
     """
 
     __slots__ = ("_dev_id",)
 
-    def __init__(self, device_id: int):
+    def __init__(self, device_id: int | Device):
+        device_id = getattr(device_id, 'device_id', device_id)
         err, self._handle = driver.cuDeviceGetMemPool(device_id)
         raise_if_driver_error(err)
         self._dev_id = device_id
@@ -481,9 +486,9 @@ class LegacyPinnedMemoryResource(MemoryResource):
 class _SynchronousMemoryResource(MemoryResource):
     __slots__ = ("_dev_id",)
 
-    def __init__(self, device_id):
+    def __init__(self, device_id : int | Device):
         self._handle = None
-        self._dev_id = device_id
+        self._dev_id = getattr(device_id, 'device_id', device_id)
 
     def allocate(self, size, stream=None) -> Buffer:
         err, ptr = driver.cuMemAlloc(size)
--- a/cuda_core/cuda/core/experimental/_memory.pyx
+++ b/cuda_core/cuda/core/experimental/_memory.pyx
@@ -4,21 +4,22 @@
 
 from __future__ import annotations
 
-from libc.stdint cimport uintptr_t
-
 from cuda.core.experimental._utils.cuda_utils cimport (
     _check_driver_error as raise_if_driver_error,
 )
 
-import abc
-from typing import Tuple, TypeVar, Union, TYPE_CHECKING
-
-if TYPE_CHECKING:
-    from cuda.core.experimental._device import Device
-
 from cuda.core.experimental._dlpack import DLDeviceType, make_py_capsule
 from cuda.core.experimental._stream import Stream, default_stream
 from cuda.core.experimental._utils.cuda_utils import driver
+from functools import wraps
+from libc.stdint cimport uintptr_t
+from typing import Tuple, TypeVar, Union, TYPE_CHECKING
+import abc
+import os
+import platform
+
+if TYPE_CHECKING:
+    from cuda.core.experimental._device import Device
 
 # TODO: define a memory property mixin class and make Buffer and
 # MemoryResource both inherit from it
@@ -326,43 +327,200 @@ class MemoryResource(abc.ABC):
         """
         ...
 
+# IPC is currently only supported on Linux. On other platforms, the IPC handle
+# type is set equal to the no-IPC handle type.
+
+_NOIPC_HANDLE_TYPE = driver.CUmemAllocationHandleType.CU_MEM_HANDLE_TYPE_NONE
+_IPC_HANDLE_TYPE = driver.CUmemAllocationHandleType.CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR \
+    if platform.system() == "Linux" else _NOIPC_HANDLE_TYPE
+
+def _requires_ipc(func):
+    """Decorator to ensure IPC support is enabled before executing a method."""
+    @wraps(func)
+    def wrapper(self, *args, **kwargs):
+        if not self.is_ipc_enabled:
+            raise RuntimeError("Memory resource is not IPC-enabled")
+        return func(self, *args, **kwargs)
+    return wrapper
+
+def _def_mempool_attr_property(scope: dict, name: str, property_type: type, doc: str):
+    """Defines a property in the supplied scope for accessing a memory pool attribute.
+
+    Args:
+        name: The name of the property (e.g., 'reuse_follow_event_dependencies').
+        doc: The docstring for the property.
+        property_type: The return type of the property (e.g., bool, int).
+
+    Returns:
+        property: A property object that retrieves the attribute value.
+    """
+    attr_enum = getattr(driver.CUmemPool_attribute, f"CU_MEMPOOL_ATTR_{name.upper()}")
+    def getter(self):
+        err, value = driver.cuMemPoolGetAttribute(self.handle, attr_enum)
+        raise_if_driver_error(err)
+        return property_type(value)
+
+    getter.__doc__ = doc
+    scope[name] = property(getter)
+
+class IPCBufferDescriptor:
+    """Serializable object describing a buffer that can be shared between processes."""
+    def __init__(self, reserved: bytes, size: int):
+        self.reserved = reserved
+        self._size = size
+
+    def __reduce__(self):
+        # This is subject to change if the CumemPoolPtrExportData struct/object changes.
+        return (self._reconstruct, (self.reserved, self._size))
+
+    @property
+    def size(self):
+        return self._size
+
+    @classmethod
+    def _reconstruct(cls, reserved, size):
+        instance = cls(reserved, size)
+        return instance
+
 
 class DeviceMemoryResource(MemoryResource):
-    """Create a device memory resource that uses the driver's stream-ordered memory pool.
+    """Create a device memory resource that uses a new stream-ordered memory pool.
 
     Parameters
     ----------
     device_id : int | Device
         Device or Device ordinal for which a memory resource is constructed.
-	The mempool that is set to *current* on ``device_id`` is used. If no
-	mempool is set to current yet, the driver would use the *default*
-	mempool on the device.
+
+    max_size : int
+        Maximum pool size. When set to 0, defaults to a system dependent value.
+
+    ipc_enabled : bool
+        Specifies whether to create an IPC-enabled memory pool. When set to
+        `True`, the memory pool and its allocations can be shared with other
+        processes.
     """
 
-    __slots__ = ("_dev_id",)
+    __slots__ = "_dev_id", "_mempool_handle", "_ipc_handle_type", "_mempool_owned"
 
-    def __init__(self, device_id: int | Device):
+    def __init__(self, device_id: int | Device, max_size: int = 0, ipc_enabled: bool = False):
         device_id = getattr(device_id, 'device_id', device_id)
-        err, self._handle = driver.cuDeviceGetMemPool(device_id)
-        raise_if_driver_error(err)
+
+        if ipc_enabled and _IPC_HANDLE_TYPE == _NOIPC_HANDLE_TYPE:
+            raise RuntimeError("IPC is not available on {platform.system()}")
+
+        properties = driver.CUmemPoolProps()
+        properties.allocType = driver.CUmemAllocationType.CU_MEM_ALLOCATION_TYPE_PINNED
+        properties.handleTypes = _IPC_HANDLE_TYPE if ipc_enabled else _NOIPC_HANDLE_TYPE
+        properties.location = driver.CUmemLocation()
+        properties.location.id = device_id
+        properties.location.type = driver.CUmemLocationType.CU_MEM_LOCATION_TYPE_DEVICE
+        properties.maxSize = max_size
+        properties.win32SecurityAttributes = 0
+        properties.usage = 0
+
         self._dev_id = device_id
+        self._mempool_handle = None
+        self._ipc_handle_type = properties.handleTypes
+        self._mempool_owned = True
+
+        err, self._mempool_handle = driver.cuMemPoolCreate(properties)
+        raise_if_driver_error(err)
+
+    def __del__(self):
+        self.close()
+
+    def close(self):
+        if self and self._mempool_owned:
+            err, = driver.cuMemPoolDestroy(self._mempool_handle)
+            raise_if_driver_error(err)
+            self._dev_id = None
+            self._mempool_handle = None
+            self._ipc_handle_type = _NOIPC_HANDLE_TYPE
+            self._mempool_owned = False
+
+    def __bool__(self):
+      return self._mempool_handle is not None
+
+    @classmethod
+    def current(cls, device_id: int | Device):
+        """Create a device memory resource that uses the driver's current stream-ordered memory pool.
+
+        The memory pool set to *current* on ``device_id`` is used. If no
+        memory pool is set to current, the driver would use the *default*
+        memory pool on the device.
+
+        The returned device memory resource does not own the memory pool, and
+        closing it has no effect.
+        """
+        device_id = getattr(device_id, 'device_id', device_id)
+
+        self = cls.__new__(cls)
+        self._dev_id = device_id
+        self._mempool_handle = None
+        self._ipc_handle_type = _NOIPC_HANDLE_TYPE # FIXME: need query this?
+        self._mempool_owned = True
+
+        err, self._mempool_handle = driver.cuDeviceGetMemPool(self.device_id)
+        raise_if_driver_error(err)
 
         # Set a higher release threshold to improve performance when there are no active allocations.
         # By default, the release threshold is 0, which means memory is immediately released back
         # to the OS when there are no active suballocations, causing performance issues.
         # Check current release threshold
         err, current_threshold = driver.cuMemPoolGetAttribute(
-            self._handle, driver.CUmemPool_attribute.CU_MEMPOOL_ATTR_RELEASE_THRESHOLD
+            self._mempool_handle, driver.CUmemPool_attribute.CU_MEMPOOL_ATTR_RELEASE_THRESHOLD
         )
         raise_if_driver_error(err)
         # If threshold is 0 (default), set it to maximum to retain memory in the pool
         if int(current_threshold) == 0:
             err, = driver.cuMemPoolSetAttribute(
-                self._handle,
+                self._mempool_handle,
                 driver.CUmemPool_attribute.CU_MEMPOOL_ATTR_RELEASE_THRESHOLD,
                 driver.cuuint64_t(0xFFFFFFFFFFFFFFFF),
             )
             raise_if_driver_error(err)
+        return self
+
+    @classmethod
+    def from_shared_handle(cls, device_id: int | Device, shared_handle: int):
+        device_id = getattr(device_id, 'device_id', device_id)
+
+        self = cls.__new__(cls)
+        self._dev_id = device_id
+        self._mempool_handle = None
+        self._ipc_handle_type = _IPC_HANDLE_TYPE
+        self._mempool_owned = True
+
+        err, self._mempool_handle = driver.cuMemPoolImportFromShareableHandle(shared_handle, _IPC_HANDLE_TYPE, 0)
+        raise_if_driver_error(err)
+
+        return self
+
+    @_requires_ipc
+    def get_shared_handle(self):
+        err, shared_handle = driver.cuMemPoolExportToShareableHandle(self._mempool_handle, _IPC_HANDLE_TYPE, 0)
+        raise_if_driver_error(err)
+        return shared_handle
+
+    def close_shared_handle(self, shared_handle):
+        assert self._ipc_handle_type == driver.CUmemAllocationHandleType.CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR
+        os.close(shared_handle)
+
+    @_requires_ipc
+    def export_buffer(self, buffer: Buffer):
+        """Export a buffer allocated from this pool for sharing between processes."""
+        err, ptr = driver.cuMemPoolExportPointer(buffer.handle)
+        raise_if_driver_error(err)
+        return IPCBufferDescriptor(ptr.reserved, buffer.size)
+
+    @_requires_ipc
+    def import_buffer(self, ipc_buffer: IPCBufferDescriptor) -> Buffer:
+        """Import a buffer that was exported from another process."""
+        share_data = driver.CUmemPoolPtrExportData()
+        share_data.reserved = ipc_buffer.reserved
+        err, ptr = driver.cuMemPoolImportPointer(self._mempool_handle, share_data)
+        raise_if_driver_error(err)
+        return Buffer.from_handle(ptr, ipc_buffer.size, self)
 
     def allocate(self, size_t size, stream: Stream = None) -> Buffer:
         """Allocate a buffer of the requested size.
@@ -383,7 +541,7 @@ class DeviceMemoryResource(MemoryResource):
         """
         if stream is None:
             stream = default_stream()
-        err, ptr = driver.cuMemAllocFromPoolAsync(size, self._handle, stream.handle)
+        err, ptr = driver.cuMemAllocFromPoolAsync(size, self._mempool_handle, stream.handle)
         raise_if_driver_error(err)
         return Buffer._init(ptr, size, self)
 
@@ -406,6 +564,10 @@ class DeviceMemoryResource(MemoryResource):
         raise_if_driver_error(err)
 
     @property
+    def handle(self):
+        return self._mempool_handle
+
+    @property
     def is_device_accessible(self) -> bool:
         """bool: this memory resource provides device-accessible buffers."""
         return True
@@ -420,6 +582,19 @@ class DeviceMemoryResource(MemoryResource):
         """int: the associated device ordinal."""
         return self._dev_id
 
+    @property
+    def is_ipc_enabled(self):
+        return self._ipc_handle_type != _NOIPC_HANDLE_TYPE
+
+    # Define additional properties corresponding to mempool attributes.
+    _def_mempool_attr_property(locals(), "reuse_follow_event_dependencies", bool, "Allow memory to be reused when there are event dependencies between streams.")
+    _def_mempool_attr_property(locals(), "reuse_allow_opportunistic", bool, "Allow reuse of completed frees without dependencies.")
+    _def_mempool_attr_property(locals(), "reuse_allow_internal_dependencies", bool, "Allow insertion of new stream dependencies for memory reuse.")
+    _def_mempool_attr_property(locals(), "release_threshold", int, "Amount of reserved memory to hold before OS release.")
+    _def_mempool_attr_property(locals(), "reserved_mem_current", int, "Current amount of backing memory allocated.")
+    _def_mempool_attr_property(locals(), "reserved_mem_high", int, "High watermark of backing memory allocated.")
+    _def_mempool_attr_property(locals(), "used_mem_current", int, "Current amount of memory in use.")
+    _def_mempool_attr_property(locals(), "used_mem_high", int, "High watermark of memory in use.")
 
 class LegacyPinnedMemoryResource(MemoryResource):
     """Create a pinned memory resource that uses legacy cuMemAllocHost/cudaMallocHost
@@ -513,3 +688,6 @@ class _SynchronousMemoryResource(MemoryResource):
     @property
     def device_id(self) -> int:
         return self._dev_id
+
+del _requires_ipc
+del _def_mempool_attr_property
--- a/cuda_core/tests/test_memory.py
+++ b/cuda_core/tests/test_memory.py
@@ -6,13 +6,19 @@ try:
 except ImportError:
     from cuda import cuda as driver
 
+import array
 import ctypes
+import multiprocessing
+import platform
 
 import pytest
 
 from cuda.core.experimental import Buffer, Device, DeviceMemoryResource, MemoryResource
 from cuda.core.experimental._memory import DLDeviceType
-from cuda.core.experimental._utils.cuda_utils import handle_return
+from cuda.core.experimental._utils.cuda_utils import get_binding_version, handle_return
+
+if platform.system() == "Linux":
+    from socket import AF_UNIX, CMSG_LEN, SCM_RIGHTS, SOCK_DGRAM, SOL_SOCKET, socketpair
 
 
 class DummyDeviceMemoryResource(MemoryResource):
@@ -277,11 +283,318 @@ def test_device_memory_resource_initialization():
 
         # Verify basic properties
         assert mr.device_id == device.device_id
-        assert mr.is_device_accessible is True
-        assert mr.is_host_accessible is False
+        assert mr.is_device_accessible
+        assert not mr.is_host_accessible
+        assert not mr.is_ipc_enabled
 
         # Test allocation/deallocation works
         buffer = mr.allocate(1024)
         assert buffer.size == 1024
         assert buffer.device_id == device.device_id
         buffer.close()
+
+
+def test_mempool():
+    if get_binding_version() < (12, 0):
+        pytest.skip("Test requires CUDA 12 or higher")
+    device = Device()
+    device.set_current()
+
+    if not device.properties.memory_pools_supported:
+        pytest.skip("Device does not support mempool operations")
+
+    pool_size = 2097152  # 2MB size
+
+    # Test basic pool creation
+    mr = DeviceMemoryResource(device, pool_size, ipc_enabled=False)
+    assert mr.device_id == device.device_id
+    assert mr.is_device_accessible
+    assert not mr.is_host_accessible
+    assert not mr.is_ipc_enabled
+
+    # Test allocation and deallocation
+    buffer1 = mr.allocate(1024)
+    assert buffer1.handle != 0
+    assert buffer1.size == 1024
+    assert buffer1.memory_resource == mr
+    buffer1.close()
+
+    # Test multiple allocations
+    buffer1 = mr.allocate(1024)
+    buffer2 = mr.allocate(2048)
+    assert buffer1.handle != buffer2.handle
+    assert buffer1.size == 1024
+    assert buffer2.size == 2048
+    buffer1.close()
+    buffer2.close()
+
+    # Test stream-based allocation
+    stream = device.create_stream()
+    buffer = mr.allocate(1024, stream=stream)
+    assert buffer.handle != 0
+    buffer.close()
+
+    # Test memory copying between buffers from same pool
+    src_buffer = mr.allocate(64)
+    dst_buffer = mr.allocate(64)
+    stream = device.create_stream()
+    src_buffer.copy_to(dst_buffer, stream=stream)
+    device.sync()
+    dst_buffer.close()
+    src_buffer.close()
+
+    # Test error cases
+    # Test IPC operations are disabled
+    buffer = mr.allocate(64)
+    ipc_error_msg = "Memory resource is not IPC-enabled"
+
+    with pytest.raises(RuntimeError, match=ipc_error_msg):
+        mr.get_shared_handle()
+
+    with pytest.raises(RuntimeError, match=ipc_error_msg):
+        mr.export_buffer(buffer)
+
+    with pytest.raises(RuntimeError, match=ipc_error_msg):
+        mr.import_buffer(None)
+
+    buffer.close()
+
+
+@pytest.mark.parametrize(
+    "property_name,expected_type",
+    [
+        ("reuse_follow_event_dependencies", bool),
+        ("reuse_allow_opportunistic", bool),
+        ("reuse_allow_internal_dependencies", bool),
+        ("release_threshold", int),
+        ("reserved_mem_current", int),
+        ("reserved_mem_high", int),
+        ("used_mem_current", int),
+        ("used_mem_high", int),
+    ],
+)
+def test_mempool_properties(property_name, expected_type):
+    """Test all properties of the DeviceMemoryResource class."""
+    # skip test if cuda version is less than 12
+    if get_binding_version() < (12, 0):
+        pytest.skip("Test requires CUDA 12 or higher")
+
+    device = Device()
+    device.set_current()
+
+    if not device.properties.memory_pools_supported:
+        pytest.skip("Device does not support mempool operations")
+
+    pool_size = 2097152  # 2MB size
+    for ipc_enabled in [True, False]:
+        mr = DeviceMemoryResource(device, pool_size, ipc_enabled=ipc_enabled)
+        assert mr.is_ipc_enabled == ipc_enabled
+
+        try:
+            # Get the property value
+            value = getattr(mr, property_name)
+
+            # Test type
+            assert isinstance(value, expected_type), f"{property_name} should return {expected_type}, got {type(value)}"
+
+            # Test value constraints
+            if expected_type is int:
+                assert value >= 0, f"{property_name} should be non-negative"
+
+            # Test memory usage properties with actual allocations
+            if property_name in ["reserved_mem_current", "used_mem_current"]:
+                # Allocate some memory and check if values increase
+                initial_value = value
+                buffer = None
+                try:
+                    buffer = mr.allocate(1024)
+                    new_value = getattr(mr, property_name)
+                    assert new_value >= initial_value, f"{property_name} should increase or stay same after allocation"
+                finally:
+                    if buffer is not None:
+                        buffer.close()
+
+            # Test high watermark properties
+            if property_name in ["reserved_mem_high", "used_mem_high"]:
+                # High watermark should never be less than current
+                current_prop = property_name.replace("_high", "_current")
+                current_value = getattr(mr, current_prop)
+                assert value >= current_value, f"{property_name} should be >= {current_prop}"
+
+        finally:
+            # Ensure we allocate and deallocate a small buffer to flush any pending operations
+            flush_buffer = mr.allocate(64)
+            flush_buffer.close()
+
+
+def mempool_child_process(importer, queue):
+    try:
+        device = Device()
+        device.set_current()
+        stream = device.create_stream()
+
+        # Get the shared handle differently based on platform
+        if platform.system() == "Windows":
+            shared_handle = queue.get()  # On Windows, we pass the handle through the queue
+        else:
+            # Unix socket handle transfer
+            fds = array.array("i")
+            _, ancdata, _, _ = importer.recvmsg(0, CMSG_LEN(fds.itemsize))
+            assert len(ancdata) == 1
+            cmsg_level, cmsg_type, cmsg_data = ancdata[0]
+            assert cmsg_level == SOL_SOCKET and cmsg_type == SCM_RIGHTS
+            fds.frombytes(cmsg_data[: len(cmsg_data) - (len(cmsg_data) % fds.itemsize)])
+            shared_handle = int(fds[0])
+
+        mr = DeviceMemoryResource.from_shared_handle(device, shared_handle)
+        ipc_buffer = queue.get()  # Get exported buffer data
+        buffer = mr.import_buffer(ipc_buffer)
+
+        # Create a new buffer to verify data using unified memory
+        unified_mr = DummyUnifiedMemoryResource(device)
+        verify_buffer = unified_mr.allocate(64)
+
+        # Copy data from imported buffer to verify contents
+        verify_buffer.copy_from(buffer, stream=stream)
+        device.sync()
+
+        # Verify the data matches expected pattern
+        verify_ptr = ctypes.cast(int(verify_buffer.handle), ctypes.POINTER(ctypes.c_byte))
+        for i in range(64):
+            assert ctypes.c_byte(verify_ptr[i]).value == ctypes.c_byte(i).value, f"Data mismatch at index {i}"
+
+        # Write new pattern to the buffer using unified memory
+        src_buffer = unified_mr.allocate(64)
+        src_ptr = ctypes.cast(int(src_buffer.handle), ctypes.POINTER(ctypes.c_byte))
+        for i in range(64):
+            src_ptr[i] = ctypes.c_byte(255 - i)  # Write inverted pattern
+
+        # Copy new pattern to the IPC buffer
+        buffer.copy_from(src_buffer, stream=stream)
+        device.sync()
+
+        verify_buffer.close()
+        src_buffer.close()
+        buffer.close()
+
+        queue.put(True)
+    except Exception as e:
+        # Capture the full traceback
+        tb_str = "".join(traceback.format_exception(type(e), e, e.__traceback__))
+        queue.put((e, tb_str))
+
+
+def test_ipc_mempool():
+    if get_binding_version() < (12, 0):
+        pytest.skip("Test requires CUDA 12 or higher")
+
+    # Check if IPC is supported on this platform/device
+    device = Device()
+    device.set_current()
+    if not device.properties.memory_pools_supported:
+        pytest.skip("Device does not support mempool operations")
+
+    if platform.system() == "Windows":
+        pytest.skip("IPC is not supported on Windows")
+
+    # Set multiprocessing start method before creating any multiprocessing objects
+    multiprocessing.set_start_method("spawn", force=True)
+
+    stream = device.create_stream()
+    pool_size = 2097152  # 2MB size
+    mr = DeviceMemoryResource(device, pool_size, ipc_enabled=True)
+    assert mr.is_ipc_enabled
+
+    # Create socket pair for handle transfer (only on Unix systems)
+    exporter = None
+    importer = None
+    if platform.system() == "Linux":
+        exporter, importer = socketpair(AF_UNIX, SOCK_DGRAM)
+
+    queue = multiprocessing.Queue()
+    process = None
+
+    try:
+        shared_handle = mr.get_shared_handle()
+
+        # Allocate and export memory
+        buffer = mr.allocate(64)
+
+        try:
+            # Fill buffer with test pattern using unified memory
+            unified_mr = DummyUnifiedMemoryResource(device)
+            src_buffer = unified_mr.allocate(64)
+            try:
+                src_ptr = ctypes.cast(int(src_buffer.handle), ctypes.POINTER(ctypes.c_byte))
+                for i in range(64):
+                    src_ptr[i] = ctypes.c_byte(i)
+
+                buffer.copy_from(src_buffer, stream=stream)
+                device.sync()
+            finally:
+                src_buffer.close()
+
+            # Export buffer for IPC
+            ipc_buffer = mr.export_buffer(buffer)
+
+            # Start child process
+            process = multiprocessing.Process(
+                target=mempool_child_process, args=(importer if platform.system() == "Linux" else None, queue)
+            )
+            process.start()
+
+            # Send handles to child process
+            if platform.system() == "Windows":
+                queue.put(shared_handle)  # Send handle through queue on Windows
+            else:
+                # Use Unix socket for handle transfer
+                exporter.sendmsg([], [(SOL_SOCKET, SCM_RIGHTS, array.array("i", [shared_handle]))])
+
+            queue.put(ipc_buffer)
+
+            # Wait for child process
+            process.join(timeout=10)
+            assert process.exitcode == 0
+
+            # Check for exceptions
+            if not queue.empty():
+                result = queue.get()
+                if isinstance(result, tuple):
+                    exception, traceback_str = result
+                    print("\nException in child process:")
+                    print(traceback_str)
+                    raise exception
+                assert result is True
+
+            # Verify child process wrote the inverted pattern using unified memory
+            verify_buffer = unified_mr.allocate(64)
+            try:
+                verify_buffer.copy_from(buffer, stream=stream)
+                device.sync()
+
+                verify_ptr = ctypes.cast(int(verify_buffer.handle), ctypes.POINTER(ctypes.c_byte))
+                for i in range(64):
+                    assert ctypes.c_byte(verify_ptr[i]).value == ctypes.c_byte(255 - i).value, (
+                        f"Child process data not reflected in parent at index {i}"
+                    )
+            finally:
+                verify_buffer.close()
+
+        finally:
+            mr.close_shared_handle(shared_handle)
+            buffer.close()
+
+    finally:
+        # Clean up all resources
+        if process is not None and process.is_alive():
+            process.terminate()
+            process.join(timeout=1)
+        queue.close()
+        queue.join_thread()
+        if exporter is not None:
+            exporter.close()
+        if importer is not None:
+            importer.close()
+        # Flush any pending operations
+        flush_buffer = mr.allocate(64)
+        flush_buffer.close()
--- a/cuda_core/tests/test_memory.py
+++ b/cuda_core/tests/test_memory.py
@@ -270,16 +270,18 @@ def test_device_memory_resource_initialization():
         pytest.skip("memory pools not supported")
     device.set_current()
 
-    # This should succeed and configure the memory pool release threshold
-    mr = DeviceMemoryResource(device.device_id)
+    # This should succeed and configure the memory pool release threshold.
+    # The resource can be constructed from either a device or device ordinal.
+    for device_arg in [device, device.device_id]:
+        mr = DeviceMemoryResource(device_arg)
 
-    # Verify basic properties
-    assert mr.device_id == device.device_id
-    assert mr.is_device_accessible is True
-    assert mr.is_host_accessible is False
+        # Verify basic properties
+        assert mr.device_id == device.device_id
+        assert mr.is_device_accessible is True
+        assert mr.is_host_accessible is False
 
-    # Test allocation/deallocation works
-    buffer = mr.allocate(1024)
-    assert buffer.size == 1024
-    assert buffer.device_id == device.device_id
-    buffer.close()
+        # Test allocation/deallocation works
+        buffer = mr.allocate(1024)
+        assert buffer.size == 1024
+        assert buffer.device_id == device.device_id
+        buffer.close()
--- a/cuda_core/cuda/core/experimental/_device.py
+++ b/cuda_core/cuda/core/experimental/_device.py
@@ -1003,7 +1003,7 @@ class Device:
                         )
                     )
                 ) == 1:
-                    dev._mr = DeviceMemoryResource(dev_id)
+                    dev._mr = DeviceMemoryResource.current(dev_id)
                 else:
                     dev._mr = _SynchronousMemoryResource(dev_id)
 


cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -323,41 +327,200 @@ class MemoryResource(abc.ABC):
...
+        self._ipc_handle_type = properties.handleTypes
+        self._mempool_owned = True
+
+        err, self._mempool_handle = driver.cuMemPoolCreate(properties)
         raise_if_driver_error(err)
+
+    def __del__(self):
+        self.close()
+
+    def close(self):
```
Andy-Jost, line 648: needs docstring

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -323,41 +327,200 @@ class MemoryResource(abc.ABC):
...
-                self._handle,
+                self._mempool_handle,
                 driver.CUmemPool_attribute.CU_MEMPOOL_ATTR_RELEASE_THRESHOLD,
                 driver.cuuint64_t(0xFFFFFFFFFFFFFFFF),
             )
             raise_if_driver_error(err)
+        return self
+
+    @classmethod
+    def from_shared_handle(cls, device_id: int | Device, shared_handle: int):
```
Andy-Jost: needs docstring

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -323,41 +330,243 @@ class MemoryResource(abc.ABC):
...
+        Specifies whether to create an IPC-enabled memory pool. When set to
+        True, the memory pool and its allocations can be shared with other
+        processes. (Default to False)
+
+    max_size : int, optional
+        Maximum pool size. When set to 0, defaults to a system-dependent value.
+        (Default to 0)
+    """
+    ipc_enabled: bool = False
+    max_size: int = 0
```
kkraus14:
```suggestion
    cdef bint ipc_enabled = False
    cdef int max_size = 0
```

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -323,41 +330,243 @@ class MemoryResource(abc.ABC):
...
-        err, self._handle = driver.cuDeviceGetMemPool(device_id)
-        raise_if_driver_error(err)
-        self._dev_id = device_id
+        If set to `None`, the memory resource uses the driver's current
+        stream-ordered memory pool for the specified `device_id`. If no memory
+        pool is set as current, the driver's default memory pool for the device
+        is used.
+
+        If not set to `None`, a new memory pool is created, which is owned by
+        the memory resource.
```
kkraus14, line 578: I understand that we need to create a new mempool in order to set options that can only be done at pool creation time, but this has some interoperability concerns. In the non-IPC case we get and modify the current device memory pool, which other libraries in the process can and may use. In this case, we create a new memory pool and don't set it as the current memory pool which means it's only usable via this `DeviceMemoryResource` instance (or by grabbing the private `_mempool_handle` attribute).

For prior art, RMM C++ has both `cuda_async_memory_resource` which always creates a new pool (https://github.com/rapidsai/rmm/blob/7cef2f5f30e962e9f3b27a3a3f2753a40277c093/cpp/include/rmm/mr/device/cuda_async_memory_resource.hpp#L132) and `cuda_async_view_memory_resource` which always needs to have a pool passed to it (https://github.com/rapidsai/rmm/blob/7cef2f5f30e962e9f3b27a3a3f2753a40277c093/cpp/include/rmm/mr/device/cuda_async_view_memory_resource.hpp#L53). This feels like less surprising behavior in my opinion, but it definitely adds complexity where instead of just memory resources, users now need to be aware of the underlying memory pools as well.

Ultimately, I think we're going to need to support being able to interact with both the default pool as well as create our own pools, where I think we may need a more explicit differentiation between them.

cc @pciolkosz

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -323,41 +330,243 @@ class MemoryResource(abc.ABC):
...
+            err, = driver.cuMemPoolDestroy(self._mempool_handle)
+            raise_if_driver_error(err)
+            self._dev_id = None
+            self._mempool_handle = None
+            self._ipc_handle_type = _NOIPC_HANDLE_TYPE
+            self._mempool_owned = False
+
+    def __bool__(self):
+        """Check if the device memory resource is valid."""
+        return self._mempool_handle is not None
```
kkraus14: Should this be defined as an abstract method on `MemoryResource` as well? I imagine we don't want to have some resources support this and others not?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -323,41 +330,243 @@ class MemoryResource(abc.ABC):
...
+        self = cls.__new__(cls)
+        self._dev_id = device_id
+        self._mempool_handle = None
+        self._ipc_handle_type = _IPC_HANDLE_TYPE
+        self._mempool_owned = True
+
+        err, self._mempool_handle = driver.cuMemPoolImportFromShareableHandle(shared_handle, _IPC_HANDLE_TYPE, 0)
+        raise_if_driver_error(err)
+
+        return self
```
kkraus14: Returning a new `DeviceMemoryResource` that you can't allocate from feels quite surprising from a user perspective. Maybe we should have some kind of `ImportedDeviceMemoryResource` or something of the like that makes it clearer?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -323,41 +330,243 @@ class MemoryResource(abc.ABC):
...
+            The shareable handle for the memory pool.
+        """
+        err, shared_handle = driver.cuMemPoolExportToShareableHandle(self._mempool_handle, _IPC_HANDLE_TYPE, 0)
+        raise_if_driver_error(err)
+        return shared_handle
+
+    def close_shareable_handle(self, shared_handle) -> None:
+        """Close a shareable handle for the memory pool."""
+        assert self._ipc_handle_type == driver.CUmemAllocationHandleType.CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR
+        os.close(shared_handle)
```
kkraus14: Would maybe be good to build a small class for wrapping the handle similar to `IPCBufferDescriptor`?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -400,21 +609,45 @@ class DeviceMemoryResource(MemoryResource):
...
         raise_if_driver_error(err)
 
+    @property
+    def device_id(self) -> int:
+        """The associated device ordinal."""
+        return self._dev_id
+
+    @property
+    def handle(self) -> cuda.bindings.driver.CUmemoryPool:
+        """The memory resource handle."""
```
kkraus14: I think we should make this docstring a bit clearer that it returns a handle to the underlying `CUmemoryPool`

cuda_core/cuda/core/experimental/_memory.pyx
pciolkosz: How it's done on the C++ side is `memory_resource` is always a memory pool view. You either construct it from a pool and it references it or you construct it from a device or no arguments for pinned/managed and then it gives you a default pool. We do not interact with the `current` pool and I don't think we want to.

Memory pool in CUDA does not go away right away on destruction, you can continue to free the allocations, you just can't allocate new memory. I think it works fine with the reference model for resources we have.

If we ever feel like we need that we could add new resource types that own the pool. On the C++ side I don't think we should have types that sometimes own things and sometimes don't. It makes it difficult to reason when resources are released, but maybe type differentiation it's less important in more dynamic language like Python.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: There appear to be some gotchas when combining Cython with the `dataclass` decorator (some details [here](https://cython.readthedocs.io/en/stable/src/userguide/extension_types.html?utm_source=chatgpt.com#dataclass-extension-types)).

The TL;DR is that annotating data members with `cdef` has limitations. Two issues stand out: 1) default values cannot be supplied on the same line as a `cdef` declaration, and 2) `cdef`-declared members are inaccessible from Python.

The best solution I could find is to change the type annotations, e.g., `int` -> `cython.int`. This does not have the above limitations, and I confirmed that the Cython-generated C classes use native types and not PyObjects.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: It turns out this is not straightforward. The handle cannot simply be serialized and deserialized on Linux because it contains a file descriptor. Instead, a socket pair along with SCM_RIGHTS (or something similar) is needed, but that assigns a new file descriptor in the receiving process.

If we want something as conceptually easy as `shareable_handle.share()`, then we would need to take ownership of the transmission process in addition to handling serialization.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: I agree that this is surprising. The fundamental problem is that a `Buffer` holds a reference to a `MemoryResource` and uses it for deallocation, so whatever object imports the buffer needs to support the `MemoryResource` API and, hence, has an `allocate` method. Perhaps it would be slightly less surprising to create a separate class for that.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: Something like the following could work:

```
  1 # Parent process.
  2 mr = DeviceMemoryResource(...)
  3 channel = IPCChannel(...) # Create a Queue (Windows) or socketpair (Linux)
  4 mr.export_mempool(channel)
  5 multiprocess.Process(target=child_main, args=(channel,))
  6
  7 # ---
  8
  9 # Child process.
 10 mr = IPCMemoryImporter.import_mempool(channel)
  ```

File: cuda_core/cuda/core/experimental/_memory.pyx (hunk 3 above)
Comment (leofang) [reply cuda_core/cuda/core/experimental/_memory.pyx]: For my own curiosity, I wonder if `cdef public` also works.

File: cuda_core/cuda/core/experimental/_memory.pyx (hunk 4 above)
Comment (leofang) [reply cuda_core/cuda/core/experimental/_memory.pyx:578]: For the record, this is what I suggested Andy when he started:
> As for “current”, it is done for a reason. For many years we’ve been battling with the fact that all projects having their own memory allocators, and this causes lots of friction because we can easily run into OOM even if the memory is not actually used, just held by a project’s caching allocator. Driver introduces the mempool as an attempt to offer the single, global source. The behavior of `cuDeviceGetMemPool` is: If anyone happens to set their pool to current (as far as the driver is concerned), we assume it is meant to be reused by all projects talking to the driver, so we reuse it. If no one has set it, then we use the driver’s default pool.
>
> We don’t want to create a new pool by default, because we never set it to current. Note that we allocate via `cuMemAllocFromPoolAsync` instead of `cuMemAllocAsync` which relies on having a pool set to current. We want to play nicely with existing projects. We only create a pool when IPC is requested, because it is disabled in the default pool.

So the intention was that if users don't pass `options` (which is default to `None`), or pass a default-constructed `DeviceMemoryResourceOptions` instance, we reuse the current pool. Only when a non-default `options` would lead to a new pool instance. 

We can consider adding a `DeviceMemoryResource.set_current()` method, similar to our `Device.set_current()`, to make this new pool visible to the driver and reusable by other projects doing `cudaMallocAsync`. However, I am not convinced that if IPC is turned on, this new pool should be made globally visible by default for performance reasons.

File: cuda_core/cuda/core/experimental/_memory.pyx [hunk 9]
```diff
@@ -323,41 +340,330 @@ class MemoryResource(abc.ABC):
...
+def _requires_ipc(func):
+    """Decorator to ensure IPC support is enabled before executing a method."""
+    @wraps(func)
+    def wrapper(self, *args, **kwargs):
+        if not self.is_ipc_enabled:
+            raise RuntimeError("Memory resource is not IPC-enabled")
+        return func(self, *args, **kwargs)
+    return wrapper
+
+def _def_mempool_attr_property(scope: dict, name: str, property_type: type, doc: str):
```
leofang: Let's turn this as `cdef`

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -4,18 +4,30 @@
 
 from __future__ import annotations
 
-from libc.stdint cimport uintptr_t
```
leofang, line 7: FYI, I think even cython-lint (#941) cannot address this. Let me explain a bit the convention I use when this project started, and we can consider develop a mini style guide later 🙂 

I use CuPy's convention where we group all `cimport` first, and then `import`. This way it's easier to tell what are Cython `cdef` types and what are Python `def` types. Within each group, we sort the imports based on [the order of stdlib -> 3rd-party -> this project](https://docs.openstack.org/hacking/latest/user/hacking.html#import-order-template).

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -323,41 +340,330 @@ class MemoryResource(abc.ABC):
...
         ...
 
+# IPC is currently only supported on Linux. On other platforms, the IPC handle
+# type is set equal to the no-IPC handle type.
+
+_NOIPC_HANDLE_TYPE = driver.CUmemAllocationHandleType.CU_MEM_HANDLE_TYPE_NONE
+_IPC_HANDLE_TYPE = driver.CUmemAllocationHandleType.CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR \
+    if platform.system() == "Linux" else _NOIPC_HANDLE_TYPE
+
+def _requires_ipc(func):
```
leofang: FYI, in the previous performance analysis we found that even after cythonization, function decorators are still slow (maybe because `functools` is pure Python). While attributes that rely on the `_requires_ipc` pre-condition may not be in the critical path, the code is simpler/cleaner if we just inline the `self.is_ipc_enabled` check. We've done this to other classes, e.g. `Device`.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -323,41 +340,330 @@ class MemoryResource(abc.ABC):
...
+    scope[name] = property(getter)
+
+class IPCBufferDescriptor:
+    """Serializable object describing a buffer that can be shared between processes."""
+    def __init__(self, reserved: bytes, size: int):
+        self.reserved = reserved
+        self._size = size
+
+    def __reduce__(self):
+        # This is subject to change if the CumemPoolPtrExportData struct/object changes.
```
leofang: nit: make it easier to grep
```suggestion
        # This is subject to change if the CUmemPoolPtrExportData struct/object changes.
```

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -323,41 +340,330 @@ class MemoryResource(abc.ABC):
...
+
+def _requires_ipc(func):
+    """Decorator to ensure IPC support is enabled before executing a method."""
+    @wraps(func)
+    def wrapper(self, *args, **kwargs):
+        if not self.is_ipc_enabled:
+            raise RuntimeError("Memory resource is not IPC-enabled")
+        return func(self, *args, **kwargs)
+    return wrapper
+
```
leofang: One thing that cython-lint will help is spacing. Ruff and other linters will catch this too: Let's make sure there are two empty lines between the blocks of function/class definitions. Right now it is a bit hard to tell if a function/class ends or not 😅

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -323,41 +340,330 @@ class MemoryResource(abc.ABC):
...
+    def __exit__(self, exc_type, exc_value, traceback):
+        """Exit the context, closing the handle."""
+        self.close()
+
+class IPCChannel(abc.ABC):
+    """Communication channel for sharing IPC-enabled memory pools."""
+    def __new__(cls):
+        if platform.system() == "Linux":
+            self = object.__new__(IPCChannelUnixSocket)
```
leofang: Q: Is this a common pattern (having an ABC instantiating itself with its concrete child class)? How about we expose `IPCChannelUnixSocket` to the end users instead, and keep `IPCChannel` internal?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -323,41 +340,330 @@ class MemoryResource(abc.ABC):
...
+        return instance
+
+class IPCAllocationHandle:
+    """Shareable handle to an IPC-enabled device memory pool."""
+    def __init__(self, handle: int):
+        self.handle = handle
+
+    def close(self):
+        """Close the handle."""
+        if self:
```
```suggestion
        if self.handle:
```

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -323,41 +340,330 @@ class MemoryResource(abc.ABC):
...
-        # Check current release threshold
-        err, current_threshold = driver.cuMemPoolGetAttribute(
-            self._handle, driver.CUmemPool_attribute.CU_MEMPOOL_ATTR_RELEASE_THRESHOLD
+        When using an existing (current or default) memory pool, the returned
+        device memory resource does not own the pool (`handle_is_owned` is
+        `False`), and closing the resource has no effect.
+    """
+    __slots__ = "_dev_id", "_mempool_handle", "_ipc_handle_type", "_mempool_owned"
+
+    def __init__(self, device_id: int | Device, options = None):
```
```suggestion
    def __init__(self, device_id: int | Device, options: Optional[DeviceMemoryResourceOptions] = None):
```

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -323,41 +340,330 @@ class MemoryResource(abc.ABC):
...
-        # By default, the release threshold is 0, which means memory is immediately released back
-        # to the OS when there are no active suballocations, causing performance issues.
-        # Check current release threshold
-        err, current_threshold = driver.cuMemPoolGetAttribute(
-            self._handle, driver.CUmemPool_attribute.CU_MEMPOOL_ATTR_RELEASE_THRESHOLD
+        When using an existing (current or default) memory pool, the returned
+        device memory resource does not own the pool (`handle_is_owned` is
+        `False`), and closing the resource has no effect.
+    """
+    __slots__ = "_dev_id", "_mempool_handle", "_ipc_handle_type", "_mempool_owned"
```
leofang: `__slots__` is a leftover back when we had this file written in pure Python. Now that we already cythonized it, let's just make them `cdef`'d and initialized in `__cinit__` (when possible).

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -323,41 +340,330 @@ class MemoryResource(abc.ABC):
...
+
+            err, self._mempool_handle = driver.cuMemPoolCreate(properties)
+            raise_if_driver_error(err)
+
+    def __del__(self):
+        self.close()
+
+    def close(self):
+        """Close the device memory resource and destroy the associated memory pool if owned."""
+        if self and self._mempool_owned:
```
leofang: Q: Not sure why `self` could be None/False?
```suggestion
        if self._mempool_owned:
```

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -400,21 +706,45 @@ class DeviceMemoryResource(MemoryResource):
...
+        """The associated device ordinal."""
+        return self._dev_id
+
+    @property
+    def handle(self) -> cuda.bindings.driver.CUmemoryPool:
+        """Handle to the underlying memory pool."""
+        return self._mempool_handle
+
+    @property
+    def handle_is_owned(self) -> bool:
```
leofang: For bool attributes we start with `is_...`
```suggestion
    def is_handle_owned(self) -> bool:
```

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -261,6 +273,11 @@ class MemoryResource(abc.ABC):
         """
         ...
 
+    @abc.abstractmethod
+    def __bool__(self):
+        """Check if the memory resource is valid."""
+        ...
```
leofang: Would it be possible to always check `self.handle` instead of implementing `__bool__`? I find this usage confusing.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: Also, can we make `is_ipc_enabled` an attribute of `MemoryResource` ABC?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -400,21 +706,45 @@ class DeviceMemoryResource(MemoryResource):
...
+
+    # Define additional properties corresponding to mempool attributes.
+    _def_mempool_attr_property(locals(), "reuse_follow_event_dependencies", bool, "Allow memory to be reused when there are event dependencies between streams.")
+    _def_mempool_attr_property(locals(), "reuse_allow_opportunistic", bool, "Allow reuse of completed frees without dependencies.")
+    _def_mempool_attr_property(locals(), "reuse_allow_internal_dependencies", bool, "Allow insertion of new stream dependencies for memory reuse.")
+    _def_mempool_attr_property(locals(), "release_threshold", int, "Amount of reserved memory to hold before OS release.")
+    _def_mempool_attr_property(locals(), "reserved_mem_current", int, "Current amount of backing memory allocated.")
+    _def_mempool_attr_property(locals(), "reserved_mem_high", int, "High watermark of backing memory allocated.")
+    _def_mempool_attr_property(locals(), "used_mem_current", int, "Current amount of memory in use.")
+    _def_mempool_attr_property(locals(), "used_mem_high", int, "High watermark of memory in use.")
```
leofang: Let's follow our convention: Group these under a `DeviceMemoryResourceAttributes` class, and access them like this
```python
threshold = mr.attributes.release_threshold
```
see e.g. `DeviceProperties` and `KernelAttributes` (yes, unfortunately we weren't being very consistent with regard to "properties" vs "attributes", we probably should fix it in the near future...)

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -508,3 +844,6 @@ class _SynchronousMemoryResource(MemoryResource):
     @property
     def device_id(self) -> int:
         return self._dev_id
+
+del _requires_ipc
```
leofang: ditto, if `_requires_ipc` is `cdef`'d and/or inline'd, this `del` is not needed

cuda_core/cuda/core/experimental/_memory.pyx
leofang: I understand the code better now. `IPCChannel` is meant to be an OS-agnostic abstraction over the underlying OS shared memory mechanism. I think it is a great idea, but making this an ABC is super confusing because this is a user-facing object that they instantiate! 

We should have two private classes implemented. One is `IPCChannelUnixSocket` (this PR) and another for Windows (future PR), assuming we have @ksimpson-work or @tpn blessing that the Windows nitty-gritty can indeed be abstracted out (I will check this later).

This class should not have any public method.  Both `receive_allocation_handle` and `send_allocation_handle` are not meant to be user-facing. Prefix them with an underscore.

leofang: Another nit on naming: How about `IPCchannel` (one uppercase C) instead of `IPCChannel` (two uppercase C)?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -323,41 +340,330 @@ class MemoryResource(abc.ABC):
...
+    attr_enum = getattr(driver.CUmemPool_attribute, f"CU_MEMPOOL_ATTR_{name.upper()}")
+    def getter(self) -> property_type:
+        err, value = driver.cuMemPoolGetAttribute(self.handle, attr_enum)
+        raise_if_driver_error(err)
+        return property_type(value)
+
+    getter.__doc__ = doc
+    scope[name] = property(getter)
+
+class IPCBufferDescriptor:
```
leofang: 1. This class is user-facing and should be documented in `cuda_core/docs/source/api_private.rst`.
2. Q: any reason this is not a cdef class, with attributes cdef'd too?
3. Let us raise `__init__` following our convention (users should not instantiate it)

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -323,41 +340,330 @@ class MemoryResource(abc.ABC):
...
+    @property
+    def size(self):
+        return self._size
+
+    @classmethod
+    def _reconstruct(cls, reserved, size):
+        instance = cls(reserved, size)
+        return instance
+
+class IPCAllocationHandle:
```
leofang: ditto, same 3 comments as with the `IPCBufferDescriptor` class.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -323,41 +340,330 @@ class MemoryResource(abc.ABC):
...
+            )
+        return self.handle
+
+    def __enter__(self):
+        """Enter the context, returning the handle object."""
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        """Exit the context, closing the handle."""
+        self.close()
```
leofang: These are quite confusing to me. I can't think of a reason we want to use this object as a ctx mgr. Let's remove them.

FYI, so far we have been avoiding adding ctx mgr support for any CUDA objects, for now. Mainly because they don't interoperate well across Python libraries/frameworks and I don't want bug reports flowing in while busy bringing up the project. We will evaluate them on a case-by-case basis in the near future.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -323,41 +340,330 @@ class MemoryResource(abc.ABC):
...
+
+    @classmethod
+    def from_shared_channel(cls, device_id: int | Device, channel: IPCChannel) -> DeviceMemoryResource:
+        """Create a device memory resource from a memory pool shared over an IPC channel."""
+        device_id = getattr(device_id, 'device_id', device_id)
+        alloc_handle = channel.receive_allocation_handle()
+        return cls.from_allocation_handle(device_id, alloc_handle)
+
+    @classmethod
+    def from_allocation_handle(cls, device_id: int | Device, alloc_handle: IPCAllocationHandle) -> DeviceMemoryResource:
```
leofang: Oh. I was wrong then, in one of my earlier comments. `from_allocation_handle` can be totally turned into a private method. Don't make it user facing. This is just a private helper. We just need `IPCChannel` and `from_shared_channel` to be user facing.

Likewise, `IPCAllocationHandle` can be totally hidden from users.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -323,41 +340,330 @@ class MemoryResource(abc.ABC):
...
+            self._mempool_handle = None
+            self._ipc_handle_type = _NOIPC_HANDLE_TYPE
+            self._mempool_owned = False
+
+    def __bool__(self):
+        """Check if the device memory resource is valid."""
+        return self._mempool_handle is not None
+
+    @classmethod
+    def from_shared_channel(cls, device_id: int | Device, channel: IPCChannel) -> DeviceMemoryResource:
```
leofang, line 662: This is fantastic 👍

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -323,41 +340,330 @@ class MemoryResource(abc.ABC):
...
+        self._ipc_handle_type = _IPC_HANDLE_TYPE
+        self._mempool_owned = True
+
+        err, self._mempool_handle = driver.cuMemPoolImportFromShareableHandle(int(alloc_handle), _IPC_HANDLE_TYPE, 0)
+        raise_if_driver_error(err)
+
+        return self
+
+    @_requires_ipc
+    def share_to_channel(self, channel : IPCChannel):
```
leofang, line 704: ditto, this is fantastic 👍

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -323,41 +340,330 @@ class MemoryResource(abc.ABC):
...
+        channel.send_allocation_handle(self.get_allocation_handle())
+
+    @_requires_ipc
+    def get_allocation_handle(self) -> IPCAllocationHandle:
```
leofang: ditto, this should be made as a private helper, not user-facing

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -323,41 +340,330 @@ class MemoryResource(abc.ABC):
...
+        Returns
+        -------
+            The shareable handle for the memory pool.
+        """
+        err, alloc_handle = driver.cuMemPoolExportToShareableHandle(self._mempool_handle, _IPC_HANDLE_TYPE, 0)
+        raise_if_driver_error(err)
+        return IPCAllocationHandle(alloc_handle)
+
+    @_requires_ipc
+    def export_buffer(self, buffer: Buffer) -> IPCBufferDescriptor:
```
leofang: This may be a bit counter-intuitive at the first glance, but it'd look much better if we move this `export_buffer` and the `import_buffer` methods to the `Buffer` class. Then it'd look very symmetric to what we add to `DeviceMemoryResource`, and we also get to turn `IPCBufferDescriptor` non-user-facing! 
```python
class Buffer:

    ...

    def share_to_channel(self, channel: IPCChannel):
        # this is your export_buffer
        # Buffer has access to the underlying mempool, so we can
        # check is_ipc_enabled and raise if it's not
        ...

    def from_shared_channel(cls, channel: IPCChannel) -> Buffer:
        # this is your import buffer
        ...
```
and you'd get my "fantastic 👍" thumbs up 😆 

(Catch: We need to figure out how to make `IPCChannel` carry extra information. In your current implementation it only carries the exported mempool handle. We'd need it to carry also the exported buffer handle. Preferably IPC event handle, too, if we believe `IPCChannel` can be one single entity bearing the responsibility of abstracting shmem import/export.)

leofang (changes requested):
Thanks a lot, @Andy-Jost! Sorry for my late reply. I've gone through the implementation and made some suggestions, in 3 categories:
1. design: I believe your `IPCChannel` abstraction is key to deliver feature parity while enabling very nice UX. I've made a few suggestions below to tidy it up and make the UI even cleaner. Lots of objects and methods can be turned to private/non-user-facing!
   - Sorry, my comments might seem inconsistent at the first glance. I reviewed `_memory.pyx` from top to bottom, so in the beginning I was not seeing the full picture. I had also forgotten the details of Keenan's PR, on which this one is based.
2. implementation: in a few places we can cythonize more, also the addition of `__bool__` is a bit nerve wrecking
3. semantics: I would prefer to save the discussion of 1. .set_current(), 2. view or owning to the very end. I think the current implementation is OK. No need to change until we are making another pass.

Tests look OK but I need to review them again later.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: Just raise in `.allocate()` if the underlying mr is not IPC-enabled, we have full information.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: Sounds good, I will change this back. Is there a reason not to sort the groups alphabetically? FWIW, I have a vim macro that I use to pipe lines like this to the `sort` program. That's how I ended up with the order here. Nice an easy and I never really think about it. But I agree that the cimport/import grouping is important.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: I noticed the `__init__` versus `_init` convention but its purpose was not clear to me. Can you explain what it is used for? From your comment, it sounds like maybe we use it when we do not expect users to create instances of a class. Anything else?

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: I agree. I meant to take out the context manager support.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: `bool(self)` means "is this object valid/not closed" and is an alias for `self._handle is not None`

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: I was thinking users that don't want to use `IPCChannel` for some reason could directly use the interface based on allocation handles. I don't feel very strongly about this though.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: Yes, and I saw your comment at the top of the file about grouping shared attributes of `Buffer` and `MemoryResource`. I was planning to get to that, too.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: `cdef public` does appear to take care of the second issue I mentioned. Thanks for pointing that out.

Andy-Jost: In that case, the declaration can be written like this:

`cdef public int max_size`
`max_size = 0`

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: see `AllocationHandle`

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: That would mean turning `MemoryResource` and `DeviceMemoryResource` into Cython classes as well. Also, if you mean turning this function into a `cdef`, that would only potentially reduce the one-time costs when the class is created and would not apply to the property invocation. I'll play around with this and see if I can find something a little better.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: Ok, that's great. I ran cython-lint and manually fixed everything I could (see next upload).

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: I've updated this, but I'm not sure I have the same intuition for the naming. I can see the issue with the repeated uppercase letters in `IPCChannel` vs. `IPCchannel`. To my taste making an exception for this case is not great. Then would we also write `IPCbufferDescriptor` or `IPCallocationHandle`? `IPCChannel` seems least-bad to my eye, but I'll change it to match the convention if you say so. LMK

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: I don't think we want to specify the type. `check_or_create_options` is pretty loose in terms of what it accepts, and this declaration matches other places in the code (e.g., see `_stream`).

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: I will work on this. It involves Cythonizing `MemoryResource`.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: I've updated this but I don't prefer the direction. To me, testing an object's validity by placing it in a Boolean context a good practice.

Take a look at what happened in `IPCAllocationHandle`. When I Cythonized that, the placeholder for an invalid handle changed from `None` to `-1`. Because I had removed the `__bool__` method, I had to update the code in several places rather than just one.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: see my other comment

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: Thanks

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: I don't think this is possible without removing the `abc.ABC` base from `MemoryResource`.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -714,6 +742,13 @@ class DeviceMemoryResource(MemoryResource):
         err, = driver.cuMemFreeAsync(ptr, stream.handle)
         raise_if_driver_error(err)
 
+    @property
+    def attributes(self) -> DeviceMemoryResourceAttributes:
+        if self._attributes is None:
+            ref = weakref.ref(self)
```
Andy-Jost, line 771: I think the `KernelAttributes` (and probably others) have a bug in that they capture internal state from the object they are nested in at construction time. To really be correct, we should capture a weak reference to the owning object like this.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: This turned into a long story. Now that the attributes are bundled into a dedicated object, I was able to try Cythonizing. I timed the property access using `timeit` with and without `cdef`. The bottom line is that Cythonization gives no observable improvement here. I think the benefit of using a helper to generate the properties, which greatly reduces the amount of repetitive code, outweighs the other considerations, namely performance, which does not materialize in this case.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: I removed the `_requires_ipc` decorator, but I'd like to set aside moving the property into `MemoryResource` for this PR.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: I think this will be tough. Can we set it aside for this PR? The issue is that sharing a mempool handle on *nix requires a socket pair because SCM_RIGHTS needs to be used. But on the other hand the allocation handles are just plain data that can be shared by any method. In particular, Python's existing facilities, such as `multiprocessing.Queue` work great on any platform. I would need to think more about whether it makes sense to "force" users into using `IPCChannel` to share allocations, which might cause us to reimplement or at least wrap a lot of things.

Andy-Jost:
This PR is ready for final review and merge. There were a lot of great ideas in the comments. I implemented many of them but not all. Below is a list of what I couldn't do. Things that were not done were skipped because of either ongoing discussion without a clear agreement or technical questions/complexity. I anticipate follow-up work.

1. I did not change the interaction with the current or default memory pool. This change does not change the default behavior, which is that a default-constructed `DeviceMemoryResource` refers to the current memory pool without ownership semantics.
2. I did not Cythonize `DeviceMemoryResource` because I could not find a way to do it. The base hierarchy through `MemoryResource` -> `abc.ABC` does not appear to be compatible with Cython.
3. I did not attempt to share memory allocations over `IPCChannel` due to technical complexity (see the related comment).

cuda_core/tests/test_ipc_mempool.py
```diff
@@ -0,0 +1,149 @@
...
+        queue.put(handle)
+
+        # Wait for the child process.
+        process.join(timeout=10)
+        assert process.exitcode == 0
+
+        # Verify that the buffer was modified.
+        protocol.verify_buffer(flipped=True)
+    finally:
+        if locals().get("buffer") is not None:
```
cpcloud: I know you're trying to get this merged, so this isn't blocking in any way.

Peeking into `locals` for this specific use case (cleaning up), strongly suggests whatever is being cleaned up should implement the context manager protocol to [support using the object in a `with` statement](https://docs.python.org/3/reference/compound_stmts.html#with) which is the closest Python comes to anything like C++'s RAII.

Then your code effectively becomes 

```python
with mr.allocate(NBYTES) as buffer:
    # everything else that requires `buffer`
# if we're here, `buffer.close()` is guaranteed to have been called
# otherwise an exception was raised
```

Andy-Jost: I totally understand the point. There was an earlier discussion on whether to implement context managers for the objects we control, but there seems to be no agreement yet (I even removed support in one place). I think this is a topic we will continue to discuss as the memory resource design matures.

cuda_core/tests/test_ipc_mempool.py
```diff
@@ -0,0 +1,149 @@
...
+POOL_SIZE = 2097152  # 2MB size
+NBYTES = 64
+
+
+@pytest.fixture(scope="function")
+def ipc_device():
+    """Obtains a device suitable for IPC-enabled mempool tests, or skips."""
+    if get_binding_version() < (12, 0):
+        pytest.skip("Test requires CUDA 12 or higher")
+
```
leofang: We now support CUDA 12+, so this can be removed
```suggestion
```

cuda_core/tests/test_memory.py
```diff
@@ -7,12 +7,30 @@
...
+
+POOL_SIZE = 2097152  # 2MB size
+
+
+@pytest.fixture(scope="function")
+def mempool_device():
+    """Obtains a device suitable for mempool tests, or skips."""
+    if get_binding_version() < (12, 0):
+        pytest.skip("Test requires CUDA 12 or higher")
+
```
leofang: ditto
```suggestion
```

cuda_core/tests/test_memory.py
```diff
@@ -270,16 +288,163 @@ def test_device_memory_resource_initialization():
         pytest.skip("memory pools not supported")
     device.set_current()
 
-    # This should succeed and configure the memory pool release threshold
-    mr = DeviceMemoryResource(device.device_id)
+    # This should succeed and configure the memory pool release threshold.
+    # The resource can be constructed from either a device or device ordinal.
+    for device_arg in [device, device.device_id]:
```
leofang: Can we parametrize the test instead of having an internal for loop?

cuda_core/tests/test_memory.py
```diff
@@ -270,16 +288,163 @@ def test_device_memory_resource_initialization():
...
+            if property_name in ["reserved_mem_high", "used_mem_high"]:
+                # High watermark should never be less than current
+                current_prop = property_name.replace("_high", "_current")
+                current_value = getattr(mr.attributes, current_prop)
+                assert value >= current_value, f"{property_name} should be >= {current_prop}"
+
+        finally:
+            # Ensure we allocate and deallocate a small buffer to flush any pending operations
+            flush_buffer = mr.allocate(64)
+            flush_buffer.close()
```
leofang: Q: Do we see any error if we don't do this?

One possible alternative is to sync the device. Any sync/alloc/free ops would trigger the mempool's recycling.

cuda_core/tests/test_memory.py
```diff
@@ -270,16 +288,163 @@ def test_device_memory_resource_initialization():
...
+        ("reserved_mem_current", int),
+        ("reserved_mem_high", int),
+        ("used_mem_current", int),
+        ("used_mem_high", int),
+    ],
+)
+def test_mempool_attributes(mempool_device, property_name, expected_type):
+    """Test all properties of the DeviceMemoryResource class."""
+    device = mempool_device
+    for ipc_enabled in [True, False]:
```
leofang: ditto, re: parametrization, we could create a cartesian product of  ipc_enabled x pool attributes

cuda_core/tests/test_ipc_mempool.py
```diff
@@ -0,0 +1,149 @@
...
+    if platform.system() == "Windows":
+        pytest.skip("IPC is not supported on Windows")
+
+    # Check if IPC is supported on this platform/device
+    device = Device()
+    device.set_current()
+    if not device.properties.memory_pools_supported:
```
leofang, line 29: technically we need to check `.handle_type_posix_file_descriptor_supported `

cuda_core/tests/test_ipc_mempool.py
```diff
@@ -0,0 +1,149 @@
...
+        self.nbytes = nbytes
+        self.stream = stream if stream is not None else device.create_stream()
+        self.scratch_buffer = DummyUnifiedMemoryResource(self.device).allocate(self.nbytes, stream=self.stream)
+
+    def fill_buffer(self, flipped=False):
+        """Fill a device buffer with test pattern using unified memory."""
+        ptr = ctypes.cast(int(self.scratch_buffer.handle), ctypes.POINTER(ctypes.c_byte))
+        op = (lambda i: 255 - i) if flipped else (lambda i: i)
+        for i in range(self.nbytes):
+            ptr[i] = ctypes.c_byte(op(i))
```
leofang, line 167: FYI (no need to change), our `Buffer` can be made zero-copy to a NumPy array via `arr = np.from_dlpack(self.scratch_buffer)`. Then the op can be easily written as
```python
arr[:] = np.arange(255, 255 - arr.size, -1, dtype=np.int8) if flipped else np.arange(arr.size, dtype=np.int8) 
```

cuda_core/tests/test_ipc_mempool.py
```diff
@@ -0,0 +1,149 @@
...
+        self.buffer.copy_from(self.scratch_buffer, stream=self.stream)
+
+    def verify_buffer(self, flipped=False):
+        """Verify the buffer contents."""
+        self.scratch_buffer.copy_from(self.buffer, stream=self.stream)
+        self.device.sync()
```
leofang:
```suggestion
        self.stream.sync()
```

cuda_core/tests/test_ipc_mempool.py
```diff
@@ -0,0 +1,149 @@
...
+        ptr = ctypes.cast(int(self.scratch_buffer.handle), ctypes.POINTER(ctypes.c_byte))
+        op = (lambda i: 255 - i) if flipped else (lambda i: i)
+        for i in range(self.nbytes):
+            assert ctypes.c_byte(ptr[i]).value == ctypes.c_byte(op(i)).value, (
+                f"Buffer contains incorrect data at index {i}"
+            )
```
leofang, line 179: ditto, we can use `all(arr1 == arr2).all()` to check

cuda_core/tests/test_ipc_mempool.py
```diff
@@ -0,0 +1,149 @@
...
+    try:
+        # Start the child process.
+        queue = multiprocessing.Queue()
+        process = multiprocessing.Process(target=child_main, args=(channel, queue))
+        process.start()
+
+        # Allocate and fill memory.
+        buffer = mr.allocate(NBYTES)
+        protocol = IPCBufferTestProtocol(ipc_device, buffer)
+        protocol.fill_buffer(flipped=False)
```
leofang: This operation is not stream-ordered. `allocate()` and `fill_buffer()` should run on the same stream.

cuda_core/tests/test_ipc_mempool.py
```diff
@@ -0,0 +1,149 @@
...
+
+        # Export the buffer via IPC.
+        handle = mr.export_buffer(buffer)
+        queue.put(handle)
```
leofang: I need to apologize for sharing my wrong understanding when syncing with you and Keenan. 

There is a reason I mentioned IPC event handle [here](https://github.com/NVIDIA/cuda-python/pull/930#discussion_r2326399207). It turns out that only the `cudaIpcMemHandle_t` is considered "unsafe"/"legacy"; `cudaIpcEventHandle_t` (and its driver counterpart) is not, and is still used in the "safe"/"modern" IPC example, [see here](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#creating-and-sharing-allocations-from-an-exported-pool). 

This is important because of the stream ordering reason as I explained during the sync-up. The mental model is the same: 
- process 1 does work on buffer on its stream
- process 1 creates an IPC event and record on its stream
- process 1 exports the buffer and the IPC event
- process 2 imports the buffer and the event
- process 2 has its stream wait on the event
- process 2 does work on buffer on its stream
- ...

So as it stands currently our test suite is not safe due to lack of stream ordering.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: Q: I don't find any comment addressing this? It is fine if we don't want to piggyback on `IPCChannel` (not sure why, though), but it still seems more natural to me to move these two methods to `Buffer` (names TBD):
- `DeviceMemoryResource.export_buffer` -> `Buffer.export` 
- `DeviceMemoryResource.import_buffer` -> `Buffer.import` or `Buffer.from_ipc` or ...

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -324,40 +331,375 @@ class MemoryResource(abc.ABC):
...
+    _defproperty(locals(), "reuse_follow_event_dependencies", bool, "Allow memory to be reused when there are event dependencies between streams.")
+    _defproperty(locals(), "reuse_allow_opportunistic", bool, "Allow reuse of completed frees without dependencies.")
+    _defproperty(locals(), "reuse_allow_internal_dependencies", bool, "Allow insertion of new stream dependencies for memory reuse.")
+    _defproperty(locals(), "release_threshold", int, "Amount of reserved memory to hold before OS release.")
+    _defproperty(locals(), "reserved_mem_current", int, "Current amount of backing memory allocated.")
+    _defproperty(locals(), "reserved_mem_high", int, "High watermark of backing memory allocated.")
+    _defproperty(locals(), "used_mem_current", int, "Current amount of memory in use.")
+    _defproperty(locals(), "used_mem_high", int, "High watermark of memory in use.")
+
+    del _defproperty
```
kkraus14: nitpick: I think it would be more readable and understandable to just implement these properties normally

cuda_core/tests/test_ipc_mempool.py
```diff
@@ -0,0 +1,149 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+try:
+    from cuda.bindings import driver
+except ImportError:
+    from cuda import cuda as driver
```
kkraus14, line 7: Do we need to still support this fallback for old versions of `cuda.bindings`?

cuda_core/tests/test_memory.py
kkraus14, line 1: Can we add a test that captures the expected failure behavior for when someone imports a `DeviceMemoryResource` via `DeviceMemoryResource.from_shared_channel` and tries to allocate using it? I'm not convinced that throwing on allocation is a good user experience here vs having something like an `ImportedDeviceMemoryResource` that would allow better introspection and IDE support if someone tried to write some code that tries to allocate from it.

cuda_core/tests/test_ipc_mempool.py
leofang: This is likely a copy-pasta from existing tests. We still need it because we want cuda.core to work with cuda-bindings < 12.6.2 (when we introduced the new layout) if possible.

cuda_core/cuda/core/experimental/_memory.pyx
cpcloud: Is there some past history of what things don't work together and why? Context managers are a **much** better alternative to, say, implementing `__del__` or dealing with weak references.

As for a reason to need them, take a look at the testing code in this PR around my comment [here](https://github.com/NVIDIA/cuda-python/pull/930/files#r2334564465).

Pretty much anything that can be "cleaned up" in some way (closed, deallocated, removed) etc. is a natural fit for a context manager.

Using `local`s is a really bad alternative. If a user asks us "How do I control the lifetime of this resource?" surely "inspect local variables and see which ones are defined and then close the one associated with the resource" is not an acceptable answer.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: I see. In this case let me quote from `import this`: "There should be one-- and preferably only one --obvious way to do it." If we find this use case later, we can move it to public. But if we later find this is a mistake, removing it would be a breaking change.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: `cdef` variables can be initialized in `__cinit__`.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: I like your reasoning. Thanks.

That reminds me we had the same struggle with the codegen aggressively changing camel to snake cases. We eventually figured out how to covert a name like `IPCChannel` to `ipc_channel`, but had to do this with a ton of crazy regex + Python logics...

cuda_core/cuda/core/experimental/_memory.pyx
leofang: interesting.. I'll have a look, thx

cuda_core/tests/test_ipc_mempool.py
leofang: This is not very nice because by the time of buffer deallocation, we don't always know which stream it'd be deallocated on, and the `with` syntax needs to make an assumption (=#779). If this is compounded with a `Stream` also being used as ctx mgr, it quickly becomes very messy. We want users to carefully ensure stream ordering if possible ("Explicit is better than implicit." -- Zen of Python).

leofang:
/ok to test [89d5e3f](https://github.com/NVIDIA/cuda-python/pull/930/commits/89d5e3fa66d6eaae314e59f37a99b653cdc9ff88)

cuda_core/tests/test_ipc_mempool.py
cpcloud: > This is not very nice because by the time of buffer deallocation, we don't always know which stream it'd be deallocated on

Then it sounds like there might be a mistake in the API design if users are forced to think about resource management of an implicit/default stream.

> the with syntax needs to make an assumption

I don't think so, unless I am misunderstanding something. If there's a lifetime relationship between streams and buffers, that should be reflected in the API, not left up to defaults or implicit finalizer behavior.

> it quickly becomes very messy

Can you write down an example of what you mean by messy? Poking around `locals` seems like a very low quality alternative to even the messiest of context managers.

We can move this discussion to #779 to avoid cluttering up the PR even more.

cuda_core/tests/test_memory.py
Andy-Jost: It passes without this. Keenan might be able to comment on the original purpose.

cuda_core/tests/test_ipc_mempool.py
Andy-Jost: Sure. I think the test for IPC is covered by the platform.system() == "Windows" check above, but I've updated it to use this device property.

cuda_core/tests/test_ipc_mempool.py
Andy-Jost: Good idea. I did run into some problems trying this, so I'll leave it out of this PR. The array created by `np.from_dlpack` is read-only and I didn't see an obvious way to make it writable (that's as deep as I went for now).

cuda_core/tests/test_ipc_mempool.py
cpcloud: To reiterate, my comments are not blocking!

cuda_core/tests/test_ipc_mempool.py
Andy-Jost: This leads to an error:

```
>>> np.from_dlpack(self.buffer)
RuntimeError: Unsupported device in DLTensor.
```

cuda_core/tests/test_ipc_mempool.py
Andy-Jost: Thanks for explaining this, Leo. I've addressed this in the tests by added the appropriate stream synchronization. That is good enough for testing the correctness. If we need a cleaner Python example using events, we can develop it outside this PR.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: Metaprogramming definitely increases the cognitive complexity. Whether that's worth the cost is a judgement call involving competing tradeoffs.

The way I see it, the benefits in this case are (1) saving around 50 lines of cut-paste code, (2) probably resulting in marginally faster code (due to caching the lookup of `attr_enum`), and (3) it being that adding a new attribute involves a one-line change.

To me, that tips the scales in favor of this approach in this case, but it's certainly not black and white.

cpcloud: It seems like you could reduce the complexity a lot by taking advantage of `property` directly:

```python
def getter(name: str, property_type):
    attr_enum = getattr(driver.CUmemPool_attribute, f"CU_MEMPOOL_ATTR_{name.upper()}")

    def impl(self) -> property_type:
        mr = self._mr()
        if mr is None:
            raise RuntimeError("DeviceMemoryResource is expired")
        err, value = driver.cuMemPoolGetAttribute(mr._mempool_handle, attr_enum)
        raise_if_driver_error(err)
        return property_type(value)

    return impl

used_mem_high = property(fget=getter("used_mem_high", int), doc="High watermark...")
```

We really should avoid `locals`.

Andy-Jost: @cpcloud I like your suggestion. This is another point in the balance between cognitive complexity and repetition. There’s no single best answer, and we shouldn’t dwell on it too much. I’m fine with making this change.

Edit: people can take a look at 80ee3c9970f20b87a188fa1279431b0f6d4dd504 if they want to see the before and after. My preference is still the original: it is more direct, less visually cluttered, and avoids a repetition in the property name.

cpcloud: There are so many ways to slice this. You can even use generic descriptors (i.e., `__get__`) or define the things programatically after the class definition.

We just need to really not peek into scope using `locals`.

Andy-Jost: > We just need to really not peek into scope using locals

Out of curiosity, why do say this? This instance seems like an extremely straightforward use case.

Python's class definition protocol:

1. Create a new dict, `d`.
2. Evaluate the class body in `d`.
3. Pass the result to the metaclass for postprocessing.

To do metaprogramming at step 2, one may need to get a reference to `d`, which is what `locals` does in that context. We could debate whether metaprogramming is a good idea, but that's another discussion. Assuming the aim is to do metaprogramming:

1. What's dangerous about using locals? What could break?
2. Is there a better alternative that allows one to generate class elements programmatically?

cuda_core/tests/test_memory.py
Andy-Jost: See f84d9d859b42013fe0732148651faed5c2f35d75 for this test.

I agree that throwing is not a great user experience. Wouldn't a `ImportedDeviceMemoryResource` still need to implement a throwing `allocate` function though?

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: See 6e86cd1243218f55eaf53a948ec01e51ffbb097d.

I agree these methods belong on the `Buffer` class and it was easy.

As for sharing the `IPCChannel`, I haven't thought about it enough. If you look at the tests, you can see that `IPCChannel` and `multiprocessing.Queue` are both used for communication. A straightforward implementation would be to roll a queue into `IPCChannel` but that seems clunky so I'd rather not do it unless it's really necessary (it's not).

I think we ought to discuss this further and, if we come up with an improvement, make a follow-on change.

Andy-Jost:
I think these changes are ready to go now.

> Thanks, Andy!
> 
> * We need to use IPC events for stream ordering, see my comments below in the test suite.

I did this by adding stream synchronization points. It is sufficient to test the functionality.

> * The question on import/export buffer methods is unanswered

Done

> 
> Also, could you add two entries (if I do not miss any) to the release note? `cuda_core/docs/source/release/0.X.Y-notes.rst`
> 
> * mention IPC is supported on Linux
> * mention MRs can also take `Device` instances

Done

cuda_core/cuda/core/experimental/_memory.pyx
cpcloud: It's not really a danger issue per se, it's primarily about using language/stdlib features that are extremely uncommon and have complex behavior.

I personally do not like `locals` for the following reasons (which I think might boil down to just the first point):

1. It's difficult to remember how it behaves in a given context. It's "whitespace sensitive" in some sense because what it returns depends on whatever is happening in the vicinity. It one thing in a function, another in a class definition, another at module scope.
2. It's mutable, and therefore requires that you understand where it was invoked to understand which scope now has things magically popping into it if you do mutate it. The behavior of `=` (for example) is IMO much easier to understand.

Andy-Jost: Thanks @cpcloud, those are good points that relate to `locals` increasing complexity and making the code more write-only.

Here's what we would get using another approach. This might represent the best trade-off as it likely satisfies @kkraus14's concern, avoids any repetition, and eliminates `locals`.

```
    def mempool_property(property_type: type):
        def decorator(stub):
            attr_enum = getattr(driver.CUmemPool_attribute, f"CU_MEMPOOL_ATTR_{stub.__name__.upper()}")

            def fget(self) -> property_type:
                mr = self._mr()
                if mr is None:
                  raise RuntimeError("DeviceMemoryResource is expired")
                err, value = driver.cuMemPoolGetAttribute(mr._mempool_handle, attr_enum)
                raise_if_driver_error(err)
                return property_type(value)
            return property(fget=fget, doc=stub.__doc__)
        return decorator

    @mempool_property(int)
    def used_mem_current(self):
        """Current amount of memory in use."""
```

Update: see e0e2175c034f67efbd7b80da9dfff150713a54de

cuda_core/cuda/core/experimental/_memory.pyx
leofang: Yes sounds good to me, thanks!

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -131,6 +131,25 @@ cdef class Buffer:
...
+    @classmethod
+    def import_(cls, mr: MemoryResource, ipc_buffer: IPCBufferDescriptor) -> Buffer:
+        """Import a buffer that was exported from another process."""
+        if not mr.is_ipc_enabled:
+            raise RuntimeError("Memory resource is not IPC-enabled")
+        share_data = driver.CUmemPoolPtrExportData()
+        share_data.reserved = ipc_buffer._reserved
+        err, ptr = driver.cuMemPoolImportPointer(mr._mempool_handle, share_data)
+        raise_if_driver_error(err)
+        return Buffer.from_handle(ptr, ipc_buffer.size, mr)
```
Andy-Jost, line 151: FYI - I'm planning to leave this as-is even though it is dangerous. If `from_handle` raises an exception, `ptr` will be leaked. I don't see any good options here, particularly because `mr.deallocate(ptr)` is stream-ordered and we do not have a stream in this context. Fortunately, `from_handle` is extremely unlikely to raise.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -299,20 +328,17 @@ class MemoryResource(abc.ABC):
         """
         ...
 
-    @property
-    @abc.abstractmethod
+    @abc.abstractproperty
```
leofang: Maybe this is another reason we need cython-lint sooner than later, not sure if it'd catch this. But `abstractproperty` is declared deprecated in favor of what's done prior to this PR. Let's not change it.

https://docs.python.org/3/library/abc.html#abc.abstractproperty

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -299,20 +328,17 @@ class MemoryResource(abc.ABC):
...
     def is_device_accessible(self) -> bool:
         """bool: True if buffers allocated by this resource can be accessed on the device."""
         ...
 
-    @property
-    @abc.abstractmethod
+    @abc.abstractproperty
```
leofang: ditto

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -299,20 +328,17 @@ class MemoryResource(abc.ABC):
...
     def is_host_accessible(self) -> bool:
         """bool: True if buffers allocated by this resource can be accessed on the host."""
         ...
 
-    @property
-    @abc.abstractmethod
+    @abc.abstractproperty
```
leofang: ditto

cuda_core/cuda/core/experimental/_memory.pyx
leofang: Hi guys, sorry I joined the party late. I agree using `locals` is nerve wrecking. 

I would like to share a bit of history and another note.
1. When we implemented the device and kernel attributes, Keenan took an adventurous step and used Cursor (or another LLM tool, I don't recall the detail) to spit out of a draft for later iterations. Turns out this was one of the best decisions that we made, and is an extremely powerful approach in 2025. We really, really don't need any metapgramming hacks to "loop over a known list" in any capacity anymore. Generate a draft code by asking LLM to read the CUDA API docs as the working basis.
2. Metaprogramming is particularly bad in a Cython module. If we know the attribute return types statically, and if this is a cdef class (which I think it is, after recent changes?), we need to make sure we write (or generate) clean code for Cython to compile into efficient C++ code. `locals` and other `def`-based constructs add to Python overhead.

leofang (changes requested):
Left a few quick notes before calling it a night, will resume tomorrow!

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: @leofang Thanks for your analysis. I appreciate the perspective, but I think there's an additional point to consider. While the time to write or generate code is important, code quality also matters. Cut-and-paste style code, even if generated quickly by an LLM, can create maintenance problems.

On the Cython point, I intentionally omitted `cdef` for this class. In my tests, Cythonizing this class actually resulted in slightly slower code. Generally, I believe performance optimizations should be driven by data and balanced against their impact on code quality. Here, attribute access times are already unlikely to be a bottleneck and Cythonizing doesn’t appear to improve them, though it does make the code more complex. Prioritizing simple code here seems more beneficial IMO.

cuda_core/tests/test_memory.py
kkraus14: @Andy-Jost yes, but at least the user would be able to introspect the class as well as type the `allocate` function as `NoReturn` (https://docs.python.org/3/library/typing.html#typing.NoReturn) which in theory should help to guide users much more nicely?

cuda_core/tests/test_ipc_mempool.py
leofang: There was a bug and we need NumPy 2.1.0+ to allow writing this test:
https://github.com/NVIDIA/cuda-python/blob/365bf079f54d10993c1e036e280c6217a20969cf/cuda_core/tests/test_event.py#L130

cuda_core/tests/test_ipc_mempool.py
leofang: Ah, for device buffers (`.is_device_accessible` is True), we wrap them as CuPy arrays not NumPy. Just change it to `cp.from_dlpack()`.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: > code quality also matters. Cut-and-paste style code, even if generated quickly by an LLM, can create maintenance problems

I think you might have misinterpreted my comment. I did not say that any copy-pasta would work out of box. It'd not fly, at least not now, and we did perform a few more manual passes before reaching the steady state. 

The point is LLM is powerful enough that it provides a good-enough **draft** code (highlighting because it's the word I used) for us to easily iterate over. It lowers significantly the mental burden of having to hand-write a massive amount of code. We can be more brave and dream bigger now.

> In my tests, Cythonizing this class actually resulted in slightly slower code.

It does not make sense to me if you just naively making this class a `cdef class` without rewriting the entire class (using LLM or other means), for reasons I described above (`def`-based constructs add to Python overhead). Naively adding `cdef` before `class` will unlikely help. It has never been what I had in mind in the context of cythonization.

cuda_core/tests/test_memory.py
```diff
@@ -259,27 +274,167 @@ def test_buffer_dunder_dlpack_device_failure():
...
+        ("reserved_mem_high", int),
+        ("used_mem_current", int),
+        ("used_mem_high", int),
+    ],
+)
+def test_mempool_attributes(ipc_enabled, mempool_device, property_name, expected_type):
+    """Test all properties of the DeviceMemoryResource class."""
+    device = mempool_device
+    if platform.system() == "Windows":
+        return  # IPC not implemented for Windows
```
leofang, line 381: nit:
```suggestion
        pytest.skip("IPC not implemented for Windows")
```
or move it to as a class decorator.

cuda_core/tests/test_ipc_mempool.py
```diff
@@ -0,0 +1,179 @@
...
+        pytest.skip("Device does not support IPC")
+
+    return device
+
+
+def test_ipc_mempool(ipc_device):
+    """Test IPC with memory pools."""
+    # Set up the IPC-enabled memory pool and share it.
+    stream = ipc_device.create_stream()
+    mr = DeviceMemoryResource(ipc_device, dict(max_size=POOL_SIZE, ipc_enabled=True))
```
leofang, line 44: would be better to get `DeviceMemoryResourceOptions` tested

cuda_core/tests/test_ipc_mempool.py
leofang: it might be a good idea to just wrap both buffers as cupy arrays instead of numpy arrays like I originally suggested, because one is backed by device memory another by managed memory, and numpy can't handle the former

cuda_core/tests/test_ipc_mempool.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def child_main1(channel, queue):
+    device = Device()
+    device.set_current()
+    stream = device.create_stream()
+
+    mr = DeviceMemoryResource.from_shared_channel(device, channel)
+    handle = queue.get()  # Get exported buffer data
+    buffer = Buffer.import_(mr, handle)
```
leofang, line 79: Thinking about this more, it's probably a better idea to avoid the whole discussion of `ImportedDeviceMemoryResource` by not making it user-visible:
```python
handle = queue.get()
buffer = Buffer.import_(device, handle, channel)
```
The MR would be a construct internal to the (imported) buffer this way.

If we manage to bypass the user-specified IPC mechanism and use `IPCChannel` to serve all needs, the code would be even more clean:
```python
buffer = Buffer.import_(device, channel)
```
But again the main benefit is to avoid unnecessary exposure of `ImportedDeviceMemoryResource`; reducing 3 lines of code to only 1 line is not a big deal.

cuda_core/tests/test_ipc_mempool.py
```diff
@@ -0,0 +1,179 @@
...
+
+    protocol = IPCBufferTestProtocol(device, buffer, stream=stream)
+    protocol.verify_buffer(flipped=False)
+    protocol.fill_buffer(flipped=True)
```
leofang, line 83: One thing I find the code a bit hard to follow is that it is unclear to me if the operations are stream-ordered. It'd be better to pass `stream` explicitly instead of holding it internally to `IPCBufferTestProtocol` instances.

cuda_core/tests/test_memory.py
```diff
@@ -259,27 +274,167 @@ def test_buffer_dunder_dlpack_device_failure():
...
+    assert buffer1.size == 1024
+    assert buffer2.size == 2048
+    buffer1.close()
+    buffer2.close()
+
+    # Test stream-based allocation
+    stream = device.create_stream()
+    buffer = mr.allocate(1024, stream=stream)
+    assert buffer.handle != 0
+    buffer.close()
```
leofang, line 334: nit
```suggestion
    buffer.close(stream)
```

cuda_core/tests/test_memory.py
```diff
@@ -259,27 +274,167 @@ def test_buffer_dunder_dlpack_device_failure():
...
+
+    # Test high watermark properties
+    if property_name in ["reserved_mem_high", "used_mem_high"]:
+        # High watermark should never be less than current
+        current_prop = property_name.replace("_high", "_current")
+        current_value = getattr(mr.attributes, current_prop)
+        assert value >= current_value, f"{property_name} should be >= {current_prop}"
+
+
+def test_mempool_attributes_ownership(mempool_device):
```
leofang, line 417: 👍

cuda_core/tests/test_memory.py
leofang: When we revisit this thread: We should keep the test, but I think we can avoid `ImportedDeviceMemoryResource`: https://github.com/NVIDIA/cuda-python/pull/930#discussion_r2354022667

leofang:
(Will merge tomorrow; see DM)

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: @leofang regarding Cythonization, I compared the code as it appears in this PR (using a custom decorator) against `cdef` code using a ordinary property decorators. In other words, I adjusted the class definition to make it friendly to Cython and I confirmed that the generated C code was fully Cythonized.

I believe the observed performance is explained by the non-Cythonized code caching the attribute look ups. To do the same think in Cython would require making a member for each property in the `cdef` variables section and adding code to initialize it. That would make the code even more complex, though, and I don't think it's worth it in this case.

PR #936, opened 2025-09-02
Fix #935: Remove bare except clauses

### LINKED ISSUE #935: [BUG]: Eliminate uses of bare exceptions
### Is this a duplicate?

### Type of Bug

Something else

### Component

General cuda-python

### Describe the bug

As @kkraus14 pointed out, catching bare exceptions and `passing` can swallow `KeyboardInterrupt` exceptions.  We have a bunch of uses of that throughout the codebase, and it would be worth updating them all to use a more specific exception.

### How to Reproduce

Theoretical problem, with an easy fix.

### Expected behavior

`KeyboardInterrupt` exceptions should not get silently ignored.

There is a [Ruff lint](https://docs.astral.sh/ruff/rules/bare-except/) about this that we may be able to turn on.

### Operating System

_No response_

### nvidia-smi output

_No response_

mdboom:
It would have been nice to add a linting rule to
catch this, but:

- ruff doesn't scan Cython
- cython-lint doesn't have a rule for this

## Description

closes 

## Checklist


mdboom:
Doing #938 instead.

PR #940, opened 2025-09-03
perf: remove generator expressions from input validation loops

cpcloud:
Add freshly generated code from updated `cython-gen` branch at https://github.com/NVIDIA/cuda-python-private/pull/138.

### DIFF
--- a/cuda_bindings/cuda/bindings/driver.pyx.in
+++ b/cuda_bindings/cuda/bindings/driver.pyx.in
@@ -363,7 +363,7 @@ class CUipcMem_flags(IntEnum):
 
 cdef dict _dict_CUipcMem_flags = {
     {{if 'CU_IPC_MEM_LAZY_ENABLE_PEER_ACCESS' in found_values}}
-    cydriver.CUipcMem_flags_enum.CU_IPC_MEM_LAZY_ENABLE_PEER_ACCESS: CUipcMem_flags.CU_IPC_MEM_LAZY_ENABLE_PEER_ACCESS
+    cydriver.CUipcMem_flags_enum.CU_IPC_MEM_LAZY_ENABLE_PEER_ACCESS: CUipcMem_flags.CU_IPC_MEM_LAZY_ENABLE_PEER_ACCESS,
     {{endif}}
 }
 {{endif}}
@@ -389,13 +389,13 @@ class CUmemAttach_flags(IntEnum):
 
 cdef dict _dict_CUmemAttach_flags = {
     {{if 'CU_MEM_ATTACH_GLOBAL' in found_values}}
-    cydriver.CUmemAttach_flags_enum.CU_MEM_ATTACH_GLOBAL: CUmemAttach_flags.CU_MEM_ATTACH_GLOBAL
-    {{endif}},
+    cydriver.CUmemAttach_flags_enum.CU_MEM_ATTACH_GLOBAL: CUmemAttach_flags.CU_MEM_ATTACH_GLOBAL,
+    {{endif}}
     {{if 'CU_MEM_ATTACH_HOST' in found_values}}
-    cydriver.CUmemAttach_flags_enum.CU_MEM_ATTACH_HOST: CUmemAttach_flags.CU_MEM_ATTACH_HOST
-    {{endif}},
+    cydriver.CUmemAttach_flags_enum.CU_MEM_ATTACH_HOST: CUmemAttach_flags.CU_MEM_ATTACH_HOST,
+    {{endif}}
     {{if 'CU_MEM_ATTACH_SINGLE' in found_values}}
-    cydriver.CUmemAttach_flags_enum.CU_MEM_ATTACH_SINGLE: CUmemAttach_flags.CU_MEM_ATTACH_SINGLE
+    cydriver.CUmemAttach_flags_enum.CU_MEM_ATTACH_SINGLE: CUmemAttach_flags.CU_MEM_ATTACH_SINGLE,
     {{endif}}
 }
 {{endif}}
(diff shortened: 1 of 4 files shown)


leofang:
Q: Do we have a rough data point on how much perf improvement we are seeing with this PR? Can we release-note it? 🙂

leofang:
/ok to test [9d8a535](https://github.com/NVIDIA/cuda-python/pull/940/commits/9d8a535bb7a7b7f9ab9143f8ab8fcd82a7f3e9ae)

cpcloud:
Looks like the segfault is only with Python 3.9 and the rest of the failures are due to a missing attribute. Investigating...

cpcloud:
Ok, it looks like we have to incref in a few places where we call `.pyObj()` and put that in another `list` (and the existing list needs to remain valid).

leofang:
Thanks, Phillip! Code-wise it LGTM. 

Could you check out my earlier question https://github.com/NVIDIA/cuda-python/pull/940#issuecomment-3251048121? If a benchmarking strategy is needed, I am thinking `nvrtcCompileProgram` is a good target. It is called by `cuda.core.experimental.Program`. Use it to compile a dummy CUDA C++ kernel (ex: `__global__ void empty() {}`) with lots of compiler options, and see how it goes before/after this change.

cpcloud:
@leofang Sure thing! Is there a tutorial I can follow (ideally showing the use of `nvrtcCompileProgram`) to get started on that benchmark?

cpcloud:
Also there appear to be some checks necessary to verify that a few of the enum flags exist, at compile time.

leofang:
> Is there a tutorial I can follow (ideally showing the use of `nvrtcCompileProgram`) to get started on that benchmark?

Yes! We have many code samples for each project. For example this one for cuda.core:
https://github.com/NVIDIA/cuda-python/blob/main/cuda_core/examples/vector_add.py
`nvrtcCompileProgram ` is called in when we call `Program.compile()`:
https://github.com/NVIDIA/cuda-python/blob/1b78bb3196a5f76750ac8882e35b75a4911b52c1/cuda_core/cuda/core/experimental/_program.py#L490

cpcloud:
@leofang

> with lots of compiler options

What does that mean 😅 ?

cpcloud:
Ok, looks like I finally got CI passing.

leofang:
I meant `nvrtcCompileProgram` takes a sequence of options (which is what we're accelerating by replacing generator with for loop), and this sequence can be passed at `Program` construction time by `Program(..., options=options)`, with `ProgramOptions`.
https://nvidia.github.io/cuda-python/cuda-core/latest/generated/cuda.core.experimental.ProgramOptions.html#cuda.core.experimental.ProgramOptions
We can make the sequence long (so that the effect is more apparent) by setting innocent compiler options, such as `ftz`, `prec_sqrt`, `debug`, ...

cpcloud:
I see. I misunderstood exactly what we're benchmarking, but I see now that having that long sequence of compiler options will hit the newly generated code path(s).

cpcloud:
I don't think this PR has much effect on anything (good or bad):

image

Here's the code I used for the benchmark:

```python
from cuda.core.experimental import Device, Program, ProgramOptions


def test_compile(benchmark):
    code = "__global__ void empty() {}"

    dev = Device()
    dev.set_current()

    # prepare program
    program_options = ProgramOptions(
        std="c++17",
        arch=f"sm_{dev.arch}",
        relocatable_device_code=True,
        debug=False,
        lineinfo=True,
        device_code_optimize=True,
        # ptxas_options: Union[str, list[str], tuple[str]] | None = None
        # max_register_count: int | None = None
        ftz=True,
        prec_sqrt=True,
        prec_div=True,
        fma=True,
        use_fast_math=True,
        extra_device_vectorization=True,
        link_time_optimization=True,
        gen_opt_lto=True,
        # define_macro: (
        #     Union[str, tuple[str, str], list[Union[str, tuple[str, str]]], tuple[Union[str, tuple[str, str]]]] | None
        # ) = None
        # undefine_macro: Union[str, list[str], tuple[str]] | None = None
        # include_path: Union[str, list[str], tuple[str]] | None = None
        # pre_include: Union[str, list[str], tuple[str]] | None = None
        no_source_include=False,
        builtin_move_forward=True,
        builtin_initializer_list=True,
        disable_warnings=False,
        # restrict: bool | None = None
        device_as_default_execution_space=True,
        device_int128=True,
        # optimization_info: str | None = None
        no_display_error_number=False,
        # diag_error: Union[int, list[int], tuple[int]] | None = None
        # diag_suppress: Union[int, list[int], tuple[int]] | None = None
        # diag_warn: Union[int, list[int], tuple[int]] | None = None
        brief_diagnostics=False,
        # time: str | None = None
        split_compile=2,
        fdevice_syntax_only=False,
        minimal=False,
    )

    prog = Program(code, code_type="c++", options=program_options)
    benchmark(prog.compile, "cubin")
```

Running it:

```
# first checkout main and build the things
pytest --benchmark-only --benchmark-enable --benchmark-autosave
```

```
# now checkout this branch `remove-generators` (re)-build all the things
pytest --benchmark-only --benchmark-enable --benchmark-compare 0001  # <- this is previous run's data
```

leofang:
Thanks, @cpcloud!

@mdboom any thoughts? Maybe Cython's handing on generators isn't as bad as it seems to be?

leofang:
@cpcloud is this read correct?
- NOW -> this PR
- 0001_d64120b -> `main` @ d64120b

It seems 0001_d64120b is actually faster? mean, max, stddev are all smaller than those of NOW. Does this make sense?

mdboom:
I read these numbers as "change is within the noise".  I think we may only see an improvement when calling functions where the argument-parsing overhead dominates the actual work done (which is probably not the case for compile).  We could disable the actual work in that function to get a read on the time of argument parsing overhead (since, IIUC, it's the Python-specific overheads that are pushing people toward other solutions).  Are there use cases that involve many calls to CUDA functions in a Python loop that we could use as a benchmark?  

I apologize if I sent @cpcloud on a wild goose chase -- we had such good results with this on #838, I thought there might be low-hanging fruit here, but hopefully at least it was a good thing to get familiarized and setup...

Additionally, I would expect the `_dict_` initializers to only have effect at import time.  We should measure whether that is having any effect.  I've been using this (`pyperf` has good support for measuring import time in a clean interpreter):

```
import sys
import pyperf

runner = pyperf.Runner()
runner.bench_command('cuda.bindings.driver import', [sys.executable, '-c', 'import cuda.bindings.driver'])
```

cpcloud:
I would be happy to abandon this PR. It doesn't seem like it's worth it at this point!

cpcloud:
(I deleted my previous comment because I hadn't seen the new comments)

cpcloud:
> It seems 0001_d64120b is actually faster? mean, max, stddev are all smaller than those of NOW. Does this make sense?

Yes, if there's comparable overhead between the two approaches. I don't think there's anything to conclude other than what @mdboom said about being within noise.

> We could disable the actual work in that function to get a read on the time of argument parsing overhead

True, but I'm not sure that would tell us much, since the overhead apparently doesn't show up in the simplest of kernels, and I'm not sure there's a use case where users are not ultimately compiling a kernel.

> I apologize if I sent @cpcloud on a wild goose chase

No worries, was a useful exercise to get set up and familiarize myself with the various components.

> Additionally, I would expect the _dict_ initializers to only have effect at import time. We should measure whether that is having any effect.

I can take a look at that a bit later today.

leofang:
Another target that allows a stress test with a sequence of arbitrary length is `nvrtcCreateProgram`. The `includeNames` argument is a sequence of header names that will be included by the source code. (The doc is a bit scarce, so the best resource is this section https://docs.nvidia.com/cuda/nvrtc/index.html#basic-usage.) 

As an (incomplete) sketch:
```python
from time import perf_counter
from cuda.bindings import nvrtc


N = 100  # sequence length
N_RUNS = 10000  # number of runs
N_WARMUPS = 10  # number of warmup runs

kernel_code = r"""
// we don't even need to autogenerate dummy headers on disk,
// because the way we test it below does not actually trigger JIT compilation
#include <dummy0.h>
#include <dummy1.h>
#include <dummy2.h>
...
#include <dummy99.h>

extern "C"
__global__ void my_dummy_kernel() {
  // do nothing
}
"""

kernel_code_in_bytes = kernel_code.encode()
program_name = b"dummy_program_name"
include_names = ["dummy0.h", "dummy1.h", ..., "dummy99.h"]

# use either your favorite tool for benchmarking, or this simple loop
t0 = t1 = t_sum = 0.
for i in range(N_RUNS + N_WARMUPS):
    if i > N_WARMUPS:
        t0 = perf_counter()
    err, program_handle = nvrtc.nvrtcCreateProgram(
        kernel_code_in_bytes, 
        program_name,
        N,
        [],
        include_names)
    if i > N_WARMUPS:
        t1 = perf_counter()
        t_sum += t1 - t0
    assert err == 0
    err, = nvrtc.nvrtcDestroyProgram(program_handle)
    assert err == 0

print("ave time:", t_sum / N_RUNS)

**Comment by cpcloud** [comment]
I did consider using `nvrtc.nvrtcCreateProgram`, but I started with a higher level API to see if a very common code path encountered by users was improved, as opposed to improving the performance of something that most people will never use directly.

**Comment by cpcloud** [comment]
I pared down the benchmark and made it use `nvrtcCreateProgram` and there's even less of difference:

image

Here's the code:

```python
from cuda.bindings import nvrtc


N = 100  # sequence length

include_names = [f"dummy{i:d}.h" for i in range(N)]
headers = "\n".join(f"#include <{header}>" for header in include_names)

kernel_code = rf"""\
{headers}

extern "C"
__global__ void my_dummy_kernel() {{}}"""

program_name = b"dummy_program_name"


def test_program(benchmark):
    def func(*, kernel_code, program_name, n, header_sources, include_names):
        err, program_handle = nvrtc.nvrtcCreateProgram(
            kernel_code, program_name, N, header_sources, include_names
        )
        assert err == 0
        (err,) = nvrtc.nvrtcDestroyProgram(program_handle)
        assert err == 0

    benchmark(
        func,
        kernel_code=kernel_code.encode(),
        program_name=program_name,
        n=N,
        header_sources=[b""] * N,
        include_names=list(map(str.encode, include_names)),
    )
```

cpcloud:
In light of that, I'd like to close this PR out. There's no evidence that generator expressions are causing any visible performance problems.

leofang:
@cpcloud With the new benchmark, it seems `NOW` (this PR) is faster than the main branch https://github.com/NVIDIA/cuda-python/commit/f317f21 (`mean` is 19.1157 vs 20.0167)? Could you remind me what's the time unit?

cpcloud:
It's a bit hard to see, but it's microseconds (`us` is how it's abbreviated).

leofang:
Ah ok, I was blind lol... Thanks. If we increase the sequence length by an order of magnitude, would it amplify the difference?

cpcloud:
Are users really including thousands of headers in one invocation of `nvrtcCreateProgram`?

leofang:
This is a proxy for us to investigate the perf difference. I was actually seeing hope here and was trying to help justify the changes made in this PR, because [your description](https://github.com/NVIDIA/cuda-python/pull/940#issuecomment-3293264421) did not match [the result you gave](https://github.com/NVIDIA/cuda-python/pull/940#issuecomment-3293261004) (which did hint that the change gives a noticeable ~5% speedup).

cpcloud:
My thought process was that I don't consider a 5% speedup at the microsecond scale, for this library's use cases to be evidence of a performance improvement.

There might be a literal difference, but 5% doesn't seem worth it without some information about how this affects an end user's experience.

At a larger scale, the effect is 16% worse for the worst case and 1% worse for the best case:

image

This strongly suggests to me that the difference isn't significant. It's sometimes better with this code and sometimes worse.

leofang:
Let us discuss this offline. I am lost.

leofang:
@cpcloud walked me through the results. One was [with N=100](https://github.com/NVIDIA/cuda-python/pull/940#issuecomment-3293261004), and the other was [with N=1000](https://github.com/NVIDIA/cuda-python/pull/940#issuecomment-3293484494), and the trend reverted. So it seems to us it's mainly affected by noise. We can revisit this in the future if needed.

PR #942, opened 2025-09-04
Fix #938: Call win32 APIs directly

### LINKED ISSUE #938: [ENH]: Remove the use of win32api
As @rwgk [pointed out](https://github.com/NVIDIA/cuda-python-private/pull/140#discussion_r2318017143), we don't need the whole `win32api` package to make the handful of Windows API calls we need to make.  Instead, we could just define `extern` functions in Cython directly.  This would be more efficient (no need to go through Python) and remove a ~9MB dependency on pywin32.

mdboom:
## Description

Instead of using `pywin32`, just calls the win32 APIs directly using Cython `extern`.

closes #938

This has a measurable impact on import time of about 9% (`import cuda.bindings.driver` in a fresh interpreter), mainly by not spending time importing `win32api`:

- Before: 92.1 ms
- After: 84.6 ms

It also improves "time to first call" by about 10% (since the first call resolves all of the dynamic function pointers and makes many win32 API calls):

- Before: 103.2 ms
- After: 93.4 ms

## Checklist


mdboom:
> It'd be ideal to also reduce the code duplication (`get_cuda_version()`, `cdef extern from "windows.h":`) in a set of follow-on PRs. I believe it'll be pretty straightforward after this PR and the associated codegen PRs are merged.

The files that are generated by cybind are including the externs in a template, so they at least aren't copy-pasted.  That is, all of the templates for each library have a `${snippet_windows_externs}` that inserts all of this boilerplate (including the `get_cuda_version` function).  

It might have been nicer to `cimport` `cuda.bindings._lib.windll` instead, but that would add a build-time dependency on cuda_bindings that not all of the cybind-generated packages have (though I might be misunderstanding that part).  So I thought it better to make them "standalone" by having cybind put this snippet in all of them.  See the cybind PR I also filed for more details.

rwgk:
Wow

[CI / Test linux-64 / py3.10, 13.0.0, wheels, GPU l4 (push)](https://github.com/NVIDIA/cuda-python/actions/runs/17473570608/job/49628613096?pr=942) Failing after 4m

```
>       assert delay_ms - generous_tolerance <= elapsed_time_ms < delay_ms + generous_tolerance
E       assert 520.4592895507812 < (500.0 + 20)
```

I haven't seen that flake for a while.

Just rerun and ignore. The timing being off by a small margin certainly isn't due to a problem in cuda-bindings.

rwgk (dismissed):
Tests passed, except for that one flake. I think a rerun of that test will resolve the flake.

leofang:
Let's ~~merge~~ **not merge** until the internal MR is approved, since it involves other teams as well.

cuda_bindings/cuda/bindings/_lib/windll.pxd
```diff
@@ -0,0 +1,36 @@
...
+
+    cdef DWORD LOAD_LIBRARY_SEARCH_SYSTEM32 = 0x00000800
+
+    HMODULE _LoadLibraryExW "LoadLibraryExW"(
+        LPCWSTR lpLibFileName,
+        HANDLE hFile,
+        DWORD dwFlags
+    ) nogil
+
+    FARPROC _GetProcAddress "GetProcAddress"(HMODULE hModule, const char* lpProcName) nogil
```
leofang: As mentioned in the other MR, I think this should be
```suggestion
    FARPROC _GetProcAddress "GetProcAddress"(HMODULE hModule, LPCWSTR lpProcName) nogil
```

cuda_bindings/cuda/bindings/_lib/windll.pxd
```diff
@@ -0,0 +1,36 @@
+# SPDX-FileCopyrightText: Copyright (c) 2021-2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
+
+from libc.stddef cimport wchar_t
+from libc.stdint cimport uintptr_t
+from cpython cimport PyUnicode_AsWideCharString
+
+cdef extern from "windows.h":
```
leofang: nit: move nogil here

cuda_bindings/cuda/bindings/_lib/windll.pxd
```diff
@@ -0,0 +1,36 @@
...
+    cdef wchar_t* wpath
+    with gil:
+        wpath = PyUnicode_AsWideCharString(path, NULL)
+    return <uintptr_t>_LoadLibraryExW(
+        wpath,
+        hFile,
+        dwFlags
+    )
+
+cdef inline FARPROC GetProcAddress(uintptr_t hModule, const char* lpProcName) nogil:
```
leofang, line 38: ditto
```suggestion
cdef inline FARPROC GetProcAddress(uintptr_t hModule, LPCWSTR lpProcName) nogil:
```

cuda_bindings/cuda/bindings/_lib/windll.pxd
```diff
@@ -0,0 +1,36 @@
...
+        HANDLE hFile,
+        DWORD dwFlags
+    ) nogil
+
+    FARPROC _GetProcAddress "GetProcAddress"(HMODULE hModule, const char* lpProcName) nogil
+
+cdef inline uintptr_t LoadLibraryExW(str path, HANDLE hFile, DWORD dwFlags) nogil:
+    cdef wchar_t* wpath
+    with gil:
+        wpath = PyUnicode_AsWideCharString(path, NULL)
```
leofang: Sorry I did not read the docs for `PyUnicode_AsWideCharString` yesterday, which says
> Returns a buffer allocated by [PyMem_New](https://docs.python.org/3/c-api/memory.html#c.PyMem_New) (use [PyMem_Free()](https://docs.python.org/3/c-api/memory.html#c.PyMem_Free) to free it) on success.

so wouldn't this be leaking memory if we return without free? Also, we should check first if `wpath` is NULL before using it.

leofang (changes requested):

mdboom:
> FYI, I noticed the cuFile module is not changed?

Ah, I think that's just an oversight on my part.  I will regenerate that as well.

cuda_bindings/cuda/bindings/_lib/windll.pxd
leofang: Ah... I misread `LPCSTR` as `LPCWSTR`, so this was in fact correct. Thanks for not following my wrong comment 😅

cuda_bindings/cuda/bindings/_lib/windll.pxd
```diff
@@ -0,0 +1,41 @@
...
+        DWORD dwFlags
+    )
+
+    FARPROC _GetProcAddress "GetProcAddress"(HMODULE hModule, LPCSTR lpProcName)
+
+cdef inline uintptr_t LoadLibraryExW(str path, HANDLE hFile, DWORD dwFlags):
+    cdef uintptr_t result
+    cdef wchar_t* wpath = PyUnicode_AsWideCharString(path, NULL)
+    if wpath is NULL:
+        raise
```
leofang: Should we raise a `RuntimeError` or something else here?

mdboom: I had to look this up...

This is declared as [except NULL](https://github.com/cython/cython/blob/master/Cython/Includes/cpython/unicode.pxd#L210) so Cython handles the exception itself.  It generates code like this:

```
__pyx_t_1 = PyUnicode_AsWideCharString(__pyx_v_path, NULL); 
if (unlikely(__pyx_t_1 == ((wchar_t *)NULL))) __PYX_ERR(2, 28, __pyx_L1_error)
```

So we actually don't need the `NULL` check (in our code) at all.

leofang: Ah, I see. Either `cimport` from Cython or `cdef extern from "Python.h"` manually would have the exception value declared (otherwise there'd be a compile-time warning), you're right. Thanks for looking into the generated code, Mike!

tpn (dismissed):

cuda_bindings/cuda/bindings/_internal/cufile_linux.pyx
```diff
@@ -97,24 +116,9 @@ cdef int _check_or_init_cufile() except -1 nogil:
...
-        if __cuDriverGetVersion == NULL:
-            __cuDriverGetVersion = dlsym(handle, "cuDriverGetVersion")
-        if __cuDriverGetVersion == NULL:
-            raise RuntimeError('something went wrong')
-        err = (<int (*)(int*) noexcept nogil>__cuDriverGetVersion)(&driver_ver)
-        if err != 0:
-            raise RuntimeError('something went wrong')
-        #dlclose(handle)
-        handle = NULL
+        driver_ver = get_cuda_version()
```
kkraus14, line 119: `driver_ver` is unused in the `load_library` call which is the only place that it's passed as far as I can tell. Should we just remove it?

cuda_bindings/cuda/bindings/_internal/nvjitlink_windows.pyx
```diff
@@ -46,110 +98,54 @@ cdef int _check_or_init_nvjitlink() except -1 nogil:
...
-            raise NotSupportedError(f'CUDA driver is not found ({e})')
-        global __cuDriverGetVersion
-        if __cuDriverGetVersion == NULL:
-            __cuDriverGetVersion = <void*><intptr_t>win32api.GetProcAddress(handle, 'cuDriverGetVersion')
-            if __cuDriverGetVersion == NULL:
-                raise RuntimeError('something went wrong')
-        err = (<int (*)(int*) noexcept nogil>__cuDriverGetVersion)(&driver_ver)
-        if err != 0:
-            raise RuntimeError('something went wrong')
+        driver_ver = get_cuda_version()
```
kkraus14, line 102: This seems unused?

cuda_bindings/cuda/bindings/_internal/nvjitlink_windows.pyx
```diff
@@ -11,19 +11,71 @@ from .utils import FunctionNotFoundError, NotSupportedError
...
 from cuda.pathfinder import load_nvidia_dynamic_lib
 
-import win32api
+from libc.stddef cimport wchar_t
+from libc.stdint cimport uintptr_t
+from cpython cimport PyUnicode_AsWideCharString, PyMem_Free
+
+from .utils import NotSupportedError
+
+cdef extern from "windows.h" nogil:
```
kkraus14, line 20: Can we use `cuda.bindings._lib.windll` instead of redefining these?

leofang: This is from a different codegen. Mike has made it such that this extern block is autogenerated.

mdboom: The codegen in cybind has to work for modules that ship with cuda_bindings and those that don't, which is why I arrived at this.  I think we can iterate on making that a bit better over time...

cuda_bindings/cuda/bindings/_internal/cufile_linux.pyx
mdboom: Fine with me to remove.  I was just trying not to change existing behavior.  I thought (perhaps) there was a reason we always "check if CUDA is installed" before loading the specific library.  @leofang: Do you know if it matters?

kkraus14 (dismissed):

mdboom:
@kkraus14: I merged main into here for one final test, but otherwise no changes since your last "approved" review.

cuda_bindings/cuda/bindings/_internal/cufile_linux.pyx
leofang: Explained and tracked in https://github.com/NVIDIA/cuda-python/issues/962

cuda_bindings/cuda/bindings/_internal/nvjitlink_windows.pyx
leofang: ditto https://github.com/NVIDIA/cuda-python/issues/962

leofang (approved):
Thanks a lot, @mdboom! Be sure to backport this PR.

PR #946, opened 2025-09-04
Fixing cuda_bindings local build manifesting with GCC-13 on Linux (WSL)

rparolin:
## Description

closes 

## Checklist

### DIFF
--- a/cuda_bindings/setup.py
+++ b/cuda_bindings/setup.py
@@ -248,12 +248,13 @@ if sys.platform != "win32":
         "-std=c++14",
         "-fpermissive",
         "-Wno-deprecated-declarations",
-        "-D _GLIBCXX_ASSERTIONS",
         "-fno-var-tracking-assignments",
     ]
     if "--debug" in sys.argv:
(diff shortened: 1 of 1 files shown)


rparolin:
pre-commit.ci autofix

rwgk:
Just sharing an observation:

> Running unit tests for `cuda_bindings` via `python -m pytest tests/` was manifesting the following error on WSL.

Interesting. I'm using "GCC-13 on Linux (WSL)", too, but never saw that error. An obvious difference at first glance is that I don't use conda very much when running the cuda_bindings unit tests.

rparolin:
> Just sharing an observation:
> 
> > Running unit tests for `cuda_bindings` via `python -m pytest tests/` was manifesting the following error on WSL.
> 
> Interesting. I'm using "GCC-13 on Linux (WSL)", too, but never saw that error. An obvious difference at first glance is that I don't use conda very much when running the cuda_bindings unit tests.

We may want to do an audit of the binaries publish via pypi and conda. I suspect we unintentionally enabled debug symbols somewhere.

PR #953, opened 2025-09-08
[scripts] Adding `run_tests.sh` script

rparolin:
[scripts] Adding `run_tests.sh` script

The `run_tests.sh` script runs all the nested cuda_python modules unit tests and reports back a unified status.

image


leofang:
Ralf is probably the main user of the toolshed scripts, so let me request his review

cuda_core/README.md
```diff
@@ -16,6 +16,12 @@ To run these tests:
 * `python -m pytest tests/` with editable installations
 * `pytest tests/` with installed packages
 
+Alternatively, from the repository root you can use a simple script:
```
rwgk, line 19: The `toolshed/` directory was meant to be non-public. Not sure about referencing it from here. See next comment.

toolshed/run_all_tests.sh
```diff
@@ -0,0 +1,264 @@
+#!/usr/bin/env bash
```
rwgk: WDYT about moving this to `tests/integration`?

rparolin: I don't have a strong opinion about where it should live.  If  there is a implication here that `toolshed` is meant for scripts that are not intended for others to use b/c we don't want to incur that support.  We could make a `scripts` folder to signal they are intended for public use -- aka by the CI -- and we guarantee we won't break them.  

I don't see a script like this being an integration test.  It would be used to eventually call the integration tests for the CUDA Python package.  But I'm open to changes if this not consistent to how y'all current work.

cryos: I would have put it in a `scripts` or `ci` folder (no strong feeling here). Totally bike shedding but as you very nicely offer parameters to run some tests and default to all it might make sense to simply call it `run_tests.sh`.

toolshed/run_all_tests.sh
```diff
@@ -0,0 +1,264 @@
...
+Targets:
+  all (default)   Run pathfinder → bindings → core
+  core            Run cuda_core tests
+  bindings        Run cuda_bindings tests
+  pathfinder      Run cuda_pathfinder tests
+  smoke           Run meta-level smoke tests (tests/integration)
+
+Options:
+  -v, --verbose       Verbose pytest output (-ra -s -v)
+      --install       Force editable install with [test] extras
```
cpcloud: This seems a bit out of scope for a script that runs tests. Personally I would be surprised if a script called run_test_all started installing stuff and assumed that `pip install -e .` is a thing from wherever I ran it.

Installing dependencies seems like a separate concern from running tests.

toolshed/run_all_tests.sh
```diff
@@ -0,0 +1,264 @@
...
+print_help() {
+  cat <<'USAGE'
+Usage: toolshed/run_all_tests.sh [options] [target]
+
+Targets:
+  all (default)   Run pathfinder → bindings → core
+  core            Run cuda_core tests
+  bindings        Run cuda_bindings tests
+  pathfinder      Run cuda_pathfinder tests
+  smoke           Run meta-level smoke tests (tests/integration)
```
cpcloud: It seems like this could all be captured with judicious use of `pytest.mark`. Is there some reason not to use that?

toolshed/run_all_tests.sh
rparolin: Done.  Thanks for the feedback.👍

cuda_core/README.md
rparolin: Moved to a `scripts` folder.

toolshed/run_all_tests.sh
rparolin: Yeah, I was on the fence about this but I was wanting this to be "one stop shop" script that handles everything a user would need when running from a clean branch.

toolshed/run_all_tests.sh
rparolin: I'll take a look at `pytest.mark`.  I've never used them before.

rparolin: I add pytest.mark support to the script.

cuda_core/README.md
rwgk: I still see `toolshed` here, oversight?

scripts/run_tests.sh
```diff
@@ -0,0 +1,263 @@
+#!/usr/bin/env bash
+
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+set -euo pipefail
+
+# Simple, dependency-free orchestrator to run tests for all packages.
+# Usage:
+#   toolshed/run_tests.sh [ -v|--verbose ] [ --install | --no-install ] [ --with-cython | --skip-cython ]
```
rwgk: `toolshed` still here, too

rparolin: oops.  Thank you!  Fixed.

tests/integration/test_smoke.py
```diff
@@ -0,0 +1,21 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import importlib
+
+
+def test_import_cuda_top_level():
+    m = importlib.import_module("cuda")
+    assert hasattr(m, "__file__")
```
cpcloud:
```suggestion
    import cuda
```

There's no point in calling `importlib.import_module` if you statically know the module name. The `hasattr` check also seem superfluous.

tests/integration/test_smoke.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def test_import_subpackages_present_if_installed():
```
cpcloud: What is the purpose of this test?

Nothing is being verified, since you're ignoring the exception that would be raised if in fact they are not installed, and if they are installed, nothing about the modules is verified.

conftest.py
```diff
@@ -0,0 +1,45 @@
...
+    cuda_home = os.environ.get("CUDA_HOME")
+    for item in items:
+        nodeid = item.nodeid.replace("\\", "/")
+
+        # Package markers by path
+        if (
+            nodeid.startswith("cuda_pathfinder/tests/")
+            or "/cuda_pathfinder/tests/" in nodeid
+        ):
+            item.add_marker(pytest.mark.pathfinder)
```
cpcloud, line 18: Is there actually any benefit to this? It seems like it's only providing the ability to not have to type `cuda_` when testing. 

For comparison:

The case of a single submodule:

```
pytest -m bindings
```

versus

```
pytest cuda_bindings
```

saves two characters


The case of multiple submodules:

```
pytest -m 'bindings or pathfiner'
```

versus

```
pytest cuda_{bindings,pathfinder}
```

I know I was the one who originally requested this, but I think there's not a lot of value in custom scripts versus learning the existing tools and how to use them effectively and conveniently.

cuda_pathfinder/tests/conftest.py
```diff
@@ -10,10 +10,7 @@ def pytest_configure(config):
 
 
 def pytest_terminal_summary(terminalreporter, exitstatus, config):  # noqa: ARG001
```
cpcloud: If this function isn't doing anything, can it be deleted?

rparolin: Deleted.

tests/integration/test_smoke.py
rparolin: I was leaving it as a place holder to encourage integration tests to be written.

cpcloud: IMO it's best to leave out placeholders and instead to do a follow up with something that isn't a placeholder but an actual integration test. This code ultimately has to be maintained, so if it isn't running anything useful then there's no reason to keep it around.

conftest.py
rparolin: Yeah, I had a similar question when I implemented this... and did some research.  The only thing I read that made me think it was worth it was marks enables us to define groups of tests down the line.  I agree if all we do is just map modules 1:1 to marks then its of limited value.  I figured thats was behind you requesting to use it. LOL.

tests/integration/test_smoke.py
rparolin: Removed.

cuda_core/README.md
rwgk: Sorry I didn't Unresolve this yesterday. There are still three `toolshed` paths here.

conftest.py
rwgk: > Is there actually any benefit to this?

Maybe if we also run the examples?

I stripped down the script we used for SWQA testing (preparation for CTK 13 release):

* comments removed
* `pip install` commands removed
* Most env vars removed

This is what was run:

```
cd cuda_pathfinder/
python -m pytest -s -vv tests/

cd cuda_bindings/
python -m pytest -s -vv tests/
CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM=1 python -m pytest -s -vv tests/
python -m pytest -s -vv examples/
python -m pytest -s -vv tests/cython/

cd cuda_core/
python -m pytest -s -vv tests/
python -m pytest -s -vv tests/example_tests/
```

Note that we ran the `cuda_bindings` tests twice, a second time with `CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM=1`

I'm pretty sure adding in running the examples will be useful.

I totally don't know how important it is to test with `CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM=1`. @leofang for input

leofang: xref https://github.com/NVIDIA/cuda-python/issues/971

cuda_core/README.md
rparolin: Should be resolved now. Sorry for the mistake.

conftest.py
rparolin: I added support .

```
(base) rparolin@NV-5FGL564:~/cuda-python$ ./scripts/run_tests.sh --help
Usage: scripts/run_tests.sh [options] [target]

Targets:
  all (default)   Run pathfinder → bindings → core
  core            Run cuda_core tests
  bindings        Run cuda_bindings tests
  pathfinder      Run cuda_pathfinder tests
  smoke           Run meta-level smoke tests (tests/integration)

Options:
  -v, --verbose       Verbose pytest output (-ra -s -v)
      --install       Force editable install with [test] extras
      --no-install    Skip install checks (assume environment is ready)
      --with-cython   Build and run cython tests (needs CUDA_HOME for core)
      --skip-cython   Skip cython tests (default)
      --with-examples Run examples where applicable (e.g., cuda_bindings/examples)
      --skip-examples Skip running examples (default)
      --with-ptds     Re-run cuda_bindings tests with PTDS (CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM=1)
  -h, --help          Show this help and exit

Examples:
  scripts/run_tests.sh --install
  scripts/run_tests.sh --no-install core
  scripts/run_tests.sh -v --with-cython bindings
  scripts/run_tests.sh smoke
(base) rparolin@NV-5FGL564:~/cuda-python$```

**Comment by rparolin** [comment]
@cpcloud @rwgk Please re-review, if this is ready to submit to main.

File: cuda_core/README.md [hunk 10]
```diff
@@ -16,6 +16,12 @@ To run these tests:
...
+
+* `./scripts/run_all_tests.sh core` to run only `cuda_core` tests
+* `./scripts/run_all_tests.sh` to run all package tests (pathfinder → bindings → core)
+* `./scripts/run_all_tests.sh smoke` to run meta-level smoke tests under `tests/integration`
```
rwgk: `run_all_tests.sh` → `run_tests.sh` (3x)

rwgk (approved):
I didn't review `run_tests.sh` in detail, but I'll try it out as soon as this PR is merged!

cuda_core/README.md
rparolin: Doh. Thanks!  Fixed.

rparolin:
@rwgk Sorry to trouble you again for another review.  Its making me get the PR re-approved after I made your requested change.

rwgk:
> @rwgk Sorry to trouble you again for another review. Its making me get the PR re-approved after I made your requested change.

Sorry I missed this before! Great to see this merged.

rparolin:
Thanks everyone!  👍 🚀

cuda_pathfinder/tests/conftest.py
```diff
@@ -9,13 +9,6 @@ def pytest_configure(config):
     config.custom_info = []
 
 
-def pytest_terminal_summary(terminalreporter, exitstatus, config):  # noqa: ARG001
```
rwgk, line 12: Sorry I overlooked this somehow:

@rparolin Removing this breaks an important part of the pathfinder tests: the `INFO` lines with paths found are no longer shown. I assume you removed this code for a reason? What was it?

Here is an example of what we are losing by removing this code:

https://github.com/NVIDIA/cuda-python/pull/956#issuecomment-3281597257

rparolin: Are those pathfinder tests not run as part of gated commit checks?  It shouldn't have let me commit if that was the case.  

I think I removed it based on review feedback -- let me double check.

rwgk: > Are those pathfinder tests not run as part of gated commit checks?

The pathfinder tests run, but the CI scripts don't check for the presence of the `INFO` lines.

I didn't want to spend time before on figuring out such a check, but I have now:

* https://github.com/NVIDIA/cuda-python/pull/979

PR #957, opened 2025-09-09
Updates KernelAttributes to avoid possible dangling handles.

Andy-Jost:
Previously a `KernelAttributes` bundle held a reference to an internal handle from a `Kernel` object. This change introduces a weak reference to ensure that a dangling handle is never used. Testing shows no statistically significant change to property access times.

### DIFF
--- a/cuda_core/cuda/core/experimental/_module.py
+++ b/cuda_core/cuda/core/experimental/_module.py
@@ -2,6 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+import weakref
 from collections import namedtuple
 from typing import Optional, Union
 from warnings import warn
@@ -60,12 +61,12 @@ class KernelAttributes:
     def __new__(self, *args, **kwargs):
         raise RuntimeError("KernelAttributes cannot be instantiated directly. Please use Kernel APIs.")
 
-    slots = ("_handle", "_cache", "_backend_version", "_loader")
+    slots = ("_kernel", "_cache", "_backend_version", "_loader")
 
     @classmethod
-    def _init(cls, handle):
+    def _init(cls, kernel):
         self = super().__new__(cls)
-        self._handle = handle
+        self._kernel = weakref.ref(kernel)
         self._cache = {}
 
         self._backend_version = "new" if (_py_major_ver >= 12 and _driver_ver >= 12000) else "old"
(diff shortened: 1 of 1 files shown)


cuda_core/cuda/core/experimental/_module.py
```diff
@@ -76,15 +77,18 @@ def _get_cached_attribute(self, device_id: int, attribute: driver.CUfunction_att
         """Helper function to get a cached attribute or fetch and cache it if not present."""
         if device_id in self._cache and attribute in self._cache[device_id]:
```
Andy-Jost: I also see an opportunity to simplify the logic throughout to use pairs as keys:

`if (device_id, attribute) in self._cache:`

Andy-Jost: Fixed in 651beb9243360c66f9482c94a98619f8a0be24bf

cuda_core/cuda/core/experimental/_module.py
```diff
@@ -74,20 +75,21 @@ def _init(cls, handle):
 
     def _get_cached_attribute(self, device_id: int, attribute: driver.CUfunction_attribute) -> int:
         """Helper function to get a cached attribute or fetch and cache it if not present."""
-        if device_id in self._cache and attribute in self._cache[device_id]:
-            return self._cache[device_id][attribute]
+        if (device_id, attribute) in self._cache:
```
rwgk: WDYT about

```
        cache_key = (device_id, attribute)  # so this tuple doesn't have to be rebuilt
        result = self._cache.get(cache_key, cache_key)  # the tuple doubles as sentinel; there is only one cache lookup
        if result is not cache_key:
            return result
```

then all the way below
```
        self.cache[cache_key] = result
        return result
```

rwgk (approved):
Looks good to me.

My suggestion is meant to be a minor by easy optimization.

cuda_core/cuda/core/experimental/_module.py
Andy-Jost: Good suggestion, thanks!

rparolin (approved):
LGTM.

Question: Is it possible to write a unit test that exercises the weakref?

Andy-Jost:
> Looks great to me!
> 
> Sorry I missed the full context: How did you discover the bug fixed in this PR? Is there a reasonably easy way to add a test that would have created a dangling handle? — I wouldn't spend more than 10 minutes on that.

Leo asked me to bundle the attributes for IPC-enabled mempools and pointed to this as a reference implementation. I noticed this issue by inspection when reading the code. There is an example test called `test_mempool_attributes_ownership` [here](https://github.com/NVIDIA/cuda-python/pull/930/files#diff-706670744803a013c5b6d93f1d156ab5caf9f355a6eae6760d37371f0cf0491f) (sorry I couldn't find a way to link directly to the code block).

Maybe to be more specific, it was this code that set off the alarm bells:

`KernelAttributes(self._handle)`

When an object (`self` here) represents resource ownership, then slicing a part off to share ownership is almost always dangerous.

Andy-Jost:
> LGTM.
> 
> Question: Is it possible to write a unit test that exercises the weakref?

I looked into this and I think it would be difficult right now. The reference management is a bit of a mess. E.g., there is a reference cycle between Kernel, which refers to a Module, which refers to its Kernels. That makes it difficult to accurately delete Kernels as we'd need for this test. If we cleaned up the references overall it would be easier.

leofang:
Sorry I did not have a chance to read this PR. It is fine. But there was not bug, and this PR is not a bug fix. 

In case it is not clear, this was implemented by design. `Kernel`'s underlying handle `CUkernel` (or `CUfunction`) is not something that users can create or destroy out of nowhere. Their lifetime is tied to the parent `ObjectCode` (with the handle `CUlibrary`, or `CUmodule`). And because we implement it (also by design, there's probably a comment somewhere in the same file) in the way that a loaded `CUlibrary` would not be unloaded (to mimic how fatbins embedded in a shared library works), it is guaranteed that there is no dangling pointer.

PR #961, opened 2025-09-10
Regen bindings to use `is`/`is not` `None` instead of equality checks

cpcloud:
This PR implements proper None checks using `is`/`is not`, contains code regenerated from https://github.com/NVIDIA/cuda-python-private/pull/146.

### DIFF
--- a/cuda_bindings/cuda/bindings/driver.pyx.in
+++ b/cuda_bindings/cuda/bindings/driver.pyx.in
@@ -24688,7 +24688,7 @@ def cuCtxCreate(ctxCreateParams : Optional[CUctxCreateParams], unsigned int flag
         pdev = int(CUdevice(dev))
     cydev = <cydriver.CUdevice>pdev
     cdef CUcontext pctx = CUcontext()
-    cdef cydriver.CUctxCreateParams* cyctxCreateParams_ptr = ctxCreateParams._pvt_ptr if ctxCreateParams != None else NULL
+    cdef cydriver.CUctxCreateParams* cyctxCreateParams_ptr = ctxCreateParams._pvt_ptr if ctxCreateParams is not None else NULL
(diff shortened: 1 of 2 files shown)


rwgk (approved):
Looks good to me.

Thoughts:

* In theory we could undo the corresponding change in the code generator, and still run pre-commit against the code generator code,
* but I think it's better to merge this PR, so we don't have to think so much what we're doing where.

leofang:
@cpcloud feel free to merge, one thing that's technically missing is a release note entry, but for this one it is up to you if you wanna document it or not

leofang:
Also plz regenerate and backport to 12.9.x

cpcloud:
Is there a checklist of things that must be done when a PR is merged that ideally we can work towards automating?

cpcloud:
backport is here https://github.com/NVIDIA/cuda-python/pull/970

leofang:
> Is there a checklist of things that must be done when a PR is merged that ideally we can work towards automating?

There is a [PR template](https://github.com/NVIDIA/cuda-python/blob/main/.github/PULL_REQUEST_TEMPLATE.md) which is ignored in this PR, though I don't think it matters much for the team members IMHO; most of time I just let it populate the text but does not read them either. I'd love to have more automation, but there are cases for which having a release note entry is too noisy, so an escape hatch is also needed.

cpcloud:
I see. It looks like the PR template could use some TLC.

leofang:
Yeah. For example, another thing that would be great to have is a backport reminder when a PR touches cuda-bindings on main, but I dunno how the PR template could help. Right now I just manually go through merged PRs every Sunday and check which one hasn't been backported 🤷  (This is tracked in https://github.com/NVIDIA/cuda-python/issues/965 FWIW.)

Issue #967, opened 2025-09-12
[FEA]: Add a `MemoryResource` which uses CUDA VMM APIs to allocate memory


benhg:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

cuda.core

### Is your feature request related to a problem? Please describe.

I would like to be able to use the equivalent of `cuMemCreate`, `cuMemMap`, and friends via a `cuda.core` `MemoryResource`

### Describe the solution you'd like

I'd like to have a `VMMAllocatedMemoryResource` which I can create on a `Device()` for which `allocate()` will use the `cuMem***` driver APIs to create memory.

### Describe alternatives you've considered

Currently the only alternative is to use the bindings APIs directly. Since the `cuMem***` functions are synchronous, there's no way to fit this with the MemPool APIs as-is (this is my current understanding, at least.)

### Additional context

This is useful to support NVSHMEM/NCCL external buffer registration, or for more interesting cases like growing allocations without changing pointer addresses, or EGM on Grace-Hopper or Grace-Blackwell systems.

leofang:
Implemented in #968.

Issue #462, opened 2025-02-22
CI: Also test Windows + WDDM mode


leofang:
#444 added a temporary Windows GPU runner with T4 so that we can finally start testing on Windows 🎉 However, it only runs on TCC mode and there is also WDDM mode that behaves differently (ex: TCC does not support mempools #206). Ideally we should test both and `nvidia-smi` allows a system to switch modes. WDDM mode is also important for applications that require handling of both graphics and compute.

However, after some desperate efforts (#461) I realized it is more complicated that just calling `nvidia-smi` inside the VM, because the default driver for data center GPUs like T4 do not support WDDM out of box. We need to install what's known as the ["vGPU" driver](https://docs.nvidia.com/vgpu/index.html) to enable the switch, and vGPU is behind a paywall (license needed)...

PR #968, opened 2025-09-12
[FEA]: Add a `MemoryResource` which uses CUDA VMM APIs to allocate memory

benhg:
## Description

This is an implementation for #967. 

Summary:

- Implement VMM grow-in-place via:
    - Fast path: contiguous VA extension with fixedAddr + cuMemMap
    - Slow path: remap to new VA, preserve data without memcpy
- Use aligned sizes everywhere and cast pointers for arithmetic
- Return proper enums for CUmemAccessDesc.flags
- Default all VMMConfig fields to common values

References:

[NVIDIA VMM blog](https://developer.nvidia.com/blog/introducing-low-level-gpu-virtual-memory-management/)
[CUDA Driver VMM docs](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__VA.html#group__CUDA__VA)

## Checklist


leofang:
@benhg the IPC PR was merged, could you resolve conflicts plz? 🙂

benhg:
> @benhg the IPC PR was merged, could you resolve conflicts plz? 🙂

Done, thanks for the reminder.

leofang:
pre-commit.ci autofix

cuda_core/cuda/core/experimental/__init__.py
```diff
@@ -20,6 +20,8 @@
     IPCChannel,
     LegacyPinnedMemoryResource,
     MemoryResource,
+    VMMAllocatedMemoryResource,
+    VMMAllocationOptions,
```
leofang: 1. These two names should align -- one is "allocated" another is "allocation"
   - @Andy-Jost we forgot to expose `DeviceMemoryResorceOptions` as public API
2. Instead of `VMMAllocatedMemoryResource`, how about a shorter, more informative one: `VirtualMemoryResource`?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,462 @@ class _SynchronousMemoryResource(MemoryResource):
...
+        addr_hint: A (optional) virtual address hint to try to reserve at. 0 -> let CUDA choose.
+        addr_align: Alignment for the VA reservation. If None, use the queried granularity.
+        peers: Extra device IDs that should be granted access in addition to `device`.
+        self_access: Access flags for the owning device ('rw', 'r', or 'none').
+        peer_access: Access flags for peers ('rw' or 'r').
+    """
+    # Human-friendly strings; normalized in __post_init__
+    allocation_type: str = "pinned"          # pinned
+    location_type: str = "device"            # device
+    handle_type: str = "posix-fd"           # posix-fd | generic | none
```
leofang: based on my limited understanding in typing, we could make it more informative
```python
VMMHandleTypeT = Literal["posix-fd", "generic", "none"]

@dataclass
class VMMAllocationOptions:
   ...
   handle_type: VMMHandleTypeT = "posix_fd"
```

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,462 @@ class _SynchronousMemoryResource(MemoryResource):
...
+            return driver.CUmemAllocationType.CU_MEM_ALLOCATION_TYPE_MANAGED
+        raise ValueError(f"Unsupported allocation_type: {spec!r}")
+
+    @staticmethod
+    def _location_type_to_driver(spec: str):
+        if spec == "device":
+            return driver.CUmemLocationType.CU_MEM_LOCATION_TYPE_DEVICE
+        if spec == "host":
+            return driver.CUmemLocationType.CU_MEM_LOCATION_TYPE_HOST
+        if spec == "host-numa":
```
leofang: nit: prefer underscore over dash

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,462 @@ class _SynchronousMemoryResource(MemoryResource):
...
+        Device ordinal for which a memory resource is constructed. The mempool that is
+        set to *current* on ``device_id`` is used. If no mempool is set to current yet,
+        the driver would use the *default* mempool on the device.
+
+    config : VMMAllocationOptions
+        A configuration object for the VMMAllocatedMemoryResource
+    """
+    def __init__(self, device, config: VMMAllocationOptions = None):
+        self.device = device
+        self.config = config or VMMAllocationOptions()
```
leofang: btw we have a `check_or_create_options` util that can be used for this

cuda_core/cuda/core/experimental/__init__.py
benhg: How's `VirtualMemoryResource`  and `VirtualMemoryResourceOptions` ?

cuda_core/cuda/core/experimental/_memory.pyx
benhg: Did not know about that feature. Will do.

cuda_core/cuda/core/experimental/_memory.pyx
benhg: Sure

leofang:
(I haven't done any serious review yet, just some drive-by comments 😅)

cuda_core/tests/test_memory.py
```diff
@@ -301,6 +307,133 @@ def test_device_memory_resource_initialization(mempool_device, use_device_object
     buffer.close()
 
 
+def test_vmm_allocator_basic_allocation():
```
Andy-Jost, line 317: I wonder how much of the nonspecific testing can be covered by parameterizing other tests on the kind of memory resource.

cuda_core/tests/test_memory.py
```diff
@@ -301,6 +307,133 @@ def test_device_memory_resource_initialization(mempool_device, use_device_object
...
+    # Grow the allocation
+    grown_buffer = vmm_mr.modify_allocation(buffer, 4 * 1024 * 1024)
+
+    # Verify growth
+    assert grown_buffer.size >= 4 * 1024 * 1024
+    assert grown_buffer.size > original_size
+
+    # The pointer should ideally be preserved (fast path)
+    # but may change if contiguous extension fails (slow path)
+    assert grown_buffer.handle is not None
```
Andy-Jost, line 438: I don't see how this test relates to the comment.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,468 @@ class _SynchronousMemoryResource(MemoryResource):
...
+        self.device = device
+        self.config = config or VirtualMemoryResourceOptions()
+
+    def _align_up(self, size: int, gran: int) -> int:
+        """
+        Align a size up to the nearest multiple of a granularity.
+        """
+        return (size + gran - 1) & ~(gran - 1)
+
+    def modify_allocation(self, buf: Buffer, new_size: int, config: VirtualMemoryResourceOptions = None) -> Buffer:
```
Andy-Jost, line 1207: Should this be named `resize`?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,468 @@ class _SynchronousMemoryResource(MemoryResource):
     @property
     def device_id(self) -> int:
         return self._dev_id
+
+VirtualMemoryHandleTypeT = Literal["posix_fd", "generic", "none"]
```
Andy-Jost: No problem with this code but I'm curious how we decide when to use enums versus strings.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,468 @@ class _SynchronousMemoryResource(MemoryResource):
...
+@dataclass
+class VirtualMemoryResourceOptions:
+    """A configuration object for the VirtualMemoryResource
+       Stores configuration information which tells the resource how to use the CUDA VMM APIs
+    """
+    """
+    Configuration for CUDA VMM allocations.
+
+    Args:
+        handle_type: Export handle type for the physical allocation. Use
```
Andy-Jost, line 1104: Formatting looks strange. Be sure to match other docstrings.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,468 @@ class _SynchronousMemoryResource(MemoryResource):
...
+    addr_hint: Optional[int] = 0
+    addr_align: Optional[int] = None
+    peers: Iterable[int] = field(default_factory=tuple)
+    self_access: VirtualMemoryAccessTypeT = "rw"
+    peer_access: VirtualMemoryAccessTypeT = "rw"
+
+    @staticmethod
+    def _access_to_flags(spec: str):
+        f = driver.CUmemAccess_flags
+        if spec == "rw":
```
Andy-Jost: nit: it would be nice to structure these string-to-enum mappings rather than write if-trees, since they appear to recur.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,468 @@ class _SynchronousMemoryResource(MemoryResource):
...
+        f = driver.CUmemAllocationGranularity_flags
+        if spec == "minimum":
+            return f.CU_MEM_ALLOC_GRANULARITY_MINIMUM
+        if spec == "recommended":
+            return f.CU_MEM_ALLOC_GRANULARITY_RECOMMENDED
+        raise ValueError(f"Unsupported granularity: {spec!r}")
+
+
+class VirtualMemoryResource(MemoryResource):
+    """Create a device memory resource that uses the CUDA VMM APIs to allocate memory.
```
Andy-Jost, line 1180: What is the relationship to `DeviceMemoryResource`?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,468 @@ class _SynchronousMemoryResource(MemoryResource):
...
+    def _grow_allocation_fast_path(self, buf: Buffer, new_size: int, prop: driver.CUmemAllocationProp,
+                                   aligned_additional_size: int, new_ptr: int) -> Buffer:
+        """
+        Fast path: extend the VA range contiguously.
+
+        This preserves the original pointer by mapping new physical memory
+        to the extended portion of the virtual address range.
+        """
+        # Create new physical memory for the additional size
+        res, new_handle = driver.cuMemCreate(aligned_additional_size, prop, 0)
```
Andy-Jost: The error handling looks scary. Perhaps structure it as a transaction?

```
with Transaction() as trans:
    trans.append(lambda: driver.cuMemAddressFree(new_ptr, aligned_additional_size)) # register undo action
    new_handle = handle_return(driver.cuMemCreate(aligned_additional_size, prop, 0)) # may raise
    trans.append(lambda: driver.cuMemRelease(new_handle)) # register undo action
    . . .
    trans.commit() # cancel undo actions
    return buf
```

cuda_core/cuda/core/experimental/_memory.pyx
benhg: This VirtualMemoryResource is conceptually similar to the DeviceMemoryResource, in that they both create memory allocations on the device side (that is, they allocate GPU Vidmem).

The difference is that DeviceMemoryResource uses `cudaMalloc` (a CUDA Runtime API) to allocate that memory, while this one uses `cuMemCreate`, `cuMemMap`, etc. (CUDA Driver APIs) to allocate that memory.

cuda_core/cuda/core/experimental/_memory.pyx
benhg: From talking to Leo offline, it seems like it's preferred to use strings in general in `cuda.core` and translate them to `cuda.bindings` Enums.

In NVSHMEM4Py (which tries to conceptually follow cuda-python as much as possible), we expose our C enums through `nvshmem.bindings`, and then re-expose them with name aliases in `nvshmem.core`

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: What I mean is that in terms of the implementation and user experience, how do they relate? Should they share underlying code? What should be the result of `isinstance(vm_mr, DeviceMemoryResource)`?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,468 @@ class _SynchronousMemoryResource(MemoryResource):
...
+
+        # Unmap the old VA range (aligned previous size)
+        aligned_prev_size = total_aligned_size - aligned_additional_size
+        result, = driver.cuMemUnmap(int(buf.handle), aligned_prev_size)
+        if result != driver.CUresult.CUDA_SUCCESS:
+            driver.cuMemAddressFree(new_ptr, total_aligned_size)
+            driver.cuMemRelease(old_handle)
+            raise Exception(f"Failed to unmap old allocation: {result}")
+
+        # Remap the old physical memory to the new VA range (aligned previous size)
```
Andy-Jost: What level of exception safety are we aiming for? At this point, for example, if mapping to the new VA range fails but the old VA range has already been unmapped what is the state of the buffer? Is it left in an invalid state? Could the program crash?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,468 @@ class _SynchronousMemoryResource(MemoryResource):
...
+                d = driver.CUmemAccessDesc()
+                d.location.type = driver.CUmemLocationType.CU_MEM_LOCATION_TYPE_DEVICE
+                d.location.id = int(peer_dev)
+                d.flags = peer_flags
+                descs.append(d)
+
+        return descs
+
+
+    def allocate(self, size: int, stream: Stream = None) -> Buffer:
```
Andy-Jost, line 1455: `stream` should always be `None` correct?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,468 @@ class _SynchronousMemoryResource(MemoryResource):
...
+        if res != driver.CUresult.CUDA_SUCCESS:
+            driver.cuMemAddressFree(ptr, aligned_size)
+            driver.cuMemRelease(handle)
+            raise Exception(f"cuMemMap failed: {res}")
+
+        # ---- Set access for owner + peers ----
+        descs = []
+
+        # Owner access
+        owner_flags = VirtualMemoryResourceOptions._access_to_flags(config.self_access)
```
Andy-Jost: Isn't this a copy-paste of _build_access_descriptors?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,468 @@ class _SynchronousMemoryResource(MemoryResource):
...
+                                   aligned_additional_size: int, total_aligned_size: int, addr_align: int) -> Buffer:
+        """
+        Slow path: full remapping when contiguous extension fails.
+
+        This creates a new VA range and remaps both old and new physical memory.
+        The buffer's pointer will change.
+        """
+        # Reserve a completely new, larger VA range
+        res, new_ptr = driver.cuMemAddressReserve(total_aligned_size, addr_align, 0, 0)
+        if res != driver.CUresult.CUDA_SUCCESS:
```
Andy-Jost: I think it would reduce a lot of clutter to use `raise_if_driver_error` throughout the code. I think we should revamp the error handling in general but that's another story.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: Python enums are generally yucky, they also happen to be not performant (although this is not in the critical path).

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: Perhaps consider the following:

```
_access_flags = {"rw": driver.CUmemAccess_flags.CU_MEM_ACCESS_FLAGS_PROT_READWRITE,
                 "r": driver.CUmemAccess_flags.CU_MEM_ACCESS_FLAGS_PROT_READ,
                 "none": 0}
```

And then:
```
flags = _access_flags.get(string)
```

benhg: That sounds like a better experience. I can do this all for all the string to enum translations.

cuda_core/cuda/core/experimental/_memory.pyx
benhg: Oh, I see. I don't think they should share code because if we were to do that, it'd result in a lot of deep if/else blocks. I think `isinstance(vm_mr, DeviceMemoryResource)` should be False, and `isinstance(vm_mr, MemoryResource)` should of course be True.

cuda_core/cuda/core/experimental/_memory.pyx
benhg: Can do.

cuda_core/cuda/core/experimental/_memory.pyx
benhg: Will do. Thanks for the pointer to `raise_if_driver_error`. I'm not yet very familiar with the internals of cuda-core so this is a learning experience for me.

cuda_core/cuda/core/experimental/_memory.pyx
benhg: Yes, adding assertion.

cuda_core/cuda/core/experimental/_memory.pyx
benhg: In principle, there are other modifications you can do. For example, if you pass `new_size` equal to the current size, but different properties, we could in theory create a new buffer with the same size but different allocation properties (for example, switching gpuDirectRDMACapable from 0 to 1). But in practice, I suspect that won't be useful very often.

Perhaps we should remove the early-return so that the property changes can happen even if the size doesn't change.

cuda_core/cuda/core/experimental/_memory.pyx
benhg: In this case, we aren't releasing the old memory until the end of this, so the state should be that the old buffer is still valid, and we have undone all of the actions we did to create the new one, so it functionally doesn't exist.

I think structuring the error path here as transactions will also help clear it up.

cuda_core/cuda/core/experimental/_memory.pyx
benhg: Other way around, I copied this into that function then forgot to move the original to calling the function :)

cuda_core/cuda/core/experimental/_memory.pyx
benhg: I removed the early-return and replaced it with a special case for smaller or equal size allocation where we just update the properties but don't do any of the remapping.

cuda_core/tests/test_memory.py
benhg: I wrote this as a justification for why we don't assert that the pointer is the same. But let me be more clear.

benhg:
pre-commit.ci autofix

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,499 @@ class _SynchronousMemoryResource(MemoryResource):
...
+    addr_hint: Optional[int] = 0
+    addr_align: Optional[int] = None
+    peers: Iterable[int] = field(default_factory=tuple)
+    self_access: VirtualMemoryAccessTypeT = "rw"
+    peer_access: VirtualMemoryAccessTypeT = "rw"
+
+    @staticmethod
+    def _access_to_flags(spec: str):
+        f = driver.CUmemAccess_flags
+        _access_flags = {"rw": f.CU_MEM_ACCESS_FLAGS_PROT_READWRITE, "r": f.CU_MEM_ACCESS_FLAGS_PROT_READ, "none": 0}
```
kkraus14: super nitpick, but it would be good to move these lookup tables outside of the functions so we don't pay the cost of creating them each invocation

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,499 @@ class _SynchronousMemoryResource(MemoryResource):
...
+
+class VirtualMemoryResource(MemoryResource):
+    """Create a device memory resource that uses the CUDA VMM APIs to allocate memory.
+
+    Parameters
+    ----------
+    device_id : int
+        Device ordinal for which a memory resource is constructed. The mempool that is
+        set to *current* on ``device_id`` is used. If no mempool is set to current yet,
+        the driver would use the *default* mempool on the device.
```
kkraus14: Was this copied from `DeviceMemoryResource`? Should it be updated based on this memory resource which isn't based on the mempool?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,499 @@ class _SynchronousMemoryResource(MemoryResource):
...
+    config : VirtualMemoryResourceOptions
+        A configuration object for the VirtualMemoryResource
+    """
+    def __init__(self, device, config: VirtualMemoryResourceOptions = None):
+        self.device = device
+        self.config = check_or_create_options(
+            VirtualMemoryResourceOptions, config, "VirtualMemoryResource options", keep_none=False
+        )
+
+    def _align_up(self, size: int, gran: int) -> int:
```
kkraus14: Should this be a static method?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,499 @@ class _SynchronousMemoryResource(MemoryResource):
...
+            int(buf.handle) + aligned_prev_size,  # fixedAddr hint - aligned end of current range
+            0
+        )
+
+        if res != driver.CUresult.CUDA_SUCCESS or new_ptr != (int(buf.handle) + aligned_prev_size):
+            # Fallback: couldn't extend contiguously, need full remapping
+            return self._grow_allocation_slow_path(buf, new_size, prop, aligned_additional_size, total_aligned_size, addr_align)
+        else:
+            # Success! We can extend the VA range contiguously
+            return self._grow_allocation_fast_path(buf, new_size, prop, aligned_additional_size, new_ptr)
```
kkraus14, line 1280: Is there a specific set of error codes or cases where we should fall into the slow path? Right now if we hit some kind of unexpected CUDA error we'll go down the slow path instead of erroring clearly to the user.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,499 @@ class _SynchronousMemoryResource(MemoryResource):
...
+            prop (driver.CUmemAllocationProp): Allocation properties for the new memory.
+            aligned_additional_size (int): The size of the new region to allocate, aligned to granularity.
+            new_ptr (int): The address of the newly reserved contiguous VA region (should be at the end of the current buffer).
+
+        Returns:
+            Buffer: The same buffer object with its size updated to `new_size`.
+        """
+        with Transaction() as trans:
+            # Create new physical memory for the additional size
+            trans.append(lambda: driver.cuMemAddressFree(new_ptr, aligned_additional_size))
```
kkraus14: Should we check / handle errors of this in some way?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,499 @@ class _SynchronousMemoryResource(MemoryResource):
...
+            res, new_handle = driver.cuMemCreate(aligned_additional_size, prop, 0)
+            raise_if_driver_error(res)
+            # Register undo for creation
+            trans.append(lambda: driver.cuMemRelease(new_handle))
```
kkraus14: +1

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,499 @@ class _SynchronousMemoryResource(MemoryResource):
...
+
+            # Map the new physical memory to the extended VA range
+            res, = driver.cuMemMap(new_ptr, aligned_additional_size, 0, new_handle, 0)
+            raise_if_driver_error(res)
+            # Register undo for mapping
+            trans.append(lambda: driver.cuMemUnmap(new_ptr, aligned_additional_size))
```
kkraus14: +1

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,499 @@ class _SynchronousMemoryResource(MemoryResource):
...
+
+        Returns:
+            Buffer: The buffer object updated with the new pointer and size.
+        """
+        with Transaction() as trans:
+            # Reserve a completely new, larger VA range
+            res, new_ptr = driver.cuMemAddressReserve(total_aligned_size, addr_align, 0, 0)
+            raise_if_driver_error(res)
+            # Register undo for VA reservation
+            trans.append(lambda: driver.cuMemAddressFree(new_ptr, total_aligned_size))
```
kkraus14: +1

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,499 @@ class _SynchronousMemoryResource(MemoryResource):
...
+
+            # Get the old allocation handle for remapping
+            result, old_handle = driver.cuMemRetainAllocationHandle(buf.handle)
+            raise_if_driver_error(result)
+            # Register undo for old_handle
+            trans.append(lambda: driver.cuMemRelease(old_handle))
```
kkraus14: +1

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,499 @@ class _SynchronousMemoryResource(MemoryResource):
...
+            aligned_prev_size = total_aligned_size - aligned_additional_size
+            result, = driver.cuMemUnmap(int(buf.handle), aligned_prev_size)
+            raise_if_driver_error(result)
+
+            def _remap_old():
+                # Try to remap the old physical memory back to the original VA range
+                try:
+                    driver.cuMemMap(int(buf.handle), aligned_prev_size, 0, old_handle, 0)
+                except Exception:
+                    pass
```
kkraus14, line 1375: The driver call wouldn't throw here but would return a CUDA error code that would need to be handled

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,499 @@ class _SynchronousMemoryResource(MemoryResource):
...
+            trans.append(_remap_old)
+
+            # Remap the old physical memory to the new VA range (aligned previous size)
+            res, = driver.cuMemMap(int(new_ptr), aligned_prev_size, 0, old_handle, 0)
+            raise_if_driver_error(res)
+
+            # Register undo for mapping
+            trans.append(lambda: driver.cuMemUnmap(new_ptr, aligned_prev_size))
```
kkraus14: +1

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,499 @@ class _SynchronousMemoryResource(MemoryResource):
...
+
+            # Create new physical memory for the additional size
+            res, new_handle = driver.cuMemCreate(aligned_additional_size, prop, 0)
+            raise_if_driver_error(res)
+
+            # Register undo for new physical memory
+            trans.append(lambda: driver.cuMemRelease(new_handle))
```
kkraus14: +1

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,499 @@ class _SynchronousMemoryResource(MemoryResource):
...
+
+            # Map the new physical memory to the extended portion (aligned offset)
+            res, = driver.cuMemMap(int(new_ptr) + aligned_prev_size, aligned_additional_size, 0, new_handle, 0)
+            raise_if_driver_error(res)
+
+            # Register undo for mapping
+            trans.append(lambda: driver.cuMemUnmap(int(new_ptr) + aligned_prev_size, aligned_additional_size))
```
kkraus14: +1

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,499 @@ class _SynchronousMemoryResource(MemoryResource):
...
+            descs = self._build_access_descriptors(prop)
+            if descs:
+                res, = driver.cuMemSetAccess(new_ptr, total_aligned_size, descs, len(descs))
+                raise_if_driver_error(res)
+
+            # All succeeded, cancel undo actions
+            trans.commit()
+
+        # Free the old VA range (aligned previous size)
+        driver.cuMemAddressFree(int(buf.handle), aligned_prev_size)
```
kkraus14: Check for errors?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,499 @@ class _SynchronousMemoryResource(MemoryResource):
...
+            d = driver.CUmemAccessDesc()
+            d.location.type = prop.location.type
+            d.location.id = prop.location.id
+            d.flags = owner_flags
+            descs.append(d)
+
+        # Peer device access
+        peer_flags = VirtualMemoryResourceOptions._access_to_flags(self.config.peer_access)
+        for peer_dev in self.config.peers:
+            if peer_flags:
```
kkraus14: Can short circuit the loop
```suggestion
        if peer_flags:
            for peer_dev in self.config.peers:
```

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,499 @@ class _SynchronousMemoryResource(MemoryResource):
...
+        """
+        if stream is not None:
+            raise NotImplementedError("Stream is not supported with VirtualMemoryResource")
+
+        config = self.config
+        # ---- Build allocation properties ----
+        prop = driver.CUmemAllocationProp()
+        prop.type = VirtualMemoryResourceOptions._allocation_type_to_driver(config.allocation_type)
+        # TODO: Support host alloation if required
+        if  prop.type != driver.CUmemLocationType.CU_MEM_LOCATION_TYPE_DEVICE:
```
kkraus14:
```suggestion
        if prop.type != driver.CUmemLocationType.CU_MEM_LOCATION_TYPE_DEVICE:
```

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,499 @@ class _SynchronousMemoryResource(MemoryResource):
...
+        aligned_size = self._align_up(size, gran)
+        addr_align = config.addr_align or gran
+
+        # ---- Transactional allocation ----
+        with Transaction() as trans:
+            # ---- Create physical memory ----
+            res, handle = driver.cuMemCreate(aligned_size, prop, 0)
+            raise_if_driver_error(res)
+            # Register undo for physical memory
+            trans.append(lambda: driver.cuMemRelease(handle))
```
kkraus14: +1

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,499 @@ class _SynchronousMemoryResource(MemoryResource):
...
+
+            # ---- Reserve VA space ----
+            # Potentially, use a separate size for the VA reservation from the physical allocation size
+            res, ptr = driver.cuMemAddressReserve(aligned_size, addr_align, config.addr_hint, 0)
+            raise_if_driver_error(res)
+            # Register undo for VA reservation
+            trans.append(lambda: driver.cuMemAddressFree(ptr, aligned_size))
```
kkraus14: +1

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,499 @@ class _SynchronousMemoryResource(MemoryResource):
...
+
+            # ---- Map physical memory into VA ----
+            res, = driver.cuMemMap(ptr, aligned_size, 0, handle, 0)
+            trans.append(lambda: driver.cuMemUnmap(ptr, aligned_size))
```
kkraus14: +1

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,499 @@ class _SynchronousMemoryResource(MemoryResource):
...
+            raise_if_driver_error(res)
+
+            # ---- Set access for owner + peers ----
+            descs = self._build_access_descriptors(prop)
+            if descs:
+                res, = driver.cuMemSetAccess(ptr, aligned_size, descs, len(descs))
```
kkraus14, line 1530: handle errors?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,499 @@ class _SynchronousMemoryResource(MemoryResource):
...
+        raise_if_driver_error(result)
+
+
+    @property
+    def is_device_accessible(self) -> bool:
+        """
+        Indicates whether the allocated memory is accessible from the device.
+
+        Returns:
+            bool: Always True for NVSHMEM memory.
```
kkraus14: This VMM memory resource isn't only for nvshmem though. Given we're only supporting device memory for now it still always returns True, but we may want to add a TODO comment to handle this based on `self.config.allocation_type`?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -900,3 +903,499 @@ class _SynchronousMemoryResource(MemoryResource):
...
+        """
+        return True
+
+    @property
+    def is_host_accessible(self) -> bool:
+        """
+        Indicates whether the allocated memory is accessible from the host.
+
+        Returns:
+            bool: Always False for NVSHMEM memory.
```
kkraus14: +1

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -10,17 +10,20 @@ from cuda.core.experimental._utils.cuda_utils cimport (
...
 import cython
 import os
 import platform
 import weakref
+
+
 from cuda.core.experimental._dlpack import DLDeviceType, make_py_capsule
 from cuda.core.experimental._stream import Stream, default_stream
-from cuda.core.experimental._utils.cuda_utils import driver
+from cuda.core.experimental._utils.cuda_utils import driver, Transaction, get_binding_version
```
kkraus14: @leofang why are we importing `driver` from here?

cuda_core/cuda/core/experimental/_memory.pyx
benhg: This is a copy-paste legacy. Actually, it's probably easy enough to just support the hostalloc flag so we may as well do it here and I'll remove this entirely.

cuda_core/cuda/core/experimental/_memory.pyx
benhg: Yes, it was copied, and I'll edit.

cuda_core/cuda/core/experimental/_memory.pyx
benhg: You mean if the rollback from the end of the Transaction ends up failing? If you mean in the case that errors occur, the idea of the Transaction is that if something raises, all the steps before it get rolled back.

If the rollback / recovery steps fail, I'm not sure there's much we can do.

kkraus14: Yea, if the `driver.cuMemAddressFree(new_ptr, aligned_additional_size)` here returns an error code here as opposed to succeeding. Since these APIs return error codes and don't throw, we should probably at the least check them and error loudly and clearly to the user if something fails unexpectedly in the roll back path?

benhg: Makes sense. I'll add typechecking that throws exceptions, so that it kicks in the transaction and prints a useful message. And I'll do that everywhere within all the transactions.

benhg: @kkraus14 could you take a look at the new `wrap_driver_function_with_error_handling` function I added which basically makes the transaction check for errors upon rollback?

cuda_core/cuda/core/experimental/_memory.pyx
benhg: Yes, I will edit this to check the error code in a less binary way.

benhg:
pre-commit.ci autofix

benhg:
pre-commit.ci autofix

cuda_core/cuda/core/experimental/_memory.pyx
kkraus14: Can we use the existing `handle_return` function?

benhg: Did not know that existed. Yes, will do.

benhg: Done

benhg:
pre-commit.ci autofix

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -25,11 +25,15 @@ import weakref
...
 from cuda.core.experimental._utils.cuda_utils import ( driver, Transaction, get_binding_version, 
-                                                       wrap_driver_function_with_error_handling,
+                                                       handle_return,
                                                      )
 
 if platform.system() == "Linux":
     import socket
+# Helper to invoke CUDA driver functions with standard error handling.
+def _driver_call(func, /, *args):
+    handle_return(func(*args))
```
kkraus14: Does this abstraction add value over just using `handle_return` directly?

We also already import `_check_driver_error as raise_if_driver_error` which if we don't need return values from the driver calls then that will be more efficient. Apologies for pointing towards `handle_return` instead of the driver specific function.

benhg: The only value add is saving me from having to pass `lambda` everywhere and making the lines really long. But I don't mind doing that not to add this one. 

I can switch to raise_if_driver_error instead of handle_return.

cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
```diff
@@ -222,3 +224,51 @@ def get_binding_version():
...
+        self._entered = False
+
+    def __enter__(self):
+        self._stack.__enter__()
+        self._entered = True
+        return self
+
+    def __exit__(self, exc_type, exc, tb):
+        # If exit callbacks remain, they'll run in LIFO order.
+        return self._stack.__exit__(exc_type, exc, tb)
```
Andy-Jost, line 269: nit: consider setting `self._entered = False` here?

benhg: Sure

cuda_core/cuda/core/experimental/_memory.pyx
benhg: Done

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -11,20 +11,25 @@ from cuda.core.experimental._utils.cuda_utils cimport (
...
+
 from cuda.core.experimental._dlpack import DLDeviceType, make_py_capsule
 from cuda.core.experimental._stream import Stream, default_stream
-from cuda.core.experimental._utils.cuda_utils import driver
+from cuda.core.experimental._utils.cuda_utils import ( driver, Transaction, get_binding_version )
 
 if platform.system() == "Linux":
     import socket
+# (Removed) helper for driver calls; use raise_if_driver_error with direct calls instead.
+
```
kkraus14:
```suggestion
```

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -906,3 +911,496 @@ class _SynchronousMemoryResource(MemoryResource):
...
+            aligned_additional_size,
+            addr_align,
+            int(buf.handle) + aligned_prev_size,  # fixedAddr hint - aligned end of current range
+            0
+        )
+
+        if res != driver.CUresult.CUDA_SUCCESS or new_ptr != (int(buf.handle) + aligned_prev_size):
+            # Check for specific errors that are not recoverable with the slow path
+            if res in (driver.CUresult.CUDA_ERROR_INVALID_VALUE, driver.CUresult.CUDA_ERROR_NOT_PERMITTED, driver.CUresult.CUDA_ERROR_NOT_INITIALIZED, driver.CUresult.CUDA_ERROR_NOT_SUPPORTED):
+                raise RuntimeError(f"Failed to extend VA range: {res}")
```
kkraus14: Maybe use `raise_if_driver_error(res)` here as well? Would be good to consistently raise the driver error.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -906,3 +911,496 @@ class _SynchronousMemoryResource(MemoryResource):
...
+            result, = driver.cuMemUnmap(int(buf.handle), aligned_prev_size)
+            raise_if_driver_error(result)
+
+            def _remap_old():
+                # Try to remap the old physical memory back to the original VA range
+                try:
+                    driver.cuMemMap(int(buf.handle), aligned_prev_size, 0, old_handle, 0)
+                except Exception:
+                    pass
+            trans.append(_remap_old)
```
kkraus14, line 1376: This still needs to be handled, where `driver.cuMemMap` wouldn't throw but would return an error code that needs to be handled

benhg: Oops, missed that one. Fixing...

leofang (changes requested):
Thanks, Ben! Sorry I was not as responsive as I'd like for reviews. Keith told me this is ready to go. We have a number of PRs that have conflicts with each other, so what I'd do is to merge all P0 tasks for the release, and then merge this the last, to minimize conflicts. By the time we get to merge this PR, I will resolve all conflicts and rerun tests. No action is needed.

Marking this PR red only to avoid unplanned merge, again no action is needed. Much appreciated, Ben!

leofang:
I am working on resolving the merge conflicts logically.

leofang (approved):

leofang:
/ok to test [6450712](https://github.com/NVIDIA/cuda-python/pull/968/commits/6450712da7e5d1e43b48d0168e534605fa5f552b)

cuda_core/cuda/core/experimental/_memory.pyx
leofang: Because it offered us a way to not worry about new/old cuda.bindings layouts. It is no longer needed after the cythonization because we need to stick to the layout. But we don't need to clean it up in this PR.

benhg:
/ok to test [4af54ac](https://github.com/NVIDIA/cuda-python/commit/4af54ac5ec9015f19bcc27e5a083fc77318c1805)

leofang:
> Vetter: a user that has permissions to leave an /ok to test <SHA> comment on a pull request. This is a user with write access (or greater) for a particular repository or a member that is listed in the additional_vetters configuration list (see note below)

I always forget that not all NV employees can launch the CI... 😞

leofang:
/ok to test [4af54ac](https://github.com/NVIDIA/cuda-python/pull/968/commits/4af54ac5ec9015f19bcc27e5a083fc77318c1805)

kkraus14 (dismissed):

leofang:
Ben, Piotr and I discussed with Vishnu from the memory team, and confirmed that VMM is not supported by the TCC mode. I suggested Ben to skip the tests on Windows for now.. perhaps also raise a warning (or exception?) on `VirtualMemoryResource` constructor. We can revisit this once we turn on WDDM (#462) and MCDM CI (#984).

leofang (approved):

leofang:
/ok to test [9db04b1](https://github.com/NVIDIA/cuda-python/pull/968/commits/9db04b10715f1c83c26cf8cbf1700bfe8f431bfa)

leofang:
Thanks Ben for your first contribution to cuda.core and Keith/Andy for review 🎉

PR #970, opened 2025-09-15
refactor: replace none equality checks with `is`/`is not` equivalents

cpcloud:
This PR implements proper None checks using `is`/`is not`, contains code regenerated from https://github.com/NVIDIA/cuda-python-private/pull/146.


cpcloud:
It does change the code, because `==` (and `!=`) must produce behavior equivalent to `PyObject_RichCompare`. Cython generates the rich comparison call for `==` and `!=` and generates C code like this:

```c
__pyx_v_x != Py_None
```

when `is not` is used to compare `None`.

PR #975, opened 2025-09-16
feat: add support for CUcheckpointRestoreArgs

cpcloud:
This PR adds the newly generated code from https://github.com/NVIDIA/cuda-python-private/pull/147.

### DIFF
--- a/cuda_bindings/setup.py
+++ b/cuda_bindings/setup.py
@@ -150,7 +150,7 @@ def parse_headers(header_dict):
         "\\(enum ": "(",
         # Since we only support 64 bit architectures, we can inline the sizeof(T*) to 8 and then compute the
         # result in Python. The arithmetic expression is preserved to help with clarity and understanding
-        r'char reserved\[52 - sizeof\(CUcheckpointGpuPair \*\)\];' : rf'char reserved[{52 - 8}];',
+        r"char reserved\[52 - sizeof\(CUcheckpointGpuPair \*\)\];": rf"char reserved[{52 - 8}];",
     }
 
     print(f'Parsing headers in "{include_path_list}" (Caching = {PARSER_CACHING})')
--- a/cuda_bindings/setup.py
+++ b/cuda_bindings/setup.py
@@ -148,6 +148,9 @@ def parse_headers(header_dict):
         "  enum ": "   ",
         ", enum ": ", ",
         "\\(enum ": "(",
+        # Since we only support 64 bit architectures, we can inline the sizeof(T*) to 8 and then compute the
+        # result in Python. The arithmetic expression is preserved to help with clarity and understanding
+        r'char reserved\[52 - sizeof\(CUcheckpointGpuPair \*\)\];' : rf'char reserved[{52 - 8}];',
     }
 
     print(f'Parsing headers in "{include_path_list}" (Caching = {PARSER_CACHING})')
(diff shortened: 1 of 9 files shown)


leofang:
Looks good! Thanks Philip! Let's add a release note entry before merging. Something along the line of "Restoring the `cuda.bindings.driver.cuCheckpointProcessRestore` API removed by mistake" should work. (It was there in 12.9.x.)

cpcloud:
Looks like I need to get it to compile first :)

cpcloud:
@leofang Is there something else I need to do to get these to compile? I'm getting the following errors:

```
      Error compiling Cython file:
      ...

          See Also
          --------
          :py:obj:`~.cuInit`
          """
          cdef cydriver.CUcheckpointRestoreArgs* cyargs_ptr = args._pvt_ptr if args is not None else NULL
                                                                                                     ^

      cuda/bindings/driver.pyx:45481:95: Cannot convert 'void *' to Python object

      Error compiling Cython file:
      ...

          See Also
          --------
          :py:obj:`~.cuInit`
          """
          cdef cydriver.CUcheckpointRestoreArgs* cyargs_ptr = args._pvt_ptr if args is not None else NULL
                                                                  ^

      cuda/bindings/driver.pyx:45481:60: Cannot convert Python object to 'CUcheckpointRestoreArgs *'

      Error compiling Cython file:
      ...

          See Also
          --------
          :py:obj:`~.cuInit`
          """
          cdef cydriver.CUcheckpointRestoreArgs* cyargs_ptr = args._pvt_ptr if args is not None else NULL
                                                                                                     ^

      cuda/bindings/driver.pyx:45481:95: Cannot convert Python object to 'CUcheckpointRestoreArgs *'

      Error compiling Cython file:
      ...

          See Also
          --------
          :py:obj:`~.cuInit`
          """
          cdef cydriver.CUcheckpointRestoreArgs* cyargs_ptr = args._pvt_ptr if args is not None else NULL
                                                              ^

      cuda/bindings/driver.pyx:45481:56: Unsafe C derivative of temporary Python reference used in conditional expression

      Error compiling Cython file:
      ...

          See Also
          --------
          :py:obj:`~.cuInit`
          """
          cdef cydriver.CUcheckpointRestoreArgs* cyargs_ptr = args._pvt_ptr if args is not None else NULL
               ^

      cuda/bindings/driver.pyx:45481:9: Storing unsafe C derivative of temporary Python reference
```

cpcloud:
Looking at other similar structs like `CUcheckpointLockArgs` the `driver.pyx.in` code is identical.

cpcloud:
Oh, it looks like `driver.pxd` doesn't contain the `cdef class` for the restore args struct.

cpcloud:
Alright, I'll figure this out, down the rabbit hole!

cpcloud:
Ok, the issue is that finding the type depends on another parse of the C files succeeding, and there are no replacements like in `cuda-python-private`, so the parse fails and thus the struct doesn't exist and is never found.

cpcloud:
For now I am going to copy that replacement in here.

cuda_bindings/setup.py
leofang, line 1: Yes. Sorry for the frustration. This is the thing we discussed earlier. We parse twice so whatever we did in the codegen carries over here. I'll create an issue.

leofang (approved):

leofang (approved):

leofang:
Since the CI was green and I only touched the docs, let me admin-merge to save resources.

cuda_bindings/setup.py
leofang: https://github.com/NVIDIA/cuda-python-private/issues/149

PR #989, opened 2025-09-19
Replace is not None checks for linker flags with booleans

mmason-nvidia:
The existing code checks for requested linker flags using "is not None" instead of testing for True/False. This can result in unwanted flags being passed to the linker in some instances. This change replaces the "is not None" check with simple boolean checks where appropriate.


leofang:
/ok to test [6a31f6b](https://github.com/NVIDIA/cuda-python/pull/989/commits/6a31f6b3046ccf942f4494e08029a795be8e6a9c)

leofang:
/ok to test [61a5fad](https://github.com/NVIDIA/cuda-python/pull/989/commits/61a5fadd4d723b5032aaeaf33ab3e7d081922367)

leofang (approved):

PR #995, opened 2025-09-19
refactor(deps): organize test and dev dependencies using dependency groups

cpcloud:
## Description
I didn't open an issue as it seemed better to discuss this in the context of
actual code instead of hypotheticals.

This PR reorganizes the `cuda_*` packages to use
[`dependency-groups`](https://peps.python.org/pep-0735/) to manage test and
development dependencies.

`optional-dependencies` (pip extras, really) was, for a long time, the only way
to spell this kind of thing, but since PEP 735, and its [corresponding
implementation in `pip`
25.1](https://ichard26.github.io/blog/2025/04/whats-new-in-pip-25.1/#dependency-groups-pep-735)
there's a better way, which among other things hides these development-only
dependencies from vanilla installations and makes it clear by reading
a `pyproject.toml` what is development-only and what isn't.

I tried to catch all the scripts that are using this, but seeing as how our CI steps are distributed
across a number of scripts which contains lots of implicit references to environment variables, I may
have missed some. Let's see what CI says.

### DIFF
--- a/.github/workflows/build-wheel.yml
+++ b/.github/workflows/build-wheel.yml
@@ -227,7 +227,7 @@ jobs:
 
       - name: Build cuda.bindings Cython tests
         run: |
-          pip install $(ls ${{ env.CUDA_BINDINGS_ARTIFACTS_DIR }}/*.whl)[test]
+          pip install ${{ env.CUDA_BINDINGS_ARTIFACTS_DIR }}/*.whl --group ./cuda_bindings/pyproject.toml:test
           pushd ${{ env.CUDA_BINDINGS_CYTHON_TESTS_DIR }}
           bash build_tests.sh
           popd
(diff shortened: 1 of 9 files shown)


.github/workflows/build-wheel.yml
```diff
@@ -227,7 +227,7 @@ jobs:
 
       - name: Build cuda.bindings Cython tests
         run: |
-          pip install $(ls ${{ env.CUDA_BINDINGS_ARTIFACTS_DIR }}/*.whl)[test]
+          pip install ${{ env.CUDA_BINDINGS_ARTIFACTS_DIR }}/*.whl --group ./cuda_bindings/pyproject.toml:test
```
kkraus14, line 230: Do we have to refer to the `pyproject.toml` here? Is there any way we can just use `--group test`?

cpcloud: We do, because `pip` only looks for `pyproject.toml` in the passed installation directory, and since we're passing a wheel and not an installation directory it needs to be told where to find the group definition. This is just an unfortunate consequence of `pip install` being a megacommand that does way too many things.

kkraus14 (approved):

leofang:
It would be nice to document how this works for internal/external developers as well as QA. Let me create an issue to follow up.

PR #999, opened 2025-09-19
ci: remove spurious free-threading builds

cpcloud:
These will not build, so they should be removed

PR #1008, opened 2025-09-22
ci: replace single gating check with `wait-for-status-check` [no-ci]

cpcloud:
This PR changes our usage of `status-check.yml` to allow skipping CI when the head commit of PR contains the text
`[skip ci]`.

Our current usage of the `wait-for-check` action doesn't sense to me.

We have as inputs all the checks that we care about `build`, `test*`, `docs`,
and we only invoke that action when those succeed.

That seems to be no different than running `exit 0` against `ubuntu-latest` and
thus defeats the purpose of using `wait-for-checks` which is to allow
finer-grained control over what is actually a required check.

In our case, there are cases such as #999 where it doesn't make sense to do
a full build test docs run, but developers are left waiting on an admin to
merge their PR.

Proposal for merging this PR:

1. Make the previously not required checks actually required in the GitHub UI.
   I don't have permissions to do that, so someone other than me needs to do
   that.
1. Merge this PR.
1. Try out a couple no-op PRs to verify that everything is working.


.github/workflows/ci.yml
```diff
@@ -109,12 +109,29 @@ jobs:
...
+        uses: poseidon/wait-for-status-checks@899c768d191b56eef585c18f8558da19e1f3e707  # v0.6.0
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          match_pattern: "CI*"
+
+  skip-checks:
+    name: Skip checking job status
+    runs-on: ubuntu-latest
+    # we can only skip if we're checking against a PR **and** its head commit says to skip ci
+    if: ${{ startsWith(github.ref_name, 'pull-request/') && contains(github.event.push.head_commit.message, '[skip ci]') }}
```
kkraus14: Instead of checking the head commit message, can we check the PR title? There are times when a PR evolves and we decide to skip CI during its lifecycle.

.github/workflows/status-check.yml
```diff
@@ -1,20 +0,0 @@
...
-  checks:
-    name: Check job status
-    runs-on: ubuntu-latest
-    steps:
-      - name: GitHub Checks
-        uses: poseidon/wait-for-status-checks@899c768d191b56eef585c18f8558da19e1f3e707  # v0.6.0
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          match_pattern: "CI*"
-          ignore_pattern: ".*Check job status.*"  # ignore self
```
leofang, line 20: As noted on Slack let's have a call and I can explain how things work. 

Just a drive-by note: This `ignore_pattern` is mistakenly dropped. It is there to work around this bug: https://github.com/poseidon/wait-for-status-checks/issues/303. (I wasn't aware that this bug report exists, I just made the same observation when setting it up 🙂)

.github/workflows/ci.yml
```diff
@@ -163,15 +163,37 @@ jobs:
...
+      - name: Get PR title
+        id: get-pr-title
+        run: echo "pr_title=$(gh pr view "$(grep -Po '(\d+)$' <<< '${{ github.ref_name }}')" --json title --jq '.title')" >> "$GITHUB_OUTPUT"
+      - name: GitHub Checks
+        if: contains(steps.get-pr-title.outputs.pr_title, '[skip ci]')
+        uses: poseidon/wait-for-status-checks@899c768d191b56eef585c18f8558da19e1f3e707  # v0.6.0
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          # ignore all checks if skipping CI
+          ignore_pattern: ".*"
```
leofang: 1. I think we can merge this step into the previous job (`checks`) and nuke the remaining job (`skip-checks`) for clarity. The if condition can be as complex as:
```yaml
- name: GitHub Checks (SKIP)
  if: contains(steps.get-pr-title.outputs.pr_title, '[skip ci]') && startsWith(github.ref_name, 'pull-request/')
  ...
```
2. When we discussed this offline, I recall you had some ideas to avoid using the `wait-for-status-checks` GHA entirely. Is it not feasible you think?

.github/workflows/ci.yml
```diff
@@ -163,15 +163,37 @@ jobs:
...
     permissions:
       checks: read
-    needs:
-      - build-linux-64
-      - test-linux-64
-      - build-linux-aarch64
-      - test-linux-aarch64
-      - build-windows
-      - test-windows
-      - doc
```
leofang, line 175: Note: By removing `needs`, the `checks` job will be launched as soon as the CI starts (building wheels) and keep spinning. On the CI workflow diagram, we won't see a flow chart that converging to the final stage. All tests will have an open end. (Not saying it's good or bad, but it is a noticeable change that's worth calling out.)

leofang:
/ok to test [6563550](https://github.com/NVIDIA/cuda-python/pull/1008/commits/65635508c7d21392799db9f62b24fafbb7734cdc)

.github/workflows/ci.yml
leofang: > Note: By removing `needs`, the `checks` job will be launched as soon as the CI starts (building wheels) and keep spinning. 

Oh... I recall why I did it the way it is today... When the `check` job starts, it is very likely that the other jobs are still being spawned, and so it waits for nothing and just completes. Example: https://github.com/NVIDIA/cuda-python/actions/runs/18147180929/job/51651080141?pr=1008

So the existing implementation we have today is correct. My statement above is not fully correct, because instead of spinning it just completes.

> On the CI workflow diagram, we won't see a flow chart that converging to the final stage. All tests will have an open end. (Not saying it's good or bad, but it is a noticeable change that's worth calling out.)

See https://github.com/NVIDIA/cuda-python/actions/runs/18147180929?pr=1008 for what I meant.

leofang (changes requested):

.github/workflows/ci.yml
cpcloud: Well, that behavior could easily result from another source: incorrect specification of the match pattern, which I am fairly certain is what went wrong there.

The current pattern is `CI*` which matches **workflow** names not checks, which are **job** names.

cpcloud: Basically, the match pattern needs to be something like `(build|test).*`.

.github/workflows/ci.yml
cpcloud: It might be possible. Let me play with it a bit.

leofang:
Check job still completes instead of spinning: https://github.com/NVIDIA/cuda-python/actions/runs/18165463144/job/51706593048?pr=1008

.github/workflows/ci.yml
cpcloud: I don't think it's possible to avoid using this action if we want to allow skipping checks based on a condition.

You'd somehow need to encode the ability to run one job based on one condition with `needs`, meaning the check jobs and the condition would need to be in the `needs` array, and if the condition is false there the remaining jobs would be forever waited on (even though the condition is false).

cpcloud:
@leofang Can you paste the names of the checks here from the UI? I can't see them because I don't have access.

leofang:
Once we are sure the check job completes the last, we can either adjust the repo setting and then merge, or admin-merge then adjust the repo setting. But we are not there yet. The repo setting is irrelevant of the failure of waiting for dependent jobs here. It is just a workflow bug that we should fix first.

.github/workflows/ci.yml
leofang: Makes sense, thanks!

cpcloud:
I'm going to try one more alternative to avoid the action entirely.

cpcloud:
I really don't like having to add an arbitrary delay to get this to work.

leofang:
Yeah adding a random delay is flakey. It could proceed to complete when build jobs finish but the test jobs are still being spawned. Not a fan either.

cpcloud:
@leofang I think this is ready for another review. It seems to be doing what (I think) we want.

.github/workflows/ci.yml
```diff
@@ -156,22 +179,20 @@ jobs:
...
     name: Check job status
-    permissions:
-      checks: read
+    runs-on: ubuntu-latest
     needs:
-      - build-linux-64
       - test-linux-64
-      - build-linux-aarch64
       - test-linux-aarch64
-      - build-windows
```
cpcloud, line 173: I removed these since the tests require the builds.

cpcloud:
The current `Check job status / Check job status` required check needs to be changed to `CI / Check job status`.

leofang:
> The current `Check job status / Check job status` required check needs to be changed to `CI / Check job status`.

I applied it. You can see it now becomes "required". I haven't removed the old check yet. 

But I am confused: If this is required, we still need to execute `/ok to test`, skip build/test/doc jobs, and make this job execute and turn green, right? But the current logic (mysteriously) also skips the status check job. Am I missing something?

cpcloud:
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks

The new status check job **should** be considered passing, because it's skipped, according to that documentation.

I'm try to taking advantage of the fact that skip == success.

cpcloud:
Figured it out: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks

cpcloud:
@leofang It looks like the `push` check is somehow different from the check that's waiting. Are push checks separate in the checks UI?

leofang:
I am checking. This is odd because as discussed earlier once a CI job is run, it should appear as a registered job in the repo setting's dropdown menu, but for some reason I had to type the full name explicitly without anything popping out for me to choose.

leofang:
This is what's in the ruleset that Keith set up. I added the last row manually. Nothing is "push"-specific 🤔 
截圖 2025-10-01 下午12 04 57

leofang:
@cpcloud I think I fixed it. It's just "Check job status" (without "CI"), see the last row:
截圖 2025-10-01 下午12 11 17
You can see your new job becomes required now. But, it does not seem to be allowing a green merge button? Is it because I still haven't removed the old check?

cpcloud:
> Is it because I still haven't removed the old check?

Yep, that has to be removed before merge is allowed.

leofang:
Let me check quickly. btw we'll need to backport this PR to 12.9.x. Keith applied the same ruleset to all branches.

leofang:
It works now! To avoid race condition I am merging immediately.

cpcloud:
Thanks for the review, much appreciated!

leofang:
Thanks for working this out Phillip!

PR #1009, opened 2025-09-23
Improve speed of init functions in cybind-generated bindings

mdboom:
This is applying the same trick in https://github.com/NVIDIA/cuda-python/pull/894 to the cybind-generated bindings.

Essentially, by inlining the boolean check in every call, we can avoid the much more expensive call through a C function pointer on every single API call.

### DIFF
--- a/cuda_bindings/cuda/bindings/_internal/nvjitlink_windows.pyx
+++ b/cuda_bindings/cuda/bindings/_internal/nvjitlink_windows.pyx
@@ -93,7 +93,7 @@ cdef void* __nvJitLinkGetInfoLog = NULL
 cdef void* __nvJitLinkVersion = NULL
 
 
-cdef int _check_or_init_nvjitlink() except -1 nogil:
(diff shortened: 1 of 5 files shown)


cuda_bindings/cuda/bindings/_internal/nvjitlink_windows.pyx
```diff
@@ -93,7 +93,7 @@ cdef void* __nvJitLinkGetInfoLog = NULL
 cdef void* __nvJitLinkVersion = NULL
 
 
-cdef int _check_or_init_nvjitlink() except -1 nogil:
+cdef int __check_or_init_nvjitlink() except -1 nogil:
     global __py_nvjitlink_init
     if __py_nvjitlink_init:
         return 0
```
leofang, line 99: btw internally we dropped these two lines

leofang:
For posterity, this PR was backported as part of https://github.com/NVIDIA/cuda-python/pull/1044.

PR #1012, opened 2025-09-24
Add doc for Git pre-push hook to block accidental pushes to main/master

rparolin:
Adds a Markdown doc showing how to set up a global Git pre‑push hook that blocks pushes to main (and master) unless the explicit override flag `--break_glass_to_push_main` is used.  The override flag is intended to be loud and nosy to avoid accidental use.

### DIFF
--- a/docs/pre-push-branch-check.md
+++ b/docs/pre-push-branch-check.md
@@ -1,0 +1,100 @@
+# 🔒 Prevent Accidental Pushes to `main`/`master` — With Emergency Override
+
+This guide sets up a **global Git pre‑push hook** that blocks pushes to protected branches (`main` and `master`) unless you explicitly use the override flag:
+
+```
+--break_glass_to_push_main
+```
+
+---
+
+## 1. Create a Global Git Hooks Directory
+```bash
+mkdir -p ~/.git-hooks
+```
+
+---
+
+## 2. Create the Pre‑Push Hook
+```bash
+vim ~/.git-hooks/pre-push
+```
+
+Paste in:
+```bash
+#!/bin/sh
+branch="$(git symbolic-ref --short HEAD)"
+override_flag="--break_glass_to_push_main"
+
+if [ "$branch" = "main" ] || [ "$branch" = "master" ]; then
+    case "$*" in
+        *"$override_flag"* )
+            echo "⚠️ Override used: pushing to '$branch'..."
+            ;;
+        *)
+            echo "❌ Push to '$branch' is disabled locally."
+            echo "   If you **really** need to, use: git push $override_flag"
+            exit 1
+            ;;
+    esac
+fi
+```
+
+---
+
+## 3. Make the Hook Executable
+```bash
+chmod +x ~/.git-hooks/pre-push
+```
+
+---
+
+## 4. Tell Git to Use It Globally
+```bash
+git config --global core.hooksPath ~/.git-hooks
+```
+
+---
+
+## 5. Test It
+1. Switch to `main`:
+    ```bash
+    git checkout main
+    ```
+2. Try to push:
+    ```bash
+    git push
+    ```
+   ➡  Should be **blocked**.
+
+3. Use the override (emergency only):
+    ```bash
+    git push --break_glass_to_push_main
+    ```
+   ➡  Allowed with a warning.
+
+---
+
+## Quick Install Script
+```bash
+mkdir -p ~/.git-hooks && \
+cat > ~/.git-hooks/pre-push <<'EOF'
+#!/bin/sh
+branch="$(git symbolic-ref --short HEAD)"
+override_flag="--break_glass_to_push_main"
+
+if [ "$branch" = "main" ] || [ "$branch" = "master" ]; then
+    case "$*" in
+        *"$override_flag"* )
+            echo "⚠️ Override used: pushing to '$branch'..."
+            ;;
+        *)
+            echo "❌ Push to '$branch' is disabled locally."
+            echo "   If you **really** need to, use: git push $override_flag"
+            exit 1
+            ;;
+    esac
+fi
+EOF
+chmod +x ~/.git-hooks/pre-push
+git config --global core.hooksPath ~/.git-hooks


docs/pre-push-branch-check.md
```diff
@@ -0,0 +1,100 @@
...
+## 1. Create a Global Git Hooks Directory
+```bash
+mkdir -p ~/.git-hooks
+```
+
+---
+
+## 2. Create the Pre‑Push Hook
+```bash
+vim ~/.git-hooks/pre-push
```
cpcloud: Won't this affect every git repo a user has cloned? That seems like a pretty ostentatious choice to make for people.

docs/pre-push-branch-check.md
```diff
@@ -0,0 +1,100 @@
...
+    git push --break_glass_to_push_main
+    ```
+   ➡  Allowed with a warning.
+
+---
+
+## Quick Install Script
+```bash
+mkdir -p ~/.git-hooks && \
+cat > ~/.git-hooks/pre-push <<'EOF'
```
rwgk: The "global protection" feature looks interesting to me ("I want this"), but unconditionally overwriting someone's `~/.git-hooks/pre-push` could go wrong.

docs/pre-push-branch-check.md
rparolin: That's intended because the purpose of the hook is to prevent accidental pushes to main.  This is more of a problem for people with admin access because we locked down direct commits to main outside of PR merges.  But also, we haven't setup those rules for all repos yet.

docs/pre-push-branch-check.md
rparolin: That's a good point about overwriting someones pre-push hooks.  I can update the docs or just remove the quick install example.

docs/pre-push-branch-check.md
cpcloud: It sounds like you're assuming that all repos anyone will clone or has cloned will be either nvidia repos or repos that they will never want to or need to force push to.

Is that not a rather strong assumption?

docs/pre-push-branch-check.md
rparolin: Updated.

docs/pre-push-branch-check.md
```diff
@@ -0,0 +1,81 @@
...
+Open your existing hook:
+```bash
+vim ~/.git-hooks/pre-push
+```
+
+Paste in:
+```bash
+#!/bin/sh
+branch="$(git symbolic-ref --short HEAD)"
+override_flag="--break_glass_to_push_main"
```
rwgk: I followed the instructions here exactly.

It looks like the `override_flag` approach does not work:

```
rwgk-win11.localdomain:~/forked/cuda-python $ git sw main
Switched to branch 'main'
Your branch is up to date with 'upstream/main'.
```

```
rwgk-win11.localdomain:~/forked/cuda-python $ git push
❌ Push to 'main' is disabled locally.
   If you **really** need to, use: git push --break_glass_to_push_main
error: failed to push some refs to 'github.com:NVIDIA/cuda-python.git'
```

```
rwgk-win11.localdomain:~/forked/cuda-python $ git push --break_glass_to_push_main
error: unknown option `break_glass_to_push_main'
usage: git push [<options>] [<repository> [<refspec>...]]

    -v, --[no-]verbose    be more verbose
<...snip...>
    -6, --ipv6            use IPv6 addresses only
```

My favorite LLM is suggesting to use an environment variable, e.g. `GIT_PUSH_ALLOW_DIRTY_STATE=1 git push`, or  the `git -c` option, e.g. `git -c hook.breakglass=true push`.

rparolin: Bah.  Thanks.  Yeah, I tried a few alternatives.  My favorite LLM suggested using the pattern `git push -o break_glass_force_push` but I couldn't get that work reliably either.

The envvar method worked consistently for me.

image

I'll update the doc.  Thank you for testing this.

docs/pre-push-branch-check.md
rparolin: Point taken.  I've added per-repo instructions to the doc as an alternative to setting this pre-push hook globally.

rparolin:
> Looks very useful to me! This is a great way to spread the trick.
> 
> FWIW, I metabolized this already into my personal config:
> 
> https://github.com/rwgk/rwgk_config/blob/09bac54bacc35b763bf76dfa5b2917a81827e5c6/git_stuff/pre-push
> 
> That code is entirely LLM-generated. I just told the LLM that I wanted to integrate the branch protection into what I had already.
> 
> I was already set up with an alias to copy that into a git repo (cwd). After learning about the global `~/.git-hooks` trick here, I think that'll actually be better ... but I'm saving that for another day.

Thank you!

I was reading as part of the research for this that global pre-push hooks will override pre-repo unless you have your global hook call into the pre-repo hook.  I never tested it so never added it to the doc but its a good piece of knowledge to file away if you need it.

leofang:
@rparolin looks nice! We should hyperlink to it in the ToC here: https://github.com/NVIDIA/cuda-python/blob/09bac290ec76c76edc0ce9b4e242bd923f5b5f2e/CONTRIBUTING.md?plain=1#L17-L23
Alternatively, we can move the entire doc to the contributing guide as a separate section

kkraus14:
I didn't get a chance to review this before we merged, but I don't think documenting git / dev configs unrelated to the cuda-python project is something we should be doing.

We should have a GitHub ruleset and proper CI config that allows us to prevent everyone, including admins, from committing directly to protected branches while still allowing us to skip CI when needed.

rparolin:
> I didn't get a chance to review this before we merged, but I don't think documenting git / dev configs unrelated to the cuda-python project is something we should be doing.
> 
> We should have a GitHub ruleset and proper CI config that allows us to prevent everyone, including admins, from committing directly to protected branches while still allowing us to skip CI when needed.

👍 I'll remove it from the repo.  https://github.com/NVIDIA/cuda-python/pull/1026

PR #1014, opened 2025-09-25
ci: remove pwd and ls commands that add noise to CI output

cpcloud:
Small QoL enhancement to reduce the amount of noise coming out of the `run-tests` CI script


kkraus14 (approved):

kkraus14:
Should this be backported to `12.9.x`?

PR #1017, opened 2025-09-25
chore(deps): bump precommit versions

cpcloud:
Run autoupdate freeze on pre-commit and autofixed the lints produced by a newer version of ruff.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ ci:
 # pre-commit autoupdate --freeze
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: 0b19ef1fd6ad680ed7752d6daba883ce1265a6de  # frozen: v0.12.2
+    rev: f298305809c552671cc47e0fec0ba43e96c146a2  # frozen: v0.13.2
     hooks:
       - id: ruff
         args: [--fix, --show-fixes]
@@ -40,7 +40,7 @@ repos:
 
   # Standard hooks
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: "v5.0.0"
+    rev: "3e8a8703264a2f4a69428a0aa4dcb512790b2c8c"  # frozen: v6.0.0
     hooks:
     - id: check-added-large-files
     - id: check-case-conflict
@@ -58,14 +58,14 @@ repos:
 
   # Checking for common mistakes
   - repo: https://github.com/pre-commit/pygrep-hooks
-    rev: "v1.10.0"
+    rev: "3a6eb0fadf60b3cccfd80bad9dbb6fae7e47b316"  # frozen: v1.10.0
     hooks:
     - id: rst-backticks
     - id: rst-directive-colons
     - id: rst-inline-touching-normal
 
   - repo: https://github.com/PyCQA/bandit
-    rev: "36fd65054fc8864b4037d0918904f9331512feb5"  # frozen: 1.7.10 KEEP IN SYNC WITH .github/workflows/bandit.yml
+    rev: "2d0b675b04c80ae42277e10500db06a0a37bae17"  # frozen: 1.8.6
     hooks:
       - id: bandit
         args:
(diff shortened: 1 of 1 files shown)


leofang:
The new grouping of import statements is very annoying, we have been grouping the imports [following PEP 8](https://peps.python.org/pep-0008/#imports) just fine and it's now broken. Can we please just skip whatever new rule is implemented in Ruff?

cpcloud:
Broken feels a bit extreme--the code still runs--but that's fine by me. PEP8 is probably not something we should religiously follow FWIW, it has its problems.

cpcloud:
In any case, I'll revert those changes.

cpcloud:
@leofang I think the import modifications are correct actually. Notice that not every file is formatted, so this is just bringing various other files in line with the rest of the codebase.

.github/workflows/bandit.yml
```diff
@@ -25,4 +25,4 @@ jobs:
         # via `pip install bandit[sarif]`. If runner Python moves to >=3.9,
         # the action will resolve to 1.8.x and you'll need to bump pre-commit.
         # (Bandit >=1.8.0 dropped Python 3.8 via Requires-Python metadata.)
-        uses: PyCQA/bandit-action@8a1b30610f61f3f792fe7556e888c9d7dffa52de  # v1.0.0
+        uses: PyCQA/bandit-action@67a458d90fa11fb1463e91e7f4c8f068b5863c7f  # v1.0.0
```
kkraus14: How did the commit change without the tag changing...?

cpcloud: Because I manually updated the hash and forgot to update the tag. This isn't automatically updated.

rwgk: Sorry the long comment here is incomplete (my fault), I forgot to add in:

To the best of my understanding: We are not free to update the `PyCQA/bandit-action` version here, but have to keep it pinned to a version sanctioned at the org level.

Long story (maybe skip over it)
-------------------------------
@kkraus14 set this up originally.

What happened after Keith set this up:

Dependabot PR #801 changed the hash/tag.

I saw failures in my CI runs, but they were not blocking. Maybe others didn't see the failures at all, but I'm not sure, and I never got a chance to get to the bottom of it.
 
My PR #911 undid PR #801 (among a couple unrelated tiny changes), which made the non-blocking CI failures I saw disappear.

My PR #913 added the `# v1.0.0` tag back, plus the long comment. Which is all true, but I realize only now, we need to add that we have to leave the version here pinned to what is sanctioned at the org level.

Upshot (I think)
----------------
* Undo the change in line 68 of .pre-commit-config.yaml
* Let's expand the comment to something like this:

```
        # Primary: KEEP THIS PINNED to a version sanctioned by the NVIDIA org.
        # Secondary: KEEP the bandit rev in .pre-commit-config.yaml IN SYNC
        # NOTE: Unfortunately the correct bandit rev is dependent on the GHA runner:
        # Current runner uses Python 3.8, so the action installs bandit==1.7.10
        # via `pip install bandit[sarif]`. If runner Python moves to >=3.9,
        # the action will resolve to 1.8.x and you'll need to bump pre-commit.
        # (Bandit >=1.8.0 dropped Python 3.8 via Requires-Python metadata.)
```

cpcloud: Does that mean we can't move to using ruff for this (it supports bandit linting)? It's much faster and would consolidate some configuration.

Is there some kind of page describing the history here of why we have to keep this pinned?

kkraus14: We can move to ruff. Third party GitHub actions are restricted at the NVIDIA org level where only certain revs are allowed. Ping me offline if you want more details.

.pre-commit-config.yaml
```diff
@@ -58,22 +58,22 @@ repos:
...
-    rev: "v1.10.0"
+    rev: "3a6eb0fadf60b3cccfd80bad9dbb6fae7e47b316"  # frozen: v1.10.0
     hooks:
     - id: rst-backticks
     - id: rst-directive-colons
     - id: rst-inline-touching-normal
 
   - repo: https://github.com/PyCQA/bandit
-    rev: "36fd65054fc8864b4037d0918904f9331512feb5"  # frozen: 1.7.10 KEEP IN SYNC WITH .github/workflows/bandit.yml
+    rev: "2d0b675b04c80ae42277e10500db06a0a37bae17"  # frozen: 1.8.6
```
kkraus14: Does this differ from the pre-commit version now?

cpcloud: Oh, it can be removed now, because we're moving this check to ruff.

cpcloud: Done.

kkraus14 (approved):

PR #1018, opened 2025-09-25
chore(pre-commit): add actionlint pre-commit-config

cpcloud:
Add `actionlint` pre-commit-config step.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,7 +4,7 @@
 
 name: "CI: Release"
 
-description: Manually-triggered release workflow. Creates a release draft if one doesn't exist for the given tag, or uses existing draft.
+# Manually-triggered release workflow. Creates a release draft if one doesn't exist for the given tag, or uses existing draft.
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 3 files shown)


cpcloud:
This would've caught my misspelling:

image

.github/workflows/release.yml
```diff
@@ -4,7 +4,7 @@
 
 name: "CI: Release"
 
-description: Manually-triggered release workflow. Creates a release draft if one doesn't exist for the given tag, or uses existing draft.
+# Manually-triggered release workflow. Creates a release draft if one doesn't exist for the given tag, or uses existing draft.
```
cpcloud, line 7: `description` is only a valid field in `action.yml`s, not workflows.

.pre-commit-config.yaml
```diff
@@ -80,5 +80,10 @@ repos:
         files: ^cuda_pathfinder/cuda/.*\.py$  # Exclude tests directory
         args: [--config-file=cuda_pathfinder/pyproject.toml]
 
+  - repo: https://github.com/rhysd/actionlint
+    rev: "03d0035246f3e81f36aed592ffb4bebf33a03106"
```
kkraus14: IIRC you need the `# frozen: x.x.x` comment otherwise pre-commit tries to update to a tag instead of a commit. The comment is apparently considered part of the format.

https://github.com/NVIDIA/cuda-python/pull/746#discussion_r2191053680

cpcloud: Ah ok. Thanks.

cpcloud:
@kkraus14 Are you comfortable merging as-is, or should we do a test run?

PR #1019, opened 2025-09-25
style: combine repeated ruff configuration to top-level module

cpcloud:
This PR moves our duplicated ruff configs to a top-level `ruff.toml` file. Commits are marked with `autofix` if the fixes were applied automatically.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import os
+
 import pytest
 
 
--- a/cuda_bindings/benchmarks/conftest.py
+++ b/cuda_bindings/benchmarks/conftest.py
@@ -3,7 +3,6 @@
 
 import numpy as np
 import pytest
-
 from cuda.bindings import driver as cuda
 from cuda.bindings import nvrtc
 from cuda.bindings import runtime as cudart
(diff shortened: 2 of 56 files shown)


ruff.toml
```diff
@@ -33,6 +33,7 @@ ignore = [
   "E741",  # ambiguous variable name such as I
   "B007",  # rename unused loop variable to _name
   "UP035", # UP006, UP007, UP035 complain about deprecated Typing.<type> use, but disregard backward compatibility of python version
+  "SIM108", # ternary if/else
```
rwgk, line 37: Hallelujah :-)

ruff.toml
```diff
@@ -33,6 +33,7 @@ ignore = [
   "E741",  # ambiguous variable name such as I
   "B007",  # rename unused loop variable to _name
   "UP035", # UP006, UP007, UP035 complain about deprecated Typing.<type> use, but disregard backward compatibility of python version
+  "UP006", # leave old-style typing as-is, e.g., Dict is allowed instead of requiring dict
```
rwgk: This one would be nice to get rid off in a follow-on PR. WDYT about creating an internal-cleanup issue, so we don't forget?

rwgk (approved):

cpcloud:
I guess we can do a test run for maximum confidence.

ruff.toml
cpcloud: I can tackle it after this is merged, I'm not sure we need an issue for it.

rwgk: Awesome.

PR #1020, opened 2025-09-25
IPC Mempool Serialization and multiprocessing Module Support

Andy-Jost:
Add serialization to memory IPC. Memory resources can be passed directly to new processes. Expands testing. Removes IPCChannel.


cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -414,70 +418,41 @@ cdef class IPCAllocationHandle:
...
         self.close()
 
+    def __reduce__(self):
+        multiprocessing.context.assert_spawning(self)
+        df = multiprocessing.reduction.DupFd(self.handle)
+        return IPCAllocationHandle._reconstruct, (df, self._uuid)
+
+    @staticmethod
+    def _reconstruct(df, uuid):
+        return IPCAllocationHandle._init(df.detach(), uuid)
```
cpcloud: This should probably be a class method.

```suggestion
    @classmethod
    def _reconstruct(cls, df, uuid):
        return cls._init(df.detach(), uuid)
```

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -414,70 +418,41 @@ cdef class IPCAllocationHandle:
...
+
     def __int__(self) -> int:
         if self._handle < 0:
             raise ValueError(
                 f"Cannot convert IPCAllocationHandle to int: the handle (id={id(self)}) is closed."
             )
         return self._handle
 
+    def detach(self):
+      handle = self._handle
```
cpcloud: The indentation doesn't match the rest of the file here.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -695,18 +711,16 @@ class DeviceMemoryResource(MemoryResource):
...
         self._mempool_owned = True
         self._is_imported = True
+        self._uuid = None
+        self._alloc_handle = None # only used for non-imported
 
         err, self._mempool_handle = driver.cuMemPoolImportFromShareableHandle(int(alloc_handle), _IPC_HANDLE_TYPE, 0)
         raise_if_driver_error(err)
-
+        if hasattr(alloc_handle, 'uuid'):
+            self.register(alloc_handle.uuid)
```
cpcloud:
```suggestion
        uuid = getattr(alloc_handle, "uuid", None)
        if uuid is not None:
            self.register(uuid)
```

This is the usual pattern for possibly-missing-attribute access in Python.

Then you're only accessing the attribute once.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -716,11 +730,20 @@ class DeviceMemoryResource(MemoryResource):
...
+            if self._is_imported:
+                raise RuntimeError("Imported memory resource cannot be exported")
+            err, alloc_handle = driver.cuMemPoolExportToShareableHandle(self._mempool_handle, _IPC_HANDLE_TYPE, 0)
+            raise_if_driver_error(err)
+            try:
+                assert self._uuid is None
+                self._uuid = uuid_module.uuid4()
+                self._alloc_handle = IPCAllocationHandle._init(alloc_handle, self._uuid)
+            except:
+                os.close(alloc_handle)
```
cpcloud, line 832: Do you actually intend to ignore **all** exceptions including those that inherit from `BaseException` (e.g., `KeyboardInterrupt` which is raised when the process receives a `SIGINT` signal)?

cuda_core/tests/memory_ipc/conftest.py
```diff
@@ -0,0 +1,33 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+
+from cuda.core.experimental import Device, DeviceMemoryResource
+
+POOL_SIZE = 2097152
+
+
+@pytest.fixture(scope="function")
```
cpcloud:
```suggestion
@pytest.fixture
```

This is the default behavior in pytest.

cuda_core/tests/memory_ipc/conftest.py
```diff
@@ -0,0 +1,33 @@
...
+
+    # Note: Linux specific. Once Windows support for IPC is implemented, this
+    # test should be updated.
+    if not device.properties.handle_type_posix_file_descriptor_supported:
+        pytest.skip("Device does not support IPC")
+
+    return device
+
+
+@pytest.fixture(scope="function")
```
cpcloud:
```suggestion
@pytest.fixture
```

This is the default behavior in pytest.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: This should reraise the exception.

cuda_core/tests/memory_ipc/test_workerpool.py
```diff
@@ -0,0 +1,103 @@
...
+    def init_worker(mrs):
+        global g_mrs
+        g_mrs = mrs
+
+    def test_ipc_workerpool(self, device, ipc_memory_resource):
+        """Test IPC with a worker pool."""
+        mr = ipc_memory_resource
+        buffers = [mr.allocate(NBYTES) for _ in range(NTASKS)]
+        with multiprocessing.Pool(processes=NWORKERS, initializer=self.init_worker, initargs=([mr],)) as pool:
+            pool.starmap(self.process_buffer, [(0, buffer.export()) for buffer in buffers])
```
rparolin: Consider: Adding a comment that map / starmap are blocking calls as a trace to the reader.

rparolin (approved):
lgtm.

cuda_core/tests/memory_ipc/test_workerpool.py
Andy-Jost: I'm leaning against this. To me it's clear these must be synchronous since they collect results.

cuda_core/cuda/core/experimental/__init__.py
leofang, line 1: Note: We forgot to expose `DeviceMemoryResourceOptions` in the previous PR.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -11,12 +11,17 @@ from cuda.core.experimental._utils.cuda_utils cimport (
...
 
 from dataclasses import dataclass
-from typing import TypeVar, Union, TYPE_CHECKING
+from typing import Optional, TypeVar, Union, TYPE_CHECKING
 import abc
 import array
 import cython
+import multiprocessing
+import multiprocessing.context
+import multiprocessing.reduction
```
leofang: Q: These seems to be private APIs? Are we allowed to use them and expect stability? Where can we find more docs?

cuda_core/cuda/core/experimental/_device.py
```diff
@@ -1160,6 +1160,16 @@ def __int__(self):
...
         return f"<Device {self._id} ({self.name})>"
 
+    def __reduce__(self):
+        return Device._reconstruct, (self.device_id,)
+
+    @staticmethod
+    def _reconstruct(device_id):
+        device = Device(device_id)
+        if not device._has_inited:
+            device.set_current()
```
leofang: I would prefer that we keep this explicit for now and avoid diverging behaviors between parent and children processes. Is it possible?
```suggestion
```

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: I don't think `multiprocessing.reduction.DupFd` is private since it is in the `__all__` list. I'm less sure about `get_spawning_popen` but I don't see an alternative.

cuda_core/cuda/core/experimental/_device.py
Andy-Jost: I don't think so. This only comes into play when spawning processes via something like `mp.Process(target=child_main, args=(buffer,))`. The resources are deserialized before entering `child_main`, so this is the only way I could find to automatically set up a device.

The simplified interfaces that "just work" out of the box rely on this. If someone wants to avoid this, they can use one of the other methods, such as passing an `alloc_handle` directly.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: By private I mean they are undocumented. I don't think CPython does an excellent codebase hygiene, so I wouldn't count on `__all__`.

It seems we need them only for
1. checking if the mp context is spawn
2. serializing the file descriptor

For 1 my concern is it is an extra safety check that we impose onto ourselves. For example, when using mpi4py I don't expect there's any spawning context, yet we still need the objects to be serializable. 

For 2 my confusion is `DupFd` does not seem to do anything special -- Can't we just make system calls ourselves?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -638,35 +622,67 @@ class DeviceMemoryResource(MemoryResource):
...
+            return DeviceMemoryResource.from_allocation_handle, (device, alloc_handle)
+        else:
+            return DeviceMemoryResource.from_registry, (self.uuid,)
+
+    @staticmethod
+    def from_registry(uuid: uuid_module.UUID):
+        try:
+            return _ipc_registry[uuid]
+        except KeyError:
+            raise RuntimeError(f"Memory resource {uuid} was not found")
```
```suggestion
            raise RuntimeError(f"Memory resource {uuid} was not found") from None
```

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -558,6 +533,12 @@ class DeviceMemoryResourceAttributes:
     del mempool_property
 
 
+# Holds DeviceMemoryResource objects imported by this process.
+# This enables buffer serialization, as buffers can reduce to a pair
+# of comprising the memory resource UUID (the key into this registry)
+# and the serialized buffer descriptor.
+_ipc_registry = {}
```
leofang, line 540: It seems there is no place where we'd purge items from this dict. I think it's an overlook? When the imported buffer is closed or goes out of scope, we should be able to do an inverse lookup and remove its MR from here, no? 

A growing internal cache could have potential performance impact.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -414,70 +418,41 @@ cdef class IPCAllocationHandle:
...
+        return cls._init(df.detach(), uuid)
+
     def __int__(self) -> int:
         if self._handle < 0:
             raise ValueError(
                 f"Cannot convert IPCAllocationHandle to int: the handle (id={id(self)}) is closed."
             )
         return self._handle
 
+    def detach(self):
```
leofang: if this is not a public API, let's prefix it with an underscore

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: The spawning check is key to how multiprocessing safely manages resources. If you look at something like a `multiprocessing.Queue`, it cannot ordinarily be pickled, but it can be passed as an argument to a spawning process, in which case it is pickled by `ForkingPickler`. This code basically follows what `multiprocessing` does for other objects.

`DupFd` is the way that handles are sent through `ForkingPickler`. When passing an object as an argument to a subprocess and having it reduce (partially) to this object, the file descriptor is sent to the new process. In this case, `ForkingPickler` takes care of everything for us. I don't see an easier way to hook into it.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -638,35 +622,67 @@ class DeviceMemoryResource(MemoryResource):
...
-        alloc_handle = channel._proxy._receive_allocation_handle()
-        return cls._from_allocation_handle(device_id, alloc_handle)
+    def __reduce__(self):
+        # If spawning a new process, serialize the resources; otherwise, just
+        # send the UUID, using the registry on the receiving end.
+        is_spawning = multiprocessing.context.get_spawning_popen() is not None
+        if is_spawning:
+            from ._device import Device
+            device = Device(self.device_id)
+            alloc_handle = self.get_allocation_handle()
```
leofang: Isn't `get_allocation_handle` already called in the constructor? Can we just use `self._alloc_handle` here?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -695,18 +711,17 @@ class DeviceMemoryResource(MemoryResource):
         self._ipc_handle_type = _IPC_HANDLE_TYPE
         self._mempool_owned = True
         self._is_imported = True
+        self._uuid = None
+        self._alloc_handle = None # only used for non-imported
```
leofang, line 800: You meant child processes do not need `self._alloc_handle`?

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: This can actually be removed.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -716,11 +731,21 @@ class DeviceMemoryResource(MemoryResource):
...
+                raise RuntimeError("Memory resource is not IPC-enabled")
+            if self._is_imported:
+                raise RuntimeError("Imported memory resource cannot be exported")
+            err, alloc_handle = driver.cuMemPoolExportToShareableHandle(self._mempool_handle, _IPC_HANDLE_TYPE, 0)
+            raise_if_driver_error(err)
+            try:
+                assert self._uuid is None
+                self._uuid = uuid_module.uuid4()
+                self._alloc_handle = IPCAllocationHandle._init(alloc_handle, self._uuid)
+            except:
```
leofang, line 831: Q: What would be the exception here?

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: For non-IPC-enabled resources, calling the function will trigger the appropriate error.

cuda_core/tests/memory_ipc/utility.py
```diff
@@ -0,0 +1,70 @@
...
+
+try:
+    from cuda.bindings import driver
+except ImportError:
+    from cuda import cuda as driver
+
+import ctypes
+
+from cuda.core.experimental import Buffer, MemoryResource
+from cuda.core.experimental._utils.cuda_utils import handle_return
```
leofang: btw I missed this during last review: 
```suggestion
import ctypes

from cuda.core.experimental import Buffer, MemoryResource
from cuda.core.experimental._utils.cuda_utils import driver, handle_return
```

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: Yes, sort of, but more specifically imported MRs do not use the alloc_handle. It is only for created (non-imported) MRs that are IPC-enabled.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: Anything uuid4 or IPCAllocationHandle._init might throw. In that case, the alloc_handle, which is a bare fd at this point, should not be leaked, and whatever that error is should propagate.

cuda_core/tests/memory_ipc/utility.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class DummyUnifiedMemoryResource(MemoryResource):
```
leofang, line 10: I also missed this: there's already a `DummyUnifiedMemoryResource` in `tests/test_memory.py`

Andy-Jost: Yes, I didn't want to try sharing.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: The more I read the more I feel this is over-engineered. 
- the existing `_send_allocation_handle` can be made as part of `__reduce__` for pickling in the parent process
- the existing `_receive_allocation_handle ` can be made as part of `_reconstruct` for unpickling in the child process

This way we don't need to rely on any internal implementation details of multiprocessing, just that our objects are serializable (which is a documented requirement in multiprocessing). Let's hop on a call on this.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: ok, thx for confirming, the inline comment is a bit confusing though.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: `__reduce__` reduces an object to a bytestream. `_send_allocation_handle` relies on a socket connection. Where does the socket come from in the context of `__reduce__`?

`DupFd` is a placeholder that `ForkingPickler` understands as a request to send a file descriptor to a new process. It owns the connection and handles the details.

cuda_core/tests/memory_ipc/conftest.py
```diff
@@ -0,0 +1,32 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+from cuda.core.experimental import Device, DeviceMemoryResource
```
leofang: Would be nice to get `DeviceMemoryResourceOptions` exercised
```suggestion
from cuda.core.experimental import Device, DeviceMemoryResource, DeviceMemoryResourceOptions
```

cuda_core/tests/memory_ipc/conftest.py
```diff
@@ -0,0 +1,32 @@
...
+    # test should be updated.
+    if not device.properties.handle_type_posix_file_descriptor_supported:
+        pytest.skip("Device does not support IPC")
+
+    return device
+
+
+@pytest.fixture
+def ipc_memory_resource(device):
+    mr = DeviceMemoryResource(device, dict(max_size=POOL_SIZE, ipc_enabled=True))
```
leofang: ditto
```suggestion
    mr = DeviceMemoryResource(device, DeviceMemoryResourceOptions(max_size=POOL_SIZE, ipc_enabled=True))
```

cuda_core/tests/memory_ipc/test_leaks.py
```diff
@@ -0,0 +1,134 @@
...
+            assert final_fds == self.initial_fds
+        return False
+
+
+prime_was_run = False
+
+
+def prime():
+    """Multiprocessing consumes a file descriptor on first launch."""
+    assert mp.get_start_method() == "spawn"
```
leofang, line 135: I think we can avoid setting spawn globally by using `mp.get_context("spawn")`

cuda_core/tests/memory_ipc/conftest.py
Andy-Jost: Does this accomplish anything? I think the implementation passes the dict to this object underneath, right?

cuda_core/tests/memory_ipc/test_leaks.py
Andy-Jost: That's true, but then we need to pass around context object (not a huge deal). I thought CUDA does not allow forks, so is there any reason not to set this globally? Spawn will be the default in a future Python BTW.

cuda_core/cuda/core/experimental/_device.py
```diff
@@ -1334,3 +1335,16 @@ def create_graph_builder(self) -> GraphBuilder:
...
+    device = Device(device_id)
+    if not device._has_inited:
+        device.set_current()
+    return device
+
+def _reduce_device(device):
+    return _reconstruct_device, (device.device_id,)
+
+multiprocessing.reduction.register(Device, _reduce_device)
+
```
Andy-Jost: This demonstrates an alternative to `__reduce__`. We can register a reduction method with `multiprocessing` to be used only when creating new processes. The advantage is the `Device` objects remain unpickleable, which might be the least surprising behavior. The cost is that `multiprocessing` needs to be imported, adding to the `cuda.core` import overhead. I could probably eliminate that with some restructuring, perhaps creating a `memory.reductions` module (or similar) that gets imported the first time IPC is used.

leofang: Seems promising, yeah would be nice to defer this registration if possible. Also as discussed offline sending an integer (dev id) over is very cheap. `Device` doesn't have to be serializable even in this use case.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -26,8 +28,9 @@ if platform.system() == "Linux":
     import socket
 
 if TYPE_CHECKING:
+    from ._device import Device
     import cuda.bindings.driver
```
leofang: nit: seems like an oversight 
```suggestion
    from cuda.core.experimental._utils.cuda_utils import driver
```

leofang:
btw while we're at this: The names `Buffer.import_` and `Buffer.export` are slightly nerve wrecking -- e.g. import from what? export to/as what? Having to ~~prefix~~ **suffix** an underscore in order to differentiate from the `import` keyword also causes annoying asymmetry and potential confusion (ex: in PyTorch methods ending with an underscore represent in-place operations). How about `Buffer.from_ipc_descriptor` and `Buffer.to_ipc_descriptor` instead?

cuda_core/tests/memory_ipc/conftest.py
leofang: 1. public APIs should be tested to ensure test coverage
2. since this is a public repo, people could peek into the test suite and learn how things are used and tested, and I don't want them to learn that passing a dict is allowed if possible. The fact that it's allowed today is only due to our choice of implementing the options using dataclasses, but we could walk away from that.

cuda_core/cuda/core/experimental/_device.py
leofang: > The resources are deserialized before entering `child_main`

This is horrifying...

cuda_core/cuda/core/experimental/_memory.pyx
leofang: Leaving a quick note to record offline discussions before calling a night.
- we don't want to create one socket pair per buffer, since this can be costly
- to be `multiprocessing`-aware, using the private APIs is the only way (plus reducing the socket creation cost to one pair per MR instead of per buffer)
- we acknowledge that the nice UX of `mp.Process(target=child_main, args=(buffer,))` is only possible for `multiprocessing`-backed IPC communications; if using mpi4py or other alternative IPC mechanisms, this UX is not possible (in general we don't want any `cuda.core` object to be picklable) but there are alternatives so it is not blocking anyone

cuda_core/cuda/core/experimental/_device.py
Andy-Jost: Can you elaborate? What problems might we encounter?

cuda_core/cuda/core/experimental/_device.py
Andy-Jost: I thought so, too, and tried that approach first, but this turned out to be more direct. If only the device ID integer is sent, then every object that makes CUDA API calls needs to check for initialization and possibly set up a device.  By sending a device as the first thing, we ensure CUDA is always initialized before any API calls are made.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: Yes, thanks. This line can be removed in fact.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: Adding to this discussion and along the lines of creating one socket pair per buffer, I created a change that sends a memory resource allocation handle with each buffer and removes all of the UUID and registry code. This should be more efficient than creating a socket pair but still involves importing a memory resource for each buffer. On my system, it increases the time to send buffers by 10x (0.012s versus 0.122s for 1000 buffers).

cuda_core/cuda/core/experimental/_device.py
leofang: This stdlib behavior can be confusing. Intuitively I'd expect the opposite order, that we drop into the user function first and then deserialize the objects sent by the parent. If the deserialization creates a global state, it might not be what the user would expect. Well, there's nothing we can do at this point, I guess.

leofang:
Seems like a Python 3.12 issue?

Andy-Jost:
> Seems like a Python 3.12 issue?

Python 3.12 + "local" CTK (but I'm not sure exactly what that means). The GPU HW and CTK version do not seem to matter.

leofang:
> but I'm not sure exactly what that means

it means if we are using CTK that we assembled to mimic a system CTK installation, or installed via pip

Andy-Jost:
It looks like this change is good to go. I'm not aware of any remaining issues.

leofang:
Awesome, thanks Andy! I'll re-review in an hour or two and then merge. I don't expect any change is needed, just wanna read the new tests once more.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -722,11 +740,22 @@ class DeviceMemoryResource(MemoryResource):
...
+            if not self.is_ipc_enabled:
+                raise RuntimeError("Memory resource is not IPC-enabled")
+            if self._is_imported:
+                raise RuntimeError("Imported memory resource cannot be exported")
+            err, alloc_handle = driver.cuMemPoolExportToShareableHandle(self._mempool_handle, _IPC_HANDLE_TYPE, 0)
+            raise_if_driver_error(err)
+            try:
+                assert self._uuid is None
+                import uuid as uuid
+                self._uuid = uuid.uuid4()
```
Andy-Jost, line 829: @ksimpson-work I ended up creating unique identifiers to track memory pools across processes. Is this something we could add to the Driver API as a mempool attribute?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -81,6 +84,9 @@ cdef class Buffer:
             self._mr = None
             self._ptr_obj = None
 
+    def __reduce__(self):
+        return Buffer.from_ipc_descriptor, (self.memory_resource, self.get_ipc_descriptor())
```
leofang, line 88: Q: I must have missed this during earlier review, but shouldn't we use `multiprocessing.reduction.register` instead of implementing `__reduce__` and always allowing `Buffer` to be serialized (in this case, when it's backed by an IPC-enabled MR)?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -384,33 +390,29 @@ cdef class IPCBufferDescriptor:
         return self
 
     def __reduce__(self):
-        # This is subject to change if the CUmemPoolPtrExportData struct/object changes.
-        return (self._reconstruct, (self._reserved, self._size))
+        return self._init, (self._reserved, self._size)
```
leofang, line 393: ditto, we should register instead of supplying `__reduce__`, we already did so for `IPCAllocationHandle` below, so let's be consistent and always register

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -644,35 +621,77 @@ class DeviceMemoryResource(MemoryResource):
...
+        if is_spawning:
+            from ._device import Device
+            device = Device(self.device_id)
+            alloc_handle = self.get_allocation_handle()
+            return DeviceMemoryResource.from_allocation_handle, (device, alloc_handle)
+        else:
+            return DeviceMemoryResource.from_registry, (self.uuid,)
+
+    @staticmethod
+    def from_registry(uuid: uuid.UUID):
```
leofang: 1. return type
```suggestion
    def from_registry(uuid: uuid.UUID) -> DeviceMemoryResource:
```
2. All public APIs should have a docstring

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -644,35 +621,77 @@ class DeviceMemoryResource(MemoryResource):
...
+        try:
+            return _ipc_registry[uuid]
+        except KeyError:
+            raise RuntimeError(f"Memory resource {uuid} was not found") from None
+
+    def register(self, uuid: uuid.UUID):
```
leofang: ditto, regarding return type & docstring

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -644,35 +621,77 @@ class DeviceMemoryResource(MemoryResource):
...
+        existing = _ipc_registry.get(uuid)
+        if existing is not None:
+            return existing
+        assert self._uuid is None or self._uuid == uuid
+        _ipc_registry[uuid] = self
+        self._uuid = uuid
+        return self
+
+    def unregister(self):
```
leofang, line 748: ditto, this is definitely a private API that's not meant to be called by the users

cuda_core/tests/memory_ipc/test_serialize.py
```diff
@@ -0,0 +1,181 @@
...
+
+from cuda.core.experimental import Buffer, Device, DeviceMemoryResource
+from utility import IPCBufferTestHelper
+
+CHILD_TIMEOUT_SEC = 20
+NBYTES = 64
+POOL_SIZE = 2097152
+
+
+class TestObjectSerializationDirect:
```
leofang, line 16: No action needed for getting this PR merged, but we should revisit. 

I feel uneasy that we use `multiprocessing.reduction` in the test code, which is potentially educating users how we intend APIs to be used. I am willing to bite the bullet and use CPython private APIs in the implementation, but I am not convinced that we have to teach users to do the same. 

In both Keenan's existing PR, as well as your previous PR, you already have a way to test this scenario. We can just launch the child process with one end of the socket, and use standard socket APIs to send the handle over. No need to call any private stdlib API.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -644,35 +621,77 @@ class DeviceMemoryResource(MemoryResource):
...
-            self._mempool_owned = False
-            self._is_imported = False
 
-    @classmethod
-    def from_shared_channel(cls, device_id: int | Device, channel: IPCChannel) -> DeviceMemoryResource:
-        """Create a device memory resource from a memory pool shared over an IPC channel."""
-        device_id = getattr(device_id, 'device_id', device_id)
-        alloc_handle = channel._proxy._receive_allocation_handle()
-        return cls._from_allocation_handle(device_id, alloc_handle)
+    def __reduce__(self):
```
leofang, line 712: ditto regarding `__reduce__` vs register, I wonder if we can just register the code block under `is_spawning`?

cuda_core/cuda/core/experimental/_memory.pyx
leofang: hey... maybe this is not a public API?! If so, let's prefix it with an underscore

cuda_core/cuda/core/experimental/_memory.pyx
leofang: Q: Can `register` be kept private?

cuda_core/tests/memory_ipc/test_errors.py
```diff
@@ -0,0 +1,139 @@
...
+        assert exc_msg == "Imported memory resource cannot be exported"
+
+
+class TestImportBuffer(ChildErrorHarness):
+    """Error when using a buffer as a buffer descriptor."""
+
+    def PARENT_ACTION(self, queue):
+        # Note: if the buffer is not attached to something to prolong its life,
+        # CUDA_ERROR_INVALID_CONTEXT is raised from Buffer.__del__
+        self.buffer = self.mr.allocate(NBYTES)
```
leofang, line 111: Q: Is this related to #1063?

cuda_core/tests/memory_ipc/test_errors.py
```diff
@@ -0,0 +1,139 @@
...
+    """
+    Error when importing a buffer object without registering its memory
+    resource.
+    """
+
+    def PARENT_ACTION(self, queue):
+        options = DeviceMemoryResourceOptions(max_size=POOL_SIZE, ipc_enabled=True)
+        mr2 = DeviceMemoryResource(self.device, options=options)
+        self.buffer = mr2.allocate(NBYTES)
+        queue.put(self.buffer)  # Note: mr2 not sent
```
leofang: I am a bit confused by this test: I thought we allow sending a single buffer over, and the register/serialization mechanism would ensure the MR is accessible on the child side? Maybe I get confused with another use case?

cuda_core/tests/memory_ipc/test_serialize.py
```diff
@@ -0,0 +1,181 @@
...
+    def child_main(self, conn):
+        # Set up the device.
+        device = Device()
+        device.set_current()
+
+        # Receive the memory resource.
+        handle = mp.reduction.recv_handle(conn)
+        uuid = conn.recv()
+        mr = DeviceMemoryResource.from_allocation_handle(device, handle)
+        mr.register(uuid)
```
leofang: To confirm my understanding: Is `register` here required because we need to import `buffer1` below?

cuda_core/tests/memory_ipc/test_workerpool.py
```diff
@@ -0,0 +1,105 @@
...
+    def test_ipc_workerpool(self, ipc_device, ipc_memory_resource):
+        """Test IPC with a worker pool."""
+        device = ipc_device
+        mr = ipc_memory_resource
+        buffers = [mr.allocate(NBYTES) for _ in range(NTASKS)]
+        with multiprocessing.Pool(processes=NWORKERS, initializer=self.init_worker, initargs=([mr],)) as pool:
+            pool.map(self.process_buffer, buffers)
+
+        for buffer in buffers:
+            IPCBufferTestHelper(device, buffer).verify_buffer(flipped=True)
```
leofang: I am staring the screen for a few mins, and I feel this test can be merged into the one below, by parametrizing  over `NMRS` (1 and 2). I think the test below supports both cases, and this one is redundant. WDYT?

cuda_core/tests/memory_ipc/test_workerpool.py
```diff
@@ -0,0 +1,105 @@
...
+    def test_ipc_workerpool(self, ipc_device, ipc_memory_resource):
+        """Test IPC with a worker pool."""
+        device = ipc_device
+        mr = ipc_memory_resource
+        buffers = [mr.allocate(NBYTES) for _ in range(NTASKS)]
+        with multiprocessing.Pool(processes=NWORKERS, initializer=self.init_worker, initargs=([mr],)) as pool:
+            pool.starmap(self.process_buffer, [(0, buffer.get_ipc_descriptor()) for buffer in buffers])
+
+        for buffer in buffers:
+            IPCBufferTestHelper(device, buffer).verify_buffer(flipped=True)
```
leofang: see below (or above) regarding parametrizing over `NMRS` and removing this test

cuda_core/cuda/core/experimental/_device.py
```diff
@@ -1334,3 +1335,17 @@ def create_graph_builder(self) -> GraphBuilder:
...
+
+def _reconstruct_device(device_id):
+    device = Device(device_id)
+    if not device._has_inited:
+        device.set_current()
+    return device
+
+
+multiprocessing.reduction.register(Device, _reduce_device)
```
leofang: Sorry I can't resist but circle back to ask this: We wanted `Device` to be serializable when only `Buffer` objects are passed over to child processes through `multiprocessing` APIs. We called `.set_current()` because the child function arguments are restored/deserialized prior to dropping into the function (xref: https://github.com/NVIDIA/cuda-python/pull/1020#discussion_r2392302541). 

Questions:
1. What happens in a multi-GPU environment? If I send over `buf0` allocated on GPU 0 and `buf1` allocated on GPU 1  to the child, which device is current by the time the child function starts executing? What if we change the order of args from `(buf0, buf1)` to `(buf1, buf0)`? We don't have CI for multi-GPU tests for now, but the design should be multi-GPU-safe.
2. If we really need a device ID, can't we just look up from `Buffer.mr.device_id`, instead of serializing `Device`?

I still feel it is unwise to set the global state behind users' back.

leofang (changes requested):
Sorry Andy, I re-read this again and still have a few questions 😓 

btw we should fix a few doc issues
- add `DeviceMemoryResourceOptions ` to `cuda_core/docs/source/api.rst`
- add `IPCAllocationHandle` to `cuda_core/docs/source/api_private.rst`
- address other comments below

The doc preview CI can be used to check what's missing.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: I opted to use `multiprocessing.reduction.register` only when `DupFd` is used. That's done to guarantee that `DupFd` is used only when the pickler can understand it.

Here, this implementation allows a buffer to be serialized apart from `multiprocessing`. For users relying on a different communication mechanism, this `__reduce__` will allow them to send buffers over any type of connection. This supports the following use case:

1. Send a memory resource via its allocation handle. It is up to the user how to open the fd in the remote process and register the memory resource (they don't need to rely on `multiprocessing` or `DupFd`).
2. Use this `Buffer.__reduce__` to convert a sharable buffer to a sequence of bytes.
3. Send the bytes by any mechanism (i.e., not relying on `multiprocessing`).
4. Reconstruct the buffer from its raw bytes. In the remote process, the memory resource is looked up by its UUID.

I don't love that some serialization methods are implemented by `multiprocessing.reduction.register` and others by `__reduce__`, however, I think this is the best compromise because it ensures `DupFd` is only used where safe while also not forcing users to rely on `multiprocessing`.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: Same comment as above. This supports sending buffers apart from `multiprocessing`.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: I lean towards keeping it public. We shouldn't assume anyone is using `multiprocessing`, and when they are not, they can use this lower-level API directly.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: I will fix return types and docstrings throughout. Same comment as above: these APIs should be public because they are not generally hidden. Only when using `multiprocessing` are they hidden.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: I think it may make sense to refactor this to make a dedicated registry singleton. That can happen after the release, though.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: Thanks for explanation/reminder, Andy. I now recall you've explained the idea offline to me.

For the benefit of reviewers/users/posterity, it'd be nice if you can expand the docstring to explain the number of ways of passing data between parent and children that we support for enabling IPC. Basically, a summary including your note above and the ways you tested in this PR.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: We try very hard to ensure all private APIs are prefixed with an underscore:
- Python helps ensure that they are not user visible (ex: with tab completion) 
- If we expose something now and hide it later, it is technically a breaking change. We should not break unless there is a strong reason.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: Do you mean the `DeviceMemoryResource` docstring?

leofang: The class docstring, yes, could be a good place. We unfortunately have not written enough docs. Ideally things we want to teach users should not live in the API reference, but it'd do the job for now...

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: Those are good points. Anyone not using `multiprocessing` might want the registry, so I've documented it in the next upload. I see how it puts us in a bind. Let's see how far I get today.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: OK, latest upload has (lengthy) documentation in the DMR docstring. +1 for writing longer docs eventually but this will have to do for now.

cuda_core/tests/memory_ipc/test_errors.py
Andy-Jost: Thanks for pointing this out. This led me to register `DeviceMemoryResource` with `multiprocessing` in the latest upload. Sorry for all of the back-and-forth in the implementation. The idea is that any serialization through `multiprocessing` should "just work," even if it incurs some overhead. With the quick exit I put into `from_allocation_handle` with the latest upload, I think the overhead should be much less than what I reported earlier. In most cases, only a few buffers should be mapped, so it won't matter. When it does matter (i.e., when a huge number of buffers are mapped), there is an out by using ordinary `pickle` methods. The latest update to this function demonstrates how to do it.

Andy-Jost: To answer your question more directly, your expectation is reasonable. So reasonable, in fact, that anything else would be too surprising. By registering the reduction method with `multiprocessing` it can be used in more contexts. Here, where a `multiprocessing.Queue` is involved, registering enables the behavior you expect.

The previous behavior enabled the "deep" transfer only when spawning a process, but now it is enabled any time `multiprocessing` facilitates the communication.

cuda_core/tests/memory_ipc/test_serialize.py
Andy-Jost: That's right. With the latest upload it is no longer required because the MR goes through a `mp.Pipe`, and I added a registration. I'll remove this line.

cuda_core/tests/memory_ipc/test_workerpool.py
Andy-Jost: These are merged now, and I added a third case to use `pickle` directly.

cuda_core/cuda/core/experimental/_device.py
Andy-Jost: The problem here is that I don't see a way to initialize CUDA (?) without calling `set_current` on a device. If I remove `set_current` from `_reconstruct_device` then I get CUDA_ERROR_INVALID_CONTEXT errors everywhere.

`Device._has_inited` is only set in `set_current`. To avoid messing with the global state, I'd rather say something like `Device(n).init()` to be sure that device n is ready to go. Then, later, when mapping a buffer, I could use `mr.device_id` to get that device. I just don't see any way to do that with the current `_device` code.

Andy-Jost: OK, I think I have something here that we can accept. I was able to remove the call to `set_current` in Device reconstruction since it turns out that memory resources can be mapped without setting a device context.

The only thing that really needs to be done implicitly when a process spawns is to call `cuInit`. The needs to be the very first thing or else all CUDA API calls (including mapping a memory resource or buffer) fails. That can be accomplished by just constructing a `Device`.

I tried removing device serialization completely but that actually makes the code quite a bit more cumbersome in several places. It's just easier to send an applicative of the form `(Device, (device_id,))` rather than send a device ID, import `Device`, and then reconstruct manually on the other end.  I also think it is conceptually sensible for `Device` objects to be serializable.

leofang: Sounds great, I'll take a look asap! A few quick notes before I forget:
1. CUDA mempools are by design not bound to any CUDA context. We (CUDA) used to have a context as the powerhouse for everything, memory, queue, module, ..., but now we are making things "context-independent," including mempools and kernel modules.
2. I read the `multiprocessing` docs more carefully, and I think we spawned the processes wrong (we should document this after this PR, in a Q&A or Tips & Tricks like in cuda-bindings). As you said calling `cuInit()` is the minimal:
   - For processes spawned via `Process`, we can inherit from the `Process` class, in the constructor `__init__()` we call `Device().set_current()` which would under the hood call `cuInit()`, and then call `super().__init__()`. My take is that this will ensure CUDA is initialized _before_ the arguments are deserialized
   - For processes spawned via a `Pool` of workers, the same can be done by passing `initializer` and `initargs`.

Prove me wrong! 🙂

Andy-Jost: I think subclassing `Process` as you describe would work. Are you suggesting we ship a subclass of `Process` to simplify this for our users? The current code places `set_current` calls within the child main functions. That's actually pretty nice, since it mirrors what we ask users to do in any process. As you point out, the memory pool and buffer mappings don't require a context, which simplifies this a lot.

When using worker pools, `initializer` and `initargs` do work (we have examples of this in `test_workpool.py`). Still, I think there's a ton of value in supporting the ultra-simple use case of mapping a workerpool over buffers without doing anything special at start up. The latest code just puts the call to `set_current` within each worker function, which seems great, actually. That way, one could mix-and-match buffers from different devices and it ought to work.

One limitation I'm facing now is that I don't know whether the test runners have multipe GPUs, so I'm not sure how to move forward with developing the multi-device scenario.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: To close the loop on this, in a perfect world, I'd like to create a class member `DeviceMemoryResource.registry` to hold a registry object and then attach management functions like `register` and `unregister` to that. There's probably not time in this release, though.

cuda_core/cuda/core/experimental/_device.py
leofang: > Are you suggesting we ship a subclass of `Process` to simplify this for our users?

Certainly not, at least not now (or never). We should only do so in our own test suite, and teach users about this. 

We did not have to teach this in most of Python GPU libraries (wearing my CuPy hat), because most libraries only interact with CUDA through runtime instead of driver, and runtime does the implicit initialization in every single cudart API. We don't and by design we always want users to call `dev.set_current()` first.

Now, with our own test suite, your other question is valid: `Device().set_current()` sets by default GPU 0 to current if there is no prior current device, this is a bit unfortunate if the child main wants to use say GPU 1. I think for now we need to call `set_current()` in both places, the process initializer (as a proxy to `cuInit()` because we definitely don't want to teach this) as well as child main (to actually select the right GPU, following the usual cuda.core requirement). We should do this for now and discuss if there is a better approach on Friday meeting. I believe cccl-rt will hit the same issue: https://github.com/NVIDIA/cccl/pull/6073.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: That is an interesting idea!!! Create an issue to make sure we don't lose track of this. Resolving.

cuda_core/cuda/core/experimental/_device.py
leofang: > Intuitively I'd expect the opposite order, that we drop into the user function first and then deserialize the objects sent by the parent. If the deserialization creates a global state, it might not be what the user would expect. Well, there's nothing we can do at this point, I guess.

For posterity: We found a solution to this, see https://github.com/NVIDIA/cuda-python/pull/1020#discussion_r2400438244. We actually _can_ do something.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -560,9 +562,63 @@ class DeviceMemoryResource(MemoryResource):
         When using an existing (current or default) memory pool, the returned
         device memory resource does not own the pool (`is_handle_owned` is
         `False`), and closing the resource has no effect.
+
+    IPC-Enabled Memory Resources
+    ----------------------------
```
leofang: The indentation is off so the doc is not rendering correctly:
https://nvidia.github.io/cuda-python/pr-preview/pr-1020/cuda-core/latest/generated/cuda.core.experimental.DeviceMemoryResource.html#cuda.core.experimental.DeviceMemoryResource

leofang: but otherwise the docstring looks great 👍 I wouldn't be as clear as what you did here, well done Andy.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -682,12 +746,18 @@ class DeviceMemoryResource(MemoryResource):
         return self
 
     def unregister(self):
-        if _ipc_registry is not None:
+        """Unregister this mapped memory resource."""
+        assert self.is_mapped
+        if _ipc_registry is not None:  # can occur during shutdown catastrophe
```
leofang, line 751: ❗

cuda_core/cuda/core/experimental/_memory.pyx
leofang: I am fixing this locally

leofang: It's not about indentation (the GH UI lied). I had forgotten that there are finite number of reserved keywords that can be used as section titles in docstrings.
https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#docstring-sections
I made a small tweak in commit 534b16a that should do the job. Well, we will hopefully re-work on the docs later...

cuda_core/tests/memory_ipc/test_errors.py
```diff
@@ -0,0 +1,143 @@
...
+        assert exc_type is RuntimeError
+        assert exc_msg == "Imported memory resource cannot be exported"
+
+
+class TestImportBuffer(ChildErrorHarness):
+    """Error when using a buffer as a buffer descriptor."""
+
+    def PARENT_ACTION(self, queue):
+        # Note: if the buffer is not attached to something to prolong its life,
+        # CUDA_ERROR_INVALID_CONTEXT is raised from Buffer.__del__
```
leofang, line 110: I suspect this comment is outdated.

leofang (approved):

PR #1021, opened 2025-09-25
ci: test free-threading builds with multiple threads

cpcloud:
This PR adds actual free-threaded test runs using `pytest-freethreaded` to the relevant CI jobs.


cpcloud:
Closing this out. I think we should wait for some usage reports before adding any specific testing strategies.

PR #1026, opened 2025-09-25
Removing pre-push doc

rparolin:


kkraus14 (approved):

kkraus14:
Fully closing the loop here, there's now a rule in place that prevents direct commits to key branches: https://github.com/NVIDIA/cuda-python/rules/8423851

PR #1032, opened 2025-09-26
Ensure allocation stream is used for buffer deallocation if no explicit stream is provided

### LINKED ISSUE #779: `Buffer` deallocation always happens on default stream

leofang:
## Description

closes #779.

## Checklist


leofang:
/ok to test [738eed4](https://github.com/NVIDIA/cuda-python/pull/1032/commits/738eed4229a6281c39b1642f905b776887176d3f)

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -84,10 +86,13 @@ cdef class Buffer:
             the behavior depends on the underlying memory resource.
         """
         if self._ptr and self._mr is not None:
+            if stream is None and self._alloc_stream is not None:
```
cpcloud:
```suggestion
            if stream is None:
                stream = self._alloc_stream
```

The `self._alloc_stream` check is superfluous: `stream`'s value remains the same with or without it.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -51,25 +51,27 @@ cdef class Buffer:
     """
 
     cdef:
-        uintptr_t _ptr
+        intptr_t _ptr
```
cpcloud: Did this change for any particular reason?

cuda_core/cuda/core/experimental/_memory.pyx
leofang: I have fixed the situation where `self._alloc_stream` could also be `None`.

leofang:
(This PR is now based entirely on #1070 except for the last commit, so that PR needs to be merged first.)

leofang:
/ok to test [605241e](https://github.com/NVIDIA/cuda-python/pull/1032/commits/605241e6877d31fa77d07785f6184af1766c9816)

cuda_core/cuda/core/experimental/_memory.pyx
rparolin: I have the same question.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -42,6 +42,29 @@ if TYPE_CHECKING:
...
+    #if defined(__GNUC__)
+      #pragma GCC diagnostic push
+      #pragma GCC diagnostic ignored "-Wint-to-pointer-cast"
+    #elif defined(_MSC_VER)
+      #pragma warning(push)
+      #pragma warning(disable : 4312)
+    #endif
+
+    void* unsafe_cast_from_int(int x) {
+        return (void*)x;
```
kkraus14: I think we can use `return (void*)(intptr_t)x;` to avoid the compilation warnings without having to use diagnostic pragmas?

Could probably do it inline in Cython below to avoid the need for this function.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: Discussed/answered in this thread: https://github.com/NVIDIA/cuda-python/pull/1070#discussion_r2408490288

cuda_core/cuda/core/experimental/_memory.pyx
leofang: Thanks, Keith, confirmed it's also working (and upcasting is safe).

kkraus14:
Changes LGTM on top of #1070

leofang:
Merged with main!

leofang:
/ok to test [33345d6](https://github.com/NVIDIA/cuda-python/pull/1032/commits/33345d6182f1e87ae7f6f21cdf427d2d24198783)

cuda_core/cuda/core/experimental/_memory.pyx
rparolin, line 939: Shouldn't we validate `stream` is not None here?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -931,10 +952,9 @@ cdef class DeviceMemoryResource(MemoryResource):
             The size of the buffer to deallocate, in bytes.
         stream : Stream, optional
             The stream on which to perform the deallocation asynchronously.
-            If None, an internal stream is used.
+            If the buffer is deallocated without an explicit stream, the allocation stream
+            is used.
         """
-        if stream is None:
```
rparolin, line 936: Did we intend to remove this? Based on the `deallocate` API signature a user would deem it valid to pass `None` as a stream argument.  That `None` is not checked in the `_deallocate` function above when the stream object is called with it.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -1030,12 +1052,10 @@ class LegacyPinnedMemoryResource(MemoryResource):
...
             The size of the buffer to deallocate, in bytes.
-        stream : Stream, optional
-            The stream on which to perform the deallocation asynchronously.
-            If None, no synchronization would happen.
+        stream : Stream
+            The stream on which to perform the deallocation synchronously.
         """
-        if stream:
-            stream.sync()
+        stream.sync()
```
rparolin, line 1058: Ditto here. We remove the `None` check but nothing about the signature change prevents a user from passing `None` as a stream object.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: This is our internal fast path and the caller is meant to check if it's `None` before calling it.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: This is a very interesting question that's worth expanding.. From my perspective `mr.deallocate()` is not meant to be called by end users because
1. it was designed to be called by `Buffer.close()` or its destructor when out of scope
2. it has such an ugly, not pythonic signature (passing pointer and size)

I don't think currently we have any test that actually exercise calling `mr.deallocate()` explicitly. 

If this can be called by the end users, the contract that we deallocate using the allocation stream would not be upheld, because this API does not take any `Buffer` argument and so the stream is not available to us.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: resolving this to keep discussion in the above thread: https://github.com/NVIDIA/cuda-python/pull/1032/files#r2412010938

rparolin (approved):

leofang:
Thanks, all! I noticed I forgot to prepare a release note for this PR. Working on it now...

leofang:
> Working on it now...

#1097

Issue #846, opened 2025-08-18
Support Python 3.14 & Drop 3.9


leofang:

cpcloud:
Let's wait until 3.14 is released before doing the work.

cpcloud:
Since we have 3.14rc builds, and 3.9 is on the way out, I'm going to PR the 3.9 drop, and we can merge it after the upcoming release.

leofang:
yes I've added a FutureWarning raised at import time to cuda.core PY39 users in #1041, we should do the same in cuda-bindings too

leofang:
Need to repeat https://github.com/NVIDIA/cuda-python/pull/1069 and drop 3.9 in the next minor releases.

cpcloud:
Closed by #1202.

Issue #1030, opened 2025-09-26
CI: Add integration tests to cover `cuda.bindings` ABI stability


leofang:
> We should add a CI job that tests running newer `cuda.bindings` + projects built against older `cuda.bindings` (from the same major release) to catch this in the future, either on a per-PR or nightly (#294) basis.
> 

 _Originally posted by @leofang in [#274](https://github.com/NVIDIA/cuda-python/issues/274#issuecomment-2824637343)_

mdboom:
A simpler approach might be to use `abigail` to detect ABI changes.  I looked at this a few weeks ago, and I think I was seeing false positives from it (changes in C ABI that don't matter in practice for Cython).  I plan to have another go at it and run over a chain of existing commits to see if it can detect legitimate ABI breakage.  If that works, that would be easier than integrating across multiple projects.

What are the intended expectations of ABI stability for CUDA bindings?  No breakage across a major version?

leofang:
> changes in C ABI that don't matter in practice for Cython

RIght. This is the unusual part of this task. We're not dealing with C ABIs (for which we could use various tools, including those from the ABI Lab), but Cython ABIs. However it is unclear what it entails for symbols that are `cimport`'d. It is never properly documented and the upstream does not want to document it either (related: https://github.com/cython/cython/issues/3077, https://github.com/cython/cython/issues/6484). 

In the past we've had Cython ABI-breaking changes at least twice, in both 11.x (https://github.com/conda-forge/cuda-python-feedstock/issues/15, @kkraus14 found this one) and 12.x (https://github.com/NVIDIA/cuda-python/issues/274), and every time it happened it caused a lot of churns. We'd really prefer to avoid another break if possible.

> No breakage across a major version?

Yes. Projects built against cuda-bindings 12.x (via `cimport`) should work with all 12.y, for arbitrary x <= y (x > y should also be allowed if no new symbol from 12.y is used).

mdboom:
I ran an experiment using `abigail` tools (the same tool CPython uses to check ABI compatibility).  It's definitely doable, but I've pretty much convinced myself it wouldn't add much value in our case to be worth it.  Cython modules really only have 2 exported symbols at the C ABI level, and they basically never change, and changing that it really under Cython's control anyway, not ours.  Anyway [here's the experiment for reference](https://github.com/NVIDIA/cuda-python/pull/1059).

```
  <elf-function-symbols>
[...]

Issue #1025, opened 2025-09-25
CI: Enforce label/milestone assignment


leofang:
We want to avoid the manual process of someone (me) tagging all issues and PRs. For issues I don't have a good way, but for PRs we can add a pre-merge check to ensure a PR has a milestone and a label (or labels) attached.

PR #994, opened 2025-09-19
ci: move cibuildwheel config out of github actions workflows and into pyproject.tomls

cpcloud:
Moves cibuildwheel config out of github actions and into corresponding pyproject.toml files.

PR #1041, opened 2025-09-28
Cythonize `cuda.core`

### LINKED ISSUE #866: RFC: Cythonize `cuda.core` while keeping it CUDA-agnostic
Today the majority of `cuda.core` is implemented in pure Python. As a result, we've been dealing with microsecond-level overhead in the past few months if not weeks (ex: https://github.com/NVIDIA/cuda-python/issues/739, https://github.com/NVIDIA/cuda-python/issues/658). As much as I think it is pre-mature optimization at this stage, I do hear the desire of keeping the performance competitive while staying productive. 

This RFC outlines one such solution to address the performance concerns. Below are the critical requirements
1. `cuda.core` continues to support multiple CUDA major versions
2. The installation UX `pip install cuda-core` stays unchanged
3. Local development workflow is uninterrupted 
4. No user- or developer- visible breaking change is introduced (despite we're still in the experimental phase)

The critical question to answer is how we'll lower to Cython while having to build against `cuda.bindings` 12x & 13.x. Here is the steps following the great work @dalcinl did for [mpi4py v4.1.0](https://mpi4py.readthedocs.io/en/stable/changes.html#release-4-1-0-2025-06-25) (to support both Open MPI and MPICH):
1. We turn all Python modules from `.py` to `.pyx` and update the build system 
   - We could consider having one mega `.pyx` with others being `.pyi` and literal-included, similar to the `mpi4py.MPI` module 
3. We build `cuda-core` twice, once against CUDA & `cuda-bindings` 12.x, and then 13.x
4. We merge two generated wheels into a single one ([script](https://github.com/mpi4py/mpi4py/blob/master/cibw/wheel-build.sh))   
   - A runtime dispatching snippet should be injected into `cuda/core/experimental/__init__.py` to decide which extension module to load, based on the installed `cuda-bindings` major version

It is worth noting that the Step 2 and 3 only happen in the public CI, so as to meet the Requirement 3 (for local development, neither internal nor external developers should need to have multiple CUDA versions installed).

Another note is that this RFC is only applicable to make our Python wheels stay **variant-free** (no -cu12/-cu13); for conda packages, it is trivial to build variant packages without changing the UX (`conda install cuda-core`), so no extra work is needed.

This RFC also mirrors our plan for `cuda-cccl` (https://github.com/NVIDIA/cccl/issues/2555).

leofang:
## Description

closes #866 

related to #846.

This PR implements the infrastructure so that all `cuda.core` modules can be properly cythonized and that accessing the `cuda.bindings` Cython layer (`from cuda.bindings.cyXXXXX cimport ...`) is safe. For local development, nothing is changed except for needing to specify `CUDA_PATH`, same as when building `cuda.bindings`.

The changes involve:
- Build system: We implement a custom build backend that defers cythonization, so that we can detect the CUDA version and install the correct `cuda-bindings` version to the build-isolation environment, prior to firing up the Cython compiler.
- Internal implementation: Many calls to Python bindings of driver APIs are now lowered to their Cython counterparts, though the change is not made exhaustively and can be followed up in future PRs. The selection of the `cu12`/`cu13` extension modules is done at the import time.
- CI: We build `cuda.core` wheels against two CUDA major versions, and use a script to merge the payloads. Only the final wheel is stored and used in the test stage (and can be released as before using the release workflow); the `cu12`/`cu13` wheels are thrown away after the merge.
   - `PIP_FIND_LINKS` is the key to allow building against the local, unreleased `cuda.bindings` wheels

## Checklist


leofang:
/ok to test [277a5b7](https://github.com/NVIDIA/cuda-python/pull/1041/commits/277a5b7c88b009ac1215465fbb215ee224ceacde)

leofang:
/ok to test [0b945cb](https://github.com/NVIDIA/cuda-python/pull/1041/commits/0b945cb5e3a256db4402afed58cd59c09cac6915)

leofang:
/ok to test [8a636e5](https://github.com/NVIDIA/cuda-python/pull/1041/commits/8a636e59f4354a9ba173ccf094afa8223907edf0)

leofang:
/ok to test [fbe9325](https://github.com/NVIDIA/cuda-python/pull/1041/commits/fbe9325e78fee2b56a9ab29b4c539553c9d393f7)

leofang:
/ok to test [d3c7877](https://github.com/NVIDIA/cuda-python/pull/1041/commits/d3c7877a92450a4f08f421c903a7d85367dbba7b)

leofang:
/ok to test [a59e7c0](https://github.com/NVIDIA/cuda-python/pull/1041/commits/a59e7c0bc2cd246dea666c0848d7a93174a05643)

leofang:
/ok to test [37fe7eb](https://github.com/NVIDIA/cuda-python/pull/1041/commits/37fe7eb541465da978f14798dd53c452caf5fe82)

leofang:
/ok to test [9e799e4](https://github.com/NVIDIA/cuda-python/pull/1041/commits/9e799e40785eac42e9bd9ed63e0ef331445b5622)

leofang:
/ok to test [d95cd91](https://github.com/NVIDIA/cuda-python/pull/1041/commits/d95cd914c221785518134fb981b9bbaf5d8c5891)

leofang:
/ok to test [93ac663](https://github.com/NVIDIA/cuda-python/pull/1041/commits/93ac6637cf316f75571cd5524f89edb613dcd0e8)

leofang:
/ok to test [d5001d4](https://github.com/NVIDIA/cuda-python/pull/1041/commits/d5001d4f8a4ef3eb9db2dcd238fda4a037100b77)

leofang:
Blocked by #1042.

leofang:
/ok to test [96ce480](https://github.com/NVIDIA/cuda-python/pull/1041/commits/96ce48055c5175047077f1e49cde00b232d4d106)

leofang:
OK I think this should be ready for the first review pass...
- ~~TODO: PY313t/314/314t are currently disabled; once #1042 is merged we should revert commit 1180ab6 to re-enable them~~
- ~~TODO: I would like to also cythonize the `Device` class, but it might be a stretching goal depending on my availability before code freeze~~
- ~~TODO: add PR description to explain how this works (see the linked issue and the links therein for now if you're interested)~~
- We can cythonize the remaining modules and classes in a separate PR

@shwina @emcastillo could you take a look?

emcastillo:
> I would like to also cythonize the Device class, but it might be a stretching goal depending on my availability before code freeze

I can try to work on it this week to take some stuff out of your hands.

leofang:
/ok to test [e702b5e](https://github.com/NVIDIA/cuda-python/pull/1041/commits/e702b5ef328819336affb0221b10cd480e73e9b2)

.github/workflows/build-wheel.yml
```diff
@@ -148,6 +124,15 @@ jobs:
         with:
           package-dir: ./cuda_bindings/
           output-dir: ${{ env.CUDA_BINDINGS_ARTIFACTS_DIR }}
+        env:
+          CIBW_BUILD: ${{ env.CIBW_BUILD }}
+          # CIBW mounts the host filesystem under /host
+          CIBW_ENVIRONMENT_LINUX: >
```
cpcloud, line 130: These are now configured in each project's respective `pyproject.toml`. Did you bring these back for a reason?

leofang:
/ok to test [566f966](https://github.com/NVIDIA/cuda-python/pull/1041/commits/566f9664273b50bb3005461c260ad10acddebd02)

leofang:
/ok to test [dc8d076](https://github.com/NVIDIA/cuda-python/pull/1041/commits/dc8d076ffa7a76524aa25d9507ee0a076f218643)

leofang:
We need a new rule: On the week of code freeze, do not alter the linter config. It is annoying 🤬

leofang:
/ok to test [04d3f5d](https://github.com/NVIDIA/cuda-python/pull/1041/commits/04d3f5dc58c88a832a79477541d9ec8ac652841b)

leofang:
/ok to test [0e6f927](https://github.com/NVIDIA/cuda-python/pull/1041/commits/0e6f9278739d75c2f62495a6770c4cc2c2395804)

leofang:
> OK I think this should be ready for the first review pass...

All TODOs are done. This PR is ready for review.

.github/workflows/build-wheel.yml
leofang: @cpcloud Yes. I did not get a chance to review #994. I have partially reverted the changes in this PR. For static properties it is appropriate to hard-code them in `pyproject.toml`, but for dynamic properties generated in the CI, it is very confusing to do so. In particular,
1. we hard-code the CI-specific information such as the CTK path `/host/...` and `cygwin ...`, which are useless outside of the CI
2. with this PR we need to alter the env vars in different job steps, making the variable-passing flow even harder to trace if they were not passed explicitly as arguments when invoking the cibuildwheel GHA

ci/tools/merge_cuda_core_wheels.py
```diff
@@ -0,0 +1,192 @@
...
+
+    # Convert wheel paths to Path objects and validate
+    wheels = []
+    for wheel_path in args.wheels:
+        wheel = Path(wheel_path)
+        if not wheel.exists():
+            print(f"Error: Wheel not found: {wheel}")
+            sys.exit(1)
+        if not wheel.name.endswith(".whl"):
+            print(f"Error: Not a wheel file: {wheel}")
```
cpcloud: This should print to `stderr`.


```suggestion
            print(f"Error: Not a wheel file: {wheel}", file=sys.stderr)
```

ci/tools/merge_cuda_core_wheels.py
```diff
@@ -0,0 +1,192 @@
...
+            sys.exit(1)
+        wheels.append(wheel)
+
+    if not wheels:
+        print("Error: No wheels provided")
```
cpcloud:
```suggestion
        print("Error: No wheels provided", file=sys.stderr)
```

ci/tools/merge_cuda_core_wheels.py
```diff
@@ -0,0 +1,192 @@
...
+    # Check that we have wheel tool available
+    try:
+        run_command(["python", "-m", "wheel", "--help"])
+    except Exception:
+        print("Error: wheel package not available. Install with: pip install wheel")
+        sys.exit(1)
+
+    # Merge the wheels
+    merged_wheel = merge_wheels(wheels, output_dir)
+    print(f"\nMerge complete! Output: {merged_wheel}")
```
cpcloud, line 186: This allows consumption of the output of this script by other tools.

```suggestion
    print(merged_wheel)
```

ci/tools/merge_cuda_core_wheels.py
```diff
@@ -0,0 +1,192 @@
...
+        print("Error: No wheels provided")
+        sys.exit(1)
+
+    output_dir = Path(args.output_dir)
+
+    # Check that we have wheel tool available
+    try:
+        run_command(["python", "-m", "wheel", "--help"])
+    except Exception:
+        print("Error: wheel package not available. Install with: pip install wheel")
```
cpcloud: Why not just import it? Also, the error should go to stderr, not stdout.

```suggestion
    try:
        import wheel
    except ImportError:
        print("Error: wheel package not available. Install with: pip install wheel", file=sys.stderr)
```

ci/tools/merge_cuda_core_wheels.py
```diff
@@ -0,0 +1,192 @@
...
+
+    print("cuda.core Wheel Merger")
+    print("======================")
+
+    # Convert wheel paths to Path objects and validate
+    wheels = []
+    for wheel_path in args.wheels:
+        wheel = Path(wheel_path)
+        if not wheel.exists():
+            print(f"Error: Wheel not found: {wheel}")
```
cpcloud:
```suggestion
            print(f"Error: Wheel not found: {wheel}", file=sys.stderr)
```

ci/tools/merge_cuda_core_wheels.py
```diff
@@ -0,0 +1,192 @@
...
+def main():
+    """Main merge script."""
+    parser = argparse.ArgumentParser(description="Merge CUDA-specific wheels into a single multi-CUDA wheel")
+    parser.add_argument("wheels", nargs="+", help="Paths to the CUDA-specific wheels to merge")
+    parser.add_argument("--output-dir", "-o", default="dist", help="Output directory for merged wheel")
+
+    args = parser.parse_args()
+
+    print("cuda.core Wheel Merger")
+    print("======================")
```
cpcloud: All output that isn't processable by other downstream text processing tools should go stderr.

```suggestion
    print("cuda.core Wheel Merger", file=sys.stderr)
    print("======================", file=sys.stderr)
```

ci/tools/merge_cuda_core_wheels.py
```diff
@@ -0,0 +1,192 @@
...
+            ]
+        )
+
+        # Find the output wheel
+        output_wheels = list(output_dir.glob("*.whl"))
+        if not output_wheels:
+            raise RuntimeError("Failed to create merged wheel")
+
+        merged_wheel = output_wheels[0]
+        print(f"Successfully merged wheel: {merged_wheel}")
```
cpcloud:
```suggestion
        print(f"Successfully merged wheel: {merged_wheel}", file=sys.stderr)
```

ci/tools/merge_cuda_core_wheels.py
```diff
@@ -0,0 +1,192 @@
...
+        # Repack the merged wheel
+        output_dir.mkdir(parents=True, exist_ok=True)
+
+        # Create a clean wheel name without CUDA version suffixes
+        base_wheel_name = wheels[0].name
+        # Remove any .cu* suffix from the wheel name
+        if ".cu" in base_wheel_name:
+            base_wheel_name = base_wheel_name.split(".cu")[0] + ".whl"
+
+        print(f"Repacking merged wheel as: {base_wheel_name}")
```
cpcloud:
```suggestion
        print(f"Repacking merged wheel as: {base_wheel_name}", file=sys.stderr)
```

ci/tools/merge_cuda_core_wheels.py
```diff
@@ -0,0 +1,192 @@
...
+                    os.remove(f_abspath)
+
+        # Repack the merged wheel
+        output_dir.mkdir(parents=True, exist_ok=True)
+
+        # Create a clean wheel name without CUDA version suffixes
+        base_wheel_name = wheels[0].name
+        # Remove any .cu* suffix from the wheel name
+        if ".cu" in base_wheel_name:
+            base_wheel_name = base_wheel_name.split(".cu")[0] + ".whl"
```
cpcloud: There's a much simpler way:


```suggestion
        base_wheel_name = wheels[0].with_suffix(".whl").name
```

E.g.,:

```
In [6]: Path("foo.cu").with_suffix(".whl")
Out[6]: PosixPath('foo.whl')
```

ci/tools/merge_cuda_core_wheels.py
```diff
@@ -0,0 +1,192 @@
...
+
+        # The base dir should only contain __init__.py, the include dir, and the versioned dirs
+        files_to_remove = os.listdir(base_wheel / base_dir)
+        for f in files_to_remove:
+            f_abspath = base_wheel / base_dir / f
+            if f not in ("__init__.py", "cu12", "cu13", "include"):
+                if os.path.isdir(f_abspath):
+                    shutil.rmtree(f_abspath)
+                else:
+                    os.remove(f_abspath)
```
cpcloud: `os.scandir` is faster

```suggestion
        files_to_remove = os.scandir(base_wheel / base_dir)
        for f in files_to_remove:
            f_abspath = f.path
            if f.name not in ("__init__.py", "cu12", "cu13", "include"):
                if f.is_dir():
                    shutil.rmtree(f_abspath)
                else:
                    os.remove(f_abspath)
```

ci/tools/merge_cuda_core_wheels.py
```diff
@@ -0,0 +1,192 @@
...
+        # Use the first wheel as the base and merge binaries from others
+        base_wheel = extracted_wheels[0]
+
+        # now copy the version-specific directory from other wheels
+        # into the appropriate place in the base wheel
+        for i, wheel_dir in enumerate(extracted_wheels):
+            cuda_version = wheels[i].name.split(".cu")[1].split(".")[0]
+            base_dir = Path("cuda") / "core" / "experimental"
+            # Copy from other wheels
+            print(f"  Copying {wheel_dir} to {base_wheel}")
```
cpcloud:
```suggestion
            print(f"  Copying {wheel_dir} to {base_wheel}", file=sys.stderr)
```

ci/tools/merge_cuda_core_wheels.py
```diff
@@ -0,0 +1,192 @@
...
+            shutil.copytree(wheel_dir / base_dir, base_wheel / base_dir / f"cu{cuda_version}")
+
+            # Overwrite the __init__.py in versioned dirs
+            open(base_wheel / base_dir / f"cu{cuda_version}" / "__init__.py", "w").close()
```
cpcloud:
```suggestion
            os.truncate(base_wheel / base_dir / f"cu{cuda_version}" / "__init__.py", 0)
```

cuda_core/cuda/core/experimental/_device.pyx
```diff
@@ -1078,20 +1078,24 @@ def uuid(self) -> str:
...
-        else:
-            uuid = handle_return(driver.cuDeviceGetUuid(self._id))
-        uuid = uuid.bytes.hex()
+        cdef cydriver.CUuuid uuid
+        IF CUDA_CORE_BUILD_MAJOR == "12":
+            HANDLE_RETURN(cydriver.cuDeviceGetUuid_v2(&uuid, self._id))
+        ELSE:  # 13.0+
+            HANDLE_RETURN(cydriver.cuDeviceGetUuid(&uuid, self._id))
+        cdef bytearray uuid_b = bytearray(sizeof(uuid.bytes))
+        uuid_b[:] = uuid.bytes
```
leofang: For some reason Cython does not generate the correct code here... 🤔 I'd expect it to use `PyBytes_FromStringAndSize` instead of `PyBytes_FromString`...
```c++
  /* "cuda/core/experimental/_device.pyx":1088
 *             HANDLE_RETURN(cydriver.cuDeviceGetUuid(&uuid, self._id))
 *         cdef bytearray uuid_b = bytearray(sizeof(uuid.bytes))
 *         uuid_b[:] = uuid.bytes             # <<<<<<<<<<<<<<
 *         cdef str uuid_hex = uuid_b.hex()
 *         # 8-4-4-4-12
*/
  __pyx_t_1 = __Pyx_PyBytes_FromString(__pyx_v_uuid.bytes); if (unlikely(!__pyx_t_1)) __PYX_ERR(0, 1088, __pyx_L1_error)
  __Pyx_GOTREF(__pyx_t_1);
  if (__Pyx_PyObject_SetSlice(__pyx_v_uuid_b, __pyx_t_1, 0, 0, NULL, NULL, NULL, 0, 0, 1) < 0) __PYX_ERR(0, 1088, __pyx_L1_error)
```

ci/tools/merge_cuda_core_wheels.py
```diff
@@ -0,0 +1,192 @@
...
+    if len(wheels) == 1:
+        raise RuntimeError("only one wheel is provided, nothing to merge")
+
+    # Extract all wheels to temporary directories
+    with tempfile.TemporaryDirectory() as temp_dir:
+        temp_path = Path(temp_dir)
+        extracted_wheels = []
+
+        for i, wheel in enumerate(wheels):
+            print(f"Extracting wheel {i + 1}/{len(wheels)}: {wheel.name}")
```
cpcloud:
```suggestion
            print(f"Extracting wheel {i + 1}/{len(wheels)}: {wheel.name}", file=sys.stderr)
```

ci/tools/merge_cuda_core_wheels.py
```diff
@@ -0,0 +1,192 @@
...
+        print("STDERR:", result.stderr)
+        result.check_returncode()
+
+    return result
+
+
+def merge_wheels(wheels: List[Path], output_dir: Path) -> Path:
+    """Merge multiple wheels into a single wheel with version-specific binaries."""
+    print("\n=== Merging wheels ===")
+    print(f"Input wheels: {[w.name for w in wheels]}")
```
cpcloud:
```suggestion
    print("\n=== Merging wheels ===", file=sys.stderr)
    print(f"Input wheels: {[w.name for w in wheels]}", file=sys.stderr)
```

ci/tools/merge_cuda_core_wheels.py
```diff
@@ -0,0 +1,192 @@
...
+
+    result = subprocess.run(cmd, cwd=cwd, env=env, capture_output=True, text=True)  # noqa: S603
+
+    if result.returncode != 0:
+        print(f"Command failed with return code {result.returncode}")
+        print("STDOUT:", result.stdout)
+        print("STDERR:", result.stderr)
+        result.check_returncode()
+
+    return result
```
cpcloud, line 46: What is this function doing over and above running the command with `subprocess.run(check=True)`?

You're printing stdout and stderr (not to stderr) and then checking the return code. That's what `run(check=True)` does.

This function just seems to add more maintenance work instead of directly using the existing functionality of the standard library.

ci/tools/merge_cuda_core_wheels.py
```diff
@@ -0,0 +1,192 @@
...
+    with tempfile.TemporaryDirectory() as temp_dir:
+        temp_path = Path(temp_dir)
+        extracted_wheels = []
+
+        for i, wheel in enumerate(wheels):
+            print(f"Extracting wheel {i + 1}/{len(wheels)}: {wheel.name}")
+            # Extract wheel - wheel unpack creates the directory itself
+            run_command(
+                [
+                    "python",
```
cpcloud:
```suggestion
                    sys.executable,
```
But also, isn't this functionality available as a module?

ci/tools/merge_cuda_core_wheels.py
```diff
@@ -0,0 +1,192 @@
...
+        # Create a clean wheel name without CUDA version suffixes
+        base_wheel_name = wheels[0].name
+        # Remove any .cu* suffix from the wheel name
+        if ".cu" in base_wheel_name:
+            base_wheel_name = base_wheel_name.split(".cu")[0] + ".whl"
+
+        print(f"Repacking merged wheel as: {base_wheel_name}")
+        run_command(
+            [
+                "python",
```
cpcloud:
```suggestion
                sys.executable,
```

cuda_core/cuda/core/experimental/_device.pyx
leofang: fixed in commit 0e28aa19

.github/workflows/build-wheel.yml
cpcloud: > with this PR we need to alter the env vars in different job steps

We're already handling all of these in pyproject.toml. There's nothing here that is dynamic beyond the capabilities of the pyproject.toml configuration.

I'm not sure how this merits special casing and duplicating all of these variables.

leofang: I find it extremely confusing when working on CI workflow. I have a strong opinion on how we pass the cibuildwheel variables 🙂

leofang:
For `ci/tools/merge_cuda_core_wheels.py` unless there is a serious bug, I prefer that we fix it in a separate PR. Right now this file is almost a copy-pasta from https://github.com/NVIDIA/cccl/blob/main/python/cuda_cccl/merge_cuda_wheels.py with very minimal changes for merging the `cuda.core` wheels (there are a few places that are hard-coded for `cuda.cccl.parallel`), and while I agree with all review comments I prefer to keep it as close as possible with the source file.

.github/workflows/build-wheel.yml
cpcloud: The current pyproject.toml scheme is working and reduced the amount of noise in our already complicated CI. There's really no reason to duplicate this config again.

cpcloud:
Why are we duplicating so much code without it going through review?

cpcloud:
> We need a new rule: On the week of code freeze, do not alter the linter config. It is annoying 🤬

Perhaps something to enforce the freeze is in order then. What's the point of a freeze if nothing is actually frozen?

.github/workflows/build-wheel.yml
leofang: This is not about duplication. This is about parameters not passed explicitly, and instead accessed as if they are global variables. I have a hard time tracking global variables this way. This is another place where I am very comfortable breaking DRY.

ci/tools/merge_cuda_core_wheels.py
leofang: Your suggestion seem to have missed the `.cuXX` part?

.github/workflows/build-wheel.yml
cpcloud: They're accessed and defined as part of a configuration file in a fairly well-scoped, specific tool (`cibuildwheel`). In fact, all the additional `CIBW_*` variables are closer to anything resembling a global variable.

ci/tools/merge_cuda_core_wheels.py
cpcloud: Yep, but I think the result is the same:

```
In [3]: def alter_wheel_name1(base_wheel_path):
   ...:     return Path(base_wheel_path.name.split(".cu")[0] + ".whl")
   ...:

In [4]: def alter_wheel_name2(base_wheel_path):
   ...:     return base_wheel_path.with_suffix(".whl")
   ...:

In [5]: from pathlib import Path

In [6]: path = Path("foo.cu13")

In [7]: alter_wheel_name1(path)
Out[7]: PosixPath('foo.whl')

In [8]: alter_wheel_name2(path)
Out[8]: PosixPath('foo.whl')
```

leofang:
> > We need a new rule: On the week of code freeze, do not alter the linter config. It is annoying 🤬
> 
> Perhaps something to enforce the freeze is in order then. What's the point of a freeze if nothing is actually frozen?

What I meant was: This week we're wrapping up a new `cuda.core` release. I expect we only merge PRs that are P0 (and then P1 if we have extra bw). I don't think linter changes should be P0.

cpcloud:
Fair enough. I don't know where we can find this information (or policy) outside of PR commentary :)

ci/tools/merge_cuda_core_wheels.py
leofang: nice thx

leofang:
> I don't know where we can find this information (or policy) outside of PR commentary :)

Before the team size expanded, I tagged every issue and PR with the P0/P1/P2 labels based on my judgment. It seems I am behind the fantastic speed at which we make new PRs lately, so Rob and I discussed and we plan to work out a new process to keep the team informed, but avoid daily triaging stand-ups if possible. https://github.com/NVIDIA/cuda-python/issues/1025 is relevant in this context. We can discuss this further offline.

leofang:
/ok to test [4ba0090](https://github.com/NVIDIA/cuda-python/pull/1041/commits/4ba0090a740cac0743186aa1bca6dd7cfc0c04fd)

.github/workflows/build-wheel.yml
```diff
@@ -170,6 +155,48 @@ jobs:
           path: ${{ env.CUDA_BINDINGS_ARTIFACTS_DIR }}/*.whl
           if-no-files-found: error
 
+      # TODO: ideally we want to build against public cuda-bindings
```
kkraus14: Do we actually? If we want to add a new feature in `cuda.bindings` and optionally take advantage of it in `cuda.core`, using the `PIP_FIND_LINKS` based approach allows us to ship that as an atomic piece of work quite nicely.

cuda_core/cuda/core/experimental/_utils/cuda_utils.pxd
```diff
@@ -2,18 +2,29 @@
...
-
 cimport cpython
-cimport libc.stdint
+from libc.stdint cimport int64_t
+
+from cuda.bindings cimport cydriver
+
+
+ctypedef fused supported_error_type:
+    cydriver.CUresult
```
kkraus14, line 12: Was making this fused type for future usage?

cuda_core/cuda/core/experimental/__init__.py
```diff
@@ -2,29 +2,64 @@
...
+else:
+    cuda_major, cuda_minor = cuda.bindings.__version__.split(".")[:2]
+    if cuda_major not in ("12", "13"):
+        raise ImportError("cuda.bindings 12.x or 13.x must be installed")
+
+import importlib
+
+subdir = f"cu{cuda_major}"
+try:
+    verioned_mod = importlib.import_module(f".{subdir}", __package__)
```
kkraus14:
```suggestion
    versioned_mod = importlib.import_module(f".{subdir}", __package__)
```

cuda_core/cuda/core/experimental/__init__.py
```diff
@@ -2,29 +2,64 @@
...
+    # Import all symbols from the module
+    globals().update(verioned_mod.__dict__)
+except ImportError:
+    # This is not a wheel build, but a conda or local build, do nothing
+    pass
+else:
+    del verioned_mod
```
kkraus14:
```suggestion
    del versioned_mod
```

cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
```diff
@@ -52,6 +52,12 @@ def _reduce_3_tuple(t: tuple):
     return t[0] * t[1] * t[2]
 
 
+cdef int HANDLE_RETURN(supported_error_type err) except?-1:
+    if supported_error_type is cydriver.CUresult:
+        if err != cydriver.CUresult.CUDA_SUCCESS:
+            return _check_driver_error(err)
```
kkraus14, line 58: Should we make this `nogil` and only acquire the gil in the case of the result not being a success in order to throw?

ci/tools/merge_cuda_core_wheels.py
```diff
@@ -0,0 +1,189 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+#!/usr/bin/env python3
```
mdboom: nit: the shebang has to be the first line, IIUC.

cuda_core/cuda/core/experimental/_device.pyx
```diff
@@ -1078,20 +1079,23 @@ def uuid(self) -> str:
...
 
         """
-        driver_ver = handle_return(driver.cuDriverGetVersion())
-        if 11040 <= driver_ver < 13000:
-            uuid = handle_return(driver.cuDeviceGetUuid_v2(self._id))
-        else:
-            uuid = handle_return(driver.cuDeviceGetUuid(self._id))
-        uuid = uuid.bytes.hex()
+        cdef cydriver.CUuuid uuid
+        IF CUDA_CORE_BUILD_MAJOR == "12":
```
kkraus14, line 1083: Cython conditional compilation is deprecated where we probably shouldn't do this: https://docs.cython.org/en/latest/src/userguide/migrating_to_cy30.html#deprecation-of-def-if

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -59,7 +64,7 @@ class IsStreamT(Protocol):
         ...
 
 
-def _try_to_get_stream_ptr(obj: IsStreamT):
+cdef cydriver.CUstream _try_to_get_stream_ptr(obj: IsStreamT) except*:
```
kkraus14, line 67: Not needed in this PR, but we should probably short circuit all of the checks if the passed object is a `cuda.bindings` / `cuda.core` type

ci/tools/merge_cuda_core_wheels.py
kkraus14, line 1: Building a single multi-CUDA wheel / pkg shouldn't be something only in CI where we should pull this script out of the `ci` folder and pull it into the general python build commands. Inevitably, someone will need to repackage `cuda.core` from source and will want to support two major CUDA versions.

cuda_core/cuda/core/experimental/__init__.py
rparolin: Isn't there linters that we can run in CI to catch typos like that?

cuda_core/cuda/core/experimental/_utils/cuda_utils.pxd
leofang: Yes I would like to cover cudart, NVRTC, nvJitLink, ... in a future PR.

cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
leofang: I have not released GILs in this PR so it is a follow-up we must do, and we can make this function `nogil` when working on this. I dunno if I have time before this release, though.

cuda_core/cuda/core/experimental/_device.pyx
leofang: I am fully aware. I was one of the first people arguing upstream: https://github.com/cython/cython/issues/4310#issuecomment-887720639. Unfortunately there is no other simple way. The upstream insisted everyone to do codegen, for example h5py is testing Tempita (Cython's template language), but I think it is an overkill for our purpose. 

If Cython really breaks us, we can always pin at a working version and work out a new solution. But they'd break CuPy first and I'll know right away if this happens and yell at them 🙂

ci/tools/merge_cuda_core_wheels.py
leofang: I would like to punt on this if that's OK. This script is currently not needed except when building wheels. For example we don't need this in the conda recipe. I don't want to commit to this support widely unless there is a real user.

cuda_core/cuda/core/experimental/_device.pyx
leofang: Also the upstream has created a compile-time option to turn off the deprecation warning (which I did set in the build backend): https://github.com/cython/cython/issues/4310#issuecomment-1640691259

ci/tools/merge_cuda_core_wheels.py
leofang: fixed in commit 20c5a99

cuda_core/cuda/core/experimental/__init__.py
leofang: fixed in commit 20c5a99

cuda_core/cuda/core/experimental/__init__.py
leofang: fixed in commit 20c5a99

> Isn't there linters that we can run in CI to catch typos like that?

I don't think we have a spellchecker added to pre-commit, would be nice to add indeed

.github/workflows/build-wheel.yml
leofang: I removed the comment in commit 20c5a991. Yes `PIP_FIND_LINKS` offers this benefit and is verified in the CI in particular with the fresh PY313t/314/314t builds, where we do not have any public releases yet and must rely on the artifacts built in this repo. 

This does mean, though, that if we use any top-of-the-tree features from cuda-bindings, we must make a joint release when releasing cuda-core and bump the lower bound on cuda-bindings. I think this is a bit too aggressive but I don't have a better way to enforce a guard. @mdboom is developing an ABI checker that we might be able to reuse here...?

leofang:
/ok to test [473f8f2](https://github.com/NVIDIA/cuda-python/pull/1041/commits/473f8f281a2c414c20baa7eee8d92b6f89206627)

cuda_core/cuda/core/experimental/_utils/cuda_utils.pxd
kkraus14: We should make sure that expanding the fused type here doesn't somehow cause compatibility issues

.github/workflows/build-wheel.yml
kkraus14: We should have some kind of CI job in the release process that verifies a planned `cuda.core` release with the oldest and newest supported `cuda.bindings` releases

cuda_core/cuda/core/experimental/_device.pyx
kkraus14: Makes sense. We need to ensure that we don't bring any of these conditional compilations into `.pxd` files otherwise they end up carrying transitively to any project that cimports the module that uses the conditional compilation.

kkraus14: Resolving this.

.github/workflows/build-wheel.yml
kkraus14: Resolving for now though.

kkraus14:
@leofang given the scope of this change, I think we should consider punting this until after the 0.4.0 release. That way we can kick the tires a bit on it in development before releasing it outwards?

leofang:
What would be the concern(s) that we must address before merging? I'd rather punt on pushing out 0.4.0 by a few more days instead of punting on this PR. It's an arbitrary timeline that we set to ourselves, and it seems even without considering this PR we have other P0 that could potentially also be slipping.

.github/workflows/build-wheel.yml
leofang: We could use https://github.com/NVIDIA/cuda-python/issues/1030 to cover this need, cc @mdboom for vis

cuda_core/cuda/core/experimental/_utils/cuda_utils.pxd
leofang: No it won't. It's effectively generating multiple copies of the same function with different input types and C names, kinda like poorman's `type_traits`. We've been using fused types in a few places, e.g. the kernel arg handling.

leofang: It might be worth noting that this PR does NOT expose any Cython-level API. We don't have such thing yet in cuda.core, and to do so it'd require careful planning.

cuda_core/cuda/core/experimental/__init__.py
```diff
@@ -2,29 +2,64 @@
...
+finally:
+    del cuda.bindings, importlib, subdir, cuda_major, cuda_minor
+
+import sys  # noqa: E402
+import warnings  # noqa: E402
+
+if sys.version_info < (3, 10):
+    warnings.warn(
+        "support for Python 3.9 and below is deprecated and subject to future removal",
+        category=UserWarning,
```
kkraus14: I think instead of `UserWarning` it's more appropriate to use either `DeprecationWarning` or `FutureWarning` (I'm not 100% clear on which one is correct for our usage here)

cuda_core/cuda/core/experimental/_device.pyx
leofang: That's a good point, but again repeating [my comment above](https://github.com/NVIDIA/cuda-python/pull/1041#discussion_r2396393639): All of the pxd headers in cuda.core currently are private. I expect we honor the Python convention even in Cython, that if a module prefixed with an underscore is (c)imported, the users are on their own.

leofang:
There is yet another merge conflict... We need to enforce the practice that only P0 PRs can be merged on the week of release 🤬

leofang:
I'll review/merge #1020 first, and then solve the merge conflicts at once.

kkraus14:
> What would be the concern(s) that we must address before merging? I'd rather punt on pushing out 0.4.0 by a few more days instead of punting on this PR. It's an arbitrary timeline that we set to ourselves, and it seems even without considering this PR we have other P0 that could potentially also be slipping.

If you feel confident we can always roll it back if it turns out to be an issue, but that sounds like a potentially painful 0.4.1.

The biggest concern that I have is that we now have two different wheel builds and only the combined version is tested in CI. Local users building from source will be using a package that isn't tested in CI.

ci/tools/merge_cuda_core_wheels.py
kkraus14: I'm okay with punting it for the future, but it is concerning that we no longer test the unmerged wheels in CI

cuda_core/cuda/core/experimental/__init__.py
leofang: My intend was to drop 3.9 support immediately in the next release (either 0.5.0 or 0.4.1), so I made it as noisy as possible. NumPy has a `VisibleDeprecationWarning` that I wish to exist as a built-in warning.. 😛 The default `DeprecationWarning` won't show up in the end user land. WDYT?

leofang:
> The biggest concern that I have is that we now have two different wheel builds and only the combined version is tested in CI. Local users building from source will be using a package that isn't tested in CI.

Ah! I guess the approach of the current PR is based on the implicit assumption that I've been sticking to... that we optimize over end user experience over developer experience 😅  

I think we will find issues during daily development if it does not work, because the expectation is that locally one would only build against a single CUDA version. But if adding an CI can help ensuring this, I can add a separate CI pipeline in the next PR? (It'd have to be a standalone pipeline, though, no participation in the build/test matrices.)

cuda_core/cuda/core/experimental/_stream.pyx
leofang: Yes. I think in `_launcher.pyx` it's already short-circuited on the call site, prior to calling this helper. 

In `_stream.pyx` we don't need short-circuiting because this helper is only called when a foreign object is passed. So if someone writes silly code like this
```python
s1 = dev.create_stream()
s1_view = dev.create_stream(obj=s1)
```
it'd work but I don't think it makes sense to optimize this case.

ci/tools/merge_cuda_core_wheels.py
leofang: > but it is concerning that we no longer test the unmerged wheels in CI

replied here: https://github.com/NVIDIA/cuda-python/pull/1041#issuecomment-3358837423

cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
leofang: Created https://github.com/NVIDIA/cuda-python/issues/1065 to track this.

.github/workflows/build-wheel.yml
leofang: Ah OK, I see why we have different perspectives. You probably have in mind that from the build system perspective (that runs inside the docker launched by cibuildwheel) it sees tons of env vars resembling global variables, is it right?

My angle is as a user of the cibuildwheel GHA. The documentation clearly states that using `env:` is the way to pass build options: https://cibuildwheel.pypa.io/en/stable/ci-services/#github-actions. As a GHA user, I want to see how an action is invoked.

kkraus14:
> Ah! I guess the approach of the current PR is based on the implicit assumption that I've been sticking to... that we optimize over end user experience over developer experience 😅
> 
> I think we will find issues during daily development if it does not work, because the expectation is that locally one would only build against a single CUDA version. But if adding an CI can help ensuring this, I can add a separate CI pipeline in the next PR? (It'd have to be a standalone pipeline, though, no participation in the build/test matrices.)

Of course we optimize for the end user, but if we accidentally break being able to build locally because its uncovered in CI that will inevitably impact the quality of software we deliver.

Adding a separate CI pipeline sounds reasonable and doing it in a follow up is fine. Ideally we cover both CUDA 12.x and 13.x on both Windows and Linux?

cuda_core/cuda/core/experimental/__init__.py
kkraus14: I believe Pandas uses `FutureWarning` which is a built in warning type we could use?

leofang:
```suggestion
        category=FutureWarning,
```

kkraus14 (approved):

cuda_core/cuda/core/experimental/__init__.py
dalcinl: `VisibleDeprecationWarning` is simply a subclass of `UserWarning`, that's the reason you see it. 

Indeed, `FutureWarning` is the proper built-in warning to use when users are supposed to see a deprecation.
https://docs.python.org/3/library/exceptions.html#FutureWarning

cuda_core/cuda/core/experimental/__init__.py
```diff
@@ -2,29 +2,64 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-from cuda.core.experimental import utils
-from cuda.core.experimental._device import Device
-from cuda.core.experimental._event import Event, EventOptions
-from cuda.core.experimental._graph import (
+try:
+    import cuda.bindings
+except ImportError:
```
rwgk, line 7: Could `ModuleNotFoundError` be better here (less likely to mask bugs)? 

(Also below, current line 21)

cuda_core/cuda/core/experimental/__init__.py
```diff
@@ -2,29 +2,64 @@
...
+    if cuda_major not in ("12", "13"):
+        raise ImportError("cuda.bindings 12.x or 13.x must be installed")
+
+import importlib
+
+subdir = f"cu{cuda_major}"
+try:
+    versioned_mod = importlib.import_module(f".{subdir}", __package__)
+    # Import all symbols from the module
+    globals().update(versioned_mod.__dict__)
```
rwgk, line 20: Could this be moved into the `else:` branch a few lines down?

rwgk (commented):
A couple minor suggestions. (This isn't a real review, I only looked through top to bottom for general background.)

cuda_core/cuda/core/experimental/__init__.py
leofang: Since we've been catching ImportError just fine and the linter does not complain, I suppose it is OK.

cuda_core/cuda/core/experimental/__init__.py
leofang: Yes we can, but my slight preference is to keep everything in the `try` block to make the intention clear, and use the `else` block to only do cleanup (to keep the main namespace tidy).

kkraus14 (approved):

PR #1045, opened 2025-09-29
Skip IPC mempool tests on WSL

rparolin:
Skip IPC mempool tests on WSL

- Skip IPC mempool tests on WSL using pytest.skipif(IS_WSL, ...).

On WSL2, cuMemPoolCreate with POSIX handle returns CUDA_ERROR_INVALID_VALUE despite capability flags indicating support. This is a platform/driver limitation, not a bug in our code. 

On my machine I confirmed the failures are driver-related rather than a bug in our code. The host is WSL2 (kernel “microsoft-standard-WSL2”) with NVIDIA driver 581.15 (CUDA 13.0). Device attributes report mempools and POSIX FD handle support, yet a minimal, direct driver repro calling cuMemPoolCreate with handleTypes=CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR consistently returns CUDA_ERROR_INVALID_VALUE, while the same call with handleTypes=CU_MEM_HANDLE_TYPE_NONE succeeds. 

```

Because this reproduces outside our code with the same CUmemPoolProps we set, the issue lies in the driver/runtime path under WSL2 (consistent with known IPC limitations there), not our implementation. Therefore, we skip IPC mempool tests on WSL to keep the suite portable and green, while leaving the tests enabled on native Linux and other environments where the driver accepts IPC pools.

### DIFF
--- a/cuda_core/tests/conftest.py
+++ b/cuda_core/tests/conftest.py
@@ -1,17 +1,35 @@
-# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+import multiprocessing
+import os
+
 import helpers
+import pytest
 
 try:
     from cuda.bindings import driver
 except ImportError:
     from cuda import cuda as driver
-import multiprocessing
-
-import pytest
 from cuda.core.experimental import Device, _device
 from cuda.core.experimental._utils.cuda_utils import handle_return
+
+
+def _detect_wsl() -> bool:
+    try:
+        with open("/proc/sys/kernel/osrelease") as f:
+            data = f.read().lower()
+        if "microsoft" in data or "wsl" in data:
+            return True
+    except Exception:
+        pass
+    # Fallback: env hints sometimes present in CI or shells
+    if any(os.environ.get(k) for k in ("WSL_DISTRO_NAME", "WSL_INTEROP")):
+        return True
+    return False
+
+
+IS_WSL = _detect_wsl()
 
 
 @pytest.fixture(scope="session", autouse=True)
--- a/cuda_core/tests/test_ipc_mempool.py
+++ b/cuda_core/tests/test_ipc_mempool.py
@@ -11,6 +11,7 @@ import multiprocessing
 
 import pytest
 from cuda.core.experimental import Buffer, Device, DeviceMemoryResource, IPCChannel, MemoryResource
+from conftest import IS_WSL
 from cuda.core.experimental._utils.cuda_utils import handle_return
 
 CHILD_TIMEOUT_SEC = 10
@@ -33,9 +34,12 @@ def ipc_device():
     if not device.properties.handle_type_posix_file_descriptor_supported:
         pytest.skip("Device does not support IPC")
 
+    # On WSL, this test is skipped via decorator; on other platforms it runs.
+
     return device
 
 
+@pytest.mark.skipif(IS_WSL, reason="WSL does not support CUDA IPC mempool sharing")
 def test_ipc_mempool(ipc_device):
     """Test IPC with memory pools."""
     # Set up the IPC-enabled memory pool and share it.
@@ -83,6 +87,7 @@ def child_main1(channel, queue):
     stream.sync()
 
 
+@pytest.mark.skipif(IS_WSL, reason="WSL does not support CUDA IPC mempool sharing")
 def test_shared_pool_errors(ipc_device):
     """Test expected errors with allocating from a shared IPC memory pool."""
     # Set up the IPC-enabled memory pool and share it.
--- a/cuda_core/tests/test_memory.py
+++ b/cuda_core/tests/test_memory.py
@@ -13,6 +13,7 @@ import pytest
 from cuda.core.experimental import Buffer, Device, DeviceMemoryResource, MemoryResource
 from cuda.core.experimental._memory import DLDeviceType, IPCBufferDescriptor
 from cuda.core.experimental._utils.cuda_utils import handle_return
+from conftest import IS_WSL
 
 POOL_SIZE = 2097152  # 2MB size
 
@@ -359,7 +360,13 @@ def test_mempool(mempool_device):
     buffer.close()
 
 
-@pytest.mark.parametrize("ipc_enabled", [True, False])
+@pytest.mark.parametrize(
+    "ipc_enabled",
+    [
+        pytest.param(True, marks=pytest.mark.skipif(IS_WSL, reason="WSL does not support CUDA IPC mempool sharing")),
+        False,
+    ],
+)
 @pytest.mark.parametrize(
     "property_name,expected_type",
     [

### CODE REVIEW

File: cuda_core/tests/conftest.py [hunk 1]
```diff
@@ -1,19 +1,37 @@
...
+        with open("/proc/sys/kernel/osrelease") as f:
+            data = f.read().lower()
+        if "microsoft" in data or "wsl" in data:
+            return True
+    except Exception:
+        pass
+    # Fallback: env hints sometimes present in CI or shells
+    if any(os.environ.get(k) for k in ("WSL_DISTRO_NAME", "WSL_INTEROP")):
+        return True
+    return False
```
cpcloud:
```suggestion
    return any(map(os.environ.get, ("WSL_DISTRO_NAME", "WSL_INTEROP")))
```

No need for the branch.

cuda_core/tests/test_ipc_mempool.py
```diff
@@ -33,9 +34,12 @@ def ipc_device():
     if not device.properties.handle_type_posix_file_descriptor_supported:
         pytest.skip("Device does not support IPC")
 
+    # On WSL, this test is skipped via decorator; on other platforms it runs.
+
```
cpcloud:
```suggestion
```

cuda_core/tests/conftest.py
```diff
@@ -1,19 +1,37 @@
...
 
 
+def _detect_wsl() -> bool:
+    try:
+        with open("/proc/sys/kernel/osrelease") as f:
+            data = f.read().lower()
+        if "microsoft" in data or "wsl" in data:
+            return True
+    except Exception:
+        pass
```
cpcloud: This is too big of an exception scope.


```suggestion
    try:
        with open("/proc/sys/kernel/osrelease") as f:
            data = f.read().lower()
    except OSError:
        pass
    else:
        if "microsoft" in data or "wsl" in data:
            return True
```

Presumably you want to skip only `OSError`s (file doesn't exist, not readable for whatever reason, etc).

cuda_core/tests/conftest.py
```diff
@@ -1,19 +1,37 @@
...
 except ImportError:
     from cuda import cuda as driver
-import multiprocessing
-
-import pytest
 from cuda.core.experimental import Device, _device
 from cuda.core.experimental._utils.cuda_utils import handle_return
 
 
+def _detect_wsl() -> bool:
```
rwgk: We have two of these floating around already:

```
$ git grep 'def platform_is_wsl'
cuda_bindings/tests/test_cufile.py:def platform_is_wsl():
cuda_core/tests/test_event.py:def platform_is_wsl():
```

It'd be ideal to consolidate them. Not sure how best.

rparolin:
Wow, you both jumped on this PR fast. LOL.  I had to double check to make sure I didn't mark this as ready by mistake.

cuda_core/tests/conftest.py
rparolin: Yeah, I moved the cuda_core version into `conftest.py` but that makes it accessible to just cuda_core.  I was going to look into making a globally accessible test utils so we don't duplicate this for every project under cuda-python.

leofang (changes requested):
In general we avoid skipping tests based on only the platform or hard-coded conditions in `cuda.core`. What we should do is to query the driver and see if a functionality is supported or not. The reason is that the driver can be updated to gain new capabilities, and checks like this would not keep up. 

@rparolin could you check what's the returned values on WSL?
```python
>>> dev = Device()
>>> dev.properties.mempool_supported_handle_types
9       
>>> dev.properties.handle_type_posix_file_descriptor_supported
True
>>> dev.properties.handle_type_win32_handle_supported
False
>>> dev.properties.handle_type_win32_kmt_handle_supported
False
```
It could be that our existing skip condition is not sufficient:
https://github.com/NVIDIA/cuda-python/blob/62d6963da3a9e33036c7984252d6661e5838714a/cuda_core/tests/test_ipc_mempool.py#L21-L36

cpcloud (dismissed):

rwgk:
> could you check what's the returned values on WSL?

This is what I'm seeing on my Windows 11 24H2 WSL2 Ubuntu 24.04 workstation:

```
>>> dev = Device()
>>> dev.properties.mempool_supported_handle_types
0
>>> dev.properties.handle_type_posix_file_descriptor_supported
True
>>> dev.properties.handle_type_win32_handle_supported
False
>>> dev.properties.handle_type_win32_kmt_handle_supported
False
```

Using current cuda-python main (at commit dbde2b4ec2c86c4c2c54ee5f27581d11721a2df8).

Ubuntu side:

```
(WslLocalCudaVenv) rwgk-win11.localdomain:~/forked/cuda-python $ nvidia-smi
Mon Sep 29 11:21:40 2025
| NVIDIA-SMI 580.82.02              Driver Version: 581.15         CUDA Version: 13.0     |
|   0  NVIDIA RTX A6000               On  |   00000000:C1:00.0 Off |                  Off |
| 30%   29C    P8             20W /  300W |    1652MiB /  49140MiB |      3%      Default |
```

Windows side:

```
PS C:\Users\rgrossekunst> nvidia-smi
Mon Sep 29 11:22:45 2025
| NVIDIA-SMI 581.15                 Driver Version: 581.15         CUDA Version: 13.0     |
| GPU  Name                  Driver-Model | Bus-Id          Disp.A | Volatile Uncorr. ECC |
|   0  NVIDIA RTX A6000             WDDM  |   00000000:C1:00.0 Off |                  Off |
| 30%   31C    P8             42W /  300W |    1729MiB /  49140MiB |      3%      Default |
```

leofang:
> ```python
> >>> dev.properties.mempool_supported_handle_types
> 0
> >>> dev.properties.handle_type_posix_file_descriptor_supported
> True
> ```

`mempool_supported_handle_types` seems to be correct, which is a bit mask for extracting the supported handle types, and `handle_type_posix_file_descriptor_supported` is where it went south. @Andy-Jost let's check if we messed up in the implementation of `handle_type_posix_file_descriptor_supported`, or if WSL driver is being inconsistent here.

keenan-simpson:
Neither are bugs. 

As stated above, it is expected that IPC mempools do not work on WSL because WSL runs on top of the WDDM driver and support for IPC mempools is not yet added. 

As for 
```dev.properties.handle_type_posix_file_descriptor_supported```
On WSL, this is also true. It doesn't, however, imply that posix file descriptors are supported as handle types for mempools via the mempool props struct.

File: cuda_python_test_helpers/__init__.py [hunk 5]
```diff
@@ -0,0 +1,66 @@
...
+            driver.cuMemPoolDestroy(pool)
+        return True
+    except Exception:
+        return False
+
+
+__all__ = [
+    "_detect_wsl",
+    "IS_WSL",
+    "skip_on_wsl",
```
rwgk: `skip_on_wsl` doesn't seem to exist here?

(I didn't review systematically; just happened to notice this)

cuda_core/tests/test_ipc_mempool.py
Andy-Jost, line 1: FYI, this file is removed in [PR #1020](https://github.com/NVIDIA/cuda-python/pull/1020), with relevant part moved to `memory_ipc/conftest.py`.

cuda_core/tests/test_memory.py
```diff
@@ -379,6 +381,9 @@ def test_mempool_attributes(ipc_enabled, mempool_device, property_name, expected
     if platform.system() == "Windows":
```
Andy-Jost, line 531: This check can probably be removed with your addition below.

cuda_python_test_helpers/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+    with suppress(Exception), open("/proc/sys/kernel/osrelease") as f:
+        data = f.read().lower()
+    if "microsoft" in data or "wsl" in data:
+        return True
+    return any(os.environ.get(k) for k in ("WSL_DISTRO_NAME", "WSL_INTEROP"))
+
+
+IS_WSL: bool = _detect_wsl()
+
+
```
Andy-Jost:
```suggestion
@functools.cache
```

Andy-Jost:
> > ```python
> > >>> dev.properties.mempool_supported_handle_types
> > 0
> > >>> dev.properties.handle_type_posix_file_descriptor_supported
> > True
> > ```
> 
> `mempool_supported_handle_types` seems to be correct, which is a bit mask for extracting the supported handle types, and `handle_type_posix_file_descriptor_supported` is where it went south. @Andy-Jost let's check if we messed up in the implementation of `handle_type_posix_file_descriptor_supported`, or if WSL driver is being inconsistent here.

I confirmed the implementation of `handle_type_posix_file_descriptor_supported` appears to be correct. @ksimpson-work pointed out that the observed behavior is intended.

cuda_python_test_helpers/__init__.py
rparolin: Thanks for pointing that out. I removed it due to feedback from Leo.  Removed.

cuda_python_test_helpers/__init__.py
rparolin: Oh interesting.  Never heard of it before.

```
functools.cache is a decorator in Python's functools module, introduced in Python 3.9, that provides a simple and efficient way to cache the results of function calls. It's essentially an unbounded version of functools.lru_cache, meaning it stores the results of all function calls without any size limit.
```

rparolin: `functools.cache is a decorator in Python's functools module, introduced in Python 3.9, that provides a simple and efficient way to cache the results of function calls. It's essentially an unbounded version of functools.lru_cache, meaning it stores the results of all function calls without any size limit.`

rparolin: Oh cool.  Never heard of this before.

cuda_core/tests/test_memory.py
rparolin: I don't believe that's the same thing.  This is skipping the test if you are running on Windows proper.  I'm running into IPC issues when running Linux on Windows via WSL -- which is technically Linux even though its passing the syscalls down to the Windows kernel.

rparolin:
> Neither are bugs.
> 
> As stated above, it is expected that IPC mempools do not work on WSL because WSL runs on top of the WDDM driver and support for IPC mempools is not yet added.
> 
> As for `dev.properties.handle_type_posix_file_descriptor_supported` On WSL, this is also true. It doesn't, however, imply that posix file descriptors are supported as handle types for mempools via the mempool props struct.

Thanks for confirming this @ksimpson-work

cuda_python_test_helpers/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+        props.win32SecurityAttributes = 0
+        props.usage = 0
+
+        res, pool = driver.cuMemPoolCreate(props)
+        if int(res.value) != 0:
+            return False
+        # Destroy created pool to avoid leaks
+        with suppress(Exception):
+            driver.cuMemPoolDestroy(pool)
+        return True
```
leofang: This is too bulky and unnecessary. I dunno if my earlier snippet informs this notion (notice `dev.set_current()` was not called), but there is a reason `cuda.core` is designed this way. Checking device attributes do not need any active CUDA context, we just need to initialize CUDA (by calling `cuInit(0)`). 

As discussed offline, the proper way to proceed is to check the dev attribute `CU_DEVICE_ATTRIBUTE_MEMPOOL_SUPPORTED_HANDLE_TYPES`, and then check the support via bit mask:
```C
int dev = 0;
int attr;
cuDeviceGetAttribute(&attr, CU_DEVICE_ATTRIBUTE_MEMPOOL_SUPPORTED_HANDLE_TYPES, dev);
if (attr & CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR) {
  // supported
} else {
  // not supported
}
```

cuda_python_test_helpers/__init__.py
leofang, line 1: Can we please not create yet another folder at the repo root? If we are hacking `sys.path` to load this module anyway, how about let's move it to `toolshed/`?

cuda_core/tests/conftest.py
```diff
@@ -1,18 +1,24 @@
-# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
leofang:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

cuda_core/tests/conftest.py
```diff
@@ -1,18 +1,24 @@
...
 except ImportError:
     from cuda import cuda as driver
-import multiprocessing
-
-import pytest
 from cuda.core.experimental import Device, _device
 from cuda.core.experimental._utils.cuda_utils import handle_return
 
+# Import shared platform helpers for tests across repos
+sys.path.insert(0, str(pathlib.Path(__file__).resolve().parents[2]))
```
cpcloud: This feels unnecessary. It's almost never the right solution to start modifying `sys.path`. Can't you just define `IS_WSL` in this module?

cuda_python_test_helpers/__init__.py
cpcloud: Let's avoid path hacking. It really is possible to use the tools as-is, without resorting to hacks on the first attempt.

Just move the WSL detection into `conftest.py` and avoid the path hacking, the unnecessary directory, and the unnecessary module.

leofang: I think path hacking is necessary and can be justified in this case. We are defining common test utilities that can be reused in both cuda.core and cuda.bindings (and potentially cuda.pathfinder) test suites. Without hacking the path we'd end up copying/pasting the same code around.

cpcloud: In that case, moving to `toolshed` is a mistake. No one will ever find it there. For shared test utilities a distinct package is the right place.

rparolin: Yes, that exactly why this folder exists.  Its holding common checks so we don't have to copy/paste variants of them in every module.

rwgk: > In that case, moving to `toolshed` is a mistake. No one will ever find it there. For shared test utilities a distinct package is the right place.

+1

When I added the `toolshed/` directory I had in mind:
* that's where we put stuff that's handy for maintenance every now and then (mainly new CTK versions)
* what's in there is NOT tested in our CI
* also not used directly from anywhere (because then we'd most likely want to test it in our CI)
* there are only pointers to the toolshed/ in production code comments

I didn't write this down. Should I do that now (`toolshed/README.txt`)?

Andy-Jost: Sorry if this is a naive question, but regarding the path hacks, couldn't we just create symlinks back from each test directory to the toolshed? That would also make things there easier to find.

rwgk:
I believe long-term it'll help us greatly if we have shared test utilities. But how best?

To get a birds-eye view, I asked my favorite LLM about it, in a completely general way, and it enumerated some options:

https://chatgpt.com/share/68dd561e-8b38-8008-948d-59b022e04c99

I was hoping there is something obvious and light-weight to avoid manipulating `sys.path` ourselves, but after reviewing the LLM output, I'm thinking there isn't.

The cleanest option would be have a formal pip-installable project ("Option 1"), but that's obviously not light-weight at all. Relying on nox/tox or similar also doesn't look light-weight.

On balance, having a one-line `sys.path` manipulation in each of our conftest.py scripts seems just right. It doesn't look pretty, but it is just one line, and far better than duplicating code.

cuda_python_test_helpers/__init__.py
leofang: Symlinks work great on Linux but unfortunately not out of box on Windows, we found it a bit annoying: https://docs.cupy.dev/en/stable/contribution.html#development-on-microsoft-windows But if we don't plan to serve native Windows developers (sounds like a bad decision, though) I do like symlinks. Alternatively, git submodules can also work, but not without its own complexity. The `sys.path` hack is the cheapest IMHO.

rwgk: Symlinks can be very tricky to work with under Windows. Unless Developer Mode is enabled, you need admin access to create a symlink. I'm not sure what the situation is on our GHA Windows runners.

Andy-Jost:
> I believe long-term it'll help us greatly if we have shared test utilities. But how best?
> 
> To get a birds-eye view, I asked my favorite LLM about it, in a completely general way, and it enumerated some options:
> 
> https://chatgpt.com/share/68dd561e-8b38-8008-948d-59b022e04c99
> 
> I was hoping there is something obvious and light-weight to avoid manipulating `sys.path` ourselves, but after reviewing the LLM output, I'm thinking there isn't.
> 
> The cleanest option would be have a formal pip-installable project ("Option 1"), but that's obviously not light-weight at all. Relying on nox/tox or similar also doesn't look light-weight.
> 
> On balance, having a one-line `sys.path` manipulation in each of our conftest.py scripts seems just right. It doesn't look pretty, but it is just one line, and far better than duplicating code.

Agreed. We need shared utilities, can't use symbolic links, and shouldn't prefer an explicit package. I also see nothing better than modifying `sys.path`.

cuda_python_test_helpers/__init__.py
leofang: > I didn't write this down. Should I do that now (`toolshed/README.txt`)?

`toolshed/README.md` is better.

cuda_core/tests/conftest.py
rparolin: I'm going to resolve this comment b/c another thread discussing the same thing sprung up.

leofang:
Since this is not testable in our CI, let's get other P0 PRs merged first, and circle back here. In the meantime, let's fix the IPC support check: https://github.com/NVIDIA/cuda-python/pull/1045#discussion_r2392397188

cuda_core/tests/conftest.py
rwgk: This is only a detail:

I'd consider this a little cleaner:

1. Change comment to:

```
# Enable `cuda_python_test_helpers` imports
```

2. Import `IS_WSL` only where it's actually needed (i.e. remove here).

It's just one of several helpers. It'll e.g. be surprising that we're picking out one here and not all.

rwgk (approved):

cuda_python_test_helpers/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+
+        # Check POSIX FD handle type support via bitmask
+        posix_fd = driver.CUmemAllocationHandleType.CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR
+        return (int(mask) & int(posix_fd)) != 0
+    except Exception:
+        return False
+
+
+__all__ = [
+    "_detect_wsl",
```
leofang: nit: we already expose `IS_WSL` below so this can be removed?
```suggestion
```

cuda_python_test_helpers/__init__.py
leofang: Thanks, Rob! Commit 2168cd1 LGTM. Resolving...

cuda_python_test_helpers/__init__.py
kkraus14: I'm -1 on modifying `sys.path`. That's a red flag that indicates that we've structured our code in the wrong way.

Given these helpers are using `cuda.core`, we probably shouldn't use them in `cuda.bindings` tests regardless to avoid a circular dependency.

the `IS_WSL` is a small enough utility that copying it around into the different modules doesn't seem like a big deal from my perspective.

rwgk: > That's a red flag that indicates that we've structured our code in the wrong way.

Complete agreement for production code.

In pytest code, I've seen this many times.

Unfortunately pytest is missing a facility for making this nicer. I don't think duplicating our test code is a healthy response. Of course, one little thing doesn't matter, but having such a hindrance tends to push tired or rushed developers (and who isn't one) into non-ideal compromises, although they may not be as obvious as a `sys.path.insert`.

leofang: > Given these helpers are using `cuda.core`, we probably shouldn't use them in `cuda.bindings` tests regardless to avoid a circular dependency.

I think the helpers are written by only calling `cuda.bindings` functions, therefore usable in both binding and core tests?

I do appreciate having a common module we can reuse. In the future I'd like to move some of the NVRTC/NVVM/PTX code snippets to this new module so that we only hard-code them in one place and reuse as much as possible.

kkraus14: I agree having common helpers is ideal, but we shouldn't need to hack `sys.path` in order to accomplish this. As @cpcloud said in https://github.com/NVIDIA/cuda-python/pull/1045#discussion_r2394793016, the correct solution here would be to put the common helpers into a distinct package and then depend on that package instead. Obviously we don't want to actually ship and publish such a package, but for pure python code only to be used locally as part of testing, the `pyproject.toml` would be <10 lines of code and in `cuda.core` / `cuda.bindings` could be included under the testing dependency groups quite easily.

kkraus14: Additionally, some of the functionality being built here is already being duplicated here: https://github.com/NVIDIA/numba-cuda/pull/488

So maybe we do want to ship some of these as something like testing utilities? 😅

cuda_core/tests/test_ipc_mempool.py
```diff
@@ -33,6 +35,10 @@ def ipc_device():
     if not device.properties.handle_type_posix_file_descriptor_supported:
         pytest.skip("Device does not support IPC")
 
+    # Final driver capability probe: skip if the driver rejects IPC mempool creation.
+    if not supports_ipc_mempool(device):
+        pytest.skip("Driver rejects IPC-enabled mempool creation on this platform")
```
kkraus14: This checks the same properties as what's already checked above. There's no situation where this can trigger.

cuda_core/tests/test_event.py
```diff
@@ -47,7 +42,7 @@ def test_timing_success(init_cuda):
     # We only want to exercise the __sub__ method, this test is not meant
     # to stress-test the CUDA driver or time.sleep().
     delay_ms = delay_seconds * 1000
-    if os.name == "nt" or platform_is_wsl():  # noqa: SIM108
+    if os.name == "nt" or IS_WSL:  # noqa: SIM108
```
kkraus14, line 45: Instead of platform checking here we should probably just make the tolerance extremely generous in general. Architecturally we know we're calling into the underlying APIs, where as long as we generally get a positive number we should be good here.

cuda_python_test_helpers/__init__.py
rwgk: Having a package with our test utils sounds great to me.

Would you want to publish that to PyPI, too?

If not, I'm not sure how easy or tricky it'll be for numba-cuda to integrate the package into the dev workflow, but it still seems like a good goal. @cpcloud have you done something like that before?

cuda_core/tests/conftest.py
```diff
@@ -1,18 +1,24 @@
...
+from cuda_python_test_helpers import IS_WSL  # noqa: F401 (imported for test modules)
```
kkraus14: Is this needed in this file? Can't tests import this as needed?

kkraus14 (changes requested):

cuda_core/tests/conftest.py
leofang: +1

I thought it's exposed based on the fact that everything listed in `conftest.py` is loaded in the scope of each test module, but it seems in one of the files we also explicitly import `IS_WSL`, then this is not needed.

leofang:
pre-commit.ci autofix

rparolin:
pre-commit.ci autofix

cuda_python_test_helpers/__init__.py
leofang, line 1: If we must make this module a package then this file should be removed. It is a duplicate.

cuda_core/tests/conftest.py
rparolin: Removed and converted into a separate package.

rparolin:
> @kkraus14 Sorry but I think making this an installable package is a very poor UX. It means every single new contributor would stumble during first test run, and have to learn that there is a separate package that we must install locally first.
> 
> Also, by doing so we need to touch the CI workflows so as to make it installed first before testing. We seem to be shooting ourselves in the foot without justifiable gain?

I had similar thoughts but figured that it could be mitigated by our `script/run_tests.sh` script which can auto install for you.  My tendency is have my build/dev environment tools script all the complexity so others don't step on landmines but I haven't developed an intuition of what is [un]familiar for python programmers yet.

leofang:
The CI shows what I noted earlier:
> Also, by doing so we need to touch the CI workflows so as to make it installed first before testing.

leofang:
/ok to test [bbe82c8](https://github.com/NVIDIA/cuda-python/pull/1045/commits/bbe82c8291d4a354cf785caab7149cc8e39694ab)

leofang:
/ok to test [7726e05](https://github.com/NVIDIA/cuda-python/pull/1045/commits/7726e0539dd40d1fe1f646a787ad01008a04ae34)

leofang:
I pushed commit [bbe82c8](https://github.com/NVIDIA/cuda-python/pull/1045/commits/bbe82c8291d4a354cf785caab7149cc8e39694ab) so that we don't get blocked by this package discussion. If the helper module is installed as a package, it's used. Otherwise, we find it via relative path. We can revisit this discussion later and find a way to avoid friction with local development and CI testing.

kkraus14:
I agree that packaging this as a separate package is a poor developer UX. Maybe we should introduce a `cuda.core.testing` module or something similar?

Regardless, we can do that in a follow up PR.

PR #990, opened 2025-09-19
Add `"-arch=compute_100"` to `nvvm.compile_program` options and restore testing with llvmlite-generated IRs

rwgk:
* Puts us in a good position to generate `MINIMAL_NVVMIR_BITCODE_STATIC` entries as needed.
* Preserves testing with llvmlite-generated IRs.

This could also work:
* Completely remove `llvmlite` from test_nvvm.py
* Move the code for generating new `MINIMAL_NVVMIR_BITCODE_STATIC` entries to a toolshed/ script.

PR #988, opened 2025-09-18
Remove `llvmlite` from `test` dependencies in `cuda_bindings/pyproject.toml`

rwgk:
This PR simply removes `llvmlite` from the `test` dependencies, with leads to turning off the failing tests.

Our existing `MINIMAL_NVVMIR_BITCODE_STATIC` dictionary is currently sufficient to exercise the `nvvm` bindings on all platforms.

For easy future reference, the [`llvmlite==0.45` release](https://pypi.org/project/llvmlite/0.45.0/) caused these test failures:

```
cuda/bindings/nvvm.pyx:54: nvvmError
============================================================================== short test summary info ==============================================================================
FAILED tests/test_nvvm.py::test_compile_program_with_minimal_nvvm_ir[bitcode_dynamic-options0] - RuntimeError: FileNameHere.ll: parse Invalid value (Producer: 'LLVM20.1.8' Reader: 'LLVM 7.0.1')
FAILED tests/test_nvvm.py::test_compile_program_with_minimal_nvvm_ir[bitcode_dynamic-options1] - RuntimeError: FileNameHere.ll: parse Invalid value (Producer: 'LLVM20.1.8' Reader: 'LLVM 7.0.1')
FAILED tests/test_nvvm.py::test_compile_program_with_minimal_nvvm_ir[bitcode_dynamic-options2] - RuntimeError: FileNameHere.ll: parse Invalid value (Producer: 'LLVM20.1.8' Reader: 'LLVM 7.0.1')
FAILED tests/test_nvvm.py::test_verify_program_with_minimal_nvvm_ir[bitcode_dynamic-options0] - cuda.bindings.nvvm.nvvmError: ERROR_INVALID_IR (6)
FAILED tests/test_nvvm.py::test_verify_program_with_minimal_nvvm_ir[bitcode_dynamic-options1] - cuda.bindings.nvvm.nvvmError: ERROR_INVALID_IR (6)
FAILED tests/test_nvvm.py::test_verify_program_with_minimal_nvvm_ir[bitcode_dynamic-options2] - cuda.bindings.nvvm.nvvmError: ERROR_INVALID_IR (6)
=========================================================================== 6 failed, 30 passed in 0.22s ============================================================================
```

PR #1047, opened 2025-09-29
test_nvvm.py simplification / use `llvmlite` in `toolshed/` only

rwgk:
This fully isolates us from `llvmlite` for regular `cuda_bindings` unit testing.

`llvmlite` needs to be installed locally only when the new `toolshed/build_static_bitcode_input.py` helper script is needed, which is expected to be rare.

See also:
* #988
* #990 (obsolete)

### DIFF
--- a/toolshed/build_static_bitcode_input.py
+++ b/toolshed/build_static_bitcode_input.py
@@ -18,18 +18,15 @@ import os
 import sys
 import textwrap
 
-try:
-    import llvmlite.binding
-except Exception:
-    sys.exit("HINT: pip install llvmlite")
-
+import llvmlite.binding  # HINT: pip install llvmlite
 from cuda.bindings import nvvm
 
 
 def get_minimal_nvvmir_txt_template():
     cuda_bindings_tests_dir = os.path.normpath("cuda_bindings/tests")
-    if not os.path.isdir(cuda_bindings_tests_dir):
-        raise RuntimeError("Please run this helper script from the cuda-python top-level directory.")
+    assert os.path.isdir(cuda_bindings_tests_dir), (
+        "Please run this helper script from the cuda-python top-level directory."
+    )
     sys.path.insert(0, os.path.abspath(cuda_bindings_tests_dir))
     import test_nvvm
 


rwgk:
Holding off running tests until after reviews.

toolshed/build_static_bitcode_input.py
```diff
@@ -0,0 +1,53 @@
...
+import binascii
+import os
+import sys
+import textwrap
+
+import llvmlite.binding  # HINT: pip install llvmlite
+from cuda.bindings import nvvm
+
+
+def get_minimal_nvvmir_txt_template():
```
cpcloud, line 25: Can't you just check in a JSON file that contains the bitcode? Why do we need the `sys.path` munging and `print`-generated dictionary?

rwgk: Where would you put the JSON file?

I had on my mind: Avoid creating extra artifacts.
But that's something I could do if you prefer.

Re printing out the generated dictionary: I don't want to over-engineer a helper script that's used only rarely (possibly only every couple years). The main goal is to archive "how it worked last time", so that future maintainers don't have to start from scratch. (Back in February it took me more than a couple hours to figure out the approach.)

cuda_bindings/tests/test_nvvm.py
```diff
@@ -130,43 +121,24 @@
...
+        return MINIMAL_NVVMIR_TXT_TEMPLATE % (major, debug_major)
+
+    bitcode_static_binascii = MINIMAL_NVVMIR_BITCODE_STATIC.get((major, debug_major))
+    if bitcode_static_binascii:
+        return binascii.unhexlify(bitcode_static_binascii)
+    raise RuntimeError(
+        "Static bitcode for NVVM IR version "
+        f"{major}.{debug_major} is not available in this test.\n"
+        "Maintainers: Please run the helper script to generate it and add the "
+        "output to the MINIMAL_NVVMIR_BITCODE_STATIC dict:\n"
```
cpcloud, line 139: I don't follow why we can't make `llvmlite` a required testing dependency and always generate the bitcode and eliminate the static testing fixture altogether.

Then we can avoid a lot of this complexity.

cpcloud: This is mutually exclusive with my JSON file suggestion.

rwgk: > I don't follow why we can't make `llvmlite` a required testing dependency and always generate the bitcode and eliminate the static testing fixture altogether.

Ah, that's whole point of this PR: remove `llvmlite` even as an optional test dependency.

@leofang please correct me if I'm wrong, but I got the impression you were skeptical of the `llvmlite` dependency all the while.

Originally I wanted to keep things simple and make `llvmlite` a required dependency.

Before we had the recent `llvmlite` `v0.45` related breakage (see #988 and numba/llvmlite#1297), I had it on my list to simplify the `test_nvvm.py` code, with `llvmlite` as a hard dependency.

But after the breakage, and seeing how @rparolin stumbled even over the optional dependency last week (see team chat), I got to think it's more trouble than it's worth to even have it as an optional dependency.

With this PR, the `cuda_bindings` unit tests will be fully isolated from `llvmlite`, so we won't stumble that much in the future.

Only every couple years probably someone has to dust off the helper script to add a new entry in `test_nvvm.py`. Estimated effort: 30-60 minutes, depending on prior background of the person who takes this on.

cpcloud: Won't we still need `llvmlite` *eventually* to regenerate the bitcode?

rwgk: Yes, but only "offline", in the sense that the routine `cuda_bindings` unit tests (CI) don't need it.

Based on the few months of related experience that I have:

* I expect that `test_nvvm.py` will only break if we're testing new CTK releases.
* Fixing up `test_nvvm.py` will be just one of potentially several adjustments we have to make for new CTK versions.
* Chances of `test_nvvm.py` breakages due to minor-version CTK releases are probably small. — Every couple of months roughly.
* Chances of `test_nvvm.py` breakages due to major-version CTK releases are significant. — Every couple years.

Making `llvmlite` a required dependency means that we'll be living at the bleeding edge: random breakages at random times (from NVIDIA's viewpoint) that need immediate attention.

With the static bitcode inputs, we'll only need to tend to the test when there are CTK changes (release schedule controlled by NVIDIA).

___

For completeness: There are other ways to convert the `txt` version to `bitcode`. I don't remember exactly, I believe there are llvm commands that could be used instead. What's most suitable might change in the future, but I'm guessing as long as `numba-cuda` uses `llvmlite`, it'll be a good choice.

rwgk:
Thanks @cpcloud!

PR #1048, opened 2025-09-29
Fix #1043: Fix memory leak in StridedMemoryView

### LINKED ISSUE #1043: [BUG]: StridedMemoryView leaks the producer memory
### Is this a duplicate?

### Type of Bug

Something else

### Component

cuda.core

### Describe the bug

Obtaining a StridedMemoryView appears to prevent the producer array from being released.

### How to Reproduce

Running the following script will throw an out of memory error:
```
cupy.cuda.memory.OutOfMemoryError: Out of memory allocating 1,073,741,824 bytes (allocated so far: 3,221,225,472 bytes).
```

```python

import cupy as xp

for idx in range(1000):
    print(idx)
    arr = xp.zeros((1024, 1024, 1024), dtype=xp.uint8)
    view = StridedMemoryView(arr, stream_ptr=-1)
```

Conversely, the following runs to completion
```python
import cupy as xp

for idx in range(1000):
    print(idx)
    arr = xp.zeros((1024, 1024, 1024), dtype=xp.uint8)
```

### Expected behavior

The script should run to completion.

### Operating System

_No response_

### nvidia-smi output

_No response_

mdboom:
## Checklist


cuda_bindings/benchmarks/test_cupy.py
leofang, line 1: Let's move the new test to cuda.core tests. We don't necessarily have cuda-core installed when testing cuda-bindings.

leofang:
btw we also need a rel-note entry for this fix

mdboom:
While this fix definitely works with the reproducer, I'm a little unsure as to why the destructor on the capsule returned by `cupy.__dlpack__` isn't getting called to begin with.  Since it's Python objects all over here, I'm not sure why explicit destruction is necessary.  I'm converting this to a draft to take some time to convince myself there isn't some other problem going on here.

leofang:
You can search the documentation that we built for DLPack, the text around `used_dltensor` is relevant:
https://dmlc.github.io/dlpack/latest/python_spec.html

mdboom:
> You can search the documentation that we built for DLPack, the text around `used_dltensor` is relevant: https://dmlc.github.io/dlpack/latest/python_spec.html

Thanks for that.  I now see why this PR makes sense in the context of what `StridedMemoryView` already does.  I have some reading to do to understand why something called a "view" would take ownership of the thing it's viewing -- I think my mental model of what's really happening just needs some filling in...

leofang:
> why something called a "view" would take ownership of the thing it's viewing -- I think my mental model of what's really happening just needs some filling in...

Yeah, this is unfortunate because in DLPack we follow what the Python buffer protocol & memoryview do. We increment the refcount of the exporting object until the view is destroyed. 

FWIW, after this fix we'd still be hitting exactly the same issue as CuTe tensors (which are also views): https://github.com/NVIDIA/cutlass/issues/2479. My advice to the CUTLASS/CuTe team was that views aren't supposed to be held indefinitely.

cuda_core/tests/test_memory.py
```diff
@@ -437,3 +441,11 @@ def test_mempool_attributes_ownership(mempool_device):
     with pytest.raises(RuntimeError, match="DeviceMemoryResource is expired"):
         _ = attributes.used_mem_high
     mr._mempool_handle = old_handle
+
+
+# Ensure that memory views dellocate their reference to dlpack/cupy tensors
+@pytest.mark.skipif(cupy is None, reason="cupy is not installed")
+def test_strided_memory_view_leak(benchmark):
```
kkraus14: Is the `benchmark` fixture intentional here?

cuda_core/tests/test_memory.py
```diff
@@ -437,3 +441,11 @@ def test_mempool_attributes_ownership(mempool_device):
...
+    for idx in range(1000):
+        arr = cupy.zeros((1024, 1024, 1024), dtype=cupy.uint8)
+        StridedMemoryView(arr, stream_ptr=-1)
```
kkraus14: Instead of doing 1000 iterations, we might be able to do something much smaller and introspect the cupy memory pool to ensure memory is being freed as expected: https://docs.cupy.dev/en/stable/reference/generated/cupy.cuda.MemoryPool.html#cupy.cuda.MemoryPool

Or if we're feeling ambitious we could use a temporary custom allocator for CuPy that we could use to track the allocations and deallocations: https://docs.cupy.dev/en/stable/reference/generated/cupy.cuda.using_allocator.html

kkraus14 (approved):
Implementation looks good, but a couple of questions / comments related to the testing

cuda_core/tests/test_memory.py
pijyoi: I was the one who raised #1043
Could the test be simplified to:
```python
arr = np.zeros(1048576, dtype=np.uint8)
before = sys.getrefcount(arr)
for idx in range(10):
    StridedMemoryView(arr, stream_ptr=-1)
after = sys.getrefcount(arr)
assert before == after
```
Using numpy also allows the test to run without cupy.

cuda_core/tests/test_memory.py
mdboom: Nope.  Removing.

cuda_core/tests/test_memory.py
mdboom: Yep, this test works.  I confirmed it breaks before this PR.

cuda_core/tests/test_memory.py
```diff
@@ -1,18 +1,24 @@
...
+
 try:
     from cuda.bindings import driver
 except ImportError:
     from cuda import cuda as driver
-
+try:
+    import numpy as np
+except ImportError:
+    np = None
```
leofang, line 13: nit: currently numpy is already a required dependency for cuda.core, so we don't need try-except here.

cuda_core/tests/test_memory.py
```diff
@@ -437,3 +443,14 @@ def test_mempool_attributes_ownership(mempool_device):
     with pytest.raises(RuntimeError, match="DeviceMemoryResource is expired"):
         _ = attributes.used_mem_high
     mr._mempool_handle = old_handle
+
+
+# Ensure that memory views dellocate their reference to dlpack tensors
+@pytest.mark.skipif(np is None, reason="numpy is not installed")
```
leofang, line 449: ditto

PR #1052, opened 2025-09-30
Fix #951: Fallback gracefully if a library can't be loaded

### LINKED ISSUE #951: Library bindings are not importable if the underlying library is not installed
During the investigation of #920 I noticed that if I remove nvJitLink from my current environment, `cuda.core` cannot fall back to use the `cuLink*` driver APIs because the pathfinder error is not caught:
```python
  File "/local/home/leof/dev/debug_920.py", line 94, in <module>
    program = Program(code, code_type="ptx", options=options)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/local/home/leof/dev/cuda-python/cuda_core/cuda/core/experimental/_program.py", line 411, in __init__
    ObjectCode._init(code.encode(), code_type), options=self._translate_program_options(options)
                                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/local/home/leof/dev/cuda-python/cuda_core/cuda/core/experimental/_program.py", line 420, in _translate_program_options
    return LinkerOptions(
           ^^^^^^^^^^^^^^
  File "<string>", line 24, in __init__
  File "/local/home/leof/dev/cuda-python/cuda_core/cuda/core/experimental/_linker.py", line 189, in __post_init__
    _lazy_init()
  File "/local/home/leof/dev/cuda-python/cuda_core/cuda/core/experimental/_linker.py", line 69, in _lazy_init
    _decide_nvjitlink_or_driver()
  File "/local/home/leof/dev/cuda-python/cuda_core/cuda/core/experimental/_linker.py", line 47, in _decide_nvjitlink_or_driver
    if inner_nvjitlink._inspect_function_pointer("__nvJitLinkVersion") == 0:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "cuda/bindings/_internal/nvjitlink.pyx", line 247, in cuda.bindings._internal.nvjitlink._inspect_function_pointer
  File "cuda/bindings/_internal/nvjitlink.pyx", line 250, in cuda.bindings._internal.nvjitlink._inspect_function_pointer
  File "cuda/bindings/_internal/nvjitlink.pyx", line 198, in cuda.bindings._internal.nvjitlink._inspect_function_pointers
  File "cuda/bindings/_internal/nvjitlink.pyx", line 70, in cuda.bindings._internal.nvjitlink._check_or_init_nvjitlink
  File "cuda/bindings/_internal/nvjitlink.pyx", line 92, in cuda.bindings._internal.nvjitlink._check_or_init_nvjitlink
  File "cuda/bindings/_internal/nvjitlink.pyx", line 58, in cuda.bindings._internal.nvjitlink.load_library
  File "/local/home/leof/miniforge3/envs/py312_cu129/lib/python3.12/site-packages/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py", line 120, in load_nvidia_dynamic_lib
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/local/home/leof/miniforge3/envs/py312_cu129/lib/python3.12/site-packages/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py", line 48, in _load_lib_no_cache
    found.raise_if_abs_path_is_None()
  File "/local/home/leof/miniforge3/envs/py312_cu129/lib/python3.12/site-packages/cuda/pathfinder/_dynamic_libs/find_nvidia_dynamic_lib.py", line 199, in raise_if_abs_path_is_None
cuda.pathfinder._dynamic_libs.load_dl_common.DynamicLibNotFoundError: Failure finding "libnvJitLink.so": No such file: libnvJitLink.so*, No such file: libnvJitLink.so*

```
This echos @kkraus14's concern earlier (#728). We want to be able to catch errors at run time when the underlying library (including the driver) does not provide the needed symbol, which implies that the exception raising should be deferred to as late as when it's actually needed.

mdboom:
## Description

Allows falling back to another library more lazily.

closes #951 

## Checklist

### DIFF
--- a/cuda_bindings/cuda/bindings/_internal/nvjitlink_linux.pyx
+++ b/cuda_bindings/cuda/bindings/_internal/nvjitlink_linux.pyx
@@ -9,7 +9,7 @@ from libc.stdint cimport intptr_t, uintptr_t
 import threading
 from .utils import FunctionNotFoundError, NotSupportedError
 
-from cuda.pathfinder import load_nvidia_dynamic_lib
+from cuda.pathfinder import load_nvidia_dynamic_lib, DynamicLibNotFoundError
 
 
 ###############################################################################
@@ -73,7 +73,11 @@ cdef void* __nvJitLinkVersion = NULL
 
 
 cdef void* load_library() except* with gil:
-    cdef uintptr_t handle = load_nvidia_dynamic_lib("nvJitLink")._handle_uint
+    cdef uintptr_t handle
+    try:
+        handle = load_nvidia_dynamic_lib("nvJitLink")._handle_uint
+    except DynamicLibNotFoundError:
+        handle = 0
     return <void*>handle
 
 
--- a/cuda_bindings/cuda/bindings/_internal/cufile_linux.pyx
+++ b/cuda_bindings/cuda/bindings/_internal/cufile_linux.pyx
@@ -9,7 +9,7 @@ import threading
 
 from .utils import FunctionNotFoundError, NotSupportedError
 
-from cuda.pathfinder import load_nvidia_dynamic_lib
+from cuda.pathfinder import load_nvidia_dynamic_lib, DynamicLibNotFoundError
 
 import cython
 
@@ -104,7 +104,11 @@ cdef void* __cuFileGetParameterPosixPoolSlabArray = NULL
 
 
 cdef void* load_library() except* with gil:
-    cdef uintptr_t handle = load_nvidia_dynamic_lib("cufile")._handle_uint
+    cdef uintptr_t handle
+    try:
+        handle = load_nvidia_dynamic_lib("cufile")._handle_uint
+    except DynamicLibNotFoundError:
+        handle = 0
     return <void*>handle
 
 
--- a/cuda_bindings/cuda/bindings/_internal/nvjitlink_windows.pyx
+++ b/cuda_bindings/cuda/bindings/_internal/nvjitlink_windows.pyx
@@ -4,12 +4,12 @@
 #
 # This code was automatically generated across versions from 12.0.1 to 13.0.1. Do not modify it directly.
 
-from libc.stdint cimport intptr_t
+from libc.stdint cimport intptr_t, uintptr_t
 
 import threading
 from .utils import FunctionNotFoundError, NotSupportedError
 
-from cuda.pathfinder import load_nvidia_dynamic_lib
+from cuda.pathfinder import load_nvidia_dynamic_lib, DynamicLibNotFoundError
 
 from libc.stddef cimport wchar_t
 from libc.stdint cimport uintptr_t
@@ -95,12 +95,14 @@ cdef void* __nvJitLinkVersion = NULL
 
 cdef int __check_or_init_nvjitlink() except -1 nogil:
     global __py_nvjitlink_init
-    if __py_nvjitlink_init:
-        return 0
+    cdef uintptr_t handle
 
     with gil, __symbol_lock:
         # Load library
-        handle = load_nvidia_dynamic_lib("nvJitLink")._handle_uint
+        try:
+            handle = load_nvidia_dynamic_lib("nvJitLink")._handle_uint
+        except DynamicLibNotFoundError:
+            handle = 0
 
         # Load function
         global __nvJitLinkCreate
--- a/cuda_bindings/cuda/bindings/_internal/nvvm_linux.pyx
+++ b/cuda_bindings/cuda/bindings/_internal/nvvm_linux.pyx
@@ -9,7 +9,7 @@ from libc.stdint cimport intptr_t, uintptr_t
 import threading
 from .utils import FunctionNotFoundError, NotSupportedError
 
-from cuda.pathfinder import load_nvidia_dynamic_lib
+from cuda.pathfinder import load_nvidia_dynamic_lib, DynamicLibNotFoundError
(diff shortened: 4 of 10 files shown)


cuda_bindings/cuda/bindings/_internal/nvjitlink_linux.pyx
```diff
@@ -73,7 +73,11 @@ cdef void* __nvJitLinkVersion = NULL
 
 
 cdef void* load_library() except* with gil:
-    cdef uintptr_t handle = load_nvidia_dynamic_lib("nvJitLink")._handle_uint
+    cdef uintptr_t handle
+    try:
+        handle = load_nvidia_dynamic_lib("nvJitLink")._handle_uint
+    except DynamicLibNotFoundError:
+        handle = 0
```
cpcloud, line 80: Is calling `dlsym` on a `NULL` pointer `handle` undefined behavior? It seems with this change that code path is now easily possible.

cpcloud: According to https://pubs.opengroup.org/onlinepubs/009604299/functions/dlsym.html, it says

> If handle does not refer to a valid object opened by [dlopen()](https://pubs.opengroup.org/onlinepubs/009604299/functions/dlopen.html), or if the named symbol cannot be found within any of the objects associated with handle, dlsym() shall return NULL.

But then there's the issue of the special values of `handle`, `RTLD_DEFAULT` and `RTLD_NEXT`. I suspect either of these could be defined as `(void*)0`, but in any case the specific value doesn't matter, you just can't use whatever their value is unless you want that behavior.

cpcloud: Seems like from the below code that `RTLD_DEFAULT` is the desired behavior, so maybe that should be returned to ensure that we're not triggering undefined behavior.

mdboom: FWIW, this is reverting to behavior prior to the move to using `pathfinder`, so it's been testing for a long time in the wild with `handle == NULL` in those situations.

But to your question, it's kinda/sorta.  On GNU, the special value `RTLD_DEFAULT` is equal to 0, so it's defined behavior to lookup the symbol in the global namespace -- for example if the library is statically linked.  All that is fine -- we are already doing that before looking in the `.so`.  Other POSIX platforms mileage may vary, of course.

mdboom: Jinx.  Yeah, I agree that maybe setting to RTLD_DEFAULT here would be better.

Windows is another kettle of fish -- something else might be better there.

leofang: Same question for `FreeLibrary` on Windows. Seems to be an undocumented feature? 😛

cpcloud: I believe it would only be defined behavior if the specific value of `RTLD_DEFAULT` is defined. Even if it's effectively guaranteed to never break, returning `RTLD_DEFAULT` is far more readable than what appears to be a `NULL` pointer (even that can technically be something other than `(void*)0` though in practice is almost always is that).

cpcloud: Ok, yeah jinxing all over the place 🤣

cpcloud: ~For Windows it looks like its `dlsym`-equivalent has a similar behavior to `RTLD_DEFAULT` specifically when it is given `NULL`.~

Not sure about this actually.

leofang: > FWIW, this is reverting to behavior prior to the move to using `pathfinder`, so it's been testing for a long time in the wild with `handle == NULL` in those situations.

I was wondering about this too. Turns out we did not test this behavior at all. This is from 12.8.0 (whose behavior we should restore to):
https://github.com/NVIDIA/cuda-python/blob/c04025db0888830cdbb769488f9b1f90ea3eb972/cuda_bindings/cuda/bindings/_internal/nvjitlink_linux.pyx#L55-L65

If a library is not loadable, we just raised an exception in `load_library`, so we never returned the null handle to complete the remaining symbol loading. I think to fix #951 we just need to swallow the exception raised by the pathfinder, and raise the one we used to raise?

mdboom: Ah, ok.  I missed that in all this.  That certainly makes this less weird.  Will change.

mdboom: Hmm...  That can't be the whole story.  A `DynamicLibNotFoundError` is a subclass of `RuntimeError`.  Puzzle to solve...

mdboom:
Closing -- I'm not sure it's fixing the right thing.  See comment: https://github.com/NVIDIA/cuda-python/issues/951#issuecomment-3353566664

Issue #851, opened 2025-08-18
[FEA]: Enhance `load_nvidia_dynamic_lib` error message if a library cannot be found


ZzEeKkAa:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

cuda.pathfinder

### Is your feature request related to a problem? Please describe.

`load_nvidia_dynamic_lib` does great job in looking for packages, runtime error could be confusing for an end user. It would be nice to have user friendly error if library was not found. For example that is how it is done at nvmath-python: https://github.com/NVIDIA/nvmath-python/blob/main/nvmath/_utils.py#L256-L262

Also some helper function (like `runtime_nvidia_library_load_error(library_name:str)` ) would be nice so user can add custom wrapper around `load_nvidia_dynamic_lib`.

### Describe the solution you'd like

n/a

rwgk:
Discussed offline:
* We want to create a troubleshooting page in the a cuda-pathfinder documentation (that still needs to be added: #760)
* Then add a URL pointing to the troubleshooting page from the `RuntimeError` reporting that the requested library could not be found.

**Mock-up** (non-existing URL):

```
cuda.pathfinder._dynamic_libs.load_dl_common.DynamicLibNotFoundError: Failure finding "libnotalib.so"

Please see https://nvidia.github.io/cuda-python/cuda-pathfinder/latest/troubleshooting.html for help.

No such file: libnotalib.so*, No such file: /usr/local/cuda-12.9/lib64/libnotalib.so
...
```

aryanputta:
I would like to work on this, following the plan in the comment above: a `troubleshooting` page in the cuda-pathfinder docs, plus the URL appended to the not-found error messages.

I have this implemented and validated locally (ruff, mypy-pathfinder, and the pathfinder unit tests pass; the new page follows the existing docs layout, so #760 being done unblocks this):

- New `docs/source/troubleshooting.rst`: how to read the "Failure finding" message, the search order (wheels, conda, OS loader, `CUDA_PATH`/`CUDA_HOME`), common causes and fixes, and how to report an issue. Content is derived from the `load_nvidia_dynamic_lib` docstring.
[...]

PR #1055, opened 2025-09-30
`_decide_nvjitlink_or_driver()` bug fix and enhancements [no ci]

### LINKED ISSUE #951: Library bindings are not importable if the underlying library is not installed

rwgk:
`RuntimeError` needs to be caught to resolve #951 

While we're at it:

The rest of the changes are enhancements for clarity (specific error messages) and safety (`ModuleNotFoundError` instead of the less specific `ImportError`).

**[no ci] justification:** CI passed before commit 5ba11be295cd6ab944decbe67f0d5ab19eb8bf7f, which just removed the new test (to be worked on separately, see #1095).

Passing CI: https://github.com/NVIDIA/cuda-python/actions/runs/18325340880?pr=1055


rwgk:
@leofang @mdboom for awareness

This PR passes CI testing, but before marking it as ready for review, I still want to do more local testing.

mdboom (changes requested):
I like this solution, not just for fixing the specific bug, but for providing a more detailed explanation of what is missing.

I think we should add a unit test.  In my earlier PR, I had written one that monkey patches `pathfinder` to always return an import error.  I think with this solution we would want to then call `_decide_nvjitlink_or_driver()` and make sure it returns `True` and doesn't raise.  That new test worked fine locally when I did `python -m pytest cuda_bindings/tests`, but failed when run in CI which does `pytest cuda_bindings/tests`.  It was somehow not loading the correct import paths in the subprocess (probably missing some of the magic that pytest does).  So we'd have to work that out, but I think that test is otherwise a good approach to testing this kind of thing.

cuda_core/cuda/core/experimental/_linker.py
```diff
@@ -37,28 +39,35 @@ def _decide_nvjitlink_or_driver() -> bool:
 
     _driver_ver = handle_return(driver.cuDriverGetVersion())
     _driver_ver = (_driver_ver // 1000, (_driver_ver % 1000) // 10)
+
+    def libname():
+        return "nvJitLink*.dll" if sys.platform == "win32" else "libnvJitLink.so*"
```
leofang: nit: this does not have to be a callable, it can just be `libname = A if condition else B`.

leofang (dismissed):
Thanks, Ralf! I think your analysis is correct.

cuda_core/cuda/core/experimental/_linker.py
```diff
@@ -37,28 +39,35 @@ def _decide_nvjitlink_or_driver() -> bool:
...
+
+    therefore_not_usable = ". Therefore cuda.bindings.nvjitlink is not usable and"
+
     try:
-        from cuda.bindings import nvjitlink as _nvjitlink
+        _nvjitlink = importlib.import_module("cuda.bindings.nvjitlink")
```
cpcloud: This is unnecessary, you have the static name of the thing you're importing, just use the standard import keyword.


```suggestion
        import cuda.bindings.nvjitlink as _nvjitlink
```

cuda_core/cuda/core/experimental/_linker.py
```diff
@@ -37,28 +39,35 @@ def _decide_nvjitlink_or_driver() -> bool:
...
-            "and the culink APIs will be used instead.",
-            stacklevel=3,
-            category=RuntimeWarning,
-        )
-        _driver = driver
-        return True
-    else:
-        return False
+
+    warn(
```
cpcloud: We should always be asking ourselves: "Is this warning actionable by our users?"

If the answer is "no" or "not really", then there's no reason to spam people with "informative" warnings.

This one looks like it is not actionable.

cuda_core/cuda/core/experimental/_linker.py
```diff
@@ -37,28 +39,35 @@ def _decide_nvjitlink_or_driver() -> bool:
 
     _driver_ver = handle_return(driver.cuDriverGetVersion())
     _driver_ver = (_driver_ver // 1000, (_driver_ver % 1000) // 10)
+
+    def libname():
+        return "nvJitLink*.dll" if sys.platform == "win32" else "libnvJitLink.so*"
+
+    therefore_not_usable = ". Therefore cuda.bindings.nvjitlink is not usable and"
```
cpcloud: Does this actually make sense as a variable name? Can we at least make this refer to some kind of noun like `message_piece`? What if we want to change the message? Then we have to change the variable name because it includes `therefore` in it.

cuda_core/cuda/core/experimental/_linker.py
leofang: It is actionable (and more so once we figure out how to address https://github.com/NVIDIA/cuda-python/issues/851) because users can update cuda-bindings and/or nvJitLink.

cpcloud: Assuming it is actionable, the warning needs to contain the action that a user can take instead of just outputting the current state of their installation.

cuda_core/cuda/core/experimental/_linker.py
rwgk: Resolved via commit 6bf421ed0bd3a2460b62c7252d54365bfe7c8fa0

cuda_core/cuda/core/experimental/_linker.py
rwgk: AFAIK this is the idiomatic way to avoid catching `ImportError`, which can mask all kinds of bugs. I don't think there is another way to narrow it down to `ModuleNotFoundError`, which is exactly what we want here.

cuda_core/cuda/core/experimental/_linker.py
rwgk: It'd be much easier if we don't actually want to produce a warning. I was just following the established path.

We could make the errors actionable in this way (for the three situations):

* cuda.bindings.nvjitlink is not available, therefore the culink APIs will be used instead. For best results, please upgrade to a more recent cuda-bindings version.

*  libnvJitLink.so* is not available. Therefore cuda.bindings.nvjitlink is not usable and the culink APIs will be used instead. For best results, please install a recent version of nvJitLink.

* libnvJitLink.so* is too old (<12.3). Therefore cuda.bindings.nvjitlink is not usable and the culink APIs will be used instead. For best results, please install a recent version of nvJitLink.

@leofang which way should I go?

A: Simply remove the warnings (and explain in source code comments instead)?

B: Append the "For best results" sentences?

cuda_core/cuda/core/experimental/_linker.py
rwgk: Resolved via commit 6bf421ed0bd3a2460b62c7252d54365bfe7c8fa0

cuda_core/cuda/core/experimental/_linker.py
cpcloud: I'm not following. A missing module will raise `ModuleNotFoundError` in both the standard import syntax case, and in the `importlib` function you're calling here:

Standard syntax:

```
>>> import cuda.bindings.nvjitlink
  File "<python-input-6>", line 1, in <module>
    import cuda.bindings.nvjitlink
ModuleNotFoundError: No module named 'cuda'
```

`importlib` call:

```
>>> importlib.import_module("cuda.bindings.nvjitlink")
  File "<python-input-8>", line 1, in <module>
    importlib.import_module("cuda.bindings.nvjitlink")
    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/829wb290i87wngxlh404klwxql5v18p4-python3-3.13.7/lib/python3.13/importlib/__init__.py", line 88, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1310, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1310, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1324, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'cuda'
```

The standard import syntax is actually better IMO since it doesn't drown the user in irrelevant import machinery details.

cuda_core/cuda/core/experimental/_linker.py
cpcloud: I'm now thinking based on this discussion that as a user of this I would 100% not care about this warning if my software is working. Why would I want to be warned to upgrade every time I run my program? We should avoid pestering our users with warnings.

cpcloud: One way I find useful to frame the utility of warnings is: if I were in a production on call setting and I saw this warning, would I ignore it, or would I do something?

In that setting, there's tons of information that you have to use to act (usually quickly) and warnings that don't indicate "something bad is about to happen" are an enormous nuisance.

cuda_core/cuda/core/experimental/_linker.py
rwgk: Oh, you're right, thanks!

I had to go back and look how I managed to confuse myself, it's this difference:

```
>>> from cuda.bindings import doesnotexist
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'doesnotexist' from 'cuda.bindings' (/home/rgrossekunst/JunkVenv/lib/python3.12/site-packages/cuda/bindings/__init__.py)
>>> import cuda.bindings.doesnotexist
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'cuda.bindings.doesnotexist'
>>>
```

Changed now: commit 9220d58d22431691151ffeb1ed084df0d12adeac

cuda_core/cuda/core/experimental/_linker.py
rwgk: Ack

I want to defer to @leofang for this, because he has significantly more relevant background.

leofang: The biggest difference is that the driver APIs do not support minor version compatibility or link LTO IRs, unlike nvJitLink which should be the majority of the use cases. It is useful to warn as a hint for debugging potential errors that are raised in later operations due to driver limitations, because the driver error msg is not super clear.

rwgk: Thanks! I worked "do not support minor version compatibility or linking LTO IRs" into the warning messages: commit 7e9fce44c5a2469e5263ec45ae1daa4dd94b37be

I added two more commits, the first tiny one (47247692d06404e8eb58f12f1ec4632bb9bd0122) to make testing easier, then the new tests, almost completely LLM generated (6630d8ebf116cb62da0bc03a62f24f531c0526b1).

cuda_core/cuda/core/experimental/_linker.py
```diff
@@ -37,28 +43,36 @@ def _decide_nvjitlink_or_driver() -> bool:
 
     _driver_ver = handle_return(driver.cuDriverGetVersion())
     _driver_ver = (_driver_ver // 1000, (_driver_ver % 1000) // 10)
+
+    warn_txt_common = (
+        "the culink APIs will be used instead, which do not support"
```
kkraus14: the culink APIs are an implementation detail that the user shouldn't need to think or worry about. We should just refer to the driver library instead of the specific culink APIs.

rwgk:
@leofang @kkraus14 This slightly tangential question keeps coming back to me: When can we declare that we don't support CTK <12.3 anymore?

Because at that point we could just rip out all the fallback (to culink) code completely.

kkraus14:
> @leofang @kkraus14 This slightly tangential question keeps coming back to me: When can we declare that we don't support CTK <12.3 anymore?
> 
> Because at that point we could just rip out all the fallback (to culink) code completely.

When we drop all of CUDA 12.x support, so quite a while away.

leofang:
> > @leofang @kkraus14 This slightly tangential question keeps coming back to me: When can we declare that we don't support CTK <12.3 anymore?
> > Because at that point we could just rip out all the fallback (to culink) code completely.
> 
> When we drop all of CUDA 12.x support, so quite a while away.

We already documented it: https://nvidia.github.io/cuda-python/cuda-bindings/latest/release/12.8.0-notes.html#highlights

cuda_core/cuda/core/experimental/_linker.py
rwgk: Done: commit 09489423db8a31dce8b6431d51a8d09cbee30c98

rwgk:
> I think we should add a unit test.

@mdboom there is a comprehensive new test now (LLM-generated).

I think this PR is ready for merging, but I'll hold off re-running the CI until there is an approval.

cuda_core/tests/test_linker_warnings.py
```diff
@@ -0,0 +1,187 @@
...
+    for modname in list(sys.modules):
+        if modname.startswith("cuda.bindings.nvjitlink") or modname == "cuda.bindings" or modname == "cuda":
+            sys.modules.pop(modname, None)
+
+    yield
+
+    # Cleanup any stubs we added
+    for modname in list(sys.modules):
+        if modname.startswith("cuda.bindings.nvjitlink") or modname == "cuda.bindings" or modname == "cuda":
+            sys.modules.pop(modname, None)
```
leofang: Please let us not meddle with `sys.modules`. If this code is wrong for any reason, it'd be very hard to track down why the Python state becomes inconsistent. 

It seems to me the whole test module is written in mind (ex: `_install_public_nvjitlink_stub` below) that it is run standalone. The import side effects to other test modules are not considered.

Don't try to replace this with `monkeypath.{setitem,delitem}`. I haven't tried it myself but `pytest.monkeypatch` does not seem to be particularly good at scoped tests: https://github.com/pytest-dev/pytest/issues/2229, https://github.com/pytest-dev/pytest/issues/945.

cuda_core/tests/test_linker_warnings.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+def _install_public_nvjitlink_stub():
```
leofang: This test modifies `sys.modules` but does not ensure restoration.

cuda_core/tests/test_linker_warnings.py
mdboom: Maybe it would be possible to run the tests in subprocesses, which would reduce the problem of side effects affecting other tests...

rwgk: The root problem here is the code-to-be-tested. Given that @kkraus14 said we'll need it for a while (I'm assuming until CTK 14 is released), I don't want to spend effort on complex and unusual testing. Probably with a similar effort, the code can be refactored for better practices (no globals), then it'll be easy to test without excessive mocking or setting up subprocesses (which tends to be brittle).

leofang:
Decision: Ship the fix, defer the tests to the next release (https://github.com/NVIDIA/cuda-python/issues/1095)

rwgk:
I merged `main`. I'll run the CI again, with the new `cuda_core/tests/test_linker_warnings.py` still in place, to confirm that it still passes.

Assuming the CI passes, I will remove the new test, then we can merge this PR without running the CI again.

rwgk:
Keeping track of four unexplained failures, JIC, for easy reference:

https://github.com/NVIDIA/cuda-python/actions/runs/18325340880/job/52189171195?pr=1055

* Test linux-aarch64 / py3.13, 12.9.1, wheels, GPU a100
* Test win-64 / py3.13t, 13.0.1, wheels, GPU t4
* Test win-64 / py3.14t, 13.0.1, wheels, GPU t4
* Test win-64 / py3.14, 13.0.1, local, GPU l4

Issue #809, opened 2025-08-06
CI: Make the backport branch generate docs too?


leofang:
The fact that we needed [this PR](https://github.com/NVIDIA/cuda-python/pull/802) was a bit surprising. We weren't prepared to generate docs from the backport branch, so that the latest patch releases of each supported "older" versions
- `cuda-bindings` v11.8.1+
- `cuda-bindings` v12.9.1+

do not generate/upload docs that can be selected from the version dropdown menu. 

I think this is a bit tricky to get right, so we probably just need to accept this fact. But it's still better to capture this in an issue.

rparolin:
Closing as done. Patch-release docs are published and in the version dropdown — `nv-versions.json` lists 12.9.0 through 12.9.7, and `docs/cuda-bindings/12.9.7` on gh-pages has 23 entries. Delivered via `release.yml:192-199` calling `build-docs.yml:61-72`.

The backport branch itself still skips docs deliberately (`12.9.x:.github/workflows/ci.yml:188`, comment "we don't build docs on the backport branch, see #809"), which is the design proposed in the issue body. 11.8.x was never covered and is legacy.

PR #1056, opened 2025-10-01
ci: replace single gating check with vanilla skipping (#1008) [no-ci]

cpcloud:
Backport of #1008

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -168,22 +168,6 @@ jobs:
       host-platform: ${{ matrix.host-platform }}
       build-ctk-ver: ${{ needs.ci-vars.outputs.CUDA_BUILD_VER }}
 
-  doc:
-    name: Docs
-    if: ${{ github.repository_owner == 'nvidia' }}
-    # Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
-    permissions:
-      id-token: write
-      contents: write
-      pull-requests: write
-    needs:
-      - ci-vars
-      - build-linux-64
-    secrets: inherit
-    uses: ./.github/workflows/build-docs.yml
-    with:
-      build-ctk-ver: ${{ needs.ci-vars.outputs.CUDA_BUILD_VER }}
-
   checks:
     name: Check job status
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -32,45 +32,121 @@ jobs:
...
       - ci-vars
-      - build
+      - build-linux-aarch64
     secrets: inherit
     uses: ./.github/workflows/test-wheel-linux.yml
     with:
       build-type: pull-request
       host-platform: ${{ matrix.host-platform }}
       build-ctk-ver: ${{ needs.ci-vars.outputs.CUDA_BUILD_VER }}
-      matrix_filter: "map(select([.CUDA_VER // empty | split(\".\")[] | tonumber] as $v | ($v[0] < 13)))"
```
leofang, line 73: This needs to be kept and applied to all test jobs because we only want to test agains CUDA 12. This was a hack we put in to avoid overhauling the test matrix json file.

.github/workflows/ci.yml
```diff
@@ -84,39 +160,38 @@ jobs:
...
-#      contents: write
-#      pull-requests: write
-#    needs:
-#      - ci-vars
-#      - build
-#    secrets: inherit
-#    uses: ./.github/workflows/build-docs.yml
-#    with:
-#      build-ctk-ver: ${{ needs.ci-vars.outputs.CUDA_BUILD_VER }}
+  doc:
```
leofang: Please disable doc builds because it does not work in the bp branch: https://github.com/NVIDIA/cuda-python/issues/809.

.github/workflows/ci.yml
```diff
@@ -84,39 +160,38 @@ jobs:
...
     needs:
-      - build
-      - test-linux
+      - test-linux-64
+      - test-linux-aarch64
       - test-windows
-#      - doc
-    secrets: inherit
-    uses: ./.github/workflows/status-check.yml
+      - doc
```
leofang: ditto

leofang (changes requested):

.github/workflows/ci.yml
```diff
@@ -93,30 +171,14 @@ jobs:
...
-#      id-token: write
-#      contents: write
-#      pull-requests: write
-#    needs:
-#      - ci-vars
-#      - build
-#    secrets: inherit
-#    uses: ./.github/workflows/build-docs.yml
-#    with:
-#      build-ctk-ver: ${{ needs.ci-vars.outputs.CUDA_BUILD_VER }}
```
cpcloud, line 110: I just deleted this and added a comment. This should lead to fewer conflicts as we change any CI things.

PR #1060, opened 2025-10-01
Add CUDA 13.0 Tests for CuFile I/O Operations

chloechia4:
## Description
This PR includes tests for the low-level bindings and the generated low-level bindings introduced in CUDA 13.0 for CUFile. 

```
test_set_stats_level
test_stats_start
test_stats_stop
test_stats_reset
test_get_stats_l1
test_get_stats_l2
test_get_stats_l3
test_get_bar_size_in_kb
test_set_parameter_posix_pool_slab_array
test_set_get_parameter_size_t
```

CUDA 13.0 CuFile Operations
Note:  The original` test_batch_io_large_operations()` did not pass once switched from CUDA 12.9 to 13.0. I realized it was because the file reads were occurring before the writes as it was submitting all operations (reads and writes) together in one batch. As a result, it was trivially failing because the reads would return as 0 bytes, since they were happening before any write I/O occurred. I changed it to so it would be separated into two phases: writes complete first in one batch handle, and then reads are submitted in another batch handle. This new test works with CUDA 12.9 versioning as well.

### DIFF
--- a/cuda_bindings/cuda/bindings/_internal/cycufile.pxd
+++ b/cuda_bindings/cuda/bindings/_internal/cycufile.pxd
@@ -1,0 +1,370 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
+#
+# This code was automatically generated with version 13.0.0. Do not modify it directly.
+
+from libc.stdint cimport uint32_t, uint64_t
+from libc.time cimport time_t
+from libcpp cimport bool as cpp_bool
+from posix.types cimport off_t
+
+cimport cuda.bindings.cydriver
+from cuda.bindings.cydriver cimport CUresult
+
+
+###############################################################################
+# Types (structs, enums, ...)
+###############################################################################
+
+# TODO: switch to "from libc.time cimport timespec" once we can use recent
+# Cython to build
+cdef extern from "<time.h>":
+    cdef struct timespec:
+        time_t tv_sec
+        long   tv_nsec
+cdef extern from "<sys/socket.h>":
+    cdef struct sockaddr:
+        unsigned short sa_family
+        char sa_data[14]
+    ctypedef sockaddr sockaddr_t
+
+
+cdef extern from '<cufile.h>':
+    # enums
+    ctypedef enum CUfileOpError:
+        CU_FILE_SUCCESS
+        CU_FILE_DRIVER_NOT_INITIALIZED
+        CU_FILE_DRIVER_INVALID_PROPS
+        CU_FILE_DRIVER_UNSUPPORTED_LIMIT
+        CU_FILE_DRIVER_VERSION_MISMATCH
+        CU_FILE_DRIVER_VERSION_READ_ERROR
+        CU_FILE_DRIVER_CLOSING
+        CU_FILE_PLATFORM_NOT_SUPPORTED
+        CU_FILE_IO_NOT_SUPPORTED
+        CU_FILE_DEVICE_NOT_SUPPORTED
+        CU_FILE_NVFS_DRIVER_ERROR
+        CU_FILE_CUDA_DRIVER_ERROR
+        CU_FILE_CUDA_POINTER_INVALID
+        CU_FILE_CUDA_MEMORY_TYPE_INVALID
+        CU_FILE_CUDA_POINTER_RANGE_ERROR
+        CU_FILE_CUDA_CONTEXT_MISMATCH
+        CU_FILE_INVALID_MAPPING_SIZE
+        CU_FILE_INVALID_MAPPING_RANGE
+        CU_FILE_INVALID_FILE_TYPE
+        CU_FILE_INVALID_FILE_OPEN_FLAG
+        CU_FILE_DIO_NOT_SET
+        CU_FILE_INVALID_VALUE
+        CU_FILE_MEMORY_ALREADY_REGISTERED
+        CU_FILE_MEMORY_NOT_REGISTERED
+        CU_FILE_PERMISSION_DENIED
+        CU_FILE_DRIVER_ALREADY_OPEN
+        CU_FILE_HANDLE_NOT_REGISTERED
+        CU_FILE_HANDLE_ALREADY_REGISTERED
+        CU_FILE_DEVICE_NOT_FOUND
+        CU_FILE_INTERNAL_ERROR
+        CU_FILE_GETNEWFD_FAILED
+        CU_FILE_NVFS_SETUP_ERROR
+        CU_FILE_IO_DISABLED
+        CU_FILE_BATCH_SUBMIT_FAILED
+        CU_FILE_GPU_MEMORY_PINNING_FAILED
+        CU_FILE_BATCH_FULL
+        CU_FILE_ASYNC_NOT_SUPPORTED
+        CU_FILE_INTERNAL_BATCH_SETUP_ERROR
+        CU_FILE_INTERNAL_BATCH_SUBMIT_ERROR
+        CU_FILE_INTERNAL_BATCH_GETSTATUS_ERROR
+        CU_FILE_INTERNAL_BATCH_CANCEL_ERROR
+        CU_FILE_NOMEM_ERROR
+        CU_FILE_IO_ERROR
+        CU_FILE_INTERNAL_BUF_REGISTER_ERROR
+        CU_FILE_HASH_OPR_ERROR
+        CU_FILE_INVALID_CONTEXT_ERROR
+        CU_FILE_NVFS_INTERNAL_DRIVER_ERROR
+        CU_FILE_BATCH_NOCOMPAT_ERROR
+        CU_FILE_IO_MAX_ERROR
+
+    ctypedef enum CUfileDriverStatusFlags_t:
+        CU_FILE_LUSTRE_SUPPORTED
+        CU_FILE_WEKAFS_SUPPORTED
+        CU_FILE_NFS_SUPPORTED
+        CU_FILE_GPFS_SUPPORTED
+        CU_FILE_NVME_SUPPORTED
+        CU_FILE_NVMEOF_SUPPORTED
+        CU_FILE_SCSI_SUPPORTED
+        CU_FILE_SCALEFLUX_CSD_SUPPORTED
+        CU_FILE_NVMESH_SUPPORTED
+        CU_FILE_BEEGFS_SUPPORTED
+        CU_FILE_NVME_P2P_SUPPORTED
+        CU_FILE_SCATEFS_SUPPORTED
+
+    ctypedef enum CUfileDriverControlFlags_t:
+        CU_FILE_USE_POLL_MODE
+        CU_FILE_ALLOW_COMPAT_MODE
+
+    ctypedef enum CUfileFeatureFlags_t:
+        CU_FILE_DYN_ROUTING_SUPPORTED
+        CU_FILE_BATCH_IO_SUPPORTED
+        CU_FILE_STREAMS_SUPPORTED
+        CU_FILE_PARALLEL_IO_SUPPORTED
+
+    ctypedef enum CUfileFileHandleType:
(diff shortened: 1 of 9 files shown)


leofang:
Thanks, Chloe! Pinning you internally...

cuda_bindings/cuda/bindings/_internal/cycufile.pxd
leofang, line 1: This file should not exist. Something is wrong when you checked in generated code.

cuda_bindings/cuda/bindings/_internal/cycufile.pyx
leofang, line 1: ditto

cuda_bindings/cuda/bindings/cycufile.pxd
leofang, line 1: @chloechia4 any reason `cuFileDriverClose_v2` is removed? I see this symbol still exists in the cuFILE header. For cuda-bindings, the Cython layer (`cyxxxxx.{pxd,pyx}`) are consider stable public APIs.

cuda_bindings/tests/test_cufile.py
leofang, line 1: @sourabgupta3 have you reviewed the tests?

cuda_bindings/tests/test_cufile.py
```diff
@@ -1511,47 +1512,72 @@ def test_batch_io_large_operations():
...
-        # Get batch status
-        min_nr = num_operations * 2  # Wait for all operations to complete
-        nr_completed = ctypes.c_uint(num_operations * 2)  # Initialize to max operations posted
-        timeout = ctypes.c_int(10000)  # 10 second timeout for large operations
+        # Verify writes succeeded
+        for i in range(nr_completed_writes.value):
+            assert io_events[i].status == cufile.Status.COMPLETE, f"Write {i} failed with status {io_events[i].status}"
 
+        # Force file sync
+        os.fsync(fd)
```
sourabgupta3: This isn't needed.

cuda_bindings/tests/test_cufile.py
```diff
@@ -1885,3 +1908,570 @@ def test_set_get_parameter_string():
...
+        # Allocate buffer for L1 statistics
+        stats_buffer = ctypes.create_string_buffer(1024)  # Allocate sufficient space
+        stats_ptr = ctypes.addressof(stats_buffer)
+
+        # Get L1 statistics (basic operation counts)
+        cufile.get_stats_l1(stats_ptr)
+
+        # Verify that statistics data was written to the buffer
+        # Convert buffer to bytes and check that it's not all zeros
+        buffer_bytes = bytes(stats_buffer)
```
sourabgupta3: Rather than just checking for buffer_bytes, can you verify by looking at actual fields of the data structure(CUfileStatsLevel1_t)?

cuda_bindings/cuda/bindings/cufile.pyx
```diff
@@ -1297,6 +1303,156 @@ cpdef set_parameter_string(int param, intptr_t desc_str):
         __status__ = cuFileSetParameterString(<_StringConfigParameter>param, <const char*>desc_str)
     check_status(__status__)
 
+    
+cpdef get_parameter_min_max_value(int param, intptr_t min_value, intptr_t max_value):
```
sourabgupta3: Add tests for this API as well.

cuda_bindings/tests/test_cufile.py
sourabgupta3: Added few comments now, otherwise looks fine.

cuda_bindings/cuda/bindings/cycufile.pxd
mdboom: @chloechia4: I left a comment on why I think this is happening in your cybind MR.  Addressing that, putting the changes here, and removing those extra files @leofang mentioned should hopefully do it.  I haven't tried testing (I'm on WSL and it looks like cuFile doesn't work there).

cuda_bindings/cuda/bindings/_internal/cycufile.pxd
chloechia4: Removed! I moved the files to the wrong folder after generating

cuda_bindings/cuda/bindings/_internal/cycufile.pyx
chloechia4: Ditto deleted

cuda_bindings/tests/test_cufile.py
leofang: Thanks, Sourab!

cuda_bindings/tests/test_cufile.py
chloechia4: Good point I added field checks in get_stats_l1/get_stats_l2/get_stats_l3. It appears that the Python bindings don't expose the CUfileStatsLevel*_t structures as ctypes classes that we can directly use. So I just added Python equivalent classes

cuda_bindings/cuda/bindings/cufile.pyx
chloechia4: ✅

cuda_bindings/cuda/bindings/cycufile.pxd
chloechia4: Addressed this

cuda_bindings/tests/test_cufile.py
chloechia4: ✅

chloechia4:
/ok to test [b307a66](https://github.com/NVIDIA/cuda-python/pull/1060/commits/b307a6612757c46fc094bd94cffe7cda99a3d729)

cuda_bindings/tests/test_cufile.py
```diff
@@ -1886,3 +1905,632 @@ def test_set_get_parameter_string():
...
+        except Exception as e:
+            logging.info(f"Correctly caught error for invalid stats level: {e}")
+
+        try:
+            assert cufile.set_stats_level(4)  # Invalid level > 3
+        except Exception as e:
+            logging.info(f"Correctly caught error for invalid stats level: {e}")
+
+        # Reset to level 0 (disabled) for cleanup
+        cufile.set_stats_level(0)
```
leofang: Q: if `0` is the default, shouldn't this line be part of `finally:` block below, to ensure cleanup regardless of success or failure?

cuda_bindings/tests/test_cufile.py
```diff
@@ -1886,3 +1905,632 @@ def test_set_get_parameter_string():
...
+    assert err == cuda.CUresult.CUDA_SUCCESS
+    (err,) = cuda.cuCtxSetCurrent(ctx)
+    assert err == cuda.CUresult.CUDA_SUCCESS
+
+    # Open cuFile driver
+    cufile.driver_open()
+
+    try:
+        # Set statistics level first (required before starting stats)
+        cufile.set_stats_level(1)  # Level 1 = basic statistics
```
leofang: Q: We don't reset stats when the test is done?

cuda_bindings/tests/test_cufile.py
```diff
@@ -1886,3 +1905,632 @@ def test_set_get_parameter_string():
...
```
leofang, line 1914: ditto, need to reset and avoid introducing side effects in any tests

cuda_bindings/tests/test_cufile.py
```diff
@@ -1886,3 +1905,632 @@ def test_set_get_parameter_string():
...
+
+    finally:
+        cufile.driver_close()
+        cuda.cuDevicePrimaryCtxRelease(device)
+
+
+@pytest.mark.skipif(
+    cufileVersionLessThan(1150), reason="cuFile parameter APIs require cuFile library version 13.0 or later"
+)
+def test_stats_start():
```
leofang: Looking at this again, this test should be removed. We should always test start-stop in pairs, not individually. This test technically introduces a side-effect (that stat collection does not stop after the test ends).

leofang: (and you already did the paired test below, just rename it to `test_stats_start_stop`)

cuda_bindings/tests/test_cufile.py
```diff
@@ -1886,3 +1905,632 @@ def test_set_get_parameter_string():
...
+
+    try:
+        # Set statistics level first (required before starting stats)
+        cufile.set_stats_level(1)  # Level 1 = basic statistics
+        # Start collecting cuFile statistics first
+
+        cufile.stats_start()
+
+        # Reset cuFile statistics to clear all counters
+        cufile.stats_reset()
```
leofang, line 1926: Q: Does this call also reset the stats level? If not, we should do it in `finally`

cuda_bindings/tests/test_cufile.py
chloechia4: Ah okay renamed

cuda_bindings/tests/test_cufile.py
chloechia4: ✅

cuda_bindings/tests/test_cufile.py
chloechia4: Added in finally:

cuda_bindings/tests/test_cufile.py
chloechia4: Added in finally:

leofang:
/ok to test [e1eabf8](https://github.com/NVIDIA/cuda-python/pull/1060/commits/e1eabf82927de4661b84ec9fae19076a5b2a44d7)

cuda_bindings/tests/test_cufile.py
leofang: This is still unresolved.

cuda_bindings/tests/test_cufile.py
```diff
@@ -1624,104 +1644,568 @@ def test_set_get_parameter_size_t():
...
+    try:
+        # Test with poll threshold parameter
+        param = cufile.SizeTConfigParameter.POLLTHRESHOLD_SIZE_KB
+
+        # Allocate ctypes variables for min and max values
+        min_value = ctypes.c_size_t()
+        max_value = ctypes.c_size_t()
+
+        # Get min/max values
+        cufile.get_parameter_min_max_value(param, ctypes.addressof(min_value), ctypes.addressof(max_value))
```
leofang: I don't like this binding. We need to tweak the codegen to return a 2-tuple `(min, max)`. Please work with @mdboom to update the codegen config.

cuda_bindings/tests/test_cufile.py
leofang: @chloechia4 Please remove this test entirely. I don't see any change?

cuda_bindings/tests/test_cufile.py
leofang: Where...? @chloechia4 if you made the changes locally, make sure you push it to remote.

cuda_bindings/tests/test_cufile.py
leofang: this has not happened yet

cuda_bindings/tests/test_cufile.py
leofang: ditto

cuda_bindings/tests/test_cufile.py
```diff
@@ -1624,104 +1644,568 @@ def test_set_get_parameter_size_t():
...
+
+    # Open cuFile driver
+    cufile.driver_open()
+
+    # Create test file directly with O_DIRECT
+    file_path = "test_stats_l2.bin"
+    fd = os.open(file_path, os.O_CREAT | os.O_RDWR | os.O_DIRECT, 0o600)
+
+    try:
+        cufile.set_stats_level(2)  # L2 = detailed performance metrics
```
leofang, line 2060: this needs to be restored

leofang (changes requested):

cuda_bindings/tests/test_cufile.py
chloechia4: Do you mean reset(). Adding a stats_reset() in "finally" section

cuda_bindings/tests/test_cufile.py
chloechia4: Yes my bad, we actually do check for level 0 in the for loop above, instead i just added stats_reset() in the finally block

leofang (approved):

PR #1067, opened 2025-10-02
Detect __pyx_capi__ changes in testing

### LINKED ISSUE #1030: CI: Add integration tests to cover `cuda.bindings` ABI stability

mdboom:
As a way of detecting changes in "Cython ABI" (which is loosely defined, see #1030), this adds the contents of the `__pyx_capi__` dictionary to JSON files in the repository, and then adds a unit test to detect any changes.

I think this may be useful as a first defense against ABI changes, however it's not a complete solution.  This can not detect changes in structs or enums used across Cython modules.

There is also the risk of annoyance any time new functions are added, which is a completely "ABI safe" operation, but we need to update the baselines in order to detect unsafe operations (deletions) in the future.  An alternative approach might be to download a baseline wheel and compare the ABI against it directly -- that could be done in CI but would be harder to run locally.

Closes #1030

### DIFF
--- a/cuda_bindings/tests/test_cython_abi.py
+++ b/cuda_bindings/tests/test_cython_abi.py
@@ -48,7 +48,7 @@ def pyx_capi_to_json(d: dict[str, object]) -> dict[str, str]:
 
     def extract_name(v: object) -> str:
         v = str(v)
-        match = re.match(r'<capsule object "([^\"]+)" at 0x[0-9a-f]+>', v)
+        match = re.match(r'<capsule object "([^\"]+)" at 0x[0-9a-fA-F]+>', v)
         assert match, f"Could not parse __pyx_capi__ entry: {v}"
         return match.group(1)
 


cpcloud:
Would this have caught the issues mentioned in #1030?

- https://github.com/conda-forge/cuda-python-feedstock/issues/15
- https://github.com/NVIDIA/cuda-python/issues/274

kkraus14:
> Would this have caught the issues mentioned in #1030?
> 
> * [11.7.1 had a huge ABI break conda-forge/cuda-python-feedstock#15](https://github.com/conda-forge/cuda-python-feedstock/issues/15)
> * [Building using versions of cuda-python with the new layout breaks runtime compatibility with older versions #274](https://github.com/NVIDIA/cuda-python/issues/274)

Yes, symbols would have been removed from `__pyx_capi__` when they were changed to be externed.

mdboom:
This theoretically would detect the change as described [here](https://github.com/conda-forge/cuda-python-feedstock/issues/15).  It's basically reverse engineering what Cython does when it does its custom "dynamic linking".  I would have liked to have tested it directly, but it's surprisingly hard to build 11.7 today, and there don't seem to be any wheels or conda packages available.

mdboom:
It occurred to me that the main downside of this approach (that you have to update the `.abi.json` files even when it intentionally changes) is made worse by the fact that we release from main.  If we moved to a model where we had a branch even for "current" (e.g. 13.x) releases, we would only need to do ABI checks there -- when we merge main into the branch or backport bugfixes -- and we could leave this off on main.  That would match how this similar thing is done in CPython.

But it's a larger discussion to have to move to a different branch model, so maybe we table this for now?  (I suspect @cpcloud has opinions about branching models, etc.)

cpcloud:
This seems to have stalled. Should we close it and accept that we won't be able to catch C ABI breakage in all cases and that we're prioritizing ease of development in this case?

mdboom:
My opinion is that for a lot of good reasons we shouldn't be releasing from main.  Once we have made that change, it would make sense to come back to this and only activate the test on maintenance branches.  I'll close for now.

mdboom:
Relevant: https://github.com/NVIDIA/cuda-python/issues/1199

cpcloud:
FWIW I don't have too many opinions on branch releases, having not worked on many teams that have released software that way. Definitely open to alternative models there!

leofang:
Let me reopen this
- @oleksandr-pavlyk had done some research that might help inform how we can detect this
- we haven't had allocated enough time for @mdboom to walk through this PR in the team sync, and we should do so when Mike is available (probably not this week?)

PR #1068, opened 2025-10-02
chore(dev-tools): only check specific files when checking for markdown in docs/source [no-ci]

cpcloud:
This speeds up the markdown file disallowance when running pre-commit. This is noticeably slower when searching through hidden `.pixi` directories because it scans every single file in the directory. We only need to check that new markdown files are not introduced, so we only need to look at passed filenames.


cpcloud:
- Verified failure by adding a new file in `cuda_python/docs/source`

PR #1069, opened 2025-10-02
refactor(pythoN): drop support for 3.9, document 3.14 support

cpcloud:
This PR removes support for Python 3.9 and fills various documentation bits adding support for 3.14. We can wait to merge this after the next release.


cuda_pathfinder/cuda/pathfinder/_utils/find_site_packages_dll.py
```diff
@@ -11,7 +11,12 @@ def find_all_dll_files_via_metadata() -> dict[str, tuple[str, ...]]:
     results: collections.defaultdict[str, list[str]] = collections.defaultdict(list)
 
     # sort dists for deterministic output
-    for dist in sorted(importlib.metadata.distributions(), key=lambda d: (d.metadata.get("Name", ""), d.version)):
+
+    for dist in sorted(
+        importlib.metadata.distributions(),
+        # `get` exists before 3.12, even though the hints only exist for Python >=3.12
+        key=lambda d: (d.metadata.get("Name", ""), d.version),  # type: ignore[attr-defined]
```
cpcloud, line 18: No code was changed here, I just reformatted it so that the `type: ignore` bit didn't apply to entire call.

We really should try harder to consolidate this and `dll` equivalent, there's a ton of repetition between the two modules.

cuda_python/pyproject.toml
```diff
@@ -22,16 +22,18 @@ classifiers = [
...
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
     "Programming Language :: Python :: Implementation :: CPython",
     "Environment :: GPU :: NVIDIA CUDA",
     "Environment :: GPU :: NVIDIA CUDA :: 12",
+    "Environment :: GPU :: NVIDIA CUDA :: 13",
 ]
 dynamic = ["version", "dependencies", "optional-dependencies"]
+requires-python = ">=3.10"
```
leofang, line 36: Q: Does this work for noarch packages?

cpcloud: What do you mean by "work"?

cpcloud: I believe `noarch` is a conda concept and not related to `pyproject.toml`, so I'm not sure what that means in this context.

Many packages that do not have any platform-specific builds set this to inform the various solvers that they only work with certain versions of Python.

I am assuming that users do things like `pip install cuda-python[cu12]` and that this field would have an effect if someone tried to run that command for Python 3.9 (after this PR).

leofang: > I believe `noarch` is a conda concept and not related to `pyproject.toml`, so I'm not sure what that means in this context.

Sorry for being imprecise, but with everything in flux I hope we'd agree this is an obvious term so that we can save a bit of typing per PR 🙂 

I meant pure Python wheel which is a well-defined term as per [PEP 425](https://peps.python.org/pep-0425/). See also [this section](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#pure-python-wheels) of the Python Packaging Guide. The exact wording differs across packaging managers, but the concept is the same.

Since I have this page opened, I see the spec does not care if it's a pure Python or platform wheel
https://packaging.python.org/en/latest/specifications/pyproject-toml/#requires-python
so I believe you're right on this. 

I see you've applied this change to all projects. LGTM! Resolving.

cpcloud:
Since 3.14 is officially out as of yesterday, this PR is ready for review.

kkraus14:
I think the `cuda.bindings` changes here should be backported to `12.9.x` as well.

leofang:
Reverted in #1109. To be re-reverted in the next release to actually drop 3.9.

Issue #1063, opened 2025-10-01
[FEA]: Update the memory module to use __dealloc__.


Andy-Jost:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

cuda.core

### Is your feature request related to a problem? Please describe.

I would like abnormal process shutdown to complete gracefully. When testing memory IPC, it was observed that a killed child process sometimes appears to invoke `Buffer.__del__` after the CUDA context has been destroyed (see additional context section for error message). This has been observed to result in hard errors including floating point exceptions and segmentation violations.

### Describe the solution you'd like

Update `Buffer.__del__` (and similar methods) to use `Buffer.__dealloc__`.

### Describe alternatives you've considered

Solutions based on `sys.is_finalizing` appear to be unreliable.

### Additional context

The following messages were output from a killed child process while running `cuda_core/tests/memory_ipc` tests:

```
	/opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/site-packages/_pytest/unraisableexception.py:67: PytestUnraisableExceptionWarning: Exception ignored in: 'cuda.core.experimental._memory.Buffer.__del__' 
	
	Traceback (most recent call last): 
	File "/opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/site-packages/cuda/core/experimental/_device.py", line 991, in __new__ 
	devices = _tls.devices 
	^^^^^^^^^^^^ 
	AttributeError: '_thread._local' object has no attribute 'devices' 
	
	During handling of the above exception, another exception occurred: 
	
	Traceback (most recent call last): 
	File "cuda/core/experimental/_memory.pyx", line 82, in cuda.core.experimental._memory.Buffer._shutdown_safe_close 
	self._mr.deallocate(self._ptr, self._size, stream) 
	File "cuda/core/experimental/_memory.pyx", line 801, in cuda.core.experimental._memory.DeviceMemoryResource.deallocate 
	raise_if_driver_error(err) 
	File "cuda/core/experimental/_utils/cuda_utils.pyx", line 69, in cuda.core.experimental._utils.cuda_utils._check_driver_error 
	raise CUDAError(f"{name}: {expl}") 
[...]

Issue #803, opened 2025-08-05
[BUG]: `cuda.core` Event failed to complete cleanup properly


yangcal:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Runtime Error

### Component

cuda.core

### Describe the bug

In our project, one of the custom object holds a reference to the `cuda.core.experimental.Event` and the following occurred during resource cleanup stage, e.,g 

> [2025-07-31T06:01:25.076Z] Traceback (most recent call last):
> [2025-07-31T06:01:25.076Z]   File "cuda/core/experimental/_event.pyx", line 114, in cuda.core.experimental._event.Event.close
> [2025-07-31T06:01:25.076Z] AttributeError: 'NoneType' object has no attribute 'cuEventDestroy'
> [2025-07-31T06:01:25.076Z] Exception ignored in: 'cuda.core.experimental._event.Event.__del__'
> [2025-07-31T06:01:25.076Z] Traceback (most recent call last):
> [2025-07-31T06:01:25.076Z]   File "cuda/core/experimental/_event.pyx", line 114, in cuda.core.experimental._event.Event.close
> [2025-07-31T06:01:25.076Z] AttributeError: 'NoneType' object has no attribute 'cuEventDestroy'
> [2025-07-31T06:01:25.076Z] Traceback (most recent call last):
> [2025-07-31T06:01:25.076Z]   File "cuda/core/experimental/_event.pyx", line 114, in cuda.core.experimental._event.Event.close
> [2025-07-31T06:01:25.076Z] AttributeError: 'NoneType' object has no attribute 'cuEventDestroy'
> [2025-07-31T06:01:25.076Z] Exception ignored in: 'cuda.core.experimental._event.Event.__del__'
> [2025-07-31T06:01:25.076Z] Traceback (most recent call last):
> [2025-07-31T06:01:25.076Z]   File "cuda/core/experimental/_event.pyx", line 114, in cuda.core.experimental._event.Event.close
> [2025-07-31T06:01:25.076Z] AttributeError: 'NoneType' object has no attribute 'cuEventDestroy'

On our side, we had to explicitly release the reference to the Event object and then the issue was resolved, e.g,

```python
def free(self):
     self.event = None # without this, the error occurred in an MPI simulation
```

### How to Reproduce

[...]

PR #1070, opened 2025-10-02
Cythonize `cuda.core` more

### LINKED ISSUE #1065: cuda-core: Release GIL when calling `cimport`'d CUDA APIs

leofang:
## Description

Close #1065 

## Checklist


leofang:
(Need #1041 to be merged first.)

leofang:
/ok to test [b713c5c](https://github.com/NVIDIA/cuda-python/pull/1070/commits/b713c5c4160324e401c0587e244f0f94097e182a)

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -168,46 +165,37 @@ cdef class Stream:
...
-            return
-
-        if self._owner is None:
-            if self._handle and not self._builtin:
-                HANDLE_RETURN(cydriver.cuStreamDestroy(self._handle))
-        else:
-            self._owner = None
-        self._handle = <cydriver.CUstream>(NULL)
+    def __dealloc__(self):
+        self.close()
```
leofang, line 180: Note: With @brandon-b-miller's reproducer (https://github.com/NVIDIA/cuda-python/issues/803#issuecomment-3292781448), I confirm this change continues to avoid shutdown issues as expected, although it'd be nice to have more ways to test this...

cpcloud: I agree.

An explanation for why this is solving the problem is also important to have. So far I haven't seen anything showing we understand why this works and other solutions don't.

leofang:
/ok to test [bf712bf](https://github.com/NVIDIA/cuda-python/pull/1070/commits/bf712bf6539d1d4eba33de16177739fb4194bfd8)

cuda_core/cuda/core/experimental/_stream.pyx
leofang: As discussed offline, I left a brief note on this: https://github.com/NVIDIA/cuda-python/issues/1063#issuecomment-3367892470

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -109,20 +108,15 @@ cdef class Stream:
...
-        object _nonblocking
-        object _priority
-        object _device_id
-        object _ctx_handle
-
     def __cinit__(self):
         self._handle = <cydriver.CUstream>(NULL)
+        self._owner = None
+        self._builtin = False
+        self._nonblocking = -1  # delayed
```
cpcloud: What does `delayed` mean here?

If it means something specific to the problem domain, then that needs additional commentary.

If it means "the definition is delayed", then the comment just adds confusion, so you can leave it out.

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -187,12 +176,14 @@ cdef class Event:
...
         """
-        HANDLE_RETURN(cydriver.cuEventSynchronize(self._handle))
+        with nogil:
+            HANDLE_RETURN(cydriver.cuEventSynchronize(self._handle))
 
     @property
     def is_done(self) -> bool:
         """Return True if all captured works have been completed, otherwise False."""
-        result = cydriver.cuEventQuery(self._handle)
+        with nogil:
```
cpcloud, line 185: I see a lot of new `nogil` sections. Is that for any particular reason?

rparolin:
@leofang Please update the PR description to include a description of the solution you are proposing.

cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
```diff
@@ -191,20 +192,23 @@ def precondition(checker: Callable[..., None], str what="") -> Callable:
     return outer
 
 
-def get_device_from_ctx(ctx_handle) -> int:
+cdef cydriver.CUdevice get_device_from_ctx(
+        cydriver.CUcontext target_ctx, cydriver.CUcontext curr_ctx) except?cydriver.CU_DEVICE_INVALID nogil:
```
Andy-Jost, line 196: When I see this level of complexity in Cython, approaching that of C++, I wonder whether it would be simpler to directly write C++ and expose it through something like pybind11. Not sure if I'm the only one who thinks this.

mdboom: You are not alone in that ;)  It's also possible to write C++ and just call it from Cython -- we have a couple small `.cpp` files in this repo already.

cuda_core/cuda/core/experimental/_device.pyx
```diff
@@ -1042,16 +1050,18 @@ class Device:
...
         cdef cydriver.CUdevice dev
-        if check_consistency:
-            HANDLE_RETURN(cydriver.cuCtxGetDevice(&dev))
-            if <int>(dev) != self._id:
-                raise CUDAError("Internal error (current device is not equal to Device.device_id)")
+        cdef cydriver.CUdevice this_dev = self._id
+        with nogil:
+            HANDLE_RETURN(cydriver.cuCtxGetCurrent(&ctx))
+            if ctx == NULL:
+                raise CUDAError("No context is bound to the calling CPU thread.")
```
oleksandr-pavlyk, line 1023: Is it OK to raise from within nogil section?

cuda_core/cuda/core/experimental/_device.pyx
```diff
@@ -985,19 +989,21 @@ class Device:
     def __new__(cls, device_id: Optional[int] = None):
         global _is_cuInit
         if _is_cuInit is False:
-            with _lock:
+            with _lock, nogil:
```
leofang, line 955: @shwina said we don't need this. We only want one thread to initialize CUDA.
```suggestion
            with _lock:
```

cuda_core/cuda/core/experimental/_device.pyx
leofang: Yes because in this and a few other limited situations Cython will ensure GIL is held again before raising. It makes the code a lot cleaner and avoid nested `nogil` blocks.

cuda_core/cuda/core/experimental/_event.pyx
leofang: This is explained in both #1065 (in particular, this https://github.com/NVIDIA/cuda-python/issues/1065#issuecomment-3363940142) as well as the offline PR review meeting.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -99,7 +95,18 @@ cdef class Buffer:
...
             the behavior depends on the underlying memory resource.
         """
-        self._shutdown_safe_close(stream, is_shutting_down=None)
+        if self._ptr and self._mr is not None:
+            if isinstance(self._mr, _cyMemoryResource):
+                # FIXME
+                if stream is None:
+                    stream = Stream.__new__(Stream)
+                    (<cyStream>(stream))._handle = <cydriver.CUstream>(0)
+                (<_cyMemoryResource>(self._mr))._deallocate(self._ptr, self._size, <cyStream>stream)
```
shwina: Is the large amount of casting here essentially because you can't `cdef` declare in this scope?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -99,7 +95,18 @@ cdef class Buffer:
             The stream object to use for asynchronous deallocation. If None,
             the behavior depends on the underlying memory resource.
         """
-        self._shutdown_safe_close(stream, is_shutting_down=None)
+        if self._ptr and self._mr is not None:
+            if isinstance(self._mr, _cyMemoryResource):
```
shwina: Can we implement `__reduce__` on `cyMemoryResource` and avoid the `isinstance` check here?

cuda_core/cuda/core/experimental/_device.pyx
kkraus14: Why wouldn't we want to release the GIL though? Other Python code completely unrelated to any CUDA APIs could happily run.

cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
```diff
@@ -52,32 +52,33 @@ def _reduce_3_tuple(t: tuple):
...
 
 
-cdef object _DRIVER_SUCCESS = driver.CUresult.CUDA_SUCCESS
 cdef object _RUNTIME_SUCCESS = runtime.cudaError_t.cudaSuccess
 cdef object _NVRTC_SUCCESS = nvrtc.nvrtcResult.NVRTC_SUCCESS
 
 
-cpdef inline int _check_driver_error(error) except?-1:
-    if error == _DRIVER_SUCCESS:
+cpdef inline int _check_driver_error(cydriver.CUresult error) except?-1 nogil:
```
kkraus14, line 65: Do we not want an explicit `with nogil:` inside of this function because it's expected to be inlined within existing `with nogil:` blocks? What is the behavior if we have an explicit `with nogil` anyway?

cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
```diff
@@ -52,32 +52,33 @@ def _reduce_3_tuple(t: tuple):
     return t[0] * t[1] * t[2]
 
 
-cdef int HANDLE_RETURN(supported_error_type err) except?-1:
+cdef int HANDLE_RETURN(supported_error_type err) except?-1 nogil:
```
kkraus14, line 55: Should we explicitly release the GIL in this function?

cuda_core/cuda/core/experimental/_device.pyx
```diff
@@ -27,6 +27,8 @@ from cuda.core.experimental._utils.cuda_utils import (
 )
 
 
+# TODO: I prefer to type these as "cdef object" and avoid accessing them from within Python,
+# but it seems it is very convenient to expose them for testing purposes...
 _tls = threading.local()
 _lock = threading.Lock()
```
kkraus14, line 33: We could `cdef object` these and then create private functions to return them for usage in testing, but we could do that in a follow up.

cuda_core/cuda/core/experimental/_device.pyx
```diff
@@ -1095,7 +1107,10 @@ class Device:
         # Use 256 characters to be consistent with CUDA Runtime
         cdef int LENGTH = 256
         cdef bytes name = bytes(LENGTH)
-        HANDLE_RETURN(cydriver.cuDeviceGetName(<char*>name, LENGTH, self._id))
+        cdef char* name_ptr = name
+        cdef cydriver.CUdevice this_dev = self._id
+        with nogil:
+            HANDLE_RETURN(cydriver.cuDeviceGetName(name_ptr, LENGTH, this_dev))
         name = name.split(b"\0")[0]
         return name.decode()
```
kkraus14, line 1078: nitpick: there's probably a way to do this without the GIL?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -66,23 +71,14 @@ cdef class Buffer:
     @classmethod
     def _init(cls, ptr: DevicePointerT, size_t size, mr: MemoryResource | None = None):
         cdef Buffer self = Buffer.__new__(cls)
-        self._ptr = <uintptr_t>(int(ptr))
```
rparolin, line 69: This looks incorrect to me.

- uintptr_t vs intptr_t - intptr_t has a reduced range of representative pointer value due to the signed bit
- int(ptr) - explicitly narrowing conversion to a 32-bit python int before promoting to a 64-bit data type. We are throwing away data in this conversion.

- uintptr_t  =           0 to 2^64 − 1
- intptr_t     = −2^63 to 2^63 − 1

rparolin (changes requested):
I left a comment asking about pointer data type changes and pointer value conversions.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -110,7 +117,13 @@ cdef class Buffer:
...
             handle, call ``int(Buffer.handle)``.
         """
-        return self._ptr_obj
+        if self._ptr_obj is not None:
+            return self._ptr_obj
+        elif self._ptr:
+            return self._ptr
+        else:
+            # contract: Buffer is closed
+            return None
```
kkraus14: Should we consider returning `NULL` instead of `None` here?

cuda_core/cuda/core/experimental/_device.pyx
leofang: > Since Cython 3.0, some simple Python statements can be used inside of nogil sections: raise, assert and print (the Py2 statement, not the function). Since they tend to be lone Python statements, Cython will automatically acquire and release the GIL around them for convenience.

https://cython.readthedocs.io/en/latest/src/userguide/external_C_code.html#releasing-the-gil

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -99,7 +95,18 @@ cdef class Buffer:
             The stream object to use for asynchronous deallocation. If None,
             the behavior depends on the underlying memory resource.
         """
-        self._shutdown_safe_close(stream, is_shutting_down=None)
+        if self._ptr and self._mr is not None:
+            if isinstance(self._mr, _cyMemoryResource):
+                # FIXME
+                if stream is None:
+                    stream = Stream.__new__(Stream)
+                    (<cyStream>(stream))._handle = <cydriver.CUstream>(0)
```
kkraus14: Do we want to address this before merging? If not, let's mark it as `#TODO` to be consistent with how we're commenting these things?

cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
rparolin: I've also been thinking this.  Basically answering this question, where is the point on the trade-off curve of using a new DSL that generates C++ code is it easier to just writing the C++ manually?

cuda_core/cuda/core/experimental/_memory.pyx
leofang: It is mostly because I am nervous about casting upfront, but we can probably simplify this. Let me push something.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -816,13 +856,19 @@ class DeviceMemoryResource(MemoryResource):
         -------
             The shareable handle for the memory pool.
         """
+        # Note: This is Linux only (int for file descriptor)
+        cdef int alloc_handle
```
kkraus14, line 860: Should we consider putting this and the below function call behind a compilation conditional?

cuda_core/cuda/core/experimental/_memory.pyx
cpcloud: I think there's a weird Cython-ism here which is that since we're in a `def` method, `int(x)` will convert to a Python `int` which is actually unlimited precision, with the performance of `int64_t` for the first 64 bits of that precision, so I don't believe this is a narrowing conversion (the cast to `intptr_t` notwithstanding).

cuda_core/cuda/core/experimental/_device.pyx
shwina: When the GIL is held and you call `with _lock`, the behaviour is clear: the thread holding the GIL acquires the lock. Here,` it's unclear and seems racy.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: With this PR, we have two kinds of memory resource implementations:
1. A pure Python one that inherits the `MemoryResource` ABC:
   - In this case, we need to ensure the Python `deallocate()` is invoked 
2. A `cuda.core`-provided one that inherits both `_cyMemoryResource` (to get a C layout & methods) and `MemoryResource` (to meet Python requirements)
   - In this case, we can call the Cython `_dealloc()`

It is unclear to me when I worked on this that if `isinstance` can be replaced by a better approach, but since it is only a pointer comparison, I assume it is OK if we keep it.

cuda_core/cuda/core/experimental/_stream.pyx
kkraus14: I think this means that these are lazily initialized, in which case `lazy` is probably a better word to describe these attributes?

cuda_core/cuda/core/experimental/_device.pyx
shwina: Perhaps we should be explicit?

```
with lock:
    with nogil:
        ...
```

cuda_core/cuda/core/experimental/_memory.pyx
leofang: This is a (arguably terrible) Cython hack to ensure correctness & backward compatibility 😓 

This is one of the dark corners of Cython: `int` represents both Python `int` and C `int`, and one has to read from the context to understand which one it means. In this case, since `ptr` is a generic Python object, `int(ptr)` invokes the CPython C API to convert it to a Python `int`, which is arbitrary length, from which we convert to `uintptr_t` through the `<uintptr_t>(...)` casting via another C API. Another case where Python `int` is guaranteed is the `isinstance(obj, int)` check (which does not make sense if it were C int).

cuda_core/cuda/core/experimental/_memory.pyx
shwina: If you have `__reduce__` implemented in `cyMemoryResource`, then wouldn't classes like `DeviceMemoryResource` automatically use that one without the need for this `isinstance` check?

`isinstance` checks are notoriously expensive when `abc`s are involved, but also it seems cleaner for `cyMemoryResource` to define its own `__reduce__`.

cuda_core/cuda/core/experimental/_device.pyx
```diff
@@ -1095,7 +1107,10 @@ class Device:
         # Use 256 characters to be consistent with CUDA Runtime
         cdef int LENGTH = 256
         cdef bytes name = bytes(LENGTH)
-        HANDLE_RETURN(cydriver.cuDeviceGetName(<char*>name, LENGTH, self._id))
+        cdef char* name_ptr = name
+        cdef cydriver.CUdevice this_dev = self._id
+        with nogil:
```
rparolin, line 1075: Nit: I'd expect that nogil annotations would be used at the parent scope (to the cython function) b/c we wanted to batch more functions into the nogil scope.  Instead of wrapping a single function with nogil. Is there a feature in Cython where you can turn off the gil in the wrapping function so you don't need to push the annotation into the parent scope?

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -338,29 +317,31 @@ cdef class Stream:
...
 
-    cdef int _get_device_and_context(Stream self) except?-1:
-        if self._device_id is None:
-            # Get the stream context first
-            self._get_context()
-            self._device_id = get_device_from_ctx(self._ctx_handle)
+    cdef int _get_device_and_context(self) except?-1:
+        cdef cydriver.CUcontext curr_ctx
+        if self._device_id == cydriver.CU_DEVICE_INVALID:
+            # TODO: It is likely faster/safer to call cuCtxGetCurrent?
```
kkraus14: I don't think this is universally true, where a context can be non-current without being destroyed. Retrieving the device and context from that stream object would be valid.

cuda_core/cuda/core/experimental/_stream.pyx
leofang: It was marked as "delayed" since cuda.core Day 1 because a `Stream` object could be wrapping all possibilities and we don't know which one it is at the allocation time (`__cinit__`/`tp_new`):
- A named built-in stream (legacy/per-thread)
- A stream allocated later by cuda.core
- A stream allocated elsewhere that implements `__cuda_stream__` protocol 

so in general we don't know, nor do we need, most of the attributes until they are actually needed, hence "delayed".

cuda_core/cuda/core/experimental/_memory.pyx
leofang: (when in doubt, the best, or probably the only, way to check is to examine the generated .cpp code)

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -617,97 +645,109 @@ class DeviceMemoryResource(MemoryResource):
...
+        self._alloc_handle = None
 
     def __init__(self, device_id: int | Device, options=None):
-        device_id = getattr(device_id, 'device_id', device_id)
+        cdef int dev_id = getattr(device_id, 'device_id', device_id)
         opts = check_or_create_options(
             DeviceMemoryResourceOptions, options, "DeviceMemoryResource options", keep_none=True
         )
+        cdef cydriver.cuuint64_t current_threshold
+        cdef cydriver.cuuint64_t max_threshold = 0xFFFFFFFFFFFFFFFF
```
rparolin: Nit: Does cython support the equivalent of a UINT_MAX constant?  Its error prone to write out the max values in hex.  Or in this case, the CUDA library has to define/export that constant as part of its public facing API?

cuda_core/cuda/core/experimental/_device.pyx
kkraus14: annotating a function with `nogil` is different than using `with nogil`:
* annotating a function with `nogil` indicates the function is callable without holding the GIL, but doesn't instruct the interpreter to release the GIL if it is held
* an explicit `with nogil:` block instructs the interpreter to release the GIL and indicates to the Cython compiler that everything within the block needs to be callable without the GIL

cuda_core/cuda/core/experimental/_device.pyx
```diff
@@ -1080,10 +1090,12 @@ class Device:
 
         """
         cdef cydriver.CUuuid uuid
-        IF CUDA_CORE_BUILD_MAJOR == "12":
-            HANDLE_RETURN(cydriver.cuDeviceGetUuid_v2(&uuid, self._id))
-        ELSE:  # 13.0+
-            HANDLE_RETURN(cydriver.cuDeviceGetUuid(&uuid, self._id))
+        cdef cydriver.CUdevice this_dev = self._id
+        with nogil:
+            IF CUDA_CORE_BUILD_MAJOR == "12":
```
oleksandr-pavlyk, line 1058: Cool, just learned that `IF/ELSE` is compile-time check feature in Cython, similar to `if constexpr` in C++.

`CUDA_CORE_BUILD_MAJOR` compile-time environment variable set to `cythonize` in `build_hook.py`.

cuda_core/cuda/core/experimental/_memory.pyx
kkraus14: Can be imported from https://github.com/cython/cython/blob/master/Cython/Includes/libc/limits.pxd

cuda_core/cuda/core/experimental/_memory.pyx
rparolin, line 860: Do we need Windows support? FD's on Windows are pointer size. Should we introduce a conditional or make this data type a uintptr_t to ensure its large enough on all platforms?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -833,6 +879,18 @@ class DeviceMemoryResource(MemoryResource):
                 raise
         return self._alloc_handle
 
+    cdef Buffer _allocate(self, size_t size, cyStream stream):
+        cdef cydriver.CUstream s = stream._handle
+        cdef cydriver.CUdeviceptr devptr
+        with nogil:
+            HANDLE_RETURN(cydriver.cuMemAllocFromPoolAsync(&devptr, size, self._mempool_handle, s))
+        cdef Buffer buf = Buffer.__new__(Buffer)
+        buf._ptr = <intptr_t>(devptr)
```
rparolin, line 888: We are sure we want intptr_t and not uintptr_t?

cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
shwina: Also it's also possible to inline C++ in Cython if needed with:

```cython
cdef extern from *:
    """
    C++ code goes here
    """
```

cuda_core/cuda/core/experimental/_device.pyx
leofang: `with A, B, ...` is functionally equivalent to
```python
with A:
    with B:
        ...
```
so I used it to avoid excessive indentation.

cuda_core/cuda/core/experimental/_stream.pyx
leofang: I will change the comment to "lazy init'd"

cuda_core/cuda/core/experimental/_memory.pyx
leofang: @shwina I might have missed something. How does `__reduce__` help `close`/`__dealloc__`?

cuda_core/cuda/core/experimental/_memory.pyx
oleksandr-pavlyk: This breaks Cython:

```
$ cython -+ b.pyx

Error compiling Cython file:
...

def example():
    return NULL
           ^

b.pyx:3:11: Cannot convert 'void *' to Python object
```

cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
leofang: As per our understanding (https://github.com/NVIDIA/cuda-python/issues/1065#issuecomment-3373942552 and https://github.com/NVIDIA/cuda-python/pull/1070#discussion_r2408239719) it is the best to keep it as the caller's responsibility for releasing the GIL, following Cython's intention on the `nogil` signature:
> Be aware that this does not release the GIL when calling the function. It merely indicates that a function is suitable for use when the GIL is released. It is also fine to call these functions while holding the GIL.

https://cython.readthedocs.io/en/latest/src/userguide/nogil.html#marking-functions-as-able-to-run-without-the-gil

cuda_core/cuda/core/experimental/_memory.pyx
leofang: This returns to the Python users, not Cython.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: This is being done in #1032. Sorry I forgot to leave a comment here...

cuda_core/cuda/core/experimental/_memory.pyx
leofang: Alternatively we could revisit this when we start working on #1028?

cuda_core/cuda/core/experimental/_stream.pyx
leofang: I did not bother doing a profiling here, just thought that it might be good to reduce a few CUDA calls... 😛

cuda_core/cuda/core/experimental/_device.pyx
leofang: Yes. Though note the caveat that the upstream wants to deprecate this feature: https://github.com/NVIDIA/cuda-python/pull/1041#discussion_r2395993944.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: resolving this to use the same thread from above: https://github.com/NVIDIA/cuda-python/pull/1070#discussion_r2408162040

cuda_core/cuda/core/experimental/_memory.pyx
oleksandr-pavlyk: We definitely want `uintptr_t` to make sure we can represent all device pointers, even those whose 64-bit is set. Such pointers are conceivably possible on systems with several large memory capacity GPUs.

```
$ cat c.pyx
from libc.stdint cimport uintptr_t, intptr_t

def roundtrip_u(ptr : int) -> int:
    cdef uintptr_t _ptr = <uintptr_t>(int(ptr))
    return _ptr

def roundtrip_i(ptr : int) -> int:
    cdef intptr_t _ptr = <intptr_t>(int(ptr))
    return _ptr

$ cython -+ c.pyx 
$ g++ c.cpp -shared -fPIC $(python3-config --includes) $(python3-config --libs) -o c$(python3-config --extension-suffix)

$ python
Python 3.12.10 | packaged by conda-forge | (main, Apr 10 2025, 22:21:13) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import c
>>> import cuda.core.experimental as cc
>>> cc.Device().set_current()
>>> stream = cc.Device().create_stream()
>>> buf = cc.DeviceMemoryResource(cc.Device()).allocate(2048, stream)

>>> c.roundtrip_u(buf.handle) # pointer here does not have the highest bit set, so both roundtrips work
17213423616
>>> c.roundtrip_i(buf.handle)
17213423616

>>> c.roundtrip_u(2**63 + 5)  # contriving and example where it breaks
9223372036854775813
>>> c.roundtrip_i(2**63 + 5)
  File "<stdin>", line 1, in <module>
  File "c.pyx", line 8, in c.roundtrip_i
    cdef intptr_t _ptr = <intptr_t>(int(ptr))
OverflowError: Python int too large to convert to C ssize_t
```

cuda_core/cuda/core/experimental/_memory.pyx
shwina: Sorry, the diff appeared to show that this was within the `__reduce__` method. Please replace all instances of `__reduce__` in my comments with `close`.

cuda_core/cuda/core/experimental/_memory.pyx
oleksandr-pavlyk: What was the rationale for switching from `uintptr_t` in the first place?

cuda_core/cuda/core/experimental/_memory.pyx
kkraus14: I thought it would implicit cast to 0. So follow up question, should we return 0 here instead of None?

cuda_core/cuda/core/experimental/_memory.pyx
leofang: Discussed this during the review meeting, but wouldn't hurt to summarize here:
- `uintptr_t` is used for all CUDA object addresses 
- `intptr_t` is used only for `Buffer` in case we need to compute pointer offsets

cuda_core/cuda/core/experimental/_memory.pyx
leofang: Ah OK. Then I have a follow-up question: How do we know for certain if we'd call Cython `close` and not `Python` close, without an explicit `isinstance` check? We just inspect the generated code?

cuda_core/cuda/core/experimental/_memory.pyx
leofang: So previously we allowed returning `driver.CUdeviceptr` here (note the return type annotation), and it had to be a generic Python object. However, come to think about it, the suggestion would allow us to **unbreak** the contract:
https://github.com/NVIDIA/cuda-python/blob/8ddbb425e104d117c0b04a1fb74a280a7b388353/cuda_core/docs/source/release/0.X.Y-notes.rst?plain=1#L25
Let me see if it is doable quickly.

cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
Andy-Jost: I'd love to see an exploration into this. Given well-defined criteria, shave off a small piece of cuda.core, try a few implementation strategies, and evaluate the results.

cuda_core/cuda/core/experimental/_memory.pyx
Andy-Jost: > `isinstance` checks are notoriously expensive when `abc`s are involved,

I've run into this also. Hugely expensive.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: @oleksandr-pavlyk good question. I think by casting a Python `int` to `intptr_t` your example constitutes a lossy conversion that Cython protected you against. If you were to do the same in C, the C/C++ compiler would catch it too:
```
/local/home/leof/dev/round_trip_cast.c:2584:29: warning: integer constant is so large that it is unsigned
 2584 |   __pyx_v__ptr = ((intptr_t)9223372036854775813LL);
      |  
```

`intptr_t` only guarantees that can be round-trip converted to/from `void*` (which is what we need, even including multi-device in mind). It does not offer any guarantees beyond that to the best of my knowledge.

We could do this but it does not make much sense depending on the use case:
```cython
# distutils: language = c++

from libc.stdint cimport uintptr_t, intptr_t

def roundtrip_u(ptr) -> int:
    cdef uintptr_t _ptr = <uintptr_t>(int(ptr))
    return _ptr


def roundtrip_i(ptr) -> int:
    cdef intptr_t _ptr = <intptr_t>(int(ptr))
    return _ptr


def roundtrip_i_v2(ptr) -> int:
    cdef intptr_t _ptr = <intptr_t>(<uintptr_t>(int(ptr)))
    return _ptr


cdef extern from * nogil:
    """
    intptr_t convert_safe(void* ptr) {
        return reinterpret_cast<intptr_t>(ptr);
    }
    """
    intptr_t convert_safe(void* ptr)


def roundtrip_i_v3(ptr) -> int:
    cdef intptr_t _ptr = convert_safe(<void*><uintptr_t>(int(ptr)))
    return _ptr
```
Output:
```
>>> import round_trip_cast
>>> round_trip_cast.roundtrip_u(2**63 + 5)
9223372036854775813
>>> round_trip_cast.roundtrip_i(2**63 + 5)
  File "<stdin>", line 1, in <module>
  File "round_trip_cast.pyx", line 11, in round_trip_cast.roundtrip_i
    cdef intptr_t _ptr = <intptr_t>(int(ptr))
OverflowError: Python int too large to convert to C ssize_t
>>> round_trip_cast.roundtrip_i_v2(2**63 + 5)
-9223372036854775803
>>> round_trip_cast.roundtrip_i_v3(2**63 + 5)
-9223372036854775803
>>> 
```

cuda_core/cuda/core/experimental/_stream.pyx
leofang: Done in commit 2dd9138.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: I moved the cast to another line in commit 2dd9138.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: To be discussed tomorrow... I raised a thread offline.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: Done in commit 2dd9138. We have now unbroken the code (which was introduced in #876 that is not yet released 🙂).

cuda_core/cuda/core/experimental/_memory.pyx
leofang: Done in commit 2dd9138

cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
leofang: The ship has sailed. Let's create an issue to track this discussion. The next time we will revisit it is when we start looking into cccl-runtime as the underlying implementation of cuda.core. I am not sure if anyone actually understands the implication of going through C++ now, after all the discussions in recent Friday co-design meetings. There are things that we need and get from cuda-bindings not yet available in native C++.

leofang:
/ok to test [a5e6bcf](https://github.com/NVIDIA/cuda-python/pull/1070/commits/a5e6bcf90370cf5a49f0ac901273547e922399b2)

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -99,7 +96,20 @@ cdef class Buffer:
             The stream object to use for asynchronous deallocation. If None,
             the behavior depends on the underlying memory resource.
         """
-        self._shutdown_safe_close(stream, is_shutting_down=None)
+        cdef _cyMemoryResource cy_mr
+        if self._ptr and self._mr is not None:
```
leofang, line 140: We probably have a bigger problem here. Regardless of how `self._mr` is implemented, it is typed as `object`, so if we are accessing it in `__dealloc__` it is unsafe. I'll also think about this in the meantime...

leofang: NVM, it is guaranteed that self._mr stays until after self is destroyed:
> You don’t need to worry about deallocating Python attributes of your object, because that will be done for you by Cython after your __dealloc__() method returns.

https://cython.readthedocs.io/en/3.1.x/src/userguide/special_methods.html#finalization-methods-dealloc-and-del

cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
leofang: As noted in another thread, the caller takes the responsibility, not the callee.

cuda_core/cuda/core/experimental/_memory.pyx
leofang: I have addressed this hopefully nicely in commit 3809a33, following @shwina's idea captured below.
> With a little bit of shifting things around in the inheritance hierarchy, it  appears possible to forgo isinstance. The key being to derive from _cyMemoryResource in both DeviceMemoryResource as well as the base MemoryResource type intended to be subclassed by users (which I have called UserMemoryResource below):
```cython
import abc

cdef class _cyMemoryResource:
    cdef void _deallocate(self):
        raise NotImplementedError("...")

cdef class UserMemoryResource(_cyMemoryResource, abc.ABC):
    cdef void _deallocate(self):
        self.deallocate()

    @abc.abstractmethod
    def deallocate(self):
        ...

cdef class FastMemoryResource(_cyMemoryResource):
    cdef void _deallocate(self):
        print("Fast library deallocate")

    cpdef deallocate(self):
        self._deallocate()

class MyMemoryResource(UserMemoryResource):
    def deallocate(self):
        print("Slow user deallocate")

cdef class Buffer:
    cdef _cyMemoryResource mr

    def __cinit__(self, _cyMemoryResource mr):
        self.mr = mr

    def __init__(self, mr):
        self.mr = mr

    cpdef void close(self):
        self.mr._deallocate()


buf1 = Buffer(FastMemoryResource())
buf2 = Buffer(MyMemoryResource())
buf1.close()
buf2.close()
```
> (In the example above, I suppose you don't even need FastMemoryResource to implement deallocate(), if it's totally hidden from the user)

After some struggle I’ve implemented it. The struggles include:
- MemoryResource (your UserMemoryResource) cannot have abstract properties. It choked Cython. So I’ve moved the properties to another ABC class MemoryResourceAttributes.
   - I wanted to do this anyway (you can see there was a TODO comment that’s removed now) because 1. it maps nicely to the C++ concepts, and 2. both Buffer and MemoryResource can inherit from it to conform to the interface.
- MemoryResource cannot mark `__init__` as an abstract method. It is useless anyway, so I removed it.
- For Buffer to inherit a Python class (MemoryResourceAttributes) while staying as a cdef class, the same trick had to be applied and so I separated out a _cyBuffer cdef class.
- All child classes that has a cdef and a Python parent must have a `__dict__`.

cuda_core/cuda/core/experimental/_stream.pyx
kkraus14: Can we remove this TODO before merge? It's not logically correct to call `cuCtxGetCurrent` here as the Stream object could be from a different context.

kkraus14 (dismissed):
Approving as the last change I'd like is trivial

cuda_core/cuda/core/experimental/_stream.pyx
leofang: I see why this TODO is confusing. We actually need the current context in order to do a safe `cuCtxGetDevice` dance (set the stream ctx to current -> get device -> revert to the current ctx). Let me clarify in the code.

kkraus14: I think we can just use `cuCtxGetDevice_v2` which allows passing a ctx explicitly and doesn't require messing with the current ctx?

leofang: See commit 22b0e2e5

leofang: > I think we can just use `cuCtxGetDevice_v2` which allows passing a ctx explicitly and doesn't require messing with the current ctx?

Unfortunately it seems to be CUDA 13 only... 😢

rparolin (approved):

Issue #692, opened 2025-06-07
Make `cuda-bindings` buildable against CTK wheels?


leofang:
The meaning of the title is that `pip install .` from source can work without a local CTK environment (no need to set `CUDA_PATH` and all that). Everything is purely pip-wheel based.

This is actually doable, I just don't believe there is a serious use case because we really want users to not build from source, nor do I want to spend my time on enabling it, but I can be convinced. Anyway it's worth writing this up somewhere in the public.

The whole thing relies on the fact that building bindings just need a host compiler; no NVCC is needed. The steps is as follows:
- We [list needed CTK wheels](https://github.com/NVIDIA/nvmath-python/blob/1cbc60be436fe98e8a08f38e9867bc6fba90f007/pyproject.toml#L8-L10) in the `[build-system.requires]` session. They will provide the needed headers.
- Since `setuptools` does not provide a way for us to programmatically query the path to the isolated build environment at build time, [we search the potential paths](https://github.com/NVIDIA/nvmath-python/blob/1cbc60be436fe98e8a08f38e9867bc6fba90f007/builder/utils.py#L12-L41) based on the assumption that `sys.path` must be updated for `setuptools` to find the build env.
- We [pass `<build env path>` to `setuptools`](https://github.com/NVIDIA/nvmath-python/blob/1cbc60be436fe98e8a08f38e9867bc6fba90f007/builder/utils.py#L118) so that the CUDA headers can be found by gcc/msvc.

I enabled this for nvmath-python and it did worked like a charm, but the situation was a lot simpler than driver/runtime modules that we have here.

kkraus14:
Sounds like a job for `path_finder` if we split it off from `cuda.bindings` 😆

leofang:
Oh, so you meant if we install `path_finder` in the isolated build env, it knows the self location and can infer other build dependencies from this knowledge without guessing. Interesting!

rparolin:
I'm going to close this issue due to inactivity. If we want to ultimately do this work please re-open the ticket.

leofang:
Took me a while to find this. Let's not close this. Keith and I discussed last week and we actually want this to happen.

leofang:
Closely related: https://github.com/NVIDIA/cuda-python/issues/2060

PR #1085, opened 2025-10-04
Add `_get_cuda_version_from_cuda_h()` in `cuda_core/build_hooks.py`

rwgk:
WIP — Will work on it more asap.

### DIFF
--- a/cuda_core/build_hooks.py
+++ b/cuda_core/build_hooks.py
@@ -12,6 +12,7 @@
 import functools
 import glob
 import os
+import pathlib
 import re
 import subprocess
 
@@ -22,6 +23,46 @@ from setuptools import build_meta as _build_meta
 prepare_metadata_for_build_wheel = _build_meta.prepare_metadata_for_build_wheel
 build_sdist = _build_meta.build_sdist
 get_requires_for_build_sdist = _build_meta.get_requires_for_build_sdist
+
+
+@functools.cache
+def _get_cuda_paths():
+    CUDA_PATH = os.environ.get("CUDA_PATH", os.environ.get("CUDA_HOME", None))
+    if not CUDA_PATH:
+        raise RuntimeError("Environment variable CUDA_PATH or CUDA_HOME is not set")
+    CUDA_PATH = CUDA_PATH.split(os.pathsep)
+    print("CUDA paths:", CUDA_PATH, flush=True)
+    return CUDA_PATH
+
+
+@functools.cache
+def _get_cuda_version_from_cuda_h(cuda_home=None):
+    """
+    Given CUDA_HOME, try to extract the CUDA_VERSION macro from include/cuda.h.
+
+    Example line in cuda.h:
+        #define CUDA_VERSION 13000
+
+    Returns the integer (e.g. 13000) or None if not found / on error.
+    """
+    if cuda_home is None:
+        cuda_home = _get_cuda_paths()[0]
+
+    cuda_h = pathlib.Path(cuda_home) / "include" / "cuda.h"
+    if not cuda_h.is_file():
+        return None
+
+    try:
+        text = cuda_h.read_text(encoding="utf-8", errors="ignore")
+    except OSError:
+        # Permissions issue, unreadable file, etc.
+        return None
+
+    m = re.search(r"^\s*#define\s+CUDA_VERSION\s+(\d+)", text, re.MULTILINE)
+    if not m:
+        return None
+    print(f"CUDA_VERSION from {cuda_h}:", m.group(1), flush=True)
+    return m.group(1)
 
 
 @functools.cache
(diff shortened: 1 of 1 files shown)


rwgk:
Local testing with a venv that does not have cuda-bindings:

```
pip install -v . --group test
```

```
...
Building wheels for collected packages: cuda-core
  Building wheel for cuda-core (pyproject.toml): started
  Running command Building wheel for cuda-core (pyproject.toml)
  CUDA paths: ['/usr/local/cuda']
  CUDA_VERSION from /usr/local/cuda/include/cuda.h: 13000
...
```

```
$ pip list
Package    Version
---------- -------
cuda-core  0.3.3a0
Cython     3.1.4
iniconfig  2.1.0
numpy      2.3.3
packaging  25.0
pip        25.2
pluggy     1.6.0
Pygments   2.19.2
pytest     8.4.2
setuptools 80.9.0
```

leofang:
Calling nvidia-smi is better because it does exactly the same thing under the hood (opening up libcuda/nvcuda and getting the driver version), but it is cross-platform and so takes much less lines of code, and it is a good habit to keep the code surface in `build_hook.py` small because it is not easy to write tests for modules outside of the source tree.

rwgk:
> Calling nvidia-smi is better because it does exactly the same thing under the hood (opening up libcuda/nvcuda and getting the driver version), but it is cross-platform and so takes much less lines of code, and it is a good habit to keep the code surface in `build_hook.py` small because it is not easy to write tests for modules outside of the source tree.

The main reason I was leaning into it is that I think it'd be a useful addition to pathfinder.

It's more lines of code only because `subprocess` is being reused.

Assume that the code I added lives in pathfinder, you'd just reuse that and have a more secure, more robust, and faster (which doesn't matter here, but is useful in general) solution with even less lines of code.

cuda_core/build_hooks.py
```diff
@@ -24,6 +26,110 @@
...
+        int  : driver version (e.g., 12040 for 12.4), if successful.
+        None : on any failure (load error, missing symbol, non-success CUresult).
+    """
+    CUDA_SUCCESS = 0
+
+    try:
+        # WinDLL => stdcall (CUDAAPI on Windows), matches CUDA Driver API.
+        lib = ctypes.WinDLL("nvcuda.dll")
+    except OSError:
+        return None
```
leofang: This snippet is all we need for Windows and the rest can go. This way we just need one function to encapsulate driver version fetching in a cross-platform fashion. The platform check can happen inside the callee, not the caller.

leofang (commented):
> The main reason I was leaning into it is that I think it'd be a useful addition to pathfinder.

Why would the pathfinder need to expose the driver version check as a public API? Or you only had the driver **loading** part in mind?

rwgk:
> > The main reason I was leaning into it is that I think it'd be a useful addition to pathfinder.
> 
> Why would the pathfinder need to expose the driver version check as a public API?

It'd be useful in general, not just here. It's just a few lines of Python code that is very similar to other functionality in pathfinder, and does not introduce any new dependencies. Having the function there, would make what's needed here a no-brainer.

cuda_core/build_hooks.py
leofang: done in commit 3a2260a3

cuda_core/build_hooks.py
```diff
@@ -23,6 +25,88 @@
 get_requires_for_build_sdist = _build_meta.get_requires_for_build_sdist
 
 
+@functools.cache
+def _get_cuda_paths():
+    CUDA_PATH = os.environ.get("CUDA_PATH", os.environ.get("CUDA_HOME", None))
+    if CUDA_PATH is None:
+        return None
```
leofang, line 32: Note: I changed this to return None if no env var is set, but I think it is a mistake because we still need paths to the headers when creating `Extension` instances below. I think it is time to evaluate using pathfinder. It is already a build-time dependency anyway (cuda-bindings brought it in).

FWIW it seems #692 suddenly becomes within reach!

cuda_core/build_hooks.py
```diff
@@ -38,15 +122,14 @@ def _get_proper_cuda_bindings_major_version() -> str:
...
-        out = subprocess.run("nvidia-smi", env=os.environ, capture_output=True, check=True)  # noqa: S603, S607
-        m = re.search(r"CUDA Version:\s*([\d\.]+)", out.stdout.decode())
-        if m:
-            return m.group(1).split(".")[0]
-    except (FileNotFoundError, subprocess.CalledProcessError):
-        # the build machine has no driver installed
-        pass
+    cuda_version = _get_cuda_driver_version()
+    if cuda_version:
+        return str(cuda_version // 1000)
```
leofang, line 132: Since we must need CUDA headers at build time, this is no longer needed in terms of deciding `cuda-bindings` version, but it can be a useful check in case there is a local driver installation that cannot support the package being built. We can raise a nice warning in such cases via this check.

rwgk:
I put this in the freezer for now, but the branch is still around: freezer/cuda_core_build_hooks_get_cuda_version_macro

Issue #1074, opened 2025-10-03
IPC tests do not exit cleanly


leofang:
After merging #1020, I see locally that the tests generate the following output with `pytest -rA`:
```
______________________________________________________ TestIPCSharedAllocationHandleAndBufferDescriptors.test_main _______________________________________________________
-------------------------------------------------------------------------- Captured stderr call --------------------------------------------------------------------------
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
________________________________________________________ TestIPCSharedAllocationHandleAndBufferObjects.test_main _________________________________________________________
-------------------------------------------------------------------------- Captured stderr call --------------------------------------------------------------------------
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
__________________________________________________________________________ test_object_passing ___________________________________________________________________________
-------------------------------------------------------------------------- Captured stderr call --------------------------------------------------------------------------
Traceback (most recent call last):
  File "cuda/core/experimental/_memory.pyx", line 82, in cuda.core.experimental._memory.Buffer._shutdown_safe_close
  File "cuda/core/experimental/_memory.pyx", line 877, in cuda.core.experimental._memory.DeviceMemoryResource.deallocate
  File "cuda/core/experimental/_utils/cuda_utils.pyx", line 75, in cuda.core.experimental._utils.cuda_utils._check_driver_error
cuda.core.experimental._utils.cuda_utils.CUDAError: CUDA_ERROR_INVALID_VALUE: This indicates that one or more of the parameters passed to the API call is not within an acceptable range of values.
Exception ignored in: 'cuda.core.experimental._memory.Buffer.__del__'
Traceback (most recent call last):
  File "cuda/core/experimental/_memory.pyx", line 82, in cuda.core.experimental._memory.Buffer._shutdown_safe_close
  File "cuda/core/experimental/_memory.pyx", line 877, in cuda.core.experimental._memory.DeviceMemoryResource.deallocate
  File "cuda/core/experimental/_utils/cuda_utils.pyx", line 75, in cuda.core.experimental._utils.cuda_utils._check_driver_error
cuda.core.experimental._utils.cuda_utils.CUDAError: CUDA_ERROR_INVALID_VALUE: This indicates that one or more of the parameters passed to the API call is not within an acceptable range of values.
Error in sys.excepthook:

Original exception was:

```
@Andy-Jost is this a symptom of #1063?

[...]

PR #1096, opened 2025-10-07
Update tests to close buffers.

Andy-Jost:
Update the memory_ipc tests to explicitly call close on buffers.

### DIFF
--- a/.github/workflows/build-wheel.yml
+++ b/.github/workflows/build-wheel.yml
@@ -230,8 +230,6 @@ jobs:
         with:
           # workaround for actions/runner-images#12377 (the cached 3.13.4 is buggy on Windows)
           python-version: ${{ matrix.python-version == '3.13' && '3.13.5' || matrix.python-version }}
-          # allow prereleases only fetches prereleases for unreleased versions of Python
-          allow-prereleases: true
 
       - name: verify free-threaded build
         if: endsWith(matrix.python-version, 't')
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -193,6 +193,7 @@ jobs:
 
   checks:
     name: Check job status
+    if: always()
     runs-on: ubuntu-latest
     needs:
       - test-linux-64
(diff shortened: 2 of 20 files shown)


cuda_core/tests/memory_ipc/test_serialize.py
```diff
@@ -178,3 +185,6 @@ def child_main(alloc_handle, mr1, buffer_desc, buffer1):
     helper1.verify_buffer(flipped=True)
     helper2.verify_buffer(flipped=True)
     helper3.verify_buffer(flipped=True)
+
+    # Close any one buffer.
+    buffer1.close()
```
Andy-Jost, line 190: This is a fresh nightmare. It is possible create aliases by mapping the same buffer multiple times. Closing more than one of them results in CUDA_ERROR_INVALID_VALUE

Andy-Jost: @ksimpson-work is this a driver problem?

leofang: Andy, as noted here https://github.com/NVIDIA/cuda-python/issues/1074#issuecomment-3367798948 and in the internal thread, this is a user problem.

leofang: It is UB to call `cudaFree{Async}` more than once on the same pointer. We must not alias the pointer.

Andy-Jost: This is a dreadful user experience. The application can embed buffers into arbitrary data structures when serializing, but this way they have to be certain there are no duplicates. It would be much better to add a reference count so that calls to `cuMemPoolImportPointer` are matched by calls to `cuMemFree(Async)`. This seems like a very standard behavior -- e.g., how Linux memory mapping works.

leofang: I saw your notes in both #1074 and here, and then discussed internally. Let's keep the status update and follow-up discussions in #1074. This PR is still a progress and let's not get distracted.

leofang:
/ok to test [40f2bf2](https://github.com/NVIDIA/cuda-python/pull/1096/commits/40f2bf23d5b52b7b4893731d06f64b07022c8e7b)

PR #1102, opened 2025-10-08
Avoid manual cleanup of object attributes

leofang:
## Description

Try to avoid cleanup errors like this
```
  File "cuda/core/experimental/_memory.pyx", line 138, in cuda.core.experimental._memory.Buffer.close
SystemError: Objects/dictobject.c:2395: bad argument to internal function
  File "cuda/core/experimental/_memory.pyx", line 751, in cuda.core.experimental._memory.DeviceMemoryResource.close
SystemError: Objects/dictobject.c:2395: bad argument to internal function
Exception ignored in: 'cuda.core.experimental._memory.DeviceMemoryResource.__dealloc__'
```

## Checklist

### DIFF
--- a/cuda_core/cuda/core/experimental/_memory.pyx
+++ b/cuda_core/cuda/core/experimental/_memory.pyx
@@ -160,9 +160,6 @@ cdef class Buffer(_cyBuffer, MemoryResourceAttributes):
                 s = <cyStream>stream
             self._mr._deallocate(self._ptr, self._size, s)
             self._ptr = 0
-            self._mr = None
-            self._ptr_obj = None
-            self._alloc_stream = None
 
     @property
     def handle(self) -> DevicePointerT:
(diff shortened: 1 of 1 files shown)


cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -160,9 +160,6 @@ cdef class Buffer(_cyBuffer, MemoryResourceAttributes):
                 s = <cyStream>stream
             self._mr._deallocate(self._ptr, self._size, s)
             self._ptr = 0
```
cpcloud, line 162: This was something I had mentioned elsewhere, that we shouldn't be accessing object properties in a `__dealloc__` code path.

Why is `_ptr` still being touched here?

From the Cython docs on `__dealloc__`:

> In particular, don’t call any other methods of the object or do anything which might cause the object to be resurrected. It’s best if you stick to just deallocating C data.

Why not just do the minimum C-API invocation required in `__dealloc__` and duplicate whatever you need to from `close`?

Then we're not going against the documented behavior and causing a crashing knowing that we could have avoided it all along.

leofang: `self._ptr` is a cdef attribute. I think it might be a confusion with `self._ptr_obj`?

Regardless, I don't believe this PR is needed. I think #1104 and #1105 together should fix everything for us.

PR #1103, opened 2025-10-08
build: enable free-threading compatibility

cpcloud:
Enable freethreading_compatibility for cuda-python extension modules


cuda_bindings/setup.py
```diff
@@ -329,7 +329,7 @@ def do_cythonize(extensions):
     return cythonize(
         extensions,
         nthreads=nthreads,
-        compiler_directives=dict(language_level=3, embedsignature=True, binding=True),
+        compiler_directives=dict(language_level=3, embedsignature=True, binding=True, freethreading_compatible=True),
```
leofang, line 332: FWIW Cython's doc is really unclear on this. I hope this flag does nothing for regular builds...

leofang (changes requested):
It looks like when we build Cython tests, we also have to pass the new flag, e.g.:
https://github.com/NVIDIA/cuda-python/actions/runs/18346151043/job/52261604753?pr=1103#step:26:340

cuda_bindings/setup.py
cpcloud: I don't even know what that would mean!

It's gated by a preprocessor define that's only set to `1` if `Py_GIL_DISABLED` is defined, and even then, that would require `Py_mod_gil` to be defined, which didn't happen until 2024 (in https://github.com/python/cpython/pull/116882). If this flag somehow allowed making it to the point where `Py_mod_gil` was accessed, then it simply wouldn't be defined in the <3.13t python headers, and the user's module would fail to compile.

I don't think we need to be worried about this.

cpcloud:
Gah, what is going on with system dependency installation?

cpcloud:
Also, I will do a follow up to turn on warnings as errors in pytest.

leofang:
> Gah, what is going on with system dependency installation?

no idea, asking in the runner channel internally

cuda_bindings/setup.py
leofang: yeah agreed, we follow whatever NumPy/SciPy do... 😛 
- https://github.com/numpy/numpy/blob/128d684bdc8d756c63341baab0fd7e4df3fbc09c/numpy/_core/tests/examples/cython/setup.py#L31-L33
- https://github.com/scipy/scipy/blob/2ee4675fb1d1cfe70828cfc287e6078b33d8be72/scipy/meson.build#L505

cpcloud:
I also handled `build_tests.bat`. I think this is ready to go now.

leofang (approved):

leofang:
Good catch!

leofang:
/ok to test [47c13be](https://github.com/NVIDIA/cuda-python/pull/1103/commits/47c13bee7d32102b1724b3622a083248a30d748f)

leofang:
I'll check the CI log later and see if the warning is gone.
> Also, I will do a follow up to turn on warnings as errors in pytest.

It might be a good idea. Though I am not sure how much work it is. For this particular case, it might be a good idea to also add a naive test like this
https://github.com/numpy/numpy/blob/128d684bdc8d756c63341baab0fd7e4df3fbc09c/tools/wheels/cibw_test_command.sh#L22-L25

leofang:
btw @cpcloud we'll need to backport the cuda-bindings changes...

leofang:
> I'll check the CI log later and see if the warning is gone.

Looks like they are gone now. I will enable auto-merge.

PR #1105, opened 2025-10-08
Avoid accessing self's Python attributes in `__dealloc__`

leofang:
## Description

closes 

## Checklist

- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #1104, opened 2025-10-08
Avoid adding `__dict__` to every child class

leofang:
## Description

closes 

## Checklist

- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #1107, opened 2025-10-08
fix: ensure that there are no module global lookups when calling `default_stream`

cpcloud:
As discussed internally, this was likely the cause of the failures we were seeing after #1104 and #1105.

### DIFF
--- a/cuda_core/cuda/core/experimental/_stream.pyx
+++ b/cuda_core/cuda/core/experimental/_stream.pyx
@@ -5,6 +5,7 @@
 from __future__ import annotations
 
 from libc.stdint cimport uintptr_t, INT32_MIN
+from libc.stdlib cimport strtol, getenv
 
 from cuda.bindings cimport cydriver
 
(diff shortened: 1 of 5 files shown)


cuda_core/tests/test_stream.py
```diff
@@ -107,11 +107,6 @@ def test_per_thread_default_stream():
     assert isinstance(PER_THREAD_DEFAULT_STREAM, Stream)
 
 
-def test_default_stream():
-    stream = default_stream()
-    assert isinstance(stream, Stream)
```
cpcloud, line 112: Since we're guaranteeing this in the type system, the runtime `isinstance` assertion is no longer useful.

leofang:
> As discussed internally, this was likely the cause of the failures we were seeing after #1104 and #1105.

FWIW would be nice to keep for a record what failures we see after the two fixes. Asking because I thought this was just (very) nice to have.

cpcloud:
The problem wasn't solved after #1104 and #1105 because global Python lookups are still happening in the `__dealloc__` code path (via `default_stream`), though of course it helps having one fewer python object to think about.

I don't have a failure that reproduces it after both of those PRs, but logically there's nothing else in the `__dealloc__` code path even *before* removing `__dict__`. All the attributes being deallocated were C-level (inside `close`) and the `close` call itself was happening in C.

This is the last remaining global object access in the `__dealloc__` code path.

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -403,8 +409,14 @@ def default_stream():
 
     """
     # TODO: flip the default
-    use_ptds = int(os.environ.get("CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM", 0))
+    cdef const char* use_ptds_raw = getenv("CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM")
```
kkraus14, line 412: Instead of looking this up in the function, we could probably resolve this as a global. Since we resolve the underlying function symbols only once we don't handle the situation where someone changes the environment variable value during process execution.

PR #1110, opened 2025-10-08
chore(dev): add pixi setup for cuda-bindings

cpcloud:
This PR adds a pixi environment for cuda-bindings

### DIFF
--- a/cuda_bindings/pyproject.toml
+++ b/cuda_bindings/pyproject.toml
@@ -37,8 +37,6 @@ all = [
 
 [dependency-groups]
 test = [
-    "cython>=3.1,<3.2",
-    "setuptools>=77.0.0",
     "numpy>=1.21.1",
     "pytest>=6.2.4",
     "pytest-benchmark>=3.4.1",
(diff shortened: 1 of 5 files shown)


cuda_bindings/pyproject.toml
```diff
@@ -67,3 +65,19 @@ archs = "native"
...
+platforms = ["linux-64", "linux-aarch64", "win-64"]
+
+[tool.pixi.pypi-dependencies]
+cuda-bindings = { path = ".", editable = true }
+
+[tool.pixi.environments]
+cu13 = { features = ["all", "test"], solve-group = "cu13" }
+
+[tool.pixi.build-dependencies]
+cuda-toolkit = ">=13,<14"
```
kkraus14: This pulls in a bunch of stuff we don't want. Can you reference the libraries in the `all` optional dependency group?

cpcloud: I can't reference them in any sort automatic way, but does it matter since this is `build-dependencies`?

cpcloud: The other thing is that including `cuda-toolkit` appears to set `CUDA_HOME` at build time, which I need to build `cuda-bindings` as part of environment activation.

cpcloud: If there's a way to do that without including the kitchen sink, that'd be great!

kkraus14: It matters for at least the cuda runtime library since we statically link that in

cpcloud: Ah, ok. Then I'll have to figure out what's happening with `CUDA_HOME` and how that magically works with `cuda-toolkit`.

cpcloud: Ok, I figured out the minimum set to build.

cuda_bindings/pyproject.toml
```diff
@@ -67,3 +65,22 @@ archs = "native"
...
+
+[tool.pixi.workspace]
+channels = ["conda-forge"]
+platforms = ["linux-64", "linux-aarch64", "win-64"]
+
+[tool.pixi.pypi-dependencies]
+cuda-bindings = { path = ".", editable = true }
+
+[tool.pixi.environments]
+cu13 = { features = ["all", "test"], solve-group = "cu13" }
```
kkraus14, line 77: Does the `all` feature end up grabbing wheels or does it map to the proper conda packages as defined in `build-dependencies` below?

cuda_bindings/pyproject.toml
```diff
@@ -67,3 +65,22 @@ archs = "native"
...
+
+[tool.pixi.build-dependencies]
+cuda-cudart-dev = ">=13,<14"
```
kkraus14: In hindsight I think we need `cuda-driver-dev` since we look for `cuda.h` at build time as well

cuda_bindings/pyproject.toml
cpcloud: All of the pyproject.toml integration maps to `pypi-dependencies`: https://pixi.sh/dev/python/pyproject_toml/#dependency-section

kkraus14: We can do it in a follow up, but we should ideally use https://pixi.sh/dev/reference/pixi_manifest/#conda-pypi-map-optional to map the pip wheels to the relevant conda packages in some kind of environment so we can support both wheel package based and conda package based environments.

cuda_bindings/pyproject.toml
```diff
@@ -67,3 +65,40 @@ archs = "native"
...
+[tool.pixi.pypi-dependencies]
+cuda-bindings = { path = ".", editable = true }
+
+[tool.pixi.environments]
+cu13 = { features = ["all", "test"], solve-group = "cu13" }
+
+[tool.pixi.feature.all.dependencies]
+libnvjitlink = "*"
+
+[tool.pixi.build-dependencies]
```
kkraus14, line 82: We need to comprehensively capture these if we're adding conda package dependencies: https://github.com/conda-forge/cuda-python-feedstock/blob/69231e7f5de1c5a1314376f34308f82141f6112e/recipe/meta.yaml#L120-L131

In general we can't mix and match conda and pypi packages where it's kinda an all-or-nothing

cpcloud:
Closing this out, going to make a new PR with `pixi.toml` to avoid the problem of mixing environments.

PR #1117, opened 2025-10-09
Add (failing) tests demonstrating that errors in Buffer.close are not raised

Andy-Jost:
Errors occurring during `Buffer.close` are not raised. This change adds tests demonstrating the issue. See #1118.

### DIFF
--- a/cuda_core/tests/memory_ipc/test_errors.py
+++ b/cuda_core/tests/memory_ipc/test_errors.py
@@ -9,6 +9,8 @@ import pytest
 from cuda.core.experimental import Buffer, Device, DeviceMemoryResource, DeviceMemoryResourceOptions
 from cuda.core.experimental._utils.cuda_utils import CUDAError, driver
 
+from cuda_python_test_helpers import supports_ipc_mempool
+
 CHILD_TIMEOUT_SEC = 20
 NBYTES = 64
 POOL_SIZE = 2097152
@@ -19,6 +21,10 @@ class ChildErrorHarness:
     PARENT_ACTION, CHILD_ACTION, and ASSERT (see below for examples)."""
 
     def test_main(self, ipc_device, ipc_memory_resource):
+        if not supports_ipc_mempool(ipc_device):
+            import pytest
+
+            pytest.skip("Driver rejects IPC-enabled mempool creation on this platform")
         """Parent process that checks child errors."""
         # Attach fixtures to this object for convenience. These can be accessed
         # from PARENT_ACTION.
--- a/.github/workflows/test-wheel-linux.yml
+++ b/.github/workflows/test-wheel-linux.yml
@@ -283,7 +283,7 @@ jobs:
         run: |
           set -euo pipefail
           pushd cuda_pathfinder
-          pip install --only-binary=:all: -v . --group "test_nvidia_wheels_cu${TEST_CUDA_MAJOR}" --group test_nvidia_wheels_host
+          pip install --only-binary=:all: -v . --group "test-cu${TEST_CUDA_MAJOR}"
           pip list
           popd
 
--- a/.github/workflows/test-wheel-windows.yml
+++ b/.github/workflows/test-wheel-windows.yml
@@ -256,7 +256,7 @@ jobs:
         shell: bash --noprofile --norc -xeuo pipefail {0}
         run: |
           pushd cuda_pathfinder
-          pip install --only-binary=:all: -v . --group "test_nvidia_wheels_cu${TEST_CUDA_MAJOR}" --group test_nvidia_wheels_host
+          pip install --only-binary=:all: -v . --group "test-cu${TEST_CUDA_MAJOR}"
           pip list
           popd
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ ci:
 # pre-commit autoupdate --freeze
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: f298305809c552671cc47e0fec0ba43e96c146a2  # frozen: v0.13.2
+    rev: 9c89adb347f6b973f4905a4be0051eb2ecf85dea  # frozen: v0.13.3
     hooks:
       - id: ruff
         args: [--fix, --show-fixes]
--- a/cuda_bindings/DESCRIPTION.rst
+++ b/cuda_bindings/DESCRIPTION.rst
@@ -5,6 +5,10 @@
 cuda-bindings: Low-level CUDA interfaces
 ****************************************
 
+.. image:: https://img.shields.io/badge/NVIDIA-black?logo=nvidia
+   :target: https://www.nvidia.com/
+   :alt: NVIDIA
+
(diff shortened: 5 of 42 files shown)


cuda_core/tests/memory_ipc/test_errors.py
```diff
@@ -147,3 +148,66 @@ def CHILD_ACTION(self, queue):
...
+
+def test_error_in_close_memory_resource(ipc_memory_resource):
+    """Test that errors when closing a memory resource are raised."""
+    mr = ipc_memory_resource
+    driver.cuMemPoolDestroy(mr.handle)
+    with pytest.raises(CUDAError, match=".*CUDA_ERROR_INVALID_VALUE.*"):
+        mr.close()
+
+
+@pytest.mark.xfail
```
rwgk: Will this work here?

```
@pytest.mark.xfail(reason="Issue #1118", strict=True)
```

The important part is `strict=True`.

cuda_core/tests/memory_ipc/test_errors.py
```diff
@@ -147,3 +148,66 @@ def CHILD_ACTION(self, queue):
...
         assert exc_type is RuntimeError
         assert re.match(r"Memory resource [a-z0-9-]+ was not found", exc_msg)
+
+
+def test_error_in_close_memory_resource(ipc_memory_resource):
+    """Test that errors when closing a memory resource are raised."""
+    mr = ipc_memory_resource
+    driver.cuMemPoolDestroy(mr.handle)
+    with pytest.raises(CUDAError, match=".*CUDA_ERROR_INVALID_VALUE.*"):
+        mr.close()
```
leofang, line 158: This is illegal and I disagree we need to test this. This is Python and we can't possibly guard against all kinds of bizarre ways of trying to mutate the state of our Python objects behind our back. In particular, as noted in both https://github.com/NVIDIA/cuda-python/issues/1074#issuecomment-3367798948 and offline discussion, errors like `CUDA_ERROR_INVALID_VALUE` are due to multiple frees. I thought we've moved on?

This test is just another instance of the same class of errors: We free the handle of an object through a direct C API call, bypassing our safeguard mechanism (under the hood we do check if the handle is already null before freeing, and then after free we set the handle to null to avoid double free), so our destructor kicks in, either through an explicit `close()` call or implicitly when going out of scope, and causes another free.

cuda_core/tests/memory_ipc/test_errors.py
```diff
@@ -147,3 +148,66 @@ def CHILD_ACTION(self, queue):
...
+
+@pytest.mark.xfail
+@pytest.mark.filterwarnings("ignore")  # Test incorrectly warns
+def test_error_in_close_buffer(ipc_memory_resource):
+    """Test that errors when closing a memory buffer are raised."""
+    mr = ipc_memory_resource
+    buffer = mr.allocate(NBYTES)
+    driver.cuMemFree(buffer.handle)
+    with pytest.raises(CUDAError, match=".*CUDA_ERROR_INVALID_VALUE.*"):
+        buffer.close()
```
leofang, line 169: ditto

cuda_core/tests/memory_ipc/test_errors.py
```diff
@@ -147,3 +148,66 @@ def CHILD_ACTION(self, queue):
...
+    def PARENT_ACTION(self, queue):
+        pass
+
+    def CHILD_ACTION(self, queue):
+        try:
+            driver.cuMemPoolDestroy(self.mr.handle)
+        except Exception:  # noqa: S110
+            pass
+        else:
+            self.mr.close()
```
leofang, line 186: ditto

leofang:
FWIW cccl-runtime, C++ stdandad library, or any high-level frameworks/libraries have the same issue. We give you the access to the underlying handle of a container, does not mean that you can free it through `free()` or `delete` behind the container's back. This is UB and by testing it we are guaranteeing certain behavior (whatever it is).

cuda_core/tests/memory_ipc/test_errors.py
cpcloud: I agree that this test is asserting that different levels of APIs offer the same guarantees, which would make developing layers of APIs really really difficult.

It seems roughly analogous to calling into the Python C API through ctypes, and expecting Python to somehow know you didn't mean to cause a segmentation violation:

```
❯ python3.13 -q
>>> x = 1
>>> import ctypes
>>> ctypes.pythonapi.Py_DecRef(x)
zsh: segmentation fault (core dumped)  python3.13 -q
```

cpcloud: I would guess there's probably another way to write this test such that the behavior is buggy without crossing into the C abyss of naked bindings.

Would it be enough to just call `close()` twice? That seems like something we should perhaps be robust to if we're not already:

```
❯ python -q
>>> f = open('/tmp/x', 'w')
>>> f.close()
>>> f.close()
```

cuda_core/tests/memory_ipc/test_errors.py
Andy-Jost: Thanks, that's good to know.

cuda_core/tests/memory_ipc/test_errors.py
Andy-Jost: There seems to be a misunderstanding. The point of this test is to check that errors occurring in `close` are raised rather than suppressed. The exact mechanism I chose to get the driver to generate an error is beside the point, If anyone sees a simpler mechanism, please point it out.

I ran into this issue (`close` not raising errors) when working with the driver bug [5570902](https://nvbugspro.nvidia.com/bug/5570902). I don't want to rely on that behavior for the test because if the driver team ever fixes the bug, the test would then break.

Let's not confuse the issue: we are not testing our robustness in the face of nonsense behavior, i.e., someone stomping around in the driver API directly. This test just does something otherwise dumb and unsupported to trigger an error, which is totally reasonable for this test.

@cpcloud  our code already tolerates double-calls to `close` and simply ignores the second call without error.

Andy-Jost:
> FWIW cccl-runtime, C++ stdandad library, or any high-level frameworks/libraries have the same issue. We give you the access to the underlying handle of a container, does not mean that you can free it through `free()` or `delete` behind the container's back. This is UB and by testing it we are guaranteeing certain behavior (whatever it is).

Respectfully, I think this misses the point. When implicitly closing resources via `__del__` or `__dealloc__`, there may be good reasons to suppress errors. However, when calling a function such as `close` directly, errors should not be suppressed.

cuda_core/tests/memory_ipc/test_errors.py
leofang: There is no misunderstanding. Undefined behavior should not be tested. It is just confusing and sends a wrong signal.

re: on allowing the behavior that calling `.close()` is allowed to raise while invoking the destructor is not, I did raise a discussion with @pciolkosz on the Tuesday meeting. It is indeed nice to have, but it is unclear semantically what we can do about the exception, and technically it does not seem possible to implement in both C++ and Python (`buffer.destroy(stream)` from cccl-runtime is also `noexcept`). I suggest we table this discussion for later, and close this PR.

PR #1123, opened 2025-10-10
Fix #1122: Reinstate graphics APIs

### LINKED ISSUE #1122: [BUG]: cudaGraphicsGLRegisterImage returns null handle with cuda-bindings 12.9.3
### Is this a duplicate?

### Type of Bug

Silent Failure

### Component

cuda.bindings

### Describe the bug

On cuda-bindings 12.9.3, `cudaGraphicsGLRegisterImage` reports `cudaSuccess` but returns a null handle.
This does not happen with cuda-bindings 12.9.2

### How to Reproduce

PySide6 is used to create the OpenGL context and texture.

```python
from cuda.bindings import runtime as cudart
from PySide6 import QtGui, QtOpenGL

class GLWidget(QtOpenGL.QOpenGLWindow):
    def initializeGL(self):
        self.m_texture = QtOpenGL.QOpenGLTexture(QtOpenGL.QOpenGLTexture.Target.Target2D)
        self.m_texture.setFormat(QtOpenGL.QOpenGLTexture.TextureFormat.RGBA8_UNorm)
        self.m_texture.setSize(512, 512)
        self.m_texture.allocateStorage()

        err, self.gfx_resource = cudart.cudaGraphicsGLRegisterImage(
            self.m_texture.textureId(), self.m_texture.target().value,
            cudart.cudaGraphicsRegisterFlags.cudaGraphicsRegisterFlagsWriteDiscard
        )
        error_name = cudart.cudaGetErrorName(err)[1].decode()
        print(error_name, self.gfx_resource)

app = QtGui.QGuiApplication([])
win = GLWidget()
win.show()

# 12.9.2: cudaSuccess <cudaGraphicsResource_t 0x1b75dca8e90>
# 12.9.3: cudaSuccess <cudaGraphicsResource_t 0x0>
```

### Expected behavior

`cudaGraphicsGLRegisterImage` should return a valid handle upon `cudaSuccess`

### Operating System

_No response_

### nvidia-smi output

_No response_

mdboom:
These inadvertently got commented out in #914.

### DIFF
--- a/cuda_bindings/tests/test_graphics_apis.py
+++ b/cuda_bindings/tests/test_graphics_apis.py
@@ -11,7 +11,7 @@ def test_graphics_api_smoketest():
         import pyglet
 
         tex = pyglet.image.Texture.create(512, 512)
-    except ImportError:
+    except (ImportError, AttributeError):
         pytest.skip("pyglet not available or could not create GL context")
         # return to make linters happy
         return


cuda_bindings/tests/test_graphics_apis.py
```diff
@@ -0,0 +1,43 @@
+# SPDX-FileCopyrightText: Copyright (c) 2021-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
leofang: nit:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

cuda_bindings/tests/test_graphics_apis.py
```diff
@@ -0,0 +1,43 @@
...
+    import PySide6
+except ImportError:
+    PySide6 = None
+
+import pytest
+
+
+@pytest.mark.skipif(PySide6 is None, reason="PySide6 not installed")
+def test_graphics_api_smoketest():
+    from PySide6 import QtGui, QtOpenGL
```
cpcloud: Seems like we can avoid a bunch of the noisy skipif and imports by using `importorskip` here:

```suggestion
    PySide6 = pytest.importorskip("PySide6")
    QtGui = PySide6.QtGui
    QtOpenGL = PySide6.QtOpenGL    
```

mdboom: I didn't know about the `importorskip` trick.

Still need to import the subpackages explicitly, though.  `PySide6` doesn't import everything automatically.

cpcloud: You can pass a dotted path to `importorskip`, making this code even shorter:

```
QtGui = pytest.importorksip("PySide6.QtGui")
# etc
```

cpcloud: Didn't realize they were subpackages, otherwise I would've made that my original suggestion!

cuda_bindings/tests/test_graphics_apis.py
```diff
@@ -0,0 +1,37 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
+
+import pytest
+from cuda.bindings import runtime as cudart
+
+
+def test_graphics_api_smoketest():
+    _ = pytest.importorskip("PySide6")
```
rparolin: I don't love skippable tests because I feel too much complexity is introduced from the lack of symmetry between local and ci builds.  For example, there is a potential for the tests to bit-rot if they aren't run frequently enough by devs or the ci runners.

mdboom:
I have updated this to use the `pyglet`-based test as [suggested by @pijyoi](https://github.com/NVIDIA/cuda-python/issues/1122#issuecomment-3392871842) (and pyglet is small enough, I think it's fine to add a a test-only dependency) and added a the "always expected to fail" test as suggested by @rparolin.  I think this is ready to go.

rwgk:
FYI

See https://github.com/NVIDIA/cuda-python/pull/1210#issuecomment-3483899752

`test_graphics_apis.py::test_graphics_api_smoketest` is **SKIPPED in all CI jobs**.

This isn't critical at the moment. However, SWQA stumbled over this test, PR #1210 is for resolving that issue.

PR #1127, opened 2025-10-11
Skipping VirtualMemoryResource tests on WSL

rparolin:
Skipping VirtualMemoryResource tests on WSL

When running the VirtualMemoryResources tests under the Windows Subsystem for Linux (WSL) allocation requests throw the following error:
```
>   raise CUDAError(f"{name.decode()}: {expl}")
E   cuda.core.experimental._utils.cuda_utils.CUDAError: CUDA_ERROR_INVALID_DEVICE: This indicates that the device ordinal supplied by the user does not correspond to a valid CUDA device or that the action requested is invalid for the specified device.

cuda/core/experimental/_utils/cuda_utils.pyx:78: CUDAError
```
See known limitations for WSL2.
https://docs.nvidia.com/cuda/wsl-user-guide/index.html#known-limitations-for-linux-cuda-applications

### DIFF
--- a/cuda_core/tests/test_memory.py
+++ b/cuda_core/tests/test_memory.py
@@ -28,7 +28,7 @@ from cuda.core.experimental._memory import DLDeviceType, IPCBufferDescriptor
 from cuda.core.experimental._utils.cuda_utils import handle_return
 from cuda.core.experimental.utils import StridedMemoryView
 
-from cuda_python_test_helpers import supports_ipc_mempool, IS_WSL
+from cuda_python_test_helpers import IS_WSL, supports_ipc_mempool
 
 POOL_SIZE = 2097152  # 2MB size
 


leofang:
@rparolin in what ways VMM raises an error on WSL? Could you show a full traceback to what you see? 

Asking because we were told that VMM is supported by the Windows WDDM driver mode (https://github.com/NVIDIA/cuda-python/pull/968#issuecomment-3379214047), and WSL is a pass-through of WDDM according to @ksimpson-work, and so by deduction it should be supported. This is a bit surprising... 🤔

keenan-simpson:
This is not the right place to be commenting, but skipping tests on windows because something isn't supported in TCC feels heavy handed. a utility function to check if the system is in TCC mode seems more appropriate. RE: VMM I would also be curious to know what the issues on WSL are for VMM, if this isn't already fully resolved

PR #1130, opened 2025-10-14
chore(dev): Run `cython-lint` against codebase, add it to pre-commit

### LINKED ISSUE #941: MNT: Set up a linter for Cython code
We should run cython-lint on hand-crafted Cython code (and skip the auto-generated ones).

MarcoGorelli:
## Description

closes 

closes #941

This PR adds `cython-lint` to `pre-commit` and fixes the flagged lint violations (mostly unused imports and variables)

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -78,5 +78,12 @@ repos:
     hooks:
       - id: actionlint
 
+  - repo: https://github.com/MarcoGorelli/cython-lint
+    rev: "d9ff7ce99ef4f2ae8fba93079ca9d76c4651d4ac"  # frozen: v0.18.0
+    hooks:
+      - id: cython-lint
+        args: [--no-pycodestyle]
+        exclude: ^cuda_bindings/
+
 default_language_version:
       python: python3
--- a/cuda_core/cuda/core/experimental/_context.pyx
+++ b/cuda_core/cuda/core/experimental/_context.pyx
@@ -4,7 +4,6 @@
 
 from dataclasses import dataclass
 
-from cuda.core.experimental._utils.clear_error_support import assert_type
 from cuda.core.experimental._utils.cuda_utils import driver
 
 
--- a/cuda_core/cuda/core/experimental/_memory.pyx
+++ b/cuda_core/cuda/core/experimental/_memory.pyx
@@ -20,8 +20,6 @@ from cuda.core.experimental._utils.cuda_utils cimport (
 )
 
 import abc
-import array
-import contextlib
 import cython
 from dataclasses import dataclass, field
 from typing import Iterable, Literal, Optional, TYPE_CHECKING, TypeVar, Union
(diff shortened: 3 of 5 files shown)


mdboom:
> * Fix these violations in `cython-gen` and `cybind` (the two generator tools that contribute these files, then regenerate and commit the result here.

Actually, it looks like just `cybind` is implicated here -- no changes to `cython-gen` required, but point still stands.

leofang:
Thanks, @MarcoGorelli! Big fan of your projects (including `cython-lint`)! 🙏

Unfortunately, in addition to the concern that Mike noted (that `cuda-bindings` code are mostly autogenerated and linting them makes little sense), we have a bigger issue, which is the legal aspect. Everywhere else in this repo accepts external contribution, except for files under the `cuda_bindings/` folder 😅 because it does not have an OSS license. https://nvidia.github.io/cuda-python/cuda-bindings/latest/contribute.html. 

Would you mind rebasing and purging all changes to `cuda_bindings/`?

MarcoGorelli:
that's very kind, thank you!

sure, I reset the commit add excluded `cuda_bindings`

leofang:
/ok to test [1fd64c8](https://github.com/NVIDIA/cuda-python/pull/1130/commits/1fd64c8cce6c8a7435ab683e5eb212f2f731fee5)

leofang (approved):

mdboom (approved):
Thanks!

rwgk:
In passing I noticed this error when adding the `--verbose` option:

```
pre-commit run --all-files --verbose
```

```
cython-lint..............................................................Passed
- hook id: cython-lint
- duration: 0.48s

Skipping file /home/rgrossekunst/forked/cuda-python/cuda_core/cuda/core/experimental/_device.pyx, as it cannot be parsed. Error: AttributeError("'_thread._local' object has no attribute 'cython_errors_listing_file'")
```

(There is no traceback.)

Maybe a bug in `cython-lint`?

MarcoGorelli:
hi @rwgk 

that means that Cython's `parse_from_strings` raised when parsing it

https://github.com/MarcoGorelli/cython-lint/blob/b9cf4701d98b34a7d5e7adbd9e446f97dd9700a4/cython_lint/cython_lint.py#L403-L412

I'd suggest reporting this to Cython itself

rwgk:
For easy reference, this is the existing `cython-lint` code:

```python
    except Exception as exp:  # pragma: no cover
        # If Cython can't parse this file, just skip it.
        print(
            f"Skipping file {filename}, as it cannot be parsed. Error: {exp!r}",
        )
        raise CythonParseError from exp
```

Unfortunately the original error and the traceback don't appear. → I'm giving up (and I think most people will give up) right there.

WDYT about (totally untested and with a huge amount of guessing):

```python
    except Exception as exp:  # pragma: no cover
        # If Cython can't parse this file, just skip it.
        sys.stderr.flush()
        print(
            f"Skipping file {filename}, as it cannot be parsed. Error: {exp!r}",
        )
        traceback.print_exc(file=sys.stdout)
        sys.stdout.flush()
        raise CythonParseError from exp
```

Will that surface the `cython` error? (That's what I'm hoping to achieve.)

I'm also wondering: `_device.pyx` builds successfully (the output pasted before is with cuda-python `main` as-is). Why does the parsing fail with `cython-lint`?

MarcoGorelli:
sure, i've reported it here: https://github.com/cython/cython/issues/7235

PR #1131, opened 2025-10-14
Update the CUDA runtime to 13.0.2

mdboom:
## Description

Fixes #1129.

Updates the statically linked CUDA runtime to 13.0.2

I used https://github.com/NVIDIA/cuda-python/pull/960 as a template for this.

## Checklist
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #1132, opened 2025-10-14
Add a script to help update to a new CTK version

mdboom:
This is the script I used to update to a new CTK version and create #1131.  I think this will be helpful going forward.  Worth adding to `toolshed`?

### DIFF
--- a/toolshed/update_ctk.py
+++ b/toolshed/update_ctk.py
@@ -150,16 +150,58 @@ def update_matrix(existing_version: str, new_version: str, matrix_path: Path) ->
         f.write(content)
 
 
-def main(version: str, cuda_python_repo: Path, cybind_repo: Path, is_prev: bool):
+def regenerate_cython_gen(cuda_python_private_repo: Path, cuda_python_repo: Path) -> int:
+    with tempfile.TemporaryDirectory() as tempdir:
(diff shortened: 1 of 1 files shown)


rwgk:
Looks useful! But I think it's important to have `cybind` in the name, maybe `update_ctk_cybind.py`?

This is to distinguish from `cython-gen` (which already has a `regenerate.py` script).

mdboom:
> This is to distinguish from `cython-gen` (which already has a `regenerate.py` script).

Hmm...  I actually didn't think of `cython-gen` here for this one because the last time this was manually done (in #960), none of the `cython-gen`-generated files needed updating.  But for completeness, this should probably /also/ regenerate with `cython-gen` (by running that `regenerate.py` script).  That's a trickier thing to automate because it requires /installing/ the CTK, rather than just fetching headers from the web, as this does.

What do others think?

rwgk:
> That's a trickier thing to automate because it requires /installing/ the CTK, rather than just fetching headers from the web, as this does.

For this time around, I'll try on the colossus workstation that I have already. I think it'll take me only a few minutes.

Not sure what'll be best long term.

mdboom:
Moving elsewhere.

PR #1133, opened 2025-10-14
Adds the cython-gen generated changes for upgrade to CUDA 13.0.2

mdboom:
As @rwgk [pointed out](https://github.com/NVIDIA/cuda-python/pull/1132#issuecomment-3402447062), we also need to update the `cython-gen` -generated files when updating the CTK version.  There was a docstring bugfix in those headers between 13.0.1 and 13.0.2 that is now reflected here.


rwgk:
Oh, sorry, I'm seeing this only now.

See #1134, which goes up all the way to 13.0.2

mdboom:
Closing in favor of #1134.  I had initially put the wrong version in the title -- I definitely generated with 13.0.2, but didn't get as many changes as @rwgk did.  Probably something with my local environment.  In the meantime, lets just use #1134.

PR #1137, opened 2025-10-14
[FIX]: Break infinite recursion in `Struct.discoverMembers()`

rwgk:
Issue discovered while testing with `ctk-next`.

The core fix is the `seen` code.

The stripping of qualifiers/pointers was invented by ChatGPT. Should we keep it?

`flush=True` systematically added to all `print()`s, for a better user experience when piping through `tee`.

### DIFF
--- a/cuda_bindings/setup.py
+++ b/cuda_bindings/setup.py
@@ -114,15 +114,28 @@ class Struct:
             self._member_names += [var_name]
             self._member_types += [var_type]
 
-    def discoverMembers(self, memberDict, prefix):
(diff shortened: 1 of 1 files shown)


rwgk:
@vzhurba01 Are you the original author of the `Struct.discoverMembers()` code?

The stripping of qualifiers/pointers was invented by ChatGPT. Should we keep it?

rwgk:
Thanks @mdboom! — I just double-checked that this is still in sync with the ctk-next branch.

PR #1145, opened 2025-10-16
Implement IPC-enabled events.

### LINKED ISSUE #1040: Enable IPC events and update tests to test stream-ordered IPC
We probably don't have time to cover it in cuda.core 0.4.0, but we need to at least confirm that a future inclusion does not break the new IPC APIs. If so, this is a blocking issue that must be resolved before code freeze.

Andy-Jost:
## Description
Adds construction keywords to `Event` to allow creation of IPC-enabled events. Adds reduce methods to allow events to be sent to other processes. Updates tests.

Closes #1040

### DIFF
--- a/.github/workflows/bandit.yml
+++ b/.github/workflows/bandit.yml
@@ -24,6 +24,8 @@ jobs:
 
       - name: Install uv
         uses: astral-sh/setup-uv@d0cc045d04ccac9d8b7881df0226f9e82c39688e  # v6.8.0
+        with:
+          enable-cache: false
 
       - name: Get ignore codes
         id: ignore-codes
--- a/cuda_bindings/cuda/bindings/_internal/cufile.pxd
+++ b/cuda_bindings/cuda/bindings/_internal/cufile.pxd
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
 #
-# This code was automatically generated across versions from 12.9.0 to 13.0.2. Do not modify it directly.
+# This code was automatically generated across versions from 12.9.1 to 13.0.2. Do not modify it directly.
 
 from ..cycufile cimport *
 
@@ -18,6 +18,7 @@ cdef CUfileError_t _cuFileBufDeregister(const void* bufPtr_base) except?<CUfileE
 cdef ssize_t _cuFileRead(CUfileHandle_t fh, void* bufPtr_base, size_t size, off_t file_offset, off_t bufPtr_offset) except* nogil
 cdef ssize_t _cuFileWrite(CUfileHandle_t fh, const void* bufPtr_base, size_t size, off_t file_offset, off_t bufPtr_offset) except* nogil
 cdef CUfileError_t _cuFileDriverOpen() except?<CUfileError_t>CUFILE_LOADING_ERROR nogil
+cdef CUfileError_t _cuFileDriverClose() except?<CUfileError_t>CUFILE_LOADING_ERROR nogil
 cdef CUfileError_t _cuFileDriverClose_v2() except?<CUfileError_t>CUFILE_LOADING_ERROR nogil
 cdef long _cuFileUseCount() except* nogil
 cdef CUfileError_t _cuFileDriverGetProperties(CUfileDrvProps_t* props) except?<CUfileError_t>CUFILE_LOADING_ERROR nogil
@@ -41,7 +42,6 @@ cdef CUfileError_t _cuFileGetParameterString(CUFileStringConfigParameter_t param
 cdef CUfileError_t _cuFileSetParameterSizeT(CUFileSizeTConfigParameter_t param, size_t value) except?<CUfileError_t>CUFILE_LOADING_ERROR nogil
 cdef CUfileError_t _cuFileSetParameterBool(CUFileBoolConfigParameter_t param, cpp_bool value) except?<CUfileError_t>CUFILE_LOADING_ERROR nogil
 cdef CUfileError_t _cuFileSetParameterString(CUFileStringConfigParameter_t param, const char* desc_str) except?<CUfileError_t>CUFILE_LOADING_ERROR nogil
-cdef CUfileError_t _cuFileDriverClose() except?<CUfileError_t>CUFILE_LOADING_ERROR nogil
 cdef CUfileError_t _cuFileGetParameterMinMaxValue(CUFileSizeTConfigParameter_t param, size_t* min_value, size_t* max_value) except?<CUfileError_t>CUFILE_LOADING_ERROR nogil
 cdef CUfileError_t _cuFileSetStatsLevel(int level) except?<CUfileError_t>CUFILE_LOADING_ERROR nogil
 cdef CUfileError_t _cuFileGetStatsLevel(int* level) except?<CUfileError_t>CUFILE_LOADING_ERROR nogil
--- a/cuda_bindings/cuda/bindings/_internal/cufile_linux.pyx
+++ b/cuda_bindings/cuda/bindings/_internal/cufile_linux.pyx
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
 #
-# This code was automatically generated across versions from 12.9.0 to 13.0.2. Do not modify it directly.
+# This code was automatically generated across versions from 12.9.1 to 13.0.2. Do not modify it directly.
 
 from libc.stdint cimport intptr_t, uintptr_t
 import threading
@@ -68,6 +68,7 @@ cdef void* __cuFileBufDeregister = NULL
 cdef void* __cuFileRead = NULL
 cdef void* __cuFileWrite = NULL
 cdef void* __cuFileDriverOpen = NULL
+cdef void* __cuFileDriverClose = NULL
 cdef void* __cuFileDriverClose_v2 = NULL
 cdef void* __cuFileUseCount = NULL
 cdef void* __cuFileDriverGetProperties = NULL
@@ -91,7 +92,6 @@ cdef void* __cuFileGetParameterString = NULL
 cdef void* __cuFileSetParameterSizeT = NULL
 cdef void* __cuFileSetParameterBool = NULL
 cdef void* __cuFileSetParameterString = NULL
-cdef void* __cuFileDriverClose = NULL
 cdef void* __cuFileGetParameterMinMaxValue = NULL
 cdef void* __cuFileSetStatsLevel = NULL
 cdef void* __cuFileGetStatsLevel = NULL
@@ -166,6 +166,13 @@ cdef int __check_or_init_cufile() except -1 nogil:
             if handle == NULL:
                 handle = load_library()
             __cuFileDriverOpen = dlsym(handle, 'cuFileDriverOpen')
+
+        global __cuFileDriverClose
+        __cuFileDriverClose = dlsym(RTLD_DEFAULT, 'cuFileDriverClose')
+        if __cuFileDriverClose == NULL:
+            if handle == NULL:
+                handle = load_library()
+            __cuFileDriverClose = dlsym(handle, 'cuFileDriverClose')
 
         global __cuFileDriverClose_v2
         __cuFileDriverClose_v2 = dlsym(RTLD_DEFAULT, 'cuFileDriverClose_v2')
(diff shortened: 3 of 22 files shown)


cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -151,6 +166,40 @@ cdef class Event:
...
+        cdef cydriver.CUipcEventHandle data
+        memcpy(data.reserved, <const void*><const char*>(ipc_descriptor._reserved), sizeof(data.reserved))
+        cdef Event self = Event.__new__(cls)
+        with nogil:
+            HANDLE_RETURN(cydriver.cuIpcOpenEventHandle(&self._handle, data))
+        self._timing_disabled = True
+        self._busy_waited = ipc_descriptor._busy_waited
+        self._ipc_enabled = True
+        self._ipc_descriptor = ipc_descriptor
+        self._device_id = -1  # ??
```
Andy-Jost, line 193: Events hold a device and context handle. I could not find these being used anywhere except the property getters in this class. For imported events, these are not available. The current implementation returns None for these properties for IPC-imported events.

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -86,24 +90,35 @@ cdef class Event:
...
+            flags |= cydriver.CUevent_flags.CU_EVENT_INTERPROCESS
+            self._ipc_enabled = True
+            if not self._timing_disabled:
+                raise TypeError("IPC-enabled events cannot use timing.")
         with nogil:
             HANDLE_RETURN(cydriver.cuEventCreate(&self._handle, flags))
         self._device_id = device_id
         self._ctx_handle = ctx_handle
+        if opts.ipc_enabled:
+            self.get_ipc_descriptor()
```
Andy-Jost, line 120: `multiprocessing` serializes arguments in a separate thread. If `get_ipc_descriptor` is called for the first time during serialization, then `cuIpcGetEventHandle` raises an error saying no CUDA context is bound to that thread. The best solution I found is this: if an event is created with IPC support, then create the descriptor right away (in the main thread) and cache it.

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -40,15 +44,15 @@ cdef class EventOptions:
         has actually been completed.
         Otherwise, the CPU thread will busy-wait until the event has
         been completed. (Default to False)
-    support_ipc : bool, optional
```
Andy-Jost, line 43: Renamed `support_ipc` --> `ipc_enabled` to match the memory IPC API. This should not be a compatibility issue because `support_ipc` was never supported.

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -86,24 +90,35 @@ cdef class Event:
...
         if not opts.enable_timing:
             flags |= cydriver.CUevent_flags.CU_EVENT_DISABLE_TIMING
             self._timing_disabled = True
         if opts.busy_waited_sync:
             flags |= cydriver.CUevent_flags.CU_EVENT_BLOCKING_SYNC
             self._busy_waited = True
-        if opts.support_ipc:
-            raise NotImplementedError("WIP: https://github.com/NVIDIA/cuda-python/issues/103")
+        if opts.ipc_enabled:
+            if is_free:
```
Andy-Jost, line 107: This is related to making IPC-enabled events race-free. See `test_event_is_monadic` in `memory_ipc/test_event_ipc.py` for details.

cuda_core/cuda/core/experimental/_event.pyx
Andy-Jost: I'm curious about the intended use of the device and context. Do we need setters for these?

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -212,12 +256,43 @@ cdef class Event:
...
+        if self._ctx_handle is not None and self._device_id >= 0:
+            return Context._from_ctx(self._ctx_handle, self._device_id)
+
+
+cdef class IPCEventDescriptor:
+    """Serializable object describing an event that can be shared between processes."""
+
+    cdef:
+        bytes _reserved
+        bint _busy_waited
```
Andy-Jost, line 274: I send the `_busy_waited` property to the child because there is no driver API to query it. I failed to devise a test that actually checks whether an event blocks or busy-waits in the child process, so I cannot confirm whether the property is accurate.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -511,7 +516,7 @@ cdef class DeviceMemoryResourceOptions:
         (Default to 0)
     """
     ipc_enabled : cython.bint = False
-    max_size : cython.int = 0
+    max_size : cython.size_t = 0
```
Andy-Jost, line 514: This is a bug fix. cython.int is 32 bits so the max pool size was limited to ~2GB. This unfortunately went out in the release.

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -260,7 +262,13 @@ cdef class Stream:
         # and CU_EVENT_RECORD_EXTERNAL, can be set in EventOptions.
         if event is None:
             self._get_device_and_context()
-            event = Event._init(<int>(self._device_id), <uintptr_t>(self._ctx_handle), options)
+            event = Event._init(<int>(self._device_id), <uintptr_t>(self._ctx_handle), options, False)
+        elif event.is_ipc_enabled:
```
Andy-Jost, line 264: This is part of enforcing monadicity.

cuda_core/tests/helpers/__init__.py
```diff
@@ -22,7 +22,7 @@
     import cuda_python_test_helpers
 except ImportError:
     # Import shared platform helpers for tests across repos
-    sys.path.insert(0, str(pathlib.Path(__file__).resolve().parents[2] / "cuda_python_test_helpers"))
+    sys.path.insert(0, str(pathlib.Path(__file__).resolve().parents[3] / "cuda_python_test_helpers"))
```
Andy-Jost, line 25: I made `helpers.py` into a package and merged the IPC `utility.py` into it.

cuda_core/tests/memory_ipc/test_event_ipc.py
```diff
@@ -0,0 +1,181 @@
...
+        log("enqueuing copy on stream2")
+        buffer.copy_from(twos, stream=stream2)
+        log("signaling parent")
+        q_out.put(None)
+        log("waiting")
+        stream2.sync()
+        log("done")
+
+
+def test_event_is_monadic(ipc_device):
```
Andy-Jost, line 91: We can discuss this design decision. My view is that it's better for users to potentially trip over these limitations (~5 minute fix) rather than risk creating a race with mutable events (hours or days).

cuda_core/tests/memory_ipc/test_memory_ipc.py
```diff
@@ -16,14 +13,11 @@
...
     def test_main(self, ipc_device, ipc_memory_resource):
-        if not supports_ipc_mempool(ipc_device):
-            import pytest
-
-            pytest.skip("Driver rejects IPC-enabled mempool creation on this platform")
         """Test IPC with memory pools."""
         # Set up the IPC-enabled memory pool and share it.
         device = ipc_device
         mr = ipc_memory_resource
+        pgen = PatternGen(device, NBYTES)
```
Andy-Jost, line 21: There are lots of changes coming from the rename of `IPCBufferTestHelper` to `PatternGen`, but the new class is much more sane and useful.

cuda_core/tests/test_event.py
```diff
@@ -121,60 +114,24 @@ def test_error_timing_recorded():
...
 
-@skipif_need_cuda_headers  # libcu++
-@pytest.mark.skipif(tuple(int(i) for i in np.__version__.split(".")[:2]) < (2, 1), reason="need numpy 2.1.0+")
 def test_error_timing_incomplete():
     device = Device()
     device.set_current()
-
-    # This kernel is designed to busy loop until a signal is received
-    code = """
-#include <cuda/atomic>
```
Andy-Jost, line 132: Factored this out to `helpers/latch.py`

cuda_core/tests/test_memory.py
```diff
@@ -95,30 +96,6 @@ def device_id(self) -> int:
         raise RuntimeError("the pinned memory resource is not bound to any GPU")
 
 
-class DummyUnifiedMemoryResource(MemoryResource):
```
Andy-Jost, line 98: Moved to `helpers/buffers.py`

cuda_core/tests/memory_ipc/test_memory_ipc.py
rwgk: Could it make sense to break the rename out as a separate PR?

Andy-Jost: I'm open to it. I think there's not much risk of this change causing trouble for being overly complex, because the actual changes to cuda.core are so small. IMO, moving to the next project has more value but I'm not set in stone.

rwgk: > because the actual changes to cuda.core are so small.

That's very difficult for me to see at the moment. Let's discuss offline. I don't want to generate extra work.

cuda_core/cuda/core/experimental/_device.pyx
```diff
@@ -28,6 +28,7 @@ from cuda.core.experimental._utils.cuda_utils import (
 from cuda.core.experimental._stream cimport default_stream
 
 
+
```
rparolin, line 31: Necessary whitespace?

cuda_core/cuda/core/experimental/_event.pxd
```diff
@@ -11,6 +11,8 @@ cdef class Event:
         cydriver.CUevent _handle
```
rparolin, line 11: Does cython throw warnings when data members inject padding overhead that could be eliminated by reordering members?

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -151,6 +165,40 @@ cdef class Event:
                 raise CUDAError(err)
             raise RuntimeError(explanation)
 
+    def get_ipc_descriptor(self) -> IPCEventDescriptor:
+        """Export an event allocated for sharing between processes."""
+        if self._ipc_descriptor is not None:
+            return self._ipc_descriptor
+        if not self.is_ipc_enabled:
```
rparolin, line 172: if `is_ipc_enabled` is `false` would we want this error to fire even if has some value in the `_ipc_descriptor` field?

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -151,6 +165,40 @@ cdef class Event:
...
+            HANDLE_RETURN(cydriver.cuIpcGetEventHandle(&data, <cydriver.CUevent>(self._handle)))
+        cdef bytes data_b = cpython.PyBytes_FromStringAndSize(<char*>(data.reserved), sizeof(data.reserved))
+        self._ipc_descriptor = IPCEventDescriptor._init(data_b, self._busy_waited)
+        return self._ipc_descriptor
+
+    @classmethod
+    def from_ipc_descriptor(cls, ipc_descriptor: IPCEventDescriptor) -> Event:
+        """Import an event that was exported from another process."""
+        cdef cydriver.CUipcEventHandle data
+        memcpy(data.reserved, <const void*><const char*>(ipc_descriptor._reserved), sizeof(data.reserved))
```
rparolin, line 185: is the explicit case to `void*` required here? It should be implicitly convertible for pointer types.

cuda_core/tests/helpers/latch.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class LatchKernel:
+    """
+    Manages a kernel that blocks stream progress until released.
+    """
+
+    def __init__(self, device):
+        if helpers.CUDA_INCLUDE_PATH is None:
+            pytest.skip("need CUDA header")
```
rparolin, line 25: Can we provide more information to the user if they hit this?  What would their next steps be?

cuda_core/cuda/core/experimental/_device.pyx
Andy-Jost: Probably not. I assume the pre-commit hooks will reformat as needed, but maybe not in this case?

cuda_core/cuda/core/experimental/_event.pxd
Andy-Jost: I don't know. FYI, CUevent is 32 bytes IIRC

cuda_core/cuda/core/experimental/_event.pyx
Andy-Jost: It should not be possible to fill the `_ipc_descriptor` field in that case. It is only ever set non-None below, in this function, after the check.

cuda_core/tests/helpers/latch.py
Andy-Jost: I'm not sure, I moved this out of the existing test_events.py test.

PR #1153, opened 2025-10-17
[no-ci] [chore] Fix the actionlint so shellcheck doesn't run even if installed

mdboom:
`actionlint` runs `shellcheck` if it's installed.  However, our Github Actions `run` sections don't currently conform to `shellcheck`.  We can decide later if we want to do that, but in the meantime, this disables `shellcheck` so the pre-commit hooks don't fail if it's installed.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -73,11 +73,11 @@ repos:
         files: ^cuda_pathfinder/cuda/.*\.py$  # Exclude tests directory
         args: [--config-file=cuda_pathfinder/pyproject.toml]
 
-  # TODO: Reinstate after the workflows have been corrected.  See #1146
-  # - repo: https://github.com/rhysd/actionlint
-  #   rev: "03d0035246f3e81f36aed592ffb4bebf33a03106"  # frozen: v1.7.7
-  #   hooks:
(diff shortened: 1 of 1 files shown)


cpcloud:
It's actually `no-ci` because of GHA special-casing `skip-ci` and a few other tokens to bypass the ... bypassing of checks using skip-as-success 😮‍💨

mdboom:
> It's actually `no-ci` because of GHA special-casing `skip-ci` and a few other tokens to bypass the ... bypassing of checks using skip-as-success 😮‍💨

Ah, no worries.  It doesn't seem to pick that up either (maybe because it was changed after the fact).  Any ideas about how to "kick it"?

cpcloud:
Hm ... not sure how to kick it. Perhaps close and reopen?

mdboom:
> Hm ... not sure how to kick it. Perhaps close and reopen?

Yeah, that didn't seem to do the trick.  Just gonna run the tests in this instance.

mdboom:
@cpcloud: Ah, I was just confused about how it worked.  You still have to run the tests with `/ok to test`, but then it skips most of the hard work, tests pass and merging becomes available.

PR #1154, opened 2025-10-17
[chore] Enable ccache in builds

mdboom:
This enables ccache on our non-Windows builds.  This should make builds significantly faster.

(I'm going to run CI twice here to confirm it works as advertised.)

### DIFF
--- a/.github/workflows/build-wheel.yml
+++ b/.github/workflows/build-wheel.yml
@@ -46,11 +46,10 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Setup ccache
-        uses: hendrikmuhs/ccache-action@bfa03e1de4d7f7c3e80ad9109feedd05c4f5a716
+      - name: Intall ccache
         if: ${{ !startsWith(inputs.host-platform, 'win') }}
-        with:
-          create-symlink: true
+        run:
+          sudo apt install -y ccache
 
       - name: Setup proxy cache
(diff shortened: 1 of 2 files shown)


cpcloud (approved):

cpcloud:
Any reason not to use https://github.com/mozilla/sccache (and https://github.com/mozilla/sccache-action) which seems to support windows?

mdboom:
> Any reason not to use https://github.com/mozilla/sccache (and https://github.com/mozilla/sccache-action) which seems to support windows?

Nah, I just wasn't aware of it.  It used to require a server component, but it looks like it defaults to local storage.  I'll file an alternative PR for that and we can compare.

mdboom:
Also, it looks like the workflow is getting stuck here.  I wonder if it's because this PR modifies a workflow and there is an additional approval required for the bot (or something)?

leofang:
This action is not yet approved by our admin, need to reach out to them first:
https://github.com/NVIDIA/cuda-python/actions/runs/18597054102

mdboom:
> This action is not yet approved by our admin, need to reach out to them first: https://github.com/NVIDIA/cuda-python/actions/runs/18597054102

Thanks.  That explains it.  Let's not reach out just yet -- the alternative of using sccache that @cpcloud suggested seems to be already approved, and if it works, we might as well go with it: https://github.com/NVIDIA/cuda-python/pull/1156

PR #1156, opened 2025-10-17
[chore] Use sccache in builds

mdboom:
This is an alternative to #1154.  Experimenting here and we should pick only one.

### DIFF
--- a/.github/workflows/build-wheel.yml
+++ b/.github/workflows/build-wheel.yml
@@ -76,41 +76,6 @@ jobs:
         if: ${{ startsWith(inputs.host-platform, 'win') }}
         uses: ilammy/msvc-dev-cmd@v1  # TODO: ask admin to allow pinning commits
 
-      # at this point, cl.exe already exists in PATH
-      - name: Find cl.exe and set CL_EXE environment variable
-        if: ${{ startsWith(inputs.host-platform, 'win') }}
-        run: |
-          # Find the full path of cl.exe
-          CL_EXE=$(which cl.exe)
-          if [ -z "${CL_EXE}" ]; then
-            echo "Error: cl.exe not found in PATH"
-            exit 1
-          fi
-          echo "Found cl.exe at: ${CL_EXE}"
-          echo "CL_EXE=$(cygpath -w "${CL_EXE}")" >> $GITHUB_ENV
-
-      #- name: Generate sccache wrapper script
-      #  if: ${{ startsWith(inputs.host-platform, 'win') }}
-      #  run: |
-      #    # Create a wrapper directory
-      #    mkdir -p sccache_wrapper
-      #    # Generate the wrapper script named cl.exe
-      #    cat > sccache_wrapper/cl.exe <<'EOF'
-      #    @echo off
-      #    sccache "%CL_EXE%" %*
-      #    EOF
-      #    cat sccache_wrapper/cl.exe
-      #    # Record the wrapper path for later use
-      #    echo "SCCACHE_WRAPPER_DIR=$(realpath "./sccache_wrapper")" >> $GITHUB_ENV
-
-      - name: Replace cl.exe by sccache.exe
-        if: ${{ startsWith(inputs.host-platform, 'win') }}
-        run: |
-          ls -lt "${{ env.CL_EXE }}"
-          cp "${{ env.SCCACHE_PATH }}" "${{ env.CL_EXE }}"
-          ls -lt "${{ env.CL_EXE }}"
-          "${{ env.SCCACHE_PATH }}" --start-server
-
       - name: Set environment variables
         env:
           CUDA_VER: ${{ inputs.cuda-version }}
@@ -190,24 +155,9 @@ jobs:
           CIBW_ENVIRONMENT_WINDOWS: >
             CUDA_PATH="$(cygpath -w ${{ env.CUDA_PATH }})"
             CUDA_PYTHON_PARALLEL_LEVEL=${{ env.CUDA_PYTHON_PARALLEL_LEVEL }}
-            SCCACHE_GHA_ENABLED=true
-            ACTIONS_RUNTIME_TOKEN=${{ env.ACTIONS_RUNTIME_TOKEN }}
-            ACTIONS_RUNTIME_URL=${{ env.ACTIONS_RUNTIME_URL }}
-            ACTIONS_RESULTS_URL=${{ env.ACTIONS_RESULTS_URL }}
-            ACTIONS_CACHE_URL=${{ env.ACTIONS_CACHE_URL }}
-            ACTIONS_CACHE_SERVICE_V2=${{ env.ACTIONS_CACHE_SERVICE_V2 }}
-            SCCACHE_DIR="$(cygpath -w ${{ env.SCCACHE_DIR }})"
-            SCCACHE_CACHE_SIZE=${{ env.SCCACHE_CACHE_SIZE }}
-            SCCACHE_LOG=debug
-          #CL_EXE="${{ env.CL_EXE }}"
-          #PATH="$(cygpath -w "${{ env.SCCACHE_WRAPPER_DIR }}");$PATH"
-          CIBW_BEFORE_ALL_WINDOWS: >
-            echo %PATH%
           # check cache stats before leaving cibuildwheel
           CIBW_BEFORE_TEST_LINUX: >
             "/host/${{ env.SCCACHE_PATH }}" --show-stats
-          CIBW_BEFORE_TEST_WINDOWS: >
-            "${{ env.SCCACHE_PATH }}" --show-stats
           # force the test stage to be run (so that before-test is not skipped)
           # TODO: we might want to think twice on adding this, it does a lot of
           # things before reaching this command.


leofang:
Looking at the post action cleanup log, we don't seem to be building the cache at all. I suspect on Linux it has to do with cibuildwheel launching a manylinux container, and if so we need to mount sccache into the container. But this does not explain the Windows situation, which does not use container and instead we build wheels on the bare VM.

leofang:
Do we need to convince setuptools to prepend `sccache` to `CC` and `CXX`?

mdboom:
> Do we need to convince setuptools to prepend `sccache` to `CC` and `CXX`?

Yeah, it looks like this action (unlike the ccache-based one) doesn't install the symlinks.  The docs have examples setting `CC` and `CXX`.  I'll try that and see if setuptools picks it up (I think it normally does).

ci/tools/env-vars
```diff
@@ -86,3 +86,5 @@ fi
   echo "CUDA_BINDINGS_ARTIFACTS_DIR=$(realpath "${REPO_DIR}/cuda_bindings/dist")"
   echo "CUDA_BINDINGS_CYTHON_TESTS_DIR=$(realpath "${REPO_DIR}/cuda_bindings/tests/cython")"
 } >> $GITHUB_ENV
+
+echo "SCCACHE_PATH=$(which sccache)"
```
cpcloud: If you throw this into the above `{}` you'll get it in the env.

Right now you're just echoing to stdout.

mdboom: Weird.  I agree now that you point that out, but it was working anyway.  Maybe the action is defining this already.

cpcloud: It does: https://github.com/Mozilla-Actions/sccache-action/blob/main/src/setup.ts#L67

So then this line can be removed.

.github/workflows/build-wheel.yml
```diff
@@ -40,12 +40,22 @@ jobs:
     runs-on: ${{ (inputs.host-platform == 'linux-64' && 'linux-amd64-cpu8') ||
                  (inputs.host-platform == 'linux-aarch64' && 'linux-arm64-cpu8') ||
                  (inputs.host-platform == 'win-64' && 'windows-2022') }}
+    env:
+      SCCACHE_GHA_ENABLED: "true"
```
cpcloud: Is this used anywhere?

mdboom: [Docs says it's required for C/C++ support](https://github.com/Mozilla-Actions/sccache-action?tab=readme-ov-file#cc-code).  Not sure why.

.github/workflows/build-wheel.yml
```diff
@@ -130,9 +141,14 @@ jobs:
           CIBW_ENVIRONMENT_LINUX: >
             CUDA_PATH=/host/${{ env.CUDA_PATH }}
             CUDA_PYTHON_PARALLEL_LEVEL=${{ env.CUDA_PYTHON_PARALLEL_LEVEL }}
+            CC="/host${{ env.SCCACHE_PATH }} cc"
+            CXX="/host${{ env.SCCACHE_PATH }} c++"
```
leofang: Q: shouldn't this be
```suggestion
            CC="/host${{ env.SCCACHE_PATH }} gcc"
            CXX="/host${{ env.SCCACHE_PATH }} g++"
```
?

leofang:
I see there's cache now! 🎉  Probably from Windows?
截圖 2025-10-17 下午4 09 58

leofang:
oh wait, it's only 13 bytes? 🤔

.github/workflows/build-wheel.yml
mdboom: This chooses the default compiler.  There are some Linux distros, especially on ARM, where the default compiler is clang.  In our case it may not matter, though.

mdboom:
> oh wait, it's only 13 bytes? 🤔

Yeah, definitely not working yet.

leofang:
> and if so we need to mount sccache into the container. 

It seems we either mount `sccache` from host, or install it as part of cibuildwheel pre-build step. This seems relevant: https://github.com/pypa/cibuildwheel/issues/1030#issuecomment-1691408281.

mdboom:
Well, this turned out to be harder than expected.  Dumping my thoughts for fresh eyes next week:

- The `SCCACHE_GHA_ENABLED` variable signals to `sccache` to store directly to GHA's cached storage.  This requires a series of special environment variables to manage the tokens to do so.  Because these are secret/token related things, GitHub makes it kind of tricky (but not impossible) to expose them inside our nested `cibuildwheel` container.  [While there are ways](https://github.com/orgs/community/discussions/42856), they smell sketchy from a security point of view.  Without those special values, starting the sccache server inside the container fails.
- I tried explicitly starting the sccache server outside of the container first, hoping to get the client inside the container to use it, but the port isn't forwarded into the container.  Docker has a config for this kind of port forwarding, but I don't think `cibuildwheel` allows for passing arbitrary arguments to Docker.
- What I haven't tried yet is /not/ setting `SCCACHE_GHA_ENABLED` and getting `sccache` to fall back to its local filesystem mode, and then explicitly caching its local directory ourselves at the end of the run.  In this case, the `sccache` github action will be working against us (since it assumes working low-level access to the GitHub caching API) - I think we would just have to do it ourselves.

For this last option, we really may be better off using the much less magical `ccache` instead (and without the GitHub action, which would have to be approved).

leofang:
I start seeing lots of caches generated in the GitHub Cache space. I think we are on the right track. I can take over and finish the last mile if you like.

.github/workflows/build-wheel.yml
```diff
@@ -130,9 +145,18 @@ jobs:
...
+            SCCACHE_GHA_ENABLED=true
+            ACTIONS_RUNTIME_TOKEN=$ACTIONS_RUNTIME_TOKEN
+            ACTIONS_RESULTS_URL=$ACTIONS_RESULTS_URL
```
leofang:
```suggestion
            ACTIONS_RUNTIME_TOKEN=${{ env.ACTIONS_RUNTIME_TOKEN }}
            ACTIONS_RESULTS_URL=${{ env.ACTIONS_RESULTS_URL }}
```

leofang:
/ok to test [b5b0af4](https://github.com/NVIDIA/cuda-python/pull/1156/commits/b5b0af48267e9d5a89be0d31407c5cfdaba462a5)

leofang:
/ok to test [b608357](https://github.com/NVIDIA/cuda-python/pull/1156/commits/b608357bbca8f2dfc89b59b69db0ae22d0d81487)

leofang:
```
  [2025-10-20T03:53:18Z DEBUG sccache::server] Error executing cache write: Cannot write to read-only storage
```

leofang:
/ok to test [ee75abe](https://github.com/NVIDIA/cuda-python/pull/1156/commits/ee75abe31e90a304b1b46f1df543fe0503352a6b)

leofang:
/ok to test [f7a986f](https://github.com/NVIDIA/cuda-python/pull/1156/commits/f7a986f5688cb3a9ea88247ebec3ff939a444d6f)

leofang:
What I am trying to do:
- Linux: Cache files were successfully written to GitHub Cache space around 1:45-1:55pm ET last Friday. I am trying to reproduce what was done back then. 
   - Currently the stats confirm my observation that we keep re-compiling instead of fetching from cache
- Fix Windows later because `setuptools` does not honor `CC`/`CXX` on Windows for MSVC

leofang:
/ok to test [dc19abe](https://github.com/NVIDIA/cuda-python/pull/1156/commits/dc19abe9b03464b09a50f364f40e232f1b425b85)

.github/workflows/build-wheel.yml
```diff
@@ -130,9 +145,34 @@ jobs:
...
+            ACTIONS_RESULTS_URL=${{ env.ACTIONS_RESULTS_URL }}
+            SCCACHE_DIR=/host/${{ env.SCCACHE_DIR }}
+            SCCACHE_CACHE_SIZE=${{ env.SCCACHE_CACHE_SIZE }}
+            SCCACHE_LOG=debug
+#            SCCACHE_NO_DAEMON=1
           CIBW_ENVIRONMENT_WINDOWS: >
             CUDA_PATH="$(cygpath -w ${{ env.CUDA_PATH }})"
             CUDA_PYTHON_PARALLEL_LEVEL=${{ env.CUDA_PYTHON_PARALLEL_LEVEL }}
+            CC="${{ env.SCCACHE_PATH }} cl"
+            CXX="${{ env.SCCACHE_PATH }} cl"
```
mdboom: Yesterday I confirmed that setuptools doesn't use these on Windows.  I'm not sure how.  The `ccache` docs (which I know is not the same as `sccache`) say to just rename `ccache.exe` to `cl.exe`.  Maybe that works with `sccache`?

leofang:
> Linux: Cache files were successfully written to GitHub Cache space around 1:45-1:55pm ET last Friday. I am trying to reproduce what was done back then.
> * Currently the stats confirm my observation that we keep re-compiling instead of fetching from cache

Still scratching my head on how the caches were generated/uploaded. My changes don't seem to fix "Cache write errors" from the stats. Maybe they came from building Cython tests (in the VM) instead of from building bindings (in the container)...?
```
  Compile requests                      41
  Compile requests executed             21
  Cache hits                             0
  Cache misses                          21
  Cache misses (C/C++)                  21
  Cache hits rate                     0.00 %
  Cache hits rate (C/C++)             0.00 %
  Cache timeouts                         0
  Cache read errors                      0
  Forced recaches                        0
  Cache write errors                    21
  Cache errors                           0
  Compilations                          21
  Compilation failures                   0
  Non-cacheable compilations             0
  Non-cacheable calls                    0
  Non-compilation calls                 20
  Unsupported compiler calls             0
  Average cache write                0.000 s
  Average compiler                  26.911 s
  Average cache read hit             0.000 s
  Failed distributed compilations        0
  Cache location                  ghac, name: sccache-v0.11.0, prefix: /sccache/
  Version (client)                0.11.0
```

.github/workflows/build-wheel.yml
leofang: `setuptools` runs `vcvarsall` in a subprocess to get all the MSVC env vars, and then inject them to the build environment. Then, it just uses whatever `cl.exe` found in `$PATH`.

I also saw a thread proposing the renaming approach. Let's worry about it later. Also it's not too bad if we can't fix Windows, because Windows builds are already faster than Linux ones for some reason...

leofang:
/ok to test [4d31536](https://github.com/NVIDIA/cuda-python/pull/1156/commits/4d315365d3255d1e9e389f4b4743812e29c6a222)

leofang:
/ok to test [93b92b7](https://github.com/NVIDIA/cuda-python/pull/1156/commits/93b92b7ea5d9629f052c7fb1710b3009da3de5a7)

leofang:
We have cache hits now! 🎉  So passing all env vars is the key for local builds (without launching a server on host beforehand).
https://github.com/NVIDIA/cuda-python/actions/runs/18666438691/job/53218694396?pr=1156#step:17:1491
```
      + sh -c '"/host/home/runner/_work/_tool/sccache/0.11.0/x64/sccache" --show-stats
  '
  [2025-10-20T22:20:36Z DEBUG sccache::config] Attempting to read config file at "/root/.config/sccache/config"
  [2025-10-20T22:20:36Z DEBUG sccache::config] Couldn't open config file: failed to open file `/root/.config/sccache/config`
  [2025-10-20T22:20:36Z DEBUG sccache::commands] request_stats
  Compile requests                      41
  Compile requests executed             21
  Cache hits                            21
  Cache hits (C/C++)                    21
  Cache misses                           0
  Cache hits rate                   100.00 %
  Cache hits rate (C/C++)           100.00 %
  Cache timeouts                         0
  Cache read errors                      0
  Forced recaches                        0
  Cache write errors                     0
  Cache errors                           0
  Compilations                           0
  Compilation failures                   0
  Non-cacheable compilations             0
  Non-cacheable calls                    0
  Non-compilation calls                 20
  Unsupported compiler calls             0
  Average cache write                0.000 s
  Average compiler                   0.000 s
  Average cache read hit             0.210 s
  Failed distributed compilations        0
  Cache location                  ghac, name: 59b10024028a23cc391dd32dc0d0ec15ef7b6fd6ef6eb72c395d01609c0ca755, prefix: /sccache/
  Version (client)                0.11.0
```

.github/workflows/build-wheel.yml
leofang: I looked into Windows but it is really tricky. There is no way to customize the `MSVCCompiler` class from `setuptools` without vendoring a large patch, which I'd like to avoid unless there is no other way. 

When [`.compile()` is invoked](https://github.com/pypa/setuptools/blob/a49f6b927d83b97630b4fb030de8035ed32436fd/setuptools/_distutils/compilers/C/msvc.py#L453) it is done like this
```python
args = [self.cc] + compile_opts + pp_opts
```
so I was hoping that I can inject (monkey-patch) `sscache` as `self.cc` and `compile_opts[0]` as `cl.exe` (the original `self.cc`). This is not doable because [`compile_opts ` must start with "/c"](https://github.com/pypa/setuptools/blob/a49f6b927d83b97630b4fb030de8035ed32436fd/setuptools/_distutils/compilers/C/msvc.py#L391), so we need to monkey-patch the entire `.compile()` method, not just `self.cc` and `self.compile_options`...

We may try either this or the renaming approach...

> because Windows builds are already faster than Linux ones for some reason...

Discussed with Mike offline. This is no longer true in the recent main branch.

leofang:
Discussed with Mike offline. I'll apply sccache to other components' Linux builds, and Mike will experiment with a disguise approach on Windows (create a .bat file that is named as `cl.exe`, and under the hood calls `sccache C:\real\path\to\cl.exe`). 

I'll also send a PR to cibuildwheels to teach about what we found here.

leofang:
/ok to test [c10dc60](https://github.com/NVIDIA/cuda-python/pull/1156/commits/c10dc6054bda434313104252c191e4c0c9e9fa07)

leofang:
/ok to test [c10dc60](https://github.com/NVIDIA/cuda-python/pull/1156/commits/c10dc6054bda434313104252c191e4c0c9e9fa07)

leofang:
> 2 files reviewed, no comments

Sorry, Greptile, I don't think you'd understand what novel exploration we are doing here 😅

cpcloud:
... what is greptile and when/where/why did we add it?

rparolin:
> ... what is greptile and when/where/why did we add it?

AI code review that I added to the repo to kick the tires of.  I was going to mention it in the team meeting.

.github/workflows/build-wheel.yml
```diff
@@ -130,9 +143,35 @@ jobs:
...
+            ACTIONS_RESULTS_URL=${{ env.ACTIONS_RESULTS_URL }}
+            ACTIONS_CACHE_URL=${{ env.ACTIONS_CACHE_URL }}
+            ACTIONS_CACHE_SERVICE_V2=${{ env.ACTIONS_CACHE_SERVICE_V2 }}
+            SCCACHE_DIR=/host/${{ env.SCCACHE_DIR }}
+            SCCACHE_CACHE_SIZE=${{ env.SCCACHE_CACHE_SIZE }}
           CIBW_ENVIRONMENT_WINDOWS: >
             CUDA_PATH="$(cygpath -w ${{ env.CUDA_PATH }})"
             CUDA_PYTHON_PARALLEL_LEVEL=${{ env.CUDA_PYTHON_PARALLEL_LEVEL }}
+            CC="${{ env.SCCACHE_PATH }} cl"
+            CXX="${{ env.SCCACHE_PATH }} cl"
```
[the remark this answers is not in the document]
leofang: Sorry for being rude, Greptile, but please kindly read all conversations throughout this PR. The real issue was already pointed out earlier, that on Windows `CC` and `CXX` are not honored by `setuptools`, so we need a different way to inject a compiler wrapper.

leofang:
/ok to test [7fd19cd](https://github.com/NVIDIA/cuda-python/pull/1156/commits/7fd19cd9b35d055d102ee7949cc4fe2ece96d537)

leofang:
/ok to test [20e0ce3](https://github.com/NVIDIA/cuda-python/pull/1156/commits/20e0ce3c606001ab71ec56f540f373d49320d4dd)

leofang:
/ok to test [ce6c03d](https://github.com/NVIDIA/cuda-python/pull/1156/commits/ce6c03d90054ceff71b8d6ddc2fe74dc434ae146)

leofang:
/ok to test [814fb82](https://github.com/NVIDIA/cuda-python/pull/1156/commits/814fb8232bf697e744d5c7d65df163500c6ab352)

leofang:
/ok to test [b9cc2b3](https://github.com/NVIDIA/cuda-python/pull/1156/commits/b9cc2b3a6a39411f49b73fc36e86cf5e5810707f)

leofang:
/ok to test [d1d804f](https://github.com/NVIDIA/cuda-python/pull/1156/commits/d1d804f0935afce70fcd87b963a79b9233f2b6d8)

leofang:
/ok to test [7de169b](https://github.com/NVIDIA/cuda-python/pull/1156/commits/7de169be0ad095c5432d78f587b472fac98d9927)

leofang:
/ok to test [7de169b](https://github.com/NVIDIA/cuda-python/pull/1156/commits/7de169be0ad095c5432d78f587b472fac98d9927)

leofang:
/ok to test [7de169b](https://github.com/NVIDIA/cuda-python/pull/1156/commits/7de169be0ad095c5432d78f587b472fac98d9927)

leofang:
/ok to test [541a9b2](https://github.com/NVIDIA/cuda-python/pull/1156/commits/541a9b24712730f0e3a494443c56ef8321fd6a5c)

leofang:
/ok to test [098feeb](https://github.com/NVIDIA/cuda-python/pull/1156/commits/098feebffb52dab4b8b52c83492d101049fb437f)

leofang:
/ok to test [8f6f008](https://github.com/NVIDIA/cuda-python/pull/1156/commits/8f6f008f5365ca2db2e28b4b31df313eaf61dd65)

leofang:
/ok to test [65beac0](https://github.com/NVIDIA/cuda-python/pull/1156/commits/65beac0413297c5532aea652472a105e83e20cee)

leofang:
/ok to test [7cac30d](https://github.com/NVIDIA/cuda-python/pull/1156/commits/7cac30df6d991ed122bcd390848e363a35426e90)

leofang:
/ok to test [6be8df7](https://github.com/NVIDIA/cuda-python/pull/1156/commits/6be8df7e6272d7817bc888c5964bd842f866c131)

leofang:
It seems neither renaming (6be8df7) nor wrapping (8f6f008) works on Windows...

leofang:
/ok to test [1eb6a1b](https://github.com/NVIDIA/cuda-python/pull/1156/commits/1eb6a1b03c78c51f50dc95b7ba900bfeed26d5cc)

leofang:
I am thinking we should purge all unsuccessful Windows attempts and get this merged.

mdboom:
> I am thinking we should purge all unsuccessful Windows attempts and get this merged.

I agree.

mdboom:
This is working with the Windows stuff removed.

However, I'm not sure how to confirm it's actually working.  The stats at the end of the log (I guess) are from the host so are basically irrelevant.

But I'm also not seeing any speed improvement.  For example:

- [This run](https://github.com/NVIDIA/cuda-python/actions/runs/18753688569/job/53499960885?pr=1156) from this PR, presumably with a hot cache, took 227s to build cuda_bindings
- [This run](https://github.com/NVIDIA/cuda-python/actions/runs/18754633555/job/53503294436?pr=1155) from another PR, took 211s to build cuda_bindings

I think the difference is within the noise, but we aren't seeing a notable speedup?  If it is really caching, maybe the remote caching overhead of large files is erasing any benefits?

leofang (approved):

leofang:
I saw Linux builds decreased from ~9 mins to 5 mins as of commit c10dc60 ([link](https://github.com/NVIDIA/cuda-python/actions/runs/18691510619)). Let me check what happened since then. 

The cache stats need to be accessed in every cibuildwheel's test step ([example](https://github.com/NVIDIA/cuda-python/actions/runs/18691510619/job/53298181250#step:17:1340)), not at the end of the build workflows. As already noted, the stats on host is irrelevant.

leofang:
@mdboom Something caused cache miss in the most recent CI run:
```
  Compile requests                      41
  Compile requests executed             21
  Cache hits                             2
  Cache hits (C/C++)                     2
  Cache misses                          19
  Cache misses (C/C++)                  19
  Cache hits rate                     9.52 %
  Cache hits rate (C/C++)             9.52 %
  Cache timeouts                         0
  Cache read errors                      0
  Forced recaches                        0
  Cache write errors                     0
  Cache errors                           0
  Compilations                          19
  Compilation failures                   0
  Non-cacheable compilations             0
  Non-cacheable calls                    0
  Non-compilation calls                 20
  Unsupported compiler calls             0
  Average cache write                0.360 s
  Average compiler                  20.770 s
  Average cache read hit             0.173 s
  Failed distributed compilations        0
  Cache location                  ghac, name: c53e4962b46713a17b429097c76c3a9dfe6c94ba4556b3b544ce89beb6872fc9, prefix: /sccache/
  Version (client)                0.12.0
```
https://github.com/NVIDIA/cuda-python/actions/runs/18753688569/job/53499960916?pr=1156#step:17:1336

I dunno what happened but I see many new caches generated in the last hour or so:
https://github.com/NVIDIA/cuda-python/actions/caches?query=sort%3Acreated-desc

Let me kick off another run to ensure we get cache hits.

mdboom:
> Let me kick off another run to ensure we get cache hits.

Cool.  Maybe the cache is somehow keyed off the contents of the `build-wheel.yml` file.  It wouldn't be a crazy thing to do.

leofang:
> It wouldn't be a crazy thing to do.

Sounds crazy to me lol... I thought it's only relevant to compiler versions, flags, args, and .cpp file contents, if the cache can be invalidated easily due to workflow changes, it is a bit annoying. GitHub Cache space is only 10 GB and we've been managing it carefully so far. It's a scarce resource.

Anyway, cache hits are back: https://github.com/NVIDIA/cuda-python/actions/runs/18755592039/job/53506633614?pr=1156#step:17:1334

mdboom:
> Anyway, cache hits are back: https://github.com/NVIDIA/cuda-python/actions/runs/18755592039/job/53506633614?pr=1156#step:17:1334

Great.  And it looks like a cuda_bindings build is now 55s vs. 220s (approx).  So, around 3 minutes saved (+ the time saved on the 2 cuda_core builds).  Not bad.

cpcloud:
I tinkered with sccache locally, and the remaining 55s is mostly spent in the linker, which is a whole other can of worms :)

Issue #1165, opened 2025-10-21
[BUG(test only)]: `test_graphics_api_smoketest` failing on headless nodes


rwgk:
I keep running into the error below on colossus machines.

I have a fix that I need to extract from my CTK-next work, but logging the error here first for visibility.

```
=================================== FAILURES ===================================
_________________________ test_graphics_api_smoketest __________________________

self = <pyglet._ModuleProxy object at 0xef3118b30b30>, name = 'Texture'

    def __getattr__(self, name: str): # noqa: ANN204
        try:
>           return getattr(self._module, name)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
E           AttributeError: 'NoneType' object has no attribute 'Texture'

../TestVenv/lib/python3.12/site-packages/pyglet/__init__.py:439: AttributeError

During handling of the above exception, another exception occurred:

    def test_graphics_api_smoketest():
        # Due to lazy importing in pyglet, pytest.importorskip doesn't work
        try:
            import pyglet

>           tex = pyglet.image.Texture.create(512, 512)
                  ^^^^^^^^^^^^^^^^^^^^

tests/test_graphics_apis.py:13:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../TestVenv/lib/python3.12/site-packages/pyglet/__init__.py:445: in __getattr__
    __import__(import_name)
../TestVenv/lib/python3.12/site-packages/pyglet/image/__init__.py:106: in <module>
    from pyglet.gl import (
../TestVenv/lib/python3.12/site-packages/pyglet/gl/__init__.py:205: in <module>
    import pyglet.window  # noqa: F401
    ^^^^^^^^^^^^^^^^^^^^
../TestVenv/lib/python3.12/site-packages/pyglet/window/__init__.py:1904: in <module>
    gl._create_shadow_window()  # noqa: SLF001
    ^^^^^^^^^^^^^^^^^^^^^^^^^^
../TestVenv/lib/python3.12/site-packages/pyglet/gl/__init__.py:179: in _create_shadow_window
    _shadow_window = ShadowWindow()
                     ^^^^^^^^^^^^^^
../TestVenv/lib/python3.12/site-packages/pyglet/gl/__init__.py:168: in __init__
    super().__init__(width=1, height=1, visible=False)
[...]

Issue #248, opened 2024-11-18
cudaError_t.cudaErrorOperatingSystem


leviome:
```
from cuda import cudart as cu

err, _, _ = cu.cudaGLGetDevices(1, cu.cudaGLDeviceList.cudaGLDeviceListAll)
print(err)

# cudaError_t.cudaErrorOperatingSystem
```                

Ubuntu 20.04, RTX 4090

leofang:
We are investigating internally.

vzhurba01:
Hi @leviome, I'm assuming that the code shared is just a snippet of the API call after you've already created a GL context. When I run the snippet as is, I get `cudaError_t.cudaErrorInvalidGraphicsContext` which seems expected since a GL context needs to be created before this call IIUC.

However, here's what I ran to attempt a repro:
```
# need to install pygame, pyopengl, cuda-python (best through conda-forge)
import pygame as pg
from OpenGL import EGL
from cuda import cudart

if __name__ == "__main__":
    pg.init()
    display_size = (640, 480)
    pg.display.set_mode(display_size, pg.OPENGL | pg.DOUBLEBUF | pg.RESIZABLE)
    # check opengl context is created
    assert EGL.eglGetCurrentContext() is not None
    print(cudart.cudaGLGetDevices(1, cudart.cudaGLDeviceList.cudaGLDeviceListAll))
```
but this returns `cudaError_t.cudaSuccess`, implying that the bindings are working as expected. Therefore it's likely there's something happening in one of your prior calls or how your system is configured.

If you think there're prior calls that complicate the binding of cudaGLGetDevices API, a full snippet would be helpful. Otherwise the bindings look ok, so this issue can be closed.

PR #1166, opened 2025-10-22
Skipping `test_graphics_apis.py` when running on WSL

rparolin:
This change is addressing a test failure when running the graphics api tests on WSL. We simply skip the tests when running on WSL due to the driver lacking pass through capabilities on WSL.

image

### DIFF
--- a/cuda_core/cuda/core/experimental/_memory.pyx
+++ b/cuda_core/cuda/core/experimental/_memory.pyx
@@ -1089,10 +1089,10 @@ class _SynchronousMemoryResource(MemoryResource):
         return self._dev_id
 
 
-VirtualMemoryHandleTypeT = Literal["posix_fd", "generic", "none", "win32", "win32_kmt", "fabric"]
+VirtualMemoryHandleTypeT = Union[Literal["posix_fd", "generic", "win32", "win32_kmt", "fabric"], None]
 VirtualMemoryLocationTypeT = Literal["device", "host", "host_numa", "host_numa_current"]
 VirtualMemoryGranularityT = Literal["minimum", "recommended"]
-VirtualMemoryAccessTypeT = Literal["rw", "r", "none"]
+VirtualMemoryAccessTypeT = Union[Literal["rw", "r"], None]
 VirtualMemoryAllocationTypeT = Literal["pinned", "managed"]
 
 
@@ -1101,18 +1101,31 @@ class VirtualMemoryResourceOptions:
     """A configuration object for the VirtualMemoryResource
        Stores configuration information which tells the resource how to use the CUDA VMM APIs
 
-    Args:
-        handle_type: Export handle type for the physical allocation. Use
-            CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR on Linux if you plan to
-            import/export the allocation (required for cuMemRetainAllocationHandle).
-            Use CU_MEM_HANDLE_TYPE_NONE if you don't need an exportable handle.
-        gpu_direct_rdma: Hint that the allocation should be GDR-capable (if supported).
-        granularity: 'recommended' or 'minimum'. Controls granularity query and size rounding.
-        addr_hint: A (optional) virtual address hint to try to reserve at. 0 -> let CUDA choose.
-        addr_align: Alignment for the VA reservation. If None, use the queried granularity.
-        peers: Extra device IDs that should be granted access in addition to `device`.
-        self_access: Access flags for the owning device ('rw', 'r', or 'none').
-        peer_access: Access flags for peers ('rw' or 'r').
+    Attributes
+    ----------
+    allocation_type: :obj:`~_memory.VirtualMemoryAllocationTypeT`
+        Controls the type of allocation.
+    location_type: :obj:`~_memory.VirtualMemoryLocationTypeT`
+        Controls the location of the allocation.
+    handle_type: :obj:`~_memory.VirtualMemoryHandleTypeT`
+        Export handle type for the physical allocation. Use
+        ``"posix_fd"`` on Linux if you plan to
+        import/export the allocation (required for cuMemRetainAllocationHandle).
+        Use `None` if you don't need an exportable handle.
+    gpu_direct_rdma: bool
+        Hint that the allocation should be GDR-capable (if supported).
+    granularity: :obj:`~_memory.VirtualMemoryGranularityT`
+        Controls granularity query and size rounding.
+    addr_hint: int
+        A (optional) virtual address hint to try to reserve at. Setting it to 0 lets the CUDA driver decide.
+    addr_align: int
+        Alignment for the VA reservation. If `None`, use the queried granularity.
+    peers: Iterable[int]
+        Extra device IDs that should be granted access in addition to ``device``.
+    self_access: :obj:`~_memory.VirtualMemoryAccessTypeT`
+        Access flags for the owning device.
+    peer_access: :obj:`~_memory.VirtualMemoryAccessTypeT`
+        Access flags for peers.
     """
     # Human-friendly strings; normalized in __post_init__
     allocation_type: VirtualMemoryAllocationTypeT = "pinned"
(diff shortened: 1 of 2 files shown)


rparolin:
@greptileai

rparolin:
@greptileai

rparolin:
@greptileai

rparolin:
@greptileai

mdboom (changes requested):
I don't think this test /always/ fails on WSL.  WSL is my daily driver, I wrote this test, and it "works for me" :tm:.  So I think there is probably something more specific/different about the WSL setup that is failing there.  Maybe there's no `$DISPLAY`, or maybe the drivers aren't installed?  Are there any clues in the way it's failing?  I think we should update this to something more narrow than just "always skip on WSL".

I also have some feedback on the `cuda_python_test_helpers` stuff, but I think that's out-of-scope for here so it can be dealt with separately.

leofang:
I agree with Mike that we shouldn't just skip. Lots of things, if not all, are supposed to just work on WSL, and I raised this offline to Rob and all in the team channel that incomplete knowledge of WSL failures are now biting us.

rparolin:
I can dig deeper into why this happening... I figured since it was throwing a clear operating system error code that WSL wasn't support.  But I'll dig deeper.

cuda_core/tests/helpers.py
```diff
@@ -22,8 +22,14 @@
...
-    sys.path.insert(0, str(pathlib.Path(__file__).resolve().parents[2] / "cuda_python_test_helpers"))
-    import cuda_python_test_helpers
+    test_helpers_path = str(pathlib.Path(__file__).resolve().parents[2] / "cuda_python_test_helpers")
+    try:
+        sys.path.insert(0, test_helpers_path)
+        import cuda_python_test_helpers
+    finally:
+        # Clean up sys.path modification
+        if test_helpers_path in sys.path:
+            sys.path.remove(test_helpers_path)
```
cpcloud, line 32: What are you trying to achieve here by removing this from `sys.path` after adding it?

The import system will keep `cuda_python_test_helpers` in `sys.modules`, so any subsequent imports of it will just pull from that.

cuda_bindings/tests/test_graphics_apis.py
cpcloud, line 13: This test should be restructured to avoid mixing up the reason for unexpected failures, expected failures, or missing dependencies.

Right now, it's just `pytest.skip`-ing and then `return`ing (I realize this is pre-existing) which is making the whole thing unnecessarily complicated.

```python
# skips if `pyglet.image` isn't importable for whatever reason, you can address
# the laziness by importing something deeper like `pyglet.image`
pyglet_image = pytest.importorskip("pyglet.image")
# if the next line raises an `AttributeError`, that should be specified in a
# `pytest.mark.xfail(raises=AttributeError)` decorator using whatever condition
# to indicate why the failure happened
tex = pyglet_image.Texture.create(512, 512)
```

Then you can probably remove the code that's mixing these reasons together.

cpcloud: You can also chain `xfail` decorators so that you can specify different conditions and types. For example, let's say `pyglet` is installed but we're running without a display. That will raises a different error than `AttributeError`. So you can do the following:

```python
@pytest.mark.xfail(condition=..., raises=AttributeError)
@pytest.mark.xfail(condition=os.environ.get("DISPLAY") is None, raises=pyglet.TheRightException)
def test_graphics_api_smoketest():
    ...
```

You can also combine these if they're not mutually exclusive (i.e., an `AttributeError` is raised when `DISPLAY` is `None`.

```python
@pytest.mark.xfail(condition=... or os.environ.get("DISPLAY") is None, raises=(AttributeError, pyglet.TheRightException))
def test_graphics_api_smoketest():
    ...
```

mdboom:
Looking at the screenshot again, I think maybe just adding `cudaErrorOperatingSystem` to the list of allowed results is probably enough.  It's letting the graphics API tell us the OS isn't sufficient rather than deciding for ourselves.  That should allow it continue to test /something/ where the WSL setup is working, and also test that it doesn't work where it isn't expected to.

leofang:
But why didn't you get the `cudaErrorOperatingSystem` error when you developed the test? Something is clearly missing.

leofang:
I don't know if it is obvious to see relations between issues/PRs, but I think this might be relevant (https://github.com/NVIDIA/cuda-python/issues/1165).

mdboom:
> But why didn't you get the `cudaErrorOperatingSystem` error when you developed the test? Something is clearly missing.

Yeah, I think @rparolin can just compare our WSL installs until we understand what might be the difference, then we can gate on that.  But generally speaking, if CUDA tells us the OS isn't suitable, that's a "reasonable" result in terms of a test that runs everywhere.

leofang:
> if CUDA tells us the OS isn't suitable, that's a "reasonable" result in terms of a test that runs everywhere.

OK. Here's the CUDA's dark corner (which we learned through debugging https://github.com/NVIDIA/cuda-python/issues/248#issuecomment-2688935641). In the graphics interop context `cudaErrorOperatingSystem` usually just means a graphics context is not correctly set up, and there could be many reasons for why it's the case. It does not mean that the OS is not supported. In the case of issue 248 we had to activate the EGL context first (`export PYOPENGL_PLATFORM="EGL"` and then `from OpenGL import EGL`), and it could not be on a headless node. I dunno what pyglet is doing under the hood, so I suggest we look at this together with https://github.com/NVIDIA/cuda-python/issues/1165 holistically and understand the real issue. The CUDA Python team needs to have a deep understanding in CUDA too.

leofang:
(I canceled all CI runs because they are hanging and blocking all other teams from getting their tests run, see the nv-gha-runners thread.)

rwgk:
I'm seeing the error in the PR description also here (ctk-next testing).

Note the `LOOOK` below: `os.environ.get("DISPLAY")=None`

This is with Windows 11 24H2 headless.

See also #1165 that @leofang pointed out already.

```
(TestVenv) C:\Users\rgrossekunst\forked\cuda-python-private\ctk-next\cuda_bindings>python -m pytest -ra -s -vv tests\test_graphics_apis.py
====================================================================== test session starts ======================================================================
platform win32 -- Python 3.13.9, pytest-8.4.2, pluggy-1.6.0 -- C:\Users\rgrossekunst\forked\cuda-python-private\ctk-next\TestVenv\Scripts\python.exe
cachedir: .pytest_cache
benchmark: 5.1.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: C:\Users\rgrossekunst\forked\cuda-python-private\ctk-next\cuda_bindings\tests
configfile: pytest.ini
plugins: benchmark-5.1.0
collected 2 items

tests\test_graphics_apis.py::test_graphics_api_smoketest
LOOOK os.environ.get("DISPLAY")=None
FAILED
tests\test_graphics_apis.py::test_cuda_register_image_invalid PASSED

=========================================================================== FAILURES ============================================================================
__________________________________________________________________ test_graphics_api_smoketest __________________________________________________________________

    def test_graphics_api_smoketest():
        pyglet = pytest.importorskip("pyglet")
        try:
            tex = pyglet.image.Texture.create(512, 512)
        except ImportError:
            pytest.skip("pyglet lazy ImportError")
        except AttributeError:
            pytest.skip("pyglet lazy import AttributeError")
        except pyglet.display.xlib.NoSuchDisplayException:
            pytest.skip("pyglet.display.xlib.NoSuchDisplayException")

        err, gfx_resource = cudart.cudaGraphicsGLRegisterImage(
            tex.id, tex.target, cudart.cudaGraphicsRegisterFlags.cudaGraphicsRegisterFlagsWriteDiscard
        )
        import os
        print(f"\nLOOOK {os.environ.get("DISPLAY")=}", flush=True)
        error_name = cudart.cudaGetErrorName(err)[1].decode()
        if error_name == "cudaSuccess":
            assert int(gfx_resource) != 0
        else:
>           assert error_name in ("cudaErrorInvalidValue", "cudaErrorUnknown")
E           AssertionError: assert 'cudaErrorOperatingSystem' in ('cudaErrorInvalidValue', 'cudaErrorUnknown')

tests\test_graphics_apis.py:28: AssertionError
==================================================================== short test summary info ====================================================================
FAILED tests\test_graphics_apis.py::test_graphics_api_smoketest - AssertionError: assert 'cudaErrorOperatingSystem' in ('cudaErrorInvalidValue', 'cudaErrorUnknown')
================================================================== 1 failed, 1 passed in 0.42s ==================================================================
```

rwgk:
This seems to work for me (same as in https://github.com/NVIDIA/cuda-python/issues/1165#issuecomment-3434240213):

```diff
diff --git a/cuda_bindings/tests/test_graphics_apis.py b/cuda_bindings/tests/test_graphics_apis.py
old mode 100644
new mode 100755
index 67f674b..d8bb2ee
--- a/cuda_bindings/tests/test_graphics_apis.py
+++ b/cuda_bindings/tests/test_graphics_apis.py
@@ -1,6 +1,8 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE

+import os
+
 import pytest
 from cuda.bindings import runtime as cudart

@@ -22,6 +24,8 @@ def test_graphics_api_smoketest():
     error_name = cudart.cudaGetErrorName(err)[1].decode()
     if error_name == "cudaSuccess":
         assert int(gfx_resource) != 0
+    elif os.environ.get("DISPLAY") is None:
+        assert error_name == "cudaErrorOperatingSystem"
     else:
         assert error_name in ("cudaErrorInvalidValue", "cudaErrorUnknown")
```

rparolin:
> Looking at the screenshot again, I think maybe just adding `cudaErrorOperatingSystem` to the list of allowed results is probably enough. It's letting the graphics API tell us the OS isn't sufficient rather than deciding for ourselves. That should allow it continue to test /something/ where the WSL setup is working, and also test that it doesn't work where it isn't expected to.

That was actually my first commit but I allowed myself to be convinced by Greptile that doing a platform skip was more clearly showing intent.  🤦

cuda_core/tests/helpers.py
rparolin: My understanding was its preventing polluting global state because the import path changes is alive for the duration of the lifetime of the interpreter process.

rparolin:
I circled back and root caused this error.  

The `test_graphics_api_smoketest` test failed because the `cudaGraphicsGLRegisterImage` function returned `cudaErrorOperatingSystem` (304) instead of expected errors. The reason it was throwing an "operating system erro" was OpenGL was using software rendering (`llvmpipe`) instead of using the available NVIDIA GPU. WSL was missing the required drivers `libGLX_nvidia.so.0` which caused it to fallback to Mesa CPU rendering, making CUDA-GL interop impossible (can't map CPU memory as GPU memory).

Once I installed the `libnvidia-gl-580` drivers the `cudaErrorOperatingSystem` changed into `cudaErrorUnknown` which the test suite was expecting.

Looking at `cudaGraphicsGLRegisterImage` function (line 37218 in runtime.pyx.in) it officially documents these possible return values: cudaSuccess, cudaErrorInvalidDevice, cudaErrorInvalidValue, cudaErrorInvalidResourceHandle, cudaErrorOperatingSystem, cudaErrorUnknown.  Which from my perspective is a strong case for adding `cudaErrorOperatingsystem` to the error tuple that swallows the OpenGL errors.

I'll close this PR as this change is not required as written.

cc: @leofang @mdboom

Issue #1170, opened 2025-10-22
Add a NumPy version check to `examples/thread_block_cluster.py`


leofang:
As noted in #1168, this test needs NumPy v2.2.5+.

leofang:
This one seems to be my oversight back in #868.

leofang:
We just need to add something like this to the example:
https://github.com/NVIDIA/cuda-python/blob/20f29e97d9e8d156ceeeb7c5021123f48208d637/cuda_core/examples/memory_ops.py#L28-L30

PR #1172, opened 2025-10-22
Add numpy version check in examples/thread_block_cluster.py

rluo8:
## Description

This is to fix issue in https://github.com/NVIDIA/cuda-python/issues/1170 .
The fix is in cuda_core/examples/thread_block_cluster.py to add a numpy version check. If numpy version is less than 2.2.5, the test will skip.
After the fix, the test could be skipped when numpy version is less than 2.2.5.

example_tests/test_basic_examples.py::TestExamples::test_example[C:\\tmp\\cuda-python\\cuda_core\\tests\\example_tests\\..\\..\\examples\\thread_block_cluster.py] SKIPPED [100%]


rluo8:
@leofang @rparolin ,
Could you please help review it?
Thanks!

cuda_core/examples/thread_block_cluster.py
```diff
@@ -21,6 +21,11 @@
     ProgramOptions,
     launch,
 )
+from packaging import version
+
+if version.parse(np.__version__) < version.parse("2.2.5"):
```
leofang: Not all users will have the `packaging` package installed. Let's use NumPy's built-in API:
```suggestion
if np.lib.NumpyVersion(np.__version__) < "2.2.5":
```

rparolin (changes requested):
lgtm, once Leo's suggested changes are merged.

cuda_core/examples/thread_block_cluster.py
cpcloud: Even better is importorskip, with minversion: https://docs.pytest.org/en/stable/how-to/skipping.html#skipping-on-a-missing-import-dependency

leofang: No, because all examples could be run outside of pytest.

rluo8: Updated the codes. Thanks for your suggestion.

leofang (approved):

cuda_core/examples/thread_block_cluster.py
cpcloud: Ah, right. I missed that it was an example!

PR #1174, opened 2025-10-22
ci: run status check always but consider cancellations failures

cpcloud:
This PR should address the issue of cancelled-due-to-failure-but-still-times-out shenanigans present in GitHub Actions. Details and links are in the comment.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -202,4 +202,28 @@ jobs:
       - doc
     steps:
       - name: Exit
-        run: exit 0
+        run: |
+          # if any dependencies were cancelled, that's a failure
+          #
+          # see https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#always
+          # and https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -202,4 +202,28 @@ jobs:
...
+        run: |
+          # if any dependencies were cancelled, that's a failure
+          #
+          # see https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#always
+          # and https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks
+          # for why this cannot be encoded in the job-level `if:` field
+          #
+          # TL; DR: `$REASONS`
+          #
+          # The intersection of skipped-as-success and required status checks
```
jrhemstad, line 214: Ah yes, I feel every proper GHA-based status check workflow needs a comment expressing the dismay and astonishment that came from this discovery. See also: https://github.com/NVIDIA/cccl/blob/main/.github/workflows/ci-workflow-pull-request.yml#L259-L263

🙂

cpcloud:
Seems like cancellation is working: https://github.com/NVIDIA/cuda-python/actions/runs/18721400957/job/53395760809?pr=1174

I manually cancelled and that caused the status check to fail.

PR #1176, opened 2025-10-22
Restore "Implement IPC-enabled events. (#1145)"

Andy-Jost:
This reverts commit bcd40ff208f750f7766ce8d1183c87424c474e56.

### DIFF
--- a/.github/workflows/build-wheel.yml
+++ b/.github/workflows/build-wheel.yml
@@ -45,6 +45,19 @@ jobs:
         with:
           fetch-depth: 0
 
+      # The env vars ACTIONS_CACHE_SERVICE_V2, ACTIONS_RESULTS_URL, and ACTIONS_RUNTIME_TOKEN
+      # are exposed by this action.
+      - name: Enable sccache
+        uses: mozilla-actions/sccache-action@7d986dd989559c6ecdb630a3fd2557667be217ad  # 0.0.9
+
+      # xref: https://github.com/orgs/community/discussions/42856#discussioncomment-7678867
+      - name: Adding addtional GHA cache-related env vars
+        uses: actions/github-script@v7
+        with:
+          script: |
+            core.exportVariable('ACTIONS_CACHE_URL', process.env['ACTIONS_CACHE_URL'])
+            core.exportVariable('ACTIONS_RUNTIME_URL', process.env['ACTIONS_RUNTIME_URL'])
+
       - name: Setup proxy cache
         uses: nv-gha-runners/setup-proxy-cache@main
         continue-on-error: true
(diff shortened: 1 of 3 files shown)


cuda_core/tests/test_helpers.py
```diff
@@ -31,8 +32,11 @@ def test_latchkernel():
     target.copy_from(ones, stream=stream)
     log("going to sleep")
     time.sleep(1)
-    log("checking target == 0")
-    assert compare_equal_buffers(target, zeros)
+    if not IS_WINDOWS and not IS_WSL:
+        # On any sort of Windows system, checking the memory before stream
+        # sync results in a page error.
+        log("checking target == 0")
+        assert compare_equal_buffers(target, zeros)
```
Andy-Jost, line 39: This was the cause of the CI crash. A surprise to me, accessing unified memory that is currently queued in a CUDA stream operation results in a page error on Windows. On Linux it works as expected (to me, anyway).

cuda_core/tests/helpers/buffers.py
```diff
@@ -0,0 +1,122 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import ctypes
+import sys
+
+from cuda.core.experimental import Buffer, MemoryResource
+from cuda.core.experimental._utils.cuda_utils import driver, handle_return
+
+if sys.platform.startswith("win"):
```
rparolin, line 10: We should migrate this to a common place so that this pattern doesn't get replicated in multiple files.  It would be a maintenance headache if we had to update it in multiple places.

cuda_core/tests/helpers/latch.py
```diff
@@ -0,0 +1,78 @@
...
+                           break;
+                       }
+
+                       // Check for timeout
+                       if (clock64() - start >= timeout_cycles) {
+                           break;  // Timeout reached
+                       }
+
+                       // Avoid 100% spin.
+                       __nanosleep(10000000); // 10 ms
```
rparolin, line 48: 10ms is a lot for a polling loop like this. You could probably get away with a 1ms and as a nit if you don't need the granularity of nanosleep you could use the `std::thread::sleep` as an alternative.

cuda_core/tests/memory_ipc/test_event_ipc.py
```diff
@@ -0,0 +1,172 @@
...
+        device.set_current()
+        stream2 = device.create_stream()
+        twos = make_scratch_buffer(device, 2, NBYTES)
+        buffer = q_in.get(timeout=CHILD_TIMEOUT_SEC)
+        log("got buffer")
+        e = q_in.get(timeout=CHILD_TIMEOUT_SEC)
+        log(f"got event ({hex(e.handle)})")
+        stream2.wait(e)
+        log("enqueuing copy on stream2")
+        buffer.copy_from(twos, stream=stream2)
```
rparolin, line 83: Shouldn't we verify that the buffer we got in the child process contains "1" before we copy "2" on top of it?

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -151,6 +165,40 @@ cdef class Event:
...
+            raise RuntimeError("Event is not IPC-enabled")
+        cdef cydriver.CUipcEventHandle data
+        with nogil:
+            HANDLE_RETURN(cydriver.cuIpcGetEventHandle(&data, <cydriver.CUevent>(self._handle)))
+        cdef bytes data_b = cpython.PyBytes_FromStringAndSize(<char*>(data.reserved), sizeof(data.reserved))
+        self._ipc_descriptor = IPCEventDescriptor._init(data_b, self._busy_waited)
+        return self._ipc_descriptor
+
+    @classmethod
+    def from_ipc_descriptor(cls, ipc_descriptor: IPCEventDescriptor) -> Event:
```
leofang, line 182: Q: it doesn't seem that this public API is tested?

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -151,6 +165,40 @@ cdef class Event:
...
+        cdef cydriver.CUipcEventHandle data
+        memcpy(data.reserved, <const void*><const char*>(ipc_descriptor._reserved), sizeof(data.reserved))
+        cdef Event self = Event.__new__(cls)
+        with nogil:
+            HANDLE_RETURN(cydriver.cuIpcOpenEventHandle(&self._handle, data))
+        self._timing_disabled = True
+        self._busy_waited = ipc_descriptor._busy_waited
+        self._ipc_enabled = True
+        self._ipc_descriptor = ipc_descriptor
+        self._device_id = -1  # ??
```
leofang, line 193: I think when we send over the handle we should also send the device id

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -151,6 +165,40 @@ cdef class Event:
...
+        self._ctx_handle = None  # ??
```
leofang, line 194: This should be (lazily? not sure if it's possible) initialized in the child process (to the specified device's current context).

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -212,12 +255,43 @@ cdef class Event:
         context is set current after a event is created.
 
         """
-
-        from cuda.core.experimental._device import Device  # avoid circular import
-
-        return Device(self._device_id)
+        if self._device_id >= 0:
```
leofang, line 258: ditto, then this awkward check can be removed.

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -212,12 +255,43 @@ cdef class Event:
...
+            from ._device import Device  # avoid circular import
+            return Device(self._device_id)
 
     @property
     def context(self) -> Context:
         """Return the :obj:`~_context.Context` associated with this event."""
-        return Context._from_ctx(self._ctx_handle, self._device_id)
+        if self._ctx_handle is not None and self._device_id >= 0:
```
leofang, line 265: ditto

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -260,7 +260,13 @@ cdef class Stream:
         # and CU_EVENT_RECORD_EXTERNAL, can be set in EventOptions.
         if event is None:
             self._get_device_and_context()
-            event = Event._init(<int>(self._device_id), <uintptr_t>(self._ctx_handle), options)
+            event = Event._init(<int>(self._device_id), <uintptr_t>(self._ctx_handle), options, False)
+        elif event.is_ipc_enabled:
+            raise TypeError(
+                "IPC-enabled events should not be re-recorded, instead create a "
+                "new event by supplying options."
+            )
```
leofang, line 268: Imagine we are doing ping pong between two processes. We should be able to reuse the same event in parent and child. Does the driver actually disallow this?

PR #1178, opened 2025-10-22
ci: remove python3.13t builds

### LINKED ISSUE #1177: ci: remove `3.13t` builds
Python 3.13t free-threading support is experimental (3.14t is no longer experimental) and a number of issues were found between 3.13t and 3.14t, the least of which was that the `ctypes` module, which is heavily used in `cuda_pathfinder` (for example) was not thread-safe.

We should drop support for 3.13t to avoid having to deal with our own experimental support of an experimental Python interpreter.

cpcloud:
Closes #1177.

### DIFF
--- a/.github/workflows/build-wheel.yml
+++ b/.github/workflows/build-wheel.yml
@@ -33,7 +33,6 @@ jobs:
           - "3.11"
           - "3.12"
           - "3.13"
-          - "3.13t"
           - "3.14"
           - "3.14t"
     name: py${{ matrix.python-version }}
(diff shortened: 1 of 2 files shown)


cpcloud:
> Since Python 3.14t provides stable free-threading support

This feels ... overly optimistic 😆

mdboom (approved):
LGTM!

.github/workflows/build-wheel.yml
```diff
@@ -287,9 +286,7 @@ jobs:
           cuda-path: "./cuda_toolkit_prev"
 
       - name: Download cuda.bindings build artifacts from the prior branch
-        if: ${{ matrix.python-version == '3.13t'
-                || matrix.python-version == '3.14'
-                || matrix.python-version == '3.14t' }}
```
leofang, line 292: Q: We don't need 3.14t here? (TBH I don't remember why this if was needed.)

PR #1179, opened 2025-10-22
Checking for RDMA support before allocating via VMM in test suite

rparolin:
This PR addresses issues with GPU Direct RDMA support validation in the Virtual Memory Resource (VMM) allocator and improves test coverage for memory management functionality. 

- Removed hardcoded platform checks: Eliminated Windows and WSL-specific skips in favor of device capability checks.
-  New test case: Added test_vmm_allocator_rdma_unsupported_exception() to verify proper error handling when RDMA is requested on unsupported devices.

image

image

image

### DIFF
--- a/cuda_core/cuda/core/experimental/_memory.pyx
+++ b/cuda_core/cuda/core/experimental/_memory.pyx
@@ -1119,7 +1119,7 @@ class VirtualMemoryResourceOptions:
     location_type: VirtualMemoryLocationTypeT = "device"
     handle_type: VirtualMemoryHandleTypeT = "posix_fd"
     granularity: VirtualMemoryGranularityT = "recommended"
-    gpu_direct_rdma: bool = True
+    gpu_direct_rdma: bool = False
     addr_hint: Optional[int] = 0
     addr_align: Optional[int] = None
     peers: Iterable[int] = field(default_factory=tuple)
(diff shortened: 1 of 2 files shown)


rparolin:
@greptileai

rparolin:
> It occurs to me that none of us (Ben, Keith, myself) read the docs when getting the VMM PR merged. The docs made it clear that there is one device attribute that we should check (which is typical to all major CUDA features, as we did in the IPC mempool test helper). https://docs.nvidia.com/cuda/cuda-c-programming-guide/#query-for-support
> 
> Translate this to cuda.core, we need to check
> 
> ```python
> dev = Device()
> if not dev.properties.virtual_memory_management_supported:
>     pytest.skip(...)
> ```

As discussed in person, migrated the majority of the test suites skip checks to use `dev.properties.virtual_memory_management_supported` but also check for device support for RDMA when the user is explicitly requesting it via our API.

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -1119,7 +1119,7 @@ class VirtualMemoryResourceOptions:
     location_type: VirtualMemoryLocationTypeT = "device"
     handle_type: VirtualMemoryHandleTypeT = "posix_fd"
     granularity: VirtualMemoryGranularityT = "recommended"
-    gpu_direct_rdma: bool = True
+    gpu_direct_rdma: bool = False
```
leofang, line 1122: cc @benhg for vis

PR #1181, opened 2025-10-23
Adding  `greptile.json` configuration file to repo

rparolin:
This enables the team to provide great customization of greptile behaviour than can be achieved through the web interface.

Behaviour changes:

- greptile reviews are no longer automatically review every PR. Users must tag review with `ai-review` label to trigger a greptile review.
- summary section will still be included when a review is requested but it will be collapsed by default to reduce noise.

### DIFF
--- a/cuda_bindings/docs/source/conf.py
+++ b/cuda_bindings/docs/source/conf.py
@@ -9,13 +9,11 @@
 
 # -- Path setup --------------------------------------------------------------
 
-# If extensions (or modules to document with autodoc) are in another directory,
-# add these directories to sys.path here. If the directory is relative to the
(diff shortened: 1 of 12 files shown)


greptile.json
```diff
@@ -0,0 +1,63 @@
...
+  ],
+  "ignorePatterns": "greptile.json\n",
+  "summarySection": {
+    "included": true,
+    "collapsible": true,
+    "defaultOopen": false
+  },
+  "includeBranches": [
+    "main",
+    "vNext"
```
cpcloud, line 22: Do we actually have a branch called `vNext` anywhere? I don't see it.

rparolin: No, but it was in the current config via the website.  We can remove it after I submit the config file if we don't need it.

PR #1182, opened 2025-10-23
Add Handwritten Bindings for CuFileRead and CuFileWrite 

chloechia4:
## Description

This PR allows us to specify that the generated Python bindings for CuFileRead and CuFileWrite return bytes read or written, respectively. This is to ensure parity with the CuFile APIs and is helpful for verifying the contents read from a buffer or written from a buffer.


cuda_bindings/cuda/bindings/cufile.pyx
```diff
@@ -2689,3 +2655,29 @@ cpdef driver_close():
     with nogil:
         status = cuFileDriverClose_v2()
     check_status(status)
+
+cpdef read(intptr_t fh, intptr_t buf_ptr_base, size_t size, off_t file_offset, off_t buf_ptr_offset):
```
sourabgupta3, line 2659: Read should also have similar documentation as Write.
Can you also updates the tests to check for bytes returned?

mdboom: Sorry, I missed that this docstring was different when I approved the related cybind MR.  We will also need to file another MR to update this over there.

chloechia4: Oh yes whoops added to this repo and corresponding cybind MR :)

rparolin (dismissed):

cuda_bindings/cuda/bindings/cufile.pxd
```diff
@@ -47,8 +47,6 @@ cpdef intptr_t handle_register(intptr_t descr) except? 0
 cpdef void handle_deregister(intptr_t fh) except*
 cpdef buf_register(intptr_t buf_ptr_base, size_t length, int flags)
 cpdef buf_deregister(intptr_t buf_ptr_base)
-cpdef read(intptr_t fh, intptr_t buf_ptr_base, size_t size, off_t file_offset, off_t buf_ptr_offset)
```
rparolin, line 50: Question: Why are we removing these function definitions?

chloechia4: They are handwritten bindings so they won't be explicitly declared in this file. This is triggered by the SKIP_LOWPP tag for read and write in cybind's config_cufile.py

mdboom: @chloechia4 is totally correct, it's triggered by that.  But thinking about this some more, I wonder if this means we /also/ need to add these declarations in the template for `cufile.pxd` in cybind so that Cython users of our library can continue to access these.

Am I right, @leofang?

leofang: We do not guarantee that the Python layer can be cimport'd in Cython.  Removing code from `cufile.pxd` in a minor release is OK. It is the Cython layer (`cycufile.pxd`) that needs a strong guarantee to not break across minor releases.

leofang:
> * Add tests that confirm we are getting the expected integer result when calling read and write.

Yes let's do this. (Sourab requested this too.)

PR #1184, opened 2025-10-24
Minor testing rework.

Andy-Jost:
Follow-on changes to #1176 based on review feedback.

### DIFF
--- a/cuda_core/tests/helpers/__init__.py
+++ b/cuda_core/tests/helpers/__init__.py
@@ -3,7 +3,6 @@
 
 import os
 import pathlib
-import platform
 import sys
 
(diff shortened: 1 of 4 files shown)


cuda_core/tests/helpers/__init__.py
```diff
@@ -20,16 +19,8 @@
 
 
 try:
-    import cuda_python_test_helpers
+    from cuda_python_test_helpers import *  # noqa: F403
```
Andy-Jost, line 22: This seems like a good place to use `import *`, since we are trying to merge modules defined in different places, we control both, and the source has an `__all__` list.

mdboom (approved):

PR #1190, opened 2025-10-24
Fix #1186: Fix segmentation fault when accessing StridedMemoryView

### LINKED ISSUE #1186: [BUG]: Segmentation fault with StridedMemoryView.shape
### Is this a duplicate?

### Type of Bug

Runtime Error

### Component

cuda.core

### Describe the bug

Hello!

There seems to be a bug in explicit shape/stride tuple construction. Program accessing the shape or stride repeatedly in a loop ends up with memory corruption. I think I tracked the problem to `carray_int64_t_to_tuple` utility, as replacing it with `tuple(<list comprehension>)` call, makes the problem go away.

Digging deeper, in the cpp code cython generates, it looks like cython decreases refcount of the newly created python int, even though the tuple is supposed to steal it (last line below).

```
    /* "cuda/core/experimental/_utils/cuda_utils.pxd":39
 *     result = cpython.PyTuple_New(length)
 *     for i in range(length):
 *         cpython.PyTuple_SET_ITEM(result, i, cpython.PyLong_FromLongLong(ptr[i]))             # <<<<<<<<<<<<<<
 *     return result
*/
    __pyx_t_1 = PyLong_FromLongLong((__pyx_v_ptr[__pyx_v_i])); if (unlikely(!__pyx_t_1)) __PYX_ERR(1, 39, __pyx_L1_error)
    __Pyx_GOTREF(__pyx_t_1);
    PyTuple_SET_ITEM(__pyx_v_result, __pyx_v_i, __pyx_t_1);
```

If I add explicit incref, the problem seems to go away

```
cdef inline tuple carray_int64_t_to_tuple(int64_t *ptr, int length):
    # Construct shape and strides tuples using the Python/C API for speed
    result = cpython.PyTuple_New(length)
    cdef object item
    for i in range(length):
        item = cpython.PyLong_FromLongLong(ptr[i])
        cpython.Py_INCREF(item)
        cpython.PyTuple_SET_ITEM(result, i, item)
    return result
```

As the low-level Python references are tricky, I'd like someone to take a look and confirm the root-cause and proposed fix. Thank you!

### How to Reproduce

Repro: the following code randomly segfaults

```
import cupy as cp

import cuda.core.experimental as ccx
d = ccx.Device()
d.set_current()
s = d.default_stream
a = cp.arange(11171, dtype=cp.uint8)
av = ccx.utils.StridedMemoryView(a, stream_ptr=s.handle)
for i in range(1000*1000):
    av.shape
```

### Expected behavior

No memory corruption:)

Possibly, by explicit incref when creation the tuple.

### Operating System

_No response_

### nvidia-smi output

_No response_

mdboom:
To avoid adding unnecessary reference counts to counteract Cython's unnecessary reference counts, this fix adds "non-magical" externs to the CPython API to make the reference counting correct.

In the process of writing a test investigating that bug, I also discovered a bug in the caching logic for `shape`.

Lastly, for the case where strides comes from `CAI` rather than `DLPack`, rather than porting this fix there, since that's already a "slow" Python API path, I just reverted that to a "tuple" comprehension.

### DIFF
--- a/cuda_core/tests/test_memory.py
+++ b/cuda_core/tests/test_memory.py
@@ -616,12 +616,12 @@ def test_strided_memory_view_leak():
 
 def test_strided_memory_view_refcnt():
     # Use Fortran ordering so strides is used
-    a = np.zeros((11171, 4), dtype=np.uint8, order="F")
+    a = np.zeros((64, 4), dtype=np.uint8, order="F")
     av = StridedMemoryView(a, stream_ptr=-1)
     # segfaults if refcnt is wrong
-    assert av.shape[0] == 11171
+    assert av.shape[0] == 64
     assert sys.getrefcount(av.shape) >= 2
 
     assert av.strides[0] == 1
-    assert av.strides[1] == 11171
+    assert av.strides[1] == 64
     assert sys.getrefcount(av.strides) >= 2


cuda_core/cuda/core/experimental/_utils/cuda_utils.pxd
```diff
@@ -32,9 +33,17 @@ cpdef int _check_nvrtc_error(error) except?-1
 cpdef check_or_create_options(type cls, options, str options_description=*, bint keep_none=*)
 
 
+# Create low-level externs so Cython won't "helpfully" handle reference counting
+# for us.  Prefixing with an underscore to distinguish it from the definition in
+# cpython.long.
+cdef extern from "Python.h":
+    PyObject *_PyLong_FromLongLong "PyLong_FromLongLong" (long long val) except NULL
+    void _PyTuple_SET_ITEM "PyTuple_SET_ITEM" (object p, Py_ssize_t pos, PyObject *o)
```
cpcloud, line 41: It is wild that *this* is the way to handle the problem.

cuda_core/tests/test_memory.py
```diff
@@ -612,3 +612,16 @@ def test_strided_memory_view_leak():
         StridedMemoryView(arr, stream_ptr=-1)
     after = sys.getrefcount(arr)
     assert before == after
+
+
+def test_strided_memory_view_refcnt():
+    # Use Fortran ordering so strides is used
+    a = np.zeros((11171, 4), dtype=np.uint8, order="F")
```
cpcloud: Any reason a smaller array won't do? I'm not that concerned about size, but this seems unnecessary :)

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -146,14 +147,11 @@ cdef class StridedMemoryView:
...
+                    # to using the optimization in cuda_utils.carray_int64_t_to_tuple
                     strides = self.metadata.get("strides")
                     if strides is not None:
-                        itemsize = self.dtype.itemsize
-                        self._strides = cpython.PyTuple_New(len(strides))
-                        for i in range(len(strides)):
-                            cpython.PyTuple_SET_ITEM(
-                                self._strides, i, strides[i] // itemsize
-                            )
+                        self._strides = tuple(x // itemsize for x in strides)
```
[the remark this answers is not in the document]
cpcloud: The bot appears to be correct. `itemsize` is declared, but never assigned anymore.

cuda_core/cuda/core/experimental/_utils/cuda_utils.pxd
mdboom: It could have been handled by adding an explicit incref to counteract the implicit decref that Cython generates, but this is the only way I could find to avoid the unnecessary nonsense.

Cython and performance optimization are at odds.

This fixes the problem at hand, but it caused me to look more closely at the reference counting the Cython generates, and there is a /lot/ of unnecessary work it does.  I don't know how much of that the C compiler can see through and optimize away, but I doubt it's everything.  Cython really needs its own reduction pass, probably.  This will become more of a problem for free-threaded builds where reference counting is more expensive.

cuda_core/tests/test_memory.py
mdboom: Nah, I just brought this over from the bug report.  Making it smaller.

cuda_core/cuda/core/experimental/_utils/cuda_utils.pxd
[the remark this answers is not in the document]
mdboom: This is a reasonable question @greptile-apps, but I confirmed it's not the case:

```
  __pyx_t_2 = __pyx_v_length;
  __pyx_t_3 = __pyx_t_2;
  for (__pyx_t_4 = 0; __pyx_t_4 < __pyx_t_3; __pyx_t_4+=1) {
    __pyx_v_i = __pyx_t_4;
    __pyx_t_5 = PyLong_FromLongLong((__pyx_v_ptr[__pyx_v_i])); if (unlikely(__pyx_t_5 == ((PyObject *)0))) __PYX_ERR(3, 48, __pyx_L1_error)
    PyTuple_SET_ITEM(__pyx_v_result, __pyx_v_i, __pyx_t_5); // <-- no refcounting here
  }
```

leofang: So what was the code that Cython generated without this `cdef extern` hack?

rparolin (approved):

cuda_core/cuda/core/experimental/_utils/cuda_utils.pxd
mdboom: This is the code it generated before:

```
    __pyx_t_1 = PyLong_FromLongLong((__pyx_v_ptr[__pyx_v_i])); if (unlikely(!__pyx_t_1)) __PYX_ERR(1, 39, __pyx_L1_error)
    __Pyx_GOTREF(__pyx_t_1);
    PyTuple_SET_ITEM(__pyx_v_result, __pyx_v_i, __pyx_t_1);
    __Pyx_DECREF(__pyx_t_1); __pyx_t_1 = 0;  // <--- wrong!
```

That DECREF is incorrect because `PyTuple_SET_ITEM` steals a reference.  @stiepan's suggestion in #1188 was to explicitly `INCREF` that value, which is totally correct, but would churn the refcount unnecessarily.  My fix here basically uses raw `PyObject *` instead to avoid Cython being "helpful".

leofang: @mdboom I see. IIUC this will generate the equivalent code with less boilerplate and without unnecessary churns (untested):
```cython
item = cpython.PyLong_FromLongLong(ptr[i])
cpython.Py_INCREF(item)
cpython.PyTuple_SET_ITEM(result, i, <cpython.PyObject*>(item))
```
This is because in Cython [the 3rd arg is typed as `object`](https://github.com/cython/cython/blob/25dab649e80db9d9bb50e582e146bbf9b4f5b2b0/Cython/Includes/cpython/tuple.pxd#L54), for which Cython handles refcount before/after the function call. Casting it to `PyObject*` bypasses Cython's mechanism and gives us full control over refcount.

mdboom: Yeah, I tried this.  The cast doesn't work because Cython still wants you to cast back into an `object` to pass to its wrapper of `PyTuple_SET_ITEM`:

```
          cpython.PyTuple_SET_ITEM(result, i, <cpython.PyObject *>cpython.PyLong_FromLongLong(ptr[i]))
                                                  ^

      ./cuda/core/experimental/_utils/cuda_utils.pxd:39:44: Casting temporary Python object to non-numeric non-Python type
```

It also causes unnecessary reference count churn.  (We incref the item only because Cython is implicitly and unnecessarily decref'ing it later).  The C compiler generally can't see through and optimize away that kind of thing because it can't reason that the reference count won't ever go to zero and that freeing the object is always a no-op.

As I look at this stuff more, I think a better approach is probably to drop all the way down to C for these performance-critical sections.  The whole `cimport cpython` tree of things is full of rough edges because it doesn't model the behavior of the Python/C API completely (it doesn't understand steals, for example).  To the point where maybe it's a good guideline to "avoid `cimport cpython`".  For example, maybe we should do this (only for these sort of localized hot path optimizations of course).  It's a lot more verbose due to the error checking, but it's a lot easier to reason about what it's doing and where the performance optimizations are:

```
cdef extern from *:
    """
    static inline PyObject* carray_int64_to_tuple(uint64_t *var, Py_ssize_t size) {
        PyObject* result = PyTuple_New(size);
        if (result == NULL) {
            return NULL;
        }
        PyObject* item;
        for (Py_ssize_t i = 0; i < size; i++) {
            item = PyLong_FromLongLong(var[i]);
            if (item == NULL) {
                Py_DECREF(result);
                return NULL;
            }
            PyTuple_SET_ITEM(result, i, item);
        }
        return result;
    }
    """
    PyObject *carray_int64_to_tuple(uintptr_t var, Py_ssize_t size)
```

leofang: > The cast doesn't work because Cython still wants you to cast back into an `object` to pass to its wrapper of `PyTuple_SET_ITEM`

Ah right. This makes me sad...

PR #1192, opened 2025-10-27
Adds use_pool option to DeviceMemoryResource.

### LINKED ISSUE #963: CUDA graph phase 2 - memory nodes
memcpy & mem ops

Andy-Jost:
## Description

closes #963

Adds `use_pool` option to `DeviceMemoryResource` to allow disabling memory pool use. This allows for graph capture of stream-ordered allocations.

### DIFF
--- a/cuda_core/tests/test_graph2.py
+++ b/cuda_core/tests/test_graph2.py
@@ -1,0 +1,119 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
+
+import numpy as np
+import pytest
+
+try:
+    from cuda.bindings import nvrtc
+except ImportError:
+    from cuda import nvrtc
+from cuda.core.experimental import (
+    Device,
+    DeviceMemoryResource,
+    DeviceMemoryResourceOptions,
+    GraphBuilder,
+    GraphCompleteOptions,
+    GraphDebugPrintOptions,
+    LaunchConfig,
+    LegacyPinnedMemoryResource,
+    Program,
+    ProgramOptions,
+    launch,
+)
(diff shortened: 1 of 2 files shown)


Andy-Jost:
This is by no means ready for merge. I'm interested in getting feedback on whether this is the right direction.

cuda_core/tests/test_graph2.py
```diff
@@ -0,0 +1,119 @@
...
+    # Begin graph capture.
+    gb = Device().create_graph_builder().begin_building(mode="thread_local")
+
+    # import code
+    # code.interact(local=dict(globals(), **locals()))
+    work_buffer = mr.allocate(NBYTES, stream=gb.stream)
+    launch(gb, LaunchConfig(grid=1, block=1), set_zero, int(work_buffer.handle), NBYTES)
+    launch(gb, LaunchConfig(grid=1, block=1), add_one, int(work_buffer.handle), NBYTES)
+    launch(gb, LaunchConfig(grid=1, block=1), add_one, int(work_buffer.handle), NBYTES)
+    target.copy_from(work_buffer, stream=gb.stream)
```
Andy-Jost, line 103: I believe this is the sort of use we are targeting.

pciolkosz:
Why not add a separate memory resource type for graphs? The behavior is different enough that a separate type might be needed

Andy-Jost:
> Why not add a separate memory resource type for graphs? The behavior is different enough that a separate type might be needed

I would much prefer to do that. When I added IPC support to `DeviceMemoryResource` I initially made `DeviceMemoryResource.__new__` a factory method that chose the correct subclass, but that idea was rejected. I certainly think the existing code is getting way too complicated but I don't have a good sense of what alternatives will be accepted.

kkraus14:
> Why not add a separate memory resource type for graphs? The behavior is different enough that a separate type might be needed

If a user tries to use a memory resource on a CUDA stream that is being captured what would be the behavior? We throw and instruct them to use a graph memory resource instead? What happens if a user is trying to use a graph memory resource on a non capturing stream?

Issue #664, opened 2025-05-30
[FEA]: cuda.core constructs which share underlying C structure should hash to same value


carterbox:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

cuda.core

### Is your feature request related to a problem? Please describe.

Both cupy and pytorch define `__hash__()` methods which use the pointer address (and sometimes device index) of the underlying CUDA API structure such two Python level objects which represent the same underlying CUDA API structure will hash the same.

https://github.com/pytorch/pytorch/blob/v2.7.0/torch/cuda/streams.py#L14
https://github.com/cupy/cupy/blob/v13.4.1/cupy/cuda/stream.pyx

### Describe the solution you'd like

cuda.core should also implement `__hash__()` such that caching and comparison are easier to implement.

leofang:
I guess we probably have to fix #465 first? (cc @rwgk for thoughts)

Issue #1075, opened 2025-10-03
Cythonize `_context.pyx` more


leofang:

Andy-Jost:
Context class is fully Cythonized. ContextOptions is a placeholder for future work.

Issue #465, opened 2025-02-24
`CUstream` does not compare equal to itself


leofang:
See CI failure: https://github.com/NVIDIA/cuda-python/actions/runs/13488648820/job/37683655417?pr=463#step:18:264

leofang:
@pciolkosz I left the meeting early today. What's the conclusion regarding stream's `operator==` behavior?

pciolkosz:
We will continue to compare the handles. We will document that using `stream_ref` member functions is UB past the lifetime of the stream with an exception for `.get()` which will always work. I will also add the stream ID getter to `stream_ref` and check with RMM if that is sufficient

leofang:
Thanks, Piotr! Just wanna make sure I understand it correctly, by handle we meant `CUstream` address instead of its unique ID, right?

pciolkosz:
Yes, that's correct.

PR #1198, opened 2025-10-29
Implement `__hash__` and `__eq__` for cuda.core classes

rparolin:
Implement `__hash__` and `__eq__` for cuda.core classes

This PR implements `__hash__()` and `__eq__()` methods for cuda.core classes (Stream, Event, Context, Device), enabling them to be used as dictionary keys and in sets.
                                                                                                                        
Fixes https://github.com/NVIDIA/cuda-python/issues/664

### DIFF
--- a/cuda_core/tests/test_comparable.py
+++ b/cuda_core/tests/test_comparable.py
@@ -65,7 +65,6 @@ def test_device_subclass_equality(init_cuda):
 
     # Since they're the same object, they're equal
     assert device == my_device
-    assert not (device != my_device)
 
 
 def test_stream_subclass_equality(init_cuda):
@@ -87,20 +86,17 @@ def test_stream_subclass_equality(init_cuda):
     # Create another Stream wrapping same handle
     stream2 = Stream.from_handle(int(stream.handle))
     assert stream == stream2, "Streams wrapping same handle are equal"
-    assert not (stream != stream2)
 
     # Create subclass instance with different handle
     my_stream = MyStream._init(options=StreamOptions(), device_id=device.device_id)
 
     # Different handles -> not equal
     assert stream != my_stream, "Streams with different handles are not equal"
-    assert not (stream == my_stream)
     assert stream.handle != my_stream.handle
 
     # sanity check: base and subclass compare equal (and hash equal)
     stream_from_handle = MyStream.from_handle(int(my_stream.handle))
     assert my_stream == stream_from_handle, "MyStream and Stream wrapping same handle compare equal"
-    assert not (my_stream != stream_from_handle)
     assert hash(my_stream) == hash(stream_from_handle)
 
 
@@ -122,7 +118,6 @@ def test_event_subclass_equality(init_cuda):
 
     # Different events should not be equal (different handles)
     assert event != my_event, "Different Event instances are not equal"
-    assert not (event == my_event)
 
     # Same subclass type with different handles
     my_event2 = MyEvent._init(device.device_id, device.context, options=EventOptions())
@@ -147,7 +142,6 @@ def test_context_subclass_equality(init_cuda):
 
     # Since both are Context instances with same handle, they're equal
     assert context == my_context, "Context instances with same handle are equal"
-    assert not (context != my_context)
 
     # Create another context from different stream
     stream2 = device.create_stream()


Andy-Jost:
+1 for this change. Longer term, it would be great to achieve this sort of thing with a mix-in class. I think what gets in the way is some internal inconsistency, e.g., using `_id` versus `_handle` for handle-like members. Perhaps something to consider if we refactor or reimplement anything.

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -165,6 +165,48 @@ cdef class Event:
...
+        int
+            Hash value based on the event handle address.
+
+        Warning
+        -------
+        Using a closed or destroyed event as a dictionary key or in a set
+        results in undefined behavior. The event handle may be reused by
+        the CUDA driver for new events.
+        """
+        return hash(<uintptr_t>(self._handle))
```
oleksandr-pavlyk:
```suggestion
        return hash(int(self._handle))
```

Since `hash` is a Python function, the code as was written would convert Python object `_handle` to native `uintptr_t` type which would then be converted to object of type `int`, and `hash` would be called on that `int` object.

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -165,6 +165,48 @@ cdef class Event:
...
+
+        Returns
+        -------
+        bool
+            True if other is an Event wrapping the same handle, False otherwise.
+            Returns NotImplemented if other is not an Event.
+        """
+        if not isinstance(other, Event):
+            return NotImplemented
+        return <uintptr_t>(self._handle) == <uintptr_t>((<Event>other)._handle)
```
oleksandr-pavlyk: Perhaps we can apply similar approach to line 49 in `_context.pyx`? I think this is faster

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -165,6 +165,48 @@ cdef class Event:
...
+            Another object to compare with.
+
+        Returns
+        -------
+        bool
+            True if other is an Event wrapping the same handle, False otherwise.
+            Returns NotImplemented if other is not an Event.
+        """
+        if not isinstance(other, Event):
+            return NotImplemented
```
oleksandr-pavlyk, line 174: Should we just return `False`?

rparolin:
> +1 for this change. Longer term, it would be great to achieve this sort of thing with a mix-in class. I think what gets in the way is some internal inconsistency, e.g., using `_id` versus `_handle` for handle-like members. Perhaps something to consider if we refactor or reimplement anything.

What would be the benefit of using a mixin? I feel like it would require a bunch of machinary of parameterize it to the point it solves the problems you pointed out (ie. being able to handle different handle member names)

cuda_core/cuda/core/experimental/_event.pyx
leofang: no because `Context` is not yet fully cythonized, and we could run into the problem described in #465 depending on the cuda-bindings version

rparolin: @leofang I'm not following your concern?   Its looks like #465 is asserting that a `Stream` object that holds the same handle value must be the same python object.

cuda_core/cuda/core/experimental/_event.pyx
rparolin: This actually causes a cython compilation error.

  Error compiling Cython file:
  ...
          -------
          Using a closed or destroyed event as a dictionary key or in a set
          results in undefined behavior. The event handle may be reused by
          the CUDA driver for new events.
          """
          return hash(int(self._handle))
                              ^
  
  cuda/core/experimental/_event.pyx:186:28: Cannot convert 'CUevent' to Python object

cuda_core/cuda/core/experimental/_event.pyx
rwgk: I wouldn't do that, unless you're sure you don't want Python to try the reflected operation:

* https://chatgpt.com/share/6902762f-bb60-8008-a945-5f806385daa0

Return value | Meaning | What Python does next

* `False` | “I know how to compare, and they’re not equal.” | Stops right there — comparison is considered handled.

* `NotImplemented` | “I don’t know how to compare this object to that one.” | Python will try the reflected operation (e.g. other.__eq__(self)), and only if that also returns NotImplemented, it will fall back to False.

cuda_core/cuda/core/experimental/_event.pyx
oleksandr-pavlyk: Then perhaps `hash(<object>self._handle)` is more appropriate.

cuda_core/cuda/core/experimental/_event.pyx
oleksandr-pavlyk: That concern would only be valid is `__eq__` was supported between different types, right? Or are we envisioning other types implementing meaningful comparison with `cuda.core.Event`?

cuda_core/cuda/core/experimental/_event.pyx
leofang: I think what Rob had was fine. In the case of `Event`, `self._handle` is a cdef attribute. Casting it via `<object>` would not work.

cuda_core/cuda/core/experimental/_event.pyx
rwgk: Using `NotImplemented` leaves the door open. No harm AFAIK. Seems like best practice to me.

cuda_core/cuda/core/experimental/_event.pyx
leofang: Sorry the conclusion is the same (no need to change) but the reason was wrong. In the case of `Context`, since we have not fully cythonized it (#1075) `self._handle` is a generic Python object. Casting it via `<uintptr_t>` would lead to a compilation error. Instead, we need to cast it to a Python it first, meaning `<uintptr_t>(int(self._handle))` would work.

cuda_core/cuda/core/experimental/_device.pyx
```diff
@@ -1145,6 +1145,48 @@ class Device:
...
+        int
+            Hash value based on the device ordinal (device_id).
+
+        Notes
+        -----
+        Device is a per-thread singleton, but equality is based on logical
+        device identity (device_id), not Python object identity. This means
+        Device(0) on thread A equals Device(0) on thread B.
+        """
+        return hash(self._id)
```
leofang: I am torn on this. The fact that we designed `Device` to be a per-thread singleton took into consideration that different threads could own different resources (e.g. `dev.memory_resource` can point to different MRs; same for `dev.context` despite we haven't really enabled it). Also, [it is a hard requirement (from CUDA)](https://developer.nvidia.com/blog/cuda-pro-tip-always-set-current-device-avoid-multithreading-bugs/) that `dev.set_current()` must be called when a new thread is spawned. However, I do think not comparing equal could leave us in a weird state. Seems like we should do this.

rparolin: Frankly, I was not sure what was the correct path here was so I made sure to document it to force the conversation.

I was leaning towards the side of the fence that included a scheme along the lines of `hash(thread_id + id)` to make them unique.

cuda_core/cuda/core/experimental/_event.pyx
mdboom: At some point in the distant future when we don't have to worry about Python < 3.13, we should use [Py_HashPointer](https://docs.python.org/3/c-api/hash.html#c.Py_HashPointer).  Creating Python `long` objects is really expensive (because they are an arbitrary precision integer), so there is a new API to avoid that when you already have a pointer at hand.

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -197,6 +197,47 @@ cdef class Stream:
...
+            Another object to compare with.
+
+        Returns
+        -------
+        bool
+            True if other is a Stream wrapping the same handle, False otherwise.
+        """
+        if not isinstance(other, Stream):
+            return False
+        return <uintptr_t>(self._handle) == <uintptr_t>((<Stream>other)._handle)
```
mdboom: Performance nit: Since `<Stream>other` is already doing an expensive `isinstance` check, we could avoid doing that twice with:

```
cdef Stream other_
try:
    other_ = <Stream>other
except TypeError:
    return False
return <uintptr_t>(self._handle) == <uintptr_t>(other_._handle)
```

cuda_core/cuda/core/experimental/_device.pyx
cpcloud: I think it's a bit more readable to package up this kind of thing with a tuple, like `hash((thread_id, id))` rather than with arithmetic, which feels much more arbitrary (e.g., why addition versus something else?).

Another thing that would probably help make the hash less likely to collide with other objects would be to include the type of the object into the hash:

```python
def __hash__(self) -> int:
    return hash((type(self), threading.current_thread().ident, self._id))
```

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -197,6 +197,47 @@ cdef class Stream:
...
+        -------
+        int
+            Hash value based on the stream handle address.
+
+        Warning
+        -------
+        Using a closed or destroyed stream as a dictionary key or in a set
+        results in undefined behavior. The stream handle may be reused by
+        the CUDA driver for new streams.
+        """
```
Andy-Jost: On this point, should we raise an exception if the handle is NULL?

cuda_core/cuda/core/experimental/_device.pyx
Andy-Jost: +1 for defining a tuple as key to serve as the basis for hash and equality.

Andy-Jost:
> > +1 for this change. Longer term, it would be great to achieve this sort of thing with a mix-in class. I think what gets in the way is some internal inconsistency, e.g., using `_id` versus `_handle` for handle-like members. Perhaps something to consider if we refactor or reimplement anything.
> 
> What would be the benefit of using a mixin? I feel like it would require a bunch of machinary of parameterize it to the point it solves the problems you pointed out (ie. being able to handle different handle member names)

The goal of a mixin would be to improve maintainability. The hash/eq code would appear in another place, shortening class definitions and ensuring consistency. The parameterization could come in the form of a property:

```
@property
def hashkey(self):
    return self.a, self.b, self.c  # whatever is needed here
```

cuda_core/cuda/core/experimental/_device.pyx
leofang: Rob pinged me offline so let me copy/paste here. Our `Device` is akin to `device_ref` in cccl-rt, which has `operator==` simply implemented as comparing the device IDs
https://github.com/NVIDIA/cccl/blob/88b95bb9a2cd29b1e249b1d5a7d4a7d57b704b18/libcudacxx/include/cuda/__device/device_ref.h#L63-L66
so we can just follow in implementing `__eq__` and `__hash__` (which is a unique thing in Python). No need to make it more complicated and include thread ID as part of hash key.

cpcloud: I would still strongly recommend hashing a tuple of `type(self)` and `self._id` for the above-stated reasons.

rparolin: Sorry, I meant the addition as pseudo code... I normally write that code in c++ and didn't know how to write in python. In my past life, I would have indivdually hashed them and use a bit-wise ops to merge them together... similar to what boost::hash_combine does.

cuda_core/cuda/core/experimental/_context.pyx
```diff
@@ -29,4 +29,35 @@ cdef class Context:
...
+        other : object
+            Another object to compare with.
+
+        Returns
+        -------
+        bool
+            True if other is a Context wrapping the same handle, False otherwise.
+        """
+        if not isinstance(other, Context):
+            return False
```
cpcloud: I'm not sure why `NotImplemented` was changed, but by removing it you're kneecapping other libraries and preventing them from implementing their own `__eq__` that could, for example, make the `__eq__` call succeed (e.g., comparing some member to the `self`).

If you want maximum interoperability here, you need to return `NotImplemented`.

IMO performance is secondary before we actually have anything to benchmark against.

cuda_core/cuda/core/experimental/_device.pyx
cpcloud: Something similar happens in [`tuple_hash`](https://github.com/python/cpython/blob/main/Objects/tupleobject.c?plain=1#L323).

Might as well just reuse that.

rparolin: yeah, its pretty much the standard mechanism for generating these aggregate style hashes.

rparolin: @leofang @cpcloud How should I move forward here? If I had to choose, I'd not include the thread_id but include the type_id which would avoid collision if different types landed in the same container.  At least their hashes would not be equivalent  if multiple types were sitting on monotonically increasing handle values starting at 0.

cuda_core/cuda/core/experimental/_stream.pyx
rparolin: Done.

cuda_core/cuda/core/experimental/_device.pyx
cpcloud: Including `type(self)` as part of a hashed `tuple` should hopefully be uncontroversial, since you're baking that in in the `isinstance` check in `__eq__`.

Including thread id in the hash seems like a bad idea if we don't also include it in `__eq__` which seems a little strange given that a device is conceptually independent of the thread it's used in for the purposes of equality, so +1 on leaving that out entirely.

cuda_core/cuda/core/experimental/_stream.pyx
rparolin: Looking at the existing semantics of the type, I'd argue against this because the handle shouldn't be null as a post-condition of the object ctor running successfully.

cuda_core/cuda/core/experimental/_context.pyx
rparolin: The feedback that caused this change wasn't from a perf perspective it was from a semantics perspective.  The argument was if the wrong type was passed to this function, it shouldn't throw, it should return `False`

cpcloud: Ah, I thought I saw something fly by about performance elsewhere.

According to https://docs.python.org/3/library/constants.html#NotImplemented, the semantics of raising are intended.

The purpose of returning `NotImplemented` is best summarized in that link:

> When a binary (or in-place) method returns NotImplemented the interpreter will try the reflected operation on the other type (or some other fallback, depending on the operator).

cpcloud: The point is that "wrong type" can't solely be decided by `cuda.core` alone, since another type, from another library, might come along and say "I know how to make this work".

Here's a (somewhat contrived) example:

```
In [14]: class MyInt:
    ...:     def __init__(self, value):
    ...:         self.value = value
    ...:     def __eq__(self, other):
    ...:         return self.value == getattr(other, "value", other)
    ...:

In [15]: 1 == MyInt(1)
Out[15]: True

In [16]: (1).__eq__(MyInt(1))
Out[16]: NotImplemented
```

This works because `int.__eq__` returns `NotImplemented`.

cpcloud: FWIW, this comment is **not** blocking. Someone else can tackle this in a follow-up.

rparolin: It definitely seems like this the idiomatic way to implement these operators.
https://docs.python.org/3/library/numbers.html#implementing-the-arithmetic-operations

I'll change it back. Thanks!

rparolin: Done.

cuda_core/cuda/core/experimental/_device.pyx
rparolin: How about I merge this PR as its ready to go... and I fast follow with a PR that introduces this change?  @cpcloud

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -197,6 +197,51 @@ cdef class Stream:
...
+            Another object to compare with.
+
+        Returns
+        -------
+        bool or NotImplemented
+            True if other is a Stream wrapping the same handle, False if not equal,
+            NotImplemented if other is not a Stream.
+        """
+        cdef Stream _other
+        try:
```
rwgk: A comment to explain why `isinstance(other, Stream)` does not work here would be useful.

leofang: Right, why do we have asymmetry here? Both `Event` and `Stream` are cdef class, but treated differently.

cuda_core/tests/test_comparable.py
leofang, line 1: Q: I don't find issues here, just a question of test organization. We have `test_XXXXX.py` file for each CUDA object that we expose. Would it make better sense here to break out `test_comparable.py` and `test_hashable.py` to these files, so that each all tests for object `XXXXX` are placed in the same file?

cuda_core/cuda/core/experimental/_device.pyx
leofang: You meant adding `type(self)`? I agree with Philip that we should do it for `Device`. Device IDs from 0-16 are very easy to collide. Let's just do it now.

cuda_core/cuda/core/experimental/_device.pyx
```diff
@@ -1145,6 +1145,49 @@ class Device:
...
```
kkraus14: Since this is the device ordinal, it can in theory change across processes if someone uses something like `CUDA_VISIBLE_DEVICES`. Should we consider hashing based on something like the UUID that remains consistent regardless of how the devices were ordered?

cuda_core/cuda/core/experimental/_event.pyx
kkraus14: Is there ever a situation where there's multiple CUDA contexts and the driver somehow reuses a handle address for a previously destroyed event on a different context? Or even reuses a handle for a previously destroyed event in general on the same context?

Not sure we can do anything about the 2nd situation, but we could include the CUDA context into the hash calculation.

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -197,6 +197,51 @@ cdef class Stream:
...
+        int
+            Hash value based on the stream handle address.
+
+        Warning
+        -------
+        Using a closed or destroyed stream as a dictionary key or in a set
+        results in undefined behavior. The stream handle may be reused by
+        the CUDA driver for new streams.
+        """
+        return hash(<uintptr_t>(self._handle))
```
kkraus14: +1 to above question with regards to contexts

cuda_core/cuda/core/experimental/_device.pyx
rparolin: Done.

cuda_core/tests/test_comparable.py
rparolin: I can move the basic functionality tests to type specific test files.  I have a test that uses multiple types... I'll leave those in the functionality specific test file.  Thanks for highlighting.

cuda_core/cuda/core/experimental/_stream.pyx
rparolin: This was due to a performance nit that @mdboom pointed out.
https://github.com/NVIDIA/cuda-python/pull/1198#discussion_r2477981885

I just missed changing this function.  I'll audit the other functions and make sure everything is consistent.  Thanks.

rparolin: Done.

cuda_core/tests/test_comparable.py
rparolin: Done.

cuda_core/cuda/core/experimental/_device.pyx
rparolin: Yikes. I didn't know about `CUDA_VISIBLE_DEVICES` -- seems like a dangerous feature.  I definitely think there is a value in utilizing a UUID that is guaranteed to uniquely identify a device across processes.  It would be great if we had a more practical use case b/c the `CUDA_VISIBLE_DEVICES` example comes across as a user intentionally trying to break the system.  That being said I'll move forward with making the change.

I ran some benchmark experiments and using an UUID in the hash has minimal impact on performance provided we cache the UUID in the device object. That enables us to avoid calling a CUDA API per hash request which is 10x slower than the hash calculation.

  • UUID fetch (CUDA API call): 0.35–0.37 microseconds
  • Hash computation (tuple hash): 0.03–0.04 microseconds

rparolin: Done.

cuda_core/cuda/core/experimental/_event.pyx
rparolin: Short answer, no handles are not reused across different contexts.

 Test results:
  • Created 2 different contexts on the same device
  • Context 1: 0x56e963d8ea30
  • Context 2: 0x56e964964ec0
  • Context 1 handles: 8 unique
  • Context 2 handles: 7 unique
  • Handle overlap: 0 handles
  
Handles are context-scoped — reused only within the same context, not across different contexts.

That being said, as a defensive strategy, we should include the `context` as part of the `event` from a semantic perspective. It bakes into the event object identity that it comes from a specific context object.

I'll make those changes.

rparolin: Done.

cuda_core/cuda/core/experimental/_stream.pyx
rparolin: Done.

cuda_core/cuda/core/experimental/_device.pyx
```diff
@@ -1145,6 +1146,60 @@ class Device:
...
+        -----
+        Uses UUID to ensure consistency across processes where
+        CUDA_VISIBLE_DEVICES may change the device ordinal mapping. The UUID
+        uniquely identifies the physical device and remains stable regardless of
+        how devices are ordered or filtered.
+
+        The UUID is cached after first access to avoid repeated CUDA API calls.
+        """
+        if self._uuid is None:
+            self._uuid = self.uuid
```
kkraus14: This should be getting memoized in the `uuid` property instead of in the `__hash__` function here and then here can just use the `self.uuid` property?

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -165,6 +165,65 @@ cdef class Event:
...
+        the context provides defense-in-depth against hash collisions.
+
+        Warning
+        -------
+        Using a closed or destroyed event as a dictionary key or in a set
+        results in undefined behavior. The event handle may be reused by
+        the CUDA driver for new events.
+        """
+        # Context should always be set as a post-condition of Event construction
+        return hash((type(self), <uintptr_t>(int(self._ctx_handle)), <uintptr_t>(self._handle)))
```
kkraus14: Since `self._ctx_handle` is a `Context` object which is now hashable, I think we can just pass it directly instead of needing to convert it to `int` and cast to `<uintptr_t>`?

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -197,6 +224,66 @@ cdef class Stream:
...
+        The context is fetched and cached during Stream construction.
+
+        Warning
+        -------
+        Using a closed or destroyed stream as a dictionary key or in a set
+        results in undefined behavior. The stream handle may be reused by
+        the CUDA driver for new streams.
+        """
+        # Context should always be set as a post-condition of Stream construction
+        return hash((type(self), <uintptr_t>(self._ctx_handle), <uintptr_t>(self._handle)))
```
kkraus14: +1

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -197,6 +224,66 @@ cdef class Stream:
...
+        int
+            Hash value based on the stream handle address and context handle.
+
+        Notes
+        -----
+        Includes the context handle in the hash to prevent collisions when
+        handles are reused across different contexts. While handles are
+        context-scoped and typically not reused across contexts, including
+        the context provides defense-in-depth against hash collisions.
+        The context is fetched and cached during Stream construction.
```
Andy-Jost: I don't quite follow this logic. The way equality is defined, any two streams with identical (reused) handles will compare equal. By including the context in the hash, don't we potentially create a situation where equivalent objects have different hashes?

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -126,27 +126,52 @@ cdef class Stream:
...
+        cdef cydriver.CUresult err
         self._handle = <cydriver.CUstream>(cydriver.CU_STREAM_LEGACY)
         self._builtin = True
+        with nogil:
+            err = cydriver.cuCtxGetCurrent(&ctx)
+        if err == cydriver.CUresult.CUDA_SUCCESS:
+            self._ctx_handle = ctx
+        else:
+            # CUDA not initialized yet, will be lazily initialized later
+            self._ctx_handle = CU_CONTEXT_INVALID
```
Andy-Jost: nit: an inline `if. . . else` would make it more obvious (to me) that the purpose here is to set `_ctx_handle`.

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -126,27 +126,52 @@ cdef class Stream:
...
+        cdef cydriver.CUresult err
         self._handle = <cydriver.CUstream>(cydriver.CU_STREAM_PER_THREAD)
         self._builtin = True
+        with nogil:
+            err = cydriver.cuCtxGetCurrent(&ctx)
+        if err == cydriver.CUresult.CUDA_SUCCESS:
+            self._ctx_handle = ctx
+        else:
+            # CUDA not initialized yet, will be lazily initialized later
+            self._ctx_handle = CU_CONTEXT_INVALID
```
Andy-Jost: ditto. Consider refactoring repeated code into a function.

cuda_core/tests/test_comparable.py
```diff
@@ -0,0 +1,35 @@
...
+    """Test that equality (==) is different from identity (is)."""
+    device = Device(0)
+    device.set_current()
+
+    # Streams: Different objects can be equal
+    s1 = device.create_stream()
+    s2 = Stream.from_handle(int(s1.handle))
+
+    assert s1 == s2, "Streams with same handle are equal"
+    assert s1 is not s2, "But they are not the same object"
```
Andy-Jost, line 31: Out of curiosity, is this desirable or accidental? Would there be any potential benefit to making `s1` and `s2` the same Python object?

cuda_core/cuda/core/experimental/_device.pyx
rparolin: Done.

cuda_core/cuda/core/experimental/_event.pyx
rparolin: Done.

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -165,6 +165,17 @@ cdef class Event:
                 raise CUDAError(err)
             raise RuntimeError(explanation)
 
+    def __hash__(self) -> int:
+        return hash((type(self), self._ctx_handle, <uintptr_t>(self._handle)))
+
+    def __eq__(self, other) -> bool:
+        cdef Event _other
+        try:
+            _other = <Event>other
```
leofang: This could segfault too and we should do a type/isinstance check first.

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -182,16 +193,20 @@ cdef class Event:
...
         """Import an event that was exported from another process."""
         cdef cydriver.CUipcEventHandle data
+        cdef cydriver.CUcontext curr_ctx
+        cdef cydriver.CUdevice dev
         memcpy(data.reserved, <const void*><const char*>(ipc_descriptor._reserved), sizeof(data.reserved))
         cdef Event self = Event.__new__(cls)
         with nogil:
             HANDLE_RETURN(cydriver.cuIpcOpenEventHandle(&self._handle, data))
+            HANDLE_RETURN(cydriver.cuCtxGetCurrent(&curr_ctx))
+            HANDLE_RETURN(cydriver.cuCtxGetDevice(&dev))
```
leofang: I feel this is problematic, unless we can gain internal knowledge and be 100% certain. (I also left a comment when this was checked in, but it was not addressed.) 

We should send the device ID over to the child process too. Unfortunately unlike streams there is no way for us to do a reverse lookup of the device that an event was created on, and so we need to do bookkeeping ourselves and make use of it. There is no guarantee that the current device when `from_ipc_descriptor` is called is the same as the one on which the event was created.

cuda_core/cuda/core/experimental/_stream.pyx
leofang, line 1: For named streams (i.e. the legacy/per-thread default streams), they are just a fixed int (1 & 2, respectively) that are recognized by all contexts. We should not try to make this more complex than it should be.

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -126,27 +126,52 @@ cdef class Stream:
...
             raise ValueError("obj and options cannot be both specified")
         if obj is not None:
             self._handle = _try_to_get_stream_ptr(obj)
             # TODO: check if obj is created under the current context/device
             self._owner = obj
+            with nogil:
+                err = cydriver.cuStreamGetCtx(self._handle, &ctx)
+                if err != cydriver.CUresult.CUDA_SUCCESS:
+                    HANDLE_RETURN(cydriver.cuCtxGetCurrent(&ctx))
+            self._ctx_handle = ctx
```
leofang: Q: Didn't we already have a lazy lookup implemented? Note that most of the `cuCtx*` API callss are very expansive (hundreds of nanoseconds), so we have gone down the routes that 
- if a context handle is already available at construction time, cache it
- otherwise, look it up lazily only when needed

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -169,10 +194,12 @@ cdef class Stream:
         cdef cydriver.CUstream s
         with nogil:
             HANDLE_RETURN(cydriver.cuStreamCreateWithPriority(&s, flags, prio))
+            HANDLE_RETURN(cydriver.cuStreamGetCtx(s, &ctx))
```
leofang: ditto this is too expansive

cuda_core/cuda/core/experimental/_stream.pyx
leofang: @mdboom `other_ = <Stream>other` is unfortunately unsafe. What you had in mind in this context would be `other_ = <Stream?>other`. But what Rob originally had (`isinstance`) was already fine.

https://cython.readthedocs.io/en/latest/src/userguide/extension_types.html#type-testing-and-casting

mdboom: (This is probably irrelevant to the discussion at hand, but for completeness).

We can ignore this if we don't think this is a hot path, but as written, this causes two isinstance checks.  The one that is written, and the one at `<Stream>other`.  (And the second one is slow because it also compares the string names of the types for some reason...)

I think those docs must only apply to `cdef struct`, not `cdef class`. I ask because in cybind we do this in a few places and I was just testing yesterday that this does raise a `TypeError` and generates this code:

```
 if (!(likely(((__pyx_t_3) == Py_None) || likely(__Pyx_TypeTest(__pyx_t_3, __pyx_mstate_global->__pyx_ptype_13test_bindings_4test_Entry))))) __PYX_ERR(0, 72, __pyx_L1_error)
  __pyx_v_other_ = ((struct __pyx_obj_13test_bindings_4test_Entry *)__pyx_t_3);
```

cuda_core/cuda/core/experimental/_event.pyx
rparolin: Done.

cuda_core/tests/test_comparable.py
rparolin: Its a good question, and it why I explicitly put test cases for it to prompt discussion. I don't have any requirements that state that the Stream objects inorder to be equal must be the same object... but its a potential footgun that I wanted to document.

kkraus14: I don't think we want to maintain a lookup table of stream handles in order to make Stream objects singletons in any kind of way, so I would expect each `Stream.from_handle` invocation to return a new Python object.

The most common case for using this is when we're trying to get a `cuda.core.Stream` object from a third party provided cuda stream.

cuda_core/cuda/core/experimental/_device.pyx
```diff
@@ -923,6 +923,21 @@ cdef cydriver.CUcontext _get_primary_context(int dev_id) except?NULL:
...
+    cdef cydriver.CUdevice dev = device_id
+    with nogil:
+        IF CUDA_CORE_BUILD_MAJOR == "12":
+            HANDLE_RETURN(cydriver.cuDeviceGetUuid_v2(&uuid, dev))
+        ELSE:  # 13.0+
+            HANDLE_RETURN(cydriver.cuDeviceGetUuid(&uuid, dev))
+    cdef bytes uuid_b = cpython.PyBytes_FromStringAndSize(uuid.bytes, sizeof(uuid.bytes))
+    cdef str uuid_hex = uuid_b.hex()
+    # 8-4-4-4-12
+    return f"{uuid_hex[:8]}-{uuid_hex[8:12]}-{uuid_hex[12:16]}-{uuid_hex[16:20]}-{uuid_hex[20:]}"
```
kkraus14: super nitpick: did we need to define this as a separate function versus just having this logic underneath the `if self._uuid is None:` block?

cuda_core/cuda/core/experimental/_device.pyx
```diff
@@ -1145,6 +1155,15 @@ class Device:
     def __repr__(self):
         return f"<Device {self._id} ({self.name})>"
 
+    def __hash__(self) -> int:
+        return hash((type(self), self.uuid))
+
+    def __eq__(self, other) -> bool:
+        try:
+            return self._id == other._id
```
kkraus14: Say someone implements a class inheriting from device, i.e. `class MyCustomDevice(Device)`, the hash will be different because we account for `type(self)` in the hash, but we don't account for it in the equality comparison here. Is this intentional and aligned with the behavior we want?

cuda_core/cuda/core/experimental/_event.pyx
```diff
@@ -165,6 +165,16 @@ cdef class Event:
...
 
+    def __hash__(self) -> int:
+        return hash((type(self), self._ctx_handle, <uintptr_t>(self._handle)))
+
+    def __eq__(self, other) -> bool:
+        # Note: using isinstance because `Event` can be subclassed.
+        if not isinstance(other, Event):
+            return NotImplemented
+        cdef Event _other = <Event>other
+        return <uintptr_t>(self._handle) == <uintptr_t>(_other._handle)
```
kkraus14, line 176: +1 on subclass

cuda_core/cuda/core/experimental/_context.pyx
```diff
@@ -29,4 +29,10 @@ cdef class Context:
         return ctx
 
     def __eq__(self, other):
-        return int(self._handle) == int(other._handle)
+        if type(other) is not Context:
+            return NotImplemented
+        cdef Context _other = <Context>other
+        return int(self._handle) == int(_other._handle)
```
kkraus14, line 35: +1 on subclass

cuda_core/tests/test_context.py
```diff
@@ -3,8 +3,88 @@
...
 
 
 def test_context_init_disabled():
     with pytest.raises(RuntimeError, match=r"^Context objects cannot be instantiated directly\."):
         cuda.core.experimental._context.Context()  # Ensure back door is locked.
+
+
+# ============================================================================
+# Context Equality Tests
+# ============================================================================
```
kkraus14, line 16: It would probably be difficult, but is there any way we can add a test for when we have different contexts and/or an invalid context in some way?

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -197,6 +197,27 @@ cdef class Stream:
...
+        # Fast path: compare handles first
+        if <uintptr_t>(self._handle) != <uintptr_t>((_other)._handle):
+            return False
+        # Ensure contexts are initialized for both streams
+        if self._ctx_handle == CU_CONTEXT_INVALID:
+            self._get_context()
+        if _other._ctx_handle == CU_CONTEXT_INVALID:
+            _other._get_context()
+        # Compare contexts as well
+        return <uintptr_t>(self._ctx_handle) == <uintptr_t>((_other)._ctx_handle)
```
kkraus14, line 219: +1 on subclass

cuda_core/cuda/core/experimental/_context.pyx
mdboom: @kkraus14: You mean `not isinstance(other, Context)` rather than `type(other) is not Context`?  I agree that is probably better.

cuda_core/cuda/core/experimental/_device.pyx
mdboom: With duck-typing, this could also do the wrong thing.  Imagine passing an instance of a totally unrelated class with an `_id` member.  Maybe we want something like:

```
return type(other) == type(self) and self._id == other._id
```

cuda_core/tests/test_comparable.py
```diff
@@ -0,0 +1,35 @@
...
+
+    # Device: Same object due to singleton (special case)
+    d1 = Device(0)
+    d2 = Device(0)
+
+    assert d1 == d2, "Devices with same ID are equal"
+    assert d1 is d2, "And they ARE the same object (singleton)"
```
mdboom, line 38: In light of @kkraus14's comment above:

> I don't think we want to maintain a lookup table of stream handles in order to make Stream objects singletons in any kind of way

should we remove this assert?  It may be true in the current implementation, but is it a commitment we want to / need to make?

mdboom (commented):
My comments aren't requirements -- I mostly defer to previous reviewers here.

cuda_core/cuda/core/experimental/_context.pyx
kkraus14: I moreso mean that say you have something that looks like (forgive the pseudocode):

```
class MyContext(Context):
    pass
    
ctx = Context()
my_ctx = MyContext(ctx)

ctx == my_ctx  # True
hash(ctx) == hash(my_ctx)  # False, due to capturing the type in the hash
```

This feels funky from my perspective.

cuda_core/tests/test_comparable.py
kkraus14: We do want `Devices` to be singletons, but we don't want Streams or Events to be singletons 😄

cuda_core/cuda/core/experimental/_device.pyx
rparolin: No, this isn't "needed" I just hoisted it out as a readability change because the inline function was long. I can easily put it back if you strongly against this change.

cuda_core/tests/test_comparable.py
rparolin: Seeing that Devices were singletons is what put on the road of questioning what types needed to be singletons -- and which didn't.

cuda_core/cuda/core/experimental/_device.pyx
rparolin: Doh. I hadn't pushed this change. Thanks for catching this.

leofang (approved):

PR #1204, opened 2025-10-31
[NVVM] Guardrail test_program for version mismatches between driver and PTX compiled binary

abhilash1910:
## Description

Internal NVBug: 5623403
Currently test_program.py does not have an exception case to handle mismatch in PTX compiled binary supported version and driver version, in order  to prevent running of [compiled  PTX code on older drivers](https://docs.nvidia.com/deploy/cuda-compatibility/minor-version-compatibility.html#) . 

@rwgk @leofang  please help to  review, thanks.

### DIFF
--- a/cuda_core/tests/test_program.py
+++ b/cuda_core/tests/test_program.py
@@ -2,14 +2,14 @@
 #
 # SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
 
+import re
 import warnings
 
 import pytest
-import re
 from cuda.core.experimental import _linker
 from cuda.core.experimental._module import Kernel, ObjectCode
 from cuda.core.experimental._program import Program, ProgramOptions
-from cuda.core.experimental._utils.cuda_utils import driver, handle_return, CUDAError
+from cuda.core.experimental._utils.cuda_utils import CUDAError, driver, handle_return
 
 cuda_driver_version = handle_return(driver.cuDriverGetVersion())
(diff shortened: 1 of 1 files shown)


abhilash1910:
pre-commit.ci autofix

leofang:
/ok to test [dd78838](https://github.com/NVIDIA/cuda-python/pull/1204/commits/dd78838afab3e74720bf7027630515405926d434)

cuda_core/tests/test_program.py
```diff
@@ -317,7 +318,12 @@ def test_nvvm_program_creation_compilation(nvvm_ir):
     assert program.backend == "NVVM"
     assert program.handle is not None
     obj = program.compile("ptx")
-    ker = obj.get_kernel("simple")  # noqa: F841
+    try:
+        ker = obj.get_kernel("simple")  # noqa: F841
+    except CUDAError as e:
+        if re.search(r"CUDA_UNSUPPORTED_PTX_VERSION", str(e)):
+            pytest.xfail("PTX version not supported by current CUDA Driver")
+        raise
```
leofang, line 326: Q: Shouldn't this be
```suggestion
        else:
            raise
```

rwgk: The `else` isn't needed: `pytest.xfail` exits the test right at the call site.

But the `else` doesn't do harm either.

I tend to fully rely on the pytest behavior, i.e. not add redundant flow control, for more compact code.

rwgk (approved):

Issue #985, opened 2025-09-18
CI: Figure out a test plan to cover all Windows driver modes


leofang:
TCC, WDDM, MCDM

leofang:
This should be considered together with #1024. Say,
- py310 + WDDM
- py311 + MCDM
- py312 + TCC
- ...

leofang:
WSL should be included once it's ready (#1039)

FurkanGozukara:
can we use MCDM on Windows with consumer gpus?

https://github.com/NVIDIA/cuda-python/issues/1207

PR #1209, opened 2025-11-03
Set CU_CTX_SCHED_BLOCKING_SYNC option in tests.

Andy-Jost:
By default, CPU threads spin while waiting for GPU synchronization. This changes the option to block instead while testing.

See [cuDevicePrimaryCtxSetFlags](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__PRIMARY__CTX.html#group__CUDA__PRIMARY__CTX_1gd779a84f17acdad0d9143d9fe719cfdf) for details.

### DIFF
--- a/cuda_core/tests/conftest.py
+++ b/cuda_core/tests/conftest.py
@@ -32,7 +32,7 @@ def init_cuda():
     device.set_current()
 
     # Set option to avoid spin-waiting on synchronization.
-    if os.environ.get("CUDA_CORE_TEST_BLOCKING_SYNC") is not None:
+    if int(os.environ.get("CUDA_CORE_TEST_BLOCKING_SYNC", 0)) != 0:
         handle_return(
             driver.cuDevicePrimaryCtxSetFlags(device.device_id, driver.CUctx_flags.CU_CTX_SCHED_BLOCKING_SYNC)
         )


leofang:
Q: Why is this needed? We don't expect our users to do this.

Andy-Jost:
This is not essential. It just reduces load on the test machines. I was annoyed by the spinning when working with certain multiprocessing tests and happened to come across this option in the docs. It's a 5-minute change so I figured why not.

leofang (changes requested):
> This is not essential. It just reduces load on the test machines.

Then we should not make this change. We need to test the real-life use scenarios. With the upcoming coverage of various Windows driver modes (https://github.com/NVIDIA/cuda-python/issues/985) our test env will get more sensitive to the underlying driver than ever. Changing the global state is not helping if any issue comes up.

leofang:
> I was annoyed by the spinning when working with certain multiprocessing tests

This is why I kept pushing for IPC events so as to enabling stream-ordered IPC tests. Device synchronization is expansive at the library test scale. We should avoid writing tests that require synchronization. We've had enough in the past to learn in a hard way that stream ordering is critical 😅

leofang:
I'll take a look at the performance issue. @Andy-Jost do you have in mind speeding up the aggregate performance of all tests under `cuda_core/tests/memory_ipc`?

Andy-Jost:
Let me clarify: I'm not saying there's any performance issue. The annoyance I ran into was during development, especially while using `code.interact` when a latch kernel was spinning. My laptop was burning up haha!

I believe most of the wait in `memory_ipc` is on `multiprocessing.Queue` or similar objects, so it is already blocking. As far as I know, we don't have any significant CUDA synchronization waits in those tests.

leofang:
I would be OK if you could guard this change behind a test-time env var (say, `CUDA_CORE_TEST_BLOCKING_SYNC`) that makes your development workflow faster. I am still concerned about the blocking nature of our tests, though, but let's make your life easier first.

Andy-Jost:
Sounds good. Latest upload adds an environment variable guard.

cuda_core/tests/conftest.py
```diff
@@ -29,6 +30,13 @@ def init_cuda():
     # TODO: rename this to e.g. init_context
     device = Device()
     device.set_current()
+
+    # Set option to avoid spin-waiting on synchronization.
+    if os.environ.get("CUDA_CORE_TEST_BLOCKING_SYNC") is not None:
```
leofang:
```suggestion
    if int(os.environ.get("CUDA_CORE_TEST_BLOCKING_SYNC", 0)) == 1:
```

leofang (approved):

leofang (approved):

PR #1216, opened 2025-11-04
[NUMBA][CUDA] Add numba_debug flag.

ayermolo:
Adds a flag numba_debug to enable NUMBA CUDA debug capabilities.

### DIFF
--- a/cuda_core/tests/test_program.py
+++ b/cuda_core/tests/test_program.py
@@ -197,7 +197,7 @@ def ptx_code_object():
             ProgramOptions(debug=True, numba_debug=True),
             marks=pytest.mark.skipif(
                 (_get_nvrtc_version_for_tests() or 0) < 13200,
-                reason="numba_debug requires libNVVM >= 13.2",
+                reason="numba_debug requires NVRTC >= 13.2",
             ),
         ),
         ProgramOptions(relocatable_device_code=True, max_register_count=32),


cpcloud:
@ayermolo Thanks! Can you add a test for this? Adding to an existing test is also fine.

ayermolo:
> @ayermolo Thanks! Can you add a test for this? Adding to an existing test is also fine.

Thanks for taking a look.

I thought I did to cuda_core/tests/test_program.py

Is there a better one?

cuda_core/cuda/core/experimental/_program.py
```diff
@@ -255,6 +255,9 @@ class ProgramOptions:
     minimal : bool, optional
         Omit certain language features to reduce compile time for small programs.
         Default: False
+    numba_debug : bool, optional
+        Enable Numba-specific flag for debugging.
+        Default: False
```
leofang: Can we keep this undocumented?

ayermolo: done

ayermolo:
> Q: What is the reason that we need this? How is `--numba-debug` consumed (and by whom)?

It will be consumed by nvrtc library, and used by: https://github.com/numba/numba/blob/main/numba/cuda/cudadrv/nvrtc.py

leofang:
/ok to test [3aed97d](https://github.com/NVIDIA/cuda-python/pull/1216/commits/3aed97dd7ff0773776d27a63ef5e7619594c49ab)

leofang:
@ayermolo could you take a look at the test failures?

ayermolo:
pre-commit.ci autofix

cuda_core/tests/test_program.py
```diff
@@ -176,6 +193,13 @@ def ptx_code_object():
     [
         ProgramOptions(name="abc"),
         ProgramOptions(device_code_optimize=True, debug=True),
+        pytest.param(
+            ProgramOptions(debug=True, numba_debug=True),
+            marks=pytest.mark.skipif(
+                (_get_nvrtc_version_for_tests() or 0) < 13200,
+                reason="numba_debug requires libNVVM >= 13.2",
```
gmarkall: The `skipIf` checks the NVRTC version when the reason given is the NVVM version. Is this an error, or is it because there's no way to get the NVVM version and the NVRTC and NVVM versions are assumed to be consistent?

ayermolo: I just forgot to change it. Originally for internal testing I was using _get_libnvvm_version_for_tests check and extending it for 13.2, but then changed to NVRTC check. Which I think should be better for this particular case?

leofang: @ayermolo using NVRTC version as the proxy is not accurate in the Python land. Often times we are dealing with packaging issues that result in mix-n-match'd versions (from different CTK versions). This is a better way to get libnvvm version: https://github.com/NVIDIA/cuda-python/blob/9aaa671dfd9699eb93f256f50d069999d38290f2/cuda_core/tests/test_program.py#L59

leofang: Sorry, I guess it is just that the "reason" string is confusing. Let's get it fixed:
```suggestion
                reason="numba_debug requires NVRTC >= 13.2",
```

leofang:
/ok to test [20ba276](https://github.com/NVIDIA/cuda-python/pull/1216/commits/20ba2761c9fc6fcdf9ba62480bdaa7f345f6d1ff)

kkraus14:
`numba-cuda` only goes through nvrtc when linking in some CUDA C++ code, which isn't the norm. It normally generates NVVM IR. Should we add this flag to a test of NVVM IR instead of for CUDA C++ code?

jiel-nv:
> `numba-cuda` only goes through nvrtc when linking in some CUDA C++ code, which isn't the norm. It normally generates NVVM IR. Should we add this flag to a test of NVVM IR instead of for CUDA C++ code?

@kkraus14 To my understanding, the numba_debug flag is specifically designed for Numba's @cuda.jit(link=[...]) decorator, so that the linked external CUDA C++ files can be complied by cuda-python's NVRTC backend with debug info. This enables proper debugging of the entire linked kernel. So I think the current CUDA C++ test is the right place to add that flag.

ayermolo:
If all the concerns are addressed @leofang can you merge this PR please. :)

leofang:
Thanks @ayermolo!

PR #1221, opened 2025-11-07
cythonize `_launch_config.py`

### LINKED ISSUE #1077: Cythonize `_launch_config.py`
@emcastillo might have some bw for this one?

rparolin:
Converts `cuda.core.experimental._launch_config` from pure Python to Cython for improved performance while maintaining full API compatibility.
 
 Closes #1077

### DIFF
--- a/cuda_core/cuda/core/experimental/_launch_config.pyx
+++ b/cuda_core/cuda/core/experimental/_launch_config.pyx
@@ -115,10 +115,7 @@ cdef class LaunchConfig:
             self.shmem_size = shmem_size
 
         # Handle cooperative_launch
-        if cooperative_launch is None:
-            self.cooperative_launch = False
-        else:
-            self.cooperative_launch = cooperative_launch
+        self.cooperative_launch = cooperative_launch
 
         # Validate cooperative launch support
         if self.cooperative_launch and not Device().properties.cooperative_launch:


cuda_core/cuda/core/experimental/_launch_config.pxd
```diff
@@ -0,0 +1,19 @@
...
+cdef bint _inited
+cdef bint _use_ex
+
+cdef void _lazy_init() except *
+
+cdef class LaunchConfig:
+    """Customizable launch options."""
+    cdef public tuple grid
+    cdef public tuple cluster
+    cdef public tuple block
```
mdboom, line 15: IIUC, these are all 3-tuples of ints.  Those are pretty expensive, particularly if used from other C or Cython code.  Could we make these `int[3]` instead?  (And of course the updates to `convert_to_3_tuple`...)

mdboom (approved):
LGTM, modulo the optional change to improve memory and construction performance by using `int[3]` (if that matters, it may not).  

But this raises a larger question in my mind that we really should create a benchmarking suite for `cuda.core` (`cuda.bindings` has one) so that we can measure whether these changes are actually improving something we care about.  I don't see anything in this PR that should significantly improve performance (there's no obvious inner loops and it's still using a lot of high-level Python datatypes), but it's easy to miss those things and I don't understand the context in which this is used.  We should be measuring so we can weigh the tradeoff of moving code to Cython (additional complexity, loss of introspection etc.)  (But that's not for this PR to solve...)

rparolin:
> LGTM, modulo the optional change to improve memory and construction performance by using `int[3]` (if that matters, it may not).
> 
> But this raises a larger question in my mind that we really should create a benchmarking suite for `cuda.core` (`cuda.bindings` has one) so that we can measure whether these changes are actually improving something we care about. I don't see anything in this PR that should significantly improve performance (there's no obvious inner loops and it's still using a lot of high-level Python datatypes), but it's easy to miss those things and I don't understand the context in which this is used. We should be measuring so we can weigh the tradeoff of moving code to Cython (additional complexity, loss of introspection etc.) (But that's not for this PR to solve...)

I totally agree that we should start with a benchmarking suite that explicitly validates Cythonizing code.  I'll start looking into that.

cuda_core/cuda/core/experimental/_launch_config.pxd
rparolin: This is a great suggestion. I'll hold off for right now b/c its a breaking API change and I don't want to impact user code.

leofang:
> I totally agree that we should start with a benchmarking suite that explicitly validates Cythonizing code. I'll start looking into that.

I don't understand the claim. We have been using short snippets to measure local cythonization impacts. Please search `timeit` across the repo: https://github.com/search?q=repo%3ANVIDIA%2Fcuda-python+timeit&type=issues. In this case, try
```python
%timeit config = LaunchConfig(grid=(1024,), block=(256,))
```
to get a feeling before/after the change. 

> I don't see anything in this PR that should significantly improve performance

Without performance data we should not merge a cythonization PR because indeed cythonization may not always help. (I was doing a bad action in pushing mine back in 0.4.0, but as I noted during meetings and in an internal thread, I did use the existing `%timeit` snippets to guide the work, I just did the work in a rush that I did not compile the results and post in my PR, but in my defense no one asked 🙂)

mdboom:
> I totally agree that we should start with a benchmarking suite that explicitly validates Cythonizing code. I'll start looking into that.

I was thinking specifically of something that could be run more frequently, like the [`pytest-benchmarking` suite in `cuda_bindings`](https://github.com/NVIDIA/cuda-python/tree/main/cuda_bindings/benchmarks).  I didn't see one for `cuda_core`.

Someday, when we feel we can spare the cycles, we should set up continuous benchmarking hardware and create reports etc., but having `pytest-benchmarking` snippets is an easy first step.

Issue #473, opened 2025-02-26
BUG: StridedMemoryView fails to wrap NumPy array


carterbox:
I'm using numpy 2.0.2 and cuda-core 0.1.1. I cannot successfully convert a numpy array into a StridedMemoryView. My understanding is that this should be possible because numpy arrays implement the dlpack interface.

```python
import nvmath  # used to patch cupy imports of nvidia-wheels
import numpy as np
import cuda.core.experimental as ccx
import cupy as cp
import torch

host_array = np.ones(50)
device_array = cp.ones(50)

device = ccx.Device()
device.set_current()

torch_array = torch.ones(50)

stream = device.create_stream()

# StridedMemoryView cannot wrap NumPy arrays because of impossible constraints on the
# stream_ptr parameter.

# Try 1: Provide a valid stream pointer
print(host_array)
wrapped_host_array = ccx.utils.StridedMemoryView(host_array, stream_ptr=stream.handle)
# RuntimeError: NumPy only supports stream=None.
print(wrapped_host_array)

# Try 2: Provide no stream pointer
print(host_array)
wrapped_host_array = ccx.utils.StridedMemoryView(host_array, stream_ptr=None)
# BufferError: stream=None is ambiguous with view()
print(wrapped_host_array)

print(torch_array)
wrapped_torch_array = ccx.utils.StridedMemoryView(torch_array, stream_ptr=None)
# BufferError: stream=None is ambiguous with view()
print(wrapped_torch_array)

# The following work

print(torch_array)
wrapped_torch_array = ccx.utils.StridedMemoryView(torch_array, stream_ptr=stream.handle)
print(wrapped_torch_array)

torch_array = torch_array.to('cuda')

print(torch_array)
wrapped_torch_array = ccx.utils.StridedMemoryView(torch_array, stream_ptr=stream.handle)
print(wrapped_torch_array)

print(device_array)
wrapped_device_array = ccx.utils.StridedMemoryView(device_array, stream_ptr=stream.handle)
print(wrapped_device_array)
```

leofang:
Could you try passing `stream_ptr=-1`?
https://github.com/NVIDIA/cuda-python/blob/5d86ebbc6b5adfea4fed8c7e79612a499a31e4ad/cuda_core/tests/test_utils.py#L50-L53
It is a documented deviation from the `__dlpack__` interface: 
[...]

Issue #285, opened 2024-12-11
`StridedMemoryView` fails with Jax arrays


leofang:
@yangcal was trying the `StridedMemoryView` but it doesn’t seem to work for jax array, at least with cuda12.2. Below is the script and the error log:
```python
import cupy as cp
import numpy as np
import jax.numpy as jnp

from cuda.core.experimental.utils import args_viewable_as_strided_memory

@args_viewable_as_strided_memory((0,))
def parse_tensor(arr):
    view = arr.view(-1)
    print(type(arr), type(view))
    print(f"shape={view.shape}")
    print(f"strides={view.strides}")

for module in (np, cp, jnp):
    arr = module.eye(2)
    print(f"module={module.__name__}")
    parse_tensor(arr)
```
Error Log:
```
E1211 13:52:21.674104 2256128 ptx_compiler_helpers.cc:71] *** WARNING *** Invoking ptxas with version 12.2.140, which corresponds to a CUDA version <=12.6.2. CUDA versions 12.x.y up to and including 12.6.2 miscompile certain edge cases around clamping.
Please upgrade to CUDA 12.6.3 or newer.
module=jax.numpy
Traceback (most recent call last):
  File "/home/scratch.yangg_sw/software/cuda-python/cuda_core/examples/tmp.py", line 18, in <module>
    parse_tensor(arr)
  File "cuda/core/experimental/_memoryview.pyx", line 372, in cuda.core.experimental._memoryview.args_viewable_as_strided_memory.wrapped_func_with_indices.wrapped_func
  File "/home/scratch.yangg_sw/software/cuda-python/cuda_core/examples/tmp.py", line 10, in parse_tensor
    view = arr.view(-1)
           ^^^^^^^^^^^^
  File "cuda/core/experimental/_memoryview.pyx", line 146, in cuda.core.experimental._memoryview._StridedMemoryViewProxy.view
  File "cuda/core/experimental/_memoryview.pyx", line 148, in cuda.core.experimental._memoryview._StridedMemoryViewProxy.view
  File "cuda/core/experimental/_memoryview.pyx", line 180, in cuda.core.experimental._memoryview.view_as_dlpack
  File "/home/Self/marie/miniconda3/envs/jax/lib/python3.12/site-packages/jax/_src/array.py", line 446, in __dlpack__
    return to_dlpack(self, stream=stream,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1224, opened 2025-11-07
feat: add from_* style constructor classmethods to StridedMemoryView and make constructor amenable to future from_*-style constructors

cpcloud:
This PR adds three methods to conform better to APIs familiar to Python users and to ease future work on `StridedMemoryView`
adding support for constructing them from other objects besides CAI and dlpack objects.

Three methods are added:

- `from_dlpack(obj, stream_ptr)` same functionality as before
- `from_cuda_array_interface(obj, stream_ptr)` same functionality as before
- `from_any_interface(obj, stream_ptr)` same behavior as the previous constructor

The constructor changes to a simple field initialization, in an effort to make
it easier to add a `from_buffer` (naming TBD) method that allows construction
from an owned raw buffer (without shape and stride information), which may
require passing in additional information such as shape and stride to the
constructor.

### DIFF
--- a/cuda_core/cuda/core/experimental/_memoryview.pyx
+++ b/cuda_core/cuda/core/experimental/_memoryview.pyx
@@ -88,20 +88,155 @@ cdef class StridedMemoryView:
     cdef DLTensor *dl_tensor
 
     # Memoized properties
-    cdef tuple _shape
-    cdef tuple _strides
-    cdef bint _strides_init  # Has the strides tuple been init'ed?
-    cdef object _dtype
+    cdef:
+        tuple _shape
+        tuple _strides
+        # a `None` value for _strides has defined meaning in dlpack and
+        # the cuda array interface, meaning C order, contiguous.
+        #
+        # this flag helps prevent unnecessary recompuation of _strides
+        bint _strides_init
+        object _dtype
 
-    def __init__(self, obj=None, stream_ptr=None):
-        if obj is not None:
-            # populate self's attributes
-            if check_has_dlpack(obj):
-                view_as_dlpack(obj, stream_ptr, self)
-            else:
-                view_as_cai(obj, stream_ptr, self)
+    def __init__(
+        self,
+        *,
+        ptr: intptr_t,
+        device_id: int,
+        is_device_accessible: bint,
+        readonly: bint,
+        metadata: object,
+        exporting_obj: object,
+        dl_tensor: intptr_t = 0,
+    ) -> None:
+        self.ptr = ptr
+        self.device_id = device_id
+        self.is_device_accessible = is_device_accessible
+        self.readonly = readonly
+        self.metadata = metadata
+        self.exporting_obj = exporting_obj
+        self.dl_tensor = <DLTensor*>dl_tensor
+        self._shape = None
+        self._strides = None
+        self._strides_init = False
+        self._dtype = None
+
+    @classmethod
+    def from_dlpack(cls, obj: object, stream_ptr: int | None=None) -> StridedMemoryView:
+        cdef int dldevice, device_id
+        cdef bint is_device_accessible, is_readonly
+        is_device_accessible = False
+        dldevice, device_id = obj.__dlpack_device__()
+        if dldevice == _kDLCPU:
+            assert device_id == 0
+            device_id = -1
+            if stream_ptr is None:
+                raise BufferError("stream=None is ambiguous with view()")
+            elif stream_ptr == -1:
+                stream_ptr = None
+        elif dldevice == _kDLCUDA:
+            assert device_id >= 0
+            is_device_accessible = True
+            # no need to check other stream values, it's a pass-through
+            if stream_ptr is None:
+                raise BufferError("stream=None is ambiguous with view()")
+        elif dldevice in (_kDLCUDAHost, _kDLCUDAManaged):
+            is_device_accessible = True
+            # just do a pass-through without any checks, as pinned/managed memory can be
+            # accessed on both host and device
         else:
-            pass
+            raise BufferError("device not supported")
+
+        cdef object capsule
+        try:
+            capsule = obj.__dlpack__(
+                stream=int(stream_ptr) if stream_ptr else None,
+                max_version=(DLPACK_MAJOR_VERSION, DLPACK_MINOR_VERSION))
+        except TypeError:
+            capsule = obj.__dlpack__(
+                stream=int(stream_ptr) if stream_ptr else None)
+
+        cdef void* data = NULL
+        cdef DLTensor* dl_tensor
+        cdef DLManagedTensorVersioned* dlm_tensor_ver
+        cdef DLManagedTensor* dlm_tensor
+        cdef const char *used_name
+        if cpython.PyCapsule_IsValid(
+                capsule, DLPACK_VERSIONED_TENSOR_UNUSED_NAME):
+            data = cpython.PyCapsule_GetPointer(
+                capsule, DLPACK_VERSIONED_TENSOR_UNUSED_NAME)
+            dlm_tensor_ver = <DLManagedTensorVersioned*>data
+            dl_tensor = &dlm_tensor_ver.dl_tensor
+            is_readonly = (dlm_tensor_ver.flags & DLPACK_FLAG_BITMASK_READ_ONLY) != 0
+            used_name = DLPACK_VERSIONED_TENSOR_USED_NAME
+        else:
+            assert cpython.PyCapsule_IsValid(
+                capsule, DLPACK_TENSOR_UNUSED_NAME)
+            data = cpython.PyCapsule_GetPointer(
+                capsule, DLPACK_TENSOR_UNUSED_NAME)
+            dlm_tensor = <DLManagedTensor*>data
+            dl_tensor = &dlm_tensor.dl_tensor
+            is_readonly = False
+            used_name = DLPACK_TENSOR_USED_NAME
+
+        cpython.PyCapsule_SetName(capsule, used_name)
+
+        return cls(
+            ptr=<intptr_t>dl_tensor.data,
+            device_id=int(device_id),
+            is_device_accessible=is_device_accessible,
+            readonly=is_readonly,
+            metadata=capsule,
+            exporting_obj=obj,
+            dl_tensor=<intptr_t>dl_tensor,
+        )
+
+    @classmethod
+    def from_cuda_array_interface(cls, obj: object, stream_ptr: int | None=None) -> StridedMemoryView:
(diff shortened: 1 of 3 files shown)


cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -88,20 +88,149 @@ cdef class StridedMemoryView:
...
+        exporting_obj: object,
+        dl_tensor: intptr_t = -1,
+    ) -> None:
+        self.ptr = ptr
+        self.device_id = device_id
+        self.is_device_accessible = is_device_accessible
+        self.readonly = readonly
+        self.metadata = metadata
+        self.exporting_obj = exporting_obj
+        self.dl_tensor = (<DLTensor*>NULL) if dl_tensor is -1 else (<DLTensor*>dl_tensor)
```
mdboom: Is there a reason to not just have a default val of `NULL` for `dl_tensor`?  It wouldn't suffer from the "don't use mutable collections as default arguments" problem that Python usually has.

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -88,20 +88,149 @@ cdef class StridedMemoryView:
...
+            metadata=capsule,
+            exporting_obj=obj,
+            dl_tensor=<intptr_t>dl_tensor,
+        )
+
+    @classmethod
+    def from_any_interface(cls, obj: object, stream_ptr: int | None = None) -> StridedMemoryView:
+        if check_has_dlpack(obj):
+            return cls.from_dlpack(obj, stream_ptr)
+        return cls.from_cuda_array_interface(obj, stream_ptr)
```
mdboom: Minor nit: Maybe put this "generic" one first or last?  I was surprised it came in the middle of two "specific" constructors.

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -136,9 +265,9 @@ cdef class StridedMemoryView:
         return self._shape
 
     @property
-    def strides(self) -> Optional[tuple[int]]:
+    def strides(self) -> tuple[int, ...]:
```
mdboom: `strides` of `NULL` (or `None`) has a defined meaning in dlpack and CAI, which is "C order, contiguous".  That's the reason for the separate flag `_strides_init`, since it could be init'ed and also legitimately `None`, to prevent unnecessary recomputation of the cached value.  Also, it does mean the return type of this property will be frequently `None`, so we shouldn't change the signature to always be a `tuple` here.

We could change this to actually compute the numeric value when it is `None`, but that could/should be done separately.

cuda_core/cuda/core/experimental/_memoryview.pyx
cpcloud: I couldn't figure out how to spell the type that would allow such a default value. Though, now that you mention it perhaps a typedef would work here since Cython doesn't parse `T*` pointer types in annotations.

Let me give that a try.

cuda_core/cuda/core/experimental/_memoryview.pyx
cpcloud: Yes, that's me just being a bit lazy about placement :)

cuda_core/cuda/core/experimental/_memoryview.pyx
cpcloud: I see. Wasn't aware of that detail.

I'll revert this change and add a comment for future folks who may think that flag isn't doing anything.

cuda_core/cuda/core/experimental/_memoryview.pyx
cpcloud: `NULL` won't work because `__init__` is in Python space, and `NULL` isn't so you get this error:

```
 │ │   Error compiling Cython file:
 │ │   ------------------------------------------------------------
 │ │   ...
 │ │           device_id: int,
 │ │           is_device_accessible: bint,
 │ │           readonly: bint,
 │ │           metadata: object,
 │ │           exporting_obj: object,
 │ │           dl_tensor: DLTensorPtr = NULL,
 │ │           ^
 │ │   ------------------------------------------------------------
 │ │   cuda/core/experimental/_memoryview.pyx:107:8: Cannot convert Python object argument to type 'DLTensorPtr'
```

where:

```
ctypedef DLTensor* DLTensorPtr
```

mdboom: Can we just use 0 instead of NULL?  I know it's not ideal, but seems better than -1 to me.  But this opinion is weakly held if `-1` is used conventionally in this kind of Cython code or something.

cpcloud: I guess, but does this have a huge impact on this API? I don't think so.

I took -1 from the other surrounding uses of it (`stream_ptr`, and `device_id`).

However, I would also really like to not die on this hill :)

cpcloud: Changed to 0

cuda_core/cuda/core/experimental/_memoryview.pyx
cpcloud: Done.

cuda_core/cuda/core/experimental/_memoryview.pyx
cpcloud: Fixed.

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -136,7 +271,7 @@ cdef class StridedMemoryView:
...
```
cpcloud: Oops, let me fix the annotation.

mdboom (approved):
LGTM

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -88,20 +88,155 @@ cdef class StridedMemoryView:
...
-    cdef bint _strides_init  # Has the strides tuple been init'ed?
-    cdef object _dtype
-
-    def __init__(self, obj=None, stream_ptr=None):
-        if obj is not None:
-            # populate self's attributes
-            if check_has_dlpack(obj):
-                view_as_dlpack(obj, stream_ptr, self)
-            else:
-                view_as_cai(obj, stream_ptr, self)
```
leofang, line 102: @cpcloud This is a breaking change. Can we please restore the old behavior and add a deprecation warning? We need a release note, too. 

If we want to push for alternative constructors, we should eventually raise in `__init__` following what we do in other objects.

cpcloud: Can you point me to downstream breakages so I can understand the impact? Across all of GitHub the only use of `StridedMemoryView` is in a comment:

https://github.com/search?q=StridedMemoryView+NOT+is%3Afork+NOT+repo%3ANVIDIA%2Fcuda-python+NOT+owner%3APopcornFX&type=code

I'd strongly prefer not to restore the behavior without evidence that this change causes problems.

Happy to add the release note.

leofang: We have real users that cannot be found this way. Examples include issues filed by people outside of the team:
- https://github.com/NVIDIA/cuda-python/issues/180
- https://github.com/NVIDIA/cuda-python/issues/1186
- https://github.com/NVIDIA/cuda-python/issues/1043
- https://github.com/NVIDIA/cuda-python/issues/473
- https://github.com/NVIDIA/cuda-python/issues/285#issue-2734178402

We have way too many invisible users at this point (check out the download counts from e.g. [pypistats](https://pypistats.org/packages/cuda-core)). We have also been slammed by our PM for breaking too often and too much (despite I disagreed in that case; in any case I was the one arguing and taking time in sorting things out, so plz back me up).

We need to make sure major breaking changes like this have a deprecation period, even before GA. In particular, in this case it is absolutely fine that we add extra constructors without breaking anything. I am not sure why we must leap so aggressively.

rparolin: +1 to @leofang guidance.  Breaking existing public facing API should go through our teams standard deprecation flow where we notify users at least a release before to when we remove the code.

mdboom: Apologies for approving while misunderstanding the backward-compat guarantees of the `experimental` namespace.

I think this is all a reasonable policy to have, but challenging to follow as long as we are releasing from main.  If we had a maintenance branch, we could put the "backward-compatible-with-deprecation-warnings" version of this on the maintenance branch, and the aspirational "where-we-want-to-be" on main.  As it stands, we need to do the first now, put a reminder in a calendar to do the second and hope that reviews get it merged in the expected window.

leofang: Sorry, I've meant to follow up here and offline with Phillip, but I got distracted (as always) by other things including the pixi stuff that Phillip started playing around with 😛 

I did not mean that this PR is bad. I apologize if my earlier comments gave this impression. The alternative constructors are totally fine. They give users finer control if they already have prior knowledge on what protocols a foreign object supports. I meant that we just need to preserve the default constructor's behavior for a little longer, at least for another minor release. We could just revert the change on `__init__()` and leave everything else as-is, though now that we have an urgent (TBD) patch release in preparation, it is good that it's [reverted for now](https://github.com/NVIDIA/cuda-python/pull/1232).

We're dealing with an interesting situation. On the one hand we're still beta/experimental software, and we do allow breaking changes in each minor release. On the other hand we are perhaps too successful in telling our CUDA Python story that we have unexpectedly high number of users, and this does increasingly keep me up at night. If we break too often, downstream projects end up pinning too tightly ([example](https://nvidia.slack.com/archives/C03R44Q1YFQ/p1763167632740679)), resulting in NVIDIA software unable to be installed in the same environment. This hurts our storytelling. 

I suggest that we clearly highlight all PRs that contain breaking changes with the ["breaking" label](https://github.com/NVIDIA/cuda-python/issues?q=state%3Aopen%20label%3Abreaking) and a rel-note entry under the Breaking Change section. With such marks we can make informed decisions and evaluate all breaking changes on a case-by-case basis. There is also no need to seek short turnaround in merging breaking PRs. We still want to design `cuda.core` properly in a future-looking fashion to avoid as much breaks as possible, it's been one of the core principles since Day 1, and for that I am fine taking a bit more time thinking through the design.

cpcloud: I didn't interpret your comments as this PR being bad, just haven't had a chance to re-PR the changes while preserving the current constructor behavior. I will PR that today.

PR #1226, opened 2025-11-07
chore: add pixi for cuda-core and cuda-bindings

cpcloud:
This PR is a redo of pixi environment setup that combines `cuda-core` and
`cuda-bindings` into a single PR.

I am not trying to capture every possible environment combination of CUDA and
Python right now, just enough to get a minimal development environment working
for both of these projects on linux-64. What's in these `pixi.toml` files is
the scope I intended.

The biggest thing that isn't working yet, but that isn't a blocker is the
ability of `cuda-core` to depend on `cuda-bindings` in the repo. During the
`cuda-core` build phase, `cuda/bindings.pyx` isn't found and that causes the
build to fail. I haven't yet investigated that.

### DIFF
--- a/cuda_core/pixi.toml
+++ b/cuda_core/pixi.toml
@@ -98,7 +98,6 @@ setuptools = ">=77"
 cython = ">=3.1"
 cuda-cudart-dev = "*"
 cuda-profiler-api = "*"
-# TODO: figure out why we can't set editable=true here
 cuda-bindings = { path = "../cuda_bindings" }
 
 [package.target.linux-64.host-dependencies]
@@ -113,7 +112,6 @@ cuda-crt-dev_win-64 = "*"
 [package.run-dependencies]
 python = "*"
 numpy = "*"
-# TODO: figure out why we can't set editable=true here
 cuda-bindings = { path = "../cuda_bindings" }
 
 [target.linux.tasks.build-cython-tests]


cuda_core/pixi.toml
```diff
@@ -0,0 +1,83 @@
...
+cuda-profiler-api = "*" # for some reason, without this, cuda.h isn't found ¯\_(ツ)_/¯
+ccache = "*"
+gxx = "*"
+gcc = "*"
+
+[package.host-dependencies]
+python = "*"
+setuptools = ">=77"
+cython = ">=3.1"
+cuda-bindings = "*"
```
cpcloud: Ideally this would work with `cuda-bindings = { path = "../cuda_bindings" }`, but it doesn't.

pytest.ini
```diff
@@ -2,6 +2,10 @@
 # SPDX-License-Identifier: Apache-2.0
 
 [pytest]
+norecursedirs =
+    cuda_bindings/examples
+    cuda_core/examples
```
leofang, line 7: Q: What does this do?

cpcloud: Ignore test discovery in those directories.

cpcloud: pytest is aggressive about importing modules, and devs shouldn't need all the stuff in those examples in their env to run the test suite.

cpcloud: This isn't strictly necessary, but it's _REALLY_ nice to be able to go into a repo and just write `pytest` and have that do the right thing. In the sense that we want people to have a pleasant development experience, it's necessary.

Right now, to avoid this config you have to know that you must write `pytest tests` which isn't unreasonable, but it's easy to make both work.

leofang: I agree, just am a bit surprised that this matters at all, since there is no file / function name that is prefixed with `test_` in the example folders, and so pytest should already ignore them by default.

leofang:
Now that we have pixi support, how about adding CI pipelines? (#280) 😉

pytest.ini
cpcloud: `*_test.py` is a matched-by-default pattern as well https://docs.pytest.org/en/stable/explanation/goodpractices.html#conventions-for-python-test-discovery

cuda_bindings/pixi.toml
```diff
@@ -0,0 +1,125 @@
...
+[workspace]
+channels = ["conda-forge"]
+platforms = ["linux-64", "linux-aarch64", "win-64"]
+preview = ["pixi-build"]
+
+[workspace.build-variants]
+python = ["3.10.*", "3.11.*", "3.12.*", "3.13.*", "3.14.*"]
+
+[target.linux.tasks.build-cython-tests]
+cmd = ["$PIXI_PROJECT_ROOT/tests/cython/build_tests.sh"]
```
leofang: This is really nice, I did not know pixi has per-arch target. How about Windows? We have `build_tests.bat` too.

cpcloud: _grumbles_ windows 🤣

cuda_bindings/pixi.toml
```diff
@@ -0,0 +1,125 @@
...
+
+[target.linux.tasks.test]
+cmd = ["pytest", "$PIXI_PROJECT_ROOT"]
```
leofang: Q: So in our case, does `$PIXI_PROJECT_ROOT` point to `cuda-python/` or `cuda-python/cuda_bindings/`?

cuda_bindings/pixi.toml
cpcloud: I can do that in a follow up

cuda_bindings/pixi.toml
cpcloud: `cuda_bindings`, `cuda_core`, etc, wherever the `pixi.toml` file is located.

cpcloud:
> Now that we have pixi support, how about adding CI pipelines? (#280) 😉

Something I will look into for sure. I would like to get docs building first (or maybe as a parallel work stream).

cuda_bindings/pixi.toml
cpcloud: The thing that's nice about that is that wherever you are in the repo, `pixi run -e cu13 test` will run the tests from the project root.

cuda_bindings/pixi.toml
```diff
@@ -0,0 +1,125 @@
...
+depends-on = [{ task = "build-cython-tests" }]
+
+[dependencies]
+cuda-bindings = { path = "." }
+libnvjitlink = "*"
+cuda-nvrtc-dev = "*"
```
leofang: At run time, we just need libnvrtc
 ```suggestion
cuda-nvrtc = "*"
```

File: cuda_bindings/pixi.toml [hunk 6]
```diff
@@ -0,0 +1,125 @@
...
+cuda-profiler-api = "*"
```
leofang: we don't need this at run time (it's header-only, so it's build-time-only)
```suggestion
```

cuda_bindings/pixi.toml
```diff
@@ -0,0 +1,125 @@
...
+cuda-nvvm = "*"
+
+[target.linux.dependencies]
+libcufile-dev = "*"
```
leofang: at run time, we just need libcufile
```suggestion
libcufile = "*"
```

cuda_bindings/pixi.toml
```diff
@@ -0,0 +1,125 @@
...
+cuda-crt-dev_linux-64 = "*"
+
+[feature.cython-tests.target.linux-aarch64.dependencies]
+cuda-crt-dev_linux-aarch64 = "*"
+
+[feature.cython-tests.target.win-64.dependencies]
+cuda-crt-dev_win-64 = "*"
+
+[feature.cu13.system-requirements]
+cuda = "13"
```
leofang, line 53: note to self: this maps to `__cuda >= 13.*`

cuda_bindings/pixi.toml
cpcloud: Did you try running these changes before suggesting them?

cuda_bindings/pixi.toml
```diff
@@ -0,0 +1,125 @@
...
+
+[feature.cu13.dependencies]
+cuda-version = "13.*"
+
+[feature.cu12.system-requirements]
+cuda = "12"
+
+[feature.cu12.dependencies]
+cuda-version = "12.*"
```
leofang: I am a bit confused by `cu12` here and below. `cuda-bindings` in the main branch does not work with CUDA 12, so how does this work ?

cuda_bindings/pixi.toml
```diff
@@ -0,0 +1,125 @@
...
+
+[target.linux-aarch64.activation.env]
+CUDA_HOME = "$CONDA_PREFIX/targets/aarch64-linux"
+
+[target.win.activation.env]
+CUDA_HOME = '%CONDA_PREFIX%\Library'
+
+[package]
+name = "cuda-bindings"
+version = "13.0.2"
```
leofang, line 65: Q: Would it be possible for pixi to read `cuda/bindings/_version.py` and avoid hard-coding?

cuda_bindings/pixi.toml
cpcloud: I only ask because I recall trying the non-dev versions before and they didn't work.

cuda_bindings/pixi.toml
```diff
@@ -0,0 +1,125 @@
...
+
+[package.build]
+backend = { name = "pixi-build-python", version = "*" }
+config.noarch = false
+
+[package.build-dependencies]
+cuda-nvrtc-dev = "*"
+cuda-profiler-api = "*"
+cuda-nvvm = "*"
```
leofang: this is OK because we don't have an easy package `cuda-nvvm-dev` to install (nvvm is considered part of the compiler toolchain, so it is structured differently from regular libraries (with -dev)

cuda_bindings/pixi.toml
cpcloud: I guess it wouldn't, good point. It was a small bit of copypasta from the numba-cuda edition of this stuff. I'll remove it here.

cuda_bindings/pixi.toml
cpcloud: Maybe, but certainly not happening in this PR!

cuda_bindings/pixi.toml
```diff
@@ -0,0 +1,125 @@
...
+
+[package]
+name = "cuda-bindings"
+version = "13.0.2"
+
+[package.build]
+backend = { name = "pixi-build-python", version = "*" }
+config.noarch = false
+
+[package.build-dependencies]
```
leofang: Q: Shouldn't we also add `cuda-cudart-static` here?

cuda_bindings/pixi.toml
```diff
@@ -0,0 +1,125 @@
...
+
+[package.host-dependencies]
+python = "*"
+setuptools = ">=77"
+cython = ">=3.1,<3.2"
+pyclibrary = ">=0.1.7"
+
+[package.run-dependencies]
+python = "*"
+cuda-pathfinder = ">=1.1,<1.2"
```
leofang:
```suggestion
cuda-pathfinder = ">=1.1,<2"
```

cuda_bindings/pixi.toml
cpcloud: https://github.com/prefix-dev/pixi/issues/2946 is related (but it's the `package` table version of it).

cuda_bindings/pixi.toml
cpcloud: I don't know, what purpose would that serve?

leofang: We statically link to cudart so that `cuda.bindings.runtime` can function. I am assuming this passes for you locally because some dependencies brought in `cuda-cudart-static` transitively, otherwise you'd hit an error at link time. It'd be good to be explicit (and in line with our conda recipe):
https://github.com/conda-forge/cuda-python-feedstock/blob/2e8725a3fc1b0f5797f0d7291e2048b3e14ac095/recipe/meta.yaml#L112

cuda_bindings/pixi.toml
leofang: No I haven't tried locally (and I should!), I am reviewing based on my understanding of how our conda CUDA packages work, and extrapolating it to pixi.

cuda_core/pixi.toml
leofang: Yeah I was puzzled why. We have to repeat a lot of things in this toml that are actually coming transitively from cuda-bindings. If this works we would have a much cleaner recipe!

cuda_bindings/pixi.toml
leofang: > No I haven't tried locally (and I should!)

@cpcloud I just tried this and hit an error, pinging you on slack... 🤔

rparolin (commented):
Would love to see some docs added to the README about how to use pixi.

cpcloud:
I will add docs in a follow-up. Let's get this merged!

cuda_bindings/pixi.toml
cpcloud: Done

leofang:
I am sorry Phillip, I did not notice you made changes this morning. I will have to force-push to your branch because there were (and still are) many issues that I just fixed locally 😓 

EDIT: for posterity, see the diff before/after my force push here: https://github.com/NVIDIA/cuda-python/compare/eda7869c153c1a16b59d4c12011e4c2b84f8a6e3..cd60970ad4cabf808d6b57882fb126b3c00bb99c

cuda_bindings/pixi.toml
```diff
@@ -0,0 +1,130 @@
...
+
+[package.target.linux.run-dependencies]
+libcufile = "*"
+
+[target.linux.tasks.build-cython-tests]
+cmd = ["$PIXI_PROJECT_ROOT/tests/cython/build_tests.sh"]
+
+[target.linux.tasks.run-cython-tests]
+cmd = ["pytest", "$PIXI_PROJECT_ROOT/tests/cython/"]
+depends-on = [{ task = "build-cython-tests" }]
```
leofang, line 126: Note: this was missing so the Cython tests were built but not run

cuda_bindings/pixi.toml
```diff
@@ -0,0 +1,130 @@
...
+[package.target.win-64.host-dependencies]
+cuda-crt-dev_win-64 = "*"
+
+[package.run-dependencies]
+python = "*"
+cuda-pathfinder = ">=1.1,<2"
+cuda-bindings = { path = "." }
+libnvjitlink = "*"
+cuda-nvrtc = "*"
+cuda-nvvm = "*"
```
leofang, line 116: I aligned pixi.toml with our feedstock recipe. All run-time dependencies are moved to under `[package.run-dependencies]` except for `cuda-bindings` itself, which cannot be placed here if we want a source build.

cuda_bindings/pixi.toml
```diff
@@ -0,0 +1,130 @@
...
+
+[package.build.target.linux-aarch64.config.env]
+CUDA_HOME = "$PREFIX/targets/sbsa-linux"
+
+[package.build.target.win-64.config.env]
+CUDA_HOME = '%PREFIX%\Library'
+
+# TODO: revisit this once prefix-dev/pixi#4906 is resolved
+# [package.build-dependencies]
+# ccache = "*"
```
leofang: Our build env should be very clean, we just need compilers and they are not declared as `build-dependencies` per the `pixi-build-python` convention (see below/above: https://github.com/NVIDIA/cuda-python/pull/1226#discussion_r2515105327).

cuda_bindings/pixi.toml
```diff
@@ -0,0 +1,130 @@
...
+name = "cuda-bindings"
+version = "13.0.2"
+
+[package.build]
+backend = { name = "pixi-build-python", version = "*" }
+
+[package.build.config]
+env = { CUDA_PYTHON_PARSER_CACHING = "1", CUDA_PYTHON_PARALLEL_LEVEL = "$(nproc)" }
+noarch = false
+compilers = ["c", "cxx"]
```
leofang, line 73: Note: this is a major fix. As per the build backend doc, we need to declare `compilers` here so that they are installed to the build env. All hacks with `CC`/`CXX` are not necessary because they are properly set by the activation script.
https://prefix-dev.github.io/pixi-build-backends/backends/pixi-build-python/#compilers

cuda_bindings/pixi.toml
```diff
@@ -0,0 +1,130 @@
...
+# TODO: check if these can be extracted from pyproject.toml
+[package]
+name = "cuda-bindings"
+version = "13.0.2"
+
+[package.build]
+backend = { name = "pixi-build-python", version = "*" }
+
+[package.build.config]
+env = { CUDA_PYTHON_PARSER_CACHING = "1", CUDA_PYTHON_PARALLEL_LEVEL = "$(nproc)" }
```
leofang, line 71: This makes re-building a lot faster.

cuda_bindings/pixi.toml
```diff
@@ -0,0 +1,130 @@
...
+
+[feature.cu13.system-requirements]
+cuda = "13"
+
+[feature.cu13.dependencies]
+# TODO: make cuda-version a build-variant
+cuda-version = "13.*"
+
+[environments]
+cu13 = { features = ["cu13", "test", "cython-tests"], solve-group = "cu13" }
```
leofang, line 60: I am a bit unhappy having to pass `-e cu13` explicitly for every pixi operation. In the next PR we should think about if this is avoidable.

cuda_bindings/pixi.toml
```diff
@@ -0,0 +1,130 @@
...
+
+[feature.cython-tests.target.win-64.activation.env]
+CUDA_HOME = '%CONDA_PREFIX%\Library'
+
+[feature.cu13.system-requirements]
+cuda = "13"
+
+[feature.cu13.dependencies]
+# TODO: make cuda-version a build-variant
+cuda-version = "13.*"
```
leofang, line 57: Note: revisit this in the next PR. It seems https://github.com/prefix-dev/pixi/issues/4139 was fixed/closed, so hopefully this becomes possible (building variants for 13.0, 13.1, 13.2, ...). But I had just resolved a few local pixi issues and I haven't had time to look into this.

cuda_bindings/pixi.toml
```diff
@@ -0,0 +1,130 @@
...
+
+# For finding headers when building the Cython tests
+[feature.cython-tests.target.linux-64.activation.env]
+CUDA_HOME = "$CONDA_PREFIX/targets/x86_64-linux"
+
+[feature.cython-tests.target.linux-aarch64.activation.env]
+CUDA_HOME = "$CONDA_PREFIX/targets/sbsa-linux"
+
+[feature.cython-tests.target.win-64.activation.env]
+CUDA_HOME = '%CONDA_PREFIX%\Library'
```
leofang, line 50: Note: I moved the `CUDA_HOME` declaration to under the test task, because cuda-bindings is supposed to work without setting it. We need it here only because we need to compile the Cython tests and for that we need to find where the CUDA headers are.

cuda_bindings/pixi.toml
```diff
@@ -0,0 +1,130 @@
...
+cuda-bindings = { path = "." }
+pip = ">=25.3,<26"  # installed to the workspace for convenience
+
+[feature.test.dependencies]
+pytest = ">=6.2.4"
+pytest-benchmark = ">=3.4.1"
+pyglet = ">=2.1.9"
+numpy = "*"
+
+[feature.cython-tests.dependencies]
```
leofang, line 23: Note: aligned with our feedstock recipe

cuda_bindings/pixi.toml
```diff
@@ -0,0 +1,130 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+[workspace]
+channels = ["conda-forge"]
+platforms = ["linux-64", "linux-aarch64", "win-64"]
+preview = ["pixi-build"]
+
+[workspace.build-variants]
+python = ["3.10.*", "3.11.*", "3.12.*", "3.13.*", "3.14.*"]
```
leofang, line 11: I haven't quite figured out how to specify the desired variant. Locally I see with either `pixi install -e cu13 -vv` or `pixi run -e cu13 test` I always get a Python 3.14 env. But this is not blocking this PR from being merged.

Another note is we should add PY314t here if it's possible.

cuda_bindings/pixi.toml
leofang: resolved, I hit two issues
1. my home dir was full and pixi by default has a cache folder under home
2. ccache does not work in my env (with hindsight it's probably related to the system compiler instead of the conda compiler being used, see https://github.com/NVIDIA/cuda-python/pull/1226#discussion_r2515105327). Also it does not seem ccache is compatible with pixi even after the fix?

cuda_bindings/pixi.toml
leofang: confirmed it is working

cuda_core/pixi.toml
leofang: I am now puzzled again, because this actually works (at least for `-e cu13`)! See commit cc835419abf69b0d5dbac0f83179aa09ce4894a2.

cuda_core/pixi.toml
```diff
@@ -0,0 +1,129 @@
...
+
+[target.linux.tasks.build-cython-tests]
+cmd = ["$PIXI_PROJECT_ROOT/tests/cython/build_tests.sh"]
+
+[target.linux.tasks.run-cython-tests]
+cmd = ["pytest", "$PIXI_PROJECT_ROOT/tests/cython/"]
+depends-on = [{ task = "build-cython-tests" }]
+
+[target.linux.tasks.test]
+cmd = ["pytest", "$PIXI_PROJECT_ROOT/tests"]
```
leofang, line 125: note: I am not sure why, but without adding the `tests` subdir to the path I hit an error:
```shell
(...)
tests/cython/test_cython.py .                                                                                                                                      [100%]

=========================================================================== 1 passed in 0.75s ============================================================================
                                                                                                                                                                          
✨ Pixi task (test in cu13): pytest $PIXI_PROJECT_ROOT                                                                                                                    import cupy as cp
from cuda.core.experimental import Device
from cuda.core.experimental import LaunchConfig, launch
from cuda.core.experimental import Program, ProgramOptions

dev = Device(0)
dev.set_current()
code = r"""
extern "C" __global__ void arange(int* a, const int N) {
    const unsigned int tid = threadIdx.x + blockIdx.x * blockDim.x;
    for (auto x=tid; x<N; x+=blockDim.x*gridDim.x) { a[x] = x; }
}
"""
prog = Program(code,
               code_type="c++",
               options=ProgramOptions(std="c++17"))
mod = prog.compile("cubin")
ker = mod.get_kernel("arange")

s = dev.create_stream()
length = 64
data_type = cp.int32
dtype = cp.dtype(data_type)
buf = dev.memory_resource.allocate(length * dtype.itemsize, s) 
config = LaunchConfig(grid=1, block=(32,), stream=s)
launch(ker, config, buf, data_type(length))
s.sync()

INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/local/home/leof/dev/cuda-python/cuda_core/.pixi/envs/cu13/lib/python3.14/site-packages/_pytest/main.py", line 314, in wrap_session
INTERNALERROR>     config._do_configure()
INTERNALERROR>     ~~~~~~~~~~~~~~~~~~~~^^
INTERNALERROR>   File "/local/home/leof/dev/cuda-python/cuda_core/.pixi/envs/cu13/lib/python3.14/site-packages/_pytest/config/__init__.py", line 1128, in _do_configure
INTERNALERROR>     self.hook.pytest_configure.call_historic(kwargs=dict(config=self))
INTERNALERROR>     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/local/home/leof/dev/cuda-python/cuda_core/.pixi/envs/cu13/lib/python3.14/site-packages/pluggy/_hooks.py", line 534, in call_historic
INTERNALERROR>     res = self._hookexec(self.name, self._hookimpls.copy(), kwargs, False)
INTERNALERROR>   File "/local/home/leof/dev/cuda-python/cuda_core/.pixi/envs/cu13/lib/python3.14/site-packages/pluggy/_manager.py", line 120, in _hookexec
INTERNALERROR>     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
INTERNALERROR>            ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/local/home/leof/dev/cuda-python/cuda_core/.pixi/envs/cu13/lib/python3.14/site-packages/pluggy/_callers.py", line 167, in _multicall
INTERNALERROR>     raise exception
INTERNALERROR>   File "/local/home/leof/dev/cuda-python/cuda_core/.pixi/envs/cu13/lib/python3.14/site-packages/pluggy/_callers.py", line 121, in _multicall
INTERNALERROR>     res = hook_impl.function(*args)
INTERNALERROR>   File "/local/home/leof/dev/cuda-python/cuda_core/.pixi/envs/cu13/lib/python3.14/site-packages/_pytest/debugging.py", line 66, in pytest_configure
INTERNALERROR>     import pdb
INTERNALERROR>   File "/local/home/leof/dev/cuda-python/cuda_core/.pixi/envs/cu13/lib/python3.14/pdb.py", line 299, in <module>
INTERNALERROR>     class _PdbInteractiveConsole(code.InteractiveConsole):
INTERNALERROR>                                  ^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR> AttributeError: module 'code' has no attribute 'InteractiveConsole'
```

cuda_core/pixi.toml
```diff
@@ -0,0 +1,129 @@
...
+#env = { PYTHONPATH = "$PIXI_PROJECT_ROOT/../cuda_python_test_helpers" }
```
leofang, line 126: This does not seem necessary...? (I commented it out.)

leofang (approved):
/ok to test cc835419

cuda_bindings/pixi.toml
cpcloud: You can't do this yet. I opened a feature request for this in pixi recently, related to the numba-cuda Pixi work.

cuda_core/pixi.toml
```diff
@@ -0,0 +1,129 @@
...
+# ccache = "*"
+
+[package.host-dependencies]
+python = "*"
+pip = "*"
+setuptools = ">=77"
+cython = ">=3.1"
+cuda-cudart-dev = "*"
+cuda-profiler-api = "*"
+# TODO: figure out why we can't set editable=true here
```
cpcloud: This comment doesn't make sense. Editable installs are a pip only thing and don't map onto any concepts in conda. The package will get rebuilt of its source changes, it's a much simpler mechanism than the way pip works.

cpcloud: Moreover, editable is a property that is only allowed in pypi-dependencies, which isn't a thing at the package build level, only the environment level.

leofang: Ah ha, yes it's what I missed, it does not apply to arbitrary `{build-,host-,run-}dependencies`, only `pypi-dependencies`. Let's remove it.

https://pixi.sh/latest/reference/pixi_manifest/#path

leofang:
```suggestion
```

cuda_core/pixi.toml
```diff
@@ -0,0 +1,129 @@
...
+[package.target.linux-aarch64.host-dependencies]
+cuda-crt-dev_linux-aarch64 = "*"
+
+[package.target.win-64.host-dependencies]
+cuda-crt-dev_win-64 = "*"
+
+[package.run-dependencies]
+python = "*"
+numpy = "*"
+# TODO: figure out why we can't set editable=true here
```
leofang:
```suggestion
```

cuda_bindings/pixi.toml
leofang: > I opened a feature request for this in pixi recently

This one? https://github.com/prefix-dev/pixi/issues/4877

cuda_bindings/pixi.toml
leofang: (The reason that this is a fix instead of an enhancement is because without the fix the system compiler is invoked.)

leofang:
weird, GitHub response is so slow, and the previous CI trigger did nothing

leofang:
/ok to test [b1834b1](https://github.com/NVIDIA/cuda-python/pull/1226/commits/b1834b1b97eadc38061e382f5a62fe34453f17d0)

leofang (approved):

cuda_bindings/pixi.toml
cpcloud: Yep!

cuda_bindings/pixi.toml
```diff
@@ -0,0 +1,130 @@
...
+cuda-nvrtc = "*"
+cuda-nvvm = "*"
+
+[package.target.linux.run-dependencies]
+libcufile = "*"
+
+[target.linux.tasks.build-cython-tests]
+cmd = ["$PIXI_PROJECT_ROOT/tests/cython/build_tests.sh"]
+
+[target.linux.tasks.run-cython-tests]
```
cpcloud, line 124: Is there a point to the complexity of creating a separate task just to run the cython tests and then making that a dependency of the `test` task?

Why not just run `pytest $PIXI_PROJECT_ROOT`? If that doesn't work for some reason (e.g., it can't discover the cython tests) then we should pass it on the command line.

leofang: Right, Cython tests aren't auto-discoverable from the project root. We added a rule to exclude them. Feel free to simplify the task dependency. I only wanted to make sure it runs as intended, and thought adding 3 lines of code was not too bad.

cuda_core/pixi.toml
cpcloud: Running tests from `$PIXI_PROJECT_ROOT` doesn't work for me without this. That's where we should run the tests from.

This was an intentional choice that I made so that `pixi run -e cu13 pytest` "just works".

PR #1229, opened 2025-11-10
[BUG] Remove the check disabling windows support, and replace by checking for CUDA VMM API support

benhg:
## Description
When we initially wrote this feature, I neglected to disable the functionality if VMM APIs are not supported, and instead disabled all Windows support because TCC Windows systems don't have this support. Instead, we should check for VMM APIs explicitly.

### DIFF
--- a/cuda_core/cuda/core/experimental/_memory.pyx
+++ b/cuda_core/cuda/core/experimental/_memory.pyx
@@ -1206,14 +1206,15 @@ class VirtualMemoryResource(MemoryResource):
         self.config = check_or_create_options(
             VirtualMemoryResourceOptions, config, "VirtualMemoryResource options", keep_none=False
         )
-        if self.config.location_type == "host":
+        # Matches ("host", "host_numa", "host_numa_current")
+        if "host" in self.config.location_type:
             self.device = None
 
-        if device and not device.properties.virtual_memory_management_supported:
-            raise RuntimeError("VirtualMemoryResource requires CUDA VMM API support")
+        if not self.device and self.config.location_type == "device":
+            raise RuntimeError("VirtualMemoryResource requires a device for device memory allocations")
 
-        if not device and self.config.location_type == "device":
-            raise ValueError("VirtualMemoryResource requires a device for device memory allocations")
+        if self.device and not self.device.properties.virtual_memory_management_supported:
+            raise RuntimeError("VirtualMemoryResource requires CUDA VMM API support")
 
         # Validate RDMA support if requested
         if self.config.gpu_direct_rdma and self.device is not None:


cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -1208,8 +1208,12 @@ class VirtualMemoryResource(MemoryResource):
         )
         if self.config.location_type == "host":
             self.device = None
-        if platform.system() == "Windows":
-            raise NotImplementedError("VirtualMemoryResource is not supported on Windows")
+
+        if device and not device.properties.virtual_memory_management_supported:
+            raise NotImplementedError("VirtualMemoryResource requires CUDA VMM API support")
+
+        if not device and self.config.location_type == "device":
```
leofang: Q: At this point shouldn't we just check `self.device`?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -1208,8 +1208,12 @@ class VirtualMemoryResource(MemoryResource):
         )
         if self.config.location_type == "host":
             self.device = None
-        if platform.system() == "Windows":
-            raise NotImplementedError("VirtualMemoryResource is not supported on Windows")
+
+        if device and not device.properties.virtual_memory_management_supported:
```
leofang: use `self.device`?

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -1208,8 +1208,12 @@ class VirtualMemoryResource(MemoryResource):
...
+            raise NotImplementedError("VirtualMemoryResource requires CUDA VMM API support")
```
```suggestion
            raise RuntimeError("VirtualMemoryResource requires CUDA VMM API support")
```

cuda_core/cuda/core/experimental/_memory.pyx
```diff
@@ -1208,8 +1208,12 @@ class VirtualMemoryResource(MemoryResource):
         )
         if self.config.location_type == "host":
             self.device = None
-        if platform.system() == "Windows":
-            raise NotImplementedError("VirtualMemoryResource is not supported on Windows")
```
leofang, line 1212: Discussed with Ben offline, this is a follow-up of https://github.com/NVIDIA/cuda-python/pull/1179. This check should have been removed.

cuda_core/cuda/core/experimental/_memory.pyx
benhg: We can split the checks. Basically self.device will be None for host allocations.

cuda_core/cuda/core/experimental/_memory.pyx
benhg: SG

leofang:
/ok to test [b5be26d](https://github.com/NVIDIA/cuda-python/pull/1229/commits/b5be26dcec74488d54a8662a14c8d431af7f93e8)

Issue #634, opened 2025-05-13
Ensure `alignment` can be passed to `MemoryResource.allocate()`


leofang:
Unlike in C++ I think it'd have to be an optional keyword argument, for a few reasons
- To preserve backward compatibility (of both `cuda.core` and RMM Python)
- To improve UX (passing an explicit alignment is less common use case, especially if the underlying MR does not support any custom alignment)

PR #1235, opened 2025-11-12
Implements GraphMemoryResource

### LINKED ISSUE #963: CUDA graph phase 2 - memory nodes

Andy-Jost:
## Description

Implements `GraphMemoryResource` for memory interactions with the graph memory allocator. Allocations from this object succeed only when graph capturing is active. Conversely, allocations from `DeviceMemoryResource` now raise an exception when graph capturing is active.

A new test module is added.

This change also simplifies and extends the logic for accepting arbitrary stream parameters as objects implementing `__cuda_stream__`. Support for that protocol was added in several places, allowing `GraphBuilder` to be used anywhere a stream is expected, including memory resource and buffer methods.

closes #963

### DIFF
--- a/cuda_core/tests/test_graph_mem.py
+++ b/cuda_core/tests/test_graph_mem.py
@@ -89,10 +89,15 @@ def test_graph_alloc(mempool_device, mode):
     set_zero = mod.get_kernel("set_zero")
     add_one = mod.get_kernel("add_one")
 
+    # Initialize out to zero.
+    config = LaunchConfig(grid=1, block=1)
+    launch(stream, config, set_zero, out, NBYTES)
+
+    # Increments out by 3
     def apply_kernels(mr, stream, out):
         buffer = mr.allocate(NBYTES, stream=stream)
-        config = LaunchConfig(grid=1, block=1)
-        for kernel in [set_zero, add_one, add_one]:
+        buffer.copy_from(out, stream=stream)
+        for kernel in [add_one, add_one, add_one]:
             launch(stream, config, kernel, buffer, NBYTES)
         out.copy_from(buffer, stream=stream)
         buffer.close()
@@ -101,18 +106,25 @@ def test_graph_alloc(mempool_device, mode):
     if mode == "no_graph":
         # Do work without graph capture.
         apply_kernels(mr=dmr, stream=stream, out=out)
+        stream.sync()
+        assert compare_buffer_to_constant(out, 3)
     else:
         # Capture work, then upload and launch.
         gb = device.create_graph_builder().begin_building(mode)
         apply_kernels(mr=gmr, stream=gb, out=out)
         graph = gb.end_building().complete()
+
+        # First launch.
         graph.upload(stream)
         graph.launch(stream)
+        stream.sync()
+        assert compare_buffer_to_constant(out, 3)
 
-    stream.sync()
-
-    # Check the result on the host.
-    assert compare_buffer_to_constant(out, 2)
+        # Second launch.
+        graph.upload(stream)
+        graph.launch(stream)
+        stream.sync()
+        assert compare_buffer_to_constant(out, 6)
 
 
 @pytest.mark.skipif(IS_WINDOWS or IS_WSL, reason="auto_free_on_launch not supported on Windows")


cuda_core/cuda/core/experimental/_memory/_graph_memory_resource.pxd
```diff
@@ -0,0 +1,13 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from cuda.bindings cimport cydriver
+from cuda.core.experimental._memory._buffer cimport MemoryResource
+from cuda.core.experimental._memory._device_memory_resource cimport DeviceMemoryResource
```
Andy-Jost: I'll remove the cruft here.

cuda_core/cuda/core/experimental/_memory/_graph_memory_resource.pyx
```diff
@@ -0,0 +1,207 @@
...
+        )
+
+        def fget(GraphMemoryResourceAttributes self) -> property_type:
+            value = GMRA_getattribute(self._dev_id, <cydriver.CUgraphMem_attribute><uintptr_t> attr_enum)
+            return property_type(value)
+
+        if _settable:
+            def fset(GraphMemoryResourceAttributes self, uint64_t value):
+                if value != 0:
+                    raise AttributeError(f"Attribute {stub.__name__!r} may only be set to zero (got {value}).")
```
Andy-Jost: The driver checks for this condition in `cuDeviceSetGraphMemAttribute` and issues a log message: "High watermark can only be reset to 0"

It's a shame we cannot access that message programmatically for use in the Python error.

cuda_core/cuda/core/experimental/_device.pyx
```diff
@@ -1295,7 +1295,7 @@ class Device:
         ctx = self._get_current_context()
         return Event._init(self._id, ctx, options, True)
 
-    def allocate(self, size, stream: Stream | None = None) -> Buffer:
```
Andy-Jost, line 1309: I updated `Stream` arguments to `IsStreamT` throughout.

cuda_core/cuda/core/experimental/_launcher.pyx
```diff
@@ -58,17 +52,7 @@ def launch(stream: Union[Stream, IsStreamT], config: LaunchConfig, kernel: Kerne
...
-    try:
-        stream_handle = stream.handle
-    except AttributeError:
-        try:
-            stream_handle = driver.CUstream(<uintptr_t>(_try_to_get_stream_ptr(stream)))
-        except Exception:
-            raise ValueError(
-                f"stream must either be a Stream object or support __cuda_stream__ (got {type(stream)})"
-            ) from None
+    stream = Stream._init(stream)
```
Andy-Jost: The canonical way to invoke the `__cuda_stream__` protocol now is to call `Stream._init`. It will either succeed in creating a `Stream` object or raise an exception.

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -62,36 +62,6 @@ class IsStreamT(Protocol):
         ...
 
 
-cdef cydriver.CUstream _try_to_get_stream_ptr(obj: IsStreamT) except*:
```
Andy-Jost, line 65: This code block was moved below without changes.

cuda_core/tests/test_graph_mem.py
```diff
@@ -0,0 +1,254 @@
...
+
+    # Capturing
+    gb = device.create_graph_builder().begin_building(mode=mode)
+    with pytest.raises(
+        RuntimeError,
+        match=r"DeviceMemoryResource cannot perform memory operations on a capturing "
+        r"stream \(consider using GraphMemoryResource\)\.",
+    ):
+        dmr.allocate(1, stream=gb)
+    gb.end_building().complete()
```
Andy-Jost, line 267: This section illustrates a drawback of not using `with` contexts. Ignore the fact that the error is caught here (that's just for testing). If an exception is thrown during graph capture, control can easily escape without making a call to `gb.end_building`. That leaves the surrounding code in an unexpected state (capturing on).

cuda_core/cuda/core/experimental/_launcher.pyx
```diff
@@ -58,17 +52,7 @@ def launch(stream: Union[Stream, IsStreamT], config: LaunchConfig, kernel: Kerne
         launching kernel.
 
     """
-    if stream is None:
-        raise ValueError("stream cannot be None, stream must either be a Stream object or support __cuda_stream__")
-    try:
-        stream_handle = stream.handle
```
Andy-Jost, line 64: This appears to be a (now fixed) bug. Any object with a handle attribute would use that preferentially over the stream protocol, even if the handle was not for a stream.

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
```diff
@@ -127,13 +128,12 @@ cdef class Buffer:
         ----------
         dst : :obj:`~_memory.Buffer`
             Source buffer to copy data from
-        stream : Stream
+        stream : IsStreamT
```
rparolin: IsStreamT is still a Stream type?

cuda_core/cuda/core/experimental/_memory/_device_memory_resource.pyx
```diff
@@ -302,14 +308,14 @@ cdef class DeviceMemoryResource(MemoryResource):
             raise RuntimeError("Imported memory resource cannot be exported")
         return self._ipc_data._alloc_handle
 
-    def allocate(self, size_t size, stream: Stream = None) -> Buffer:
+    def allocate(self, size_t size, stream: Optional[IsStreamT] = None) -> Buffer:
```
rparolin: Does this allocate member function need to take an alignment parameter?

cuda_core/cuda/core/experimental/_memory/_device_memory_resource.pyx
```diff
@@ -334,15 +339,17 @@ cdef class DeviceMemoryResource(MemoryResource):
...
         size : int
             The size of the buffer to deallocate, in bytes.
-        stream : Stream, optional
+        stream : IsStreamT, optional
             The stream on which to perform the deallocation asynchronously.
             If the buffer is deallocated without an explicit stream, the allocation stream
             is used.
         """
-        DMR_deallocate(self, <uintptr_t>ptr, size, <Stream>stream)
+        stream = Stream._init(stream) if stream is not None else default_stream()
```
rparolin: Is there a way to capture the user error of neglecting to pass the stream that memory allocation came from?  I'm thinking of a debug assert that verify if the allocated memory address came from the provided stream?  Another potential option, is if the user neglects providing a stream, we do a look to determine where the address came from. Not sure if that is possible given the current lower level API.

cuda_core/cuda/core/experimental/_memory/_device_memory_resource.pyx
```diff
@@ -460,10 +467,21 @@ cdef void DMR_init_create(
...
+# A result of CU_STREAM_CAPTURE_STATUS_INVALIDATED is considered an error.
+cdef inline int check_not_capturing(cydriver.CUstream s) except?-1 nogil:
+    cdef cydriver.CUstreamCaptureStatus capturing
+    HANDLE_RETURN(cydriver.cuStreamIsCapturing(s, &capturing))
+    if capturing != cydriver.CUstreamCaptureStatus.CU_STREAM_CAPTURE_STATUS_NONE:
+        raise RuntimeError("DeviceMemoryResource cannot perform memory operations on "
+                           "a capturing stream (consider using GraphMemoryResource).")
+
+
+cdef inline Buffer DMR_allocate(DeviceMemoryResource self, size_t size, Stream stream):
```
rparolin, line 487: What does inline do here?

cuda_core/cuda/core/experimental/_memory/_graph_memory_resource.pyx
```diff
@@ -0,0 +1,207 @@
...
+    capturing is enabled results in a runtime error.
+
+    Parameters
+    ----------
+    device_id : int | Device
+        Device or Device ordinal for which a graph memory resource is obtained.
+    """
+
+    def __new__(cls, device_id: int | Device):
+        cdef int c_device_id = getattr(device_id, 'device_id', device_id)
```
rparolin, line 169: This is where I wish we had function overloading in the language..

cuda_core/cuda/core/experimental/_device.pyx
rparolin: I was confused by this change.  Why did we do that?

rparolin:
Comment: I'd recommend in future chunking  large PRs into smaller reviewable chunks.  IMHO, it makes it more approachable and consumable for reviewers.

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
Andy-Jost: `IsStreamT` identifies any type that supports conversion to `Stream` via `__cuda_stream__`, including `Stream` itself.

cuda_core/cuda/core/experimental/_memory/_device_memory_resource.pyx
Andy-Jost: Since the stream only defines an ordering, it is legal to allocate on one stream and deallocate on another.

That said, I think the overall memory management scheme could be clarified and potentially improved.

cuda_core/cuda/core/experimental/_memory/_device_memory_resource.pyx
Andy-Jost: Cython will mark the function `CYTHON_INLINE` in the generated C++.

```
#ifndef CYTHON_INLINE
  #if defined(__clang__)
    #define CYTHON_INLINE __inline__ __attribute__ ((__unused__))
  #else
    #define CYTHON_INLINE inline
  #endif
#endif
```

cuda_core/cuda/core/experimental/_memory/_graph_memory_resource.pyx
Andy-Jost: Yes. It can also be done with decorators and/or pybind11. The main problem with general overloading for us is that the signature matching is done at runtime in Python and we are very sensitive to performance.

In many simple cases, arguments are orthogonal and we just want to reduce a set of types to a certain type. That's what we have here (`{Device, int} -> Device`) and elsewhere in this change with stream conversions (`IsStreamT -> Stream`). For these situations, I think the best solution is to make a standard conversion for each type: to convert "any supported object," `s`, to a `Stream`, use `Stream._init(s)`. Similarly, we should probably update code like this to `Device._init(d)` or `Device(d)`. We could also consider adding `IsDeviceT` and `__cuda_device__` for symmetry.

(I would actually prefer to remove the `_init` methods in favor of constructors, but I understand why we have those and it's a different discussion, anyway.)

cuda_core/cuda/core/experimental/_device.pyx
Andy-Jost: The `GraphBuilder` object contains a stream and can be used in place of streams. In `test_graph.py` you can find lots of statements like this `launch(gb, LaunchConfig(grid=1, block=1), empty_kernel)`.

This was only partly implemented. For instance, `launch` contained logic to convert arguments to streams, but nothing in the memory module did.

This change extends the support. In general, we should uniformly use `IsStreamT` for any API function that accepts a stream-like argument and then call `Stream._init(arg)` to convert it.

cuda_core/cuda/core/experimental/_memory/_graph_memory_resource.pyx
Andy-Jost: Good news: CUDA 13 adds functions for error log management. It looks like cuLogsRegisterCallback might help here.

cuda_core/cuda/core/experimental/_device.pyx
```diff
@@ -1242,7 +1242,7 @@ class Device:
         """
         raise NotImplementedError("WIP: https://github.com/NVIDIA/cuda-python/issues/189")
 
-    def create_stream(self, obj: IsStreamT | None = None, options: StreamOptions | None = None) -> Stream:
+    def create_stream(self, obj: Optional[IsStreamT] = None, options: StreamOptions | None = None) -> Stream:
```
leofang: Q: We seem to be introducing inconsistent syntax preference here, any reason for this change? The old typing and the new one are equivalent. Also, we still keep the typing for `options` the same. 

I think if ruff were able to lint Cython code, this would have been flagged.

leofang: IIRC in "modern" (in relative terms, since Python typing changes so rapidly) typing using `|` is preferred over `Optional`. We should get a typing expert to help review (I am not one) 🙂

cuda_core/cuda/core/experimental/_device.pyx
```diff
@@ -1295,7 +1295,7 @@ class Device:
         ctx = self._get_current_context()
         return Event._init(self._id, ctx, options, True)
 
-    def allocate(self, size, stream: Stream | None = None) -> Buffer:
+    def allocate(self, size, stream: Optional[IsStreamT] = None) -> Buffer:
```
leofang: ditto
```suggestion
    def allocate(self, size, stream: IsStreamT | None = None) -> Buffer:
```

cuda_core/cuda/core/experimental/_device.pyx
leofang: Sorry, I recalled this past discussion https://github.com/NVIDIA/cuda-python/pull/693#discussion_r2136758919 and we should not support `IsStreamT` here. We should tighten it up to `stream: Stream | GraphBuilder = None` instead.

cuda_core/cuda/core/experimental/_device.pyx
leofang: nvm, see my comment above.

cuda_core/cuda/core/experimental/_memory/_device_memory_resource.pyx
leofang: Great question, @rparolin why do you ask?

The answer is not yet: https://github.com/NVIDIA/cuda-python/issues/634.

cuda_core/cuda/core/experimental/_launcher.pyx
leofang: Which part was a bug?

Andy-Jost: If an object provided both `handle` and `__cuda_stream__` the handle would be used. So, say for the sake of argument the `GraphBuilder` provided a `handle` attribute, this would break.

cuda_core/cuda/core/experimental/_device.pyx
Andy-Jost: Thanks for the context, Leo. It does seem that `__cuda_stream__` is the wrong thing to use here. I agree, let's reserve it for conversions from external libraries.

I'm interested in how we convert our own objects internally. We have objects like `Graph` that should be usable anywhere a stream can be used. While we can update signatures to accept `Stream | Graph`, I'm worried about the possible maintenance implications. I'd like to find a solution that "closes" this section of code to future modifications.

How about this: we add a property such as `__cudacore_stream__` that returns a `Stream` object and teach the `Stream._init` method to respect it. Then, any public function that accepts a stream-like argument can call `Stream._init(arg)` to perform the conversion (or raise a type error). There's no significant overhead, since it's just a property access, and we can keep this as an internal implementation detail.

A possible alternative would be to name the property `stream`. We already have `Graph.stream`, and we could add `Stream.stream` returning `self`.

cuda_core/cuda/core/experimental/_device.pyx
Andy-Jost: Yeah, it's a mess. There was at least one place where the lint tool changed a `|` to an `Optional` but this doesn't seem to be it. I'll change this back and in a separate PR I can go through to update everything to the new style. This will only deteriorate if we don't make it consistent.

cuda_core/tests/test_graph_mem.py
leofang: General policy aside, I don't think for this particular case a `with` context makes sense? During capturing once an error is raised (see the crafted example below) we cannot successfully end capturing.
```python
>>> from cuda.bindings import runtime
>>> runtime.cudaSetDevice(0)
(<cudaError_t.cudaSuccess: 0>,)
>>> _, s = runtime.cudaStreamCreate()
>>> s
<CUstream 0x559fca4376f0>
>>> # capturing in the global mode to make a silly but concise example 
>>> runtime.cudaStreamBeginCapture(s, runtime.cudaStreamCaptureMode.cudaStreamCaptureModeGlobal)
(<cudaError_t.cudaSuccess: 0>,)
>>> runtime.cudaMalloc(1024)  # not allowed in the global mode
(<cudaError_t.cudaErrorStreamCaptureUnsupported: 900>, None)
>>> runtime.cudaStreamEndCapture(s)  # cannot end capture
(<cudaError_t.cudaErrorStreamCaptureInvalidated: 901>, None)
```

leofang: In other words, once an error is raised, the driver switches from capturing on to capturing off, so this statement
> That leaves the surrounding code in an unexpected state (capturing on).

does not hold.

cuda_core/cuda/core/experimental/_launcher.pyx
leofang: interesting, thanks!

cuda_core/cuda/core/experimental/_device.pyx
leofang: @Andy-Jost I'll let you summarize our offline discussion if it's OK 🙂 I will try to complete my review before I go to bed.

Andy-Jost: Sure thing. We only need to support types `Stream | GraphBuilder`, which should suffice for the foreseeable future. I could add a helper type definition and conversion function to make this easier to support.

This approach matches what CCCL does in `launch.cuh`:

```
_GraphInserter&& __forward_or_cast_to_stream_ref(_GraphInserter&& __inserter);
::cuda::stream_ref __forward_or_cast_to_stream_ref(::cuda::stream_ref __stream)
```

For the `launch` API that accepts objects with `__cuda_stream__`, we'll continue to accept `IsStreamT` but raise a deprecation warning if such an object is passed. We will tighten that up in the next minor release.

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -445,3 +415,65 @@ cpdef Stream default_stream():
...
+        else:
+            warnings.warn(
+                "Passing foreign stream objects to this function via the "
+                "stream protocol is deprecated. Convert the object explicitly "
+                "using Stream(obj) instead.",
+                stacklevel=2,
+                category=DeprecationWarning,
+            )
+            return <Stream> stream
+    raise TypeError(f"Stream or GraphBuilder expected, got {type(arg).__name__}")
```
Andy-Jost: Here's the routine for accepting stream-like arguments.

cuda_core/tests/test_graph_mem.py
Andy-Jost: The error could be any Python error, not necessarily a CUDA driver error. What if you replaced `runtime.cudaMalloc` with something like `len(1)`? Then the stream capture would never be turned off.

It seems to me that graphs ought to have a context manager and the exit should try to end capture and silently tolerate CUDA_ERROR_STREAM_CAPTURE_INVALIDATED.

Example:
```
gb = device.create_graph_builder().begin_building()
len(1) # raises, as an example
gb.end_building().complete()
```

versus

```
with device.create_graph_builder().begin_building() as gb:
    len(1)
```

Only the second one guarantees that capture is ended.

leofang: No it still doesn't matter, because when any error happens (CUDA or Python) a Python exception is bubbling up and `gb` and its internal stream get deallocated nicely.

This `with` example can be re-written as 
```python
gb = device.create_graph_builder().begin_building()
try:
    # do work with gb
except Exception:
    try:
        # I do not believe this is needed given the earlier cudart example. This is 
        # only to illustrate that the with statement has an equivalent way to express. 
        #
        # Regardless if it is a Python or CUDA error, the gb is left in an undefined 
        # state that we should not even attempt to build a graph from. Just let it go
        # out of scope and let destructors kick in.
        gb.end_building().complete()
    except Exception:
        pass
else:
    # gb is ready for use
```
so there is always a way 🙂 It is only a syntactic sugar.

Forgive me for being repetitive. As a design guideline, under no circumstances in which context managers are a must-have in the design. They can be easily implemented on top of methods like `.begin()` and `.end()`, but not the other way around and so in any case we need to implement the latter anyway, meaning we can always add context managers later as a syntactic sugar for the latter. The Python language and standard library has shown us a clear path. So let us punt on this for a little longer. 
 
For the particular case of stream capturing, the stream can be destroyed just fine even before capturing ends, so the destructor will work as intended.
```python
>>> from cuda.bindings import runtime
>>> runtime.cudaSetDevice(0)
(<cudaError_t.cudaSuccess: 0>,)
>>> _, s = runtime.cudaStreamCreate()
>>> runtime.cudaStreamBeginCapture(s, runtime.cudaStreamCaptureMode.cudaStreamCaptureModeGlobal)
(<cudaError_t.cudaSuccess: 0>,)
>>> runtime.cudaStreamDestroy(s)  # destroy stream without ending the capturing state
(<cudaError_t.cudaSuccess: 0>,)
```

Andy-Jost: I agree that we should punt: let's take this discussion offline and follow up at our leisure.

Let me add one observation, which is what motivated the original comment. If I insert a Python error right before line 254, all subsequent tests fail with `ValueError: device_id must be within [0, 0), got 0`. Here's a sampling:

```
test_graph_mem.py .............FEE                                           [ 64%]
test_hashable.py EEEEEE                                                      [ 88%]
test_helpers.py FFF                                                          [100%]
```

This is apparently due to `GraphBuilder` having no `__del__` method.

cuda_core/cuda/core/experimental/_stream.pyx
Andy-Jost: Latest update removes `allow_default` and `default_value`. Default arguments can be handled more easily by the caller.

cuda_core/tests/test_graph_mem.py
leofang: > This is apparently due to `GraphBuilder` having no `__del__` method.

Yeah we should fix this asap. Hopefully once we have your proposal implemented this class of errors will go away systematically 🙂 

Before that happens, I wonder if there is a way we can enforce the linter to check if `__del__` is missing 🤔

cuda_core/cuda/core/experimental/_memory/_device_memory_resource.pyx
```diff
@@ -481,16 +499,19 @@ cdef Buffer DMR_allocate(DeviceMemoryResource self, size_t size, Stream stream):
...
+cdef inline void DMR_deallocate(
     DeviceMemoryResource self, uintptr_t ptr, size_t size, Stream stream
 ) noexcept:
     cdef cydriver.CUstream s = stream._handle
     cdef cydriver.CUdeviceptr devptr = <cydriver.CUdeviceptr>ptr
+    cdef cydriver.CUresult r
     with nogil:
-        HANDLE_RETURN(cydriver.cuMemFreeAsync(devptr, s))
+        r = cydriver.cuMemFreeAsync(devptr, s)
+        if r != cydriver.CUDA_ERROR_INVALID_CONTEXT:
```
leofang, line 510: Q: mempools are not tied to a CUDA context, so when will this happen?

cuda_core/cuda/core/experimental/_memory/_graph_memory_resource.pyx
```diff
@@ -0,0 +1,210 @@
...
+        """Return False. This memory resource does not provide host-accessible buffers."""
+        return False
+
+
+class GraphMemoryResource(cyGraphMemoryResource):
+    """
+    A memory resource managing the graph-specific memory pool.
+
+    Graph-captured memory operations use a special internal memory pool, which
+    is a per-device singleton. This class serves as the interface to that pool.
```
leofang: Q: Do we want to document this (driver) implementation detail?

cuda_core/cuda/core/experimental/_memory/_graph_memory_resource.pyx
```diff
@@ -0,0 +1,210 @@
...
+    The only supported operations are allocation, deallocation, and a limited
+    set of status queries.
+
+    This memory resource should be used to allocate memory when graph capturing
+    is enabled. Using this when graphs are not being captured will result in a
+    runtime error.
```
leofang: nit: "Building" instead of "capturing" a graph? We may want to support explicit graph construction in the future.

cuda_core/cuda/core/experimental/_memory/_virtual_memory_resource.py
```diff
@@ -460,7 +465,7 @@ def _build_access_descriptors(self, prop: driver.CUmemAllocationProp) -> list:
 
         return descs
 
-    def allocate(self, size: int, stream: Stream = None) -> Buffer:
+    def allocate(self, size: int, stream: Stream | GraphBuilder | None = None) -> Buffer:
```
leofang: Q: I don't think VMM is capturable ? Does this actually work?

cuda_core/cuda/core/experimental/_memory/_virtual_memory_resource.py
```diff
@@ -544,7 +549,7 @@ def allocate(self, size: int, stream: Stream = None) -> Buffer:
         buf = Buffer.from_handle(ptr=ptr, size=aligned_size, mr=self)
         return buf
 
-    def deallocate(self, ptr: int, size: int, stream: Stream = None) -> None:
+    def deallocate(self, ptr: int, size: int, stream: Stream | GraphBuilder | None = None) -> None:
```
leofang: ditto

cuda_core/cuda/core/experimental/_stream.pyx
```diff
@@ -445,3 +415,59 @@ cpdef Stream default_stream():
...
+        )
+    return <cydriver.CUstream><uintptr_t>(info[1])
+
+# Helper for API functions that accept either Stream or GraphBuilder. Performs
+# needed checks and returns the relevant stream.
+cdef Stream Stream_accept(arg, bint allow_stream_protocol=False):
+    if isinstance(arg, Stream):
+        return <Stream> arg
+    elif isinstance(arg, GraphBuilder):
+        return <Stream> arg.stream
```
leofang: nit
```suggestion
        return <Stream>(arg.stream)
```

cuda_core/tests/helpers/misc.py
leofang, line 1: 👍

cuda_core/tests/test_graph_mem.py
```diff
@@ -0,0 +1,256 @@
...
+    __global__ void add_one(char *a, size_t nbytes) {
+        size_t idx = blockIdx.x * blockDim.x + threadIdx.x;
+        size_t stride = blockDim.x * gridDim.x;
+        for (size_t i = idx; i < nbytes; i += stride) {
+            a[i] += 1;
+        }
+    }
+    """
+    arch = "".join(f"{i}" for i in Device().compute_capability)
+    program_options = ProgramOptions(std="c++17", arch=f"sm_{arch}")
```
leofang:
```suggestion
    program_options = ProgramOptions(std="c++17", arch=f"sm_{Device().arch}")
```

cuda_core/tests/test_graph_mem.py
leofang, line 1: Tests look good! Not sure if I missed it, can we add one more test for launching multiple times the same graph with alloc/dealloc nodes?

cuda_core/cuda/core/experimental/_memory/_device_memory_resource.pyx
Andy-Jost: This is a reactive fix as I was seeing many of these errors at the end of a full pytest run. I think this issue exists prior to this PR, so I could separate this if you prefer.

cuda_core/cuda/core/experimental/_memory/_graph_memory_resource.pyx
Andy-Jost: The driver documentation says memory "related to graphs" and "for use with graphs." I can update this to use similar language.

cuda_core/cuda/core/experimental/_memory/_virtual_memory_resource.py
Andy-Jost: Good catch, let me remove this.

cuda_core/tests/test_graph_mem.py
```diff
@@ -0,0 +1,256 @@
...
+    graph.launch(stream)
+    stream.sync()
+    assert compare_buffer_to_constant(out, 1)
+
+    # Update the input buffer and rerun the graph.
+    set_buffer(in_, 5)
+    graph.upload(stream)
+    graph.launch(stream)
+    stream.sync()
+    assert compare_buffer_to_constant(out, 6)
```
Andy-Jost, line 167: Graph re-launch

cuda_core/tests/test_graph_mem.py
Andy-Jost: Is this what you have in mind? https://github.com/NVIDIA/cuda-python/pull/1235/files#r2551298378

cuda_core/tests/test_graph_mem.py
```diff
@@ -0,0 +1,267 @@
...
+        graph.upload(stream)
+        graph.launch(stream)
+        stream.sync()
+        assert compare_buffer_to_constant(out, 3)
+
+        # Second launch.
+        graph.upload(stream)
+        graph.launch(stream)
+        stream.sync()
+        assert compare_buffer_to_constant(out, 6)
```
Andy-Jost, line 127: Graph re-launch (with dealloc in the graph).

cuda_core/tests/test_graph_mem.py
Andy-Jost: Here's a second one: https://github.com/NVIDIA/cuda-python/pull/1235/files#r2551353767

leofang: Yes, many thanks!

cuda_core/cuda/core/experimental/_memory/_device_memory_resource.pyx
leofang: @Andy-Jost Let's save this to a separate PR. I think this one is clean, and I'd like to merge it asap (after I read it one more time).

cuda_core/cuda/core/experimental/_memory/_graph_memory_resource.pyx
```diff
@@ -0,0 +1,207 @@
...
+from functools import cache
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from cuda.core.experimental._memory.buffer import DevicePointerT
+
+__all__ = ['GraphMemoryResource']
+
+
+cdef class GraphMemoryResourceAttributes:
```
leofang, line 23: @pciolkosz this is yet another reason why CUDA graphs should support mempools... We have to use another class to wrap the attribute access, instead of reusing `DeviceMemoryResourceAttributes`, just because from the driver perspective they are different... 😢 

@Andy-Jost do you think apart from making the code base dirty the two attribute classes can be merged into one, and we dispatch to the right driver APIs internally?

cuda_core/tests/test_launcher.py
```diff
@@ -140,6 +141,14 @@ def test_launch_invalid_values(init_cuda):
     with pytest.raises(TypeError):
         launch(stream, ker, None)
 
+    msg = (
+        r"Passing foreign stream objects to this function via the stream "
+        r"protocol is deprecated\. Convert the object explicitly using "
+        r"Stream\(obj\) instead\."
+    )
+    with pytest.warns(DeprecationWarning, match=msg):
+        launch(StreamWrapper(stream), config, ker)
```
leofang, line 150: I missed this test! Thanks! 🙏

cuda_core/tests/test_memory.py
```diff
@@ -288,13 +279,18 @@ def test_buffer_dunder_dlpack_device_failure():
...
     This test verifies that the DeviceMemoryResource initializes properly,
     including the release threshold configuration for performance optimization.
     """
-    device = mempool_device
+    device = Device()
+
+    if not device.properties.memory_pools_supported:
+        pytest.skip("Device does not support mempool operations")
+
+    device.set_current()
```
leofang, line 293: Q: Any reason we don't wanna move this logic (check attribute first, and then either skip or set current) in the `mempool_device` fixture, and then use the fixture here?

cuda_core/tests/test_memory.py
```diff
@@ -519,11 +515,16 @@ def test_vmm_allocator_rdma_unsupported_exception():
...
 
-def test_mempool(mempool_device):
-    device = mempool_device
+def test_device_memory_resource():
+    device = Device()
+
+    if not device.properties.memory_pools_supported:
+        pytest.skip("Device does not support mempool operations")
+
+    device.set_current()
```
leofang, line 524: ditto

PR #1238, opened 2025-11-14
Standardize use of Device arguments

Andy-Jost:
## Description

Standardizes the way device arguments are accepted in public APIs. Previously, some functions could accept either a `Device` or `int` representing a device ordinal, whereas other would accept only an ordinal. This change updates everything to accept either.

Also, rather than use `getattr` to convert argument, this change uses `Device(arg)` to convert any supported argument to a device (or raise an exception), matching the way `Stream` arguments are converted.

### DIFF
--- a/cuda_core/cuda/core/experimental/_device.pyx
+++ b/cuda_core/cuda/core/experimental/_device.pyx
@@ -948,9 +948,16 @@ class Device:
         Default value of `None` return the currently used device.
 
     """
-    __slots__ = ("_id", "_mr", "_has_inited", "_properties", "_uuid")
+    __slots__ = ("_id", "_memory_resource", "_has_inited", "_properties", "_uuid")
 
-    def __new__(cls, device_id: int | None = None):
+    def __new__(cls, device_id: Device | int | None = None):
+        # Handle device_id argument.
+        if isinstance(device_id, Device):
+            return device_id
+        else:
+            device_id = getattr(device_id, 'device_id', device_id)
+
+        # Initialize CUDA.
         global _is_cuInit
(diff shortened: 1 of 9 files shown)


cuda_core/cuda/core/experimental/_device.pyx
```diff
@@ -948,9 +948,16 @@ class Device:
         Default value of `None` return the currently used device.
 
     """
-    __slots__ = ("_id", "_mr", "_has_inited", "_properties", "_uuid")
+    __slots__ = ("_id", "_memory_resource", "_has_inited", "_properties", "_uuid")
```
Andy-Jost, line 951: I renamed this to match a request in a recent change to _memory/_buffer.*

cuda_core/cuda/core/experimental/_device.pyx
```diff
@@ -986,21 +993,7 @@ class Device:
...
-                            &attr, cydriver.CUdevice_attribute.CU_DEVICE_ATTRIBUTE_MEMORY_POOLS_SUPPORTED, dev_id
-                        )
-                    )
-                if attr == 1:
-                    from cuda.core.experimental._memory import DeviceMemoryResource
-                    device._mr = DeviceMemoryResource(dev_id)
-                else:
-                    from cuda.core.experimental._memory import _SynchronousMemoryResource
-                    device._mr = _SynchronousMemoryResource(dev_id)
-
```
Andy-Jost, line 1003: Moved to `@property memory_resource` (lazy init). This was resulting in a circularity with `DeviceMemoryResource` using `Device`.

Issue #1199, opened 2025-10-29
Revisit `cuda-bindings` branching strategy


leofang:
Right now we have `cuda-bindings` 12.9.x maintained in the `12.9.x` branch. This causes frictions for 
- understanding the maintenance scope
   - it is not obvious at all that only `.github/`, `ci/`, `cuda_bindings/`, and `cuda_python/` are maintained there
- tracking and backpoting fixes (https://github.com/NVIDIA/cuda-python/issues/965)
- maintaining CI infra (https://github.com/NVIDIA/cuda-python/issues/1058) and docs (https://github.com/NVIDIA/cuda-python/issues/809)
- releasing 12.x packages (https://github.com/NVIDIA/cuda-python/issues/1120)

We would like to lower the burden and the mental overhead on the team. Possible solutions include:
- Eliminate the backport needs by hosting two directories `cuda_bindings_12/` and `cuda_bindings_13/` on the `main` branch 
- Generate the same binding code that have a bunch of macros that can be toggled at build time to target CUDA 12/13
- ...

rparolin:
## TL;DR

The pain comes from **having multiple long-lived branches**, not from how the source is laid out. Move to a **single branch with per-major directories** and most of it goes away. Build-time macros (proposal #2) is a nicer end-state but the wrong first step.

## What I found

- The `.pyx.in`/`.pxd.in` files are **not hand-written templates**. They're generated, with a banner: *"automatically generated… Do not modify it directly."*
- What differs between `main` (13.x) and `12.9.x` is **real CUDA 12-vs-13 API surface** (new enums, structs, renumbered values) — emitted by the generator. ~600 lines in `cydriver.pxd.in` alone.
- All generated output (56 `.pyx/.pxd/.cpp` + 18 `.in`) is **git-tracked**. The build already regenerates from the installed toolkit's headers, so the committed copies are effectively cached.

**So:** every pain you listed (#965, #1058, #809, #1120) exists because there are multiple branches. Collapse to one branch and all four go away — independent of dirs-vs-macros.

## The two proposals

**#1 — per-major directories on `main`** (recommended first step)

- One branch. Each CUDA major is a generated directory: `gen/v12/`, `gen/v13/`. New major = copy + regenerate. EOL = delete the directory. No branch is ever cut or retired.
- Hand-written code (`__init__.py`, `_lib/`, `build_hooks.py`, `ci/`, `.github/`, docs) lives **once**. A fix is one PR — no backport.
[...]

PR #1243, opened 2025-11-17
Update external library bindings

mdboom:
This updates external (nvvm, nvjitlink and cufile) based on a few bugfixes.  This wraps up 3 separate changes:

- Fix a race condition where the init function could be called multiple times (increasing startup times)
 
- Adds `cdef _get_ptr` to class wrappers as a way to get a pointer to the underlying buffer without making a Python call (this is largely in support of future work to refactor how the buffer is allocated, in support of the future NVML bindings) -- In this case, there are no functions here that make use of `_get_ptr`, but since they aren't exposed, Cython generates a `static` C function which the linker will remove because no one uses it.  For example:

```
static intptr_t __pyx_f_4cuda_8bindings_6cufile_13_py_anon_pod1__get_ptr(struct __pyx_obj_4cuda_8bindings_6cufile__py_anon_pod1 *__pyx_v_self);
```

- Fixes a bug where string buffers that have invalid UTF-8 in "garbage" memory after the null-terminator character would raise a `UnicodeDecodeError`.

### DIFF
--- a/cuda_bindings/tests/test_cufile.py
+++ b/cuda_bindings/tests/test_cufile.py
@@ -121,20 +121,6 @@ def isSupportedFilesystem():
 pytestmark = pytest.mark.skipif(not cufileLibraryAvailable(), reason="cuFile library not available on this system")
 
 
-def safe_decode_string(raw_value):
-    """Safely decode a string value from ctypes buffer."""
-    # Find null terminator if present
-    null_pos = raw_value.find(b"\x00")
-    if null_pos != -1:
-        raw_value = raw_value[:null_pos]
-    # Decode with error handling
-    try:
-        return raw_value.decode("utf-8", errors="ignore")
-    except UnicodeDecodeError:
-        # If UTF-8 fails, try to decode as bytes
-        return str(raw_value)
-
-
 def test_cufile_success_defined():
     """Check if CUFILE_SUCCESS is defined in OpError enum."""
     assert hasattr(cufile.OpError, "SUCCESS")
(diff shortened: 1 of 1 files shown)


leofang:
Q: Should we turn this function into a no-op (or remove it) to confirm the string fix?
https://github.com/NVIDIA/cuda-python/blob/6ed6f6c8a863d32cc34523f7337e1e8a90fe4616/cuda_bindings/tests/test_cufile.py#L124-L135

mdboom:
@leofang: WRT the "to-be-backported" label...  do we usually backport these non-runtime/non-driver changes?

> Q: Should we turn this function into a no-op (or remove it) to confirm the string fix?
> 
> https://github.com/NVIDIA/cuda-python/blob/6ed6f6c8a863d32cc34523f7337e1e8a90fe4616/cuda_bindings/tests/test_cufile.py#L124-L135

Yeah, I was just thinking the same thing.

leofang:
> WRT the "to-be-backported" label... do we usually backport these non-runtime/non-driver changes?

Yes. We backport all bug fixes and enhancements to the 12.9 branch. For this PR since the diff is simple (not touching GHA files or build system), the bot can automatically backport it after the PR is merged.

mdboom:
> > WRT the "to-be-backported" label... do we usually backport these non-runtime/non-driver changes?
> 
> Yes. We backport all bug fixes and enhancements to the 12.9 branch. For this PR since the diff is simple (not touching GHA files or build system), the bot can automatically backport it after the PR is merged.

Ah got it.  That was the source of my confusion.

leofang:
> Ah got it. That was the source of my confusion.

Hopefully https://github.com/NVIDIA/cuda-python/issues/1199 can help eliminate the confusion 🙂

PR #1244, opened 2025-11-17
Use new faster auto lowpp class implementation

mdboom:
This is still somewhat a WIP, but I would like to get more CI.


cuda_bindings/cuda/bindings/cycufile.pxd
```diff
@@ -156,53 +167,61 @@ cdef extern from '<cufile.h>':
...
+cdef extern from '<cufile.h>':
     ctypedef enum CUFileArrayConfigParameter_t:
         CUFILE_PARAM_POSIX_POOL_SLAB_SIZE_KB
         CUFILE_PARAM_POSIX_POOL_SLAB_COUNT
 
     # types
-    ctypedef void* CUfileHandle_t 'CUfileHandle_t'
-    ctypedef void* CUfileBatchHandle_t 'CUfileBatchHandle_t'
+ctypedef void* CUfileHandle_t 'CUfileHandle_t'
+ctypedef void* CUfileBatchHandle_t 'CUfileBatchHandle_t'
```
leofang: Q: Shouldn't these two be wrapped in a `cdef extern from '<cufile.h>'` block too?

cuda_bindings/cuda/bindings/cycufile.pxd
```diff
@@ -156,53 +167,61 @@ cdef extern from '<cufile.h>':
...
         CUfileOpError err
         CUresult cu_err
-    cdef struct _anon_pod0 '_anon_pod0':
-        unsigned int major_version
-        unsigned int minor_version
-        size_t poll_thresh_size
-        size_t max_direct_io_size
-        unsigned int dstatusflags
-        unsigned int dcontrolflags
+cdef struct _anon_pod0 '_anon_pod0':
```
leofang, line 193: nit: can we generate one empty line above, to separate out each code block?

cuda_bindings/cuda/bindings/cycufile.pxd
```diff
@@ -156,53 +167,61 @@ cdef extern from '<cufile.h>':
...
+    unsigned int major_version
+    unsigned int minor_version
+    size_t poll_thresh_size
+    size_t max_direct_io_size
+    unsigned int dstatusflags
+    unsigned int dcontrolflags
+cdef extern from '<cufile.h>':
```
leofang, line 201: ditto, re: empty line

cuda_bindings/cuda/bindings/cufile.pyx
```diff
@@ -39,95 +53,109 @@ cdef class _py_anon_pod1:
...
+        object _owner
+        bint _readonly
+        dict _refs
 
     def __init__(self):
-        arr = _numpy.empty(1, dtype=_py_anon_pod1_dtype)
-        self._data = arr.view(_numpy.recarray)
-        assert self._data.itemsize == sizeof((<CUfileDescr_t*>NULL).handle), \
-            f"itemsize {self._data.itemsize} mismatches union size {sizeof((<CUfileDescr_t*>NULL).handle)}"
+        self._ptr = <_anon_pod1 *>calloc(1, sizeof((<CUfileDescr_t*>NULL).handle))
```
leofang, line 64: Q: Do we have to zero-initialize the memory? It was not a guarantee that we provided.

cuda_bindings/cuda/bindings/cycufile.pxd
kkraus14, line 193: Was it intentional to move these from being externed to now no longer being externed? We should make sure this doesn't cause an ABI break if so.

leofang: These are inner, anonymous structs split apart by the codegen with a unique name assigned. They are not supposed to be used by the end users, and so while moved out of the `extern` block changes the name mangling, I don't think we should consider them API/ABI breaking.

cuda_bindings/cuda/bindings/cycufile.pxd
mdboom: Yes, probably.  But they aren't strictly needed since they are opaque.  For correctness/completeness they could be though.

cuda_bindings/cuda/bindings/cufile.pyx
mdboom: We either need to do this, set only pointer fields to NULL, or provide a public API to init the memory.  `calloc` just seemed like the most solid option.  Consider this:

```
x = MyClass()
x.array_member = [1, 2, 3]
```

How do we know whether there is already a value at `array_member` that needs to be freed, if it isn't initially NULL?

cuda_bindings/cuda/bindings/cycufile.pxd
mdboom: @kkraus14: The `cdef extern from '<cufile.h>':` means "this struct is external, use it from the header", not "make this an external symbol".  If you do that, Cython assumes the C struct exists in the header and won't generate one.  In this case, we need to generate one since the generator creates these as an implementation detail, and we need the structs in order to access their members.

kkraus14: My concern is that existing Cython generated code from using an older `cuda.bindings` module references the existing symbols generated from these structs via some `__pyx_import` call or otherwise and that changing whether it was externed or not would change the symbols (as we previously saw during the 12.6 ABI breakage).

I wasn't aware that this struct didn't exist in the header and that it wasn't actually used in any of our `.pyx` code so at no point is a symbol ever actually generated.

kkraus14: Resolving this.

cuda_bindings/cuda/bindings/cufile.pyx
leofang: Oh I was asking this: Can we use `malloc` instead of using `calloc`? We need to allocate memory but not initialize it? The old code uses `np.empty` which does not initialize memory either, but we'd get a valid pointer while skipping one memory write.

In any case this is not blocking.

mdboom: I don't see how we can safely handle array members of structs without setting at least the pointer field to NULL.  I know the previous code basically did that with `np.empty`, but it means you can't safely use the object until (potentially) every member has been initialized.  Maybe we could use the existence of a key in `self._refs` to determine if the array member contains data.

We at minimum would need to do something like:

```
ptr = malloc(...)
ptr.arrayMember = NULL
# or
ptr.arrayMemberSize = 0
```

It makes the generation a little complicated, and I'm also not sure it would be faster than calloc.

mdboom: And thinking about this more, string members (whether pointers or inline buffers) also require `calloc`.

cuda_bindings/cuda/bindings/cufile.pyx
```diff
@@ -39,95 +53,109 @@ cdef class _py_anon_pod1:
     .. seealso:: `_anon_pod1`
     """
     cdef:
-        readonly object _data
+        _anon_pod1 *_ptr
+        object _owner
+        bint _readonly
+        dict _refs
```
leofang: Q: Based on the diff I don't see `_refs` is used anywhere?

cuda_bindings/cuda/bindings/cufile.pyx
```diff
@@ -39,95 +53,109 @@ cdef class _py_anon_pod1:
...
             return False
-        if self._data.size != other._data.size:
-            return False
-        if self._data.dtype != other._data.dtype:
-            return False
-        return bool((self._data == other._data).all())
+        other_ = other
+        return (memcmp(<void *><intptr_t>(self._ptr), <void *><intptr_t>(other_._ptr), sizeof((<CUfileDescr_t*>NULL).handle)) == 0)
+
+    def __setitem__(self, key, val):
```
leofang, line 99: Q:
1. what is special about key == 0?
2. I am trying to understand the `setattr` path below -- wouldn't it lead to infinite recursion?

mdboom: The way the diff is rendered makes it a bit hard to see, but the previous implementation of `__setitem__` was:

```
def __setitem__(self, key, val):
        self._data[key] = val
```

(where `self._data` is a Numpy array).

This let you do the following:

```
data = np.empty(1, dtype=test.auto_lowpp_class_dtype)
obj = test.AutoLowppClass()
obj[0] = data
```

"0" is special because `AUTO_LOWPP_CLASS` behaves like an array of length 1.

The `setattr` thing is likewise to support a Numpy-ism:

```
obj = test.AutoLowppClass()
obj['field'] = 42
# equivalent to
obj.field = 42
```

For the record, I don't think this any of this is a good idea, but I am trying not to remove API.  On the other hand, I don't think the entirety of the Numpy `__setitem__` API was ever /intentionally/ exposed, it just was.

If it were just me, and we didn't care about backward-compat, I wouldn't implement a `__setitem__` at all here.  What do you think, @leof?

cuda_bindings/cuda/bindings/cufile.pyx
mdboom: That seems like a bug.  Will fix.

cuda_bindings/cuda/bindings/cufile.pyx
leofang: Ah... Thanks for detailed explanation, Mike! I think what you have is fine. Let's get this PR in.

It was designed to allow `obj.field = 42` because for single structs this is more intuitive than `obj['field'] = 42`. I think at the codegen level we used to ensure internally we don't just have a structured array, but a record array that allows attribute access. I don't think introducing breaking changes is a good idea. The new code that you generated seems reasonable to maintain. 

> I wouldn't implement a `__setitem__` at all here.

What would be the alternative if we were to go with this?

leofang (approved):

cuda_bindings/cuda/bindings/cufile.pyx
mdboom: > It was designed to allow obj.field = 42 because for single structs this is more intuitive than obj['field'] = 42.

I think it was actually the opposite.  The `__setitem__` supports the `obj['field']` syntax, which would otherwise not work.

> What would be the alternative if we were to go with this?

I'm suggesting that /if designed today/ we probably wouldn't want to allow `obj['field']` to work, since, like you say, it's less intuitive.

I'm fine with merging this as-is, but we may want to deprecate `__setitem__` on autogen classes, and do the usual "raise a warning for a few releases and then remove it" thing.  But that can be a separate follow-on discussion.

leofang: Makes sense. It was done like this because in the beginning we did not have a clean separation between `CLASS` and `ARRAY`. The kv-access like `obj['field']` is a must for the latter because of how NumPy array slicing works, but not so for the former. Now that we have a clean separation, indeed we should think about how to deprecate and remove this by v14 or v15.

mdboom:
I have confirmed with the tool in https://github.com/NVIDIA/cuda-python/pull/1067 that there is no change in the ABI.

Unfortunately, the changes in this PR are /not/ tested in CI, given this on all of the tests that use these types.  We will need to perform some manual testing.

```
@pytest.mark.skipif(not isSupportedFilesystem(), reason="cuFile handle_register requires ext4 or xfs filesystem")
```

kkraus14 (approved):

cuda_bindings/docs/source/release/13.X.Y-notes.rst
```diff
@@ -3,7 +3,7 @@
 
 .. module:: cuda.bindings
 
-``cuda-bindings`` 13.X.Y Release notes
+````cuda-bindings```` 13.X.Y Release notes
```
rwgk: There are a few quadruple backticks in this commit. (I noticed in passing.)

cuda_bindings/cuda/bindings/cufile.pyx
```diff
@@ -1845,113 +2211,151 @@ cdef class StatsLevel2:
...
+        cdef StatsLevel1 val_ = val
+        memcpy(<void *>&(self._ptr[0].basic), <void *>(val_._get_ptr()), sizeof(CUfileStatsLevel1_t) * 1)
 
     @property
     def read_size_kb_hist(self):
         """~_numpy.uint64: (array of length 32)."""
-        return self._data.read_size_kb_hist
+        cdef view.array arr = view.array(shape=(32,), itemsize=sizeof(uint64_t), format="Q", mode="c", allocate_buffer=False)
+        arr.data = <char *>(&(self._ptr[0].read_size_kb_hist))
+        return arr
```
leofang: Q: Can we wrap it as a numpy array here to avoid breaking?

cuda_bindings/cuda/bindings/cufile.pyx
```diff
@@ -1845,113 +2211,151 @@ cdef class StatsLevel2:
...
+        arr[:] = _numpy.asarray(val, dtype=_numpy.uint64)
+        memcpy(<void *>(&(self._ptr[0].read_size_kb_hist)), <void *>(arr.data), sizeof(uint64_t) * len(val))
 
     @property
     def write_size_kb_hist(self):
         """~_numpy.uint64: (array of length 32)."""
-        return self._data.write_size_kb_hist
+        cdef view.array arr = view.array(shape=(32,), itemsize=sizeof(uint64_t), format="Q", mode="c", allocate_buffer=False)
+        arr.data = <char *>(&(self._ptr[0].write_size_kb_hist))
+        return arr
```
leofang: ditto

mdboom:
I have reverted the backward compatibility by returning numpy arrays, rather than Cython arrays, from members that have numeric array types.

I tested this again on a machine that doesn't skip the `cufile` tests and all is passing there.

There is still one breaking change here, noted in the release notes, that IMHO is just a real bug -- `PerGpuStats` was declared as a `AUTO_LOWPP_CLASS` rather than an `AUTO_LOWPP_ARRAY`, but it is always used as an array.  You can see in the test code the weird backflips that were required to access it as an array that go away if you just declare it as such.  Unfortunately, this is required here because I can't make `PerGpuStats[0]` return a Numpy array as before under the new implementation, but that never really made sense anyway.

PR #1245, opened 2025-11-17
Upgrade Cython dependency to 3.2

mdboom:
The NVML bindings are so large that they are hitting [this bug](https://github.com/cython/cython/commit/c3594439237fe694021b684e525775fcfb14fc44), which was fixed in Cython 3.2.  I'm upgrading to Cython 3.2 here to see if that breaks anything in our CI.

### DIFF
--- a/cuda_bindings/pyproject.toml
+++ b/cuda_bindings/pyproject.toml
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2023-2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
 [build-system]
-requires = ["setuptools>=77.0.0", "cython>=3.1,<3.2", "pyclibrary>=0.1.7"]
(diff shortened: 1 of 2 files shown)


mdboom:
From another conversation with @vyasr, Cython 3.2 may have a bug affecting free-threaded builds, [fixed by this PR](https://github.com/cython/cython/pull/7313#issuecomment-3515406116), not yet released.  The cuda-bindings CI is unlikely to pick up this bug, which still may have real consequences for our users if they are unlucky.

mdboom:
Another note -- the workaround suggested by the Cython team (to increase `sys.maxrecursionlimit`) doesn't work for the NVML bindings -- they are too large.  So I do think we need this fixed "the right way".

cpcloud:
Re free threaded builds, is there anything we can do other than simply wait for the fix to land? I suspect we have zero user using our free-threaded builds at the moment.

mdboom:
> Re free threaded builds, is there anything we can do other than simply wait for the fix to land? I suspect we have zero user using our free-threaded builds at the moment.

Yeah, given we don't expect it to affect many, we could put a note in the "known issues".  I'm not sure how best to word it -- it's sort of "if you are using a lot of threads on a free threaded build, it could segfault", but that's kind of par for the course anyway right now...?

PR #1250, opened 2025-11-17
feat: add from_* style constructor classmethods to StridedMemoryView and deprecate `__init__` constructor

cpcloud:
Redo of #1224 without breaking changes.

### DIFF
--- a/cuda_core/tests/test_utils.py
+++ b/cuda_core/tests/test_utils.py
@@ -82,7 +82,7 @@ class TestViewCPU:
 
     def test_strided_memory_view_cpu_init(self, in_arr):
         # stream_ptr=-1 means "the consumer does not care"
-        with pytest.warns(UserWarning, match="deprecated"):
+        with pytest.deprecated_call(match="deprecated"):
             view = StridedMemoryView(in_arr, stream_ptr=-1)
         self._check_view(view, in_arr)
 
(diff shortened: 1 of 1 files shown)


cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -78,30 +79,72 @@ cdef class StridedMemoryView:
...
+
+    def __init__(self, obj: object = None, stream_ptr: int | None = None) -> None:
+        cdef str clsname = self.__class__.__name__
         if obj is not None:
             # populate self's attributes
             if check_has_dlpack(obj):
+                warnings.warn(
+                    f"Constructing a {clsname} directly from a DLPack-supporting object is deprecated; "
+                    "Use `StridedMemoryView.from_dlpack` or `StridedMemoryView.from_any_interface` instead."
+                )
```
kkraus14, line 112: Can we make this a `FutureWarning` (https://docs.python.org/3/library/exceptions.html#FutureWarning) or whatever warning type is most appropriate for this deprecation?

mdboom: I think the one we want is `DeprecationWarning`.  Those are usually off by default, but tools like `pytest` will collect and display them at the end, so they get shown to developers using the library during their development rather than to "end users".  `FutureWarning` is always on, so is meant for non-developers to see.  (Think of like a data processing script that will require the user to change the format of their data in the future...)  Fuzzy boundaries between those, of course.

leofang: @cpcloud let's add the warning and get the PR merged

cpcloud: Sounds good!

cpcloud: Done.

PR #1258, opened 2025-11-18
BUG-FIX: Remove duplicate try_site_packages() call in load_nvidia_dynamic_lib

rparolin:
BUG-FIX: Remove duplicate try_site_packages() call in load_nvidia_dynamic_lib

Remove redundant call to `try_site_packages()` that was immediately overwritten by the same call on the next line. This duplicate call was wasteful (unnecessary filesystem traversal).


rwgk:
Thanks @rparolin for fixing that oversight! (I just checked, I overlooked it when reviewing #1049.)

PR #1267, opened 2025-11-19
test: add `pytest-randomly` to `cuda_bindings`

cpcloud:
Add pytest-randomly to cuda_bindings. Tests are randomized by default. [`pytest-randomly` docs](https://github.com/pytest-dev/pytest-randomly/blob/main/README.rst)

### DIFF
--- a/cuda_bindings/pixi.lock
+++ b/cuda_bindings/pixi.lock
@@ -60,6 +60,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/gxx_impl_linux-64-15.2.0-hda75c37_15.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/harfbuzz-12.2.0-h15599e2_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/icu-75.1-he02047a_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/importlib-metadata-8.7.0-pyhe01879c_1.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/iniconfig-2.3.0-pyhd8ed1ab_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/intel-gmmlib-22.8.2-hb700be7_0.conda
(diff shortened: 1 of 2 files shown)


rwgk:
Looks like this particular CI run uncovered 3 types of errors.

It seems to be — random — unsurprisingly I guess, e.g. https://github.com/NVIDIA/cuda-python/actions/runs/19505585558?pr=1268 happened to succeed, and one of the errors here was also triggered under https://github.com/NVIDIA/cuda-python/actions/runs/19505615062?pr=1269

At the initial stage introducing `pytest-randomly`, maybe it'll help us to start with a fixed seed, e.g.
```
export RANDOMLY_SEED=12345
```
?

cpcloud:
> At the initial stage introducing pytest-randomly, maybe it'll help us to start with a fixed seed, e.g.

In practice this leads to never changing that because we'll forget about it, so we should fix the errors and then let it continue to run in a random order. The idea is to be robust to the randomness, acknowledging that this can make things seem flaky.

While I don't like leaving PRs open, this one seems like it might need to be open to get to a point where it's worth merging.

rwgk:
> so we should fix the errors and then let it continue to run in a random order.

This could be exhausting, and we risk not getting this done in a reasonable timeframe.

It could also be distracting in unfortunate ways, e.g. around releases.

> because we'll forget about it,

I think we should do this in a controlled way and create a bug to track the stages, roughly:

* Get tests working with fixed random seed.
* After working with that for about a month, change the random seed (but still fixed).
* After no significant distractions for a month, remove the fixed random seed entirely.

kkraus14:
@cpcloud there's an inherent level of statefulness in CUDA, i.e. driver initialization, context creation and setting it to be current, etc. This statefulness is also quite expensive where we absolutely can't afford to create and tear down contexts for each test for example. Is there a way we can randomize at the module or class level instead of the test level so that we can have more control over some assumptions and control some of these costs?

cpcloud:
This PR depends on https://github.com/NVIDIA/cuda-python/pull/1313, since that isolates fixture state from cufile tests, allowing them to run in a random order without depending on initialization details

PR #1269, opened 2025-11-19
test: add `pytest-randomly` to `cuda_pathfinder`

cpcloud:
Add pytest-randomly to cuda_pathfinder. Tests are randomized by default. [`pytest-randomly` docs](https://github.com/pytest-dev/pytest-randomly/blob/main/README.rst)


rparolin:
Is this going to randomize the order of our unit tests for ever pytest run? I thought we were going to have a specific CI job that randomized the order to avoid inflicting flakiness on people while we work out bugs.

cpcloud:
Ended up just disabling randomly for bindings core for now, since this PR is only about pathfinder.

cpcloud:
@rparolin Missed your comment. Happy to revert, but it might be easier to just see whether it's a burden. If it is, let's revert then, is my thinking.

Issue #1143, opened 2025-10-16
[FEA]: Improve CI coverage of cuFile


mdboom:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

cuda.bindings

### Is your feature request related to a problem? Please describe.

The majority of the tests in `test_cufile.py` are skipped because:

`@pytest.mark.skipif(not isSupportedFilesystem(), reason="cuFile handle_register requires ext4 or xfs filesystem")`

Maybe we should try to add a worker to the mix that /is/ on one of these filesystems to get some better coverage there.  I don't know what filesystem our existing workers have (maybe btrfs?)

### Describe the solution you'd like

Find a worker that meets the requirements of cuFile that we can run on.

leofang:
@vyasr @kingcrimsontianyu How do you guys test kvikio with nv-gha-runners? Are you able to get full test coverage over everything cuFILE supports?

kingcrimsontianyu:
My underwhelming answer :smile: is that we haven't tested actual GDS I/O on cloud/CI. When KvikIO detects at runtime that the system setup does not meet GDS requirements, it will switch to the parallel I/O implementation which is based on the standard Linux system calls.

leofang:
Thanks, Tianyu! Sounds like this is blocked 🥲

leofang:
Blocked by https://github.com/nv-gha-runners/roadmap/issues/318

PR #1271, opened 2025-11-19
test: fix overly restrictive cufile skipping logic

cpcloud:
# Summary

## Problem Statement

The cuFile test suite was incorrectly skipping valid filesystem tests due to a flawed mount point detection algorithm.

### Root Cause

The previous implementation had a critical bug in the `isSupportedFilesystem()` function:

1. **Incorrect Logic**: It assumed that if any mount point from `/proc/mounts` was a prefix of the current directory, that mount point was the one being used.

2. **The Bug**: Since `/` is a prefix of all paths, the function always matched the root filesystem (`/`) regardless of where the actual test directory was mounted.

3. **Impact**: Tests were incorrectly skipped based on the root filesystem type instead of the actual filesystem where the test directory resided. This meant:
   - Valid ext4/xfs filesystems mounted elsewhere (e.g., `/mnt/data`, `/home/user/project`) were incorrectly skipped
   - Only tests running directly on the root filesystem would work correctly
   - **Critical**: The bug was actually *hiding test failures* in CI environments (see "CI Environment Context" below)

### Example Scenario
```
/mnt/ext4_data (ext4 filesystem)  ← Test directory here
/ (root filesystem, could be any type)
```
The old code would check if `/` is a prefix of `/mnt/ext4_data` → **YES** → incorrectly use root filesystem type instead of ext4.

### Key Changes

1. **Replaced Manual Parsing with `findmnt`**
   - **Before**: Manually parsed `/proc/mounts` and used flawed prefix matching
   - **After**: Uses the `findmnt` command-line tool with `-T .` flag to find the mount point for the current directory

2. **Proper Mount Point Resolution**
   - `findmnt -no FSTYPE -T .` correctly identifies the filesystem type for the current directory
   - Leverages kernel's mount table logic for accurate resolution
   - Handles nested mounts and bind mounts correctly

## Benefits

1. **Correctness**: Tests now accurately detect filesystem type regardless of mount point location
2. **Reliability**: Uses kernel-provided tools instead of manual parsing
3. **Maintainability**: Simpler code (reduced from ~30 lines to ~8 lines)
4. **Robustness**: Handles edge cases like bind mounts, nested mounts, and network filesystems correctly

## Testing Impact

- **Before**: Tests on non-root ext4/xfs filesystems were incorrectly skipped
- **After**: Tests correctly identify and run on supported filesystems regardless of mount point
- **Revealed Issue**: The fix exposes that tests fail in CI even on supported filesystems (hence xfail markers)

### The Hidden Failure Problem

This bug fix reveals a critical issue that was previously masked:

1. **GitHub Actions Runners**: Run code inside containers, but the code location is mounted on an **ext4 filesystem** (not overlayfs)

2. **Previous Behavior** (with buggy code):
   - Buggy detection matched `/` first (root filesystem)
   - In containers, root is often overlayfs
   - Tests were skipped because overlayfs is not supported
   - **Tests never ran, so failures were hidden**

3. **Current Behavior** (with fixed code):
   - Correct detection identifies ext4 filesystem
   - Tests are no longer skipped
   - **Tests now run but fail** (hence xfail markers)
   - The failures would have been present all along if detection had been correct

4. **Tests Fail in CI**:
   - Even though the filesystem is ext4 (nominally supported), cuFile operations fail in containerized CI environments
   - The xfail markers acknowledge this known limitation while allowing the fix to proceed

### Key Insight

**The bug was hiding test failures.** Had the filesystem mount point discovery been correct from the start, these tests would have been failing in CI. The fix:
- ✅ Corrects the filesystem detection logic
- ✅ Reveals the pre-existing CI failures
- ✅ Documents them with xfail markers for future investigation

### DIFF
--- a/cuda_bindings/pixi.lock
+++ b/cuda_bindings/pixi.lock
@@ -3,6 +3,8 @@ environments:
   cu13:
     channels:
     - url: https://conda.anaconda.org/conda-forge/
+    options:
+      pypi-prerelease-mode: if-necessary-or-explicit
     packages:
       linux-64:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
@@ -521,6 +523,8 @@ environments:
   default:
     channels:
     - url: https://conda.anaconda.org/conda-forge/
+    options:
+      pypi-prerelease-mode: if-necessary-or-explicit
     packages:
       linux-64:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
@@ -947,6 +951,8 @@ packages:
   version: 13.0.3
   build: py314hf14b7c0_0
   subdir: linux-64
+  variants:
+    python: 3.14.*
   depends:
   - python
   - cuda-pathfinder >=1.1,<2
@@ -962,7 +968,7 @@ packages:
   - python_abi 3.14.* *_cp314
   license: LicenseRef-NVIDIA-SOFTWARE-LICENSE
   input:
-    hash: 5b9ca621be4a7a80d774ea4b72d3a88bcfbf672781f7d43b46293b41121afc66
+    hash: 82cc3b40bfe11e69e9d2054bc6f71d78370192927f2541b1921842b25bc59594
     globs:
     - pyproject.toml
 - conda: .
@@ -970,6 +976,8 @@ packages:
   version: 13.0.3
   build: py314hf14b7c0_0
   subdir: linux-aarch64
+  variants:
+    python: 3.14.*
   depends:
   - python
   - cuda-pathfinder >=1.1,<2
@@ -985,7 +993,7 @@ packages:
   - python_abi 3.14.* *_cp314
   license: LicenseRef-NVIDIA-SOFTWARE-LICENSE
   input:
-    hash: 5b9ca621be4a7a80d774ea4b72d3a88bcfbf672781f7d43b46293b41121afc66
+    hash: 82cc3b40bfe11e69e9d2054bc6f71d78370192927f2541b1921842b25bc59594
     globs:
     - pyproject.toml
 - conda: .
@@ -993,6 +1001,8 @@ packages:
   version: 13.0.3
   build: py314hf14b7c0_0
   subdir: win-64
+  variants:
+    python: 3.14.*
   depends:
   - python
   - cuda-pathfinder >=1.1,<2
@@ -1005,7 +1015,7 @@ packages:
   - python_abi 3.14.* *_cp314
   license: LicenseRef-NVIDIA-SOFTWARE-LICENSE
   input:
-    hash: 5b9ca621be4a7a80d774ea4b72d3a88bcfbf672781f7d43b46293b41121afc66
+    hash: 82cc3b40bfe11e69e9d2054bc6f71d78370192927f2541b1921842b25bc59594
     globs:
     - pyproject.toml
 - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-cccl_linux-64-13.0.85-ha770c72_0.conda


cpcloud:
Turns out these tests are never run on `main` because they are skipped. See https://github.com/NVIDIA/cuda-python/actions/runs/19510167362/job/55859660002 and search for `cuda.bindings`, you'll see this:

image

cpcloud:
The logic I have here is I believe correct, but I'm not sure if these tests are just broken, or this is a latent bug that we've been living with for a while.

rwgk:
> Turns out these tests are never run on main because they are skipped.

Tangential:

Yeah, I ran into that twice with the ctk-next testing. SWQA was the first to run in environments where they are not skipped, but I didn't think of checking first if they actually ran in the cuda-python CI and started debugging painstakingly assuming they did.

It'd be awesome to have some sort of automatic flagging, after all jobs finish, to report which tests

* are skipped everywhere
* are always skipped on linux-64, linux-aarch64, win-64

leofang:
1. For cuFILE related changes, let us always rope in the GDS team for consultation (cc @sourabgupta3). 
2. Currently lots of things cannot be run in the CI (https://github.com/NVIDIA/cuda-python/issues/1143), not even in cuFILE's compat mode IIUC.

cpcloud:
It's really an issue of incorrect skipping logic, not whether a thing runs in CI. This *should* have been running, because the runners have a compatible file system.

This wasn't running, but it was unintentionally not running.

cpcloud:
Given all the failures happen on 13.0.2, and the 12.9.x tests all pass, that seems like a good place to start digging.

cpcloud:
Ah, well those are not running the cuda.bindings tests, so that wouldn't explain the failure.

leofang:
Let's give Sourab some time. I pinged him offline and he's looking into it.

leofang:
> Let's give Sourab some time. I pinged him offline and he's looking into it.

Ping'd you both in an internal thread.

cpcloud:
What's blocking this PR? This PR is only about fixing the testing problem, and it shouldn't be blocked on any fixes/problems that were discovered here. This PR isn't changing any existing behavior, except that now the tests that were skipped when they should've been running (because the skipping criteria checking function was broken) are now xfailing.

We can merge this PR without waiting for cufile fixes, and we can continue the discussion on that elsewhere.

leofang:
Asked internally but to recap:
- We observed and agreed `isSupportedFilesystem` should return `False` in the CI, but then the xfail condition would not be triggered, so why suddenly we need to add them now?
- We proved that the filesystem in the CI is ext4, so why wouldn't `isSupportedFilesystem` return `True`?

rwgk:
I'm trying to understand. What I believe I'm seeing:

* Phillip observed that the logic in `isSupportedFilesystem()` is faulty and tried to fix it.

* After the fix, the tests are not skipped anymore, but fail, when they are expected to succeed.

* Phillip wants to separate:
    * fixing `isSupportedFilesystem()` (the work is done already) from
    * fixing the bug that makes the tests fail unexpectedly.

* Phillip's method for achieving that is to mark the tests as `xfail`.

* The net result is that, across all test logs, 169 lines like this one
```
2025-12-02T17:51:18.7693251Z SKIPPED [1] tests/test_cufile.py:156: cuFile handle_register requires ext4 or xfs filesystem
```
are replaced by 169 lines like this one:
```
2025-12-02T19:18:15.6874406Z XFAIL tests/test_cufile.py::test_handle_register - handle_register call fails in CI for unknown reasons
```

* No test that was incorrectly skipped before passes as expected.

@cpcloud is that a correct understanding?

cpcloud:
That is correct.

rwgk:
I looked around some more and met (huddle) with Phillip. We think this will be the most robust and simplest replacement for the `isSupportedFilesystem()` function:

```
@cache
def isSupportedFilesystem():
    """Check if the current filesystem is supported (ext4 or xfs).

    This uses `findmnt` so the kernel's mount table logic owns the decoding of the filesystem type.
    """

    fs_type = (
        subprocess.check_output(
            ["findmnt", "-no", "FSTYPE", "-T", "."],
            text=True,
        )
        .strip()
    )
    logging.info(f"Current filesystem type (findmnt): {fs_type}")

    return fs_type in ("ext4", "xfs")
```

For easy reference, I'm attaching what Cursor told me about availability of `findmnt`:

```
How common is `findmnt`?

  Big picture: findmnt is part of util-linux, which is installed by default on essentially all mainstream Linux distros. So in practice it's very likely to be present anywhere you can
  realistically run cuFile.

  1. GitHub-hosted runners (Ubuntu)

  • Likelihood: extremely high.
  • GitHub's Ubuntu images (20.04, 22.04, etc.) are full server-ish installs with util-linux by default.
  • findmnt is used by a lot of system tooling; stripping it out would be unusual. I'd treat it as "present unless someone intentionally broke the image," which GitHub won't.


  2. "Random" Linux dev workstations

  • Likelihood: very high on mainstream distros.
  • On Ubuntu, Debian, Fedora, RHEL, SUSE, Arch, etc., util-linux is part of the base system. Desktop/WS installs have findmnt out of the box.
  • The main exceptions would be:
    • Really minimal/container images (Alpine-style, scratch images, custom stripped-down builds).
    • Unusual non-glibc environments.
  • But for a CUDA/cuFile dev machine, it's overwhelmingly likely to be a normal distro with findmnt installed.


  3. Jetson (L4T)

  • Likelihood: very high.
  • L4T is an Ubuntu-derived root filesystem + NVIDIA bits. Ubuntu's base system includes util-linux, and I'd strongly expect findmnt to be there on stock Jetson images.
  • If someone has drastically pruned the rootfs, they could remove it, but that would be an atypical, highly customized setup.
```

cpcloud:
SGTM, thanks @rwgk!

cpcloud:
I'll change `"."` to `os.getcwd()` but that's the only major alteration.

cpcloud:
Created a follow-up here: https://github.com/NVIDIA/cuda-python/issues/1307

rwgk (approved):
LGTM - Great baseline for cufile team to start investigating #1307

cpcloud:
> I could not reproduce this observation in a similar local env, but I am giving up digging

I finally had a chance to get to the bottom of this and it was actually a bug in my `find_mount_point` implementation.

My approach of looking at `Path(...).parents` doesn't search the current directory--which itself could be a mount point--so when you were running from `/home/leofang`, the search would start at `/home` (the first element of `.parents`) which wasn't a mount point, and then terminate at `/` (which of course is, but is overlayfs), hence the `False` we observed.

`findmnt` doesn't have this problem, so switching to that actually fixed a bug!

PR #1272, opened 2025-11-19
Fix an invalid memory access in call to cuMemPoolGetAttribute

Andy-Jost:
## Description

Changes the argument size from 32-bits to 64-bits, as expected by the driver.

closes NVIDIA/cuda-python-private#197

### DIFF
--- a/cuda_core/cuda/core/experimental/_memory/_device_memory_resource.pyx
+++ b/cuda_core/cuda/core/experimental/_memory/_device_memory_resource.pyx
@@ -5,7 +5,7 @@
 from __future__ import annotations
 
 from libc.limits cimport ULLONG_MAX
-from libc.stdint cimport uintptr_t
+from libc.stdint cimport uintptr_t, uint64_t
 from libc.string cimport memset
 
 from cuda.bindings cimport cydriver
@@ -116,10 +116,10 @@ cdef DMRA_mempool_attribute(property_type: type):
     return decorator
 
 
-cdef int DMRA_getattribute(
+cdef uint64_t DMRA_getattribute(
     cydriver.CUmemoryPool pool_handle, cydriver.CUmemPool_attribute attr_enum
 ):
-    cdef int value
+    cdef uint64_t value = 0
     with nogil:
         HANDLE_RETURN(cydriver.cuMemPoolGetAttribute(pool_handle, attr_enum, <void *> &value))
     return value


rwgk:
I see this in the documentation:

* https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MALLOC__ASYNC.html#group__CUDA__MALLOC__ASYNC_1gd45ea7c43e4a1add4b971d06fa72eda4

___


Description
Supported attributes are:

[CU_MEMPOOL_ATTR_RELEASE_THRESHOLD](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1gg5af6ea9ddd7633be98cb7de1bbf1d9f050fa2c731b01422ffbeb8c16ce0ba9a8): (value type = cuuint64_t) Amount of reserved memory in bytes to hold onto before trying to release memory back to the OS. When more than the release threshold bytes of memory are held by the memory pool, the allocator will try to release memory back to the OS on the next call to stream, event or context synchronize. (default 0)

[CU_MEMPOOL_ATTR_REUSE_FOLLOW_EVENT_DEPENDENCIES](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1gg5af6ea9ddd7633be98cb7de1bbf1d9f0e11d636e9d56f9ce8a449b887fe2917f): (value type = int) Allow [cuMemAllocAsync](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MALLOC__ASYNC.html#group__CUDA__MALLOC__ASYNC_1g13413273e84a641bce1929eae9e6501f) to use memory asynchronously freed in another stream as long as a stream ordering dependency of the allocating stream on the free action exists. Cuda events and null stream interactions can create the required stream ordered dependencies. (default enabled)

[CU_MEMPOOL_ATTR_REUSE_ALLOW_OPPORTUNISTIC](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1gg5af6ea9ddd7633be98cb7de1bbf1d9f0d1016ed326b53c1eb06e8da5d40359cb): (value type = int) Allow reuse of already completed frees when there is no dependency between the free and allocation. (default enabled)

[CU_MEMPOOL_ATTR_REUSE_ALLOW_INTERNAL_DEPENDENCIES](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1gg5af6ea9ddd7633be98cb7de1bbf1d9f03b7973c1b89f3a05f26702ade1124e9f): (value type = int) Allow [cuMemAllocAsync](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MALLOC__ASYNC.html#group__CUDA__MALLOC__ASYNC_1g13413273e84a641bce1929eae9e6501f) to insert new stream dependencies in order to establish the stream ordering required to reuse a piece of memory released by [cuMemFreeAsync](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MALLOC__ASYNC.html#group__CUDA__MALLOC__ASYNC_1g41acf4131f672a2a75cd93d3241f10cf) (default enabled).

[CU_MEMPOOL_ATTR_RESERVED_MEM_CURRENT](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1gg5af6ea9ddd7633be98cb7de1bbf1d9f0ecd4938eb06ce224e04d4c56fea476c6): (value type = cuuint64_t) Amount of backing memory currently allocated for the mempool

[CU_MEMPOOL_ATTR_RESERVED_MEM_HIGH](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1gg5af6ea9ddd7633be98cb7de1bbf1d9f05ccb2736772fd27a735f15905118cbf6): (value type = cuuint64_t) High watermark of backing memory allocated for the mempool since the last time it was reset.

[CU_MEMPOOL_ATTR_USED_MEM_CURRENT](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1gg5af6ea9ddd7633be98cb7de1bbf1d9f03f17742ca47490659f188bc75be9b85c): (value type = cuuint64_t) Amount of memory from the pool that is currently in use by the application.

[CU_MEMPOOL_ATTR_USED_MEM_HIGH](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1gg5af6ea9ddd7633be98cb7de1bbf1d9f027fd3cf6d5a152572e55e9736da6987b): (value type = cuuint64_t) High watermark of the amount of memory from the pool that was in use by the application.

___

There are three attributes with value type `int`.

Is that real?

If it's real, could we still get into trouble because of big-endian vs little-endian?

cuda_core/cuda/core/experimental/_memory/_device_memory_resource.pyx
```diff
@@ -116,10 +116,10 @@ cdef DMRA_mempool_attribute(property_type: type):
...
 
-cdef int DMRA_getattribute(
+cdef uint64_t DMRA_getattribute(
     cydriver.CUmemoryPool pool_handle, cydriver.CUmemPool_attribute attr_enum
 ):
-    cdef int value
+    cdef uint64_t value
     with nogil:
         HANDLE_RETURN(cydriver.cuMemPoolGetAttribute(pool_handle, attr_enum, <void *> &value))
     return value
```
leofang, line 125: This technically is still not correct, because the driver sometimes does want just an `int`.
https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MALLOC__ASYNC.html#group__CUDA__MALLOC__ASYNC_1gd45ea7c43e4a1add4b971d06fa72eda4
Wouldn't this result in only the first 32 bits being written in those cases?

rwgk:
> Wouldn't this result in only the first 32 bits being written in those cases?

Or the last 32 bits, depending on endian-ness.

leofang:
Right. 

Another issue: Why don't we see any segfault in the CI? We even turned on compute-sanitizer in some CI runs.

cuda_core/cuda/core/experimental/_memory/_device_memory_resource.pyx
Andy-Jost: Yes, I need to initialize to zero.

Andy-Jost:
> > Wouldn't this result in only the first 32 bits being written in those cases?
> 
> Or the last 32 bits, depending on endian-ness.

It appears that the properties using `int` are using it in liu of `bool`. We should be okay in those cases since we convert to `bool` before returning to Python.

Andy-Jost:
> Right.
> 
> Another issue: Why don't we see any segfault in the CI? We even turned on compute-sanitizer in some CI runs.

Valgrind also could not detect this, since the bad write clobbers part of a local variable on the stack. When it crashes, a pointer is sliced as follows: 0x7fffxxxxxxxx -> 0x7fff00000000. I cannot explain why the crash is intermittant.

leofang:
I propose an alternative solution (https://github.com/NVIDIA/cuda-python/pull/1274) that should be safer, more extensible, and with the bonus of slightly faster.

cuda_core/cuda/core/experimental/_memory/_device_memory_resource.pyx
```diff
@@ -116,10 +116,10 @@ cdef DMRA_mempool_attribute(property_type: type):
     return decorator
 
 
-cdef int DMRA_getattribute(
+cdef uint64_t DMRA_getattribute(
     cydriver.CUmemoryPool pool_handle, cydriver.CUmemPool_attribute attr_enum
 ):
-    cdef int value
+    cdef uint64_t value = 0
```
rwgk, line 122: This looks ok, I believe it'll work.

But at a minimum there should be a comment/warning somewhere, explaining that

* we know we're implicitly casting `int*` to `cuuint64_t*` for some attributes, and

* we decided to not worry about endianness here because the `int`s we know about today are only converted to bool.

Better IMO would be to just do it right, so that any changes to the enums in the future will lead to obvious failures that are most likely easily fixed.

For this one case it may not matter much, but in aggregate it'll makes a huge difference in long-term reliability and maintainability of cuda-python as a whole.

It's really straightforward to achieve, although it's significantly more code:

* https://chatgpt.com/share/691e57f4-17e0-8008-ade1-2e8d5fc590d9

This would give us complete safety. If there are changes to the API in the future, it will fail loudly and we'll know immediately that we need to review and adjust.

leofang: In terms of safety, I left a comment above (#1274) probably while you were leaving the review here 🙂

Andy-Jost:
Closing this since we have a better solution.

PR #1274, opened 2025-11-19
Cythonize device memory resource attributes

leofang:
## Description

There are a few pure Python constructions that are getting in the way, and those also forced us to use a single value type (was `int` and being changed to `uint64_t` in #1272). We have LLM now so we really should just spell it out once and tweak them quickly manually, like what we do for device attributes and program options. 

Before (`main`):
```python
In [6]: %timeit mr.attributes.reuse_allow_internal_dependencies
265 ns ± 6.03 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

After this PR:
```python
In [6]: %timeit mr.attributes.reuse_allow_internal_dependencies
116 ns ± 0.63 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

By construction this also fixes https://github.com/NVIDIA/cuda-python-private/issues/197.

## Checklist

### DIFF
--- a/cuda_core/cuda/core/experimental/_memory/_device_memory_resource.pyx
+++ b/cuda_core/cuda/core/experimental/_memory/_device_memory_resource.pyx
@@ -20,7 +20,6 @@ from cuda.core.experimental._utils.cuda_utils cimport (
 
 from dataclasses import dataclass
 from typing import Optional, TYPE_CHECKING
-import cython
 import platform  # no-cython-lint
 import uuid
 import weakref
@@ -49,8 +48,8 @@ cdef class DeviceMemoryResourceOptions:
         Maximum pool size. When set to 0, defaults to a system-dependent value.
         (Default to 0)
     """
-    ipc_enabled : cython.bint = False
-    max_size : cython.size_t = 0
+    ipc_enabled : bool = False
+    max_size : int = 0
 
 
 cdef class DeviceMemoryResourceAttributes:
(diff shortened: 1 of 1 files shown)


leofang:
/ok to test [74c6611](https://github.com/NVIDIA/cuda-python/pull/1274/commits/74c6611c73831326147c56a07608bd77ae1e6ff1)

cuda_core/cuda/core/experimental/_memory/_device_memory_resource.pyx
```diff
@@ -49,8 +48,8 @@ cdef class DeviceMemoryResourceOptions:
         Maximum pool size. When set to 0, defaults to a system-dependent value.
         (Default to 0)
     """
-    ipc_enabled : cython.bint = False
-    max_size : cython.size_t = 0
+    ipc_enabled : bool = False
+    max_size : int = 0
```
leofang, line 52: I took the opportunity to also fix the signature errors (we don't want users to see Cython here):
https://nvidia.github.io/cuda-python/cuda-core/0.4.2/generated/cuda.core.experimental.DeviceMemoryResourceOptions.html

cuda_core/cuda/core/experimental/_memory/_device_memory_resource.pyx
```diff
@@ -66,63 +65,70 @@ cdef class DeviceMemoryResourceAttributes:
         self._mr_weakref = mr
         return self
 
-    @DMRA_mempool_attribute(bool)
+    cdef int _getattribute(self, cydriver.CUmemPool_attribute attr_enum, void* value) except?-1:
```
leofang, line 68: The return type is made `int` so that we have a very simple error check `except?-1`. If we were to use `void` (which made sense back in Cython 0.29.x) then we would have to use `except*` here, which adds a slight overhead.

leofang:
Thanks guys! Merging.

PR #1242, opened 2025-11-16
Expand Windows test matrix to reproduce and fix nvbugs 5630448

leofang:
## Description

close #462
close #984
closes #985 
close #1024 
close #1241

expand Windows test matrix to cover
- 6 Python versions 
- 2 CUDA versions
- 3 driver modes
   - datacenter and Quadro cards cover TCC & MCDM
   - gamer cards cover WDDM ~~& MCDM~~
- 1 driver version
   - unlike on Linux, for now we can control what driver versions we install on Windows, so I'll work on this next

Using this PR, we can reproduce nvbugs 5630448 and I worked out a fix for it.

## Checklist

- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #1277, opened 2025-11-20
Fix VMM tests

### LINKED ISSUE #1276: Fix VMM test suite
Using `SECURITY_ATTRIBUTES` was wrong:
https://github.com/NVIDIA/cuda-python/blob/b1a6baf15994f67950194bc1c2a7704e2475c8ea/cuda_core/tests/test_memory.py#L318-L334
we should use `OBJECT_ATTRIBUTES` instead. See discussion starting here: https://github.com/NVIDIA/cuda-python/issues/1263#issuecomment-3555406115.

leofang:
## Description

follow-up of #1242
closes #1276

## Checklist

### DIFF
--- a/cuda_core/pyproject.toml
+++ b/cuda_core/pyproject.toml
@@ -51,7 +51,8 @@ cu12 = ["cuda-bindings[all]==12.*"]
 cu13 = ["cuda-bindings[all]==13.*"]
 
 [dependency-groups]
-test = ["cython>=3.2,<3.3", "setuptools", "pytest>=6.2.4"]
+# use cffi for VMM tests on Windows
+test = ["cython>=3.2,<3.3", "setuptools", "pytest>=6.2.4", "cffi; platform_system == \"Windows\""]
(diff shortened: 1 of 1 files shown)


leofang:
/ok to test [3dd1bc5](https://github.com/NVIDIA/cuda-python/pull/1277/commits/3dd1bc5a2799318ce501944478a98056c6fae27b)

leofang:
Ah... On the self-hosted Windows runners we don't have MSVC, so CFFI would not work 😞

kkraus14:
My 2c: having this `OBJECT_ATTRIBUTE` directly exposed to `cuda.core` doesn't make sense where I wonder if we can pull out the key functionality that is needed from it and support those more directly via args/kwargs to hide this as an implementation detail from users. If they need the low level functionality they can always reach for `cuda.bindings.driver`.

leofang:
Discussed during the meeting, we want to hide this for now and raise an exception asking users to reach out to us if they use `handle_type="win32"`. I'll prepare another PR to do this.

PR #1205, opened 2025-10-31
Memory refactor

Andy-Jost:
### Major refactoring of the memory package.

**Overview**
This PR refactors the `_memory.pyx` module into a dedicated package (`_memory/`) to address its growing size and complexity, which were hindering further development. The primary goals are to physically separate the code into more manageable submodules, simplify the internal logic, and enhance the overall structure, including the addition of `.pxd` headers for better Cython integration.

**Major Changes**

- Split `_memory.pyx` into submodules, the major ones being the following:
    - Buffers: `_buffer.*`
    - Device memory resources: `_dmr.*`
    - IPC (Inter-Process Communication): `_ipc.*`
    - Virtual memory management: `_vmm.*`
- Introduced Cython headers (`.pxd`) for public definitions to improve modularity and type safety.
- Refactored `DeviceMemoryResource` to isolate IPC-related code, reducing coupling.
- Simplified IPC implementation by adding an `IPCData` class to encapsulate relevant data members and eliminating a redundant `uuid` field.
- Streamlined the class hierarchy by removing unnecessary classes.
- Simplified the Cython interface for memory allocation and deallocation operations.

**Minor Improvements**

- Added `__all__` lists to modules for explicit control over exports.
- Extracted long implementation functions from class definitions to make classes more concise and readable.
- Renamed various private attributes and methods for consistency (e.g., `_handle` instead of `_mempool_handle`).
- Consolidated and alphabetized property definitions for better organization.
- Converted additional classes and functions to Cython for performance gains.

Issue #674, opened 2025-06-02
Add `cuda.core.NDBuffer`


leofang:
Moving the tracking issue to the public. Unlike the existing `Buffer` which is a 1D contiguous, untyped blob, we want to have a typed buffer that feels more natural to use as underlying memory of tensors/arrays. The requirements include
- Arbitrary dimension 
- Holds the underlying pointer
- Holds shape
- Holds strides
- Exposes same attributes as `Buffer` does

The only bit that needs discussion during design meetings: Is it owning or non-owning? 
- I think it should be owning (I was wrong in one of the meetings today) because `Buffer` is owning too. That'd make this `NDBuffer` look more like `mdarray` (https://github.com/NVIDIA/cccl/issues/2474; see also [Raft's counterpart](https://docs.rapids.ai/api/raft/stable/cpp_api/mdspan_mdarray/#mdarray-multi-dimensional-owning-container) and [p1684](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p1684r5.html)).
- If it's non-owning, then it's a view but we already have `StridedMemoryView` (#180)

leofang:
Capturing what Satya noted in an offline meeting:
> I think NDBuffer is useful for two reasons:
> 
> - While it's possible for users to compute the offset themselves, it's better to do that as part of the type so that each user is not burdened with it.
> -  We can offer efficient copy functions across layouts that users can take advantage of. While it's straightforward to write a copy function, it's hard to write a performant copy function.

kkraus14:
> * I think it should be owning (I was wrong in one of the meetings today) because `Buffer` is owning too. That'd make this `NDBuffer` look more like `mdarray` ([[FEA]: Provide `mdarray` cccl#2474](https://github.com/NVIDIA/cccl/issues/2474); see also [Raft's counterpart](https://docs.rapids.ai/api/raft/stable/cpp_api/mdspan_mdarray/#mdarray-multi-dimensional-owning-container) and [p1684](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p1684r5.html)).

The name `NDBuffer` is very confusing to me as I think of a Buffer as an untyped region of memory. I think using `ndarray` (more Python than `mdarray`?) is a more natural name, though we don't want to give the impression we will have all of the functionality that Python users expect of an array and we are purely a data container.

[...]

PR #1283, opened 2025-11-21
feat: Introduce StridedLayout, support wrapping external allocations in Buffer, add StridedMemoryView.from_buffer

stiepan:
## Description

1. Adds ccx.utils.StridedLayout for describing ndim-tensor layout (shape, strides in counts, itemsize).
    * Creating layout directly from shape/strides:
    ``StridedLayout(shape, strides, itemsize)``,  ``StridedLayout(a.shape, a.strides, a.itemsize, divide_strides=True)``
    * Creating dense layout with specific stride order (C|F or permutation) ``StridedLayout.dense(shape, itemsize, stride_order)``
    * Creating dense layout from another one: ``StridedLayout.dense_like`` and ``self.to_dense``
    * Implements properties: ndim, shape, volume, strides, strides_in_bytes, stride_order, min_offset, max_offset, is_contiguous(c|f|any), is_unique
    * required_size_in_bytes method for required allocation size
    * Stride manipulation helpers for reshaping, (un)squeezing, permuting, flattening, repacking (changing itemsize, as in viewing float tenor as complex one), broadcasting, slicing.

  From Python, StridedLayout is immutable, stride manipulation methods return a new instance. In Cython, to avoid temporary objects in a sequence of operations, layout manipulations methods can be run in place.
  Please take a look at the StridedLayout docs for more details and examples.

2.  Enables wrapping external allocation into Buffer (``Buffer.from_handle(ptr, owner=obj)``). The owner and memory resource cannot be specified together. The owner reference is kept until the Buffer is closed. Without the memory resource, Buffer now queries driver for host/device accessibility and device_id of the ptr.

3. StridedMemoryView uses now StridedLayout to represent the shape/strides. 
    * For DLPack/CAI imported tensors, the layout is lazily created if needed.
    * There's a new class method ``from_buffer(buffer, layout, optional dtype)`` to create SMV from Buffer and StridedLayout. For example to implement empty_like() method for numpy array, but allocated on a device, one could:
    ```
    def device_tensor_like(a : numpy.ndarray, device : ccx.Device) -> StridedMemoryView:
        a_view = StridedMemoryView(a, -1)
        # get the original layout of ``a`` and convert it to a dense layout
        # to avoid overallocating memory (e.g. if the ``a`` was sliced)
        layout = a_view.layout.to_dense()
        # get the required size in bytes to fit the tensor
        required_size = layout.required_size_in_bytes()
        # allocate the memory on the device
        device.set_current()
        mem = device.allocate(required_size)
        # create a view on the newly allocated device memory
        b_view = StridedMemoryView.from_buffer(mem, layout, a_view.dtype)
        return b_view
    ```
    * ~~The StridedMemoryView can be now exported via dlpack.~~ (delayed for later)
    * ~~The StridedMemoryView.copy_from, StridedMemoryView.copy_to allow copying data between views~~ (in a follow-up PR).
  

## Checklist


cuda_core/docs/source/api.rst
```diff
@@ -75,6 +75,7 @@ Utility functions
 
    args_viewable_as_strided_memory
 
-   :template: dataclass.rst
```
stiepan, line 78: * dataclass.rst does not render methods.
* class.rst omits cythonized properties

cyclass places attributes section just after the main class docstring. this way we can document the actual attributes at the end of the main docstring and they are followed by docstring of all the properties.

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -120,63 +138,229 @@ cdef class StridedMemoryView:
...
               + f"                  shape={self.shape},\n"
               + f"                  strides={self.strides},\n"
+              + f"                  itemsize={self.layout.itemsize},\n"
               + f"                  dtype={get_simple_repr(self.dtype)},\n"
               + f"                  device_id={self.device_id},\n"
               + f"                  is_device_accessible={self.is_device_accessible},\n"
               + f"                  readonly={self.readonly},\n"
               + f"                  exporting_obj={get_simple_repr(self.exporting_obj)})")
 
+    def __dlpack__(
```
stiepan: I followed the Buffer example here - no synchronization is done here.

cuda_core/cuda/core/experimental/_dlpack.pyx
```diff
@@ -91,14 +76,103 @@ cpdef object make_py_capsule(object buf, bint versioned):
...
+        dl_tensor.data = <void*>data_ptr
+
+        _setup_dl_tensor_device(dl_tensor, buf)
+        _setup_dl_tensor_layout(dl_tensor, buf, layout)
+        _setup_dl_tensor_dtype(dl_tensor, dtype)
+
+        return cpython.PyCapsule_New(tensor_ptr, capsule_name, pycapsule_deleter)
+    except:
+        cleanup(dlm_tensor)
+        cleanup_versioned(dlm_tensor_ver)
```
stiepan: those two are almost like the deleters, just with extra ifs to handle incomplete initialization in case of exceptions.

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -120,63 +138,229 @@ cdef class StridedMemoryView:
             dlm_tensor = <DLManagedTensor*>data
             dlm_tensor.deleter(dlm_tensor)
 
+    @classmethod
+    def from_buffer(
+        cls, buffer : Buffer, layout : StridedLayout,
+        dtype : numpy.dtype | None = None,
```
stiepan, line 164: A dtype is optional, in principle, for a copy or layout manipulation we don't need it. This way, we don't have to immediately worry about copying tensors with dtypes that are not in standard numpy.
The dtype is needed only for exporting the view with dlpack.

cuda_core/cuda/core/experimental/_layout.pxd
```diff
@@ -0,0 +1,680 @@
...
+        return self._py_shape
+
+    cdef inline tuple get_strides_tuple(StridedLayout self):
+        if not _has_valid_property(self, PROP_STRIDES):
+            if self.base.strides == NULL:
+                self._py_strides = None
+            else:
+                self._py_strides = cuda_utils.carray_integer_t_to_tuple(self.base.strides, self.base.ndim)
+            _mark_property_valid(self, PROP_STRIDES)
+        return self._py_strides
```
leofang, line 190: Previously, StridedMemoryView had a performance optimization that once Python shape and strides are generated, we cache the generated tuples. Can we keep that optimization?

stiepan: The layout caches those here, so SMV trying to access the layout.shape/strides triggers this creation here once and in subsequent calls it will get the cached tuple. Isn't that enough? (I.e. should we cache twice, here and in SMV?)

The cache here is a bit verbose: i.e. not a plain is None check but checking if a bit is set in a mask. 
1. this provides uniform caching pattern that includes things that cannot be None (like boolean contiguity flags) and things for which None is a valid cached value (strides=None).
2. for Cython API, where the layout can be mutated inplace, a single mask gives an easy way to invalidate all cached props with single mask = 0 assignment.

leofang: Thanks, Kamil, I misread!

cuda_core/cuda/core/experimental/_memoryview.pyx
leofang: I think it is needed because a user can check `view.dtype`?

cuda_core/cuda/core/experimental/_dlpack.pyx
```diff
@@ -33,6 +34,16 @@ cdef void deleter(DLManagedTensor* tensor) noexcept with gil:
     stdlib.free(tensor)
 
 
+cdef void cleanup(DLManagedTensor* tensor) noexcept with gil:
```
leofang: Q: Can we merge this with `deleter` above?

cuda_core/cuda/core/experimental/_dlpack.pyx
```diff
@@ -41,43 +52,17 @@ cdef void versioned_deleter(DLManagedTensorVersioned* tensor) noexcept with gil:
     stdlib.free(tensor)
 
 
-cpdef object make_py_capsule(object buf, bint versioned):
-    cdef DLManagedTensor* dlm_tensor
-    cdef DLManagedTensorVersioned* dlm_tensor_ver
-    cdef DLTensor* dl_tensor
-    cdef void* tensor_ptr
-    cdef const char* capsule_name
+cdef void cleanup_versioned(DLManagedTensorVersioned* tensor) noexcept with gil:
```
leofang: ditto, with `versioned_deleter` above

cuda_core/cuda/core/experimental/_dlpack.pyx
```diff
@@ -91,14 +76,103 @@ cpdef object make_py_capsule(object buf, bint versioned):
...
-    return cpython.PyCapsule_New(tensor_ptr, capsule_name, pycapsule_deleter)
+    try:
+        if versioned:
+            dlm_tensor_ver = <DLManagedTensorVersioned*>(
+                stdlib.malloc(sizeof(DLManagedTensorVersioned)))
+            dlm_tensor_ver.dl_tensor.shape = NULL
+            dl_tensor = &dlm_tensor_ver.dl_tensor
+            dlm_tensor_ver.version.major = DLPACK_MAJOR_VERSION
+            dlm_tensor_ver.version.minor = DLPACK_MINOR_VERSION
+            cpython.Py_INCREF(buf)
```
leofang, line 125: We need to call decref in the `except:` branch.

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
```diff
@@ -271,27 +287,82 @@ cdef class Buffer:
...
+        out.device_id = attrs[1]
+    return 0
+
+
+cdef inline _query_memory_attrs(uintptr_t ptr):
+    cdef tuple attrs = (
+        driver.CUpointer_attribute.CU_POINTER_ATTRIBUTE_MEMORY_TYPE,
+        driver.CUpointer_attribute.CU_POINTER_ATTRIBUTE_DEVICE_ORDINAL,
+    )
+    return driver.cuPointerGetAttributes(len(attrs), attrs, ptr)
```
leofang: TODO: cimport this from cydriver

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
```diff
@@ -271,27 +287,82 @@ cdef class Buffer:
...
+    raise_if_driver_error(ret)
+    memory_type = attrs[0]
+
+    if memory_type == 0:
+        # unregistered host pointer
+        out.is_host_accessible = True
+        out.is_device_accessible = False
+        out.device_id = -1
+    elif (
+        memory_type == driver.CUmemorytype.CU_MEMORYTYPE_HOST
```
leofang: Shouldn't this type set `out.is_device_accessible` to False?

cuda_core/cuda/core/experimental/include/layout.hpp
```diff
@@ -0,0 +1,54 @@
...
+#ifndef CUDA_CORE_LAYOUT_HPP
+#define CUDA_CORE_LAYOUT_HPP
+
+#include <cmath>
+#include <algorithm>
+#include <vector>
+#include <numeric>
+
+
+#define STRIDED_LAYOUT_MAX_NDIM 32
```
leofang: Q: is there any particular reason (apart from using 32-bit masks) we want to limit this to 32? There are libraries that use a higher limit, for example cuTENSOR (@yangcal would remember what the current limit is).

cuda_core/cuda/core/experimental/include/layout.hpp
```diff
@@ -0,0 +1,54 @@
...
+    return std::abs(x);
+}
+
+template <typename T>
+void _swap(T &a, T &b) noexcept
+{
+    std::swap(a, b);
+}
+
+inline void _order_from_strides(std::vector<int>& indices, const int64_t* shape, const int64_t* strides, int ndim)
```
leofang: For my own curiosity: Most of the code in this PR was written in Cython, is using pure C++ critical here?

cuda_core/cuda/core/experimental/_utils/cuda_utils.pxd
```diff
@@ -13,6 +13,11 @@ ctypedef fused supported_error_type:
     cydriver.CUresult
 
 
+ctypedef fused integer_t:
+    int64_t
+    int
```
leofang: nit: can we use int32_t instead?

cuda_core/cuda/core/experimental/include/layout.hpp
stiepan: Frankly, I needed some low-level sorting (vectors of integers), the ``std::sort`` with custom compare cb seemed like the straightforward tool, and when the first attempt to make it work with Cython did not work I just went for the header. :)

cuda_core/cuda/core/experimental/include/layout.hpp
stiepan: Initially, precompiling the copy kernels was on the table, so sticking to 32-ndim was more appealing. Now, I don't think there's anything preventing us from going to 64-dims. Especially, if there's really a need for that.

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
stiepan: 🙃 

I've actually had that this way initially, but seeing all the cdriver imports are gone from buffer, went along with the Python API. I can undig the previous variant.

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
stiepan: For the regular host allocation that has nothing to do with CUDA, the `cuPointerGetAttribute` fails with an error. The plural counterpart `cuPointerGetAttributes`, succeeds and returns `memory_type==0` (not enumerated in CUmemorytype). Calling this on the side, the CU_MEMORYTYPE_HOST is returned for pinned memory.

NB. The somehow relevant note from the driver docs:
Unlike [cuPointerGetAttribute](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__UNIFIED.html#group__CUDA__UNIFIED_1g0c28ed0aff848042bc0533110e45820c), this function will not return an error when the ptr encountered is not a valid CUDA pointer. Instead, the attributes are assigned default NULL values and CUDA_SUCCESS is returned.

cuda_core/cuda/core/experimental/_dlpack.pyx
stiepan: The cleaners helpers do so. I moved the cpython.Py_INCREF call next to assigning its address to manager_ctx, for the clearer "release" ptr semantics. If the manager_ctx is set, the DECREF needs to be called.

cuda_core/cuda/core/experimental/_memoryview.pyx
stiepan: Currently, if it is not specified, view.dtype is just None. For dlpack/CAI imported views, there's no change in behavior.

cuda_core/cuda/core/experimental/_dlpack.pyx
stiepan: As we're delaying the user-facing dlpack export for now, I reverted dlpack changes here not to inflate the PR or introduce unused code-paths.

I think the dlpack capsule creation needs to be fixed not to leak stuff if exception happens during its creation. I'll open another small PR for that.

cuda_core/cuda/core/experimental/_utils/cuda_utils.pxd
stiepan: done

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
leofang: Yeah, sorry. I think it's not "gone" gone, most likely @Andy-Jost found that we don't need many driver API calls in this file after the refactoring (#1205). But pointer attribute checks are in the hot path so we should cythonize it. 

In fact, I am trying to catch up with what @fbusato is doing in C++ (https://github.com/NVIDIA/cccl/pull/6733), which is an equivalent check (but for C++ mdspan instead of Python SMV).

cuda_core/cuda/core/experimental/_memoryview.pyx
leofang: I see. OK.

stiepan:
The dlpack fix moved out of this PR is here https://github.com/NVIDIA/cuda-python/pull/1291

cuda_core/cuda/core/experimental/include/layout.hpp
stiepan: [53568b5](https://github.com/NVIDIA/cuda-python/pull/1283/commits/53568b5f947fb38e797b25f77388a740b55c953f)

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
stiepan, line 1: @leofang,

NB.

```
driver.cuPointerGetAttributes(1, [driver.CUpointer_attribute.CU_POINTER_ATTRIBUTE_DEVICE_ORDINAL], 0)
```
returns
>(<CUresult.CUDA_SUCCESS: 0>, [4294967294])
while the actual value returned by the underlying API is -2. Looks like some unsigned/signed type mismatch.

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
stiepan: Thanks for the reference! Looking at the logic in cccl, I adjusted managed memory "discovery". I am not sure if we need to go into so much details as trying to get particular memory pool and check if readability flag is set there, I did not add this, but can adjust if needed.

In any case, I moved back to cydriver API and added tests with host/device/managed/pinned from cuda malloc and pinned from cuda register. 

For pinned memory, I am not 100% sure what happens on devices for which
[CU_DEVICE_ATTRIBUTE_CAN_USE_HOST_POINTER_FOR_REGISTERED_MEM](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1gge12b8a782bebe21b1ac0091bf9f4e2a3904c097da7a1891f9904b3e6a49e4cdd) is false. I.e. if one registers host memory with cuda host register and passes the original pointer, while the pointer to access on device is different. I.e. would it be still memory_type = 0 or memory_type = host and what would be a desired is_device_accessible value.

leofang:
Let's kick off CI

leofang:
/ok to test [3a904e7](https://github.com/NVIDIA/cuda-python/pull/1283/commits/3a904e7533eb12887063f81f798abba071e31663)

stiepan:
A single test case failed - testing pointer attributes for host memory "manually pinned" with cuMemHostRregister. It failed on pre-condition assert that the memory is not device accessible before it is registered. And it failed in a second of two cases testing this. The test did not clean-up properly - it was missing unregister call. I am guessing we ended up with the same pointer in the second case. The https://github.com/NVIDIA/cuda-python/pull/1283/commits/38ddb36acfcd093cda7f21c02b7f4c5f823121fe should fix that.

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -433,3 +593,76 @@ def args_viewable_as_strided_memory(tuple arg_indices):
...
         return wrapped_func
     return wrapped_func_with_indices
+
+
+cdef inline StridedLayout layout_from_dlpack(DLTensor* dl_tensor):
+    cdef StridedLayout layout = StridedLayout.__new__(StridedLayout)
+    cdef int nbits = dl_tensor.dtype.bits
+    cdef int itemsize = nbits >> 3
+    if (itemsize << 3) != nbits:
+        raise ValueError("dl_tensor.dtype.bits must be a multiple of 8")
```
kkraus14, line 646: How should we handle things like 6-bit and 4-bit floats? https://github.com/dmlc/dlpack/blob/main/include/dlpack/dlpack.h#L171-L181

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -433,3 +593,76 @@ def args_viewable_as_strided_memory(tuple arg_indices):
...
+    # alone is not enough if the layout can map to negative offsets, i.e.:
+    # the valid range is not the [ptr, ptr + size - 1], but
+    # [ptr - offset, ptr + size - offset - 1]. The offset is not reported
+    # by the packages.
+    if is_allocated and buffer.size < layout.get_required_size_in_bytes():
+        raise ValueError(
+            f"Buffer size is too small for the layout. "
+            f"Expected at least {layout.get_required_size_in_bytes()} bytes, "
+            f"got {buffer.size} bytes."
+        )
```
kkraus14: These checks seem somewhat expensive to do on the path of getting the ptr to the data. Should we validate these only at construction time and avoid them in a hot path like this?

stiepan: Those checks happen now for new SMV created with `from_buffer` or `view` methods. We could limit those to from_buffer only or get rid of them completely. I don't have strong preference either way here.

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
```diff
@@ -267,20 +286,93 @@ cdef class Buffer:
...
+    vals[1] = <uintptr_t><void*>&is_managed
+    vals[2] = <uintptr_t><void*>&device_id
+
+    cdef cydriver.CUresult ret
+    ret = cydriver.cuPointerGetAttributes(3, attrs, <void**>vals, ptr)
+    if ret == cydriver.CUresult.CUDA_ERROR_NOT_INITIALIZED:
+        with cython.gil:
+            # Device class handles the cuInit call internally
+            from cuda.core.experimental import Device
+            Device()
```
cpcloud, line 448: Can we call `cuInit` if that's what we're after, instead of calling `Device`?

cuda_core/tests/helpers/layout.py
```diff
@@ -0,0 +1,162 @@
...
+    Whether to initialize the dense layout in C or F order.
+    For C, the strides can be explicit or implicit (None).
+    """
+
+    C = "C"
+    IMPLICIT_C = "implicit_c"
+    F = "F"
+
+
+class _S:
```
cpcloud: We're already importing `numpy` where this is being used, can't we use `np.s_` for this instead of duplicating that functionality solely for testing?

cuda_core/cuda/core/experimental/_memoryview.pyx
stiepan: Good question, I am not sure, really. AFAIK, we don't support it now anyway, the dtype parsing won't accept it. 

Playing with https://github.com/jax-ml/ml_dtypes I see that numpy: 
1. does not allow exporting such types through dlpack
2. does not pack those - one element corresponds to one byte offset. 

We could work around it in different ways, just looking at the landscape now I am not sure what's the desired behavior.  E.g. would a dlpack dtype for fractional fps have nbits * lanes divisible by 8 or the numpy behavior is a common thing - we should round the itemsize to 1?

cuda_core/tests/helpers/layout.py
stiepan: Cool, I wasn't aware of this utility.

cuda_core/cuda/core/experimental/_memoryview.pyx
kkraus14: My 2c:
* If Buffer isn't validating ownership, neither should SMV. If we wanted to require an owner we should do so at the Buffer level. Let's remove the `is_allocated` check entirely?
* Given we're allowing effectively slicing a buffer at SMV construction time using offset and size, I think the layout size validation makes sense at `from_buffer` and `view` construction time.

cuda_core/cuda/core/experimental/_memoryview.pyx
kkraus14: cc @seberg for visibility and any thoughts

Given JAX and `ml_dtypes` uses unpacked layouts and generally we want to support packed layouts that sounds like a bit of an impedance mismatch.

seberg: NumPy's problem with packed layout is that the array object's ABI is all byte-strided/pointer based. There might be paths to address this.[^1]
One could create e.g. a `2xfloat4_e2m1fn` with explicit functions to pack and unpack. Not sure that a dtype helps a lot if you need to unpack for some or even most operations (compared to packing/unpacking with a custom converter function).

Enabling `ml_dtypes` to import/export for DLPack in NumPy, we can do, I think. Export would be easy, even a `__dlpack_dtype__` would work. Unfortunately, a proper import implementation needs some form of registration so it probably not very quick.

FWIW, DLPack can allow `bitsize == 8` to represent the padded versions. I don't think there was a particular reason not to just spell that out.
But by itself that doesn't actually quite help us (not without NumPy allowing export or some hacks here to make it work).

[^1]: I'll think about bit-sized a bit more. Even without a new type, it may be plausible to assume nobody really accesses an array that they don't know the dtype of (i.e. a bit-sized dtype) -- i.e. it's a bit of an ABI break, but likely doesn't really matter.
A normal 64bit pointer is probably enough to include a bit-offset (or element size?) than things may actually work with surprisingly few tweaks (I assume 32bit systems may not enough space in practice to store any offset in the pointer, but that seems OK).
(Unless something about passing `char *`/`void *` breaks this.)

stiepan: Thanks @seberg for detailed overview.

So when receiving fp4/fp6 tensor through dlpack we can expect 3 cases:

1. The "padded/unpacked" tensor (what numpy + ml_dtypes does now). We can represent them with no changes here, as soon as we start accepting the dtype. In dlpack, setting bitlen to 8 for those seems to be a bit of grey area, but if it's there, the StridedLayout won't be a hindrance. It looks like copying those does not require anything special.

2. The packed, vectorized dtype (2xfloat4_e2m1fn, 4xfloat6_e2m3fn?). We just need to take lanes into account, i.e. ``itemsize=nbits*lanes//8; assert itemsize * 8 == nbits*lanes``. I omitted the lanes, because we forbid ``lanes > 1`` when creating the numpy.dtype, but this is a simple adjustment. Again, it looks like copying those does not require anything special.

3. Receiving the packed, non-vectorized dtype. 
Little effort workaround would be to enforce scenario 2. I guess the actual tensors must have nice strides and round number of elements to make them performant, so this should be possible in any real-life case. We could either require users to pass views like in scenario 2 or stipulate that we do such a conversion and repack the layout automatically. If really necessary we could special-case the SMV to keep the original shape and strides and use the repacked StridedLayout only internally if a copy is requested.

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
stiepan: Device has a lock around the cuInit call, so I wanted to stick to that single section handling the call. I guess calling cuInit more than once wouldn't hurt us either. Or I could move that part from ``Device.__new__`` into separate utility. I just though this is one-time thingy, trying to recover from an error, so being a bit lavish about the import and Device call is not that bad. What do you think?

cuda_core/cuda/core/experimental/_memoryview.pyx
stiepan: I removed the ``.owner`` check.

cuda_core/tests/helpers/layout.py
stiepan: Replaced it with numpy.s_

cuda_core/cuda/core/experimental/_memoryview.pyx
stiepan: I adjusted the itemsize computation to take lanes into account, this should cover 1 and 2 with no extra changes once we start to recognize the dtype.

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -615,20 +615,13 @@ cdef StridedLayout layout_from_cai(object metadata):
...
-    # [ptr - offset, ptr + size - offset - 1]. The offset is not reported
-    # by the packages.
+    cdef bint is_allocated = buffer.memory_resource is not None
+    # Check the layout's offset range [min_offset, max_offset] fits
+    # within the [0, buffer.size - 1] range.
+    # The required_size_in_bytes fails if min_offset < 0.
+    # NB. For external memory, both positive and negative offsets can be valid,
+    # but for a proper check we'd need to know both size and data offset,
+    # while neither is reported by the packages.
     if is_allocated and buffer.size < layout.get_required_size_in_bytes():
```
kkraus14: Do we need this check at the time of getting the ptr as opposed to at the time of constructon as well?

Andy-Jost:
## Issue Analysis

The test `test_buffer_external_pinned_registered` is failing in the second parametrized case (`change_device=True`) because Python's memory allocator may reuse the same memory location for the new `(ctypes.c_byte * 20)()` object across test runs. 

**Root Cause:**
When the first parametrized run (`change_device=False`) completes, it unregisters the pointer in the `finally` block. However, when the second run (`change_device=True`) starts, if Python reuses the same memory location, `ctypes.addressof(a)` returns the same pointer value. If there's any timing issue or if the unregister didn't complete properly, the pointer may still appear registered, causing the first assertion `assert not buffer.is_device_accessible` to fail.

**Solution:**
Ensure the pointer is unregistered at the start of each test run, before checking its attributes. This guarantees a clean state regardless of memory reuse.

cuda_core/cuda/core/experimental/_memoryview.pyx
stiepan: Right, I moved the validation to ``view_buffer_strided`` - the common utility used by from_buffer and view when creating a new SMV.

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
leofang: Thanks for reporting @stiepan! Tracked in https://github.com/NVIDIA/cuda-python/issues/1336.

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
leofang: my 2c is let's keep this for now, I don't have clever thoughts here 😅

leofang:
@stiepan would you mind resolving the conflicts?

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -61,18 +71,12 @@ cdef class StridedMemoryView:
...
-    obj : Any
-        Any objects that supports either DLPack (up to v1.0) or CUDA Array
-        Interface (v3).
-    stream_ptr: int
-        The pointer address (as Python `int`) to the **consumer** stream.
-        Stream ordering will be properly established unless ``-1`` is passed.
     """
     cdef readonly:
-        intptr_t ptr
+        uintptr_t ptr
```
leofang: Q: Not sure if I missed this, but it seems we went great lengthy in changing this to unsigned. What's the reason...? 

It was made as signed following `Buffer.handle` because I anticipate people could do pointer arithmetic and usually we can't assume the pointer offsets would be positive.

stiepan: @leofang, thanks, good catch.

I haphazardly chnaged it here following the change in Buffer.ptr, making it at best brittle wrt ptr arithemtic. Reverted.

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
```diff
@@ -277,20 +286,93 @@ cdef class Buffer:
...
+    vals[0] = <uintptr_t><void*>&memory_type
+    vals[1] = <uintptr_t><void*>&is_managed
+    vals[2] = <uintptr_t><void*>&device_id
+
+    cdef cydriver.CUresult ret
+    ret = cydriver.cuPointerGetAttributes(3, attrs, <void**>vals, ptr)
+    if ret == cydriver.CUresult.CUDA_ERROR_NOT_INITIALIZED:
+        with cython.gil:
+            # Device class handles the cuInit call internally
+            from cuda.core.experimental import Device
```
kkraus14: Can this import be moved to the module level instead of in this inline function? Unfortunately these types of imports aren't free even when cached.

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
```diff
@@ -277,20 +286,93 @@ cdef class Buffer:
...
+        or memory_type == cydriver.CUmemorytype.CU_MEMORYTYPE_HOST
+    ):
+        # For pinned memory allocated with cudaMallocHost or paged-locked
+        # with cudaHostRegister, the memory_type is
+        # cydriver.CUmemorytype.CU_MEMORYTYPE_HOST.
+        # TODO(ktokarski): In some cases, the registered memory requires
+        # using different ptr for device and host, we could check
+        # cuMemHostGetDevicePointer and
+        # CU_DEVICE_ATTRIBUTE_CAN_USE_HOST_POINTER_FOR_REGISTERED_MEM
+        # to double check the device accessibility.
```
kkraus14, line 420: Do you happen to know what cases these are? This used to be the case with non-unified addressing but I don't think any platforms that CUDA supports are non-unified addressing these days.

stiepan: I did not find a comprehensive list, but digging a bit I learnt one notable exception for modern gpus: running on WSL. Indeed, trying to access cudahostregistered ptr on WSL fails (if the memory is allocated with cuda from the start, using the same pointer is fine).

```
import cuda.core.experimental as ccx
from cuda.bindings import runtime
import cupy as cp
import numpy as np

d = ccx.Device()
d.set_current()

def query_memory_attrs(ptr):
    attrs = (
        driver.CUpointer_attribute.CU_POINTER_ATTRIBUTE_MEMORY_TYPE,
        driver.CUpointer_attribute.CU_POINTER_ATTRIBUTE_DEVICE_ORDINAL,
        driver.CUpointer_attribute.CU_POINTER_ATTRIBUTE_IS_MANAGED,
    )
    ret, attr = driver.cuPointerGetAttributes(len(attrs), attrs, ptr)
    assert ret == 0
    return attr

a_np = np.empty(5, dtype=np.int32)
cpu_ptr = a_np.ctypes.data
ret, = runtime.cudaHostRegister(cpu_ptr, 20, 0)
assert ret == 0
assert query_memory_attrs(cpu_ptr)[0] == driver.CUmemorytype.CU_MEMORYTYPE_HOST
ret, attr = runtime.cudaPointerGetAttributes(cpu_ptr)
assert ret == 0
print(attr.devicePointer == cpu_ptr)
# On WSL, accessing cpu_ptr instead of attr.devicePointer fails
um = cp.cuda.UnownedMemory(cpu_ptr, 20, a_np, 0)
mem = cp.cuda.MemoryPointer(um, 0)
a_cp = cp.ndarray(shape=(5,), dtype=cp.int32, memptr=mem)
a_cp[:] = 1
print(a_np)
print(a_cp)
```

stiepan: At the same time, driver's cuPointerGetAttributes still reports that pointer as CU_MEMORYTYPE_HOST.

stiepan: So, what should be the meaning of the is_device_accessible, is_host_accessible in this case? 

1. Should we check the device attribute and, if the attribute is 0, follow-up by retreiving host_ptr, device_ptr and set is_host_accessible=host_ptr==ptr, is_device_accessible=device_ptr==ptr?
2. Or expect user to pass the correct pointer in a correct context, i.e. if the buffer is to be consumed on the gpu, user is expected to pass the device ptr?
3. Or (not a fan) have buffer.device_ptr, buffer.host_ptr attributes?

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
stiepan: done

leofang:
/ok to test [639ee5f](https://github.com/NVIDIA/cuda-python/pull/1283/commits/639ee5f8ded8a71041128de63f86652d53f2b84f)

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
leofang: Let's table this discussion for now. I'll create an issue to track this. I think the strided layout itself is already big enough that we want to keep the scope limited.

cuda_core/cuda/core/experimental/_memoryview.pyx
leofang: Ah, it's probably a miss then. We made sure `Buffer.handle` is a `intptr_t` in the beginning. Most likely it was changed recently to `uintptr_t` and I failed to capture it. Let me create an issue.

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
kkraus14: Given WSL is a relatively common platform, it would be nice to do the `CU_DEVICE_ATTRIBUTE_CAN_USE_HOST_POINTER_FOR_REGISTERED_MEM` and if it's `false` then we at least set the `is_device_accessible` to `false`.

stiepan: Simply setting false would be too strict, ptr coming from cudamallochost memory works just fine on both device and host ends.

stiepan:
> /ok to test [639ee5f](https://github.com/NVIDIA/cuda-python/pull/1283/commits/639ee5f8ded8a71041128de63f86652d53f2b84f)

>cudaErrorNoDevice: no CUDA-capable device is detected

when collecting cases in test_linkier, test_program, and test_utils. Could it be unrelated?

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
kkraus14: If we checked `memory_type == cydriver.CUmemorytype.CU_MEMORYTYPE_HOST` and `CU_DEVICE_ATTRIBUTE_CAN_USE_HOST_POINTER_FOR_REGISTERED_MEM`, I would assume the first check would return `True`, and the second check would return `True` if an allocation made from `cudaMallocHost` can use the same ptr for device and host, so it would still return `True` for `is_device_accessible`?

stiepan: On WSL, the `CU_DEVICE_ATTRIBUTE_CAN_USE_HOST_POINTER_FOR_REGISTERED_MEM` is False.

If a ptr comes from `cudaMallocHost` or was passed to `cudaHostRegister`, the  `memory_type == cydriver.CUmemorytype.CU_MEMORYTYPE_HOST` is True.

For `cudaMallocHost`, the ptr is truely device and host accessible, only the `cudaHostRegister-ed` one is troublesome - even though the memory type is CU_MEMORYTYPE_HOST, it cannot be used to access the mem from device. So my point was that if we were to say `is_device_accessible is False` whenever `CU_DEVICE_ATTRIBUTE_CAN_USE_HOST_POINTER_FOR_REGISTERED_MEM` is False, we would break otherwise valid `cudaMallocHost` usages.

```
import cuda.core.experimental as ccx
from cuda.bindings import runtime
import cupy as cp
import numpy as np
import ctypes


def query_memory_attrs(ptr):
    attrs = (
        driver.CUpointer_attribute.CU_POINTER_ATTRIBUTE_MEMORY_TYPE,
        driver.CUpointer_attribute.CU_POINTER_ATTRIBUTE_DEVICE_ORDINAL,
        driver.CUpointer_attribute.CU_POINTER_ATTRIBUTE_IS_MANAGED,
    )
    ret, attr = driver.cuPointerGetAttributes(len(attrs), attrs, ptr)
    assert ret == 0
    return attr

def as_numpy(ptr, shape, dtype):
    size = np.prod(shape) * dtype.itemsize
    return np.ndarray(
        shape=shape,
        dtype=dtype,
        buffer=memoryview((ctypes.c_char * size).from_address(ptr))
    )

def as_cupy(ptr, shape, dtype):
    size = np.prod(shape) * dtype.itemsize
    um = cp.cuda.UnownedMemory(ptr, size, owner=None, device_id=0)
    mem = cp.cuda.MemoryPointer(um, 0)
    return cp.ndarray(shape=shape, dtype=dtype, memptr=mem)

d = ccx.Device()
d.set_current()

# On WSL this is 0
print(driver.cuDeviceGetAttribute(driver.CUdevice_attribute.CU_DEVICE_ATTRIBUTE_CAN_USE_HOST_POINTER_FOR_REGISTERED_MEM, 0))

shape = (5,)
dtype = np.dtype(np.int32)
size = np.prod(shape) * dtype.itemsize

# But this works
l = ccx.LegacyPinnedMemoryResource()
alloc_mem = l.allocate(np.prod(shape) * dtype.itemsize)
alloc_ptr = int(alloc_mem.handle)
# the pinned ptr is CU_POINTER_ATTRIBUTE_MEMORY_TYPE, as expected, 1 (aka CU_MEMORYTYPE_HOST) 
assert query_memory_attrs(alloc_ptr)[0] == driver.CUmemorytype.CU_MEMORYTYPE_HOST

a_np = as_numpy(alloc_ptr, shape, dtype)
a_cp = as_cupy(alloc_ptr, shape, dtype)
a_np[:] = 1
print(a_np)
print(a_cp)

# The problem is when we register the memory
a_np = np.empty(shape, dtype=dtype)
cpu_ptr = a_np.ctypes.data
ret, = runtime.cudaHostRegister(cpu_ptr, size, 0)
assert ret == 0
assert query_memory_attrs(cpu_ptr)[0] == driver.CUmemorytype.CU_MEMORYTYPE_HOST
reg_np = as_numpy(cpu_ptr, shape, dtype)
reg_cp = as_cupy(cpu_ptr, shape, dtype)
reg_np[:] = 2
print(reg_np)
# Here we end up with invalid access
print(reg_cp)
```

cuda_core/cuda/core/experimental/_memoryview.pyx
leofang: > Let me create an issue.

https://github.com/NVIDIA/cuda-python/issues/1342

leofang:
> > cudaErrorNoDevice: no CUDA-capable device is detected
> 
> when collecting cases in test_linkier, test_program, and test_utils. Could it be unrelated?

Yes. It's likely a known glitch. Pinged you in an internal gha-runner thread. We can consider the CI is green.

@cpcloud and I discussed, we are still reviewing the PR but we'd like to get it merged tomorrow the latest.

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
kkraus14: Just to confirm my understanding is correct, on WSL `CU_DEVICE_ATTRIBUTE_CAN_USE_HOST_POINTER_FOR_REGISTERED_MEM` is `False` for both `cudaMallocHost` memory as well as `cudaHostRegister` memory, but the ptr returned from `cudaMallocHost` is in fact usable in device code while the ptr used for `cudaHostRegister` is not usable in device code?

kkraus14: Alternatively, we could query the `CU_POINTER_ATTRIBUTE_DEVICE_POINTER` and `CU_POINTER_ATTRIBUTE_HOST_POINTER` attributes. On my local WSL setup it yields:
* Same ptr for pinned host memory
* Same ptr for managed memory
* Different ptrs for device memory (0 for the `CU_POINTER_ATTRIBUTE_HOST_POINTER` attribute, as expected)
* Different ptrs for registered host memory (neither are 0)

Our logic could be that we return `is_device_accessible == True` only when the ptr is equal to the ptr returned from `CU_POINTER_ATTRIBUTE_DEVICE_POINTER` and `is_host_accessible == True` when the ptr is equal to the ptr returned from `CU_POINTER_ATTRIBUTE_HOST_POINTER`.

That being said, querying these attributes are expensive and not sure if we want to pay this penalty...

stiepan: >Just to confirm my understanding is correct, on WSL CU_DEVICE_ATTRIBUTE_CAN_USE_HOST_POINTER_FOR_REGISTERED_MEM is False

That's right.

>for both cudaMallocHost memory as well as cudaHostRegister memory

`CU_DEVICE_ATTRIBUTE_CAN_USE_HOST_POINTER_FOR_REGISTERED_MEM ` is a device attribute, not a memory ptr attribute

>the ptr returned from cudaMallocHost is in fact usable in device code while the ptr used for cudaHostRegister is not usable in device code

That's right. And using memory type is not enough to distinguish the two.

>Alternatively, we could query the CU_POINTER_ATTRIBUTE_DEVICE_POINTER and CU_POINTER_ATTRIBUTE_HOST_POINTER attributes.
 
Yeah, I've been thinking about similar approach. According to cuMemHostGetDevicePointer, there is still a catch, though. In some cases, the `device_ptr != host_ptr`, even though the memory can be accessed through the host pointer from the device. 🥲 If I read the docs right (and assuming that's the only edge-case), we'd need to boundle it with the `CU_DEVICE_ATTRIBUTE_CAN_USE_HOST_POINTER_FOR_REGISTERED_MEM ` check, so that the `CU_DEVICE_ATTRIBUTE_CAN_USE_HOST_POINTER_FOR_REGISTERED_MEM or ptr == device_ptr` would be accurate enough.

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
```diff
@@ -346,20 +365,92 @@ cdef class Buffer:
...
+            s = Stream_accept(stream) if stream is not None else self._alloc_stream
+            self._memory_resource.deallocate(self._ptr, self._size, s)
         self._ptr = 0
         self._memory_resource = None
+        self._owner = None
         self._ptr_obj = None
         self._alloc_stream = None
 
 
+cdef Buffer_init_mem_attrs(Buffer self):
```
oleksandr-pavlyk, line 389: Nit: `cdef void Buffer_init_mem_attrs(Buffer self):`

stiepan: Hmm, I recall some weird issues with `void` ret type when it comes to exception propagation with cython. Won't this require except* clause?

leofang: Yeah it'd have to be 
```cython
cef int Buffer_init_mem_attrs(Buffer self) except?-1:
    ...
    return 0
```
if we want to do this and gain a bit of perf. I am fine with the status quo.

leofang:
/ok to test [66fc6e8](https://github.com/NVIDIA/cuda-python/pull/1283/commits/66fc6e8027e49883c361fdb14d4f95ac061e5b53)

leofang (approved):

leofang:
Many thanks for @stiepan for migrating the strided layout to cuda-core and everyone for helping review!!! 🔥

leofang:
Unfortunately, as per https://github.com/NVIDIA/cuda-python/issues/674#issuecomment-3656821898 we'll have to revert this PR, see #1370.

PR #1284, opened 2025-11-24
Prerelease NVML bindings

mdboom:


cuda_bindings/cuda/bindings/_internal/nvml_windows.pyx
```diff
@@ -0,0 +1,6000 @@
...
+        if __cuDriverGetVersion == NULL:
+            __cuDriverGetVersion = <void*><intptr_t>win32api.GetProcAddress(handle, 'cuDriverGetVersion')
+            if __cuDriverGetVersion == NULL:
+                raise RuntimeError('something went wrong')
+        err = (<int (*)(int*) noexcept nogil>__cuDriverGetVersion)(&driver_ver)
+        if err != 0:
+            raise RuntimeError('something went wrong')
+
+        # Load library
+        handle = load_nvidia_dynamic_lib("nvml")._handle_uint
```
rwgk: @kkraus14 What's your opinion about supporting this in pathfinder?

Under both Linux and Windows we can rely on the library to be available through the system search, usually:

* `C:\Windows\System32\nvml.dll`
* `/lib/x86_64-linux-gnu/libnvidia-ml.so.1`
* `/usr/lib/wsl/lib/libnvidia-ml.so.1`

But we don't even have to know these paths, we can go straight to the system search, just based on the filenames.

It would be pretty easy to support `nvml` in pathfinder, but then again, this could also be handled via a couple simple helper functions that live in cybind.

kkraus14: We should add support for both `nvml` and `libcuda` in pathfinder, there's other implications for `libcuda` in relation to `cuda-compat`, but adding `nvml` would be a great addition and should be straightforward, I agree with you.

leofang: It should be trivial because both libcuda/nvcuda and nvml are part of the driver and cannot be installed to site-packages. The former can come from cuda-compat but it must be in the system path, so just let the dynamic linker do the job and no need to search in multiple places.

rwgk: Thanks! I just created #1288 for this work.

leofang:
Let's get #1244 merged first to reduce the diff (from cufile).

mdboom:
> Let's get #1244 merged first to reduce the diff (from cufile).

Yes, if we want to merge #1244 before 13.1 (I'm not sure that we do).

cuda_bindings/tests/nvml/test_pynvml.py
```diff
@@ -0,0 +1,283 @@
...
+import pytest
+from cuda.bindings import _nvml as nvml
+from looseversion import LooseVersion
+
+from . import util
+
+XFAIL_LEGACY_NVLINK_MSG = "Legacy NVLink test expected to fail."
+
+
+def test_system_get_nvml_version(nvml_init):
```
rwgk, line 17: It'd be great to avoid the new `looseversion` dependency just for these two trivial use cases.

E.g.:

```python
@pytest.mark.parametrize(
    "get_version",
    [
        nvml.system_get_nvml_version,
        nvml.system_get_driver_version,
    ],
)
def test_system_get_version(nvml_init, get_version):
    vsn = get_version()
    assert isinstance(vsn, str)
    assert tuple(int(p) for p in vsn.split(".")[:2]) > (0, 0)
```

cuda_bindings/tests/nvml/conftest.py
```diff
@@ -0,0 +1,139 @@
...
+            assert 1 <= arch_id <= 12, "Architecture not found, presumably something newer"
+            # arch_name = (utils.nvml_architecture_name.get(archID)).split("_")[-1]
+            # archName = archName[0] + archName[1:].lower()
+
+            BoardCfg = namedtuple("BoardCfg", "name, ids_arr")
+            board = BoardCfg(name, ids_arr=[(pci_info.pci_device_id, pci_info.pci_sub_system_id)])
+
+            try:
+                serial = nvml.device_get_serial(dev)
+            except:
```
rwgk: Does
```
            except nvml.NvmlError:
```
work here?

cuda_bindings/tests/nvml/conftest.py
```diff
@@ -0,0 +1,139 @@
...
+            try:
+                dev = nvml.device_get_handle_by_index_v2(i)
+            except nvml.NoPermissionError:
+                continue
+            pci_info = nvml.device_get_pci_info_v3(dev)
+
+            name = nvml.device_get_name(dev)
+            # Get architecture name ex: Ampere, Kepler
+            arch_id = nvml.device_get_architecture(dev)
+            # 1 = NVML_DEVICE_ARCH_KEPLER and 12 = NVML_DEVICE_ARCH_COUNT
```
rwgk: I couldn't find `NVML_DEVICE_ARCH_COUNT` in CTK 12.9, 13.0, 13.1, or with Google.

I see there are only `NVML_DEVICE_ARCH-*` preprocessor defines, and there is no `COUNT`-kind sentinel.

Do we actually need the `assert` for the maximum here?

leofang:
Backported in #1349.

PR #1285, opened 2025-11-24
Replace OS sleep with GPU nanosleep kernel in event timing test

rwgk:
This change replaces the timing-based event test's use of `time.sleep()` with a GPU-side nanosleep kernel, eliminating flakiness from OS/driver timing characteristics while maintaining deterministic test behavior.

## Changes

- Added `NanosleepKernel` helper that uses `__nanosleep()` to create a guaranteed 20 ms GPU-side delay
- Updated `test_timing_success` to use the nanosleep kernel instead of `time.sleep()`
- Removed OS-specific timing tolerance logic (Windows/WSL special cases)
- Simplified assertions to check for finite elapsed time and a minimum threshold (>10ms)

## Benefits

- **Deterministic**: GPU-side delay is consistent across platforms, eliminating flakiness on Windows/WDDM and WSL
- **Simpler**: Removes platform-specific tolerance calculations and OS timing dependencies
- **Reliable**: Tests `Event.__sub__` functionality without depending on OS timer resolution or driver scheduling behavior

This PR makes PR #1279 obsolete.

### DIFF
--- a/cuda_core/tests/helpers/nanosleep_kernel.py
+++ b/cuda_core/tests/helpers/nanosleep_kernel.py
@@ -14,7 +14,7 @@ class NanosleepKernel:
     Manages a kernel that sleeps for a specified duration using clock64().
     """
 
-    def __init__(self, device, sleep_duration_ms=20):
+    def __init__(self, device, sleep_duration_ms: int = 20):
         """
         Initialize the nanosleep kernel.
 
@@ -22,14 +22,13 @@ class NanosleepKernel:
             device: CUDA device to compile the kernel for
             sleep_duration_ms: Duration to sleep in milliseconds (default: 20)
         """
-        # clock_rate is in kHz
-        sleep_cycles = int(sleep_duration_ms * device.properties.clock_rate) + 1
         code = f"""
         extern "C"
         __global__ void nanosleep_kernel() {{
-            long long int start = clock64();
-            while (clock64() - start < {sleep_cycles}) {{
-                __nanosleep(1000000); // 1 ms yield to avoid 100% spin
+            // The maximum sleep duration is approximately 1 millisecond.
+            unsigned int one_ms = 1000000U;
+            for (unsigned int i = 0; i < {sleep_duration_ms}; ++i) {{
+                __nanosleep(one_ms);
             }}
         }}
         """


leofang:
LGTM but it'd be nice for @kkraus14 to take another look.

cuda_core/tests/helpers/__init__.py
```diff
@@ -5,7 +5,7 @@
 import pathlib
 import sys
 
-CUDA_PATH = os.environ.get("CUDA_PATH")
+CUDA_PATH = os.environ.get("CUDA_HOME", os.environ.get("CUDA_PATH"))
```
rparolin: Should we have this somewhere central so all tests get this automatically?
Also, should this be in a different PR as it doesn't appear related to fixing the invalid timing assumptions in the unit test?

cuda_core/tests/test_event.py
```diff
@@ -13,35 +13,30 @@
...
+    # Sanity check: cuEventElapsedTime should always return a finite float for two completed
+    # events. This guards against unexpected driver/HW anomalies (e.g. NaN or inf) or general
+    # undefined behavior, without asserting anything about the magnitude of the measured time.
+    assert math.isfinite(delta_ms)
+    # cudaEventElapsedTime() has a documented resolution of ~0.5 microseconds. With two back-to-back
+    # event records on the same stream, their timestamps may fall into the same tick, producing
+    # an elapsed time of exactly 0.0 ms. The only reliable way to force delta_ms > 0 would be to
+    # insert real GPU work (e.g. a __nanosleep kernel), which is more complexity than this test
+    # should depend on. For correctness we only assert that the elapsed time is non-negative.
+    assert delta_ms >= 0
```
rparolin: You could introduce a side effect from that kernel that you could validate against to ensure the kernel code ran.  For example, have the kernel write `42` to a variable and validate the value was set in the test.

cuda_core/tests/test_event.py
```diff
@@ -13,35 +13,30 @@
...
-    else:
-        # Most modern Linux kernels have a default timer resolution of 1 ms.
-        generous_tolerance = 20
-    assert delay_ms - generous_tolerance <= elapsed_time_ms < delay_ms + generous_tolerance
+    delta_ms = e2 - e1
+    assert isinstance(delta_ms, float)
+    # Sanity check: cuEventElapsedTime should always return a finite float for two completed
+    # events. This guards against unexpected driver/HW anomalies (e.g. NaN or inf) or general
+    # undefined behavior, without asserting anything about the magnitude of the measured time.
+    assert math.isfinite(delta_ms)
```
rparolin: This is great to check but in a CI run it won't tell the engineer which even originated the problem.

```
# Verify that each timestamp is a finite number
assert math.isfinite(e1), f"e1 is not finite: {e1}"
assert math.isfinite(e2), f"e2 is not finite: {e2}"
```

or you could avoid the subtraction all together:
```assert e2 >= e1, f"Event timestamps out of order: e1={e1}, e2={e2}"```

File: cuda_core/tests/helpers/__init__.py (hunk 1 above)
Comment (rwgk) [reply cuda_core/tests/helpers/__init__.py]: I reverted the commit. The rest of the PR is more important.

File: cuda_core/tests/test_event.py (hunk 2 above)
Comment (rwgk) [reply cuda_core/tests/test_event.py]: In this particular case, the side-effect is not interesting: we don't want to exercise that a kernel runs, we're interested in getting a result that is different from zero from `e2 - e1`.

If we add a kernel, `__nanosleep` should make the behavior deterministic, e.g. if we sleep for 1000 nanoseconds, `delta_ms >= 500` should be true going by the documentation; but I'd give it `2000` nanoseconds to steer clear of edge cases.

File: cuda_core/tests/test_event.py (hunk 3 above)
Comment (rwgk) [reply cuda_core/tests/test_event.py]: In this case `e1` and `e2` are `Event` objects:

```
>       assert math.isfinite(e1)
               ^^^^^^^^^^^^^^^^^
E       TypeError: must be real number, not cuda.core.experimental._event.Event
```

The code exercised here is this part of the `Event.__sub__` implementation (in `_event.pyx`):

```
    def __sub__(self, other: Event):
        # return self - other (in milliseconds)
        cdef float timing
        with nogil:
            err = cydriver.cuEventElapsedTime(&timing, other._handle, self._handle)
        if err == 0:
            return timing
```

I.e. only `delta_ms` is a `float`, exactly as produced by `cuEventElapsedTime()`.

We could write `assert math.isfinite(e2 - e1)`, but that's only a tiny step closer to the `cuEventElapsedTime()` call. I think it'll already be very obvious where the problem is if this `assert` fires, even with `math.isfinite(delta_ms)`, which is better because then it's unambiguous that the three references to `delta_ms` are all exactly the same object.

`pytest` is also very good at automagically exposing all relevant information with the **`--showlocals`** option, e.g.:

```
(TestVenv) smc120-0004.ipp2a2.colossus.nvidia.com:/wrk/forked/cuda-python/cuda_core $ pytest --showlocals tests/test_event.py::test_a_minus_b
...
    def test_a_minus_b():
        a = float('inf')
        b = 1
        c = a - b
>       assert math.isfinite(c)
E       assert False
E        +  where False = <built-in function isfinite>(inf)
E        +    where <built-in function isfinite> = math.isfinite

a          = inf
b          = 1
c          = inf

tests/test_event.py:46: AssertionError
```

We should add that to all our `pytest.ini` files (separate PR).

Comment (rwgk) [reply cuda_core/tests/test_event.py]: I created PR #1286 to add the `--showlocals` option.

**Review by kkraus14 — DISMISSED** [review]
LGTM! Thanks @rwgk

**Comment by rwgk** [comment]
I (actually cursor-agent) added a couple commits to insert a kernel that causes a delay of 20 ms. I was surprised to see that `__nanosleep` by itself didn't lead to predictable behavior, is that expected?

**Comment by rwgk** [comment]
> LGTM! Thanks @rwgk

Oops, sorry, I somehow missed this response.

I was curious to see how much trouble it is to add the kernel, and apart from the nanosleep surprise, cursor did that in seconds.

I'll let the test finish, then we can still decide if we want to keep the kernel, or remove it again.

**Comment by kkraus14** [comment]
My 2c: is that we should remove it in the name of simplicity. This adds a lot of machinery and a lot of things that can go wrong in order to test event timing, but I don't have a strong opinion if you think that this is valuable.

I would move all of the kernel definition and compilation out of the test into the module if we decide to keep it though.

File: cuda_core/tests/test_event.py (hunk 3 above)
Comment (leofang) [reply cuda_core/tests/test_event.py]: You can't compare event or print event, CUDA does not offer this capability. Or the relative time difference as reported by `cuEventElapsedTime` can be accessed.

File: cuda_core/tests/test_event.py [hunk 4]
```diff
@@ -1,47 +1,71 @@
...
+
+    # Create a simple kernel that sleeps for 20 ms to ensure a measurable delay
+    # This guarantees delta_ms > 10 without depending on OS/driver timing characteristics
+    # Use clock64() in a loop to ensure we actually wait for the full duration
+    clock_rate_hz = device.properties.clock_rate * 1000
+    sleep_cycles = int(0.020 * clock_rate_hz)  # 20 ms in clock cycles
+    code = f"""
+    extern "C"
+    __global__ void nanosleep_kernel() {{
+        unsigned long long start = clock64();
```
leofang: Q: I think it's `long long int` not `unsigned long long`?
https://docs.nvidia.com/cuda/cuda-c-programming-guide/#time-function

cuda_core/tests/test_event.py
```diff
@@ -1,47 +1,71 @@
...
+        while (clock64() - start < {sleep_cycles}) {{
+            __nanosleep(1000000); // 1 ms yield to avoid 100% spin
+        }}
+    }}
+    """
+    arch = "".join(f"{i}" for i in device.compute_capability)
+    program_options = ProgramOptions(std="c++17", arch=f"sm_{arch}")
```
leofang: nit:
```suggestion
    program_options = ProgramOptions(std="c++17", arch=f"sm_{device.arch}")
```

cuda_core/tests/test_event.py
```diff
@@ -1,47 +1,71 @@
...
+    # Sanity check: cuEventElapsedTime should always return a finite float for two completed
+    # events. This guards against unexpected driver/HW anomalies (e.g. NaN or inf) or general
+    # undefined behavior, without asserting anything about the magnitude of the measured time.
+    assert math.isfinite(delta_ms)
+    # With the nanosleep kernel between events, we can assert a positive elapsed time.
+    # The kernel sleeps for 20 ms using clock64(), so delta_ms should be at least ~10 ms.
+    # Using a 10 ms threshold (half the sleep duration) provides a large safety margin above
+    # the ~0.5 microsecond resolution of cudaEventElapsedTime, making this test deterministic
+    # and non-flaky.
+    assert delta_ms > 10
```
leofang: Q: Should equality be included?
```suggestion
    assert delta_ms >= 10
```

rwgk:
I verified conclusively that `test_event_elapsed_time_basic` (as of commit 5eba5ac7de8633b57ca7881f6e0312597d49c545)
PASSES with the Windows **WDDM** driver model, but **HAGS OFF**.

Command used:

```
python -m pytest -ra -s -vv tests\test_event.py::test_event_elapsed_time_basic --count=1000
```

Complete test script: [test_event_elapsed_time_basic.cmd](https://github.com/user-attachments/files/23754389/test_event_elapsed_time_basic.cmd.txt)

Full log: [test_event_elapsed_time_basic_hags_off_2025-11-25+105446_log.txt](https://github.com/user-attachments/files/23754400/test_event_elapsed_time_basic_hags_off_2025-11-25%2B105446_log.txt)

For completeness: The same pytest command also passes with HAGS ON.

rwgk:
> I would move all of the kernel definition and compilation out of the test into the module if we decide to keep it though.

Given all the effort that went into this already (mostly before this PR, including the back and forth with SWQA) I don't want to stop a few inches before the finish line.

I'll move the new kernel code into a module and address Leo's comments.

cuda_core/tests/test_event.py
rwgk: Nice catch, thanks!

Done: commit ad169339ce884e3655f211803e7fb7b4a93edb85

cuda_core/tests/test_event.py
rwgk: Much better!

Done: commit 55d9b44c190e4b7408bf068b12d6d0fba5262572

cuda_core/tests/test_event.py
rwgk: Technically: Because of the large safety margin (expected 10 ms) it shouldn't matter at all.

Readability aspect: Making an effort to be precise here would send the wrong message, by distracting from the large safety margin.

rwgk:
> I would move all of the kernel definition and compilation out of the test into the module if we decide to keep it though.

Done. That particular step was again just seconds worth of cursor-agent effort.

I changed the naming back to be more similar to the original code, and polished the comments manually, to convince myself it's all accurate.

Interactive testing with HAGS OFF still passes (`pytest -ra -s -v tests\test_event.py::test_timing_success --count=1000`).

I'll reboot my machine to get it back to HAGS ON (where I want to keep it) and then test again, while the CI is running here.

cuda_core/tests/helpers/nanosleep_kernel.py
```diff
@@ -0,0 +1,44 @@
...
+    def __init__(self, device, sleep_duration_ms=20):
+        """
+        Initialize the nanosleep kernel.
+
+        Args:
+            device: CUDA device to compile the kernel for
+            sleep_duration_ms: Duration to sleep in milliseconds (default: 20)
+        """
+        # clock_rate is in kHz
+        sleep_cycles = int(sleep_duration_ms * device.properties.clock_rate) + 1
```
kkraus14: Given we're just trying to wait ~20ms, I don't think we need to use the clock rate here to try to get something cycle accurate. We should be able to simplify the loop around `__nanosleep` to just go through 20 iterations?

rwgk: Done: commit 35510e6b620729480b1f1a38a11748efb930b884

Thanks, I should have looked at the `__nanosleep` documentation before:

> The maximum sleep duration is approximately 1 millisecond.

No wonder that the large values I tried to use yesterday didn't make sense!

While I was at it, I printed out the measured elapsed times:

```
   n: 1000
 min: 19.9619
mean: 20.5188
 max: 21.0688
```

Just as documented (~0.5 microsecond resolution of `cudaEventElapsedTime()`). Nice.

kkraus14 (approved):
LGTM!

rwgk:
Thanks for all the feedback!

PR #1289, opened 2025-11-26
Add peer access control for DeviceMemoryResource

### LINKED ISSUE #479: Add peer access support
In a multi-GPU setting, right now it's not possible to memcpy directly between the two devices. A couple of checks would have to be done by cuda.core as part of enabling this:
- cuDeviceCanAccessPeer
- cuCtxEnablePeerAccess

Andy-Jost:
## Description

closes #479

## Add peer access control for DeviceMemoryResource

This PR implements peer access control for `DeviceMemoryResource`, allowing memory pool allocations to be accessed from devices other than the owner device.

### Overview

Memory pools created by `DeviceMemoryResource` can now grant peer access permissions to other GPUs, enabling multi-GPU workflows where allocations on one device need to be accessed by kernels running on another device.

### Key Changes

**1. New `peer_accessible_by` property**
- Get/set which devices can access allocations from a memory pool
- Returns a tuple of sorted device IDs that currently have peer access
- Setter accepts a sequence of Device objects or device IDs
- Automatically excludes the owner device from the peer access list
- Uses incremental updates (only changes permissions for devices being added/removed)

**Example usage:**
```
dmr = DeviceMemoryResource(0)
dmr.peer_accessible_by = [1, 2]  # Grant access to devices 1 and 2
assert dmr.peer_accessible_by == (1, 2)
dmr.peer_accessible_by = []      # Revoke all peer access
```

**2. Implementation details**
- Uses `cuMemPoolSetAccess` to modify peer access permissions at runtime
- Properly allocates C arrays for passing `CUmemAccessDesc` structures to the driver API
- Memory-safe with try/finally blocks ensuring allocated memory is freed
- Supports both granting (`CU_MEM_ACCESS_FLAGS_PROT_READWRITE`) and revoking (`CU_MEM_ACCESS_FLAGS_PROT_NONE`) access in a single transaction

**3. Debugging support**
- Added `DMR_mempool_get_access()` helper function to probe actual access permissions via `cuMemPoolGetAccess`
- Useful for debugging and verifying expected access states

**4. Workaround for driver bug (nvbug 5698116)**
- Added cleanup in `__dealloc__` to reset peer access before destroying memory pools
- Works around an issue where recycled memory pool handles inherit peer access state from previous handles

### Technical Notes

- Required adding `malloc`/`free` imports from `libc.stdlib` for C array allocation
- The implementation properly handles the conversion from Python collections to C structures required by the low-level CUDA driver API
- No trailing whitespace in source files (adheres to project code style)

### Testing

This feature is designed to work with the existing multi-GPU test suite and enables peer access scenarios for buffer copy operations between devices.

### DIFF
--- a/cuda_bindings/pyproject.toml
+++ b/cuda_bindings/pyproject.toml
@@ -69,5 +69,5 @@ repair-wheel-command = "delvewheel repair --namespace-pkg cuda -w {dest_dir} {wh
 
 [tool.pytest.ini_options]
 required_plugins = "pytest-benchmark"
-addopts = "--benchmark-disable"
+addopts = "--benchmark-disable --showlocals"
 norecursedirs = ["tests/cython", "examples"]
--- a/cuda_core/cuda/core/experimental/_dlpack.pyx
+++ b/cuda_core/cuda/core/experimental/_dlpack.pyx
@@ -26,49 +26,54 @@ cdef void pycapsule_deleter(object capsule) noexcept:
 
 
 cdef void deleter(DLManagedTensor* tensor) noexcept with gil:
-    stdlib.free(tensor.dl_tensor.shape)
-    if tensor.manager_ctx:
-        cpython.Py_DECREF(<object>tensor.manager_ctx)
-        tensor.manager_ctx = NULL
-    stdlib.free(tensor)
+    if tensor:
+        if tensor.dl_tensor.shape:
+            stdlib.free(tensor.dl_tensor.shape)
+        if tensor.manager_ctx:
+            cpython.Py_DECREF(<object>tensor.manager_ctx)
+            tensor.manager_ctx = NULL
+        stdlib.free(tensor)
 
 
 cdef void versioned_deleter(DLManagedTensorVersioned* tensor) noexcept with gil:
-    stdlib.free(tensor.dl_tensor.shape)
-    if tensor.manager_ctx:
-        cpython.Py_DECREF(<object>tensor.manager_ctx)
-        tensor.manager_ctx = NULL
-    stdlib.free(tensor)
+    if tensor:
+        if tensor.dl_tensor.shape:
+            stdlib.free(tensor.dl_tensor.shape)
+        if tensor.manager_ctx:
+            cpython.Py_DECREF(<object>tensor.manager_ctx)
+            tensor.manager_ctx = NULL
+        stdlib.free(tensor)
 
 
-cpdef object make_py_capsule(object buf, bint versioned):
-    cdef DLManagedTensor* dlm_tensor
-    cdef DLManagedTensorVersioned* dlm_tensor_ver
-    cdef DLTensor* dl_tensor
-    cdef void* tensor_ptr
-    cdef const char* capsule_name
-
-    if versioned:
+cdef inline DLManagedTensorVersioned* allocate_dlm_tensor_versioned() except? NULL:
+    cdef DLManagedTensorVersioned* dlm_tensor_ver = NULL
+    try:
         dlm_tensor_ver = <DLManagedTensorVersioned*>(
             stdlib.malloc(sizeof(DLManagedTensorVersioned)))
-        dlm_tensor_ver.version.major = DLPACK_MAJOR_VERSION
-        dlm_tensor_ver.version.minor = DLPACK_MINOR_VERSION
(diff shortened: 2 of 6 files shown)


cuda_core/cuda/core/experimental/_device.pyx
```diff
@@ -1035,6 +1035,27 @@ class Device:
...
+        """
+        peer = Device(peer)
+        cdef int d1 = <int> self.device_id
+        cdef int d2 = <int> peer.device_id
+        if d1 == d2:
+            return True
+        cdef int value = 0
+        with nogil:
+            HANDLE_RETURN(cydriver.cuDeviceCanAccessPeer(&value, d1, d2))
+        return bool(value)
```
leofang, line 1057: @pciolkosz can correct me, but I think this is to check peer accessibility enabled via the old/legacy, per-context approach, instead of the new per-mempool approach? I don't think both are compatible?

Andy-Jost: I didn't see anything in the new interface to check whether one device is accessible from another. I inferred from this function that (at least at some point) it was possible for one device to be unreachable from another. We definitely need to resolve this.

Andy-Jost: This page from the C++ Programming Guide has what we need: "To enable access from other devices, the accessing device must be peer capable with the memory pool’s device; check with cudaDeviceCanAccessPeer"

https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#device-accessibility-for-multi-gpu-support

cuda_core/cuda/core/experimental/_memory/_device_memory_resource.pyx
```diff
@@ -408,6 +410,69 @@ cdef class DeviceMemoryResource(MemoryResource):
...
+        """
+        return self._peer_accessible_by
+
+    @peer_accessible_by.setter
+    def peer_accessible_by(self, devices):
+        """Set which devices can access this memory pool."""
+        from .._device import Device
+
+        # Convert all devices to device IDs
+        cdef set target_ids = {Device(dev).device_id for dev in devices}
```
Andy-Jost: I think I should filter this set by cudaDeviceCanAccessPeer

cuda_core/cuda/core/experimental/_memory/_device_memory_resource.pxd
```diff
@@ -14,4 +14,8 @@ cdef class DeviceMemoryResource(MemoryResource):
         bint                  _mempool_owned
         IPCData               _ipc_data
         object                _attributes
+        object                _peer_accessible_by
```
Andy-Jost, line 17: I will address peer access with IPC memory pools in a follow-up change. The peer access attributes are not inherited when an allocation is sent to another process via IPC, but access can be set. It will require a new test and possibly a small code change.

cuda_core/cuda/core/experimental/_device.pyx
pciolkosz: Yeah, I think this is correct. `cuDeviceCanAccessPeer` should work for both legacy and mempool path, since the peer access requirements should be the same for both

cuda_core/cuda/core/experimental/_memory/_device_memory_resource.pyx
Andy-Jost: Done in the latest. I added a check that raises an error if a non-accessible device is added to the list. Unfortunately, I don't see a good way to test it.

cuda_core/tests/test_memory_peer_access.py
```diff
@@ -0,0 +1,169 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import cuda.core.experimental
+import pytest
+from cuda.core.experimental import Device, DeviceMemoryResource
+from cuda.core.experimental._utils.cuda_utils import CUDAError
+from helpers.buffers import PatternGen, compare_buffer_to_constant, make_scratch_buffer
+
+NBYTES = 1024
```
rparolin, line 10: I'd make this a local variable to the unit test instead of a global. bonus points if you make the pytest test multiple memory block sizes.

rparolin (approved):

PR #1291, opened 2025-11-27
Fix potential leaks for interrupted dlpack capsule creation

stiepan:
## Description

If dlpack's make_py_capsule is interrupted with an exception, it does not free memory malloced for the dl tensor and layout data.
This PR:
* splits the make_py_capsule into a couple of different helpers (allocating the dlm tensor, layout setup, device setup, dtype setup) for readability
* adjusts deleters so that they can be safely run with partially initialized object
* adds a test that checks if the cleanup was run when exception is raised by checking the refcount of the buffer object. As the decref is not left to the deleter, this assures the deleter was run.

## Checklist


leofang:
/ok to test [044f179](https://github.com/NVIDIA/cuda-python/pull/1291/commits/044f179998fc2ed232792a31cdb060234489087e)

leofang (approved):

PR #1292, opened 2025-12-01
chore: Replace isinstance(obj, T) with type(obj) is T comparisons

### LINKED ISSUE #1282: Replace `isinstance` checks by `type ==` in `ParamHolder`
During the meeting today, @mdboom suggested that in Cython `type(obj) == T` is faster than `isinstance(obj, T)`. We should profile it and make changes if this is true, since this is inside `launch(s, config, ker, *args)` and is a hot path.

bharatr21:
## Description

closes #1282 

Replace `isinstance(obj, T)` checks with `type(obj) is T` to optimize `cuda.core.launch()`

------------------------------------
#### Additional Notes
------------------------------------ 
I made a benchmarking script in Cython to prove the speedup of using `type()` in place of `isinstance()` checks since the original issue requested profiling which resulted in an **~5x** speedup.  
Appreciate some guidance to know if I've done the profiling right

Created a file `benchmark_isinstance_cython.pyx` :

 ```pyx
from cpython.mem cimport PyMem_Malloc, PyMem_Free
from libc.stdint cimport (intptr_t,
                          int8_t, int16_t, int32_t, int64_t,
                          uint8_t, uint16_t, uint32_t, uint64_t)
from libcpp cimport bool as cpp_bool
from libcpp.complex cimport complex as cpp_complex
from libcpp.vector cimport vector

import ctypes
import numpy
import time
from statistics import mean, stdev


ctypedef cpp_complex.complex[float] cpp_single_complex
ctypedef cpp_complex.complex[double] cpp_double_complex

# Cache type objects
cdef object ctypes_bool = ctypes.c_bool
cdef object ctypes_int8 = ctypes.c_int8
cdef object ctypes_int16 = ctypes.c_int16
cdef object ctypes_int32 = ctypes.c_int32
cdef object ctypes_int64 = ctypes.c_int64
cdef object ctypes_uint8 = ctypes.c_uint8
cdef object ctypes_uint16 = ctypes.c_uint16
cdef object ctypes_uint32 = ctypes.c_uint32
cdef object ctypes_uint64 = ctypes.c_uint64
cdef object ctypes_float = ctypes.c_float
cdef object ctypes_double = ctypes.c_double
cdef object numpy_bool = numpy.bool_
cdef object numpy_int8 = numpy.int8
cdef object numpy_int16 = numpy.int16
cdef object numpy_int32 = numpy.int32
cdef object numpy_int64 = numpy.int64
cdef object numpy_uint8 = numpy.uint8
cdef object numpy_uint16 = numpy.uint16
cdef object numpy_uint32 = numpy.uint32
cdef object numpy_uint64 = numpy.uint64
cdef object numpy_float16 = numpy.float16
cdef object numpy_float32 = numpy.float32
cdef object numpy_float64 = numpy.float64
cdef object numpy_complex64 = numpy.complex64
cdef object numpy_complex128 = numpy.complex128

# Limitation due to cython/cython#534
ctypedef void* voidptr


# ============================================================================
# Version 1: Current implementation using isinstance()
# ============================================================================

cdef inline int prepare_ctypes_arg_isinstance(
        vector[void*]& data,
        vector[void*]& data_addresses,
        arg,
        const size_t idx) except -1:
    cdef void* ptr

    if isinstance(arg, ctypes_bool):
        ptr = PyMem_Malloc(sizeof(cpp_bool))
        (<cpp_bool*>ptr)[0] = <cpp_bool>(arg.value)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif isinstance(arg, ctypes_int8):
        ptr = PyMem_Malloc(sizeof(int8_t))
        (<int8_t*>ptr)[0] = <int8_t>(arg.value)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif isinstance(arg, ctypes_int16):
        ptr = PyMem_Malloc(sizeof(int16_t))
        (<int16_t*>ptr)[0] = <int16_t>(arg.value)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif isinstance(arg, ctypes_int32):
        ptr = PyMem_Malloc(sizeof(int32_t))
        (<int32_t*>ptr)[0] = <int32_t>(arg.value)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif isinstance(arg, ctypes_int64):
        ptr = PyMem_Malloc(sizeof(int64_t))
        (<int64_t*>ptr)[0] = <int64_t>(arg.value)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif isinstance(arg, ctypes_uint8):
        ptr = PyMem_Malloc(sizeof(uint8_t))
        (<uint8_t*>ptr)[0] = <uint8_t>(arg.value)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif isinstance(arg, ctypes_uint16):
        ptr = PyMem_Malloc(sizeof(uint16_t))
        (<uint16_t*>ptr)[0] = <uint16_t>(arg.value)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif isinstance(arg, ctypes_uint32):
        ptr = PyMem_Malloc(sizeof(uint32_t))
        (<uint32_t*>ptr)[0] = <uint32_t>(arg.value)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif isinstance(arg, ctypes_uint64):
        ptr = PyMem_Malloc(sizeof(uint64_t))
        (<uint64_t*>ptr)[0] = <uint64_t>(arg.value)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif isinstance(arg, ctypes_float):
        ptr = PyMem_Malloc(sizeof(float))
        (<float*>ptr)[0] = <float>(arg.value)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif isinstance(arg, ctypes_double):
        ptr = PyMem_Malloc(sizeof(double))
        (<double*>ptr)[0] = <double>(arg.value)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    else:
        return 1


cdef inline int prepare_numpy_arg_isinstance(
        vector[void*]& data,
        vector[void*]& data_addresses,
        arg,
        const size_t idx) except -1:
    cdef void* ptr

    if isinstance(arg, numpy_bool):
        ptr = PyMem_Malloc(sizeof(cpp_bool))
        (<cpp_bool*>ptr)[0] = <cpp_bool>(arg)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif isinstance(arg, numpy_int8):
        ptr = PyMem_Malloc(sizeof(int8_t))
        (<int8_t*>ptr)[0] = <int8_t>(arg)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif isinstance(arg, numpy_int16):
        ptr = PyMem_Malloc(sizeof(int16_t))
        (<int16_t*>ptr)[0] = <int16_t>(arg)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif isinstance(arg, numpy_int32):
        ptr = PyMem_Malloc(sizeof(int32_t))
        (<int32_t*>ptr)[0] = <int32_t>(arg)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif isinstance(arg, numpy_int64):
        ptr = PyMem_Malloc(sizeof(int64_t))
        (<int64_t*>ptr)[0] = <int64_t>(arg)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif isinstance(arg, numpy_uint8):
        ptr = PyMem_Malloc(sizeof(uint8_t))
        (<uint8_t*>ptr)[0] = <uint8_t>(arg)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif isinstance(arg, numpy_uint16):
        ptr = PyMem_Malloc(sizeof(uint16_t))
        (<uint16_t*>ptr)[0] = <uint16_t>(arg)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif isinstance(arg, numpy_uint32):
        ptr = PyMem_Malloc(sizeof(uint32_t))
        (<uint32_t*>ptr)[0] = <uint32_t>(arg)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif isinstance(arg, numpy_uint64):
        ptr = PyMem_Malloc(sizeof(uint64_t))
        (<uint64_t*>ptr)[0] = <uint64_t>(arg)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif isinstance(arg, numpy_float32):
        ptr = PyMem_Malloc(sizeof(float))
        (<float*>ptr)[0] = <float>(arg)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif isinstance(arg, numpy_float64):
        ptr = PyMem_Malloc(sizeof(double))
        (<double*>ptr)[0] = <double>(arg)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif isinstance(arg, numpy_complex64):
        ptr = PyMem_Malloc(sizeof(cpp_single_complex))
        (<cpp_single_complex*>ptr)[0] = cpp_complex.complex[float](arg.real, arg.imag)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif isinstance(arg, numpy_complex128):
        ptr = PyMem_Malloc(sizeof(cpp_double_complex))
        (<cpp_double_complex*>ptr)[0] = cpp_complex.complex[double](arg.real, arg.imag)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    else:
        return 1


# ============================================================================
# Version 2: Optimized implementation using type() is with isinstance fallback
# ============================================================================

cdef inline int prepare_ctypes_arg_type_is(
        vector[void*]& data,
        vector[void*]& data_addresses,
        arg,
        const size_t idx) except -1:
    cdef void* ptr
    cdef object arg_type = type(arg)

    if arg_type is ctypes_bool:
        ptr = PyMem_Malloc(sizeof(cpp_bool))
        (<cpp_bool*>ptr)[0] = <cpp_bool>(arg.value)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif arg_type is ctypes_int8:
        ptr = PyMem_Malloc(sizeof(int8_t))
        (<int8_t*>ptr)[0] = <int8_t>(arg.value)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif arg_type is ctypes_int16:
        ptr = PyMem_Malloc(sizeof(int16_t))
        (<int16_t*>ptr)[0] = <int16_t>(arg.value)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif arg_type is ctypes_int32:
        ptr = PyMem_Malloc(sizeof(int32_t))
        (<int32_t*>ptr)[0] = <int32_t>(arg.value)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif arg_type is ctypes_int64:
        ptr = PyMem_Malloc(sizeof(int64_t))
        (<int64_t*>ptr)[0] = <int64_t>(arg.value)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif arg_type is ctypes_uint8:
        ptr = PyMem_Malloc(sizeof(uint8_t))
        (<uint8_t*>ptr)[0] = <uint8_t>(arg.value)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif arg_type is ctypes_uint16:
        ptr = PyMem_Malloc(sizeof(uint16_t))
        (<uint16_t*>ptr)[0] = <uint16_t>(arg.value)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif arg_type is ctypes_uint32:
        ptr = PyMem_Malloc(sizeof(uint32_t))
        (<uint32_t*>ptr)[0] = <uint32_t>(arg.value)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif arg_type is ctypes_uint64:
        ptr = PyMem_Malloc(sizeof(uint64_t))
        (<uint64_t*>ptr)[0] = <uint64_t>(arg.value)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif arg_type is ctypes_float:
        ptr = PyMem_Malloc(sizeof(float))
        (<float*>ptr)[0] = <float>(arg.value)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif arg_type is ctypes_double:
        ptr = PyMem_Malloc(sizeof(double))
        (<double*>ptr)[0] = <double>(arg.value)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    else:
        # Backward compatibility: fallback to isinstance for subclasses
        if isinstance(arg, ctypes_bool):
            ptr = PyMem_Malloc(sizeof(cpp_bool))
            (<cpp_bool*>ptr)[0] = <cpp_bool>(arg.value)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        elif isinstance(arg, ctypes_int8):
            ptr = PyMem_Malloc(sizeof(int8_t))
            (<int8_t*>ptr)[0] = <int8_t>(arg.value)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        elif isinstance(arg, ctypes_int16):
            ptr = PyMem_Malloc(sizeof(int16_t))
            (<int16_t*>ptr)[0] = <int16_t>(arg.value)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        elif isinstance(arg, ctypes_int32):
            ptr = PyMem_Malloc(sizeof(int32_t))
            (<int32_t*>ptr)[0] = <int32_t>(arg.value)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        elif isinstance(arg, ctypes_int64):
            ptr = PyMem_Malloc(sizeof(int64_t))
            (<int64_t*>ptr)[0] = <int64_t>(arg.value)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        elif isinstance(arg, ctypes_uint8):
            ptr = PyMem_Malloc(sizeof(uint8_t))
            (<uint8_t*>ptr)[0] = <uint8_t>(arg.value)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        elif isinstance(arg, ctypes_uint16):
            ptr = PyMem_Malloc(sizeof(uint16_t))
            (<uint16_t*>ptr)[0] = <uint16_t>(arg.value)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        elif isinstance(arg, ctypes_uint32):
            ptr = PyMem_Malloc(sizeof(uint32_t))
            (<uint32_t*>ptr)[0] = <uint32_t>(arg.value)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        elif isinstance(arg, ctypes_uint64):
            ptr = PyMem_Malloc(sizeof(uint64_t))
            (<uint64_t*>ptr)[0] = <uint64_t>(arg.value)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        elif isinstance(arg, ctypes_float):
            ptr = PyMem_Malloc(sizeof(float))
            (<float*>ptr)[0] = <float>(arg.value)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        elif isinstance(arg, ctypes_double):
            ptr = PyMem_Malloc(sizeof(double))
            (<double*>ptr)[0] = <double>(arg.value)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        else:
            return 1


cdef inline int prepare_numpy_arg_type_is(
        vector[void*]& data,
        vector[void*]& data_addresses,
        arg,
        const size_t idx) except -1:
    cdef void* ptr
    cdef object arg_type = type(arg)

    if arg_type is numpy_bool:
        ptr = PyMem_Malloc(sizeof(cpp_bool))
        (<cpp_bool*>ptr)[0] = <cpp_bool>(arg)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif arg_type is numpy_int8:
        ptr = PyMem_Malloc(sizeof(int8_t))
        (<int8_t*>ptr)[0] = <int8_t>(arg)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif arg_type is numpy_int16:
        ptr = PyMem_Malloc(sizeof(int16_t))
        (<int16_t*>ptr)[0] = <int16_t>(arg)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif arg_type is numpy_int32:
        ptr = PyMem_Malloc(sizeof(int32_t))
        (<int32_t*>ptr)[0] = <int32_t>(arg)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif arg_type is numpy_int64:
        ptr = PyMem_Malloc(sizeof(int64_t))
        (<int64_t*>ptr)[0] = <int64_t>(arg)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif arg_type is numpy_uint8:
        ptr = PyMem_Malloc(sizeof(uint8_t))
        (<uint8_t*>ptr)[0] = <uint8_t>(arg)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif arg_type is numpy_uint16:
        ptr = PyMem_Malloc(sizeof(uint16_t))
        (<uint16_t*>ptr)[0] = <uint16_t>(arg)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif arg_type is numpy_uint32:
        ptr = PyMem_Malloc(sizeof(uint32_t))
        (<uint32_t*>ptr)[0] = <uint32_t>(arg)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif arg_type is numpy_uint64:
        ptr = PyMem_Malloc(sizeof(uint64_t))
        (<uint64_t*>ptr)[0] = <uint64_t>(arg)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif arg_type is numpy_float32:
        ptr = PyMem_Malloc(sizeof(float))
        (<float*>ptr)[0] = <float>(arg)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif arg_type is numpy_float64:
        ptr = PyMem_Malloc(sizeof(double))
        (<double*>ptr)[0] = <double>(arg)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif arg_type is numpy_complex64:
        ptr = PyMem_Malloc(sizeof(cpp_single_complex))
        (<cpp_single_complex*>ptr)[0] = cpp_complex.complex[float](arg.real, arg.imag)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    elif arg_type is numpy_complex128:
        ptr = PyMem_Malloc(sizeof(cpp_double_complex))
        (<cpp_double_complex*>ptr)[0] = cpp_complex.complex[double](arg.real, arg.imag)
        data_addresses[idx] = ptr
        data[idx] = ptr
        return 0
    else:
        # Backward compatibility: fallback to isinstance for subclasses
        if isinstance(arg, numpy_bool):
            ptr = PyMem_Malloc(sizeof(cpp_bool))
            (<cpp_bool*>ptr)[0] = <cpp_bool>(arg)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        elif isinstance(arg, numpy_int8):
            ptr = PyMem_Malloc(sizeof(int8_t))
            (<int8_t*>ptr)[0] = <int8_t>(arg)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        elif isinstance(arg, numpy_int16):
            ptr = PyMem_Malloc(sizeof(int16_t))
            (<int16_t*>ptr)[0] = <int16_t>(arg)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        elif isinstance(arg, numpy_int32):
            ptr = PyMem_Malloc(sizeof(int32_t))
            (<int32_t*>ptr)[0] = <int32_t>(arg)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        elif isinstance(arg, numpy_int64):
            ptr = PyMem_Malloc(sizeof(int64_t))
            (<int64_t*>ptr)[0] = <int64_t>(arg)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        elif isinstance(arg, numpy_uint8):
            ptr = PyMem_Malloc(sizeof(uint8_t))
            (<uint8_t*>ptr)[0] = <uint8_t>(arg)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        elif isinstance(arg, numpy_uint16):
            ptr = PyMem_Malloc(sizeof(uint16_t))
            (<uint16_t*>ptr)[0] = <uint16_t>(arg)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        elif isinstance(arg, numpy_uint32):
            ptr = PyMem_Malloc(sizeof(uint32_t))
            (<uint32_t*>ptr)[0] = <uint32_t>(arg)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        elif isinstance(arg, numpy_uint64):
            ptr = PyMem_Malloc(sizeof(uint64_t))
            (<uint64_t*>ptr)[0] = <uint64_t>(arg)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        elif isinstance(arg, numpy_float32):
            ptr = PyMem_Malloc(sizeof(float))
            (<float*>ptr)[0] = <float>(arg)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        elif isinstance(arg, numpy_float64):
            ptr = PyMem_Malloc(sizeof(double))
            (<double*>ptr)[0] = <double>(arg)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        elif isinstance(arg, numpy_complex64):
            ptr = PyMem_Malloc(sizeof(cpp_single_complex))
            (<cpp_single_complex*>ptr)[0] = cpp_complex.complex[float](arg.real, arg.imag)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        elif isinstance(arg, numpy_complex128):
            ptr = PyMem_Malloc(sizeof(cpp_double_complex))
            (<cpp_double_complex*>ptr)[0] = cpp_complex.complex[double](arg.real, arg.imag)
            data_addresses[idx] = ptr
            data[idx] = ptr
            return 0
        else:
            return 1


# ============================================================================
# Benchmark functions
# ============================================================================

def benchmark_isinstance(kernel_args, int iterations):
    """Benchmark the isinstance() approach."""
    cdef size_t n_args = len(kernel_args)
    cdef size_t i, j
    cdef int not_prepared
    cdef vector[voidptr] data
    cdef vector[voidptr] data_addresses
    cdef double start, end

    # Warmup
    for _ in range(100):
        data = vector[voidptr](n_args, NULL)
        data_addresses = vector[voidptr](n_args)
        for i, arg in enumerate(kernel_args):
            if isinstance(arg, int):
                continue
            elif isinstance(arg, float):
                continue
            elif isinstance(arg, complex):
                continue
            elif isinstance(arg, bool):
                continue

            not_prepared = prepare_numpy_arg_isinstance(data, data_addresses, arg, i)
            if not_prepared:
                not_prepared = prepare_ctypes_arg_isinstance(data, data_addresses, arg, i)

        for data_ptr in data:
            if data_ptr:
                PyMem_Free(data_ptr)

    # Actual benchmark
    start = time.perf_counter()
    for j in range(iterations):
        data = vector[voidptr](n_args, NULL)
        data_addresses = vector[voidptr](n_args)

        for i, arg in enumerate(kernel_args):
            if isinstance(arg, int):
                continue
            elif isinstance(arg, float):
                continue
            elif isinstance(arg, complex):
                continue
            elif isinstance(arg, bool):
                continue

            not_prepared = prepare_numpy_arg_isinstance(data, data_addresses, arg, i)
            if not_prepared:
                not_prepared = prepare_ctypes_arg_isinstance(data, data_addresses, arg, i)

        for data_ptr in data:
            if data_ptr:
                PyMem_Free(data_ptr)

    end = time.perf_counter()
    return end - start


def benchmark_type_is(kernel_args, int iterations):
    """Benchmark the type() is approach with isinstance fallback."""
    cdef size_t n_args = len(kernel_args)
    cdef size_t i, j
    cdef int not_prepared
    cdef vector[voidptr] data
    cdef vector[voidptr] data_addresses
    cdef double start, end
    cdef object arg_type

    # Warmup
    for _ in range(100):
        data = vector[voidptr](n_args, NULL)
        data_addresses = vector[voidptr](n_args)
        for i, arg in enumerate(kernel_args):
            arg_type = type(arg)
            if arg_type is int:
                continue
            elif arg_type is float:
                continue
            elif arg_type is complex:
                continue
            elif arg_type is bool:
                continue

            not_prepared = prepare_numpy_arg_type_is(data, data_addresses, arg, i)
            if not_prepared:
                not_prepared = prepare_ctypes_arg_type_is(data, data_addresses, arg, i)

        for data_ptr in data:
            if data_ptr:
                PyMem_Free(data_ptr)

    # Actual benchmark
    start = time.perf_counter()
    for j in range(iterations):
        data = vector[voidptr](n_args, NULL)
        data_addresses = vector[voidptr](n_args)

        for i, arg in enumerate(kernel_args):
            arg_type = type(arg)
            if arg_type is int:
                continue
            elif arg_type is float:
                continue
            elif arg_type is complex:
                continue
            elif arg_type is bool:
                continue

            not_prepared = prepare_numpy_arg_type_is(data, data_addresses, arg, i)
            if not_prepared:
                not_prepared = prepare_ctypes_arg_type_is(data, data_addresses, arg, i)

        for data_ptr in data:
            if data_ptr:
                PyMem_Free(data_ptr)

    end = time.perf_counter()
    return end - start


def run_benchmark():
    """Main benchmark runner."""
    print("=" * 70)
    print("Cython Benchmark: isinstance() vs type() is + isinstance fallback")
    print("Kernel Argument Handling Hot Path")
    print("=" * 70)
    print()

    # Create realistic kernel arguments
    kernel_args = [
        numpy.int32(100),
        numpy.float32(2.5),
        numpy.float64(1.23),
        numpy.complex64(1+1j),
        numpy.int64(999),
        numpy.uint32(255),
        ctypes.c_int32(50),
        ctypes.c_float(1.5),
        ctypes.c_double(2.7),
        numpy.int8(10),
        numpy.int16(20),
        numpy.uint8(5),
        numpy.uint16(30),
        numpy.float32(0.5),
    ]

    iterations = 50000
    num_runs = 10

    print(f"Configuration:")
    print(f"  - Arguments per launch: {len(kernel_args)}")
    print(f"  - Simulated launches per run: {iterations:,}")
    print(f"  - Number of runs: {num_runs}")
    print(f"  - Total argument processing: {len(kernel_args) * iterations * num_runs:,}")
    print()

    print("Running isinstance() benchmark...")
    isinstance_times = []
    for i in range(num_runs):
        t = benchmark_isinstance(kernel_args, iterations)
        isinstance_times.append(t)
        print(f"  Run {i+1}: {t:.4f}s")
    isinstance_mean = mean(isinstance_times)
    isinstance_stdev = stdev(isinstance_times)
    print(f"  Mean: {isinstance_mean:.4f}s ± {isinstance_stdev:.6f}s")
    print()

    print("Running type() is + isinstance fallback benchmark...")
    type_is_times = []
    for i in range(num_runs):
        t = benchmark_type_is(kernel_args, iterations)
        type_is_times.append(t)
        print(f"  Run {i+1}: {t:.4f}s")
    type_is_mean = mean(type_is_times)
    type_is_stdev = stdev(type_is_times)
    print(f"  Mean: {type_is_mean:.4f}s ± {type_is_stdev:.6f}s")
    print()

    print("=" * 70)
    print("RESULTS")
    print("=" * 70)
    print(f"isinstance():                 {isinstance_mean:.4f}s ± {isinstance_stdev:.6f}s")
    print(f"type() is + isinstance fallback: {type_is_mean:.4f}s ± {type_is_stdev:.6f}s")
    print()

    speedup = isinstance_mean / type_is_mean
    time_saved = isinstance_mean - type_is_mean
    percent_faster = (speedup - 1) * 100

    if speedup > 1.02:
        print(f"✓ type() is is {speedup:.2f}x FASTER ({percent_faster:.1f}% improvement)")
        print(f"  Time saved per 1M launches: {time_saved / (iterations * num_runs) * 1e6 * 1000:.2f}ms")
        print()
        print("RECOMMENDATION: Replace isinstance() with type() is")
    elif speedup < 0.98:
        print(f"⚠️  isinstance() is {1/speedup:.2f}x FASTER")
        print()
        print("RECOMMENDATION: Keep using isinstance()")
    else:
        print(f"≈ Performance is similar (difference < 2%)")
        print()
        print("RECOMMENDATION: Keep using isinstance() for clarity")
    print()

    # Per-launch cost
    total_launches = iterations * num_runs
    print(f"Per-launch argument processing cost:")
    print(f"  isinstance(): {isinstance_mean / total_launches * 1e6:.2f} µs")
    print(f"  type() is:    {type_is_mean / total_launches * 1e6:.2f} µs")
    print()
 ```
 
 I mainly used the compiler flags `-O3` and `-march=native` and compiled and ran the above benchmark via this setup script `setup_benchmark.py`:

 
 ```py
 #!/usr/bin/env python3
"""
Setup script for building the Cython benchmark extension.
"""

from setuptools import setup, Extension
from Cython.Build import cythonize
import numpy

extensions = [
    Extension(
        "benchmark_isinstance_cython",
        ["benchmark_isinstance_cython.pyx"],
        include_dirs=[numpy.get_include()],
        extra_compile_args=["-O3", "-march=native"],
        language="c++",
    )
]

setup(
    name="benchmark_isinstance_cython",
    ext_modules=cythonize(
        extensions,
        compiler_directives={
            'language_level': 3,
            'boundscheck': False,
            'wraparound': False,
            'cdivision': True,
        }
    ),
)
```
The script was then run with `python setup_benchmark.py build_ext --inplace`

**UPDATE:** Updated the benchmark script to include the `isinstance()` fallback and it still results in an **~3x** speedup

## Checklist

### DIFF
--- a/cuda_core/docs/source/release/0.5.x-notes.rst
+++ b/cuda_core/docs/source/release/0.5.x-notes.rst
@@ -16,10 +16,7 @@ None.
 Breaking Changes
 ----------------
 
-- Python ``bool`` objects are now converted to C++ ``bool`` type when passed as kernel
-  arguments.  Previously, they were converted to ``int``.  This brings them inline
-  with ``ctypes.c_bool`` and ``numpy.bool_``.
-
+None.
 
 New features
 ------------
@@ -36,4 +33,6 @@ None.
 Fixes and enhancements
 ----------------------
 
-None.
+- Python ``bool`` objects are now converted to C++ ``bool`` type when passed as kernel
+  arguments.  Previously, they were converted to ``int``.  This brings them inline
+  with ``ctypes.c_bool`` and ``numpy.bool_``.

### CODE REVIEW

**Comment by kkraus14** [comment]
> My only concern with this PR is backward compatibility. It is technically possible to subclass either a numpy or ctypes datatype right now and it would be accepted and work here with the `isinstance` check but would no longer be accepted after this change. I don't know how often that actually happens in practice, and our test suite obviously doesn't do that. I'm not sure how to assess how much we care about this -- it seems hard to do a GitHub code search for, for example. @leofang, thoughts?
> 
> If we determine we do want to be strict about backward compatibility, we could probably do:
> 
> ```
> if arg_type is ctypes_bool:
>    ...
> elif ...
>    ...
> else:
>     # If no exact types are found, fallback to slower `isinstance` check
>     if isinstance(arg_type, ctypes_bool):
>         ...
>     elif:
>         ...
>     else:
>         return 1
> ```
> 
> Note that the fallback cases are in a separate `if/elif/else` block so that Cython can still optimize the outer one to a C switch statement.

Outside of backward compatibility, I think it's reasonable to continue supporting subclasses but to fast path the most common types that we expect users to use, so I'm in support of taking this approach.

**Comment by bharatr21** [comment]
> > My only concern with this PR is backward compatibility. It is technically possible to subclass either a numpy or ctypes datatype right now and it would be accepted and work here with the `isinstance` check but would no longer be accepted after this change. I don't know how often that actually happens in practice, and our test suite obviously doesn't do that. I'm not sure how to assess how much we care about this -- it seems hard to do a GitHub code search for, for example. @leofang, thoughts?
> > If we determine we do want to be strict about backward compatibility, we could probably do:
> > ```
> > if arg_type is ctypes_bool:
> >    ...
> > elif ...
> >    ...
> > else:
> >     # If no exact types are found, fallback to slower `isinstance` check
> >     if isinstance(arg_type, ctypes_bool):
> >         ...
> >     elif:
> >         ...
> >     else:
> >         return 1
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Note that the fallback cases are in a separate `if/elif/else` block so that Cython can still optimize the outer one to a C switch statement.
> 
> Outside of backward compatibility, I think it's reasonable to continue supporting subclasses but to fast path the most common types that we expect users to use, so I'm in support of taking this approach.

To clarify my understanding, would it be enough to do:
```py
if arg_type is ctypes_bool:
   ...
elif isinstance(arg_type, ctypes_bool):
   ...
else:
     return 1
```

or should the check be done manually for all possible subclasses of type `T`, like
```py
if arg_type is T:
   ...
elif arg_type is subclass_1:
   ...
elif arg_type is subclass_2:
   ...
elif isinstance(arg_type, T):
   ...
else:
     return 1
```

bharatr21:
/ok to test [`0fd52b4`](https://github.com/NVIDIA/cuda-python/pull/1292/commits/0fd52b44c343e5e85bd3abaa3be767b641dbb002)

kkraus14:
> To clarify my understanding, would it be enough to do:
> 
> ```python
> if arg_type is ctypes_bool:
>    ...
> elif isinstance(arg_type, ctypes_bool):
>    ...
> else:
>      return 1
> ```
> 
> or should the check be done manually for all possible subclasses of type `T`, like
> 
> ```python
> if arg_type is T:
>    ...
> elif arg_type is subclass_1:
>    ...
> elif arg_type is subclass_2:
>    ...
> elif isinstance(arg_type, T):
>    ...
> else:
>      return 1
> ```

The subclasses would likely be user created / controlled, so we can't possibly cover them all, so we should go with the first option. If we find there's some kind of common subclass then it's something we could explore fast pathing in the future.

leofang:
I am not too worried about backward compatibility in this case, because it is rare in practice that someone would e.g. subclass `np.int32`. I have never seen such usage. I am fine merging this as-is. 

If we want to keep bc, Mike's suggestion on shuffling all existing `isinstance` checks to the very end of the big if-else block makes perfect sense. Do not interleave `is` and `isinstance` for each single type.

bharatr21:
> I am not too worried about backward compatibility in this case, because it is rare in practice that someone would e.g. subclass `np.int32`. I have never seen such usage. I am fine merging this as-is.
> 
> If we want to keep bc, Mike's suggestion on shuffling all existing `isinstance` checks to the very end of the big if-else block makes perfect sense. Do not interleave `is` and `isinstance` for each single type.

So basically retaining this structure, but adding all the `isinstance` checks after the last `elif` block? Like:
```pyx
if arg_type is T1:
 ...
elif arg_type is T2:
 ...
elif arg_type is T3:
 ...
else:
    if isinstance(arg_type, T1):
      ...
    elif isinstance(arg_type, T2):
      ...
    elif isinstance(arg_type, T3):
      ...
    else:
       return 1
```

leofang:
Yes, this was [what Mike suggested](https://github.com/NVIDIA/cuda-python/pull/1292#pullrequestreview-3530207413) 🙂

bharatr21:
/ok to test [`0b9d8ad`](https://github.com/NVIDIA/cuda-python/pull/1292/commits/0b9d8ad5378389b23be95c8329b79cb01eb25e50)

mdboom (changes requested):
Looks like there are 2 test failures here that will need to be addressed.  I'm not exactly sure why they are failing, but let me know if I can help once you've had a chance to look.

bharatr21:
/ok to test [`5c72f9f`](https://github.com/NVIDIA/cuda-python/pull/1292/commits/5c72f9f08502576d4442da4e7c16dc00a708ea25)

@mdboom let me know how I can kick off the CI/CD test runs. If it's intentional that only maintainers/owners do it, then that's ok

leofang:
/ok to test [5c72f9f](https://github.com/NVIDIA/cuda-python/pull/1292/commits/5c72f9f08502576d4442da4e7c16dc00a708ea25)

bharatr21:
@leofang could you please start one more CI/CD run?

leofang (changes requested):
I don't think commit [ed81c11](https://github.com/NVIDIA/cuda-python/pull/1292/commits/ed81c11fb5e5f990c1bd5fe353476b9713cee1be) is correct. The bool treatment should not be changed. @bharatr21 have you been able to run the test suite locally and confirm everything passes before pushing?

bharatr21:
> I don't think commit [ed81c11](https://github.com/NVIDIA/cuda-python/pull/1292/commits/ed81c11fb5e5f990c1bd5fe353476b9713cee1be) is correct. The bool treatment should not be changed. @bharatr21 have you been able to run the test suite locally and confirm everything passes before pushing?

It looks like this kernel logic is treating the Python primitive `bool` also to be 4 bytes, unlike `numpy_bool` or `ctypes_bool` which is why this change is necessary. I don't have CUDA hardware locally so I was unable to run the tests locally, I'm trying to setup some cloud GPU environment so I can run it now.

leofang:
Let us know if you have trouble setting up an environment!

I tried bisecting locally and it seems the test failures in `test_graph.py` started from the first commit 874b7f4e4414786eae2ab7d3edb99460c3027ea8, but it is not obvious to me what could go wrong. Either Mike or I can take a look tomorrow if you haven't found the root cause by then already 🙂

bharatr21:
> Let us know if you have trouble setting up an environment!
> 
> I tried bisecting locally and it seems the test failures in `test_graph.py` started from the first commit [874b7f4](https://github.com/NVIDIA/cuda-python/commit/874b7f4e4414786eae2ab7d3edb99460c3027ea8), but it is not obvious to me what could go wrong. Either Mike or I can take a look tomorrow if you haven't found the root cause by then already 🙂

May I get some guidance on how to setup the environment on a system without GPUs? 
Or does that defeat the purpose :sweat_smile:
 
I'm also slightly surprised/confused, as all CI checks here seem to have passed since [`ed81c11`](https://github.com/NVIDIA/cuda-python/pull/1292/commits/ed81c11fb5e5f990c1bd5fe353476b9713cee1be)

mdboom:
The old set of checks was like this:

```
            if isinstance(arg, Buffer):
                ...
            elif isinstance(arg, int):
                ...
            elif isinstance(arg, float):
                ...
            elif isinstance(arg, complex):
                ...
            elif isinstance(arg, bool):
                ...
```

Since `isinstance(True, int) == True` (surprising, but true!), it is impossible to hit the last branch here (it will always hit `isinstance(arg, int)` first), and the code in the bool branch was probably never tested.

This PR moves this to:

```
            if arg_type is Buffer:
                ...
            elif arg_type is bool:
                ...
            elif arg_type is int:
                ...
            elif arg_type is float:
                ...
            elif arg_type is complex:
                ...
```

The ordering has changed, but the ordering also should no longer matter because it's now doing exact type checks.

When Cython converts a Python bool to C, it is defined as being an `int` [here](https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#types): "The special bint type is used for C boolean values (int with 0/non-0 values for False/True)"

So, I think @bharatr21's change here is correct.  I played a bit with specifying it this way:

```
prepare_arg[bint](self.data, self.data_addresses, arg, i)
```

but this doesn't work. `bint` is sort of a "special" type in Cython and doesn't behave like the other integer typedefs.  Likewise, since `bint` is defined as "whatever C `int` is" (which may not always be 32 bits, even though it is on today's mainstream platforms). I tried:

```
prepare_arg[int](self.data, self.data_addresses, arg, i)
```

but that also doesn't work.  I can't find a way to specify "C int" from Cython.  However, `uint32_t` is [defined as](https://github.com/cython/cython/blob/master/Cython/Includes/libc/stdint.pxd#L15) `ctypedef unsigned int   uint32_t` by Cython, so what @bharatr21 has here is probably the best we can do:

```
prepare_arg[uint32_t](self.data, self.data_addresses, int(arg), i)
```

The only small tweak to make is that it seems that the explicit Python cast from a `bool` to an `int` is not necessary.

cuda_core/cuda/core/experimental/_kernel_arg_handler.pyx
```diff
@@ -207,43 +263,68 @@ cdef class ParamHolder:
...
                     # see note below on handling int arguments
                     prepare_arg[intptr_t](self.data, self.data_addresses, arg.handle, i)
                     continue
                 else:
                     # it's a CUdeviceptr:
                     self.data_addresses[i] = <void*><intptr_t>(arg.handle.getPtr())
                 continue
-            elif isinstance(arg, int):
+            elif arg_type is bool:
+                prepare_arg[int32_t](self.data, self.data_addresses, int(arg), i)
```
mdboom:
```suggestion
                prepare_arg[int32_t](self.data, self.data_addresses, arg, i)
```

cuda_core/cuda/core/experimental/_kernel_arg_handler.pyx
```diff
@@ -207,43 +263,68 @@ cdef class ParamHolder:
...
+                # If no exact types are found, fallback to slower `isinstance` check
+                elif isinstance(arg, Buffer):
+                    if isinstance(arg.handle, int):
+                        prepare_arg[intptr_t](self.data, self.data_addresses, arg.handle, i)
+                        continue
+                    else:
+                        self.data_addresses[i] = <void*><intptr_t>(arg.handle.getPtr())
+                        continue
+                elif isinstance(arg, bool):
+                    prepare_arg[int32_t](self.data, self.data_addresses, int(arg), i)
```
mdboom:
```suggestion
                    prepare_arg[int32_t](self.data, self.data_addresses, arg, i)
```

mdboom:
> May I get some guidance on how to setup the environment on a system without GPUs?
> Or does that defeat the purpose 😅

Yeah, I think unfortunately we can't really test this without a GPU.  There may be other suggestions, but you can get 30 days/$200 free credits on Azure to at least get started.  Not a long term solution, though.

cuda_core/cuda/core/experimental/_kernel_arg_handler.pyx
leofang: It **must** be using `cpp_bool` as before
```suggestion
                prepare_arg[cpp_bool](self.data, self.data_addresses, arg, i)
```
because the whole idea of the launch interface is to ensure a Python `bool` on the host side maps to the C++ bool on the device side, and users don't have to think or look up the docs. If it does not work, then it is a bug that is exposed by this PR and we should use the opportunity to fix it.

cuda_core/cuda/core/experimental/_kernel_arg_handler.pyx
leofang: ditto
```suggestion
                    prepare_arg[cpp_bool](self.data, self.data_addresses, arg, i)
```

leofang:
> Since `isinstance(True, int) == True` (surprising, but true!), it is impossible to hit the last branch here (it will always hit `isinstance(arg, int)` first), and the code in the bool branch was probably never tested.

Ah! I was playing around with this last night, but for some reason I missed checking this combination. Good catch, Mike!

> The only small tweak to make is that it seems that the explicit Python cast from a `bool` to an `int` is not necessary.

That's right. The `prepare_arg` helper is designed to do the type casting correctly. IIRC in order to prepare for kernel launch we already have to cast each arg from Python to C++ and we should not add another cast.

cuda_core/cuda/core/experimental/_kernel_arg_handler.pyx
mdboom: Yes, I think if we update the test (which currently takes an `int` where a Python `bool` is being passed) and add a breaking change note, we should fix this bug.  I will add a commit to this PR to show what I mean.

mdboom:
I've added commit 0fc1ed8d12161f8dbefca68f9bf9cf55f85fe637 here to fix the failing test (so it expects a bool) and added a changelog comment (since that's technically a breaking change).  It remains a mystery to me why this test, testing passing bool values to a kernel, didn't break. https://github.com/NVIDIA/cuda-python/blob/c860f3f9a5420779b192f0969dcae69b2b1e6c28/cuda_core/tests/test_launcher.py#L157  Maybe there is something different about the argument passing to a `set_handle` function vs. an actual kernel.

mdboom:
> Maybe there is something different about the argument passing to a `set_handle` function vs. an actual kernel.

I figured out why this is the case.

On `main`:

- `test_graph_conditional_if` was passing a 4-byte bool to a 4-byte int.  This works because they match.
- `test_launch_scalar_argument` was passing a 4-byte bool to a 1-byte bool.  This works because the first byte of a little-endian 4-byte bool will always have the same value as a 1-byte bool

On this `PR` (without fixing the single test):

- `test_graph_conditional_if` was passing a 1-byte bool to a 4-byte int.  If there is any non-zero garbage in the 3 bytes following the 1-byte passed in, a False could read as True.
- `test_launch_scalar_argument` was passing a 1-byte bool to a 1-byte bool.  This works because they match.

cuda_core/tests/test_graph.py
```diff
@@ -41,7 +43,9 @@ def _common_kernels_conditional():
                                                                                     unsigned int value);
     __global__ void empty_kernel() {}
     __global__ void add_one(int *a) { *a += 1; }
-    __global__ void set_handle(cudaGraphConditionalHandle handle, int value) { cudaGraphSetConditional(handle, value); }
+    __global__ void set_handle(cudaGraphConditionalHandle handle, bool value) {
```
leofang: I think this was already wrong 🙃 `cudaGraphSetConditional` takes an `unsigned int`
```suggestion
    __global__ void set_handle(cudaGraphConditionalHandle handle, unsigned int value) {
```

cuda_core/cuda/core/experimental/_kernel_arg_handler.pyx
```diff
@@ -207,43 +263,68 @@ cdef class ParamHolder:
...
-                continue
 
             not_prepared = prepare_numpy_arg(self.data, self.data_addresses, arg, i)
             if not_prepared:
                 not_prepared = prepare_ctypes_arg(self.data, self.data_addresses, arg, i)
             if not_prepared:
                 # TODO: revisit this treatment if we decide to cythonize cuda.core
-                if isinstance(arg, driver.CUgraphConditionalHandle):
+                if arg_type is driver.CUgraphConditionalHandle:
+                    prepare_arg[intptr_t](self.data, self.data_addresses, <intptr_t>int(arg), i)
```
leofang: Looks like in the header it's a typedef of `unsigned long long`, let's be explicit:
```suggestion
                    prepare_arg[unsigned long long](self.data, self.data_addresses, arg, i)
```
alternatively, I think this should work too
```suggestion
                    prepare_arg[cydriver.CUgraphConditionalHandle](self.data, self.data_addresses, arg, i)
```

cuda_core/cuda/core/experimental/_kernel_arg_handler.pyx
```diff
@@ -207,43 +263,68 @@ cdef class ParamHolder:
...
+                    prepare_arg[intptr_t](self.data, self.data_addresses, arg, i)
+                    continue
+                elif isinstance(arg, float):
+                    prepare_arg[double](self.data, self.data_addresses, arg, i)
+                    continue
+                elif isinstance(arg, complex):
+                    prepare_arg[cpp_double_complex](self.data, self.data_addresses, arg, i)
+                    continue
+                elif isinstance(arg, driver.CUgraphConditionalHandle):
                     prepare_arg[intptr_t](self.data, self.data_addresses, <intptr_t>int(arg), i)
```
leofang: ditto

cuda_core/tests/test_graph.py
```diff
@@ -216,7 +220,9 @@ def test_graph_capture_errors(init_cuda):
     gb.end_building().complete()
 
 
-@pytest.mark.parametrize("condition_value", [True, False])
+@pytest.mark.parametrize(
+    "condition_value", [True, False, ctypes.c_bool(True), ctypes.c_bool(False), np.bool_(True), np.bool_(False)]
```
leofang: 👍

leofang:
Thanks, @mdboom @bharatr21. Looks like this PR is now both an enhancement (better perf) and bug fix (exposing two type casting bugs) 😄

cuda_core/tests/test_graph.py
mdboom: Yes, that's true, and we obviously can't change `cudaGraphSetConditional`.  But if we change this back to a 4-byte int, when a 1-byte bool is passed to it in the test, the test fails.  If we want to make this change, we also would need to change the type of the values passed to it.

```
@pytest.mark.parametrize(
    "condition_value", [True, False, ctypes.c_bool(True), ctypes.c_bool(False), np.bool_(True), np.bool_(False)]
)
```

to

```
@pytest.mark.parametrize(
    "condition_value", [1, 0]
)
```

Personally, I think it's better to test that bools are working correctly then to match the API of `cudaGraphSetConditional`.  Though we could test it with both `int`s and `bool`s (and have a separate definition of `set_handle` for each case).

cuda_core/cuda/core/experimental/_kernel_arg_handler.pyx
mdboom: The first option doesn't work because the Cython parser can't handle `unsigned long long` in that place.  It would have to be some sort of typedef to `unsigned long long`.  But I like the second option a lot better anyway -- seems more explicit about what it needs to match.

mdboom:
> Thanks, @mdboom @bharatr21. Looks like this PR is now both an enhancement (better perf) and bug fix (exposing two type casting bugs) 😄

My last commit improves the tests to test creating conditional handles in both ways: `int` and `bool`.  The types being sent need to match the type specified in the argument to `set_handle`, so the kernel is generated both ways depending on the input type.

I think to strictly follow what upstream cudaGraphConditionalHandle does, `set_handle` should take an `unsigned int`, and then if the user wants to pass a bool they would need to cast it to an int like so:

```
condition_value = True
...
launch(gb, LaunchConfig(grid=1, block=1), set_handle, handle, int(condition_value))
```

...and this is now being tested.

cuda_core/cuda/core/experimental/_kernel_arg_handler.pyx
leofang: > The first option doesn't work because the Cython parser can't handle `unsigned long long` in that place.

Ah yes. Need to add it to the list under the `supported_type` fused type. But I agree the 2nd option is better.

cuda_core/cuda/core/experimental/_kernel_arg_handler.pyx
leofang: @mdboom I think the fix was applied to the wrong place and this line is still unfixed 😅

cuda_core/cuda/core/experimental/_kernel_arg_handler.pyx
```diff
@@ -207,43 +264,68 @@ cdef class ParamHolder:
...
         for i, arg in enumerate(kernel_args):
-            if isinstance(arg, Buffer):
+            arg_type = type(arg)
+            if arg_type is Buffer:
                 # we need the address of where the actual buffer address is stored
-                if isinstance(arg.handle, int):
+                if type(arg.handle) is int:
                     # see note below on handling int arguments
-                    prepare_arg[intptr_t](self.data, self.data_addresses, arg.handle, i)
+                    prepare_arg[cydriver.CUgraphConditionalHandle](self.data, self.data_addresses, arg.handle, i)
```
leofang: ditto, this change is misplaced (Buffer.handle should be cast to `intptr_t` not the driver type)

leofang (changes requested):

cuda_core/cuda/core/experimental/_kernel_arg_handler.pyx
leofang:
```suggestion
                    prepare_arg[intptr_t](self.data, self.data_addresses, arg.handle, i)
```

cuda_core/tests/test_graph.py
leofang: The new test does seem to be more robust! Thanks, Mike.

cuda_core/cuda/core/experimental/_kernel_arg_handler.pyx
leofang:
```suggestion
                    prepare_arg[cydriver.CUgraphConditionalHandle](self.data, self.data_addresses, arg, i)
```

cuda_core/cuda/core/experimental/_kernel_arg_handler.pyx
```diff
@@ -128,67 +129,123 @@ cdef inline int prepare_ctypes_arg(
...
         return prepare_arg[double](data, data_addresses, arg, idx)
-    elif isinstance(arg, numpy_complex64):
+    elif arg_type is numpy_complex64:
         return prepare_arg[cpp_single_complex](data, data_addresses, arg, idx)
-    elif isinstance(arg, numpy_complex128):
+    elif arg_type is numpy_complex128:
         return prepare_arg[cpp_double_complex](data, data_addresses, arg, idx)
     else:
-        return 1
+        # If no exact types are found, fallback to slower `isinstance` check
```
leofang, line 218: Note: this block might add penalty to ctypes and graph conditional handle dispatch, since they are tried after NumPy. I dunno how much this costs, though.

cuda_core/docs/source/release/0.5.x-notes.rst
```diff
@@ -0,0 +1,39 @@
...
+
+None.
+
+
+Breaking Changes
+----------------
+
+- Python ``bool`` objects are now converted to C++ ``bool`` type when passed as kernel
+  arguments.  Previously, they were converted to ``int``.  This brings them inline
+  with ``ctypes.c_bool`` and ``numpy.bool_``.
```
leofang: Q: I think this no longer applies? Should we just add another entry to mention the perf improvement

leofang:
/ok to test [d9a7c14](https://github.com/NVIDIA/cuda-python/pull/1292/commits/d9a7c14e6c97f414be68c69186e3fb01301ce890)

bharatr21:
I updated the branch from `main` and all CI checks passed for the last commit. 
@leofang are we waiting/blocked with something else or can it be merged once the latest round of CI checks pass?

leofang:
Hi @bharatr21 I'd like to let @mdboom take a look at the question above (https://github.com/NVIDIA/cuda-python/pull/1292#discussion_r2594511906), and then we can merge.

cuda_core/docs/source/release/0.5.x-notes.rst
mdboom: No, it still applies.  That is why an update to the test kernel generated in `_common_kernels_conditional` was required -- the Python `bool` is now passed as a C++ `bool`, rather than an `int32_t` as it was before this change.

leofang: Ahhhh ok. Let us document this as a bug fix instead of breaking change. It was not designed to be casting a Python bool to c++ int32. We can wait until the CI finishes, change the rel notes, and admin-merge.

leofang:
Many thanks @bharatr21 @mdboom!

PR #1293, opened 2025-12-01
test: replace cuda driver, device, and context setup with fixtures where relevant

cpcloud:
Factor out repetitive cuda setup in `cuda_bindings` into pytest fixtures.

### DIFF
--- a/cuda_bindings/tests/conftest.py
+++ b/cuda_bindings/tests/conftest.py
@@ -18,7 +18,7 @@ def device(cuda_driver):
     return device
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture(scope="module", autouse=True)
 def ctx(device):
     # Construct context
     err, ctx = cuda.cuCtxCreate(None, 0, device)
--- a/cuda_bindings/pixi.lock
+++ b/cuda_bindings/pixi.lock
@@ -962,7 +962,7 @@ packages:
   - python_abi 3.14.* *_cp314
   license: LicenseRef-NVIDIA-SOFTWARE-LICENSE
   input:
-    hash: 225edd459102d1f609dc61be2335826c3aaec36d76fb657faf6559efe1937aca
+    hash: 551bbbc879e5fefd687d45528e4876eb2383a17c2d292c200e92b369eead4289
     globs:
     - pyproject.toml
 - conda: .
@@ -985,7 +985,7 @@ packages:
   - python_abi 3.14.* *_cp314
   license: LicenseRef-NVIDIA-SOFTWARE-LICENSE
   input:
-    hash: 225edd459102d1f609dc61be2335826c3aaec36d76fb657faf6559efe1937aca
+    hash: 551bbbc879e5fefd687d45528e4876eb2383a17c2d292c200e92b369eead4289
     globs:
     - pyproject.toml
 - conda: .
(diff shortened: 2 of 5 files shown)


cuda_bindings/tests/conftest.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+@pytest.fixture(scope="module")
+def ctx(device):
+    # Construct context
+    err, ctx = cuda.cuCtxCreate(None, 0, device)
+    assert err == cuda.CUresult.CUDA_SUCCESS
+    yield ctx
+    (err,) = cuda.cuCtxDestroy(ctx)
+    assert err == cuda.CUresult.CUDA_SUCCESS
```
kkraus14: Just a note that this has a side effect on tests without this fixture as calling `cuCtxCreate` pushes the created context onto the calling CPU thread as the current context, which will be used for all driver and runtime APIs.

Maybe it makes sense to make this an autouse fixture or something?

mdboom: Would `scope='function'` help here?  (Then it will be a new context for each scope, cleanly cleaning up after itself).

kkraus14: Potentially, but if I remember correctly context creation and destruction is quite expensive, where doing it at a function scope may blow up test time.

cpcloud: Module scope was intentional to speed up tests. It's quite a large improvement.

cpcloud: If a test isn't using that fixture then why would it matter whether there's some state changing?

cpcloud: Is any of this blocking?

If the test suite is passing then we should trust that and not get bogged down on trying to fix something that isn't a demonstrable problem.

I'm not saying this is perfect or even 100% foolproof, but I would like to get to +1 or -1 as is, instead of changing things without strong motivation.

cpcloud: That said, I am happy to move to function scope if that moves things along.

kkraus14: > If a test isn't using that fixture then why would it matter whether there's some state changing?

Because even though it isn't using the fixture, it is implicitly using the context that is created and set to current by the fixture. I.E. some errors are "sticky" and more or less poison the Context and will continue to return errors for all operations.

I'm +1 to moving this forward, but trying to think through the possible implications while concerns like these are top of mind.

kkraus14: Final thought, I think we should make this an autouse fixture given essentially everything interacts with the CUDA context either explicitly or implicitly

cpcloud: Happy to make it an autouse fixture. That will reduce the amount of `usefixtures` markers as well.

cuda_bindings/tests/conftest.py
```diff
@@ -0,0 +1,28 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
+
+import cuda.bindings.driver as cuda
+import pytest
+
+
+@pytest.fixture(scope="module")
+def cuda_driver():
+    (err,) = cuda.cuInit(0)
```
leofang, line 10: FWIW `cuInit` cannot be undone once it's called. Might as well make it `scope='session'`.

cpcloud: Sounds good, I will put up a follow-up for that in a bit.

leofang:
LGTM too! Thanks, Phillip!

cpcloud:
This should help us be more confident about pytest-randomly usage inside cuda-bindings. I will look into factoring out fixtures where possible in cuda-core as well.

PR #1298, opened 2025-12-02
cuda-core experimental namespace deprecation (attempt v0)

rwgk:
This was the initial attempt — continuing the work under #1377 (starting over fresh from `main`)

### DIFF
--- a/ci/tools/merge_cuda_core_wheels.py
+++ b/ci/tools/merge_cuda_core_wheels.py
@@ -107,15 +107,19 @@ def merge_wheels(wheels: List[Path], output_dir: Path) -> Path:
 
             # Copy only version-specific binaries (.so, .pyd, .dll files) from the source wheel
             # Python modules (.py, .pyx, .pxd) remain in cuda/core/
+            # Exclude versioned directories (cu12/, cu13/) to avoid recursion
             source_dir = wheel_dir / base_dir
             for item in source_dir.rglob("*"):
                 if item.is_dir():
                     continue
 
+                # Skip files in versioned directories to avoid recursion
+                rel_path = item.relative_to(source_dir)
+                if any(part in ("cu12", "cu13") for part in rel_path.parts):
+                    continue
+
                 # Only copy binary files, not Python source files
                 if item.suffix in (".so", ".pyd", ".dll"):
-                    # Preserve directory structure relative to base_dir
-                    rel_path = item.relative_to(source_dir)
                     dest_item = versioned_dir / rel_path
                     dest_item.parent.mkdir(parents=True, exist_ok=True)
                     shutil.copy2(item, dest_item)


cuda_core/tests/test_experimental_backward_compat.py
```diff
@@ -0,0 +1,157 @@
...
+    """Test that imports from experimental namespace still work."""
+    with warnings.catch_warnings(record=True) as w:
+        warnings.simplefilter("always")
+        
+        # Test main module import
+        import cuda.core.experimental
+        
+        # Should emit deprecation warning
+        assert len(w) >= 1
+        assert issubclass(w[0].category, DeprecationWarning)
```
kkraus14: We should be able to leverage `with pytest.deprecated_call` https://docs.pytest.org/en/stable/reference/reference.html#pytest.deprecated_call

cuda_core/cuda/core/experimental/__init__.py
```diff
@@ -48,11 +85,30 @@
...
+
+# Also create forwarding stubs for submodules
+# These will be imported lazily when accessed
+def __getattr__(name):
+    """Forward attribute access to the new location with deprecation warning."""
+    if name in ("_device", "_event", "_graph", "_launch_config", "_launcher", 
+                "_linker", "_memory", "_module", "_program", "_stream", "_system", 
+                "_utils", "_context", "_dlpack", "_kernel_arg_handler", 
+                "_launch_config", "_memoryview"):
+        _warn_deprecated()
```
kkraus14, line 100: Does this trigger a warning on each call as opposed to just at import time?

rwgk:
Remark: I haven't looked at **any** of the changes myself yet. Everything to this point was done by cursor-agent. I just gave it a few prompts to explain what I want to do.

cuda_core/tests/test_experimental_backward_compat.py
rwgk: cursor-agent:

Good catch! I've updated the tests to use `pytest.deprecated_call()` instead of manual warning catching. This is cleaner and more idiomatic pytest usage.
___

human:

I'll work in this auto-pilot mode until I get everything passing here (CI). Then I'll review myself.

cuda_core/cuda/core/experimental/__init__.py
rwgk: cursor-agent:

The deprecation warning is only emitted once at import time when `cuda.core.experimental` is first imported, not on each attribute access. This is because:
- The warning is emitted at module initialization (line 64: `_warn_deprecated()`)
- utils is already set as an attribute in the module namespace (line 68), so accessing it doesn't go through `__getattr__`
- Only accessing submodules via `__getattr__` (like `_device`, `_utils`) triggers additional warnings

I've added documentation to clarify this behavior.

Andy-Jost: 1. The call to `_warn_deprecated` in this function will never cause a warning to be issued because the warning was already issued at module scope when loading this module. It should be removed.
2. Do we really intend to publish all submodules by stripping the leading underscores? That would make public many things that are currently private.
3. This implementation pays for the extra attribute look-up with each access. What about something like the following?

```
            # Fallback to underscore-prefixed name
            attr = importlib.import_module(f"mymodule.{name}")
            # Set the attribute globally to avoid __getattr__ next time
            setattr(sys.modules[__name__], name, attr)
            # alternative: globals()[name] = attr
            return attr
```

However, at that point, it might be better to just say this:

```from .. import _device, _event, # etc.```

**Comment by Andy-Jost** [comment]
Another approach would be to change absolute imports to relative, e.g.:

```
- from cuda.core.experimental._utils.cuda_utils import driver
+ from ._utils.cuda_utils import driver
```

An advantage of this is that it decouples the code changes from the file movements, so it could be checked into the current main without eliminating the `experimental` namespace. After that, the files could be moved without changing any import statements.

I've been planning to propose this change but haven't had time to ask the team for input yet.

**Comment by leofang** [comment]
Usually absolute import is encouraged (ex: PEP 8 and Ruff's [TID252](https://docs.astral.sh/ruff/rules/relative-imports/) rule). I have a slight preference for abs import because it gives a clear view on what comes from where. But I don't feel strongly about this.

**Comment by Andy-Jost** [comment]
From PEP 8:

> However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose.

My 2 cents regarding the package structure:

- We have a "complex package layout," particularly as we consider namespace migration.
- Absolute imports being "unnecessarily verbose" also applies in our situation (for example, `cuda.core.experimental` is quite long).

This is why I slightly favor relative imports here, particularly as a way to simplify the namespace migration. However, I also don't feel strongly on this point overall.

**Comment by rwgk** [comment]
Regarding absolute vs relative imports:

I've been following the direct guidance of @kkraus14, which he gave me/us around the time when I started working on cuda-pathfinder (~spring this year). It was that we should use absolute imports throughout.

For things like this, I usually put my personal preference aside completely. There will always be valid arguments one way or the other. In the end what I think is most valuable is consistency, ideally as enforced by tooling (with overrides that must be explicit, e.g. `noqa`).

I'd be a big fan of adding [TID252](https://docs.astral.sh/ruff/rules/relative-imports/) to our ruff.toml, and then agree that everything that passes that check is fine.

File: cuda_core/cuda/core/experimental/__init__.py (hunk 2 above)
Comment (leofang) [reply cuda_core/cuda/core/experimental/__init__.py:100]: I prefer strongly that we just remove this entirely. Underscored modules are never considered public APIs.

File: cuda_core/cuda/core/experimental/__init__.py [hunk 3]
```diff
@@ -2,43 +2,90 @@
...
+        Program,
+        ProgramOptions,
+        Stream,
+        StreamOptions,
+        VirtualMemoryResource,
+        VirtualMemoryResourceOptions,
+        launch,
+        system,
+        utils,
+    )
```
leofang: Q: Why do we need this new block, when they are already imported below?

cuda_core/cuda/core/include/utility.hpp
leofang, line 1: We probably need to document that everything under `cuda_core/cuda/core/include/` are considered internal (for now). There is currently no plan to maintain any non-Python (such as C, C++, and Cython) public APIs in cuda-core, and when we do we will revise this document.

cuda_core/cuda/core/_memory/_buffer.pyx
```diff
@@ -8,18 +8,18 @@ cimport cython
...
 
 from cuda.bindings cimport cydriver
-from cuda.core.experimental._memory._device_memory_resource cimport DeviceMemoryResource
-from cuda.core.experimental._memory._ipc cimport IPCBufferDescriptor, IPCDataForBuffer
-from cuda.core.experimental._memory cimport _ipc
-from cuda.core.experimental._stream cimport Stream_accept, Stream
-from cuda.core.experimental._utils.cuda_utils cimport HANDLE_RETURN
+from cuda.core._memory._device_memory_resource cimport DeviceMemoryResource
+from cuda.core._memory._ipc cimport IPCBufferDescriptor, IPCDataForBuffer
+from cuda.core._memory cimport _ipc
```
rparolin, line 13: Should we be removing the prefix underscore from the module name? For example, `cuda.core._memory` -> `cuda.core.memory`.

cuda_core/cuda/core/experimental/__init__.py
```diff
@@ -48,11 +95,47 @@
...
+
+# Also create forwarding stubs for submodules
+# These will be imported lazily when accessed
+def __getattr__(name):
+    """Forward attribute access to the new location with deprecation warning."""
+    if name in (
+        "_device",
+        "_event",
+        "_graph",
+        "_launch_config",
```
rparolin, line 88: `_launch_config` is listed twice in the list.

rparolin: Lines: 116, and 128.

cuda_core/cuda/core/experimental/__init__.py
rwgk: Fixed: commit 798ce3d4259dd42cde334076e373921c61067df7

cuda_core/cuda/core/experimental/__init__.py
rwgk: Fixed: commit 4b9796b8f7a1a51682eb43adeb167ea150f17e3e

Amazing that you spotted this. I sorted the list alphabetically now, to make it less likely that we're missing duplicates.

cuda_core/cuda/core/__init__.py
```diff
@@ -3,3 +3,59 @@
...
+from cuda.core._memory import (  # noqa: E402
+    Buffer,
+    DeviceMemoryResource,
+    DeviceMemoryResourceOptions,
+    GraphMemoryResource,
+    LegacyPinnedMemoryResource,
+    MemoryResource,
+    VirtualMemoryResource,
+    VirtualMemoryResourceOptions,
+)
```
leofang, line 53: This list needs to be updated to capture a few new things.

cuda_core/cuda/core/__init__experimental.pxd
leofang, line 1: Cython needs an exact filename `__init__.pxd` so as to find the current modules. This renaming does not make sense to me. What does it do?

cuda_core/cuda/core/_memory/_buffer.pyx
leofang: No. We don't want these modules to become public APIs. Tab auto-completion should be kept clean.

cuda_core/build_hooks.py
```diff
@@ -84,11 +84,16 @@ def get_cuda_paths():
         print("CUDA paths:", CUDA_PATH)
         return CUDA_PATH
 
+    # Add local include directory for cuda/core/include
+    local_include_dirs = ["cuda/core"]
```
leofang, line 88: Q: Why is this needed? We do not keep any headers under `cuda/core` or `cuda/core/experimental`. If without adding this Cython runs into compilation errors, I suspect this has to do with the `__init_experimental.pxd` file below.

cuda_core/cuda/core/__init__experimental.pxd
rwgk: Probably an accident.

I had to git merge main already several times, a likely cause of accidents. The merges are messy. I'm using Cursor and it seems to miss some details sometimes.

This PR has become much more complex than it looked like initially (tests passed a couple weeks ago).

E.g. I don't understand why ci/tools/merge_cuda_core_wheels.py (last changed in Oct) worked before but now needs a change (it makes sense that it needs a change). Testing it locally isn't easy because the input wheels only live on the runners and disappear when the jobs finish. So I'd have to figure out how to build them locally (currently I'm not trying to).

I'll keep beating on it until I'll get all tests to pass again. Maybe I'll start over fresh. LLMs seems to be less successful when working incrementally.

Thanks for the hints.

leofang: Ah! Leaving a quick note before signing off... 

The merge script can be tested locally easily. Create two conda envs, one with CUDA 13 and another with CUDA 12, and build cuda.core wheels in each env via `pip wheel --no-deps --no-build-isolation -v .`. The `cuda-bindings` package can just be conda-installed together with the CTK for simplicity. No need to build it from source. Then, we take the two wheels and rename them before running the merge script. This logic is relevant:
https://github.com/NVIDIA/cuda-python/blob/83eaec1061b22c386b38b754ce859d3811176ed8/.github/workflows/build-wheel.yml#L430-L439 

Once this PR (https://github.com/NVIDIA/cuda-python/pull/1085) is revived/merged, we will be able to build cuda-core in a pure-wheel env and no need to use conda.

rwgk:
I'm continuing the work under the new PR #1377 

I think it was the right decision to start over. Many things got lost in the repeated `git merge main` steps here:

```
Summary of Differences: cuda_core_experimental_deprecation_v0 vs cuda_core_experimental_deprecation_v1
========================================================================================================

This summary compares the cuda_core/ directory between the two branches to identify
important changes that might need to be transferred from v0 to v1.

KEY DIFFERENCES IDENTIFIED:
============================

1. **StridedLayout Naming**
   - v0: Uses public `StridedLayout` class
   - v1: Uses private `_StridedLayout` class (with leading underscore)
   - Impact: This is intentional - v1 follows main branch which privatized StridedLayout
   - Status: v1 is correct (matches main branch)

2. **New Memory Resources (from main)**
   - v1 has: `ManagedMemoryResource`, `ManagedMemoryResourceOptions`, 
             `PinnedMemoryResource`, `PinnedMemoryResourceOptions`
   - v0: Missing these
   - Impact: v1 includes newer features from main branch
   - Status: v1 is ahead (has newer features)

3. **StridedMemoryView Exports**
   - v1: Exports `StridedMemoryView` and `args_viewable_as_strided_memory` from `cuda.core`
   - v0: Missing these exports
   - Impact: v1 makes these utilities available at top level
   - Status: v1 is correct

4. **Device.get_all_devices() Method**
   - v1: Has `Device.get_all_devices()` classmethod
   - v0: Missing this method
   - Impact: v1 provides convenience method to get all devices
   - Status: v1 is ahead (has newer feature)

5. **CUDA_CORE_BUILD_MAJOR Type**
   - v1: Uses `int(_get_proper_cuda_bindings_major_version())` (integer comparison)
   - v0: Uses `_get_proper_cuda_bindings_major_version()` (string comparison)
   - v1: Cython code uses `IF CUDA_CORE_BUILD_MAJOR == 12:` (integer)
   - v0: Cython code uses `IF CUDA_CORE_BUILD_MAJOR == "12":` (string)
   - Impact: Type consistency fix
   - Status: v1 is correct (integer comparison is more appropriate)

6. **Include Directory Path**
   - v1: Uses `_include/` (with leading underscore for implementation detail)
   - v0: Uses `include/` (without underscore)
   - Impact: v1 follows the design decision to mark headers as implementation detail
   - Status: v1 is correct (matches user's requirement)

7. **pyproject.toml Package Data**
   - v1: `"cuda.core._include"` 
   - v0: `"cuda.core.include"`
   - Impact: Matches the _include directory rename
   - Status: v1 is correct

8. **build_hooks.py Comments**
   - v1: Updated comment to reference `_include/` instead of `include/`
   - v0: Older comment
   - Impact: Documentation accuracy
   - Status: v1 is correct

STATISTICS:
==========
- ~131 Python/Cython files changed
- Most changes are due to:
  * File moves (experimental/ -> root)
  * StridedLayout -> _StridedLayout renaming
  * Include path updates (include/ -> _include/)
  * New memory resource types added

CONCLUSION:
===========

The v1 branch appears to be MORE UP-TO-DATE than v0, incorporating:
- New memory resource types (ManagedMemoryResource, PinnedMemoryResource)
- Device.get_all_devices() convenience method
- Proper type handling for CUDA_CORE_BUILD_MAJOR
- Correct include directory naming (_include/)
- Proper exports of StridedMemoryView utilities

The main difference is that v0 still uses public `StridedLayout` while v1 uses private
`_StridedLayout`, but this is intentional as v1 follows the main branch which
privatized this class.

RECOMMENDATION:
===============
No important changes need to be transferred from v0 to v1. The v1 branch is
more complete and up-to-date with main branch changes. The v0 branch appears
to be an older incremental migration attempt, while v1 is a fresh, complete
migration that properly incorporates all the latest changes from main.
```

leofang:
Closing in favor of #1377.

PR #1299, opened 2025-12-02
Allow buffers imported via IPC to be re-exported

Andy-Jost:
### Summary

This PR enhances the CUDA IPC (Inter-Process Communication) functionality to support re-exporting buffers that were originally imported from another process. It also introduces a new `Buffer.is_mapped` property to track whether a buffer was mapped into the current process via IPC.

#### 1. **New `Buffer.is_mapped` Property**
- Added `is_mapped` property to `Buffer` class to indicate if a buffer was imported via IPC
- Returns `True` for buffers imported via `Buffer.from_ipc_descriptor()`
- Returns `False` for locally-allocated buffers

#### 2. **Enable Re-export of Imported Buffers**
- **Removed restriction**: Previously, attempting to export a memory resource that was imported via IPC would raise `RuntimeError("Imported memory resource cannot be exported")`
- Buffers can now be re-exported after being imported, enabling multi-hop IPC chains (e.g., Process A → Process B → Process C)
- IPC descriptors are now cached on first export to avoid redundant operations

#### 3. **IPC Data Structure Improvements**
- Rename `IPCData` into `IPCDataForMR`: Tracks IPC state for memory resources (allocation handle + mapped flag)
- Added `IPCDataForBuffer`: Tracks IPC state for individual buffers (descriptor + mapped flag)
- Added `_ipc_data` member to `Buffer` class to track IPC-related metadata
- IPC descriptors are cached on buffers to support efficient re-export

#### 4. **Enhanced Testing**
- **New test**: `TestIpcReexport` validates the 3-process chain scenario:
  1. Process A allocates a buffer and exports it to Process B
  2. Process B receives the buffer and re-exports it to Process C
  3. Process C receives the buffer, modifies it, and signals completion
  4. Process A verifies that C's modifications are visible
- Added `is_mapped` assertions throughout IPC tests
- Converted `test_ipc_send_buffers` to class-based test structure
- Removed obsolete `TestExportImportedMR` error test (no longer an error)

### DIFF
--- a/cuda_core/cuda/core/experimental/_memory/_buffer.pyx
+++ b/cuda_core/cuda/core/experimental/_memory/_buffer.pyx
@@ -63,7 +63,7 @@ cdef class Buffer:
         self._ptr_obj = ptr
         self._size = size
         self._memory_resource = mr
-        self._ipc_data = IPCDataForBuffer(ipc_descriptor, mapped=True) if ipc_descriptor is not None else None
+        self._ipc_data = IPCDataForBuffer(ipc_descriptor, True) if ipc_descriptor is not None else None
         self._alloc_stream = <Stream>(stream) if stream is not None else None
         return self
 
@@ -103,7 +103,7 @@ cdef class Buffer:
     def get_ipc_descriptor(self) -> IPCBufferDescriptor:
         """Export a buffer allocated for sharing between processes."""
         if self._ipc_data is None:
-            self._ipc_data = IPCDataForBuffer(_ipc.Buffer_get_ipc_descriptor(self), mapped=False)
+            self._ipc_data = IPCDataForBuffer(_ipc.Buffer_get_ipc_descriptor(self), False)
         return self._ipc_data.ipc_descriptor
 
     def close(self, stream: Stream | GraphBuilder | None = None):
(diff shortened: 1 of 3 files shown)


cuda_core/cuda/core/experimental/_memory/_buffer.pyx
```diff
@@ -55,13 +56,14 @@ cdef class Buffer:
...
         cls, ptr: DevicePointerT, size_t size, mr: MemoryResource | None = None,
-        stream: Stream | None = None
+        stream: Stream | None = None, ipc_descriptor: IPCBufferDescriptor | None = None
     ):
         cdef Buffer self = Buffer.__new__(cls)
         self._ptr = <uintptr_t>(int(ptr))
         self._ptr_obj = ptr
         self._size = size
         self._memory_resource = mr
+        self._ipc_data = IPCDataForBuffer(ipc_descriptor, True) if ipc_descriptor is not None else None
```
rwgk, line 66: Optional: This would be more readable with `is_mapped=True`.

I'd do that systematically for all `IPCDataForBuffer` and `IPCDataForMR` calls (I think there are only four calls).

rwgk (approved):
LGTM

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
Andy-Jost: Thanks. I actually had exactly that initially but noticed that Cython passes the keyword arguments as true keywords (in a dict) -- I had hoped that it would perform some magic, since the call target is a cdef function. Within this project, it's hard to know how to trade readability off against performance.

PR #1304, opened 2025-12-03
ci: Replace CI JSON files with YAML

### LINKED ISSUE #1297: CI: Replace json by yaml
With works happening both internally (ex: our codegen) and externally (ex: numba-cuda), it has become clear that [yaml files like this](https://github.com/NVIDIA/numba-cuda/blob/main/ci/matrix.yml) (from numba-cuda) are better than json ones because
- it allows comments
- it avoids weird syntax errors (ex: no comma in the last line)

Let's just unify our infrastructure and always use yaml instead of json. The only drawback that I am aware of is that GHA does not provide a built-in `fromYAML` function (whereas there is [`fromJSON`](https://docs.github.com/en/enterprise-cloud@latest/actions/reference/workflows-and-actions/expressions#fromjson)), but I am sure we can work it around.

bharatr21:
## Description

Closes #1297 

Modify the JSON workflow files to YAML and adjust the references to those files itself (using `yq` instead of `jq` to extract values or converting them to JSON first before calling `jq` or `fromJSON` )

## Checklist

- [x] Documentation update is not required for this change.

### DIFF
--- a/ci/test-matrix.yml
+++ b/ci/test-matrix.yml
@@ -14,7 +14,6 @@
 #
 # Notes:
 # - DRIVER: 'earliest' does not work with CUDA 12.9.1
-# - LOCAL_CTK: 0 does not work with CUDA 12.0.1
 
 linux:
   pull-request:


ci/test-matrix.yml
```diff
@@ -0,0 +1,60 @@
...
+#
+# Please keep the matrices sorted in ascending order by the following:
+#
+# [ARCH, PY_VER, CUDA_VER, LOCAL_CTK, GPU, DRIVER]
+#
+# Windows entries also include DRIVER_MODE.
+#
+# Notes:
+# - DRIVER: 'earliest' does not work with CUDA 12.9.1
+# - LOCAL_CTK: 0 does not work with CUDA 12.0.1
```
bharatr21: Please let me know if all these notes and comments are really necessary

mdboom: I think the second comment (`- LOCAL_CTK: 0 does not work with CUDA 12.0.1`) can be removed since we no longer support `12.0.1`, but the other should remain.

mdboom:
```suggestion
```

PR #1309, opened 2025-12-03
Warn when multiprocessing start method is 'fork'

### LINKED ISSUE #1136: [FEA]: Warn when the multiprocessing start method is "fork"
### Is this a duplicate?

### Area

cuda.core

When using IPC capabilities in `cuda.core`, it is easy to make the mistake of starting new processes by the "fork" method, especially since that is the default for the `multiprocessing` module. When this occurs, child processes fail to construct `Device` objects, generating a CUDA_ERROR_NOT_INITIALIZED error. The solution is to use the "spawn" start method. Users may not be aware of this and the error message does not clearly indicate the problem.

There are key points in `cuda.core` where it is clear that `multiprocessing` is being used. For instance, when calling a reduction (i.e., pickling) method that was registered with `multiprocessing`. In those cases, the library should perform a one-time check of the start method and warn if it is set incorrectly. Note that the start method can only be set once, so a one-time check while serializing is sufficient.

_No response_

### Additional context

_No response_

Andy-Jost:
## Summary

CUDA does not support the fork() system call. Forked subprocesses exhibit undefined behavior, including failure to initialize CUDA contexts and devices. This PR adds warnings when IPC objects are serialized with the fork multiprocessing start method.

## Changes

- Added `_check_multiprocessing_start_method()` function in `cuda_utils.pyx` that checks if the multiprocessing start method is 'fork' and emits a one-time warning
- Updated multiprocessing reduction functions to call the warning check:
  - `_reduce_allocation_handle` in `_ipc.pyx`
  - `_deep_reduce_device_memory_resource` in `_ipc.pyx`
  - `_reduce_event` in `_event.pyx`
- Warning message explains that CUDA doesn't support fork, describes undefined behavior, and recommends using 'spawn' method

## Test Coverage

Added comprehensive test suite `test_multiprocessing_warning.py` with 5 tests:
- Warning emitted for DeviceMemoryResource pickling with fork method
- Warning emitted for IPCAllocationHandle pickling with fork method
- Warning emitted for Event pickling with fork method
- No warning when start method is 'spawn'
- Warning emitted only once (one-time check)

Tests run in subprocesses to avoid interference from conftest.py's session fixture that sets spawn method.

## Related Work

Fixes #1136

### DIFF
--- a/cuda_core/tests/conftest.py
+++ b/cuda_core/tests/conftest.py
@@ -12,6 +12,7 @@ try:
 except ImportError:
     from cuda import cuda as driver
 
+import cuda.core.experimental
 from cuda.core.experimental import Device, DeviceMemoryResource, DeviceMemoryResourceOptions, _device
 from cuda.core.experimental._utils.cuda_utils import handle_return
 
@@ -28,7 +29,7 @@ def session_setup():
 @pytest.fixture(scope="function")
 def init_cuda():
     # TODO: rename this to e.g. init_context
-    device = Device()
+    device = Device(0)
     device.set_current()
 
     # Set option to avoid spin-waiting on synchronization.
@@ -83,7 +84,7 @@ def deinit_all_contexts_function():
 def ipc_device():
     """Obtains a device suitable for IPC-enabled mempool tests, or skips."""
     # Check if IPC is supported on this platform/device
-    device = Device()
+    device = Device(0)
     device.set_current()
 
     if not device.properties.memory_pools_supported:
@@ -113,13 +114,43 @@ def ipc_memory_resource(ipc_device):
 @pytest.fixture
 def mempool_device():
     """Obtains a device suitable for mempool tests, or skips."""
-    device = Device()
+    device = Device(0)
     device.set_current()
 
     if not device.properties.memory_pools_supported:
         pytest.skip("Device does not support mempool operations")
 
     return device
+
+
+def _mempool_device_impl(num):
+    num_devices = len(cuda.core.experimental.system.devices)
+    if num_devices < num:
+        pytest.skip(f"Test requires at least {num} GPUs")
+
+    devs = [Device(i) for i in range(num)]
+    for i in reversed(range(num)):
+        devs[i].set_current()  # ends with device 0 current
+
+    if not all(devs[i].can_access_peer(j) for i in range(num) for j in range(num)):
+        pytest.skip("Test requires GPUs with peer access")
+
+    if not all(devs[i].properties.memory_pools_supported for i in range(num)):
+        pytest.skip("Device does not support mempool operations")
+
+    return devs
+
+
+@pytest.fixture
+def mempool_device_x2():
+    """Fixture that provides two devices if available, otherwise skips test."""
+    return _mempool_device_impl(2)
+
+
+@pytest.fixture
+def mempool_device_x3():
+    """Fixture that provides three devices if available, otherwise skips test."""
+    return _mempool_device_impl(3)
 
 
 skipif_need_cuda_headers = pytest.mark.skipif(helpers.CUDA_INCLUDE_PATH is None, reason="need CUDA header")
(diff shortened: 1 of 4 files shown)


cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
```diff
@@ -283,3 +286,39 @@ class Transaction:
...
+
+# Track whether we've already warned about fork method
+_fork_warning_emitted = False
+
+
+def _check_multiprocessing_start_method():
+    """Check if multiprocessing start method is 'fork' and warn if so."""
+    global _fork_warning_emitted
+    if _fork_warning_emitted:
+        return
```
leofang: Instead of caching it, would it be better to always call `multiprocessing.get_start_method()` and check it?  I worry that it is a global state that we don't own and it could be changed at arbitrary point in time by the user or any package.

Andy-Jost: In the official `multiprocessing` library [docs](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.set_start_method), the description of `multiprocessing.set_start_method()` explains that:

- The start method is a global setting that can only be chosen a single time in a given Python process.
- Calling `set_start_method()` again without forcing will raise a `RuntimeError` indicating that the context has already been set.
- Recent CPython implementations mention an internal/undocumented way to override this with a `force` argument, but this is not part of the public, documented API and should be avoided in normal code.

Andy-Jost:
The previous test methodology of starting a subprocess did not work on the test runners. The latest upload avoids using a subprocess, instead mocking up key methods.

cuda_core/cuda/core/experimental/_utils/cuda_utils.pyx
leofang: Thanks, Andy. Good to know Python by default checks this. However, I do see the `force` argument being documented, so it is part of the public API. And it does allow overwriting:
```python
>>> import multiprocessing
>>> multiprocessing.set_start_method("spawn")
>>> multiprocessing.set_start_method("spawn")
  File "<python-input-2>", line 1, in <module>
    multiprocessing.set_start_method("spawn")
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/local/home/leof/miniforge3/envs/py313_cu130/lib/python3.13/multiprocessing/context.py", line 247, in set_start_method
    raise RuntimeError('context has already been set')
RuntimeError: context has already been set
>>> multiprocessing.set_start_method("fork", force=True)
>>> multiprocessing.set_start_method("spawn", force=True)
>>> multiprocessing.set_start_method("forkserver", force=True)
>>> 
```
If this is not on the hot path, I think not caching the result and always checking is safer. 

Alternatively, we could mention in the warning message that setting `set_start_method(..., force=True`) is dangerous because we cannot help capture issues.

leofang: One more thing, if we already warned the next `warn()` call is a no-op so adding the `_fork_warning_emitted` guard is redundant 😆 
```python
import warnings

def f():
    warnings.warn("oh no", UserWarning, stacklevel=3)

f()  # only this call would generate a warning 
f()
f()
```

leofang: Ah I see, you want to limit the warning to only once per process lifetime, regardless of which object raises the warning from. NVM.

mdboom: On minor detail -- since `set_start_method` can only be set once, I don't think we need to check it multiple times.  In other words, we can set `_fork_warning_emitted = True` unconditionally rather than only when a warning is emitted so we don't check again.  And maybe renaming the flag to `_fork_warning_checked` for clarity.  Unless I'm missing some case where it might be set exceptionally late and it does need to be checked multiple times.

Andy-Jost: Thanks for the comments. I've made adjustments to the name and conditionality as Mike suggested. I left this as a one-time check because there does not seem to be a strong consensus, but I don't feel strongly about that and wouldn't mind changing it if more discussion goes in that direction.

Incidentally, a separate warning is issued when fork is called in multithreaded programs (like those using CUDA), so if a user somehow sidesteps this warning by setting the start method multiple times, they will still get another nasty message indicating their program is invalid.

PR #1313, opened 2025-12-04
chore: clean up cufile fixtures

cpcloud:
# Clean up cuFile test fixtures

## Summary

This PR refactors the cuFile test fixtures to follow pytest best practices, reducing code duplication and improving maintainability. The changes modernize the test infrastructure while maintaining the same test coverage.

### Import Handling
- **Replaced try/except ImportError with `pytest.importorskip`**: More idiomatic pytest approach for handling optional dependencies
- **Moved WSL skip check to pytest mark**: Converted `platform_is_wsl()` function check into a `pytestmark` decorator for better integration with pytest's skip mechanism

### Fixture Improvements
- **`cufile_env_json` fixture**: Now uses `monkeypatch` for environment variable management instead of manual `os.environ` manipulation, ensuring proper cleanup and isolation
- **New `raw_driver()` context manager**: Extracted driver open/close logic into a reusable context manager
- **`driver` fixture refactoring**: Uses the new context manager pattern for cleaner resource management

### Test Cleanup
- **Async test functions**: Removed redundant `driver_open()`/`driver_close()` calls by leveraging the `driver` fixture via `@pytest.mark.usefixtures`
- **Parameter test refactoring**:
  - Extracted slab sizes and counts into reusable fixtures (`slab_sizes`, `slab_counts`)
  - Created `driver_config` fixture to handle parameter configuration
  - Simplified assertions using list comparison instead of manual loops
  - Used context manager pattern for driver lifecycle management

## Benefits

- **Reduced code duplication**: Eliminated repeated `driver_open()`/`driver_close()` calls across multiple tests
- **Better resource management**: Context managers ensure proper cleanup even if tests fail
- **Improved test isolation**: `monkeypatch` ensures environment variables don't leak between tests
- **More maintainable**: Fixtures make it easier to modify test setup in the future
- **Cleaner test code**: Tests focus on their logic rather than setup/teardown boilerplate

## Testing

All existing tests pass with the refactored fixtures. No functional changes to test behavior.

## Statistics

- **1 file changed**: `cuda_bindings/tests/test_cufile.py`
- **62 insertions(+), 91 deletions(-)**: Net reduction of 29 lines while improving code quality

### DIFF
--- a/cuda_bindings/tests/test_cufile.py
+++ b/cuda_bindings/tests/test_cufile.py
@@ -1,7 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
 
-import contextlib
 import ctypes
 import errno
 import logging
@@ -122,19 +121,11 @@ def ctx():
     cuda.cuDevicePrimaryCtxRelease(device)
 
 
-@contextlib.contextmanager
-def raw_driver():
-    cufile.driver_open()
-    try:
-        yield
-    finally:
-        cufile.driver_close()
-
-
 @pytest.fixture
 def driver(ctx):
-    with raw_driver():
-        yield
+    cufile.driver_open()
+    yield
+    cufile.driver_close()
 
 
 @pytest.mark.skipif(not isSupportedFilesystem(), reason="cuFile handle_register requires ext4 or xfs filesystem")
@@ -1891,8 +1882,11 @@ def test_set_parameter_posix_pool_slab_array(slab_sizes, slab_counts, driver_con
     retrieved_counts_addr = ctypes.addressof(retrieved_counts)
 
     # Open cuFile driver AFTER setting parameters
-    with raw_driver():
+    cufile.driver_open()
+    try:
         cufile.get_parameter_posix_pool_slab_array(retrieved_sizes_addr, retrieved_counts_addr, n_slab_sizes)
+    finally:
+        cufile.driver_close()
 
     # Verify they match what we set
     assert list(retrieved_sizes) == slab_sizes


cuda_bindings/tests/test_cufile.py
```diff
@@ -24,46 +24,18 @@
...
-    """Check if running on Windows Subsystem for Linux (WSL)."""
-    return platform.system() == "Linux" and "microsoft" in pathlib.Path("/proc/version").read_text().lower()
-
-
-if platform_is_wsl():
-    pytest.skip("skipping cuFile tests on WSL", allow_module_level=True)
-
-
-from cuda.bindings.cufile import cuFileError
+cufile = pytest.importorskip("cuda.bindings.cufile", reason="skipping tests on Windows")
```
rwgk, line 27: This is a duplicate, see line 18.

I think it's best to not give a `reason`, the default message is fine:

```
SKIPPED [1] tests\test_cufile.py:18: could not import 'cuda.bindings.cufile': No module named 'cuda.bindings.cufile'
```

Note that I had reasons to temporarily disable building the cufile bindings in the CTK 13.1 pre-release testing on platforms other than Windows. In such cases a manually specified reason can be misleading.

cuda_bindings/tests/test_cufile.py
```diff
@@ -108,11 +80,18 @@ def isSupportedFilesystem():
 
 
 # Global skip condition for all tests if cuFile library is not available
-pytestmark = pytest.mark.skipif(not cufileLibraryAvailable(), reason="cuFile library not available on this system")
+pytestmark = [
```
rwgk, line 83: `pytestmark` is not referenced. I removed it locally and `pre-commit` didn't complain:

```diff
-pytestmark = [
+[
```

cuda_bindings/tests/test_cufile.py
```diff
@@ -108,11 +80,18 @@ def isSupportedFilesystem():
...
+    pytest.mark.skipif(not cufileLibraryAvailable(), reason="cuFile library not available on this system"),
+    pytest.mark.skipif(
+        platform.system() == "Linux" and "microsoft" in pathlib.Path("/proc/version").read_text().lower(),
```
rwgk, line 86: For readability, and to make it easier to experiment when requirements change (e.g. new platforms), I'd prefer if we kept the helper functions and call them from here.

cuda_bindings/tests/test_cufile.py
cpcloud: That will break the test suite, because `pytestmark` is a special variable: https://docs.pytest.org/en/stable/example/markers.html#marking-whole-classes-or-modules.

cuda_bindings/tests/test_cufile.py
cpcloud: I find them incredibly noisy. Can't you just add a line to the `pytestmark` list when you're experimenting?

cpcloud: Then you can comment things in and out without having to jump around to function definitions.

cpcloud: That said, I don't really have a major issue with the function version, I just really don't want to do the actual _skipping_ there (the previous pattern).

PR #1316, opened 2025-12-04
[doc-only] Add developer guide for cuda.core

### LINKED ISSUE #1101: Update `cuda.core` developer guide
- Explain the build system change/expectation
- Explain a typical development cycle (pure Python -> Cython -> C++)
- Explain the steps for cythonization

Andy-Jost:
## Summary

This PR adds a developer guide for Python and Cython code in `cuda/core`. The guide establishes conventions and best practices to maintain code quality and consistency across the codebase.

Closes #1101.

## What's Included

The guide covers 10 major areas:

1. **File Structure** - Ordering of elements (SPDX headers, imports, `__all__`, classes, functions)
2. **Package Layout** - Organization of `.pyx`, `.pxd`, and `.py` files
3. **Import Statements** - Five-group organization with alphabetical sorting
4. **Class and Function Definitions** - Ordering within classes (dunder methods, methods, properties)
5. **Naming Conventions** - Project-specific patterns (Cython `c_` prefix, type suffixes)
6. **Type Annotations and Declarations** - Modern PEP 604 union syntax, Cython declarations
7. **Docstrings** - NumPy docstring style with Sphinx cross-references
8. **Errors and Warnings** - Custom CUDA exceptions, error handling patterns
9. **CUDA-Specific Patterns** - GIL management for CUDA driver API calls
10. **Development Lifecycle** - Two-phase development (Python first, then Cython optimization)

## Key Principles

- **PEP 8 and PEP 257 Base**: References these as normative standards, documenting only project-specific extensions
- **Guidelines, Not Rules**: Emphasizes readability and maintainability over rigid enforcement
- **Codebase-Aligned**: Examples reflect actual patterns from the codebase

## File Added

- `cuda_core/docs/source/developer-guide.rst`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,6 +40,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       skip: ${{ steps.get-should-skip.outputs.skip }}
+      doc-only: ${{ steps.get-should-skip.outputs.doc_only }}
     steps:
       - name: Checkout repository
         uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8  # v6.0.1
@@ -50,11 +51,16 @@ jobs:
         run: |
           set -euxo pipefail
           if ${{ startsWith(github.ref_name, 'pull-request/') }}; then
-            skip="$(gh pr view "$(grep -Po '(\d+)$' <<< '${{ github.ref_name }}')" --json title --jq '.title | contains("[no-ci]")')"
+            pr_number="$(grep -Po '(\d+)$' <<< '${{ github.ref_name }}')"
+            pr_title="$(gh pr view "${pr_number}" --json title --jq '.title')"
+            skip="$(echo "${pr_title}" | grep -q '\[no-ci\]' && echo true || echo false)"
+            doc_only="$(echo "${pr_title}" | grep -q '\[doc-only\]' && echo true || echo false)"
           else
             skip=false
+            doc_only=false
           fi
           echo "skip=${skip}" >> "$GITHUB_OUTPUT"
+          echo "doc_only=${doc_only}" >> "$GITHUB_OUTPUT"
 
   # WARNING: make sure all of the build jobs are in sync
   build-linux-64:
@@ -86,7 +92,7 @@ jobs:
         host-platform:
           - linux-aarch64
     name: Build ${{ matrix.host-platform }}, CUDA ${{ needs.ci-vars.outputs.CUDA_BUILD_VER }}
-    if: ${{ github.repository_owner == 'nvidia' && !fromJSON(needs.should-skip.outputs.skip) }}
+    if: ${{ github.repository_owner == 'nvidia' && !fromJSON(needs.should-skip.outputs.skip) && !fromJSON(needs.should-skip.outputs.doc-only) }}
     secrets: inherit
     uses: ./.github/workflows/build-wheel.yml
     with:
@@ -105,7 +111,7 @@ jobs:
         host-platform:
           - win-64
     name: Build ${{ matrix.host-platform }}, CUDA ${{ needs.ci-vars.outputs.CUDA_BUILD_VER }}
-    if: ${{ github.repository_owner == 'nvidia' && !fromJSON(needs.should-skip.outputs.skip) }}
+    if: ${{ github.repository_owner == 'nvidia' && !fromJSON(needs.should-skip.outputs.skip) && !fromJSON(needs.should-skip.outputs.doc-only) }}
     secrets: inherit
     uses: ./.github/workflows/build-wheel.yml
     with:
@@ -121,11 +127,12 @@ jobs:
         host-platform:
           - linux-64
     name: Test ${{ matrix.host-platform }}
-    if: ${{ github.repository_owner == 'nvidia' }}
+    if: ${{ github.repository_owner == 'nvidia' && !fromJSON(needs.should-skip.outputs.doc-only) }}
     permissions:
       contents: read  # This is required for actions/checkout
     needs:
       - ci-vars
+      - should-skip
       - build-linux-64
     secrets: inherit
     uses: ./.github/workflows/test-wheel-linux.yml
@@ -142,6 +149,8 @@ jobs:
         host-platform:
           - linux-aarch64
     name: Test ${{ matrix.host-platform }}
+    # Note: No doc-only check needed here - if build-linux-aarch64 is skipped,
+    # this job is automatically skipped due to the dependency.
     if: ${{ github.repository_owner == 'nvidia' }}
     permissions:
       contents: read  # This is required for actions/checkout
@@ -162,6 +171,8 @@ jobs:
         host-platform:
           - win-64
     name: Test ${{ matrix.host-platform }}
+    # Note: No doc-only check needed here - if build-windows is skipped,
+    # this job is automatically skipped due to the dependency.
     if: ${{ github.repository_owner == 'nvidia' }}
     permissions:
       contents: read  # This is required for actions/checkout
@@ -196,6 +207,7 @@ jobs:
     if: always()
     runs-on: ubuntu-latest
     needs:
+      - should-skip
       - test-linux-64
       - test-linux-aarch64
       - test-windows
@@ -219,11 +231,18 @@ jobs:
           # failing job(s) will timeout causing a cancellation here and the
           # build to succeed which we don't want (originally this was just
           # 'exit 0')
-          if ${{ needs.test-linux-64.result == 'cancelled' ||
-                 needs.test-linux-aarch64.result == 'cancelled' ||
-                 needs.test-windows.result == 'cancelled' ||
-                 needs.doc.result == 'cancelled' }}; then
+          #
+          # Note: When [doc-only] is in PR title, test jobs are intentionally
+          # skipped and should not cause failure.
+          doc_only=${{ needs.should-skip.outputs.doc-only }}
+          if ${{ needs.doc.result == 'cancelled' }}; then
             exit 1
-          else
-            exit 0
           fi
+          if [[ "${doc_only}" != "true" ]]; then
+            if ${{ needs.test-linux-64.result == 'cancelled' ||
+                   needs.test-linux-aarch64.result == 'cancelled' ||
+                   needs.test-windows.result == 'cancelled' }}; then
+              exit 1
+            fi
+          fi
+          exit 0
--- a/cuda_bindings/pixi.lock
+++ b/cuda_bindings/pixi.lock
@@ -26,7 +26,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-cudart-static-13.1.80-hecca717_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-cudart-static_linux-64-13.1.80-h376f20c_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-cudart_linux-64-13.1.80-h376f20c_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-nvrtc-13.1.80-hecca717_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-nvrtc-13.1.115-hecca717_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-nvvm-13.1.80-h69a702a_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-nvvm-dev_linux-64-13.1.80-ha770c72_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-nvvm-impl-13.1.80-h4bc722e_0.conda
@@ -72,7 +72,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libblas-3.11.0-5_h4a7cf45_openblas.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libcap-2.77-h3ff7636_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libcblas-3.11.0-5_h0358290_openblas.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/libcufile-1.16.0.49-hd07211c_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/libcufile-1.16.1.26-hd07211c_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libdeflate-1.25-h17f619e_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libdrm-2.4.125-hb03c661_1.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libegl-1.7.0-ha4b6fd6_2.conda
@@ -220,7 +220,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-aarch64/cuda-cudart-static-13.1.80-h8f3c8d4_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-cudart-static_linux-aarch64-13.1.80-h8f3c8d4_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-cudart_linux-aarch64-13.1.80-h8f3c8d4_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-aarch64/cuda-nvrtc-13.1.80-h8f3c8d4_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-aarch64/cuda-nvrtc-13.1.115-h8f3c8d4_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-aarch64/cuda-nvvm-13.1.80-he9431aa_100.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-nvvm-dev_linux-aarch64-13.1.80-h579c4fd_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-aarch64/cuda-nvvm-impl-13.1.80-h7b14b0b_0.conda
@@ -263,7 +263,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-aarch64/libblas-3.11.0-5_haddc8a3_openblas.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-aarch64/libcap-2.77-h68e9139_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-aarch64/libcblas-3.11.0-5_hd72aa62_openblas.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-aarch64/libcufile-1.16.0.49-hbf501ad_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-aarch64/libcufile-1.16.1.26-hbf501ad_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-aarch64/libdeflate-1.25-h1af38f5_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-aarch64/libdrm-2.4.125-he30d5cf_1.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-aarch64/libegl-1.7.0-hd24410f_2.conda
@@ -401,7 +401,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/win-64/cuda-cudart-static-13.1.80-hac47afa_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-cudart-static_win-64-13.1.80-hac47afa_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-cudart_win-64-13.1.80-hac47afa_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/win-64/cuda-nvrtc-13.1.80-hac47afa_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/win-64/cuda-nvrtc-13.1.115-hac47afa_0.conda
       - conda: https://conda.anaconda.org/conda-forge/win-64/cuda-nvvm-13.1.80-h719f0c7_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-nvvm-dev_win-64-13.1.80-h57928b3_0.conda
       - conda: https://conda.anaconda.org/conda-forge/win-64/cuda-nvvm-impl-13.1.80-h2466b09_0.conda
@@ -542,7 +542,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-cudart-static-13.1.80-hecca717_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-cudart-static_linux-64-13.1.80-h376f20c_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-cudart_linux-64-13.1.80-h376f20c_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-nvrtc-13.1.80-hecca717_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-nvrtc-13.1.115-hecca717_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-nvvm-13.1.80-h69a702a_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-nvvm-dev_linux-64-13.1.80-ha770c72_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-nvvm-impl-13.1.80-h4bc722e_0.conda
@@ -588,7 +588,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libblas-3.11.0-5_h4a7cf45_openblas.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libcap-2.77-h3ff7636_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libcblas-3.11.0-5_h0358290_openblas.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/libcufile-1.16.0.49-hd07211c_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/libcufile-1.16.1.26-hd07211c_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libdeflate-1.25-h17f619e_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libdrm-2.4.125-hb03c661_1.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libegl-1.7.0-ha4b6fd6_2.conda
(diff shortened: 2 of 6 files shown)


Andy-Jost:
Link to rendered document: [style-guide.md](https://github.com/Andy-Jost/cuda-python/blob/style-guide/cuda_core/cuda/core/style-guide.md)

Andy-Jost:
I'm interested in feedback on this idea. With a comprehensive style guide, aside from the other benefits listed in the description, we can use LLMs to help refactor, clean, and document the code. If the team is comfortable with this approach, I can begin to apply rules conservatively to the code base.

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1818 @@
...
+17. [CUDA-Specific Patterns](#cuda-specific-patterns)
+18. [Copyright and Licensing](#copyright-and-licensing)
+
+---
+
+## File Structure
+
+Files in `cuda/core/experimental` must follow a consistent structure. The ordering of elements within a file is as follows:
+
+### 1. SPDX Copyright Header
```
cpcloud: I'm -1 on including things in this guide that can be automated, and I realize that results in a lack of completeness.

In fact, I'd rather our lint rule for this particular thing just add it to the top of the file when it fails.

I think this can be done for all file types, including pyx files.

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1818 @@
...
+
+The file must begin with the SPDX copyright header as specified in [Copyright and Licensing](#copyright-and-licensing).
+
+### 2. Import Statements
+
+Import statements come immediately after the copyright header. Follow the import ordering conventions specified in [Import Statements](#import-statements).
+
+### 3. `__all__` Declaration
+
+If the module exports public API elements, include an `__all__` list after the imports and before any other code. This explicitly defines the public API of the module.
```
cpcloud: This is slightly misleading, since `__all__` is about [what happens when you star-import a module](https://docs.python.org/3/tutorial/modules.html#importing-from-a-package).

+1 on defining this, but stating it defines the public API I would remove from the description.

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1818 @@
...
+Type aliases and module-level constants may immediately follow `__all__` (if present) or come after imports:
+
+```python
+DevicePointerT = driver.CUdeviceptr | int | None
+"""Type union for device pointer representations."""
+
+LEGACY_DEFAULT_STREAM = C_LEGACY_DEFAULT_STREAM
+```
+
+### 5. Principal Class or Function
```
cpcloud: I'm a bit mixed on enforcing ordering of elements here. I do think it can help cut review time, but in practice I find that this kind of strict ordering only starts mattering when files get huge, to which I would probably say "Split the file up" 😆

cuda_core/style-guide.md
```diff
@@ -0,0 +1,1818 @@
...
+    """Private implementation helper."""
+    # ...
+```
+
+### Notes
+
+- Not every file will have all sections. For example, a utility module may not have a principal class.
+- The distinction between "principal" and "other" classes is based on the file's primary purpose. If a file exists primarily to define one class, that class is the principal class.
+- Private implementation functions should be placed at the end of the file to keep the public API visible at the top.
+- **Within each section**, classes and functions should be sorted alphabetically by name. The principal class or function is an exception to this rule, as it appears first in its respective section.
```
cpcloud: I'm not sure this should be enforced without a linter. It feels a bit draconian to me to require someone to reorder their methods.

For example, I like to order methods by "relevance radius", which is a squishy concept that is something like "methods that are related to each other are close to each other". I find this makes debugging way easier, if only because I can see more related code in a single vim window.

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1818 @@
...
+
+**Migration guidance**: Simple top-level modules can be migrated to this subpackage structure when they become sufficiently complex (e.g., when a module grows to multiple related classes or when logical grouping would improve maintainability).
+
+### Guidelines
+
+1. **Always create `.pxd` files for shared Cython types**: If a class or function is `cimport`ed by other modules, provide a `.pxd` declaration file.
+
+2. **Keep `.pxd` files minimal**: Only include declarations needed for Cython compilation. Omit implementation details, docstrings, and Python-only code.
+
+3. **Use `__all__` in submodules**: Each submodule should define `__all__` to explicitly declare its public API.
```
cpcloud: The public API bit seems to be repeated in a bunch of places, and it's misleading. That said, I'm unsure of a better way to phrase it, but perhaps there's a slightly less "sure" way of stating it?

How about just "Each submodule should define `__all__`"?

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1818 @@
...
+
+4. **Use `from ._module import *` in subpackage `__init__.py`**: This pattern assembles the subpackage API from its submodules. Use `# noqa: F403` to suppress linting warnings about wildcard imports.
+
+5. **Migrate to subpackage structure when complex**: When a top-level module becomes complex (multiple related classes, logical grouping needed), consider refactoring to the subpackage pattern.
+
+6. **Separate concerns**: Use `.py` files for pure Python utilities, `.pyx` files for Cython implementations that need C-level performance.
+
+## Import Statements
```
cpcloud: How do we avoid making this conflict with formatting tools like `ruff`?

I would hate for someone to follow the guide and then get a lint error on commit. I also would really really like to avoid thinking about this too hard (which I guess the LLM helps with here)

I feel like you can tell an LLM to look at the ruff config, and give it the ruff rule documentation, and tell it to format cython imports as if ruff could format them.

rwgk:
Will look more closely asap.

Drive-by comment: I was surprised to see the new file under `cuda_core/cuda/core/`. I'd expect it to be under `cuda_core/`, like other meta-kind-of files, e.g. `LICENSE`, `README.md`, etc.

Andy-Jost:
> Will look more closely asap.
> 
> Drive-by comment: I was surprised to see the new file under `cuda_core/cuda/core/`. I'd expect it to be under `cuda_core/`, like other meta-kind-of files, e.g. `LICENSE`, `README.md`, etc.

I thought about that, too. I agree it belongs at the root.

rparolin:
> > Will look more closely asap.
> > Drive-by comment: I was surprised to see the new file under `cuda_core/cuda/core/`. I'd expect it to be under `cuda_core/`, like other meta-kind-of files, e.g. `LICENSE`, `README.md`, etc.
> 
> I thought about that, too. I agree it belongs at the root.

Do we want this in the root? I'd expect it within a docs folder. That shouldn't impact an LLMs ability to discover and utilize it.

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+
+# 5. cuda-core imports
+from cuda.core.experimental._context import Context
+from cuda.core.experimental._device import Device
+from cuda.core.experimental._utils.cuda_utils import driver
+```
+
+## Class and Function Definitions
+
+### Class Definition Order
```
cpcloud: I don't know about this. I'm not sure there's a lot of value in enforcing an arbitrary order. Sometimes it makes sense to order things in a different way than alphabetically, for example.

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+
+    @property
+    def size(self) -> int:
+        """Size property."""
+        # ...
+```
+
+### Helper Functions
+
+Sometimes, implementation details are moved outside of the class definition to improve readability. Helper functions should be placed at the end of the file (in the private/implementation section) when:
```
cpcloud: Why wouldn't we want to place helper functions as close to their call sites as possible (assuming that they're private to a module)?

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+```
+
+The `stacklevel` value depends on the call depth. Use `stacklevel=2` for direct function calls, `stacklevel=3` for calls through helper functions.
+
+#### One-Time Warnings
+
+For warnings that should only be emitted once per process, use a module-level flag:
+
+```python
+_fork_warning_checked = False
```
cpcloud: Why can't we use `simplefilter` for this scenario as well?

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+        return
+    _fork_warning_checked = True
+
+    # ... check condition and emit warning ...
+    warnings.warn(message, UserWarning, stacklevel=3)
+```
+
+#### Deprecation Warnings
+
+For deprecation warnings, use `warnings.simplefilter("once", DeprecationWarning)` to ensure each deprecation message is shown only once:
```
cpcloud: This doesn't give any guidance on scope, which matters a lot here.

For example:

```python
import warnings


warnings.simplefilter("once", UserWarning)
def foo():
    warnings.warn("tsk tsk", UserWarning)
    warnings.warn("tsk tsk", UserWarning)

foo()
foo()
```

behaves differently than

```python
import warnings


def foo():
    warnings.simplefilter("once", UserWarning)
    warnings.warn("tsk tsk", UserWarning)
    warnings.warn("tsk tsk", UserWarning)

foo()
foo()
```

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+
+def some_function():
+    if not hasattr(_tls, 'devices'):
+        _tls.devices = []
+    return _tls.devices
+```
+
+### Locks for Shared State
+
+Use `threading.Lock()` to protect shared mutable state:
```
cpcloud: I'm not sure this actually adds a ton of value without the details that are available in the documentation, which we would just be repeating here if we included them.

We don't really want LLMs blindly using locks where the only criteria is "is it shared mutable state"? Shared in what sense? What about sharing across processes? I'm not sure that even to a first approximation this is useful.

What about recursive locks? How does GC behavior play in here? This goes way beyond style IMO and the scenario for dealing with shared mutable state is too complex to include such a simplistic rule.

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+
+```python
+def thread_safe_cuda_operation():
+    with _lock, nogil:
+        HANDLE_RETURN(cydriver.cuSomeOperation())
+```
+
+### One-Time Initialization
+
+For one-time initialization that must be thread-safe, use a lock with a flag:
```
cpcloud: I dunno. I could easily see atomics being used for this case. -1 on all these specific suggestions. I would probably just remove the whole thread safety and concurrencyd section (since this is all LLM-generated there's not much to lose by underspecifying and solving the threading-and-concurrency-guideline-problem later).

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+```python
+cdef:
+    int device_id
+    size_t buffer_size
+    cydriver.CUdeviceptr ptr
+```
+
+### Inline Functions
+
+Use `inline` for small, frequently-called functions:
```
cpcloud: Does this still matter in 2025? I am pretty dubious (but of course open to being wrong) that a modern C compiler is going to produce meaningfully faster code versus leaving it to choose whether to inline.

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+
+2. **Use UPPER_SNAKE_CASE**: Follow Python convention for module-level constants.
+
+3. **Group related constants**: Define related constants together, optionally in a dictionary or class.
+
+4. **Document non-obvious constants**: If a constant's purpose isn't immediately clear, add a comment explaining it.
+
+5. **Prefer CUDA bindings**: Use constants from `cuda.bindings` directly when possible rather than redefining them.
+
+## Comments and Inline Documentation
```
cpcloud: This feels overly prescriptive to me.

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+# NOTE: match this behavior to DeviceMemoryResource.allocate()
+stream = default_stream()
+
+# NOTE: this is referenced in instructions to debug nvbug 5698116
+cpdef DMR_mempool_get_access(DeviceMemoryResource dmr, int device_id):
+```
+
+### Implementation Comments
+
+Add comments to explain complex logic or non-obvious behavior:
```
cpcloud: I really question whether this is useful for humans. I suppose it may be useful for an LLM 🤷🏻

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+```python
+cdef:
+    int device_id
+    size_t buffer_size
+    cydriver.CUdeviceptr ptr
+```
+
+### Prefer `cdef` for Internal Functions
+
+Use `cdef` functions for internal operations that don't need to be callable from Python:
```
cpcloud: Isn't this already specified above?

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+Avoid creating Python objects in hot paths:
+
+```python
+# Avoid: Creates Python list
+result = []
+for i in range(n):
+    result.append(i)
+
+# Preferred: Use C array or pre-allocate
+cdef int* c_result = <int*>malloc(n * sizeof(int))
```
cpcloud: Seeing as how these are not equivalent, it seems odd to say "prefer X over Y" when X and Y are very different. Sometimes you want a list, sometimes you don't need it and can use a C array.

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+6. **Cache expensive lookups**: Cache results of expensive operations (e.g., dictionary lookups, attribute access) when used repeatedly.
+
+## API Design Principles
+
+### Public vs Private API
+
+Use naming conventions to distinguish public and private APIs:
+
+- **Public API**: No leading underscore, documented in docstrings, included in `__all__`
+- **Private API**: Leading underscore (`_`), may have minimal documentation, not in `__all__`
```
cpcloud: Since `__all__` **only** affects star import behavior, this is redundant, since leading underscore names are not included in star imports.

cuda_core/style-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+
+```python
+# SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+```
+
+### Guidelines
+
+1. **Placement**: The copyright header must be the first lines of the file, before any imports or other code.
```
cpcloud: Isn't this covered by the "all files must include a copyright header"?

cuda_core/style-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+
+### 6. Other Public Classes and Functions
+
+Following the principal class or function, define other public classes and functions. These include:
+
+- **Auxiliary classes**: Supporting classes that are part of the public API (e.g., `DeviceMemoryResourceOptions` is an auxiliary class used by `DeviceMemoryResource`)
+- **Abstract base classes**: ABCs that define interfaces (e.g., `MemoryResource` in `_buffer.pyx`)
+- **Other public classes**: Additional classes exported by the module
+
+**All classes and functions in this section should be sorted alphabetically by name**, regardless of their relationship to the principal class. The principal class appears first as an exception to this rule.
```
mdboom: -1 on alphabetical ordering.  I think we should reflect the logical relationships between classes instead (which I understand is a bit mushy as a concept).  It's easy enough in an IDE to jump to the one you are looking for using search.

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+
+After all classes, define public module-level functions that are part of the API.
+
+### 8. Private or Implementation Functions
+
+Finally, define private functions and implementation details. These include:
+
+- Functions with names starting with `_` (private)
+- `cdef inline` functions used for internal implementation
+- Helper functions not part of the public API
```
mdboom: I suppose this is fine, but seems backward to me, given that languages like C have always required forward declarations to do things like this, code tends to be ordered from low-level to high-level, with the "principal public thing" at the bottom, not the top.  I realize Python has never had this problem, but it does seems reversed from almost anything else.

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+- **Within each section**, classes and functions should be sorted alphabetically by name. The principal class or function is an exception to this rule, as it appears first in its respective section.
+
+## Package Layout
+
+### File Types
+
+The `cuda/core/experimental` package uses three types of files:
+
+1. **`.pyx` files**: Cython implementation files containing the actual code
+2. **`.pxd` files**: Cython declaration files containing type definitions and function signatures for C-level access
```
mdboom: (For the future): We really should be requiring that all `.pyx` files have a corresponding `.pyi` file.  Without it, modern IDEs (which don't actually import modules) can not perform auto completion and cross referencing.  But we can add that later as an effort to improve the IDE experience.

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+- **Declaration files**: Use `.pxd` for Cython type declarations
+- **Private modules**: Prefix with underscore (e.g., `_buffer.pyx`, `_device.pyx`)
+- **Public modules**: No underscore prefix (e.g., `utils.py`)
+
+### Relationship Between `.pxd` and `.pyx` Files
+
+For each `.pyx` file that defines classes or functions used by other Cython modules, create a corresponding `.pxd` file:
+
+- **`.pxd` file**: Contains `cdef` class declarations, `cdef`/`cpdef` function signatures, and `cdef` attribute declarations
+- **`.pyx` file**: Contains the full implementation including Python methods, docstrings, and implementation details
```
mdboom: (As above): The better way is to include docstrings and type annotations in a `.pyi` file and just implementation in the `.pyx` file.

cuda_core/docs/developer-guide.md
mdboom: If I were writing this for a human, I would take this section (and some other sections below that are basically "PEP8") and just link to them so we aren't maintaining duplication of community standards.  If the LLM can handle something like that, I think we should do that here.  Given "agents", shouldn't they just be able to run `ruff`?

cuda_core/style-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+
+Within a class definition, elements must be organized in the following order:
+
+1. **Special (dunder) methods**: Methods with names starting and ending with double underscores (e.g., `__init__`, `__cinit__`, `__dealloc__`, `__reduce__`, `__dlpack__`)
+
+2. **Methods**: Regular instance methods, class methods (`@classmethod`), and static methods (`@staticmethod`)
+
+3. **Properties**: Properties defined with `@property` decorator
+
+**Note**: Within each section (dunder methods, methods, properties), elements should be sorted alphabetically by name.
```
mdboom: -1 again to anything alphabetical.  Search is easy in most editors/IDEs, but the goal of ordering in the file should be for reading for understanding.

cuda_core/style-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+
+    def __dealloc__(self):
+        """Cleanup."""
+        # ...
+
+    def __dlpack__(self, *, stream=None):
+        """DLPack protocol support."""
+        # ...
+
+    def __init__(self, *args, **kwargs):
```
mdboom: Additionally on the alphabetical thing, `__init__` is conventionally /always/ the first method (though I suppose with Cython `__cinit__` muddies that a bit).  But I definitely wouldn't want to see `__buffer__` first.

cuda_core/style-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+
+### Function Definitions
+
+For module-level functions (outside of classes), follow the ordering specified in [File Structure](#file-structure): principal functions first (if applicable), then other public functions, then private functions. Within each group, sort alphabetically.
+
+## Naming Conventions
+
+### Class Names
+
+Use **PascalCase** (also known as CapWords) for class names.
```
mdboom: I most often see this called `CamelCase`.

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+    cdef:
+        uintptr_t _ptr
+        size_t _size
+        MemoryResource _memory_resource
+        object _ipc_data
+```
+
+#### Cython `cdef` Variables
+
+Consider prefixing `cdef` variables with `c_` to distinguish them from Python variables. This improves code readability by making it clear which variables are C-level types.
```
mdboom: 👍

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+
+The `c_` prefix is particularly helpful when mixing Python and Cython variables in the same scope, or when the variable name would otherwise be ambiguous.
+
+### Constants
+
+Use **UPPER_SNAKE_CASE** for module-level constants.
+
+```python
+LEGACY_DEFAULT_STREAM = C_LEGACY_DEFAULT_STREAM
+PER_THREAD_DEFAULT_STREAM = C_PER_THREAD_DEFAULT_STREAM
```
mdboom: Additionally, related constants should be grouped into enums.

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+```python
+cdef class DeviceMemoryResource:
+    def close(self):
+        """Close the memory resource and release CUDA handles."""
+        DMR_close(self)
+
+    def __dealloc__(self):
+        """Automatic cleanup when object is garbage collected."""
+        DMR_close(self)
+```
```
mdboom: Should we recommend that all such objects that require cleanup define a context manager, i.e. `__enter__` and `__exit__` methods?

cuda_core/docs/developer-guide.md
```diff
@@ -0,0 +1,1980 @@
...
+
+2. **Implement `__dealloc__` as safety net**: Use `__dealloc__` to ensure cleanup happens even if users forget to call `close()`, but don't rely on it for normal operation.
+
+3. **Document cleanup behavior**: Clearly document when cleanup happens automatically versus when it must be called explicitly.
+
+4. **Handle cleanup errors gracefully**: Cleanup methods should be idempotent (safe to call multiple times) and handle errors without raising exceptions when possible.
+
+5. **Use stream-ordered deallocation**: When deallocating buffers, use the appropriate stream for asynchronous cleanup to avoid blocking operations.
+
+6. **Track resource ownership**: Clearly document which objects own CUDA handles and are responsible for cleanup.
```
mdboom: Can't we do better than just documenting, by specifying all objects contain a strong reference to their owner to prevent use-after-free errors etc.?

EDIT: Though maybe for a style guide this is too complex of an issue.  For style, documenting that such a situation exists is the first requirement for actually patching these sort of holes.

cuda_core/docs/developer-guide.md
mdboom: This is for Cython, which is not a modern C compiler ;)  IME, sometimes the C compiler can see through it and make the right choice and other times not.  The salient point for me is that we would ideally backup the choice of inline with a benchmark.

cuda_core/docs/developer-guide.md
leofang: In 2025, is there a way we can auto-generate `.pyi` from `.py`/`.pxd`/`.pyx`? If not, let's not make it a must.

One thing I hate the most is Python typing. Does not make any sense to me to maintain 2-3 files.

cuda_core/docs/developer-guide.md
leofang: I don't think Cython would honor `.pyi` files? To the very least, I don't think it is friendly to write `.pyi` files when the types are needed to complete a `cdef` function?

cuda_core/docs/developer-guide.md
leofang: Not for now. I can write a section explaining the current rationale in a separate PR if it is OK.

leofang:
Merging my offline comment to here. My 2c:
- If possible I’d like to make part (or all) of this PR rendered as Sphinx docs, for both internal and external contributors to use as developer guide
- I’d also like all LLM to be able to read it when working on the codebase, including but not limited to: Cursor, CodeRabbit, Copilot, …

cuda_core/docs/developer-guide.md
Andy-Jost: FWIW, some of the things I'd like to document would more properly belong in a "Developer's Guide." I am therefore considering whether to expand the scope of this document. I'd restructure it into sections where "Style" is one section.

cuda_core/style-guide.md
Andy-Jost: I have the opposite sentiment, tending to prefer a concrete rule over an imprecise one. The idea of grouping by functionality opens the door to endless fiddling, as functional groups can involve overlapping subsets.

As an example, I like that the CUDA Driver documentation alphabetizes, as seen [here](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MALLOC__ASYNC.html#group__CUDA__MALLOC__ASYNC).

Sure, the alphabetical rule puts, say, `cuMemPoolGetAttribute` far away from `cuMemPoolSetAttribute`, but that's not a problem because the document sections limit scope (there are only 17 items in this section). Since classes and modules also limit scope in a similar way, I prefer to alphabetize locally within them.

That's my 2 cents, anyway. Definitely not a hill I'd die on.

Andy-Jost:
Moved to cuda_core/docs

cuda_core/style-guide.md
Andy-Jost: This is a great point. My reason for spelling out the order this way is not to create unnecessary rigidity for human developers, but to make a clear instruction for LLMs to follow.

I see it this way:
 - This rule is made to be broken. It should be disregarded anytime something else is better in a specific case.
 - When there is no particular reason to care, spelling out a default order for the LLMs to follow allows us to automatically clean up.

The intent is to *reduce* the load on human developers. My hope is that we can worry less about ordering things manually and rely more on LLMs to clean up after the fact. The language here probably could be relaxed.

cuda_core/docs/developer-guide.md
Andy-Jost: I removed the actual contents and made a note that pre-commit adds this where needed.

cuda_core/docs/developer-guide.md
Andy-Jost: I like the idea of using `.pyi` files. I can look into auto-generating those. After we have an established practice, we can update this document.

Andy-Jost: https://github.com/NVIDIA/cuda-python/issues/1442

cuda_core/docs/developer-guide.md
Andy-Jost: I removed several sections below to eliminate generic content. The new version should be much tighter. I also removed certain things that might conflict with or become out of date with the linters (e.g., the verbatim contents of SPDX headers).

As for the import statements, there is partial overlap with ruff: it enforces alphabetization but not the "5 sections" format we follow. I think it makes sense to describe those here.

cuda_core/docs/developer-guide.md
Andy-Jost: I softened the language here quite a bit. There is a suggested order, but the guide makes it clear that developers are free to deviate according to their best judgement. The goal is not to constrain humans but to give guidance to machines.

cuda_core/docs/developer-guide.md
Andy-Jost: I've been finding that some of our larger class defs (>400 LOC) are difficult to navigate, particularly when methods are very long, properties are mixed in between methods, and nothing is alphabetized. It's not a huge deal, but I'd love to see a move towards a bit more structure.

cuda_core/docs/developer-guide.md
Andy-Jost: I removed generic things like this. There was a fair bit of AI "slop" in the previous version. This one should be much more concise and focused.

cuda_core/docs/developer-guide.md
Andy-Jost: removed

cuda_core/style-guide.md
Andy-Jost: I rephrased throughout to make it clear that the suggested order is not a requirement, that logical ordering specifically is acceptable, and that alphabetical is a possible fallback if nothing better applies.

cuda_core/docs/developer-guide.md
Andy-Jost: I added a general statement about this in the "File Structure" section. Many of us have grown accustomed to C/C++-style declare-before-use rules, but I think putting the most important thing first is a better rule when we can follow it because it is better for readability.

Andy-Jost:
The doc is now included with the Sphinx docs. Link above but also [here ](https://nvidia.github.io/cuda-python/pr-preview/pr-1316/cuda-core/latest/). Click through CUDA Core Developer Guide.

cuda_core/docs/developer-guide.md
Andy-Jost: I dug into this a little more. It appears that docstrings and type annotations need to be *repeated* in `.pyi`; we cannot move content there from `.pyx` files. More info in #1442. Maintaining these by hand would be a headache. If we want them, we should look for a way to generate `.pyi` files.

Andy-Jost:
I've addressed all the feedback from the first round. Please take another look so we can get this one out the door. @leofang @mdboom @cpcloud

Andy-Jost:
Closing after discussion; will not publish.

Issue #1314, opened 2025-12-04
CUDA Graph phase 3 - memcpy nodes


rparolin:
Cover cuMemcpyAsync and cover cuMemsetAsync.

https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MEM.html#group__CUDA__MEM_1g5f26aaf5582ade791e5688727a178d78

cuMemsetAsync has multiple bit-width APIs. We can wrap them nicely from the python side so the user doesn't need to worry about what bit-width variant they need to call.

https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MEM.html#group__CUDA__MEM_1g7d805e610054392a4d11e8a8bf5eb35c

Andy-Jost:
Is `cuMempyAsync` covered by `Buffer.copy_to` and `Buffer.copy_from` or is something special needed for graphs?

Could we similarly cover `cuMemsetAsync` by adding a method to `Buffer`?

rparolin:
> Is `cuMempyAsync` covered by `Buffer.copy_to` and `Buffer.copy_from` or is something special needed for graphs?
@leofang  please advise.
 
> Could we similarly cover `cuMemsetAsync` by adding a method to `Buffer`?
Yes, that's my understanding. We could have a call `fill` or `clear` ?

Andy-Jost:
Ok, follow-up question:

> We can wrap them nicely from the python side so the user doesn't need to worry about what bit-width variant they need to call

How do we envision this working? If someone says `buf.fill(1)` we still don't have information about the intended bit-width. Are we thinking something like `buf.fill(value=1, width=4)` ?

If the type is a specific NumPy dtype we can do it automatically. Is there a precedent for anything like that, or do we have a list of frameworks/dtypes that we support?

Andy-Jost:
Please see the attached draft PR. Is that what we have in mind for `Buffer.fill`?

leofang:
Sorry for late reply. Continuing the discussion in your PR 🙂 https://github.com/NVIDIA/cuda-python/pull/1318#issuecomment-3629515796

leofang:
@Andy-Jost can correct me if I am wrong, but I think this is done.

PR #1318, opened 2025-12-04
Add Buffer.fill() method for cuMemsetAsync support (#1314)

Andy-Jost:
## Summary

This PR implements `Buffer.fill(value, width, *, stream)` method that wraps CUDA's `cuMemsetAsync` functions, supporting `cuMemsetD8Async`, `cuMemsetD16Async`, and `cuMemsetD32Async` based on the width parameter.

Part of issue #1314: CUDA Graph phase 3 - memcpy nodes

## Changes

- **Added `Buffer.fill()` method** in `cuda/core/experimental/_memory/_buffer.pyx`:
  - Supports width=1 (byte fill via `cuMemsetD8Async`)
  - Supports width=2 (16-bit fill via `cuMemsetD16Async`)
  - Supports width=4 (32-bit fill via `cuMemsetD32Async`)
  - Validates width (must be 1, 2, or 4)
  - Validates value range for each width
  - Validates buffer size is divisible by width
  - Follows same pattern as `copy_to`/`copy_from` methods

- **Added comprehensive tests** in `tests/test_memory.py`:
  - Tests all three widths (1, 2, 4 bytes)
  - Tests error cases (invalid width, value out of range, size not divisible)
  - Tests with different memory resource types
  - Includes verification for host-accessible memory

## Implementation Details

The method automatically selects the appropriate CUDA driver API function based on the `width` parameter:
- `width=1`: Uses `cuMemsetD8Async` with N = buffer_size (bytes)
- `width=2`: Uses `cuMemsetD16Async` with N = buffer_size // 2 (16-bit elements)
- `width=4`: Uses `cuMemsetD32Async` with N = buffer_size // 4 (32-bit elements)

## Example Usage

```python
buffer = mr.allocate(1024, stream=stream)
buffer.fill(0x42, width=1, stream=stream)  # Fill with byte value
buffer.fill(0x1234, width=2, stream=stream)  # Fill with 16-bit value
buffer.fill(0xDEADBEEF, width=4, stream=stream)  # Fill with 32-bit value
```

## Testing

All tests pass with comprehensive coverage of:
- Success cases for all widths
- Error validation (width, value range, size divisibility)
- Multiple memory resource types (device, unified, pinned)

### DIFF
--- a/cuda_core/tests/test_memory.py
+++ b/cuda_core/tests/test_memory.py
@@ -778,7 +778,7 @@ def test_strided_memory_view_leak():
     arr = np.zeros(1048576, dtype=np.uint8)
     before = sys.getrefcount(arr)
     for idx in range(10):
-        StridedMemoryView(arr, stream_ptr=-1)
+        StridedMemoryView.from_any_interface(arr, stream_ptr=-1)
     after = sys.getrefcount(arr)
     assert before == after
 
@@ -786,7 +786,7 @@ def test_strided_memory_view_leak():
 def test_strided_memory_view_refcnt():
     # Use Fortran ordering so strides is used
     a = np.zeros((64, 4), dtype=np.uint8, order="F")
-    av = StridedMemoryView(a, stream_ptr=-1)
+    av = StridedMemoryView.from_any_interface(a, stream_ptr=-1)
     # segfaults if refcnt is wrong
     assert av.shape[0] == 64
     assert sys.getrefcount(av.shape) >= 2
--- a/cuda_bindings/cuda/bindings/_version.py
+++ b/cuda_bindings/cuda/bindings/_version.py
@@ -1,4 +1,4 @@
 # SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
 
-__version__ = "13.1.0"
+__version__ = "13.1.1a0"
--- a/cuda_bindings/pixi.lock
+++ b/cuda_bindings/pixi.lock
@@ -12,32 +12,32 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/alsa-lib-1.2.14-hb9d3cd8_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/aom-3.9.1-hac33072_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/attr-2.5.2-h39aace5_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/binutils_impl_linux-64-2.45-h9d8b0ac_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/binutils_impl_linux-64-2.45-default_hfdba357_104.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/bzip2-1.0.8-hda65f42_8.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/ca-certificates-2025.11.12-hbd8a1cb_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/cairo-1.18.4-h3394656_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/colorama-0.4.6-pyhd8ed1ab_1.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/conda-gcc-specs-15.2.0-h56430cd_7.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-cccl_linux-64-13.0.85-ha770c72_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-crt-dev_linux-64-13.0.88-ha770c72_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-cudart-13.0.96-hecca717_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-cudart-dev-13.0.96-hecca717_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-cudart-dev_linux-64-13.0.96-h376f20c_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-cudart-static-13.0.96-hecca717_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-cudart-static_linux-64-13.0.96-h376f20c_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-cudart_linux-64-13.0.96-h376f20c_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-nvrtc-13.0.88-hecca717_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-nvvm-13.0.88-h69a702a_4.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-nvvm-dev_linux-64-13.0.88-ha770c72_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-nvvm-impl-13.0.88-h4bc722e_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-nvvm-tools-13.0.88-h4bc722e_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-pathfinder-1.3.2-pyhcf101f3_0.conda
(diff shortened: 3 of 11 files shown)


cuda_core/cuda/core/experimental/_memory/_buffer.pyx
```diff
@@ -167,15 +173,83 @@ cdef class Buffer:
...
+        cdef Stream s_stream = <Stream>stream
+        cdef size_t buffer_size = self._size
+        cdef unsigned char c_value8
+        cdef unsigned short c_value16
+        cdef unsigned int c_value32
+        cdef size_t N
+        cdef cydriver.CUstream s = s_stream._handle
+
+        # Validate width
+        if width not in (1, 2, 4):
```
rparolin, line 219: I'd put the validation code closer to the top of the function so we avoid any setup work in the error case where the user passes an unsupported size to the function.

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
```diff
@@ -167,15 +173,83 @@ cdef class Buffer:
...
+            raise ValueError(f"width must be 1, 2, or 4, got {width}")
+
+        # Validate value fits in width
+        if width == 1:
```
rparolin, line 233: You could hoist this logic into a validation function to remove the magic numbers in the code.

```
def _validate_value_against_bitwidth(bitwidth, value, signed=False):
    max_bits = bitwidth
    if signed:
        min_value = -(1 << (max_bits - 1))
        max_value = (1 << (max_bits - 1)) - 1
    else:
        min_value = 0
        max_value = (1 << max_bits) - 1

    if not min_value <= value <= max_value:
        raise ValueError(
            f"value ({value}) is outside the representable range for {bitwidth}-bit integers "
            f"[{min_value}, {max_value}]"
        )
```

cuda_core/tests/test_memory.py
```diff
@@ -219,6 +219,88 @@ def test_buffer_copy_from():
...
 
+def buffer_fill(dummy_mr: MemoryResource, device: Device, check=False):
+    stream = device.create_stream()
+
+    # Test width=1 (byte fill)
+    buffer1 = dummy_mr.allocate(size=1024)
+    buffer1.fill(0x42, width=1, stream=stream)
+    device.sync()
+
+    if check:
```
rparolin, line 230: Why are parametrizing these value checks in a test suite?  The memory sizes don't strike me as so large that these operations would be slow.

Andy-Jost: The values are only checked when the memory allocation is pinned. This follows the existing pattern.

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
Andy-Jost: Ok. I converted it to Cython, too.

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
Andy-Jost: OK. I fiddled with it to simplify the logic. To be honest, I don't see a big improvement, here, since most of the preceding statements just declare stack variables.

kkraus14:
cc @pciolkosz who implemented CCCL's `cuda::fill_bytes`

I think something what would make more sense would be for `Buffer.fill()` to only work for 1 byte width and then have a `StridedMemoryView.fill()` that uses the dtype of the `StridedMemoryView` to control the byte width used for filling as well as interpreting the signed-ness of the input. I think this has a couple of benefits:
* No need for user to pass a width explicitly and inherently no need to ensure that the Buffer is aligned with a passed width
* Makes it clear when a Python integer is desired to be treated as an unsigned vs signed integer

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
```diff
@@ -167,15 +173,78 @@ cdef class Buffer:
...
+            raise ValueError(f"width must be 1, 2, or 4, got {width}")
+
+        # Validate buffer size modulus.
+        cdef size_t buffer_size = self._size
+        if buffer_size % width != 0:
+            raise ValueError(f"buffer size ({buffer_size}) must be divisible by width ({width})")
+
+        # Map width (bytes) to bitwidth and validate value
+        cdef int bitwidth = width * 8
+        _validate_value_against_bitwidth(bitwidth, value, is_signed=False)
```
kkraus14, line 229: I think this is a bit of unexpected behavior that this will throw if someone passes a negative integer

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
```diff
@@ -340,3 +409,43 @@ cdef class MemoryResource:
             and document the behavior.
         """
         ...
+
+
+# Helper Functions
+# ----------------
+cdef void _validate_value_against_bitwidth(int bitwidth, int64_t value, bint is_signed=False) except *:
```
kkraus14, line 416: nitpick: we should probably make this an inline function for performance reasons

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
```diff
@@ -340,3 +409,43 @@ cdef class MemoryResource:
...
+    cdef uint64_t max_value_unsigned
+    cdef int64_t max_value
+
+    if is_signed:
+        min_value = -(<int64_t>1 << (max_bits - 1))
+        max_value = (<int64_t>1 << (max_bits - 1)) - 1
+    else:
+        min_value = 0
+        max_value_unsigned = (<uint64_t>1 << max_bits) - 1
+        max_value = <int64_t>max_value_unsigned
```
kkraus14, line 446: nitpick: getting the min and max values here can be cimported from the built in libc module: https://github.com/cython/cython/blob/master/Cython/Includes/libc/limits.pxd

i.e. `from libc.limits cimport INT_MAX`

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
```diff
@@ -167,15 +173,78 @@ cdef class Buffer:
...
+                s
+            ))
+
+    def fill(self, value: int, width: int, *, stream: Stream | GraphBuilder):
+        """Fill this buffer with a value pattern asynchronously on the given stream.
+
+        Parameters
+        ----------
+        value : int
+            Integer value to fill the buffer with
```
kkraus14, line 199: I know the underlying driver APIs expect an integer, but I think for `Buffer.fill()` it would be good to support a byte-like input as well, maybe via buffer protocol?

Andy-Jost: To confirm my understanding, I believe the suggestion here is to accept arguments of type `bytes` or that provide bytes via the buffer protocol (where the number of bytes equals the buffer size) and fill our `Buffer` from that. Would that be better to implement as `Buffer.copy_from` rather than `Buffer.fill`? Or possibly a new function `Buffer.copy_from_bytes` or `Buffer.copy_from_host`.

Andy-Jost: The internal logic will be different from `Buffer.fill`, since it requires copying the bytes to a staging area, unlike `cuMemset*`.

leofang:
From https://github.com/NVIDIA/cuda-python/issues/1314#issuecomment-3614720861:
>> How do we envision this working? If someone says `buf.fill(1)` we still don't have information about the intended bit-width. Are we thinking something like `buf.fill(value=1, width=4)` ?
>> 
>> If the type is a specific NumPy dtype we can do it automatically. Is there a precedent for anything like that, or do we have a list of frameworks/dtypes that we support?

and https://github.com/NVIDIA/cuda-python/pull/1318#issuecomment-3627494735:

> and then have a `StridedMemoryView.fill()` that uses the dtype of the `StridedMemoryView` to control the byte width used for filling as well as interpreting the signed-ness of the input. I think this has a couple of benefits:

My thinking was: We follow what we did for `cuda.core.launch()` and use a dispatcher to infer the bit width from the scalar type. So we'll do 
```python
def fill(self, value, stream: Stream | GraphBuilder):
    ...
```
and the user code can be:
```python
buf.fill(np.int16(8), s)
```
This would provide a uniform UX across cuda.core APIs that accept Python/NumPy scalars. @kkraus14 @Andy-Jost WDYT?

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
leofang: and avoid `except*`, so it should return an int instead, something like 
```suggestion
cdef int _validate_value_against_bitwidth(int bitwidth, int64_t value, bint is_signed=False) except?-1:
```

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
kkraus14: I was referring to allowing a user to pass a memoryview object of 1, 2, or 4 bytes in length to be used as the fill value. I.e. `Buffer.fill(b'abcd')`.

rparolin:
My bad, I didn't check if you had auto-merged enabled. @Andy-Jost please make sure any feedback from Keith and Leo is incorporated.

leofang:
No worries. Tracking this in #1345.

Andy-Jost:
> My bad, I didn't check if you had auto-merged enabled. @Andy-Jost please make sure any feedback from Keith and Leo is incorporated.

Will follow up to fix, no worries.

PR #1319, opened 2025-12-05
Bump cuda-pathfinder, cuda-bindings versions to a0

rwgk:
Standard approach to avoid confusion between released and development versions.


rwgk:
> Thanks, Ralf!
> 
> Q: Couldn't we just admin-merge this without running CI?

Yes, let's do that next time.

PR #1323, opened 2025-12-05
Fix #1320: Backport cybind changes to 12.9.x branch

### LINKED ISSUE #1320: Backport cuFILE `cdef class` updates

mdboom:


leofang:
> If we don't plan to add another PR for 12.9.5, then we need to add a `12.9.5-notes.rst` to both https://github.com/NVIDIA/cuda-python/tree/main/cuda_python/docs/source/release and https://github.com/NVIDIA/cuda-python/tree/main/cuda_bindings/docs/source/release. Once #1326 is completed we will be able to catch these, at the release time.

NVM, @rwgk reminded me offline that this PR targets the bp branch, and we can't generate docs from there (https://github.com/NVIDIA/cuda-python/issues/809). Let's update the docs in a separate PR. 

Merging, thanks @mdboom!

PR #1340, opened 2025-12-08
Add coverage automation

mdboom:
This is deploying to the GitHub Pages branch at: https://nvidia.github.io/cuda-python/coverage/

This generates a combined coverage report for all of the tests in this repo (cuda_pathfinder, cuda_bindings and cuda_core), which is probably the right thing for a monorepo like this.  We could, alternatively, create separate coverage reports for each of them.

All of these coverage tools are very sensitive to the `cwd`, which is why the GHA config looks a bit funny in that regard.  It seems like the solution is to install all of the packages, and then run the tests from the installation directory, which feels wrong, but I couldn't find any other combination (including path remapping, editable installs) that did the right thing to combine results from 3 test suites correctly and didn't cause the Cython coverage plugin to raise an exception.

The biggest shortcoming is that we only run in one platform and hardware config, so it would miss anything that is platform or config specific.  I think we can get to that, but it would involve some shuttling around of artifacts between different jobs and combining results from Linux and Windows is likely to be quite fiddly.  I think it's still worth merging this as-is with that known issue.

I envision this being set up as a nightly job (with a manual trigger as well).  It's pretty expensive, since you have to rebuild everything from source with a special Cython config and compiler flags, and then runtimes are also slower.  Therefore, I don't think it's suitable for per-PR CI.  There are nifty tools like coveralls or codecov.io that will do graphs with per-PR changes in coverage, but IME those are a bit noisy and full of false positives, even notwithstanding the slowness.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -207,6 +207,12 @@ flowchart TD
 - **Parallel execution**: Matrix builds across Python versions and platforms
 - **Component isolation**: Each component (core, bindings, pathfinder, python) can be built/released independently
 
+## Code coverage
+
+Code coverage reports are produced nightly and posted to [GitHub Pages](https://nvidia.github.io/cuda-python/coverage).
+
+Known limitations: Code coverage is only run on Linux x86_64 with an a100 GPU.  We plan to add more platform and GPU coverage in the future.
+
 ---
 
 <a>1</a>: The `cuda-python` meta package shares the same license and the contributing guidelines as those of `cuda-bindings`.


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,138 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+name: "CI: Coverage"
+
+on:
+  # TODO: Remove this -- this just makes it possible to test during review
+  pull_request:
+    branches:
+      - main
```
cryos: Our CI works a little differently, remove `pull_request`, add in under `branches`:

```
    - "pull-request/[0-9]+"
```

mdboom:
> The only thing that caught my eye was in the gh workflow yml it appears that we are requiring a node with a GPU. Do we need to run this coverage on a node that has a GPU attached or can it run on a lighter weight node and keep the GPU nodes available for other workloads?

We need to run on a GPU, otherwise most of the tests will be skipped, and the coverage information won't be meaningful.

We could do the /builds/ on commodity hardware, and pass those to GPU machines as artifacts.  We can do this when we add Windows support and coverage on a broader range of GPUs.  But for now that's the reason I'm choosing to do this nightly and on-demand rather than on every pull request -- in order to reduce usage of the GPU machines.

mdboom:
I'm removing the bit that makes this run on pull requests (included only for testing this PR) before merging.

PR #1347, opened 2025-12-09
Un-disable ruff E722 Do not use bare `except`

rwgk:
1. Remove `E722` from `[lint.per-file-ignores]` in ruff.toml
2. Fix one bare `except` error in test_cufile.py
3. It turns out the skip condition in cuda_bindings/examples/3_CUDA_Features/simpleCudaGraphs_test.py is not needed in our CI testing.

### DIFF
--- a/cuda_bindings/examples/3_CUDA_Features/simpleCudaGraphs_test.py
+++ b/cuda_bindings/examples/3_CUDA_Features/simpleCudaGraphs_test.py
@@ -5,7 +5,6 @@ import ctypes
 import random as rnd
 
 import numpy as np
-import pytest
 from common import common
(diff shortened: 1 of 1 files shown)


rwgk:
@leofang Can you think of a situation in which the skip condition in cuda_bindings/examples/3_CUDA_Features/simpleCudaGraphs_test.py is needed?

leofang:
It is needed in CUDA 12.x wheels, where the needed headers are splayed into different paths under site-packages. This is not caught in the CI because we don't run cuda-bindings 12.x tests on main. But I think it is OK to removed it from main, now that this issue is fixed with CUDA 13.x.

PR #1349, opened 2025-12-09
Add experimental NVML bindings to 12.9.x branch

mdboom:


cuda_bindings/cuda/bindings/_internal/_nvml.pxd
```diff
@@ -0,0 +1,360 @@
...
+cdef nvmlReturn_t _nvmlDeviceGetPdi(nvmlDevice_t device, nvmlPdi_t* pdi) except?_NVMLRETURN_T_INTERNAL_LOADING_ERROR nogil
+cdef nvmlReturn_t _nvmlDeviceSetHostname_v1(nvmlDevice_t device, nvmlHostname_v1_t* hostname) except?_NVMLRETURN_T_INTERNAL_LOADING_ERROR nogil
+cdef nvmlReturn_t _nvmlDeviceGetHostname_v1(nvmlDevice_t device, nvmlHostname_v1_t* hostname) except?_NVMLRETURN_T_INTERNAL_LOADING_ERROR nogil
+cdef nvmlReturn_t _nvmlDeviceGetNvLinkInfo(nvmlDevice_t device, nvmlNvLinkInfo_t* info) except?_NVMLRETURN_T_INTERNAL_LOADING_ERROR nogil
+cdef nvmlReturn_t _nvmlDeviceReadWritePRM_v1(nvmlDevice_t device, nvmlPRMTLV_v1_t* buffer) except?_NVMLRETURN_T_INTERNAL_LOADING_ERROR nogil
+cdef nvmlReturn_t _nvmlDeviceGetGpuInstanceProfileInfoByIdV(nvmlDevice_t device, unsigned int profileId, nvmlGpuInstanceProfileInfo_v2_t* info) except?_NVMLRETURN_T_INTERNAL_LOADING_ERROR nogil
+cdef nvmlReturn_t _nvmlDeviceGetSramUniqueUncorrectedEccErrorCounts(nvmlDevice_t device, nvmlEccSramUniqueUncorrectedErrorCounts_t* errorCounts) except?_NVMLRETURN_T_INTERNAL_LOADING_ERROR nogil
```
kkraus14, line 360: This API was added in CUDA 13.0, I don't think it should be in the 12.9.x branch?

mdboom: Got it.  Regenerating without the 13.0.1 headers engaged.

mdboom: It also looks like we don't have a good story in the generator for hand-written APIs that don't exist across all versions.  I've hand-removed them for now, but that's going to be a maintenance headache.

mdboom: So it turns out (in a side conversation with @leofang) -- this is working as intended.  The "megaheader" approach will include APIs from all versions, but the ones that don't exist on 12.9.x will raise "not found" exceptions.  /Not/ doing this creates a whole other set of problems, since then types go missing and our hand-written APIs no longer compile, which then need to be conditionally compiled out (which we don't really have infrastructure for).

cuda_bindings/cuda/bindings/_internal/_nvml.pxd
```diff
@@ -0,0 +1,360 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
+#
+# This code was automatically generated across versions from 12.9.1 to 13.0.1. Do not modify it directly.
```
kkraus14: Should this be generated through 13.1.0 to be consisent with `main`?

mdboom: It was consistent with main at the time of filing :)  I'll merge the 13.1 and 12.9 cybind MRs and regenerate, then merge this.

PR #1352, opened 2025-12-10
Fix NVML tests for machines with more than one device

mdboom:
## Description

As @rwgk discovered, all of the tests using the `for_all_devices` fixture fail with `fixture function has more than one 'yield'`.  I didn't realize that wasn't allowed.  This changes the fixture to just yield the whole list of devices.


mdboom:
Good idea.

> I'd try to give that as a prompt to Cursor. Good chance it'll "just do it".

I did one of these, and VSCode/Copilot was able to infer the rest.

PR #1353, opened 2025-12-10
Add new NVML APIs in CTK 13.1

mdboom:


cuda_bindings/cuda/bindings/_nvml.pyx
leofang, line 1: The diff for this file is confusing. I expect to see a ton of added lines and no other changes or removals. But the diff is nontrivial to me. What is happening...?

mdboom: AFAICT, types are being emitted in a different order. Since an `AUTO_LOWPP_CLASS` has a lot of common boilerplate, the diff algorithm is getting confused.  Looking closely, I don't see anything that has been removed.  Maybe we should sort the types before generating them to prevent this in the future?

mdboom: Ah, actually, I don't think the order is changing, so sorting wouldn't help.  It's really just types being "inserted" and the diff algo getting confused.

leofang: Ahhh ok, thanks for reminding me. I recall we discussed this recently. We shouldn't enforce the sort order (it's nontrivial). Resolving.

leofang (approved):

PR #1357, opened 2025-12-10
Design proposal: Buffer.fill() redesign

### LINKED ISSUE #1345: Revisit `Buffer.fill()`
PR #1318 was mistakenly prematurely merged: https://github.com/NVIDIA/cuda-python/pull/1318#issuecomment-3629730971

xref: https://github.com/NVIDIA/cuda-python/pull/1318#issuecomment-3629515796 and other open threads in the same PR. 

This must be resolved before we can cut a release, otherwise we'd have to make a breaking change in the next release (v0.6.0) which is not ideal.

Andy-Jost:
## Summary

This PR contains a design document proposing changes to `Buffer.fill()` to address feedback from PR #1318 that was prematurely merged.

**Please review the design document and provide feedback before implementation proceeds.**

## Related

- Closes #1345
- Follow-up to #1318

cc @leofang @kkraus14 @rparolin

### DIFF
--- a/cuda_core/docs/design/buffer-fill-redesign.md
+++ b/cuda_core/docs/design/buffer-fill-redesign.md
@@ -1,0 +1,77 @@
+# Buffer.fill() Redesign Proposal
+
+**Issue**: [#1345 - Revisit `Buffer.fill()`](https://github.com/NVIDIA/cuda-python/issues/1345)  
+**Author**: Andy Jost  
+**Date**: December 10, 2025
+
+## Background
+
+PR #1318 implemented `Buffer.fill(value, width, *, stream)` but was merged before review feedback was addressed. This document proposes a simplified API based on that feedback.
+
+## Proposed API
+
+```python
+def fill(self, value, *, stream: Stream | GraphBuilder):
+    """Fill buffer with a repeating byte pattern.
+    
+    Parameters
+    ----------
+    value : int or buffer-protocol object
+        - int: Must be in range [0, 256). Converted to 1 byte.
+        - buffer-protocol object: Must be 1, 2, or 4 bytes.
+    stream : Stream | GraphBuilder
+        Stream for the asynchronous fill operation.
+    
+    Raises
+    ------
+    TypeError
+        If value is not an int and does not support the buffer protocol.
+    ValueError
+        If value byte length is not 1, 2, or 4.
+        If buffer size is not divisible by value byte length.
+    OverflowError
+        If int value is outside [0, 256).
+    """
+```
+
+## Implementation
+
+```python
+def get_fill_pattern(value):
+    if isinstance(value, int):
+        return value.to_bytes(1, 'little')  # Raises OverflowError if not in [0, 256)
+    mv = memoryview(value)
+    return mv.tobytes()
+
+pattern = get_fill_pattern(value)
+L = len(pattern)
+if L not in (1, 2, 4):
+    raise ValueError(f"value must be 1, 2, or 4 bytes, got {L}")
+if buffer_size % L != 0:
+    raise ValueError(f"buffer size ({buffer_size}) must be divisible by {L}")
+
+# Call appropriate cuMemsetD{8,16,32}Async based on L
+```
+
+## Examples
+
+```python
+# Byte fill (1 byte)
+buffer.fill(0, stream=stream)              # Zero memory
+buffer.fill(0xFF, stream=stream)           # Fill with 0xFF
+
+# Multi-byte fill via numpy scalars
+buffer.fill(np.uint16(0x1234), stream=stream)      # 2-byte pattern
+buffer.fill(np.float32(1.0), stream=stream)        # 4-byte pattern
+
+# Raw bytes
+buffer.fill(b'\xDE\xAD\xBE\xEF', stream=stream)    # 4-byte pattern
+```
+
+## Changes from Current API
+
+| Current | Proposed |
+|---------|----------|
+| `fill(value, width, *, stream)` | `fill(value, *, stream)` |
+| Explicit `width` parameter | Width inferred from value |
+| Only accepts `int` | Accepts `int` or buffer-protocol objects |


Andy-Jost:
@kkraus14 @leofang @rparolin

I don't plan to commit the proposal document buffer-fill-redesign.md. Just posting this to close on the redesign before reimplementing the feature.

You can toggle between "source diff" and "rich diff" for commenting and rendering, respectively.

cuda_core/docs/design/buffer-fill-redesign.md
```diff
@@ -0,0 +1,350 @@
...
+   - `numpy.int8/uint8` → 1-byte fill
+   - `numpy.int16/uint16` → 2-byte fill
+   - `numpy.int32/uint32` → 4-byte fill
+   - `bytes` (length 1, 2, or 4) → raw pattern fill
+
+2. **StridedMemoryView.fill()** (new):
+   - Uses the view's dtype to determine width and signedness
+   - Value is validated against the dtype's range
+
+3. **Remove `width` parameter** from current API (breaking change, but before release)
```
Andy-Jost: Here's the main part of the proposal.

cuda_core/docs/design/buffer-fill-redesign.md
```diff
@@ -0,0 +1,350 @@
...
+## Recommendation
+
+**Option C (Hybrid Approach)** provides the best balance:
+
+1. **Buffer.fill()** accepts:
+   - `int` → 1-byte fill (value 0-255)
+   - `numpy.int8/uint8` → 1-byte fill
+   - `numpy.int16/uint16` → 2-byte fill
+   - `numpy.int32/uint32` → 4-byte fill
+   - `bytes` (length 1, 2, or 4) → raw pattern fill
```
kkraus14: I think all 4 of these could be rolled into "objects that support Python buffer protocol" where as long as the object supports Python buffer protocol and is either 1, 2, or 4 bytes we match that to the corresponding fill ~(We could also simulate doing 8 byte fills using the 4 byte fill CUDA API under the hood?)~

Numpy scalars support Python buffer protocol so it would allow us to support those types without adding to our numpy dependency footprint in any way.

cuda_core/docs/design/buffer-fill-redesign.md
```diff
@@ -0,0 +1,350 @@
...
+
+2. **StridedMemoryView.fill()** (new):
+   - Uses the view's dtype to determine width and signedness
+   - Value is validated against the dtype's range
```
kkraus14: If the `StridedMemoryView` is non-contiguous, do we not support it for now or does the fill operate ignoring strides?

cuda_core/docs/design/buffer-fill-redesign.md
```diff
@@ -0,0 +1,350 @@
...
+    # ... proceed with cuMemsetD*Async
+```
+
+**Future enhancement**: A kernel-based fallback could be added later to support strided fills if there's user demand. This would require launching a simple fill kernel that iterates over elements using the stride information.
+
+## Open Questions
+
+1. **Should plain Python `int` default to 1-byte or 4-byte?**
+   - Proposal: 1-byte (consistent with "Buffer is untyped bytes")
+   - Alternative: Error if int > 255, requiring explicit dtype for larger values
```
kkraus14: Python integers have a built-in `to_bytes` function that errors if you try to call it with something larger than 1-byte where it makes sense to limit it to 1-byte in my opinion.

cuda_core/docs/design/buffer-fill-redesign.md
```diff
@@ -0,0 +1,350 @@
...
+
+2. **Should we support `numpy.int64/uint64`?**
+   - CUDA only provides cuMemsetD{8,16,32}Async, no 64-bit variant
+   - Proposal: Error for 64-bit types with clear message
```
kkraus14: This is quite unfortunate, but I agree this is the right behavior for now.

kkraus14:
@Andy-Jost I'm aligned to the approach you've detailed out here. My only suggested change is related to the use of Python buffer protocol instead of specifically handling bytes and numpy scalar objects. The rest looks good to me!

cuda_core/docs/design/buffer-fill-redesign.md
Andy-Jost: Ok, I see now. This seems like a great approach. I'll look into it.

I don't think there is a straightforward way to support 8-byte values.

Andy-Jost:
> @Andy-Jost I'm aligned to the approach you've detailed out here. My only suggested change is related to the use of Python buffer protocol instead of specifically handling bytes and numpy scalar objects. The rest looks good to me!

Thanks for the quick feedback, Keith. I'll follow up with a code change based on these suggestions.

cuda_core/docs/design/buffer-fill-redesign.md
kkraus14: > I don't think there is a straightforward way to support 8-byte values.

Yup, apologies, that was a brain fart on my end. Will edit it to strikethrough.

leofang:
@kkraus14 @Andy-Jost 

Question: 
1. Wearing my array API hat, wouldn't Option C trap us in the unfortunate business of implementing [type promotion rules](https://data-apis.org/array-api/latest/API_specification/type_promotion.html)? What if `SMV.dtype` is not compatible with the value type? Is it something that we want to handle at the cuda.core level?
2. Can we verify using Python buffer protocol is faster than the ParamHolder dispatcher? My impression is that the buffer protocol overhead is nontrivial. Conversely, if it is faster we should create a task to track rewriting the dispatcher to use buffer protocol.

Andy-Jost:
> @kkraus14 @Andy-Jost
> 
> Question:
> 
> 1. Wearing my array API hat, wouldn't Option C trap us in the unfortunate business of implementing [type promotion rules](https://data-apis.org/array-api/latest/API_specification/type_promotion.html)? What if `SMV.dtype` is not compatible with the value type? Is it something that we want to handle at the cuda.core level?

Let me update the proposal and to strip it down (e.g., Option C is off the table now).

If we go with Keith's suggestion to rely on the buffer protocol, we can sidestep type promotion for the basic `Buffer.fill`. However, there are gotchas with `StridedMemoryView.fill` and maybe this is what you have in mind. For instance:

```
view = StridedMemoryView.from_buffer(buffer, layout, dtype=np.float32)
view.fill(1, stream=stream)
```

Is the fill value `np.float32(1)` or the byte `0x01`?

Given the added complexity, I'm inclined to:

  1. Implement `Buffer.fill` with the buffer protocol approach.
  2. Defer `StridedMemoryView.fill` for now, as it needs more design


> 2. Can we verify using Python buffer protocol is faster than the ParamHolder dispatcher? My impression is that the buffer protocol overhead is nontrivial. Conversely, if it is faster we should create a task to track rewriting the dispatcher to use buffer protocol.

I don't have any sense of how the buffer protocol will compare with the `ParamHolder` dispatch. `Buffer.fill` is potentially simpler because it only supports 1/2/4 bytes. How about this: I can reimplement `Buffer.fill` based on the buffer protocol and benchmark the performance against `ParamHolder`. That at least allows us to fix the API right away. Then, based on the benchmark results we can decide what to do next with `ParamHolder`.

Andy-Jost:
New (much shorter) proposal doc is live.

cuda_core/docs/design/buffer-fill-redesign.md
```diff
@@ -0,0 +1,77 @@
...
+
+```python
+def fill(self, value, *, stream: Stream | GraphBuilder):
+    """Fill buffer with a repeating byte pattern.
+    
+    Parameters
+    ----------
+    value : int or buffer-protocol object
+        - int: Must be in range [0, 256). Converted to 1 byte.
+        - buffer-protocol object: Must be 1, 2, or 4 bytes.
```
kkraus14, line 21: super nitpick, but there's the `collections.abc.Buffer` type that exists specifically to indicate objects that support buffer protocol

kkraus14:
> 2\. Can we verify using Python buffer protocol is faster than the ParamHolder dispatcher? My impression is that the buffer protocol overhead is nontrivial. Conversely, if it is faster we should create a task to track rewriting the dispatcher to use buffer protocol.

I think we could special case and fast path certain object types and then fall back to buffer protocol if desired? I.E. support `bytes` objects zero copy directly via the CPython API first, then support `int` objects via a fastpath using the CPython API, and then fall back to buffer protocol?

leofang:
There is one thing that came to my mind last night. @Andy-Jost you saw @chloechia4's design doc on pythonic cuFILE support. There is one bit of cuFILE that I like very much (at least the concept of it), which is they support **deferred** C scalar values by accepting pointers in their read/write APIs. 

Our current `Buffer.fill` design cannot be deferred. The value must be to be materialized at the invocation time, due to the underlying memset C API constraint. We could work around this by conditionally dispatching to memcpy, but then we need a way to express this special case and I suspect we'd have to use Python int to serve as the pointer address (like in `launch`). WDYT?

Andy-Jost:
> There is one thing that came to my mind last night. @Andy-Jost you saw @chloechia4's design doc on pythonic cuFILE support. There is one bit of cuFILE that I like very much (at least the concept of it), which is they support **deferred** C scalar values by accepting pointers in their read/write APIs.
> 
> Our current `Buffer.fill` design cannot be deferred. The value must be to be materialized at the invocation time, due to the underlying memset C API constraint. We could work around this by conditionally dispatching to memcpy, but then we need a way to express this special case and I suspect we'd have to use Python int to serve as the pointer address (like in `launch`). WDYT?

Discussed: not something we can support due to driver API limitations

PR #1360, opened 2025-12-11
Forward port of #1349

mdboom:
This adds the changes in #1349 that make NVML dynamically support 12.9 correctly to the main branch.  This should have no net effect on 13.0+ -- it is just to make the branches match.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: 9c89adb347f6b973f4905a4be0051eb2ecf85dea  # frozen: v0.13.3
     hooks:
-      - id: ruff
+      - id: ruff-check
         args: [--fix, --show-fixes]
       - id: ruff-format
 
--- a/cuda_pathfinder/pixi.lock
+++ b/cuda_pathfinder/pixi.lock
@@ -1,0 +1,942 @@
+version: 6
+environments:
+  cu13:
+    channels:
+    - url: https://conda.anaconda.org/conda-forge/
+    packages:
+      linux-64:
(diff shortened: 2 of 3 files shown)


cuda_bindings/cuda/bindings/_nvml.pxd
```diff
@@ -32,19 +32,16 @@ ctypedef nvmlProcessInfo_v2_t ProcessInfo_v2
 ctypedef nvmlNvLinkUtilizationControl_t NvLinkUtilizationControl
 ctypedef nvmlViolationTime_t ViolationTime
 ctypedef nvmlUUIDValue_t UUIDValue
-ctypedef nvmlPlatformInfo_v1_t PlatformInfo_v1
```
kkraus14, line 35: Were these removed due to codegen issues and something we intend to readd in the future?

mdboom: They were changed from a type-erased type to a `AUTO_LOWPP_CLASS`.  It looks like a deletion in the low-level, but it's actually a move from the low-level to the high-level.  

Initially this type was not wrapped specially because it was deprecated in 13.0.  In order to support 12.9, we needed to wrap this class as well and use it as a return type on 12.9.

mdboom: That said, is it a bug that `AUTO_LOWPP_CLASS` doesn't include a declaration in the `.pxd` file?  Seems to be long-standing behavior in any event.  @leofang: any insights?

leofang: This is Python-layer pxd which we don't guarantee any stability. I think it might just be a simple oversight that the new CLASS is not exposed here (if this is what you meant), but the reality is no one is supposed to use these other than ourselves if we like. I don't think it is a bug to sweat on unless it's trivial to fix.

PR #1361, opened 2025-12-11
[cuda_pathfinder] Adding pixi support

rparolin:
This PR adds pixi build system support to cuda_pathfinder, enabling it to be built and tested using pixi alongside cuda_bindings and cuda_core. The changes introduce a new pixi.toml configuration file with support for multiple environments and platforms and the pixi.lock file for reproducible builds/tests.

### DIFF
--- a/cuda_pathfinder/pixi.toml
+++ b/cuda_pathfinder/pixi.toml
@@ -1,0 +1,48 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+[workspace]
+channels = ["conda-forge"]
+platforms = ["linux-64", "linux-aarch64", "win-64"]
+preview = ["pixi-build"]
+
+[workspace.build-variants]
+python = ["3.10.*", "3.11.*", "3.12.*", "3.13.*", "3.14.*"]
+
+[dependencies]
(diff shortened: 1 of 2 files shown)


rparolin:
@greptileai please review

cuda_pathfinder/pixi.toml
```diff
@@ -0,0 +1,48 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+[workspace]
+channels = ["conda-forge"]
+platforms = ["linux-64", "linux-aarch64", "win-64"]
+preview = ["pixi-build"]
+
+[workspace.build-variants]
+python = ["3.10.*", "3.11.*", "3.12.*", "3.13.*", "3.14.*"]
```
kkraus14, line 11: cuda pathfinder is pure python where I think there's a way to do a `noarch` package to avoid unnecessary build variants by python version?

cuda_pathfinder/pixi.toml
```diff
@@ -0,0 +1,48 @@
...
+
+[feature.cu13.system-requirements]
+cuda = "13"
+
+[feature.cu13.dependencies]
+cuda-version = ">=13.1,<14"
+
+[environments]
+default = { features = ["test"], solve-group = "default" }
+cu13 = { features = ["cu13", "test"], solve-group = "cu13" }
```
kkraus14, line 27: We need to support `cu12` for pathfinder as well

PR #1363, opened 2025-12-11
feat: add `size` property to `StridedMemoryView`

cpcloud:
Add a `size` property to `StridedMemoryView` to avoid having dig into `layout` to get this information.

### DIFF
--- a/cuda_core/cuda/core/experimental/_memoryview.pyx
+++ b/cuda_core/cuda/core/experimental/_memoryview.pyx
@@ -297,6 +297,10 @@ cdef class StridedMemoryView:
         return self.get_layout()
 
     @property
+    def size(self) -> int:
+        return self.get_layout().get_volume()
+
+    @property
     def shape(self) -> tuple[int, ...]:
         """
         Shape of the tensor.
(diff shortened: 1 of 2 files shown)


cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -296,6 +296,10 @@ cdef class StridedMemoryView:
         """
         return self.get_layout()
 
+    @property
+    def size(self) -> int:
```
kkraus14, line 300: Should we consider caching this?

kkraus14 (approved):

cuda_core/cuda/core/experimental/_memoryview.pyx
cpcloud: I think so. This property is accessed on every kernel invocation in numba-cuda, so I would probably make the case that all the shape/stride information should be cached/made into a instance attribute.

I can experiment with that in a follow-up.

cpcloud: Actually, this is all going through `cdef` functions, so I'd just leave it as is for now, plus there's the fact that `cdef` classes don't have a `__dict__` attribute by default, which `functools.cached_property` requires. I'd rather not use `functools.cache` just for this one property.

Issue #1359, opened 2025-12-11
Is `MemoryResource` base class changed on the main branch on purpose?


leofang:
```
 ✓ Check 0.4.0 MemoryResource
   $ cd /local/home/leof/dev/cuda-python && python -c "
   import sys
   sys.path.insert(0, '/tmp/cuda-core-0.4.0')

   from cuda.core.experimental._memory import MemoryResource
   import inspect

   print('cuda-core 0.4.0 MemoryResource:')
   print('=' * 60)
   print(f'Is abstract: {inspect.isabstract(MemoryResource)}')
   print(f'Has __abstractmethods__: {hasattr(MemoryResource, \"__abstractmethods__\")}')
   if hasattr(MemoryResource, '__abstractmethods__'):
       print(f'Abstract methods: {sorted(MemoryResource.__abstractmethods__)}')

   print(f'\\nAll public methods and properties:')
   for name in sorted(dir(MemoryResource)):
       if not name.startswith('_'):
           print(f'  - {name}')
   "
   ╭───────────────────────────────────────────────────────────────────────────────────╮
   │ cuda-core 0.4.0 MemoryResource:                                                   │
   │ ============================================================                      │
   │ Is abstract: False                                                                │
   │ Has __abstractmethods__: False                                                    │
   │                                                                                   │
   │ All public methods and properties:                                                │
   │   - allocate                                                                      │
   │   - deallocate                                                                    │
   │   - device_id                                                                     │
   │   - is_device_accessible                                                          │
   │   - is_host_accessible                                                            │
   │ <exited with exit code 0>                                                         │
   ╰───────────────────────────────────────────────────────────────────────────────────╯

 ● Interesting! So:

     - 0.3.0: MemoryResource IS abstract, has allocate, deallocate, device_id, is_device_accessible, is_host_accessible as abstract
     - 0.4.0: MemoryResource is NOT abstract, still has the 5 members but NOT abstract
     - Current (main branch): MemoryResource has abstract methods but missing the 3 property definitions

   This is a breaking change. The ABC interface is broken because the properties are missing. Let me add them back:

 ○ Edit cuda/core/experimental/_memory/_buffer.pyx

 ╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
 │ Edit file                                                                                                                                                            │
 │                                                                                                                                                                      │
 │ _buffer.pyx +18 -0                                                                                                                                                   │
[...]

PR #1364, opened 2025-12-11
Add `PinnedMemoryResource` and `ManagedMemoryResource`

### LINKED ISSUE #718: [FEA]: Add a `ManagedMemoryResource`
### Is this a duplicate?

### Area

cuda.bindings

We don't expose a memory resource for allocating managed memory, it would be good to expose something like the [`UnifiedMemoryResource`](https://github.com/NVIDIA/cuda-python/blob/24fde1755ad1efa6e7d73b2236dad9681fa0b727/cuda_core/tests/test_memory.py#L68) that we wrote for testing.

Expose the memory resource we wrote for tests.

_No response_

### Additional context

_No response_

### LINKED ISSUE #1074: IPC tests do not exit cleanly

### LINKED ISSUE #1141: Add a `PinnedMemoryResource`
Counterpart of #718 but for pinned memory.

We already have `LegacyPinnedMemoryResource` that wraps `cuMemHostAlloc` under the hood. Both issues (this and #718) are about exposing the CUDA 13 driver features for asynchronous allocating and deallocating pinned/managed memory.

leofang:
## Description

closes #1141 
closes #718 
closes #1074

This pull request introduces significant refactoring and new features to the CUDA experimental memory resource subsystem. The main highlights include the introduction of a shared internal base class (`_MemPool`) for device, managed, and pinned memory resources, the addition of new managed and pinned memory resource types, and the unification and simplification of IPC (Inter-Process Communication) logic for memory pools. These changes improve extensibility, code clarity, and set the groundwork for future features.

Key changes:

**Refactoring and Internal Architecture:**

* Introduced a new internal `_MemPool` base class (`cuda_core/cuda/core/experimental/_memory/_memory_pool.pxd`) to unify the implementation for all memory resource types (device, managed, pinned). Device, managed, and pinned memory resources now inherit from this shared class, reducing code duplication and simplifying future extensions. [[1]](diffhunk://#diff-e5c7f9d5c55d1b3c7a64c323b322eaf329c85ab2af9deed7c8fe8c27feb77e17R1-R28) [[2]](diffhunk://#diff-f4b16f8b68b018dc749335e2aecf287ba341c3442d14a062332c42a7d8ca28f0L5-R10) [[3]](diffhunk://#diff-4532a4fb53ba205439e4861e2facf2c3a549b5d7b3642dacea997d7a7c7f9f23R1-R9) [[4]](diffhunk://#diff-a567cc0b447e21bb80719f6aa4ac880d4a5394fe2c26d7d38a25662e821165aaR1-R10)

* Refactored IPC logic to operate on `_MemPool` instead of `DeviceMemoryResource`, updating all relevant Cython signatures and implementations to use the new base class. This change unifies IPC handling for all memory resource types and simplifies buffer import/export across processes. [[1]](diffhunk://#diff-a82607f79a1dea4549aebabd41c58a179901e147e35791f3be0e031270a2e826L7-R10) [[2]](diffhunk://#diff-a82607f79a1dea4549aebabd41c58a179901e147e35791f3be0e031270a2e826L56-R64) [[3]](diffhunk://#diff-d09e4be2286a3f86012cc0648c90fb000f1eb4339e8799c95f5c95362630d100L145-L155) [[4]](diffhunk://#diff-d09e4be2286a3f86012cc0648c90fb000f1eb4339e8799c95f5c95362630d100L172-R167) [[5]](diffhunk://#diff-d09e4be2286a3f86012cc0648c90fb000f1eb4339e8799c95f5c95362630d100L192-R184) [[6]](diffhunk://#diff-d09e4be2286a3f86012cc0648c90fb000f1eb4339e8799c95f5c95362630d100L212-R202) [[7]](diffhunk://#diff-d09e4be2286a3f86012cc0648c90fb000f1eb4339e8799c95f5c95362630d100L234-R228) [[8]](diffhunk://#diff-d09e4be2286a3f86012cc0648c90fb000f1eb4339e8799c95f5c95362630d100L251-R238)

**New Features:**

* Added support for managed and pinned memory resources, including their corresponding options and registration in the experimental API. This includes new Cython and Python files for `ManagedMemoryResource` and `PinnedMemoryResource`, and their exposure in the public API. [[1]](diffhunk://#diff-4aba28d70ee29dbb0ea8357bd51cd7f5295f013b2d0a5a78e99ab34de8a83bd9R1-R195) [[2]](diffhunk://#diff-4532a4fb53ba205439e4861e2facf2c3a549b5d7b3642dacea997d7a7c7f9f23R1-R9) [[3]](diffhunk://#diff-a567cc0b447e21bb80719f6aa4ac880d4a5394fe2c26d7d38a25662e821165aaR1-R10) [[4]](diffhunk://#diff-63e2cfe3cd454af380d1c38c78ba71d64b8096fbae6673e061c354b5a809c477R47-R51) [[5]](diffhunk://#diff-5c64e4f9464500a75054b90e1cc489c234c376f946d95b22a54c4e46b7b5db56R10-R11)

* Implemented `ManagedMemoryResourceOptions` for customizable managed memory resource creation, and ensured correct handling of host/device preferences and IPC limitations for managed memory pools.

**API and Type Improvements:**

* Updated buffer import logic to accept both `DeviceMemoryResource` and `PinnedMemoryResource`, reflecting the new extensible memory resource architecture. [[1]](diffhunk://#diff-0c575d585121f3088d1a460ea98deef018725b103c6957a3fd816f29cae5595eL11-R12) [[2]](diffhunk://#diff-0c575d585121f3088d1a460ea98deef018725b103c6957a3fd816f29cae5595eL109-R110)

* Ensured type consistency and correctness in Cython code, such as using integer major versions for CUDA build checks and compile-time environments. [[1]](diffhunk://#diff-f548f2eb738fc5eb41d25700e66f4c87c32cd6da393a2aae3120eb45a9279ab4L98-R98) [[2]](diffhunk://#diff-ff5555a8b93fbe7cd2e8944979f745f2efb8a552fdad968dcfe5843aefc78868L1083-R1083)

These changes collectively modernize the memory resource subsystem, improve maintainability, and add new capabilities for advanced CUDA memory management.

### DIFF
--- a/cuda_core/tests/conftest.py
+++ b/cuda_core/tests/conftest.py
@@ -26,37 +26,7 @@ from cuda.core.experimental import (
 from cuda.core.experimental._utils.cuda_utils import handle_return
 
 
-def _check_pinned_memory_available():
-    """Check if PinnedMemoryResource is available (CUDA 13.0+)."""
-    try:
-        device = Device()
-        return hasattr(device.properties, "host_memory_pools_supported")
-    except Exception:
-        return False
-
-
-def _check_managed_memory_available():
-    """Check if ManagedMemoryResource is available (CUDA 13.0+)."""
-    try:
-        device = Device()
-        return hasattr(device.properties, "memory_pools_supported") and hasattr(device.properties, "managed_memory")
-    except Exception:
-        return False
-
-
-# Skip marks for tests requiring CUDA 13.0+
-skipif_pinned_memory_unavailable = pytest.mark.skipif(
-    not _check_pinned_memory_available(), reason="PinnedMemoryResource requires CUDA 13.0 or later"
-)
-
-skipif_managed_memory_unavailable = pytest.mark.skipif(
-    not _check_managed_memory_available(), reason="ManagedMemoryResource requires CUDA 13.0 or later"
-)
-
-
-# Helper functions for runtime checks within tests
 def skip_if_pinned_memory_unsupported(device):
-    """Skip test if device doesn't support host memory pools or CUDA < 13."""
     try:
         if not device.properties.host_memory_pools_supported:
             pytest.skip("Device does not support host mempool operations")
@@ -65,16 +35,14 @@ def skip_if_pinned_memory_unsupported(device):
 
 
 def skip_if_managed_memory_unsupported(device):
-    """Skip test if device doesn't support managed memory pools or CUDA < 13."""
     try:
-        if not device.properties.memory_pools_supported or not device.properties.managed_memory:
+        if not device.properties.memory_pools_supported or not device.properties.concurrent_managed_access:
             pytest.skip("Device does not support managed memory pool operations")
     except AttributeError:
         pytest.skip("ManagedMemoryResource requires CUDA 13.0 or later")
 
 
 def create_managed_memory_resource_or_skip(*args, **kwargs):
-    """Create ManagedMemoryResource, skipping test if CUDA 13.0+ required."""
     try:
         return ManagedMemoryResource(*args, **kwargs)
     except RuntimeError as e:


cuda_core/tests/test_memory.py
```diff
@@ -512,35 +518,52 @@ def test_buffer_dlpack_failure_clean_up():
...
-
-    This test verifies that the DeviceMemoryResource initializes properly,
-    including the release threshold configuration for performance optimization.
-    """
+def test_modern_device_memory_resource_initialization(memory_resource_factory_with_device):
     device = Device()
+    MR, MRops, use_device_object = memory_resource_factory_with_device
 
-    if not device.properties.memory_pools_supported:
+    if MR is DeviceMemoryResource and not device.properties.memory_pools_supported:
```
cpcloud: I find the repetition across tests and large numbers of conditions very difficult to read and would find it hard to maintain these tests in general.

I think the parameterized fixture `memory_resource_factory_with_device` as-is is not the right approach to making the tests readable.

How about splitting up the tests based on types, along with a fixtures that depend on a (maybe new?) `device` fixture that can be used to invoke `pytest.skip` based on properties? (`device` appears to be used in all of these tests)

In other words, instead of branching, write a test for each branch and move as much of the skipping logic as you can out of the test and into fixtures.

The shared asserts are already repeated, and given that's true, it's more readable to repeat them without branching.

cuda_core/tests/test_memory.py
```diff
@@ -512,35 +518,52 @@ def test_buffer_dlpack_failure_clean_up():
...
-@pytest.mark.parametrize("use_device_object", [True, False])
-def test_device_memory_resource_initialization(use_device_object):
-    """Test that DeviceMemoryResource can be initialized successfully.
-
-    This test verifies that the DeviceMemoryResource initializes properly,
-    including the release threshold configuration for performance optimization.
-    """
+def test_modern_device_memory_resource_initialization(memory_resource_factory_with_device):
     device = Device()
+    MR, MRops, use_device_object = memory_resource_factory_with_device
```
cpcloud: The `use_device_object` only being used in this one test suggests it's not the right abstraction for what you're trying to do. See my comment below. I think additional tests will be readable as opposed to all the branching inside the tests to figure out what assertions are valid.

Fixtures should produce the "what", and the test should most have API calls and assertions. I am well aware this isn't always possible, but I think we can improve the situation here using a different approach.

cuda_core/cuda/core/experimental/_device.pyx
```diff
@@ -1080,7 +1080,7 @@ class Device:
         if self._uuid is None:
             dev = self._id
             with nogil:
-                IF CUDA_CORE_BUILD_MAJOR == "12":
+                IF CUDA_CORE_BUILD_MAJOR == 12:
```
cpcloud, line 1083: This seems like a meaningful change. Was it just silently failing or testing false before?

cuda_core/cuda/core/experimental/_memory/_device_memory_resource.pyx
```diff
@@ -261,7 +160,7 @@ cdef class DeviceMemoryResource(MemoryResource):
         RuntimeError
             If no mapped memory resource is found in the registry.
         """
-        return _ipc.DMR_from_registry(uuid)
+        return <DeviceMemoryResource>(_ipc.MP_from_registry(uuid))
```
cpcloud, line 163: Why all the new casts?

cuda_core/cuda/core/experimental/_device.pyx
leofang: Previously we only allow equality check, so making it a str or int doesn't matter. Now we need inequality check (pinned/managed mempools start with CUDA >=13), so an int is needed.

cuda_core/cuda/core/experimental/_memory/_device_memory_resource.pyx
leofang: Because all MRs are now subclasses of `_MemPool`, and it is simpler to retain Andy's original design by storing the pool instances in the registry, but we know they are of concrete MR types.

cuda_core/tests/test_memory.py
leofang: done in [28138b9](https://github.com/NVIDIA/cuda-python/pull/1364/commits/28138b97f033ac100f1acc5f838869dc88c174df)

cuda_core/tests/test_memory.py
leofang: done in [28138b9](https://github.com/NVIDIA/cuda-python/pull/1364/commits/28138b97f033ac100f1acc5f838869dc88c174df)

cuda_core/tests/memory_ipc/test_serialize.py
```diff
@@ -151,41 +151,20 @@ def test_main(self, ipc_device, ipc_memory_resource):
         pgen.verify_buffer(buffer, seed=True)
         buffer.close()
 
-    def child_main(self, alloc_handle, mr1, buffer_desc, buffer1):
+    def child_main(self, alloc_handle, mr1, buffer_desc, buffer):
```
leofang, line 154: This closes #1074.

This `TestObjectPassing` test is known to have the triple-free problem that is currently deemed as a bug by the CUDA driver. For `DeviceMemoryResource` it is still OK because all API calls return success. For `PinnedMemoryResource` a `CUDA_ERROR_ALREADY_MAPPED` error cannot be worked around and the test without this rewriting would fail. Below is an LLM-generated summary.

***

Root Cause (Issue #1074):

The CUDA driver does NOT refcount imported pointers. The original test had a flawed assumption: "Successful calls to cuMemPoolImportPointer correctly matched with calls to cuMemFree(Async) ought to work" similar to malloc/free. However:

- Each call to cuMemPoolImportPointer with the same descriptor creates a NEW handle to the same memory
- Each handle's destructor calls cuMemFree(Async)  
- Without driver-level refcounting, this causes multiple frees of the same memory

Original Test Problem - Triple Free:

- buffer1 passed as argument → serialized → imported in child (Import No.1)
- buffer2 created from mr1 and descriptor (Import No.2)
- buffer3 created from mr2 and descriptor (Import No.3)
- All three destructors call free → 3 frees, 1 allocation = crash/errors

Why DMR "worked" but PMR failed:
- DMR: Triple-free caused CUDA_ERROR_INVALID_VALUE during cleanup (ignored exceptions in destructors), test logic completed
- PMR: More strict validation, failed immediately with CUDA_ERROR_ALREADY_MAPPED

The Fix: Changed test to pass only MR and descriptor to child, not the buffer itself:
- Parent: Creates buffer, gets descriptor, passes (mr, buffer_desc) to child
- Child: Creates ONE buffer from descriptor → 1 import, 1 free ✅

Key Takeaway: Never import the same IPC pointer multiple times without explicit coordination. The CUDA driver doesn't track this, leading to multiple frees. Each unique buffer object should have exactly one import/free pair.

leofang:
/ok to test [0c22b31](https://github.com/NVIDIA/cuda-python/pull/1364/commits/0c22b3139205449e2723b6927412722ba86b26ec)

leofang:
/ok to test [f9e3c55](https://github.com/NVIDIA/cuda-python/pull/1364/commits/f9e3c55f80e708f1838071886aea77a7b3ad807a)

leofang:
@Andy-Jost PTAL

cuda_core/cuda/core/experimental/_memory/_managed_memory_resource.pyx
```diff
@@ -0,0 +1,119 @@
...
+cdef class ManagedMemoryResourceOptions:
+    """Customizable :obj:`~_memory.ManagedMemoryResource` options.
+
+    Attributes
+    ----------
+    preferred_location : int | None, optional
+        The preferred device location for the managed memory.
+        Use a device ID (0, 1, 2, ...) for device preference, -1 for CPU/host,
+        or None to let the driver decide.
+        (Default to None)
```
kkraus14, line 29: I know passing `-1` for CPU is how the underlying CUDA API handles things, but it doesn't feel particularly Pythonic. Can be addressed in a follow up.

cuda_core/cuda/core/experimental/_memory/_memory_pool.pyx
```diff
@@ -0,0 +1,438 @@
...
+        return self._peer_accessible_by
+
+    @peer_accessible_by.setter
+    def peer_accessible_by(self, devices):
+        """Set which devices can access this memory pool."""
+        from .._device import Device
+
+        # Convert all devices to device IDs
+        cdef set[int] target_ids = {Device(dev).device_id for dev in devices}
+        target_ids.discard(self._dev_id)  # exclude this device from peer access list
```
kkraus14, line 243: nitpick, can make this a smidge more efficient and still readable using a walrus operator:

```suggestion
        cdef set[int] target_ids = {device_id for dev in devices if (device_id := Device(dev).device_id != self._dev_id}
```

cuda_core/cuda/core/experimental/_memory/_memory_pool.pyx
```diff
@@ -0,0 +1,438 @@
...
+            raise ValueError(f"Device {self._dev_id} cannot access peer(s): {', '.join(map(str, bad))}")
+        cdef set[int] cur_ids = set(self._peer_accessible_by)
+        cdef set[int] to_add = target_ids - cur_ids
+        cdef set[int] to_rm = cur_ids - target_ids
+        cdef size_t count = len(to_add) + len(to_rm) # transaction size
+        cdef cydriver.CUmemAccessDesc* access_desc = NULL
+        cdef size_t i = 0
+
+        if count > 0:
+            access_desc = <cydriver.CUmemAccessDesc*>PyMem_Malloc(count * sizeof(cydriver.CUmemAccessDesc))
```
kkraus14, line 256: Is there any way we could use something like a `std::vector` here or something else that has proper RAII semantics instead of having to manually allocate and free? I.E.
```
access_desc = cdef vector[CUmemAccessDesc] access_descs
access_desc.reserve(count)
```

It looks like you have things handled here in the error cases, but it always makes me wary of doing manual memory management if we don't have to.

cuda_core/cuda/core/experimental/_memory/_memory_pool.pyx
```diff
@@ -0,0 +1,438 @@
...
+            properties.maxSize = opts._max_size
+    ELSE:
+        properties.maxSize = opts._max_size
+
+    self._dev_id = dev_id
+    self._mempool_owned = True
+
+    with nogil:
+        HANDLE_RETURN(cydriver.cuMemPoolCreate(&(self._handle), &properties))
+        # TODO: should we also set the threshold here?
```
kkraus14, line 373: Just to confirm, we're currently setting the threshold higher up still, correct?

cuda_core/cuda/core/experimental/_memory/_memory_pool.pyx
```diff
@@ -0,0 +1,438 @@
...
+    buf._ptr_obj = None
+    buf._size = size
+    buf._memory_resource = self
+    buf._alloc_stream = stream
+    return buf
+
+
+cdef inline void _MP_deallocate(
+    _MemPool self, uintptr_t ptr, size_t size, Stream stream
+) noexcept:
```
kkraus14, line 409: Can we mark this function as `nogil`? I don't see anywhere needing the GIL here?

cuda_core/cuda/core/experimental/_memory/_pinned_memory_resource.pyx
```diff
@@ -0,0 +1,274 @@
...
+        return
+
+    if platform.system() != "Linux":
+        return
+
+    numa_count = None
+
+    # Try /sys filesystem first (most reliable and doesn't require external tools)
+    try:
+        import os
```
kkraus14, line 44: nitpick: any reason for this not to be a top level import?

cuda_core/cuda/core/experimental/_memory/_pinned_memory_resource.pyx
```diff
@@ -0,0 +1,274 @@
...
+from cuda.core.experimental._utils.cuda_utils import check_multiprocessing_start_method
+
+
+# Cache to ensure NUMA warning is only raised once per process
+cdef bint _numa_warning_shown = False
+
+
+def _check_numa_nodes():
+    """Check if system has multiple NUMA nodes and warn if so."""
+    global _numa_warning_shown
```
kkraus14, line 33: Not a big deal since it's just a warning, but should we have some kind of lock around this to prevent multi-threading from possibly warning multiple times?

kkraus14 (approved):
Looks great! A few nitpicks / comments, but no blockers where they can all be addressed later if desired.

cuda_core/tests/memory_ipc/test_serialize.py
Andy-Jost: FYI, I have a change implementing resource handles on hold until after the release. It fixes this triple-free issue by adding reference counting outside of the driver. From what I'm seeing, it should significantly improve or eliminate the appearance of CUDA driver errors at shutdown.

cuda_core/cuda/core/experimental/_memory/_managed_memory_resource.pyx
```diff
@@ -0,0 +1,119 @@
...
+        """Whether this memory resource has IPC enabled."""
+        return self._ipc_data is not None
+
+    @property
+    def is_mapped(self) -> bool:
+        """
+        Whether this is a mapping of an IPC-enabled memory resource from
+        another process.  If True, allocation is not permitted.
+        """
+        return self._ipc_data is not None and self._ipc_data._is_mapped
```
Andy-Jost, line 119: Should these be provided by the `_MemPool` base?

cuda_core/tests/memory_ipc/test_serialize.py
Andy-Jost: I wonder if we should find a (future) workaround that permits duplicate mappings both for DMR and PMR. The driver team already confirmed that the DMR behavior is buggy. It seems like a very annoying restriction.

cuda_core/tests/memory_ipc/test_event_ipc.py
```diff
@@ -22,7 +22,12 @@ def test_main(self, ipc_device, ipc_memory_resource):
         device = ipc_device
         mr = ipc_memory_resource
         stream1 = device.create_stream()
-        latch = LatchKernel(device)
+        # TODO: We pick a timeout here to ensure forward progress (it needs to be
+        # less than CHILD_TIMEOUT_SEC) when a pinned memory resource is in use,
+        # in which case the call to buffer.copy_from(...) below is a synchronous
+        # operation that blocks the host. But calling the latch kernel here does not
+        # make any sense. We should refactor this test.
+        latch = LatchKernel(device, timeout_sec=5)
```
Andy-Jost, line 30: Would it make more sense to skip this test in the case of PMR?

Andy-Jost (approved):
LGTM

cuda_core/cuda/core/experimental/_memory/_managed_memory_resource.pyx
leofang: Let me put something together tomorrow when addressing https://github.com/NVIDIA/cuda-python/issues/1359.

cuda_core/tests/memory_ipc/test_event_ipc.py
leofang: Let us address this in a follow-up PR. I am thinking this ~~kernel~~ **line** should be dropped, since all it does is launching a kernel, but ~~there is no synchronization/communication between host and device~~ **it is not busy-waiting for anything**. The alternative is to replace it by a dummy kernel.

cuda_core/cuda/core/experimental/_memory/_memory_pool.pyx
leofang: Will address this in a follow-up PR. `_memory_pool.pyx` in this PR just a copy-pasta, not new code.

cuda_core/cuda/core/experimental/_memory/_managed_memory_resource.pyx
leofang: Right, though I don't have good ideas either. I don't think we want to use Python enums, so our options are limited.

cuda_core/cuda/core/experimental/_memory/_memory_pool.pyx
leofang: Will address this in a follow-up PR. `_memory_pool.pyx` in this PR just a copy-pasta, not new code.

cuda_core/cuda/core/experimental/_memory/_memory_pool.pyx
leofang: No. Currently (before/after this PR), we only crank up the threshold when getting the current mempool, not when creating a new mempool. 

I think this is an oversight but I will address this in a follow-up PR. `_memory_pool.pyx` in this PR just a copy-pasta, not new code. Also, for new mempool instances I am not sure if setting threshold to the maximum is the safest thing to do. I'll play around with it but I suspect this won't be done by ~tomorrow.

cuda_core/cuda/core/experimental/_memory/_pinned_memory_resource.pyx
leofang: This is my oversight.

leofang:
Let me merge and follow up.

cuda_core/cuda/core/experimental/_memory/_memory_pool.pyx
leofang: It turns out that this is an optimal implementation. We creates the C args from the Python objects at the call site, so even if we mark this as nogil, we can't possibly relax the gil at the call site anyway.

cuda_core/cuda/core/experimental/_memory/_pinned_memory_resource.pyx
leofang: Addressed in #1388.

leofang: In general, `cuda.core` has a few internal, module-level states that should be guarded by a lock. We haven't done that extensively. IIRC there should be an issue tracking the work for thread-safety, if not I'll create one.

cuda_core/cuda/core/experimental/_memory/_managed_memory_resource.pyx
leofang: Addressed in #1388. This is a good catch. I moved `is_ipc_enabled`, `is_mapped`, and `uuid` to `_MemPool`.

cuda_core/cuda/core/experimental/_memory/_pinned_memory_resource.pyx
leofang: Addressed in https://github.com/NVIDIA/cuda-python/pull/1388.

cuda_core/cuda/core/experimental/_memory/_pinned_memory_resource.pyx
leofang: > IIRC there should be an issue tracking the work for thread-safety, if not I'll create one.

https://github.com/NVIDIA/cuda-python/issues/1389

PR #1365, opened 2025-12-11
[cuda_pathfinder] Add CUDA 12 support to cuda-pathfinder pixi environments and removing build-variants

rparolin:
- Add new `cu12` pixi environment for CUDA 12 testing
- Set `noarch = "python"` in package configuration

### DIFF
--- a/cuda_pathfinder/pixi.lock
+++ b/cuda_pathfinder/pixi.lock
@@ -3,6 +3,8 @@ environments:
   cu12:
     channels:
     - url: https://conda.anaconda.org/conda-forge/
+    options:
+      pypi-prerelease-mode: if-necessary-or-explicit
     packages:
       linux-64:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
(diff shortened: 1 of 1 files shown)


cuda_pathfinder/pixi.toml
cpcloud: Since you're removing build variants I think you need a lower bound on the Python version.

```suggestion
python = ">=3.10"
setuptools = ">=64"

[package.run-dependencies]
python = ">=3.10"
```

cpcloud (changes requested):
Nothing major, but we should probably put a lower bound on the Python version.

cuda_pathfinder/pixi.toml
cpcloud: I'll just accept this for now, so that we can get it merged.

cuda_pathfinder/pixi.toml
```diff
@@ -39,7 +43,11 @@ python = "*"
 setuptools = ">=64"
 
 [package.run-dependencies]
-python = "*"
+python = ">=3.10"
+setuptools = ">=64"
+
+[package.run-dependencies]
+python = ">=3.10"
```
cpcloud: These are duplicated, which is a bug. I'll add a suggestion here to fix it.

cpcloud:
```suggestion
```

cpcloud (approved):

kkraus14:
/ok to test [aa681fd](https://github.com/NVIDIA/cuda-python/pull/1365/commits/aa681fd943a62163899fbb022f2347b391229f57)

PR #1366, opened 2025-12-11
Simplify Buffer.fill() API

### LINKED ISSUE #1345: Revisit `Buffer.fill()`

Andy-Jost:
## Summary

- Remove the explicit `width` parameter from `Buffer.fill()` — width is now inferred from the value
- Accept `int` in `[0, 256)` for 1-byte fills, or `collections.abc.Buffer` objects (1, 2, or 4 bytes) for multi-byte patterns
- Enables cleaner usage: `buffer.fill(0, stream=s)` or `buffer.fill(np.float32(1.0), stream=s)`

Closes #1345

## Test Plan

- [x] Updated `test_buffer_fill` with new API and additional test cases for numpy types and raw bytes
- [x] Updated `test_graph_mem.py` fill tests
- [x] All fill-related tests pass locally


Andy-Jost:
This is ready to go AFAIK

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
```diff
@@ -203,58 +209,75 @@ cdef class Buffer:
...
+        """Fill this buffer with a repeating byte pattern.
 
         Parameters
         ----------
-        value : int
-            Integer value to fill the buffer with
-        width : int
-            Width in bytes for each element (must be 1, 2, or 4)
+        value : int | :obj:`collections.abc.Buffer`
+            - int: Must be in range [0, 256). Converted to 1 byte.
```
leofang, line 205: Forgot to ask yesterday:
1. if we can copy up to 4 bytes, why do we restrict ourselves to 1 byte only?
2. why is the value required to be unsigned?

leofang: ^^^^^ @Andy-Jost would be nice to get an answer on this before merging

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
```diff
@@ -203,58 +209,75 @@ cdef class Buffer:
...
+            # int.to_bytes raises OverflowError if not in [0, 256)
+            pattern = value.to_bytes(1, "little")
+        else:
+            try:
+                mv = memoryview(value)
+            except TypeError:
+                raise TypeError(
+                    f"value must be an int or support the buffer protocol, got {type(value).__name__}"
+                ) from None
+            pattern = mv.tobytes()
```
kkraus14: Instead of doing these via Python functions and getting intermediate Python objects, can we use Cython built-ins / CPython APIs?

For `int.to_bytes` we can use `_PyLong_AsByteArray` (Python < 3.13) and `PyLong_AsNativeBytes` (Python >=3.13).

For the `memoryview` we can use Cython typed memoryviews (https://cython.readthedocs.io/en/latest/src/userguide/memoryviews.html)?

Both of these should be considerably more efficient than using the Python functions?

kkraus14: This causes a deepcopy of the memory since `bytes` is purely owning object.

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
Andy-Jost: For (1), the goal is to keep a simple mental model: an `int` is a single byte. If `buf.fill(0xFF)` were interpreted as 4 bytes, the behavior could be silently surprising, and we'd run into questions about interleaved zeros and endianness. In contrast, if we keep this as one byte, then an incorrect expression such as `buf.fill(0xFFFFFFFF)` raises, which is helpful because the noisy failure can be addressed.

For (2), allowing something like `[-128, 256)` is unusual and difficult to reason about. Bytes are almost always understood as `[0, 256)`, so mixing in "some" negative values raises questions and makes the contract harder to reason about. This matches the behavior of `int.to_bytes`.

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
Andy-Jost: I'll look into that today.

Andy-Jost: Good catch. I think `mv.tobytes()` can be removed, since the `memoryview` ought to work with `len` and `int.from_bytes`.

Andy-Jost: Ok, I reworked this section. Let me know what you think.

I was able to avoid `int.to_bytes` by letting Cython do the integer conversion after checking its range.

For the `memoryview` conversion, the new code avoids the copy via `tobytes` altogether, unless `memoryview.cast` fails (possible for non-contiguous ranges).

This also fixes an important failure path: large buffers were copied before their length was validated, so in the worst case, a buffer of any size might be copied before generating an error. That's fixed now.

I did work up a version using Cython typed memoryview, but it added complexity and did not deliver any performance gains I could observe. I think that compared to the other costs here (buffer acquisition and CUDA API call), the gains are negligible at best. LMK if you want to reconsider.

For reference, here's a sketch of that:
```
cdef size_t width
cdef unsigned int int_value
cdef const unsigned char[::1] b  # 1-D, contiguous bytes

if isinstance(value, int):
...
else:
    # Typed-memoryview acquisition
    b = value

    width = b.shape[0]
    if width not in (1, 2, 4):
        raise ValueError(f"value must be 1, 2, or 4 bytes, got {width}")

    if width == 1:
        int_value = b[0]
    elif width == 2:
        int_value = (<unsigned int>b[0]) | ((<unsigned int>b[1]) << 8)
    else:  # width == 4
        int_value = (<unsigned int>b[0]) | ((<unsigned int>b[1]) << 8) | ((<unsigned int>b[2]) << 16) | ((<unsigned int>b[3]) << 24)
```

Andy-Jost: Removed `mv.tobytes`

kkraus14: I was hoping that the general flow would look like:
* Given a Python integer object --> use a Cython function that takes `uint8_t` (will automatically throw Overflow error if not in range) --> `reinterpret_cast` to `unsigned char` (the type expected by the CUDA API)
* (Optional) Given a Python bytes object --> get void* (via `PyBytes_AsString`) --> `reinterpret_cast` / `memcpy` (assuming it gets optimized out) to `int{8,16,32}_t` (the type expected by the CUDA API)
* Given a buffer protocol compatible object --> get void* via buffer protocol --> `reinterpret_cast` / `memcpy` (assuming it gets optimized out) to `int{8,16,32}_t` (the type expected by the CUDA API)

There's a lot of intermediate object creation in the current implementation that I suspect adds non-trivial overhead.

kkraus14: @Andy-Jost can you create a follow-up issue since we seemingly auto-merged before concluding this discussion?

Andy-Jost: https://github.com/NVIDIA/cuda-python/issues/1375

PR #1367, opened 2025-12-11
Support H2D/D2D/D2H SMV copy

stiepan:
## Description

This PR implements copy_from/copy_to methods for StridedMemoryView.
* For D2D we either run memcopyasync or launch a simple cuda kernel to elementwise copy between the views.
   * Layouts are first permuted so that dst is as close as to C-stride as possible.
   * Layouts are squeezed, flattened and vecotrized if possible.
   * Each specialization is compiled once per arch and stored in a in-process cache.
* For H2D/D2H, ideally we just run memcopyasync. If necessary, one or both H2H and D2D copies are run to coalesce, transpose, or scatter the data. The H2H copy is run with numpy.
* If extra copies are needed and user did not specified allocators, we allocate: for device - from Device().memory_resource, for host - simply creating numpy array. Otherwise we use specified allocator, for host, the memory is ten wrapped into a numpy array.

Time-permittion or as a follow-up:
* Add examples
* Migrate tiled kernel from nvmath to speed-up transposing D2D copies.
* When available (CTK 13), use default pinned memory pool instead of numpy as default host allocator (when https://github.com/NVIDIA/cuda-python/pull/1364 is merged)

closes 

## Checklist

### DIFF
--- a/cuda_core/cuda/core/experimental/_memoryview.pyx
+++ b/cuda_core/cuda/core/experimental/_memoryview.pyx
@@ -301,6 +301,10 @@ cdef class StridedMemoryView:
         return self.get_layout()
 
     @property
+    def size(self) -> int:
+        return self.get_layout().get_volume()
+
+    @property
     def shape(self) -> tuple[int, ...]:
         """
         Shape of the tensor.
--- a/cuda_bindings/cuda/bindings/_nvml.pxd
+++ b/cuda_bindings/cuda/bindings/_nvml.pxd
@@ -32,7 +32,6 @@ ctypedef nvmlProcessInfo_v2_t ProcessInfo_v2
 ctypedef nvmlNvLinkUtilizationControl_t NvLinkUtilizationControl
 ctypedef nvmlViolationTime_t ViolationTime
 ctypedef nvmlUUIDValue_t UUIDValue
-ctypedef nvmlPlatformInfo_v1_t PlatformInfo_v1
 ctypedef nvmlVgpuPlacementList_v1_t VgpuPlacementList_v1
 ctypedef nvmlNvLinkPowerThres_t NvLinkPowerThres
 ctypedef nvmlSystemEventData_v1_t SystemEventData_v1
@@ -40,11 +39,9 @@ ctypedef nvmlGpuInstanceProfileInfo_t GpuInstanceProfileInfo
 ctypedef nvmlComputeInstanceProfileInfo_t ComputeInstanceProfileInfo
 ctypedef nvmlMask255_t Mask255
 ctypedef nvmlHostname_v1_t Hostname_v1
-ctypedef nvmlNvLinkInfo_v1_t NvLinkInfo_v1
 ctypedef nvmlPowerValue_v2_t PowerValue_v2
 ctypedef nvmlVgpuProcessUtilizationSample_t VgpuProcessUtilizationSample
 ctypedef nvmlGpuFabricInfo_t GpuFabricInfo
-ctypedef nvmlGpuFabricInfo_v2_t GpuFabricInfo_v2
 ctypedef nvmlSystemEventSetCreateRequest_v1_t SystemEventSetCreateRequest_v1
 ctypedef nvmlSystemEventSetFreeRequest_v1_t SystemEventSetFreeRequest_v1
 ctypedef nvmlSystemRegisterEventRequest_v1_t SystemRegisterEventRequest_v1
@@ -266,7 +263,6 @@ cpdef unsigned int device_get_pcie_link_max_speed(intptr_t device) except? 0
 cpdef unsigned int device_get_pcie_speed(intptr_t device) except? 0
 cpdef unsigned int device_get_adaptive_clock_info_status(intptr_t device) except? 0
 cpdef unsigned int device_get_bus_type(intptr_t device) except? 0
-cpdef object device_get_gpu_fabric_info_v(intptr_t device)
 cpdef object system_get_conf_compute_capabilities()
 cpdef object system_get_conf_compute_state()
 cpdef object device_get_conf_compute_mem_size_info(intptr_t device)
@@ -293,7 +289,6 @@ cpdef object device_get_row_remapper_histogram(intptr_t device)
 cpdef unsigned int device_get_architecture(intptr_t device) except? 0
 cpdef object device_get_clk_mon_status(intptr_t device)
 cpdef object device_get_process_utilization(intptr_t device, unsigned long long last_seen_time_stamp)
-cpdef object device_get_platform_info(intptr_t device)
 cpdef unit_set_led_state(intptr_t unit, int color)
 cpdef device_set_persistence_mode(intptr_t device, int mode)
 cpdef device_set_compute_mode(intptr_t device, int mode)
@@ -445,7 +440,6 @@ cpdef object device_get_repair_status(intptr_t device)
 cpdef object device_get_power_mizer_mode_v1(intptr_t device)
 cpdef device_set_power_mizer_mode_v1(intptr_t device, intptr_t power_mizer_mode)
 cpdef object device_get_pdi(intptr_t device)
-cpdef object device_get_nvlink_info(intptr_t device)
 cpdef device_read_write_prm_v1(intptr_t device, intptr_t buffer)
 cpdef object device_get_gpu_instance_profile_info_by_id_v(intptr_t device, unsigned int profile_id)
 cpdef object device_get_unrepairable_memory_flag_v1(intptr_t device)
--- a/cuda_bindings/cuda/bindings/_nvml.pyx
+++ b/cuda_bindings/cuda/bindings/_nvml.pyx
@@ -13,6 +13,8 @@ from ._internal.utils cimport (get_buffer_pointer, get_nested_resource_ptr,
 
 from enum import IntEnum as _IntEnum
 
+from cuda.bindings.cydriver cimport CUDA_VERSION
+
 
 from libc.stdlib cimport calloc, free, malloc
 from cython cimport view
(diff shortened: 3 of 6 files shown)


cuda_core/pyproject.toml
```diff
@@ -69,7 +69,7 @@ issues = "https://github.com/NVIDIA/cuda-python/issues/"
 include = ["cuda.core*"]
 
 [tool.setuptools.package-data]
-"cuda.core.experimental.include" = ["*.h", "*.hpp", "*.cuh"]
+"cuda.core.experimental.include" = ["*.h", "*.hpp", "*.cuh", "strided_copy/*.h", "strided_copy/impl/*.h"]
```
leofang: nit: I think this should work (untested)
```suggestion
"cuda.core.experimental.include" = ["**/*.h", "**/*.hpp", "**/*.cuh"]
```

stiepan: Nice, much more concise! Seems to work fine.

cuda_core/cuda/core/experimental/_strided_copy/_jit.pyx
```diff
@@ -0,0 +1,105 @@
...
+
+
+cpdef str get_strided_copy_include_dir(object logger):
+    """
+    Finds and caches the absolute path for the strided copy includes.
+    """
+    # TODO(ktokarski) Once Program API supports passing includes as strings and names,
+    # read all the headers once and cache them.
+    if hasattr(_tls, "strided_copy_include_dir"):
+        return _tls.strided_copy_include_dir
```
cpcloud:
```suggestion
    if (strided_copy_include_dir := getattr(_tls, "strided_copy_include_dir", None)) is not None:
        return strided_copy_include_dir
```

This is one attribute access versus two (one inside `hasattr`, and then the attribute access itself).

Feel free to not use the `:=` walrus operator if you don't like it.

cuda_core/cuda/core/experimental/_strided_copy/_jit.pyx
```diff
@@ -0,0 +1,105 @@
...
+    cdef str current_dir = os.path.dirname(os.path.abspath(__file__))
+    cdef str copy_kernel_dir = os.path.normpath(os.path.join(current_dir, "..", "include", "strided_copy"))
```
cpcloud:
```suggestion
    cdef str copy_kernel_dir = os.path.normpath(os.path.join(current_dir, os.pardir, "include", "strided_copy"))
```

Probably not strictly necessary, but avoids a hardcoded string.

cuda_core/cuda/core/experimental/_strided_copy/_jit.pyx
```diff
@@ -0,0 +1,105 @@
...
+    cdef str arch = device_ccs.get(device_id)
+    if arch is None:
+        arch = f"sm_{Device(device_id).arch}"
+        device_ccs[device_id] = arch
+    return arch
+
+
+cdef compile_load_kernel(str kernel_code, str arch, object logger):
+    cdef str include_dir = get_strided_copy_include_dir(logger)
+    cdef options = ProgramOptions(std="c++17", arch=arch, include_path=include_dir)
```
cpcloud: Hardcoding `c++17' feels a bit brittle here, but I certainly won't block on that.

cuda_core/cuda/core/experimental/_layout.pxd
```diff
@@ -375,21 +393,28 @@ cdef inline stride_t *get_strides_ptr(BaseLayout& base) except? NULL nogil:
...
         return False
     for i in range(a.ndim):
         if a.shape[i] != b.shape[i]:
             return False
-    if a.strides != NULL or b.strides != NULL:
-        if a.strides == NULL or b.strides == NULL:
+    return True
+
+
+cdef inline bint _base_equal_strides(BaseLayout& a, BaseLayout& b) noexcept nogil:
```
cpcloud, line 405: Why is one underscore-prefixed (`_base_equal_strides`) and the other not (`base_equal_shapes`)?

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -235,8 +236,8 @@ cdef class StridedMemoryView:
         return view
 
     def copy_from(
-        self, other : StridedMemoryView, stream : Stream,
-        allocator = None,
+        self, other : StridedMemoryView, Stream stream,
+        allocator : CopyAllocatorOptions | dict[str, MemoryResource] | None = None,
```
cpcloud, line 240: I don't think the syntax variation matters here. Cython can handle both variations. Slight preference for leaving things as they are here.

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -702,3 +702,51 @@ cdef inline int view_buffer_strided(
     view._layout = layout
     view._dtype = dtype
     return 0
+
+
+cdef inline copy_into(
```
cpcloud: `inline`-ing here seems a bit pointless and since many of these would be transitively inline'd (because many of the methods/functions are themselves `inline`), the code will be bigger.

Do we know if `inline` actually helps anything here? My preference would be to leave `inline` off unless it can be demonstrated that it actually meaningfully helps performance and doesn't bloat shared objects. It's very low risk to leave it off.

cpcloud (commented):
Mostly just nits, slight preference to avoid so much inlining unless we can demonstrate its effectiveness in a realistic benchmark.

cuda_core/cuda/core/experimental/_layout.pxd
stiepan: It's because the shapes check the ndim equality first, while strides don't. I needed the shape equality for the copy, while strides not really. Maybe for completness it makes sense to add "public" version with ndim check too. :D

cuda_core/cuda/core/experimental/_memoryview.pyx
stiepan: Right! I checked what resulting cpp looks like and indeed, it seems that cython can figure out if the annotation is cimported (and use struct ptr instead of a generic PyObject*). On the other hand, this provides kind of extra compile-time check, as `Type arg` does not work with regular Python classes.

cuda_core/cuda/core/experimental/_memoryview.pyx
stiepan: Yeah, here it's probably just unnecessary. Leaving definitions in `layout.pxd` file for basic shape/strides utilities actually made a perf difference, but for a function definied in the same pyx file I would guess compiler doesn't care if the inline is there. Removed it.

cuda_core/cuda/core/experimental/_strided_copy/_jit.pyx
leofang: C++17 is the minimal starting CUDA 13, so I am fine with this.

stiepan: The cpp code actually uses some of c++17 (`if constexpr` for sure). I looked into ProgramOption docs, it says c++17 is a default, so I removed it. If the default is not going to drop, I guess we're fine.

cuda_core/cuda/core/experimental/_memoryview.pyx
leofang: Agreed with Kamil. If it's a cdef class, let's use the Cython-style typing, otherwise we do Python type annotation.

leofang:
/ok to test [c761437](https://github.com/NVIDIA/cuda-python/pull/1367/commits/c7614373322153c51a5319a1271b6b5ccbd5e1da)

stiepan:
>Error: Unexpected HTTP response: 504

Looks like some (transient?) connectivity issues

cuda_core/cuda/core/experimental/include/strided_copy/impl/array_view.h
```diff
@@ -0,0 +1,63 @@
...
+#else
+// Fallback for other compilers, or if restrict is not supported
+#define RESTRICT
+#endif
+
+namespace cuda_core
+{
+
+template <typename T, typename _coords_t>
+struct array_view
```
rparolin, line 27: I wouldn't overload the name `array_view` since the `std::array_view` was the historical name for `std::span`.   Given that this is implemented multi-dimensions I'd call it something like `mdspan` which would allow us to ultimately remove the implementation after the `std::mdspan` is usable in our code base.
https://en.cppreference.com/w/cpp/container/mdspan.html

cuda_core/cuda/core/experimental/include/strided_copy/impl/utils.h
```diff
@@ -0,0 +1,180 @@
...
+};
+
+template <class T, class F>
+struct conditional<false, T, F>
+{
+    using type = F;
+};
+
+template <bool B, class T = void>
+struct enable_if
```
rparolin, line 43: What is the benefit or reimplementing these core meta programming primitives?  https://en.cppreference.com/w/cpp/types/enable_if.html

cuda_core/cuda/core/experimental/include/strided_copy/impl/utils.h
```diff
@@ -0,0 +1,180 @@
...
+
+template <typename T>
+struct min_val
+{
+};
+
+template <>
+struct min_val<int32_t>
+{
+    static constexpr int32_t value = -2147483648;
```
rparolin, line 95: Why is std::numeric_limits not being used instead?
https://en.cppreference.com/w/cpp/types/numeric_limits.html

```
auto min_val = std::numeric_limits<int32_t>::min();
```

leofang:
As per https://github.com/NVIDIA/cuda-python/issues/674#issuecomment-3656821898 let us close this PR. Thanks @stiepan for all the works, and sorry that we decided to not accept it in cuda.core.

PR #1370, opened 2025-12-15
Revert `StridedLayout`

leofang:
## Description

As per https://github.com/NVIDIA/cuda-python/issues/674#issuecomment-3656821898 this PR reverts #1283, since we'll move `StridedLayout` to elsewhere. 

The commit from #1363 was reverted first so as to make the revert clean, but I added it back so that it's intact. The commit history should be clean enough to see how this is done.

## Checklist

### DIFF
--- a/cuda_core/cuda/core/experimental/_memoryview.pyx
+++ b/cuda_core/cuda/core/experimental/_memoryview.pyx
@@ -170,6 +170,14 @@ cdef class StridedMemoryView:
             dlm_tensor.deleter(dlm_tensor)
 
     @property
+    def size(self) -> int:
+        cdef int i
+        cdef size_t s = 1
+        for i in self.shape:
+            s *= i
+        return s
+
+    @property
     def shape(self) -> tuple[int, ...]:
         if self._shape is None:
             if self.exporting_obj is not None:
--- a/cuda_core/tests/test_utils.py
+++ b/cuda_core/tests/test_utils.py
@@ -90,6 +90,7 @@ class TestViewCPU:
         assert isinstance(view, StridedMemoryView)
         assert view.ptr == in_arr.ctypes.data
         assert view.shape == in_arr.shape
+        assert view.size == in_arr.size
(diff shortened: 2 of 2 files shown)


leofang:
/ok to test [412ba2c](https://github.com/NVIDIA/cuda-python/pull/1370/commits/412ba2cc2d08c7256871ee49fda03d92e882e715)

leofang:
cc @stiepan for vis

cuda_core/docs/source/api.rst
```diff
@@ -75,7 +75,6 @@ Utility functions
 
    args_viewable_as_strided_memory
 
-   :template: autosummary/cyclass.rst
+   :template: dataclass.rst
```
cpcloud, line 78: Didn't the original PR fix a docs issue?

stiepan: In the reverted state, no docs for methods are rendered in the SMV page. It could have been intentional, but, FYI currently it seems there is no mention of from_dlpack/from_cai that were added in the meantime.

stiepan:
> cc @stiepan for vis

It seems clean. Good that we moved dlpack fix out of this PR after all. 🥲 

@leofang, as @cpcloud pointed out, please double check the docs page.

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
stiepan, line 1: This part may be needed, if you're going to bring back from_buffer with explicit shape, strides, dtype. If the Buffer does not come from memory resource, but wraps external pointer (Buffer.from_handle), the changes here: 1. allowed to pass owner obj to keep that allocation alive, 2. inferred device_id and device/host accesibility.

FYI @cpcloud

cpcloud: I guess I'll just revert revert this part and change the signature in a follow-up PR.

leofang: This PR is now reverted in #1373  😅

PR #1372, opened 2025-12-15
test: avoid leaving files around during `ProgramOptions` testing

cpcloud:
Small test refactor to avoid dumping files in `$PWD` during `ProgramOptions` testing.

### DIFF
--- a/cuda_core/tests/test_program.py
+++ b/cuda_core/tests/test_program.py
@@ -232,13 +232,6 @@ def ptx_code_object():
         # ProgramOptions(pre_include="cuda_runtime.h"),
         ProgramOptions(no_cache=True),
         pytest.param(
-            ProgramOptions(fdevice_time_trace="trace.json"),
-            marks=pytest.mark.skipif(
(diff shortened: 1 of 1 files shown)


cuda_core/tests/test_program.py
```diff
@@ -308,6 +287,36 @@ def test_cpp_program_with_various_options(init_cuda, options):
     assert program.handle is None
 
 
+@pytest.mark.skipif(
+    (_get_nvrtc_version_for_tests() or 0) < 13000,
+    reason="buggy with NVRTC < 13.0 (File 'trace.json.json' could not be opened)",
```
kkraus14, line 292: nitpick but can we make this `trace.json` instead of `trace.json.json`? 😆

kkraus14 (approved):

cuda_core/tests/test_program.py
leofang: That is the bug that led to this skip condition. We set the filename to `trace.json` but NVRTC 12.9 insisted to look for `trace.json.json`.

PR #1373, opened 2025-12-15
refactor: privatize `StridedLayout` as `_StridedLayout`

cpcloud:
# Revert the revert of StridedLayout and privatize as `_StridedLayout`

## Summary

This PR reverts the removal of `StridedLayout` (reverting #1370) and
immediately privatizes it to `_StridedLayout` to unblock ongoing work while
avoiding premature API stabilization.

## Motivation

The original `StridedLayout` implementation was reverted in #1370, but this
blocked work that depends on the internal layout infrastructure. Rather than
keeping the code removed, we've chosen to:

1. **Restore the implementation** - Bring back all `StridedLayout` functionality
2. **Privatize the API** - Rename `StridedLayout` to `_StridedLayout` (making it internal-only)
3. **Remove from documentation** - Ensure no public API promises are made

This approach allows internal and experimental code to continue using the
layout functionality while clearly signaling that the API is not yet stable.

## Changes

- Reverted commit d8e931761 which removed `StridedLayout`
- Renamed `StridedLayout` to `_StridedLayout` throughout the codebase
- Removed `_StridedLayout` from public API documentation
- Updated all tests to use the privatized name

## Follow-up Work

An immediate follow-up PR will refactor `StridedMemoryView.from_buffer` to
remove the `_StridedLayout` parameter and replace it with explicit `stride` and
`shape` arguments. This will further reduce the API surface of the internal
layout implementation.

### DIFF
--- a/cuda_core/cuda/core/experimental/_memory/_buffer.pyx
+++ b/cuda_core/cuda/core/experimental/_memory/_buffer.pyx
@@ -5,7 +5,7 @@
 from __future__ import annotations
 
 cimport cython
-from libc.stdint cimport uintptr_t, int64_t, uint64_t
+from libc.stdint cimport uintptr_t
 
 from cuda.bindings cimport cydriver
 from cuda.core.experimental._memory._device_memory_resource cimport DeviceMemoryResource


cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -152,6 +158,48 @@ cdef class StridedMemoryView:
             return cls.from_dlpack(obj, stream_ptr)
         return cls.from_cuda_array_interface(obj, stream_ptr)
 
+    @classmethod
+    def from_buffer(
+        cls, buffer : Buffer, layout : _StridedLayout,
```
kkraus14, line 163: Happy to defer it to a follow up but having this constructor require a private class isn't ideal. Should we make this private as well?

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -169,71 +217,153 @@ cdef class StridedMemoryView:
             dlm_tensor = <DLManagedTensor*>data
             dlm_tensor.deleter(dlm_tensor)
 
+    def view(
```
kkraus14, line 220: +1

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -169,71 +217,153 @@ cdef class StridedMemoryView:
...
+    ):
+        """
+        Copies the data from this view into the ``other`` view.
+
+        For details, see :meth:`copy_from`.
+        """
+        raise NotImplementedError("Sorry, not supported: copy_to")
+
+    @property
+    def layout(self) -> _StridedLayout:
```
kkraus14, line 292: Should we make this private for now?

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -169,71 +217,153 @@ cdef class StridedMemoryView:
...
               + f"                  shape={self.shape},\n"
               + f"                  strides={self.strides},\n"
+              + f"                  itemsize={self.layout.itemsize},\n"
               + f"                  dtype={get_simple_repr(self.dtype)},\n"
               + f"                  device_id={self.device_id},\n"
               + f"                  is_device_accessible={self.is_device_accessible},\n"
               + f"                  readonly={self.readonly},\n"
               + f"                  exporting_obj={get_simple_repr(self.exporting_obj)})")
 
+    cdef inline _StridedLayout get_layout(self):
```
kkraus14, line 335: This is a Cython function, but should we make it private for now? We technically allow cimporting our StridedMemoryView cython class.

kkraus14 (approved):

cuda_core/cuda/core/experimental/_memoryview.pyx
cpcloud: I will address these comments in the `from_buffer` PR follow-up.

cuda_core/cuda/core/experimental/_memoryview.pyx
cpcloud: Definitely, this was a search and replace job, I will do some more precise surgery in a follow-up.

cuda_core/cuda/core/experimental/_memoryview.pyx
cpcloud: Will do in a follow-up.

cuda_core/cuda/core/experimental/_memoryview.pyx
leofang: > We technically allow cimporting our StridedMemoryView cython class.

Actually we don't. `cuda.core` currently has 0 Cython APIs.

kkraus14: Sorry, I was looking at Buffer instead of StridedMemoryView and saw the `buffer.pxd` file. @leofang is correct and we can ignore this.

leofang:
Let me admin-merge this PR. All other CI pipelines are green but one, which seems to be causing the main branch to fail as well. I've raised a thread internally.

PR #1376, opened 2025-12-16
Optimize Buffer.fill() to avoid intermediate object creation

### LINKED ISSUE #1375: [FEA]: Reduce Python object creation in `Buffer.fill`
See open question in https://github.com/NVIDIA/cuda-python/pull/1366.

Comment: https://github.com/NVIDIA/cuda-python/pull/1366#discussion_r2620860178

Adjust flow in `Buffer.fill` to reduce intermediate object creation

Andy-Jost:
## Summary

- Use Cython typed parameters and C buffer API to eliminate Python object creation overhead in `Buffer.fill()`
- int path: `uint8_t` parameter for automatic overflow checking
- bytes path: direct `char*` pointer access  
- general buffer path: `PyObject_GetBuffer` for direct `void*` access

Closes #1375

## Test Plan

- [x] All 99 `test_buffer_fill` tests pass

### DIFF
--- a/cuda_core/cuda/core/experimental/_memory/_buffer.pyx
+++ b/cuda_core/cuda/core/experimental/_memory/_buffer.pyx
@@ -4,6 +4,7 @@
 
 from __future__ import annotations
 
+cimport cython
 from libc.stdint cimport uint8_t, uint16_t, uint32_t, uintptr_t
 from cpython.buffer cimport PyObject_GetBuffer, PyBuffer_Release, Py_buffer, PyBUF_SIMPLE
 
@@ -25,6 +26,7 @@ else:
 
 from cuda.core.experimental._dlpack import DLDeviceType, make_py_capsule
 from cuda.core.experimental._utils.cuda_utils import driver
+from cuda.core.experimental._device import Device
 
 __all__ = ['Buffer', 'MemoryResource']
 
@@ -54,6 +56,8 @@ cdef class Buffer:
         self._ipc_data = None
         self._ptr_obj = None
         self._alloc_stream = None
+        self._owner = None
+        self._mem_attrs_inited = False
 
     def __init__(self, *args, **kwargs):
         raise RuntimeError("Buffer objects cannot be instantiated directly. "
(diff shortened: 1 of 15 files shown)


Andy-Jost:
@kkraus14 I think this implementation follows the approach you outlined.

kkraus14 (approved):

kkraus14:
LGTM, this was basically exactly what I had in mind except your implementation was more elegantly using Cython instead of jumping straight down to the CPython API 😄

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
```diff
@@ -419,6 +382,36 @@ cdef inline void Buffer_close(Buffer self, stream):
         self._alloc_stream = None
 
 
+cdef inline void Buffer_fill_uint8(Buffer self, uint8_t value, cydriver.CUstream s):
+    with nogil:
+        HANDLE_RETURN(cydriver.cuMemsetD8Async(<cydriver.CUdeviceptr>self._ptr, value, self._size, s))
```
leofang, line 387: nit: we want the exception clause `except*` since `HANDLE_RETURN` can raise, but we don't want Cython to warn, so with Cython 3 we general want to avoid using `void` as the return type and use this instead:
```suggestion
cdef inline int Buffer_fill_uint8(Buffer self, uint8_t value, cydriver.CUstream s) except?-1:
    with nogil:
        HANDLE_RETURN(cydriver.cuMemsetD8Async(<cydriver.CUdeviceptr>self._ptr, value, self._size, s))
    return 0
```

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
```diff
@@ -419,6 +382,36 @@ cdef inline void Buffer_close(Buffer self, stream):
...
+
+
+cdef inline void Buffer_fill_from_ptr(
+    Buffer self, const char* ptr, size_t width, cydriver.CUstream s
+) except *:
```
leofang, line 392: ditto
```suggestion
cdef inline int Buffer_fill_from_ptr(
    Buffer self, const char* ptr, size_t width, cydriver.CUstream s
) except?-1:
```

cuda_core/cuda/core/experimental/_memory/_buffer.pyx
```diff
@@ -419,6 +382,36 @@ cdef inline void Buffer_close(Buffer self, stream):
...
+            HANDLE_RETURN(cydriver.cuMemsetD16Async(
+                <cydriver.CUdeviceptr>self._ptr, (<uint16_t*>ptr)[0], buffer_size // 2, s))
+    elif width == 4:
+        if buffer_size & 0x3:
+            raise ValueError(f"buffer size ({buffer_size}) must be divisible by 4")
+        with nogil:
+            HANDLE_RETURN(cydriver.cuMemsetD32Async(
+                <cydriver.CUdeviceptr>self._ptr, (<uint32_t*>ptr)[0], buffer_size // 4, s))
+    else:
+        raise ValueError(f"value must be 1, 2, or 4 bytes, got {width}")
```
leofang, line 412:
```suggestion
        raise ValueError(f"value must be 1, 2, or 4 bytes, got {width}")
    return 0
```

PR #1380, opened 2025-12-16
refactor: update `StridedMemoryView.from_buffer` to accept tuples and dtype instead of now-private `_StridedLayout`

cpcloud:
Change the definition of `StridedMemoryView.from_buffer` to accept shape, strides, and order to match the functionality of `_StridedLayout` without exposing it in the public API.

### DIFF
--- a/cuda_core/cuda/core/experimental/_layout.pxd
+++ b/cuda_core/cuda/core/experimental/_layout.pxd
@@ -297,9 +297,7 @@ cdef class _StridedLayout:
             raise ValueError(
                 f"Allocation size for a layout that maps elements "
                 f"to negative memory offsets is ambiguous. "
-                f"The layout's min_offset is {min_offset}. "
-                f"To create a supported layout with the same shape "
-                f"please use _StridedLayout.to_dense()."
+                f"The layout's min_offset is {min_offset}."
             )
         if max_offset < min_offset:
             return 0
--- a/cuda_core/tests/test_utils.py
+++ b/cuda_core/tests/test_utils.py
@@ -312,7 +312,7 @@ def test_from_buffer_disallowed_negative_offset():
     d = Device()
     d.set_current()
     buffer = d.memory_resource.allocate(20)
-    with pytest.raises(ValueError, match="please use _StridedLayout.to_dense()."):
+    with pytest.raises(ValueError):
         StridedMemoryView.from_buffer(
             buffer,
             shape=layout.shape,


cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -160,8 +160,13 @@ cdef class StridedMemoryView:
...
     @classmethod
     def from_buffer(
-        cls, buffer : Buffer, layout : _StridedLayout,
-        dtype : numpy.dtype | None = None,
+        cls,
+        buffer : Buffer,
+        shape : tuple[int, ...],
+        strides : tuple[int, ...] | None = None,
+        *,
+        dtype : numpy.dtype,
```
leofang: btw Kamil made `dtype` optional because there are types that are not currently representable by NumPy's built-in dtypes (ex: bf16). Would it be possible to preserve it?

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -160,8 +160,13 @@ cdef class StridedMemoryView:
...
+        order : str = "C",
```
leofang: If we accept `strides` we should not accept `order`, otherwise we'd have to worry about consistency checks (ex: checking if `strides` matches `order`).

cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -185,19 +190,32 @@ cdef class StridedMemoryView:
...
             Whether the mark the view as readonly.
         """
         cdef StridedMemoryView view = StridedMemoryView.__new__(cls)
-        view_buffer_strided(view, buffer, layout, dtype, is_readonly)
+        cdef int64_t itemsize = dtype.itemsize
+        view_buffer_strided(
+            view,
+            buffer,
+            _StridedLayout.dense(shape=shape, itemsize=itemsize, stride_order=order) if strides is None
+            else _StridedLayout(shape=shape, strides=strides, itemsize=itemsize),
```
leofang: ditto, here we do not check if `order` and `strides` are both being given but inconsistent, the simplest (and most general) solution is to just accept `strides`

cuda_core/cuda/core/experimental/_memoryview.pyx
cpcloud: Only if we also add `itemsize` to the signature.

cuda_core/cuda/core/experimental/_memoryview.pyx
cpcloud: Ok, that will lead to a bunch of test changes.

cuda_core/cuda/core/experimental/_memoryview.pyx
cpcloud: Fixed.

cuda_core/cuda/core/experimental/_memoryview.pyx
cpcloud: Not as a bad as I thought it would be!

cuda_core/cuda/core/experimental/_memoryview.pyx
cpcloud: Fixed.

leofang:
LGTM! One question: There are still a few references to `_StridedLayout` in the docs:
https://nvidia.github.io/cuda-python/pr-preview/pr-1380/cuda-core/latest/search.html?q=_StridedLayout
What to do with them? 

I think we should at least comment out `SMV.layout` and make it private.

cuda_core/cuda/core/experimental/_memoryview.pyx
leofang: Looks great! Thanks, Phillip.

cpcloud:
Let me see if making `layout` private is going to be a huge effort.

leofang:
yeah if it's too much churn then let's just raise a warning if users try to access it

cpcloud:
I removed I think all public references to `StridedLayout` as well as its mentions in various error messages.

Issue #1389, opened 2025-12-16
Address known thread-safety issues in `cuda.core`


leofang:
> In general, `cuda.core` has a few internal, module-level states that should be guarded by a lock. We haven't done that extensively. 

_Originally posted by @leofang in https://github.com/NVIDIA/cuda-python/pull/1364#discussion_r2624683212_

leofang:
Example:
https://github.com/NVIDIA/cuda-python/blob/027ba105fc33ac148c6d7343a9f71801e9f5ff72/cuda_core/cuda/core/experimental/_launcher.pyx#L20-L35

leofang:
Sebastian's comment here is relevant: https://github.com/NVIDIA/cuda-python/pull/1390#discussion_r2626380632

seberg:
I did an audit before for this type of issues but will have another sweep soon. Similar code still exists around `_decide_nvjitlink_or_driver` (and `_lazy_init` in `_linker.pyx`), but only a minor issue.

I.e. this doesn't use `cdef ... _inited` so is no additional free-threading issue.  The function may get called twice but looks idempotent _except_ for issuing a warning which is a nuisance.

That said: I'll look into establishing a `call-once` pattern which wouldn't just fix the nuisance but hopefully also make this clean (or at least obviously so).

PR #1388, opened 2025-12-16
Restore v0.3.x `MemoryResource` behaviors + MR minor follow-ups

leofang:
## Description

closes #1359

This PR attempts to partially restore the v0.3.x behavior, where initializing a `MemoryResource` instance would raise `TypeError`. We cannot actually do this because it is now a `cdef class`, so we do lose some protection, but at least we can raise the same exception when attempting to access the 5 methods/attributes.

This PR also addresses a few review comments from #1364.

Below is LLM-generated PR description: This pull request refactors the memory resource interface and its implementations to improve error handling and code clarity, while also making minor improvements to thread safety and test coverage. The most important changes are summarized below.

### MemoryResource interface changes

* Removed use of the `abc` module and replaced abstract methods in the `MemoryResource` class with concrete methods that raise `TypeError` if not implemented by subclasses, ensuring clearer error messages when methods are not overridden. [[1]](diffhunk://#diff-0c575d585121f3088d1a460ea98deef018725b103c6957a3fd816f29cae5595eL19) [[2]](diffhunk://#diff-0c575d585121f3088d1a460ea98deef018725b103c6957a3fd816f29cae5595eL495) [[3]](diffhunk://#diff-0c575d585121f3088d1a460ea98deef018725b103c6957a3fd816f29cae5595eL514-L516) [[4]](diffhunk://#diff-0c575d585121f3088d1a460ea98deef018725b103c6957a3fd816f29cae5595eL531-R543)
* Added property stubs (`is_device_accessible`, `is_host_accessible`, `device_id`) to `MemoryResource`, each raising `TypeError` to enforce implementation in subclasses.

### Memory resource implementation cleanup

[...]

Issue #980, opened 2025-09-17
[ENH]: Robust imports of optional dependencies (nvvm, nvJitLink)


rwgk:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

cuda.core

### Is your feature request related to a problem? Please describe.

xref: https://github.com/NVIDIA/cuda-python/pull/907#discussion_r2356628633

Copy-pasting for easy reference:

When we do it, we should:
- apply the suggestion you have above for NVVM
- apply the same approach to nvJitLink: https://github.com/NVIDIA/cuda-python/blob/7a24bd80b908e42f45ba610445998af383b2bbd3/cuda_core/cuda/core/experimental/_linker.py#L32
- ensure cuda-pathfinder is a dependency of cuda-core (today we get it indirectly through cuda-bindings)

### Describe the solution you'd like

Based on POC code posted here: https://github.com/NVIDIA/cuda-python/pull/907#discussion_r2356514889

brandon-b-miller:
Could we make this a utility function that's public? Numba-cuda [checks](https://github.com/NVIDIA/numba-cuda/blob/fc8b05025ced7375ea106a79d4c44dfcda7c25c5/numba_cuda/numba/cuda/cudadrv/driver.py#L121) for nvjitlink's availability to determine if LTO should be enabled in various places, currently through a workaround pulled from cuda-python. It'd be useful if we could check it through `cuda-python`/pathfinder.

rwgk:
> Could we make this a utility function that's public? Numba-cuda [checks](https://github.com/NVIDIA/numba-cuda/blob/fc8b05025ced7375ea106a79d4c44dfcda7c25c5/numba_cuda/numba/cuda/cudadrv/driver.py#L121) for nvjitlink's availability to determine if LTO should be enabled in various places, currently through a workaround pulled from cuda-python. It'd be useful if we could check it through `cuda-python`/pathfinder.

I labeled this issue additionally with cuda.pathfinder. I like the idea. It expands the scope of pathfinder somewhat, but I think it's close enough to the original purpose and doesn't introduce additional dependencies.

leofang:
Why is a new API needed? What's wrong with the existing pathfinder loading API?

rwgk:
[...]

PR #1390, opened 2025-12-16
Cythonize `launch` & `LaunchConfig` more

### LINKED ISSUE #1078: Cythonize `_launcher.pyx` more
@emcastillo might also have bw for this one?

leofang:
## Description

closes #1078

this PR:
```python
In [1]: %timeit launch(s, config, kernel, 0, 0)
2.6 μs ± 0.656 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

main branch (commit 027ba105fc)
```python
In [1]: %timeit launch(s, config, kernel, 0, 0)
4.02 μs ± 5.8 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

cuda.core v0.4.2 (**regression**):
```python
In [1]: %timeit launch(s, config, kernel, 0, 0)
4.42 μs ± 6.69 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

cuda.core v0.3.2:
```python
In [1]: %timeit launch(s, config, kernel, 0, 0)
4.18 μs ± 4.13 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

script:
```python
from cuda.core.experimental import Device, Program, launch, LaunchConfig


dev = Device()
dev.set_current()


prog = Program(
    r"""
    extern "C" __global__ void empty_kernel(const float* a, const double* b) {
    }
    """, code_type="c++"
)
obj = prog.compile(target_type="cubin")
kernel = obj.get_kernel("empty_kernel")

config = LaunchConfig(grid=64, block=256, shmem_size=0)
s = dev.default_stream
launch(s, config, kernel, 0, 0)
dev.sync()
```
Execute it via `ipython -i launch_perf.py` to warm up and then repeat the launch via `%timeit launch(s, config, kernel, 0, 0)`. 

## Checklist

### DIFF
--- a/cuda_core/cuda/core/experimental/_launch_config.pyx
+++ b/cuda_core/cuda/core/experimental/_launch_config.pyx
@@ -2,34 +2,44 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+from cuda.core.experimental._utils.cuda_utils cimport (
+    HANDLE_RETURN,
+)
+
+import threading
+
 from cuda.core.experimental._device import Device
 from cuda.core.experimental._utils.cuda_utils import (
     CUDAError,
     cast_to_3_tuple,
     driver,
     get_binding_version,
-    handle_return,
 )
 
-# TODO: revisit this treatment for py313t builds
+
 cdef bint _inited = False
 cdef bint _use_ex = False
+cdef object _lock = threading.Lock()
 
 
-cdef void _lazy_init() except *:
-    """Initialize module-level globals for driver version checks."""
+cdef int _lazy_init() except?-1:
     global _inited, _use_ex
     if _inited:
-        return
+        return 0
 
     cdef tuple _py_major_minor
     cdef int _driver_ver
+    with _lock:
+        if _inited:
+            return 0
 
-    # binding availability depends on cuda-python version
-    _py_major_minor = get_binding_version()
-    _driver_ver = handle_return(driver.cuDriverGetVersion())
-    _use_ex = (_driver_ver >= 11080) and (_py_major_minor >= (11, 8))
-    _inited = True
+        # binding availability depends on cuda-python version
+        _py_major_minor = get_binding_version()
+        HANDLE_RETURN(cydriver.cuDriverGetVersion(&_driver_ver))
+        _use_ex = (_driver_ver >= 11080) and (_py_major_minor >= (11, 8))
+        _inited = True
+
+    return 0
 
 
 cdef class LaunchConfig:
--- a/cuda_core/cuda/core/experimental/_launcher.pyx
+++ b/cuda_core/cuda/core/experimental/_launcher.pyx
@@ -14,14 +14,14 @@ from cuda.core.experimental._utils.cuda_utils cimport (
     HANDLE_RETURN,
 )
 
+import threading
+
 from cuda.core.experimental._module import Kernel
 from cuda.core.experimental._stream import Stream
 from cuda.core.experimental._utils.cuda_utils import (
     _reduce_3_tuple,
     get_binding_version,
 )
-
-import threading
 
 
 cdef bint _inited = False


leofang:
/ok to test [51635bf](https://github.com/NVIDIA/cuda-python/pull/1390/commits/51635bf21c506495fd258701b7e080964969c55e)

leofang:
For comparison, this is with `cuda.bindings`:
```python
In [16]: %timeit driver.cuLaunchKernel(kernel._handle, *config.grid, *config.block, 0, s.handle, ((0, 0), (ctypes.c_void_p, ctypes.c_void_p)), 0)
3.98 μs ± 27.6 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

cuda_core/cuda/core/experimental/_launch_config.pyx
```diff
@@ -2,34 +2,44 @@
...
-    """Initialize module-level globals for driver version checks."""
+cdef int _lazy_init() except?-1:
     global _inited, _use_ex
     if _inited:
-        return
+        return 0
 
     cdef tuple _py_major_minor
     cdef int _driver_ver
+    with _lock:
```
rparolin, line 32: Question: Was this broken previously? Why did we have to add double checked locking?

cuda_core/cuda/core/experimental/_launch_config.pyx
```diff
@@ -127,7 +137,42 @@ cdef class LaunchConfig:
                 f"block={self.block}, shmem_size={self.shmem_size}, "
                 f"cooperative_launch={self.cooperative_launch})")
 
+    cdef cydriver.CUlaunchConfig _to_native_launch_config(self):
+        _lazy_init()
+        # TODO: memset to zero?
+        cdef cydriver.CUlaunchConfig drv_cfg
```
rparolin, line 143: Do these types have a deterministic state after they've been created? Or does no ctor run leaving them in whatever state the bits on the stacks dictate?  If its the latter, I'd memset the structures to be zero.

cuda_core/cuda/core/experimental/_launcher.pyx
```diff
@@ -1,38 +1,51 @@
...
-_use_ex = None
+
+cdef bint _inited = False
+cdef bint _use_ex = False
+cdef object _lock = threading.Lock()
 
 
-def _lazy_init():
-    global _inited
+cdef int _lazy_init() except?-1:
```
rparolin, line 32: Is this violating DRY? Could we a hoist this implementation into a common place or create a primitive that implements the double check locking but accepts a user lambda to preps the data its fetching.

cuda_core/cuda/core/experimental/_launch_config.pyx
leofang: Sort of. It would be "broken" in free threading Python, so we've been gradually adding locks here and there to get prepared (https://github.com/NVIDIA/cuda-python/issues/1389, https://github.com/NVIDIA/cuda-python/pull/1388#discussion_r2624693133). 

After this release we should plan for a global sweep and catch as many thread-safety issues as we can. FWIW @seberg has done quite a bit of work for CuPy lately, and I think in general it is a good improvement even for single-threaded program.

cuda_core/cuda/core/experimental/_launch_config.pyx
leofang: Yeah this is Cython saying "create this struct on the stack." I actually don't think we need to zero-initialize them. I read the API reference and don't find any requirement for these two structs, and in the end we overwrite all of their members so we are cool.

cuda_core/cuda/core/experimental/_launcher.pyx
leofang: Let me keep this thread open. IIRC Ralf has a similar issue tracking such needs. I'll dig it out.

leofang:
Thanks, Rob!

cc @kkraus14 @cpcloud @emcastillo for vis

cuda_core/cuda/core/experimental/_launch_config.pxd
```diff
@@ -2,18 +2,23 @@
...
-    cdef public tuple cluster
-    cdef public tuple block
-    cdef public int shmem_size
-    cdef public bint cooperative_launch
+    cdef:
+        public tuple grid
+        public tuple cluster
+        public tuple block
+        public int shmem_size
+        public bint cooperative_launch
```
leofang, line 17: Leaving a quick note here in case I forget. 

I wasn't super happy about this PR and it's why I was on the fence of pushing this forward: Our design of `LaunchConfig` allows reusing the Python object across multiple launches. But, it is too flexible that we pay the price of maintaining the public attributes (both readable and writable in Python) associated with Python overhead. If we don't think this is reasonable, we should break it before GA, and turn `LaunchConfig` into an immutable object.

kkraus14: In theory we could store the native types and have getter / setter properties that translate to/from tuples as needed? I agree we should avoid paying the Python overhead here if possible.

cuda_core/cuda/core/experimental/_launch_config.pyx
seberg: Nothing too important, but just some FYIs in case it is interesting:
* Even on normal Python `get_binding_version()` might be releasing the GIL at some point (it looks complicated enough), so it may be less likely, but in this case the situation did likely not change much.
  (For pure C-functions, this is clear-cut, I think even very simple Python methods can safe due to the GIL.  For functions where the GIL is enough `with /@cython.critical_section` is a good tool)
* In this particular case, I guess things are likely OK because `cuDriverGetVersion` is presumably thread-safe and there is no problem if this init code runs more than once (there is no convenience "exactly once" functionality in Python/Cython yet).
* In theory, this may still not fully safe (in free-threaded!). I wouldn't be surprised to find this kind of issue in CuPy still (hopefully it doesn't matter in practice and thread-sanitizer eventually helps flush them out).
  Just because `_inited` is set and visible to the another thread does _not_ imply a change like `_use_ex` is too (i.e. they might be written out-of-order and not a single write). Making `_inited` an `std::atomic<bool>` would fix this.
  (From the C++ docs: in practice, but on non-ARM, if the compiler doesn't re-order those writes the CPU won't either. Also the default `atomic` load/store guarantees are actually stronger than what would be needed here.)

I am saying mainly FYI, because I suspect for true confidence thread-sanitizer will be needed and hopefully finds this type of issue (in this case, we are writing on the C-level, if Python just did two inits here it would be thread-safe from a TSAN perspective).

cuda_core/cuda/core/experimental/_launcher.pyx
leofang: I think I had https://github.com/NVIDIA/cuda-python/issues/980 in mind. It makes sense to consolidate these "is something available?" init checks for DRY.

cuda_core/cuda/core/experimental/_launch_config.pxd
leofang: Yes. Though in the case of launch config here, we essentially are wrapping a struct with a flexible array member (the attributes array can be arbitrarily long), which is annoying.

Last night I was thinking about thread safety. But in the present case we still do not offer thread safety anyway (ex: when two threads set `grid` member at the same time, for example). Something to think about in #1389.

cuda_core/cuda/core/experimental/_launcher.pyx
Andy-Jost: FWIW, double-checking a guard variable when locking is standard. In this case, the lock and init check seem fine (they cause no harm) but also probably unnecessary, since racing threads would be racing to set two global variables (`_inited` and `_use_ex`) to identical values.

Andy-Jost (commented):
LGTM

PR #1393, opened 2025-12-17
Initial import of `cuda.core.system`

mdboom:
(Marked as draft as a reminder to not merge until after the 0.5.0 release...)

Prerequisites to get this PR to pass:

- #1404
- #1405 
- #1406 
- #1407
- #1411
- #1415

This is the first landing of `cuda.core.system`, with all of the features in the `nvutil` prototype (which sort of has an arbitrary collection of the most core things in NVML, but is a reasonable starting point for a first PR).

This requires a generator change (not yet merged) to include `AUTO_LOWPP_*` classes in the `.pxd` file so they can be `cimport`'ed.  I know we don't usually do that, but it seems important to be able to use those high-level bindings and not repeat ourselves.  ABI stability there should be ok -- I don't anticipate needing to change anything on the `.pxd` side of those classes.

Following the `nvutil` design, this initializes NVML immediately upon import of `cuda.core.system`.  That feels convenient and may be the right choice, but it will be hard to walk that back.  Questions the NVML docs don't answer for me: are there any use cases where you would want to init/shutdown NVML repeatedly.  The `cuda.bindings.nvml` tests do this, so I know it works.  Is there any harm in `init'ing` and never shutting down -- we could add an `atexit` handler, but I don't know if it's required.

### DIFF
--- a/cuda_core/docs/source/api.rst
+++ b/cuda_core/docs/source/api.rst
@@ -70,7 +70,6 @@ CUDA system information and NVIDIA Management Library (NVML)
 .. autosummary::
    :toctree: generated/
 
-   system.initialize
    system.get_driver_version
    system.get_driver_version_full
    system.get_num_devices


cuda_core/cuda/core/system/system.pyx
```diff
@@ -0,0 +1,73 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+
+from cuda.bindings cimport _nvml as nvml
+
+
+def get_driver_version() -> tuple[int, int]:
```
mdboom: This is a bit confusing.  This is an existing API that returns the CUDA version.  It really should be called `get_cuda_version` to avoid confusion, but that would be a breaking change.  There is a new API to return the driver version called `get_gpu_driver_version` below, but that naming isn't great.

cuda_core/cuda/core/system/__init__.py
leofang, line 1: FYI, `cuda.core` supports any `cuda-bindings`/`cuda-python` 12.x and 13.x, many of which do not have the NVML bindings available. So, we need a version guard here before importing anything that would expect the bindings to exist, and raise an exception in such cases.

mdboom: Ah, good reminder.  I guess that precludes `cimport`'ing anything from `cuda.bindings._nvml`, since `_nvml` is a moving target.  Will just take that out for now...

cuda_core/cuda/core/experimental/system/device.pyx
```diff
@@ -0,0 +1,306 @@
...
+            if isinstance(uuid, bytes):
+                uuid = uuid.decode("ascii")
+            self._handle = nvml.device_get_handle_by_uuid(uuid)
+
+    @property
+    def handle(self) -> int:
+        return self._handle
+
+    @classmethod
+    def get_all_devices(cls) -> Iterable[device]:
```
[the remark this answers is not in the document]
mdboom: Sorry, @Copilot, that isn't true for Cython code.

rparolin:
"Following the nvutil design, this initializes NVML immediately upon import of cuda.core.system. That feels convenient and may be the right choice, but it will be hard to walk that back."

Even if we decide to ship with a backwards compatible version that initializes on import and never shuts down.  It would be if we at least had API that we used internally to test the system init/shutdown for multiple test suites so we can gain confidence that is stable.  Then we can always provide an API to people as an alternative to the legacy behaviour.

mdboom:
@rparolin wrote:
> It would be if we at least had API that we used internally to test the system init/shutdown for multiple test suites so we can gain confidence that is stable.

That is a good point, and we are already doing that as part of the `cuda.bindings.nvml` tests [here](https://github.com/NVIDIA/cuda-python/blob/main/cuda_bindings/tests/nvml/conftest.py#L10).  If a user uses `cuda.bindings.nvml` directly, they would have to manually initialize -- and repeatedly init/shutdown in the same process seems to be tested and working.  It's only when importing the high-level `cuda.core.system` that initialization would be implicit (and permanent to the process).  This seems sort of inline with high-level vs. low-level expectations.

> Then we can always provide an API to people as an alternative to the legacy behaviour.

I'm not sure I follow on this point.  If we make initialization implicit now, we will break everyone in the future if we decide to require it to be explicit.  I guess the alternative is we create a new module that is identical to `cuda.core.system` but doesn't initialize.

Elsewhere in `cuda.core`, it seems to be a combination of either initialization on the first API call, or just checking and raising a helpful error if the user didn't initialize.  I don't know if we try to be consistent with that, or decide on a library-by-library basis here depending on whether there is any need to uninitialize.

mdboom:
I think the answer to whether to implicitly initialize has become clear: If we do that, you can't do things like import the module to build the docs on a machine without CTK/CUDA installed.  So, we need to at minimum do lazy init on first call that needs it...  Rejiggering this PR...

```
  File "/home/runner/miniconda3/envs/cuda-python-docs/lib/python3.12/site-packages/cuda/core/__init__.py", line 31, in <module>
    from cuda.core import system, utils  # noqa: E402
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/miniconda3/envs/cuda-python-docs/lib/python3.12/site-packages/cuda/core/cu13/system/__init__.py", line 25, in <module>
    initialize()
  File "cuda/core/system/_nvml_context.pyx", line 40, in cuda.core.system._nvml_context.initialize
  File "cuda/core/system/_nvml_context.pyx", line 49, in cuda.core.system._nvml_context.initialize
  File "cuda/bindings/_nvml.pyx", line 21323, in cuda.bindings._nvml.init_v2
  File "cuda/bindings/_nvml.pyx", line 21329, in cuda.bindings._nvml.init_v2
  File "cuda/bindings/cy_nvml.pyx", line 15, in cuda.bindings.cy_nvml.nvmlInit_v2
  File "cuda/bindings/_internal/_nvml.pyx", line 3971, in cuda.bindings._internal._nvml._nvmlInit_v2
cuda.bindings._internal.utils.FunctionNotFoundError: function nvmlInit_v2 is not found
```

cuda_core/cuda/core/system/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+    "get_gpu_driver_version",
+    "get_num_devices",
+    "get_process_name",
+    "HAS_WORKING_NVML",
+]
+
+
+from .system import *
+
+if HAS_WORKING_NVML:
```
rparolin: Opinion: `HAS_WORKING_NVML` strikes me as an odd name. I'd be more inclined to name this `IS_NVML_AVAILABLE` or `IS_NVML_INSTALLED`?

mdboom: Yeah, I agree it's an odd name.  It's really "is the cuda bindings that's installed new enough to correctly use NVML", not whether NVML is installed.  Maybe `CUDA_BINDINGS_IS_NVML_COMPATIBLE` is a better name?

cuda_core/cuda/core/system/_device.pyx
```diff
@@ -0,0 +1,309 @@
...
+
+    @property
+    def device_id(self) -> int:
+        """
+        The PCI device id of the device
+        """
+        return self._pci_info.pci_device_id >> 16
+
+
+cdef class Device:
```
rparolin, line 174: Question: cuda.core.system.Device vs cuda.core.Device?  Does it make for there to be two core Device objects? Or should the implementation be merged into cuda.core.system?

cuda_core/cuda/core/system/system.pyx
rparolin: Could we provide those functions for backwards compat and provide a better named alternative that we are willing to support longer term?

cuda_core/cuda/core/system/utils.pyx
```diff
@@ -0,0 +1,46 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from cpython cimport array
+from libc.stdint cimport uint64_t
+
+
+def unpack_bitmask(x: list[int] | array.array) -> list[int]:
```
rparolin: Request: I'd rename the file from `utils.pyx` because util files tend to turn into dumping grounds.  Rename this to something like bit-twiddling.pyx, bit-tricks.pyx, or bitmasks.pyx.

cuda_core/cuda/core/system/_device.pyx
mdboom: IIUC, there is a requirement to be able to use NVML for device information without importing and using all of CUDA (or requiring a CTK install).  So blending the two is explicitly off the table, or could be rather painful.  That said, there is potential for confusion here -- there might be some documentation / naming changes to make to be clearer.  I even am starting to wonder if a separate (non-`cuda.core`) package would be more appropriate.

cuda_core/cuda/core/system/utils.pyx
mdboom: Good point.  Alternatively, since this is also only used by `device.pyx`, it could just be moved there.  It probably makes the most sense as a `.pxi` that is included by `device.pyx`.

cuda_core/cuda/core/system/device_utils.pxi
```diff
@@ -0,0 +1,46 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from cpython cimport array
+from libc.stdint cimport uint64_t
+
+
+def _unpack_bitmask(x: list[int] | array.array) -> list[int]:
```
leofang: cdef inline?

cuda_core/cuda/core/system/_device.pyx
leofang, line 1: Important: `device.pyx` -> `_device.pyx` (don't let users access `cuda.core.system.device` via tab completion).

cuda_core/cuda/core/system/system.pyx
```diff
@@ -0,0 +1,103 @@
...
+    Tuple in the format `(CUDA_MAJOR, CUDA_MINOR)`.
+    """
+    return get_driver_version_full()[:2]
+
+
+def get_driver_version_full() -> tuple[int, int, int]:
+    """
+    The CUDA driver version.
+
+    Tuple in the format `(CUDA_MAJOR, CUDA_MINOR, CUDA_PATCH)`.
```
leofang: Q: Does it ever return `CUDA_PATCH` != 0? I have never seen that.

cuda_core/cuda/core/system/system.pyx
leofang: It is actually confusing to call it `get_cuda_version`. This used to return the UMD (user-mode driver) version, e.g. 13.0, 13.1, .... Now that we have access to NVML, we gain in addition the capability of querying the KMD (kernel-mode driver) version, e.g. 580.65.06. So I'd suggest something like
```python
def get_driver_version(kernel_mode: bool = False) -> tuple[int, ...]:
```
for users to query either UMD or KMD version. Alternatively, we can allow users to query both at once.

cuda_core/cuda/core/system/system.pyx
```diff
@@ -0,0 +1,103 @@
...
+    cdef int v
+    if HAS_WORKING_NVML:
+        initialize()
+        v = nvml.system_get_cuda_driver_version()
+    else:
+        v = handle_return(driver.cuDriverGetVersion())
+    return (v // 1000, (v // 10) % 100, v % 10)
+
+
+def get_gpu_driver_version() -> tuple[int, ...]:
```
leofang: ditto, having another `get_gpu_driver_version` function is confusing and should be combined into the above one. This is for querying KMD version.

cuda_core/cuda/core/system/system.pyx
leofang: In any case, `get_cuda_version` is too ambiguous. Depending on the context a CUDA version could refer to any of the versions of KMD, UMD, CTK, NVCC, NVRTC, CUDART, ... and we should help avoid spreading such confusion.

cuda_core/cuda/core/system/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+    from ._nvml_context import initialize
+    from .device import Device, DeviceArchitecture
+    from .exceptions import *
+
+    __all__.extend(
+        [
+            "initialize",
```
leofang: Q: Can we do lazy init when invoking any NVML functionalities, instead of exposing `initialize()`? For example, we do not expose `cuInit()` in cuda.core.

cuda_core/cuda/core/system/exceptions.py
leofang, line 1: ditto, prefix the module name with an underscore

cuda_core/cuda/core/system/_system.pyx
leofang, line 1: ditto, prefix the module name with an underscore

cuda_core/cuda/core/system/system.pyx
```diff
@@ -0,0 +1,103 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+
+# This file needs to either use NVML exclusively, or when `cuda.bindings._nvml`
+# isn't available, fall back to non-NVML-based methods for backward
+# compatibility.
+
+
+from cuda.bindings._version import __version_tuple__ as _BINDINGS_VERSION
```
leofang: Important: I don't think `__version_tuple__` exists before we switched to generate `_version.py` using setuptools-scm?

cuda_core/cuda/core/system/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+            "get_nvml_version",
+            "Device",
+            "DeviceArchitecture",
+            "UninitializedError",
```
leofang: Q: Starting here, instead of duplicating the list is it possible to simply extend this list with `exceptions.__all__`?

cuda_core/cuda/core/system/exceptions.py
```diff
@@ -0,0 +1,73 @@
...
+ArgumentVersionMismatchError = nvml.ArgumentVersionMismatchError
+DeprecatedError = nvml.DeprecatedError
+NotReadyError = nvml.NotReadyError
+GpuNotFoundError = nvml.GpuNotFoundError
+InvalidStateError = nvml.InvalidStateError
+ResetTypeNotSupportedError = nvml.ResetTypeNotSupportedError
+UnknownError = nvml.UnknownError
+
+
+__all__ = [
```
leofang, line 42: Q: Should we also expose the base exception `NvmlError`?

cuda_core/cuda/core/system/system.pyx
mdboom: Thanks for the context.  Makes sense.  I think I've addressed this now in the latest commit.

cuda_core/cuda/core/system/__init__.py
mdboom: Sure -- That's probably fine.  It does mean it will be impossible to uninitialize, and I didn't know if that was a use case we needed to support.

cuda_core/cuda/core/system/system.pyx
mdboom: Good point.

cuda_core/cuda/core/system/__init__.py
leofang: We already don't uninitialize in this PR using `nvmlShutdown()` anyway. Also I'd be curious how NVUtils handles this global state. (CUDA libs usually use library handles so that we can do library init/deinit on a per-handle basis; NVML and cuFILE were unfortunately designed with global states.) Would be nice to have @pentschev chiming in here 🙂

cuda_core/cuda/core/system/system.pyx
mdboom: Yeah, I don't see that either, but the nvml.h header says it's possible, so I think we should make it possible in the API.  And for the kernel mode, it definitely has a meaningful third number...

cuda_core/cuda/core/system/__init__.py
mdboom: I will need to rejigger this to use the double-checking pattern, otherwise we will have a lock on every NVML call which won't perform well at all.

cuda_core/cuda/core/system/_device.pyx
mdboom: To summarize the offline discussion:

1) cuda.core.system definitely needs to be free of importing/initializing CUDA
2) The naming is confusing.  We can at least document why there are two different `Device` classes, but there doesn't seem to be an obvious way to name this better (though suggestions welcome).  Shipping it as a separate package doesn't seem desirable either.
3) As a follow-on to this PR, we can add helpers to move between each of these `Device` types.

cuda_core/cuda/core/system/__init__.py
mdboom: nvutils initializes at import and doesn't ever shutdown.  Initializing at import caused a bunch of issues here (can't generate docs on a machine without a GPU, for one...), so initializing at first call (or requiring explicit initialization) seems like the only options here.

cuda_core/cuda/core/system/_device.pyx
rparolin: sgtm! Thanks for pulling this together Mike!

rparolin:
lgtm to me. Just make sure the spelling mistakes that copilot flagged are resolved.  Thank you!

cuda_core/cuda/core/system/__init__.py
pentschev: I think for `nvmlShutdown()` to properly function we would need to refcount with a singleton and implement a [`weakref.finalize`](https://docs.python.org/3/library/weakref.html#weakref.finalize) to clear. However, I _think_ we're probably not doing `nvmlShutdown()` properly anywhere, it's hard to ensure all libraries agree on a resource management for NVML. With that said, I believe we have never seen issues because of improper/missing `nvmlShutdown()` uses, so to the best of my knowledge I'd say we're probably fine just "forgetting" (i.e., ignoring) to shut it down for now.

mdboom: Thanks for the context @pentschev.  That's the easiest thing to do ;)

PR #1377, opened 2025-12-16
cuda-core experimental namespace deprecation (attempt v1)

rwgk:
Result of starting over fresh based on experience gathered under #1298

This PR adds `print_wheel_directory_structure()` in `ci/tools/merge_cuda_core_wheels.py`, which helped greatly figuring out the required changes to that script. The output of the function is likely to be useful for troubleshooting in the future.

Questions:

* What is preferred? `cuda/core/_include` or `cuda/core/inclulde`? — Currently it is `_include`, based on [this feedback under #1298](https://github.com/NVIDIA/cuda-python/pull/1298#discussion_r2604293064).
* Do we want to keep `cuda/core/__init__.pyx` in the wheels (it does not seem to be needed)? — Keep or revert commit a63913c3d81a893e5f844ed2314dee61ed206176

TODO:
* ~Restore `cuda_core/tests/test_experimental_backward_compat.py` from #1298 (it somehow got lost).~
* In a temporary branch, revert all changes under `cuda_core/tests/`. The original tests should still pass. — See #1394

PR #1397, opened 2025-12-17
fix: support zero-sized arrays as input to `StridedMemoryView.from_cuda_array_interface`

### LINKED ISSUE #1396: [BUG]: `StridedMemoryView.from_cuda_array_interface` doesn't support zero-sized arrays
### Is this a duplicate?

### Type of Bug

Runtime Error

### Component

cuda.core

### Describe the bug

Calling `StridedMemoryView.from_cuda_array_interface` raises an exception when given a zero-sized array. `StridedMemoryView.from_dlpack` works.

### How to Reproduce

```python
import cupy as cp

x = cp.zeros(0, dtype="int64")
StridedMemoryView.from_cuda_array_interface(x, stream_ptr=0)
```

### Expected behavior

I would expect the behavior to be effective the same as `StridedMemoryView.from_dlpack`.

### Operating System

_No response_

### nvidia-smi output

_No response_

cpcloud:
Add support for zero-sized array input to `StridedMemoryView.from_cuda_array_interface`. Closes #1396.

### DIFF
--- a/cuda_core/docs/source/release/0.5.0-notes.rst
+++ b/cuda_core/docs/source/release/0.5.0-notes.rst
@@ -61,7 +61,7 @@ None.
 Fixes and enhancements
 ----------------------
 
-- Zero-size arrays are now supported as inputs to ``StridedMemoryView.from_cuda_array_interface``.
+- Zero-size arrays are now supported as inputs when constructing ``StridedMemoryView``.
 - Most CUDA resources can be hashed now.
 - Python ``bool`` objects are now converted to C++ ``bool`` type when passed as kernel arguments (previously converted to ``int``).
 - Restored v0.3.x :class:`MemoryResource` behaviors and missing MR attributes for backward compatibility.


cuda_core/cuda/core/experimental/_memoryview.pyx
```diff
@@ -582,10 +582,13 @@ cpdef StridedMemoryView view_as_cai(obj, stream_ptr, view=None):
...
-        driver.cuPointerGetAttribute(
-            driver.CUpointer_attribute.CU_POINTER_ATTRIBUTE_DEVICE_ORDINAL,
-            buf.ptr))
+    if buf.ptr != 0:
+        buf.device_id = handle_return(
+            driver.cuPointerGetAttribute(
+                driver.CUpointer_attribute.CU_POINTER_ATTRIBUTE_DEVICE_ORDINAL,
+                buf.ptr))
+    else:
+        buf.device_id = -1
```
kkraus14: Is there a precedent for using `-1` for the device_id here? Since CAI doesn't actually capture the device, my assumption would be that it uses the current Device which can be retrieved via `driver.cuCtxGetDevice`.

Maybe a moot point since this is a null ptr anyway.

cpcloud: It's a bit ambiguous, but I went with the convention for `device_id` coming from CPU arrays, from this bit of the `StridedMemoryView` docs:

```
    device_id : int
        The device ID for where the tensor is located. It is -1 for CPU tensors
        (meaning those only accessible from the host).
```

I _did_ hit this in the context of a numba kernel test, so maybe it does make sense to set the device id to the current device?

cpcloud: Changed to use `cuCtxGetDevice()`.

kkraus14 (approved):

kkraus14:
@cpcloud can you add a line to the 0.5.0 release notes for this and then we'll admin merge?

cpcloud:
@kkraus14 done

leofang:
Actually, thinking about this more... why do we need a release note entry for this fix? `from_cuda_array_interface` is new, anyway, so I wouldn't even call it a fix. This is more like a follow-up PR.

leofang:
> Actually, thinking about this more... why do we need a release note entry for this fix? `from_cuda_array_interface` is new, anyway, so I wouldn't even call it a fix. This is more like a follow-up PR.

@cpcloud @kkraus14 let's come up with a conclusion (mention or not) and I'll incorporate it in my follow-up PR.

leofang:
oh... @cpcloud there is a merge conflict now because of #1377 😛 Was wondering why I can't admin-merge...

kkraus14:
> > Actually, thinking about this more... why do we need a release note entry for this fix? `from_cuda_array_interface` is new, anyway, so I wouldn't even call it a fix. This is more like a follow-up PR.
> 
> @cpcloud @kkraus14 let's come up with a conclusion (mention or not) and I'll incorporate it in my follow-up PR.

I think this bug existed previously in the `StridedMemoryView.__init__` before we had the `from_*` constructors.

cpcloud:
I added some less specific language to the release notes.

leofang:
/ok to test [5a33e7a](https://github.com/NVIDIA/cuda-python/pull/1397/commits/5a33e7a306015e9eaef0516ea10a66a4903cdcd6)

leofang:
Let me admin-merge since most of the CI runs pass and we hit the hanging issue, xref: https://github.com/NVIDIA/cuda-python/pull/1377#issuecomment-3666850503

PR #1398, opened 2025-12-17
[pixi] Add pixi support for running all sub-package tests

rparolin:
# Add pixi support for running all sub-package tests

## Summary

Adds a root-level `pixi.toml` that enables running tests for all sub-packages (`cuda_pathfinder`, `cuda_bindings`, `cuda_core`) via `pixi run test`. Supports both default and CUDA version-specific environments (`cu12`, `cu13`).

## Changes

- Added root-level `pixi.toml` with `cu12` and `cu13` environments matching sub-packages
- Added `test` tasks that orchestrate sub-package operations

### Run tests
```bash
# Run all tests (default environment)
pixi run test

# Run tests with CUDA 12 environment
pixi run -e cu12 test

# Run tests with CUDA 13 environment
pixi run -e cu13 test
```

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -22,7 +22,6 @@ PIXI_ENVIRONMENT_NAME = "${PIXI_ENVIRONMENT_NAME/default/cu13}"
 #
 # Usage: pixi run test | pixi run -e cu12 test | pixi run -e cu13 test
 [target.linux.tasks.test-pathfinder]
-env.PIXI_ENVIRONMENT_NAME = "$PIXI_ENVIRONMENT_NAME"
 cmd = [
     "pixi",
     "run",
@@ -32,7 +31,6 @@ cmd = [
 ]
 
 [target.linux.tasks.test-bindings]
-env.PIXI_ENVIRONMENT_NAME = "$PIXI_ENVIRONMENT_NAME"
 cmd = [
     "pixi",
     "run",
@@ -42,7 +40,6 @@ cmd = [
 ]
 
 [target.linux.tasks.test-core]
-env.PIXI_ENVIRONMENT_NAME = "$PIXI_ENVIRONMENT_NAME"
 cmd = ["pixi", "run", "--manifest-path", "$PIXI_PROJECT_ROOT/cuda_core", "test"]
 
 [target.linux.tasks.test]


rparolin:
@greptileai please review

pixi.toml
```diff
@@ -0,0 +1,95 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+[workspace]
+channels = ["conda-forge"]
+platforms = ["linux-64", "linux-aarch64", "win-64"]
+preview = ["pixi-build"]
```
cpcloud: You're not building a package anywhere, so this is unnecessary.

```suggestion
```

pixi.toml
```diff
@@ -0,0 +1,95 @@
...
+cuda-version = ">=13.1,<14"
+
+[environments]
+cu12 = { features = ["cu12"], solve-group = "cu12" }
+cu13 = { features = ["cu13"], solve-group = "cu13" }
+
+# Install tasks: Set up all sub-package environments
+# Usage: pixi run install (default) or pixi run -e cu12 install or pixi run -e cu13 install
+# Detects environment from CONDA_DEFAULT_ENV (set by pixi when using -e flag)
+[target.linux.tasks.install-pathfinder]
```
cpcloud: Maybe you already tried this, but the "pixi way" to do this would be something like a separate feature for each library, and then you combine features to make environments that you run stuff in.

If you want different cuda versions, you make different environments.

It's actually designed explicitly so that you end up with matrix'd environments, e.g., `cu-13-0-py313` (from `numba-cuda`).

Can we try that approach instead of embedding pixi inside of bash inside of pixi?

pixi.toml
```diff
@@ -0,0 +1,95 @@
...
+    "set -e; ENV=\"\"; [[ \"${CONDA_DEFAULT_ENV:-}\" == *\"cu12\"* ]] && ENV=\"cu12\"; [[ \"${CONDA_DEFAULT_ENV:-}\" == *\"cu13\"* ]] && ENV=\"cu13\"; cd cuda_pathfinder && pixi ${ENV:+--environment $ENV} install && cd ../cuda_bindings && pixi ${ENV:+--environment $ENV} install && cd ../cuda_core && pixi ${ENV:+--environment $ENV} install"
+]
+
+
+# Test tasks: Run tests (assumes environments are already installed)
+# Order: pathfinder -> bindings -> core (matching scripts/run_tests.sh)
+# Each sub-package has its own pixi.toml and environment
+# Usage: pixi run test (default) or pixi run -e cu12 test or pixi run -e cu13 test
+[target.linux.tasks.test-pathfinder]
+cmd = ["bash", "-c", "ENV=\"\"; [[ \"${CONDA_DEFAULT_ENV:-}\" == *\"cu12\"* ]] && ENV=\"cu12\"; [[ \"${CONDA_DEFAULT_ENV:-}\" == *\"cu13\"* ]] && ENV=\"cu13\"; cd cuda_pathfinder && if [ -n \"$ENV\" ]; then pixi run --environment \"$ENV\" test; else pixi run test; fi"]
```
cpcloud: This also makes me think you may not be aware that tasks can be given an environment to run in:

```
pixi run -e my-env my-task
```

EDIT: I see you're using this, but I still don't understand why `pixi run -e my-env my-task` isn't just the way we do this instead of looking at environment variables.

cpcloud: I also don't think it's really that bad to have to pass the environment around. Is it less convenient? Maybe, but IMO it's worth it to not have to read bash LOL

rparolin: I greatly simplified the bash expression.

cpcloud (approved):

PR #1409, opened 2025-12-18
Handle unsupported device attributes gracefully

Andy-Jost:
## Summary

Adds defensive error handling in `DeviceProperties._get_attribute()` to gracefully handle cases where `cuDeviceGetAttribute` returns `CUDA_ERROR_INVALID_VALUE` for unsupported attributes.

## Changes

- Modified `_get_attribute` in `_device.pyx` to check for `CUDA_ERROR_INVALID_VALUE` before raising an exception
- When the driver returns this error, the method now returns `0` as a conservative default
  - For boolean attributes (e.g., `vulkan_cig_supported`), `0` means `False`
  - For integer attributes, `0` indicates "not supported" or "disabled"

## Rationale

This addresses two scenarios:

1. **Version mismatch**: When cuda-core is compiled against a newer CUDA toolkit (e.g., 12.9) but runs on an older driver (e.g., 12.8), newer attribute IDs may not be recognized by the driver
2. **Driver bugs**: Cases like nvbug5605010 where the driver advertises CUDA 12.9 capability but incorrectly rejects attribute 138 (`CU_DEVICE_ATTRIBUTE_VULKAN_CIG_SUPPORTED`) with `CUDA_ERROR_INVALID_VALUE`

Previously, these scenarios would cause `DeviceProperties` properties to raise `CUDAError` exceptions, breaking tests and user code. With this change, unsupported attributes return sensible defaults (`0`/`False`) instead of raising exceptions.

## Test Coverage

- Existing tests should continue to pass
- Tests that previously failed with `CUDA_ERROR_INVALID_VALUE` (e.g., `test_device.py`) will now pass by returning `0`
- No new tests added (defensive handling preserves existing behavior for supported attributes)

## Related Work

- Addresses nvbug5605010 (driver 575.66 incorrectly rejecting attribute 138)
- Handles version mismatches observed on CUDA 12.8 driver machines rejecting attributes 138, 141, 142 from `cuda-core` built against CUDA 12.9

### DIFF
--- a/.github/ISSUE_TEMPLATE/release_checklist.yml
+++ b/.github/ISSUE_TEMPLATE/release_checklist.yml
@@ -20,7 +20,6 @@ body:
         - label: File an internal nvbug to communicate test plan & release schedule with QA
         - label: Ensure all pending PRs are reviewed, tested, and merged
         - label: Check (or update if needed) the dependency requirements
-        - label: Bump the version
         - label: Create a public rc tag
         - label: "Point QA to fetch public artifacts (wheels) from the GHA run ID, example: `gh run download 12323257563 -p \"cuda-core*\" -R NVIDIA/cuda-python`"
         - label: Wait for QA reports and fix any issues found
--- a/.github/workflows/test-wheel-linux.yml
+++ b/.github/workflows/test-wheel-linux.yml
@@ -278,7 +278,9 @@ jobs:
         run: |
           set -euo pipefail
           pushd cuda_pathfinder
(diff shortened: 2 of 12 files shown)


leofang:
I think this was an oversight when the new attributes were added. We should not always return 0 because it'd shadow the actual issue (driver version insufficient) and our standard treatment is to add version checks in both the codebase and the corresponding test suite to ensure cuda-core runs with any CUDA driver 12.x and 13.x. This needs to be revisited before the next release is out.

PR #1411, opened 2025-12-18
Use setuptools-scm to determine versions

mdboom:
This uses the `setuptools-scm` package to automatically determine the version of each package from the corresponding git tag.

This is motivated by a couple of things:

- Removing more steps in our release project (in this case, manually updating version numbers)
- Making the version number of work on main always be greater than the latest release, so we can dynamically turn on new features in `cuda.core` based on in-development work in `cuda.bindings`.

From an offline discussion with @leofang, we have a requirement here that release candidate builds can be shipped as-is as final releases.  Therefore, the tag matching regexes have been written so that:

| tag | version |
| --- | --- |
| v13.1.1rc1 | 13.1.1 |
| v13.1.1 | 13.1.1 |
| (no tag) | 13.1.2.dev34+gecfc79fb9 |

This does have the downside that the first and second builds above, even if the hashes the tags point at are different, will have the same filename.  The builds will be distinguishable by inspecting `cuda.bindings._version.__commit_id__`, however.

Other implications of this change:

Packages that are locally installed from git from an untagged commit will have complicated-but-helpful version numbers like `13.1.2.dev34+gecfc79fb9`, so parsing them is a little more complicated than `.split('.')`.  To help with that, `setuptools-scm` also generates a `version_tuple` that dependent projects can use:

```
__version__ = version = '13.1.2.dev34+gecfc79fb9'
__version_tuple__ = version_tuple = (13, 1, 2, 'dev34', 'gecfc79fb9')

__commit_id__ = commit_id = 'gecfc79fb9'
```

but that should only effect developers installing non-released versions of our software.  People installing from official packages should continue to see simple version numbers.

Building from an `sdist` continues to work -- the tarball will contain the generated `_version.py` files, and will be (as it should be) hardcoded with the version that the `sdist` was generated from.

### DIFF
--- a/.github/workflows/test-wheel-linux.yml
+++ b/.github/workflows/test-wheel-linux.yml
@@ -36,7 +36,7 @@ jobs:
       OLD_BRANCH: ${{ steps.compute-matrix.outputs.OLD_BRANCH }}
     steps:
       - name: Checkout ${{ github.event.repository.name }}
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8  # v5.0.0
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8  # v6.0.1
 
       - name: Validate Test Type
         run: |
@@ -93,7 +93,7 @@ jobs:
         run: nvidia-smi
 
       - name: Checkout ${{ github.event.repository.name }}
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8  # v5.0.0
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8  # v6.0.1
 
       - name: Setup proxy cache
         uses: nv-gha-runners/setup-proxy-cache@main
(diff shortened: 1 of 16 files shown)


rwgk:
I'm also a bit worried about doing something unusual for the `rc` suffix (my general thinking: "every wrinkle we add is a stumbling hazard") but our constraints are also unusual. Overall, this looks like a big step in the right direction, I think we should give this a try and fix forward as needed.

mdboom:
Cc: @leofang, @cryos.  Dismissed the approval because I had to make a couple more changes to get CI to pass.  Also see the backport of this to the 12.9.x branch in #1415.

leofang:
> Cc: @leofang, @cryos. Dismissed the approval because I had to make a couple more changes to get CI to pass. Also see the backport of this to the 12.9.x branch in #1415.

@mdboom sorry I could not respond promptly. I see the automerge kicked in and this PR is now merged -- is it intended?

leofang:
> Overall, this looks like a big step in the right direction, I think we should give this a try and fix forward as needed.

Yes, this was also the conclusion we reached on the Thursday team-sync.

.github/workflows/test-wheel-linux.yml
leofang, line 1: Thanks for the bug fix!

PR #1412, opened 2025-12-18
Add CUDA version compatibility check

Andy-Jost:
## Summary

Adds a version compatibility check that warns users when cuda-bindings was compiled against a newer CUDA major version than the installed driver supports.

### cuda-bindings
- Added `check_cuda_version_compatibility()` function in `cuda/bindings/utils/_version_check.py`
- Compares compile-time `CUDA_VERSION` vs runtime `cuDriverGetVersion()`
- Exported from `cuda.bindings.utils`
- Added comprehensive unit tests in `tests/test_version_check.py`

### cuda-core
- `Device.__new__` calls `check_cuda_version_compatibility()` after `cuInit` succeeds
- Imports the function directly from `cuda.bindings.utils`

## Rationale

When cuda-bindings is built against CUDA 13 headers but the user's driver only supports CUDA 12, many features will silently fail or behave unexpectedly. This check provides early, clear feedback:

```
UserWarning: cuda-bindings was built against CUDA 13.0, but the installed driver 
only supports CUDA 12.8. Some features may not work correctly. Consider updating 
your NVIDIA driver. Set CUDA_PYTHON_DISABLE_VERSION_CHECK=1 to suppress this warning.
```

## Design

- **Provided by cuda-bindings**: The version check implementation lives in `cuda.bindings.utils` since it checks cuda-bindings' compile-time version
- **Invoked by cuda-core**: Called when `Device` first triggers CUDA initialization
- **Runs once**: Uses a module-level flag to ensure the check runs only once per process
- **Non-blocking**: Warning only, does not prevent operation
- **Suppressible**: Set `CUDA_PYTHON_DISABLE_VERSION_CHECK=1` to disable
- **Major version only**: No warning for minor version differences (handled by graceful degradation per PR #1409)

## Future Work

We could not find a suitable place to invoke the version check automatically within cuda-bindings itself (e.g., hooking into `cuInit`), so the check is currently triggered by cuda-core. This may be revisited in the future.

## Test Coverage

7 tests in cuda-bindings covering:
- No warning when driver is newer
- No warning when same major version
- Warning when compile major > driver major
- Warning only issued once per process
- Suppression via environment variable
- Silent handling of driver errors
- Silent handling of missing CUDA_VERSION attribute

## Related Work

- Complements PR #1409 which handles graceful degradation for minor version mismatches


cuda_core/cuda/core/_utils/cuda_utils.pyx
```diff
@@ -288,6 +289,67 @@ class Transaction:
...
+        return  # Can't check, skip silently
+
+    compile_major = compile_version // 1000
+    runtime_major = runtime_version // 1000
+
+    if compile_major > runtime_major:
+        compile_minor = (compile_version % 1000) // 10
+        runtime_minor = (runtime_version % 1000) // 10
+        warnings.warn(
+            f"cuda-python was built against CUDA {compile_major}.{compile_minor}, "
```
leofang: Important: `cuda-python` is a meta package. We need to be clear what is actually being used.
```suggestion
            f"cuda-bindings was built against CUDA {compile_major}.{compile_minor}, "
```

leofang (changes requested):

cuda_bindings/docs/source/environment_variables.rst
```diff
@@ -9,6 +9,8 @@ Runtime Environment Variables
 
 - ``CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM`` : When set to 1, the default stream is the per-thread default stream. When set to 0, the default stream is the legacy default stream. This defaults to 0, for the legacy default stream. See `Stream Synchronization Behavior <https://docs.nvidia.com/cuda/cuda-runtime-api/stream-sync-behavior.html>`_ for an explanation of the legacy and per-thread default streams.
 
+- ``CUDA_PYTHON_DISABLE_VERSION_CHECK`` : When set to 1, suppresses the warning that is issued when ``cuda.core`` detects that ``cuda-bindings`` was compiled against a newer CUDA major version than the installed driver supports. This warning helps identify version mismatches that may cause features to not work correctly.
```
leofang: This env var is added to cuda-bindings docs, but the actual change is made to cuda-core? Can we move them to the right place?

Actually, thinking about this we probably should do this check in cuda-bindings instead of cuda-core, since it's what actually matters. The question is if there is a good place that would allow lazy-init. Remember, both cuda-core and cuda-bindings can be installed and imported on CPU-only machines, without driver or a physical GPU installed. I can't think of an obvious solution off top of my head.

cuda_core/cuda/core/_utils/cuda_utils.pyx
```diff
@@ -288,6 +289,67 @@ class Transaction:
...
+            f"but the installed driver only supports CUDA {runtime_major}.{runtime_minor}. "
+            f"Some features may not work correctly. Consider updating your NVIDIA driver. "
```
leofang: This statement is nerve wrecking considering that we support CUDA Minor Version Compatibility. In a few cases, we check the minor version in cuda-core to ensure what we need is available at run time, but only when we actually need them. I think we should rephrase this warning.

Andy-Jost: This warning is only issued in the case of a MAJOR version mismatch.

cuda_bindings/docs/source/environment_variables.rst
Andy-Jost: What about `cuInit`? That seems like a logical place for the version check, though the change might be more complex.

Andy-Jost: As discussed, cuda-bindings provides the check; cuda-core invokes it. Other consumers of cuda-bindings can also invoke the check.

cuda_bindings/cuda/bindings/utils/_version_check.py
```diff
@@ -0,0 +1,75 @@
...
+    try:
+        compile_version = driver.CUDA_VERSION  # e.g., 13010
+    except AttributeError:
+        # Older cuda-bindings may not expose CUDA_VERSION
+        return
+
+    # Get runtime driver version
+    err, runtime_version = driver.cuDriverGetVersion()
+    if err != driver.CUresult.CUDA_SUCCESS:
+        return  # Can't check, skip silently
```
rparolin: Not being able to query from the driver version is worthy of a warning to the user instead of silently eating it.

cuda_bindings/cuda/bindings/utils/_version_check.py
```diff
@@ -0,0 +1,75 @@
...
+            f"cuda-bindings was built against CUDA {compile_major}.{compile_minor}, "
+            f"but the installed driver only supports CUDA {runtime_major}.{runtime_minor}. "
+            f"Some features may not work correctly. Consider updating your NVIDIA driver. "
+            f"Set CUDA_PYTHON_DISABLE_VERSION_CHECK=1 to suppress this warning.",
+            UserWarning,
+            stacklevel=3,
+        )
+
+
+def _reset_version_compatibility_check():
```
rparolin: Is there a real use case where production code would call this function? If it’s only used by tests, then it seems reasonable to rely implicitly on the global and move the function into the test code instead.

cuda_bindings/cuda/bindings/utils/_version_check.py
```diff
@@ -0,0 +1,75 @@
...
+# SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
+
+import os
+import warnings
+
+# Track whether we've already checked version compatibility
+_version_compatibility_checked = False
+
+
+def check_cuda_version_compatibility():
```
rwgk: I'd definitely add `major` into the name for clarity, and make it more obvious what the function actually does:

```
def warn_if_cuda_major_version_mismatch()
```

Also `CUDA_PYTHON_DISABLE_MAJOR_VERSION_WARNING`

cuda_bindings/cuda/bindings/utils/_version_check.py
```diff
@@ -0,0 +1,75 @@
...
+        return
+
+    # Import here to avoid circular imports and allow lazy loading
+    from cuda.bindings import driver
+
+    # Get compile-time CUDA version from cuda-bindings
+    try:
+        compile_version = driver.CUDA_VERSION  # e.g., 13010
+    except AttributeError:
+        # Older cuda-bindings may not expose CUDA_VERSION
```
rwgk: Hm ... this code is in the cuda-bindings sources, it cannot possibly be in older cuda-bindings releases.

I think the try-except can go completely?

cuda_bindings/cuda/bindings/utils/_version_check.py
rwgk: I'd even assume here that this should never error, and surface an exception if it does, thinking something is very wrong at that point, and flagging it immediately is most helpful.

cuda_bindings/cuda/bindings/utils/_version_check.py
```diff
@@ -0,0 +1,75 @@
...
+        return  # Can't check, skip silently
+
+    compile_major = compile_version // 1000
+    runtime_major = runtime_version // 1000
+
+    if compile_major > runtime_major:
+        compile_minor = (compile_version % 1000) // 10
+        runtime_minor = (runtime_version % 1000) // 10
+        warnings.warn(
+            f"cuda-bindings was built against CUDA {compile_major}.{compile_minor}, "
```
rwgk: I think it'd be most helpful to be clear that the major version mismatch is the primary concern. The minor version information is more distracting than helpful. Suggested warning message:

```
cuda-bindings was built for CUDA major version {compile_major}, but the
NVIDIA driver only supports up to CUDA {runtime_major}. Some cuda-bindings
features may not work correctly. Consider updating your NVIDIA driver,
or using a cuda-bindings version built for CUDA {runtime_major}. (Set
CUDA_PYTHON_DISABLE_MAJOR_VERSION_WARNING=1 to suppress this warning.)
```

cuda_bindings/tests/test_version_check.py
```diff
@@ -0,0 +1,98 @@
...
+from cuda.bindings import driver
+from cuda.bindings.utils import _version_check, warn_if_cuda_major_version_mismatch
+
+
+class TestVersionCompatibilityCheck:
+    """Tests for CUDA major version mismatch warning function."""
+
+    def setup_method(self):
+        """Reset the version compatibility check flag before each test."""
+        _version_check._major_version_compatibility_checked = False
```
rparolin: This is effectively going to reset if the version was checked already and after the unit tests executes the check could potentially execute again.  Shouldn't we cache the current state and before setting it to false for the unit tests and then restore to its previous value on unit test teardown?

Andy-Jost: Someone with a dodgy build might get multiple warnings during testing. Not sure it's worth fixing.

Andy-Jost:
/ok to test https://github.com/NVIDIA/cuda-python/commit/acadb3183278aca6201be0d1b733c2eb03b57dd5

cuda_bindings/cuda/bindings/utils/_version_check.py
```diff
@@ -0,0 +1,56 @@
...
+
+    The check runs only once per process. Subsequent calls are no-ops.
+
+    The warning can be suppressed by setting the environment variable
+    ``CUDA_PYTHON_DISABLE_MAJOR_VERSION_WARNING=1``.
+    """
+    global _major_version_compatibility_checked
+    if _major_version_compatibility_checked:
+        return
+    _major_version_compatibility_checked = True
```
leofang: 1. need a lock + double check pattern
2. move `_major_version_compatibility_checked = True` to the code later because we could raise below

cuda_bindings/docs/source/environment_variables.rst
```diff
@@ -9,6 +9,8 @@ Runtime Environment Variables
 
 - ``CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM`` : When set to 1, the default stream is the per-thread default stream. When set to 0, the default stream is the legacy default stream. This defaults to 0, for the legacy default stream. See `Stream Synchronization Behavior <https://docs.nvidia.com/cuda/cuda-runtime-api/stream-sync-behavior.html>`_ for an explanation of the legacy and per-thread default streams.
 
+- ``CUDA_PYTHON_DISABLE_MAJOR_VERSION_WARNING`` : When set to 1, suppresses warnings about CUDA major version mismatches between ``cuda-bindings`` and the installed driver.
```
leofang, line 12: TBH I am not sure if this env var is really useful, because the added API is entirely opt-in?

cuda_bindings/tests/test_version_check.py
leofang: We should either mock it or restore whatever the original value was during teardown, I agree with Rob.

cuda_core/cuda/core/_device.pyx
```diff
@@ -1393,3 +1352,61 @@ class Device:
         """
         self._check_context_initialized()
         return GraphBuilder._init(stream=self.create_stream(), is_stream_owner=True)
+
+
+cdef inline void Device_ensure_cuda_initialized() except *:
```
leofang: We prefer `cdef inline int Device_ensure_cuda_initialized() except? -1:` to avoid an unconditional exception check (which Cython warns at build time).

cuda_core/cuda/core/_device.pyx
```diff
@@ -1393,3 +1352,61 @@ class Device:
...
+    if _is_cuInit is False:
+        with _lock, nogil:
+            HANDLE_RETURN(cydriver.cuInit(0))
+            _is_cuInit = True
+        try:
+            from cuda.bindings.utils import warn_if_cuda_major_version_mismatch
+        except ImportError:
+            pass
+        else:
+            warn_if_cuda_major_version_mismatch()
```
leofang, line 1390: We want to move imports like this to the top so as to avoid the little overheads in a hot loop. Please benchmark the constructor to see the perf diff before/after this PR. As per @mdboom such imports, despite they are cached, have a noticeable perf overhead (on the order that we care about).

leofang (changes requested):
Sorry for my late review.

cuda_core/cuda/core/_device.pyx
Andy-Jost: I agree with the general advice. Note that in this case, the block is guarded by `_is_cuInit` so the import is only attempted once. I decided to bury this import because 1) it would otherwise add clutter to the top-of-file import list, which human readers parse relatively often, 2) an import failure is ignored, so it can't hurt anything, and 3) there is no performance impact, since the import is only attempted once.

cuda_bindings/cuda/bindings/utils/_version_check.py
Andy-Jost: 1. Added a lock
2. This was intentional. If the code attempting to perform the check fails for some reason, we would never want to retry it.

cuda_bindings/docs/source/environment_variables.rst
Andy-Jost: The version check can be triggered indirectly from the end user's POV. For example, by using cuda-core. Without this, each library that calls the version-check function would need its own opt-out, and users might need to set them all. To me it makes sense to place the opt-out alongside the implementation and there is little/no cost.

cuda_core/cuda/core/_device.pyx
rwgk: I agree this really is fine. This is not in a hot loop.

The only change I'd make here, to make this industrial strength:

```
try:
    from cuda.bindings.utils import warn_if_cuda_major_version_mismatch
except ModuleNotFoundError:
    pass
except Exception as e:
    warnings.warn(
        f"Failed to run CUDA version check: {e}",
        RuntimeWarning,
    )
else:
    warn_if_cuda_major_version_mismatch()
```

This one-minute read has the rationale:

https://chatgpt.com/share/6968476c-2120-8008-a133-1bb4aabfb959

The key point:
___

If the only goal is “produce a warning”

Ironically, that makes silently swallowing all failures worse, not better.

A warning system that:

fails quietly

gives no hint it failed

…doesn’t actually help users.

rwgk (approved):

cuda_bindings/docs/source/environment_variables.rst
leofang: I insist that this is useless because this is listed in the cuda-bindings docs, not cuda-core. But cuda-bindings does not call this function anywhere, whereas cuda-core users don't necessarily look at cuda-bindings docs. Can we please create a new `environment_variables.rst` for cuda-core if we think this is helpful there?

cuda_core/cuda/core/_device.pyx
leofang: OK. FWIW after grabbing the lock we could check `_is_cuInit` again. 

I don't think the `Failed to run CUDA version check: ...` warning is actionable. We don't need to add it.

cuda_bindings/docs/source/environment_variables.rst
leofang: xref: https://github.com/NVIDIA/cuda-python/pull/1412#discussion_r2632529697

leofang: Even better: Document this env var in both docs.

Andy-Jost: Good point. The same argument applies to all cuda.bindings runtime variables, so I:

1. duplicated both `CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM` and `CUDA_PYTHON_DISABLE_MAJOR_VERSION_WARNING` in the cuda.core documentation; and
2. added a note referencing the cuda.bindings runtime variables and linking to the relevant documentation page.

PR #1413, opened 2025-12-19
[FEA] Add from_handle() methods for Kernel. 

### LINKED ISSUE #192: Implement `from_handle()` for all `cuda.core` objects
We wanted to do this but it seems so far we've only covered `Stream.from_handle`. We want to also cover these objects:
- `Program`
- `ObjectCode`
- `Kernel`
- `Buffer`

rparolin:
Fixes #192 

This PR implements from_handle() static methods for Kernel classes enabling interoperability with foreign CUDA handles.

✨ Key Changes

1. Kernel.from_handle()
2. Thread Safety Improvements
2.1. Added thread-safe lazy initialization using double-checked locking pattern
2.2. Refactored global state access through helper functions that ensure initialization

### DIFF
--- a/cuda_core/cuda/core/_module.py
+++ b/cuda_core/cuda/core/_module.py
@@ -7,7 +7,6 @@ import threading
 import weakref
 from collections import namedtuple
 from typing import Union
-from warnings import warn
 
 from cuda.core._device import Device
 from cuda.core._launch_config import LaunchConfig, _to_native_launch_config


cuda_core/cuda/core/_module.py
```diff
@@ -436,7 +488,43 @@ def occupancy(self) -> KernelOccupancy:
             self._occupancy = KernelOccupancy._init(self._handle)
         return self._occupancy
 
-    # TODO: implement from_handle()
+    @staticmethod
+    def from_handle(handle: int, mod: "ObjectCode" = None) -> "Kernel":
```
leofang, line 490: This would open a can of worms and we should be careful. Mainly because
1. we cover two sets of driver APIs (function/module vs kernel/library)
2. function/kernel are mostly interchangeable, but their parents (module/library) are not IIRC
3. bidirectional lookup cannot be done between function and module, but can be between kernel and library

We should discuss this in depth after the holidays.

rparolin: Happy New Year! 🎄

rparolin: @leofang updated based on our conversation

cuda_core/cuda/core/_module.py
```diff
@@ -27,35 +36,75 @@
...
+    _lazy_init()
+    return _kernel_ctypes
+
+
+def _get_backend_version():
+    """Get the backend version ("new" or "old") based on CUDA version.
+
+    Returns "new" for CUDA 12.0+ (uses cuLibrary API), "old" otherwise (uses cuModule API).
+    """
+    return "new" if (_get_py_major_ver() >= 12 and _get_driver_ver() >= 12000) else "old"
```
leofang: nit: let's also memoize this

cuda_core/cuda/core/_module.py
```diff
@@ -436,7 +488,43 @@ def occupancy(self) -> KernelOccupancy:
...
+        else:
+            # Use CUfunction for older versions
+            kernel_obj = driver.CUfunction(handle)
+
+        # If no module provided, create a placeholder
+        if mod is None:
+            # Create a placeholder ObjectCode that won't try to load anything
+            mod = ObjectCode._init(b"", "cubin")
+            # Set a dummy handle to prevent lazy loading
+            mod._handle = 1  # Non-null placeholder
```
leofang: - for `CUkernel`, we can do an inverse lookup by calling `cuKernelGetLibrary`, but we'll need a check for cuda-bindings>=12.5 and driver version >=12.5 (it's a new API)
- for `CUfunction`, there is no way to do an inverse lookup, so setting `_handle` = 1 might be OK (btw are we sure we won't use this pointer in any situation and it's just a placeholder?), but in practice we won't have any test coverage for this since we dropped support for CUDA 11. Might be a good idea to just kill the `CUfunction` branch.

cuda_core/tests/test_module.py
```diff
@@ -420,3 +420,90 @@ def test_module_serialization_roundtrip(get_saxpy_kernel_cubin):
...
+
+def test_kernel_from_handle(get_saxpy_kernel_cubin):
+    """Test Kernel.from_handle() with a valid handle"""
+    original_kernel, objcode = get_saxpy_kernel_cubin
+
+    # Get the handle from the original kernel
+    handle = int(original_kernel._handle)
+
+    # Create a new Kernel from the handle
+    kernel_from_handle = cuda.core._module.Kernel.from_handle(handle, objcode)
```
leofang: For all new tests added: We should access `Kernel` directly from `cuda.core`, instead of `cuda.core._module`. We already expose it to the main namespace (for typing/doc purposes).

I did not check if existing tests already use this pattern. If so we can either have another cleanup PR, or just fix it in this PR.

cuda_core/cuda/core/_module.py
rparolin: Done.

PR #1414, opened 2025-12-19
chore: move to default envs

cpcloud:
This PR replaces cu13 environments with `default`, allowing `pixi run test` to work everywhere in `cuda-python`, using cuda 13.x

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -7,15 +7,20 @@ channels = ["conda-forge"]
 platforms = ["linux-64", "linux-aarch64", "win-64"]
 
 [feature.cu12.dependencies]
+[feature.cu13.dependencies]
 
 [environments]
 cu12 = { features = ["cu12"] }
+cu13 = { features = ["cu13"] }
+
+[activation.env]
+PIXI_ENVIRONMENT_NAME = "${PIXI_ENVIRONMENT_NAME/default/cu13}"
 
 # Test Tasks
 # Runs tests across all sub-packages: pathfinder → bindings → core (dependency order)
-# Each sub-package has its own pixi.toml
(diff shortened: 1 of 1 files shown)


rparolin:
Does this mean that there is no longer a cu13 environment? Or did we made cu13 the default?  @cpcloud

cpcloud:
I got rid of cu13 to avoid some duplication, but I can easily bring it back.

rparolin:
> I got rid of cu13 to avoid some duplication, but I can easily bring it back.

I think we should bring it back so the name `cu13` exists. We will have to use it directly when `cu14` is the new default and we want to do `cu13` specific testing.  So I don't see any harm is having the name being available. 

Indeed we should make it the default so users aren't required to type it.  If users type nothing, `pixi run test` then they should get the default we centrally control which at the time of writing is `cu13`.

cpcloud:
@rparolin I brought back `cu13`. Would you mind reviewing again?

cuda_core/pixi.toml
```diff
@@ -10,18 +10,16 @@ preview = ["pixi-build"]
 [workspace.build-variants]
 python = ["3.10.*", "3.11.*", "3.12.*", "3.13.*", "3.14.*"]
 
-[dependencies]
```
rparolin, line 13: What was this even doing before? cuda-core was a dependency of itself? 🤔

pixi.toml
```diff
@@ -7,20 +7,15 @@ channels = ["conda-forge"]
 platforms = ["linux-64", "linux-aarch64", "win-64"]
 
 [feature.cu12.dependencies]
-[feature.cu13.dependencies]
 
 [environments]
 cu12 = { features = ["cu12"] }
-cu13 = { features = ["cu13"] }
```
rparolin, line 14: Don't we want to keep cu13 environment?

PR #1415, opened 2025-12-19
Backport #1411

mdboom:


cuda_pathfinder/cuda/pathfinder/__init__.py
```diff
@@ -1,10 +1,11 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+from cuda.pathfinder._version import __version__ as __version__
```
cpcloud: Not blocking, but the `X as X` seems redundant.

cpcloud (approved):

PR #1425, opened 2026-01-05
feat: improve structured dtype array support in `StridedMemoryView`

cpcloud:
# Improve Support for Structured Dtype Arrays in StridedMemoryView

## Summary
Extends `StridedMemoryView` to handle NumPy structured dtypes (e.g., `dtype([("field1", float64), ("field2", int64)])`) instead of only built-in numeric types.

## Changes

**`cuda/core/_memoryview.pyx`** (30 lines, +20 -10)
- Converted `_typestr2dtype` from static dict to function with lazy caching
- Converted `_typestr2itemsize` from static dict to function with lazy caching
- Falls back to `numpy.dtype(typestr)` for non-builtin types (e.g., structured dtypes)
- Renamed globals: `_typestr2dtype` → `_TYPESTR_TO_DTYPE`, `_typestr2itemsize` → `_TYPESTR_TO_ITEMSIZE`
- Removed TODO comment about built-in-only limitation (now resolved)

**`tests/test_utils.py`** (+17 lines)
- Added `test_struct_array()` validating structured dtype handling
- Tests array with `dtype=[("array1", np.float64), ("array2", np.int64)]`
- Verifies size, shape, and dtype preservation (note: full dtype info lost due to typestr encoding)

## Key Behavior
Structured dtypes now work but are represented as void types (e.g., `V16`) due to inherent lossiness in typestr encoding (`dtype([("a", "int")]).str == "V8"`).

### DIFF
--- a/cuda_core/cuda/core/_memoryview.pyx
+++ b/cuda_core/cuda/core/_memoryview.pyx
@@ -485,41 +485,14 @@ cdef StridedMemoryView view_as_dlpack(obj, stream_ptr, view=None):
     return buf
 
 
-_builtin_numeric_dtypes = [
-    numpy.dtype("uint8"),
-    numpy.dtype("uint16"),
-    numpy.dtype("uint32"),
-    numpy.dtype("uint64"),
-    numpy.dtype("int8"),
-    numpy.dtype("int16"),
-    numpy.dtype("int32"),
-    numpy.dtype("int64"),
-    numpy.dtype("float16"),
-    numpy.dtype("float32"),
-    numpy.dtype("float64"),
-    numpy.dtype("complex64"),
-    numpy.dtype("complex128"),
-    numpy.dtype("bool"),
-]
-# Doing it once to avoid repeated overhead
-_TYPESTR_TO_DTYPE = {dtype.str: dtype for dtype in _builtin_numeric_dtypes}
-_TYPESTR_TO_ITEMSIZE = {dtype.str: dtype.itemsize for dtype in _builtin_numeric_dtypes}
-
-cpdef object _typestr2dtype(str typestr):
-    if (dtype := _TYPESTR_TO_DTYPE.get(typestr)) is not None:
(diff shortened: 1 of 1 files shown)


cuda_core/tests/test_utils.py
```diff
@@ -430,3 +430,20 @@ def test_view_zero_size_array(api, shape, dtype):
...
+    y = cp.empty(2, dtype=x.dtype)
+    y.set(x)
+
+    smv = StridedMemoryView.from_cuda_array_interface(y, stream_ptr=0)
+    assert smv.size * smv.dtype.itemsize == x.nbytes
+    assert smv.size == x.size
+    assert smv.shape == x.shape
+    # full dtype information doesn't seem to be preserved due to use of type strings,
+    # which are lossy, e.g., dtype([("a", "int")]).str == "V8"
+    assert smv.dtype == np.dtype(f"V{x.itemsize}")
```
mdboom, line 462: Mostly for paranoia reasons and because it's something that dlpack can't do, maybe test for strides that are not a multiple of itemsize?  The fact that both elements of this array are the same size and 8-byte aligned my hide some potential problems.  For example:

```
x = np.array([(1.0, 2), (2.0, 3)], dtype=[("array1", np.float64), ("array2", np.int32)])

y = cp.empty(2, dtype=x.dtype)
y.set(x)

y = y["array1"]
...
```

cpcloud: `y["array1"]` doesn't work, seems like that's unsupported in cupy.

I also can't seem to slice a structured array, so I'm not entirely sure how to construct such an array with existing cupy APIs.

mdboom: Got it.  If it's not supported by cupy, then probably not important here.

leofang: it will be supported by CuPy v14.0.0

PR #1428, opened 2026-01-06
feat: allow constructing SMV from numpy arrays

cpcloud:
## Summary

Add the `StridedMemoryView.from_array_interface()` class method to support
creating memory views from objects implementing the NumPy Array Interface
protocol (`__array_interface__`).

## Changes

**New API** (`cuda_core/cuda/core/_memoryview.pyx:156-163`)
- Added `StridedMemoryView.from_array_interface(obj)` classmethod for creating
  memory views from array interface objects
- Implemented `view_as_array_interface()` helper function that:
  - Validates array interface version (requires v3+)
  - Rejects masked arrays per existing behavior in CAI APIs
  - Extracts pointer, `shape`, `dtype`, `strides`, and `readonly` flag from `__array_interface__` dict
  - Marks buffers as not device-accessible

**Test Coverage** (`cuda_core/tests/test_utils.py:464-515`)
- 25+ parameterized test cases covering:
  - Multiple dtypes (int8-64, uint8-64, float32/64, complex64/128)
  - Various array shapes (scalar, 1D, 2D, 3D, empty)
  - Memory layouts (C-contiguous, Fortran-contiguous, strided)
  - Structured dtypes
- Validates: size, dtype, shape, pointer address, device accessibility, readonly flag, strides

---

~~**Note:** This PR builds on #1425 (structured dtype support). Only review the latest commit `dd893ce` - previous commits are from the base PR.~~

### DIFF
--- a/cuda_core/cuda/core/_memoryview.pyx
+++ b/cuda_core/cuda/core/_memoryview.pyx
@@ -139,6 +139,8 @@ cdef class StridedMemoryView:
     def from_dlpack(cls, obj: object, stream_ptr: int | None=None) -> StridedMemoryView:
         cdef StridedMemoryView buf
         with warnings.catch_warnings():
+            # ignore the warning triggered by calling the constructor
+            # inside the library we're allowed to do this
             warnings.simplefilter("ignore", DeprecationWarning)
             buf = cls()
         view_as_dlpack(obj, stream_ptr, buf)


cuda_core/cuda/core/_memoryview.pyx
```diff
@@ -153,6 +153,15 @@ cdef class StridedMemoryView:
         view_as_cai(obj, stream_ptr, buf)
         return buf
 
+    @classmethod
+    def from_array_interface(cls, obj: object) -> StridedMemoryView:
+        cdef StridedMemoryView buf
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
```
mdboom: Maybe leave a comment here about what warning we are trying to suppress here?  Could we have a more specific filter so this won't suppress future, useful, warning?

cuda_core/tests/test_utils.py
```diff
@@ -460,3 +461,56 @@ def test_struct_array():
...
+    assert smv.dtype == np.dtype(expected_dtype)
+    assert smv.shape == x.shape
+    assert smv.ptr == x.ctypes.data
+    assert smv.device_id == init_cuda.device_id
+    assert smv.is_device_accessible is False
+    assert smv.exporting_obj is x
+    assert smv.readonly is not x.flags.writeable
+    # Check strides
+    strides_in_counts = convert_strides_to_counts(x.strides, x.dtype.itemsize)
+    assert (x.flags.c_contiguous and smv.strides is None) or smv.strides == strides_in_counts
```
mdboom, line 516: Since strides in SMV are defined as counts of `itemsize`, they are a subset of what Numpy can handle and we should make sure arrays we can't handle are properly rejected.  It seems that this is /mostly/ the case, but due to the lazy construction, the exception happens a little later than ideal.  For example:

```
def test_from_array_interface_exception(init_cuda):
    # Create an array with strides that aren't a multiple of itemsize
    x = np.array([(1, 2.0), (3, 4.0)], dtype=[("a", "i4"), ("b", "f8")])["b"]
    # with pytest.raises(ValueError) as e:
    smv = StridedMemoryView.from_array_interface(x)
    with pytest.raises(ValueError) as e:
        smv.strides
    assert "strides must be divisible by itemsize" in str(e)
```

In user code, this could make it harder for the user to find the source of the error.

But given this is a latent bug that I think also applies to CAI, and not introduced by this PR, I'd be fine with (a) adding this test above, (b) creating a follow-up issue to detect this case during construction.

mdboom (changes requested):

cuda_core/cuda/core/_memoryview.pyx
cpcloud: Yep, I'll see about adding a more specific warning.

cuda_core/tests/test_utils.py
cpcloud: Makes sense. I'll add the test and create a follow-up issue to track raising an exception on construction.

cuda_core/cuda/core/_memoryview.pyx
cpcloud: Done. Added a comment and made the warning specific.

cuda_core/tests/test_utils.py
cpcloud: Done. Added the test. Created https://github.com/NVIDIA/cuda-python/issues/1429 to track failing earlier.

leofang:
Ideally we want to always use `from_dlpack` to cover NumPy arrays. @cpcloud is there a reason that we need to accept array interfaces?

cpcloud:
Yes, structured arrays, strings, bytes, and whatever other things DLPack doesn't support.

leofang:
Ah right, thx for reminder...

PR #1448, opened 2026-01-09
cuda.core.system: Add APIs related to events

mdboom:
Prerequsite: #1446

## Description

- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #1449, opened 2026-01-09
cuda.bindings._nvml changes for cuda.core.system work

mdboom:
As an artifact of how our CI works, any changes required to `cuda.bindings` required by changes to `cuda.core` must be merged first, and not as part of the same PR.

I think it's worth working on that problem independently, but in the meantime, I am filing this PR as a prerequisite to #1448 and #1446.


mdboom:
Actually, maybe I don't need to do this separately, now that I think about it...  Closing for now.

PR #1454, opened 2026-01-10
Add `__version__` validation: import-time assertions, unit tests, PyPI fallback prevention, CI hardening

rwgk:
This PR adds safety mechanisms to prevent version-related issues during builds and ensures version detection from git tags is working as intended.

## Version Number Validation

**Problem**: `setuptools-scm` silently falls back to version `0.1.x` when git tags are unavailable (e.g., due to shallow clones), which can lead to incorrect version detection and unexpected dependency resolution. Silent failures in the procedure producing `__version__` can lead to highly confusing behavior and potentially invalidate elaborate QA testing.

**Solution**: We implement a two-layer defense strategy to catch invalid versions at multiple stages:

### First Line of Defense: Import-Time Assertions

Fail-fast assertions are added immediately after importing `__version__` in all three package `__init__.py` files:

- `cuda_bindings/cuda/bindings/__init__.py`
- `cuda_core/cuda/core/__init__.py`
- `cuda_pathfinder/cuda/pathfinder/__init__.py`

Each file includes a minimal one-liner assertion:
```python
assert tuple(int(_) for _ in __version__.split(".")[:2]) > (0, 1), "FATAL: invalid __version__"
```

This ensures that any attempt to import a package with an invalid version (e.g., `0.1.dev...`) fails immediately with a clear error, preventing the package from being used at all. The assertion checks that `major.minor > (0, 1)`, which is sufficient since all three packages are already at higher versions.

### Second Line of Defense: Unit Tests

As a backup, we also implement late-stage detection via regular unit tests that validate version numbers after installation:

- Adds `validate_version_number()` function to `cuda_python_test_helpers` for centralized validation logic
- Creates minimal `test_version_number.py` files in `cuda_bindings`, `cuda_core`, and `cuda_pathfinder` that import the version and call the validation function
- Tests are split into separate functions (`test_bindings_version`, `test_core_version`, `test_pathfinder_version`) so all invalid versions are reported in a single test run
- Each test suite validates its own package version plus dependency versions (e.g., `cuda_bindings` tests check both `cuda-bindings` and `cuda-pathfinder` versions)
- Provides clear error messages explaining the issue without referencing setuptools-scm internals

The unit tests run during the test phase and provide explicit test coverage with clearer error messages in CI logs.

### Why Two Layers?

While the import-time assertions provide immediate feedback and prevent invalid packages from being imported, we maintain the unit tests as a second line of defense because:

1. **Redundancy**: If the assertions somehow fail to catch an issue (e.g., due to import path quirks or edge cases), the unit tests provide a backup check
2. **Explicit Test Coverage**: Unit tests make version validation an explicit, testable requirement rather than an implicit assertion
3. **CI Visibility**: Test failures in CI logs are more visible and easier to debug than import-time assertion failures
4. **Defense in Depth**: Silent failures in the procedure producing `__version__` can lead to highly confusing behavior and potentially invalidate elaborate QA testing. Multiple detection points reduce the risk of invalid versions going undetected.

### CI Workflow Hardening

To ensure version validation works correctly in CI environments, we've hardened the test workflows:

- **Intentional Shallow Clone**: Test workflows explicitly use `fetch-depth: 1` (the default) with a comment emphasizing that shallow cloning is intentional. This ensures we're testing wheel installation without full git history, which is the correct behavior for testing pre-built artifacts.

- **Wheel-Only Installation**: The "Ensure cuda-python installable" step uses `--only-binary=:all:` to ensure we only test wheels, never build from source. This prevents pip from building packages from source when source code is present, which could lead to version issues in shallow clones.

- **Immediate Version Verification**: After installing `cuda-python`, a new "Verify installed package versions" step imports `cuda.pathfinder` and `cuda.bindings` to trigger the `__version__` assertions immediately. This provides early detection of invalid versions right after installation, before any tests run.

These changes ensure that:
- We're testing the actual wheel artifacts, not building from source
- Version issues are caught as early as possible in the CI pipeline
- The test workflows are resilient even with shallow clones

### Why Not Early-Stage Detection?

We initially attempted early-stage detection in build hooks (`build_wheel`, `build_editable`) to catch fallback versions during the build process. However, this approach proved too fragile:

1. **Timing Issues**: `_version.py` files are written by `setuptools-scm` during `prepare_metadata_for_build_wheel`, but validation needs to run at the right time in the build process. Attempting to validate too early results in "file not found" errors, while validating too late allows builds to complete with invalid versions.

2. **Shallow Clone Handling**: When `setuptools-scm` detects a shallow clone, it bypasses `git_describe_command` entirely and falls back to `0.0` or `0.1.x` versions before our validation can run. This makes build-time detection unreliable in CI environments that use shallow clones.

3. **Complexity**: The build hook approach required careful coordination between PEP 517 hooks (`prepare_metadata_for_build_wheel`, `build_wheel`) and custom validation logic, making it error-prone and difficult to maintain.

Given these challenges, we decided to use the simpler and more certain approach: import-time assertions for immediate feedback, unit tests for explicit coverage, and CI workflow hardening to ensure the validation works correctly in all environments.

## PyPI Fallback Prevention

**Problem**: During isolated (PEP 517) builds, a just-built `cuda-bindings` installation could be incorrectly replaced with a PyPI wheel if the installed version didn't match expectations.

**Solution**: Enhanced `cuda_core/build_hooks.py` to:
- Check installed `cuda-bindings` version using direct import of `cuda.bindings._version` (note that `importlib.metadata` cannot be used in isolated environments)
- Detect editable installs by checking if the `_version.py` file path is within the repository root
- Prevent replacement of editable installs
- Ensure version compatibility: if `cuda-bindings` is installed (non-editable) and its major version doesn't match the CUDA major version, raise an exception

This prevents accidental installation of incompatible `cuda-bindings` versions from PyPI during builds.

___

Piggy-backed:

## Import Sorting Fix

**Problem**: Ruff's import sorting (`I001`) was inconsistently reordering the `_version` import in `cuda_pathfinder/cuda/pathfinder/__init__.py`, depending on whether `_version.py` exists or not (e.g., after `git clean -fdx`).

**Solution**: Added `# isort: skip` directive to the `_version` import line to prevent ruff from moving it. This ensures consistent import ordering regardless of build state.

## Note on setuptools-scm RuntimeWarning

We've observed `RuntimeWarning`s from `setuptools-scm` that incorrectly display package versions instead of `setuptools` versions (e.g., `ERROR: setuptools==0.5.1.dev20+gf8dddb370 is used in combination with setuptools-scm>=8.x`). This appears to be a known issue in `setuptools-scm` when using custom build backends (see [setuptools-scm issue #1192](https://github.com/pypa/setuptools-scm/issues/1192)). A minimal reproducer has been created at [github.com/rwgk/setuptools-scm-issue-1192](https://github.com/rwgk/setuptools-scm-issue-1192).

These warnings don't affect functionality but are noisy. They occur on both `main` and this branch.

### DIFF
--- a/cuda_core/tests/helpers/__init__.py
+++ b/cuda_core/tests/helpers/__init__.py
@@ -9,14 +9,6 @@ from typing import Union
 
 from cuda.core._utils.cuda_utils import handle_return
 
-try:
-    from cuda_python_test_helpers import *  # noqa: F403
-except ModuleNotFoundError:
-    # Import shared platform helpers for tests across repos
-    sys.path.insert(0, str(pathlib.Path(__file__).resolve().parents[3] / "cuda_python_test_helpers"))
-    from cuda_python_test_helpers import *  # noqa: F403
-
-
 CUDA_PATH = os.environ.get("CUDA_PATH")
 CUDA_INCLUDE_PATH = None
 CCCL_INCLUDE_PATHS = None
@@ -28,6 +20,14 @@ if CUDA_PATH is not None:
         path = os.path.join(path, "cccl")
         if os.path.isdir(path):
             CCCL_INCLUDE_PATHS = (path,) + CCCL_INCLUDE_PATHS
+
+
+try:
+    from cuda_python_test_helpers import *  # noqa: F403
+except ModuleNotFoundError:
+    # Import shared platform helpers for tests across repos
+    sys.path.insert(0, str(pathlib.Path(__file__).resolve().parents[3] / "cuda_python_test_helpers"))
+    from cuda_python_test_helpers import *  # noqa: F403
 
 
 @functools.cache
--- a/cuda_core/cuda/core/system/_device.pyx
+++ b/cuda_core/cuda/core/system/_device.pyx
@@ -15,6 +15,7 @@ from ._nvml_context cimport initialize
 include "_device_utils.pxi"
 
 
+BrandType = nvml.BrandType
 FieldId = nvml.FieldId
 
 
@@ -174,6 +175,77 @@ cdef class PciInfo:
         return self._pci_info.pci_device_id >> 16
 
 
+cdef class DeviceAttributes:
+    """
+    Various device attributes.
+    """
+    def __init__(self, attributes: nvml.DeviceAttributes):
+        self._attributes = attributes
+
+    @property
+    def multiprocessor_count(self) -> int:
+        """
+        The streaming multiprocessor count
+        """
+        return self._attributes.multiprocessor_count
+
+    @property
+    def shared_copy_engine_count(self) -> int:
+        """
+        The shared copy engine count
+        """
+        return self._attributes.shared_copy_engine_count
+
+    @property
+    def shared_decoder_count(self) -> int:
+        """
+        The shared decoder engine count
+        """
+        return self._attributes.shared_decoder_count
+
+    @property
+    def shared_encoder_count(self) -> int:
+        """
+        The shared encoder engine count
+        """
+        return self._attributes.shared_encoder_count
+
+    @property
+    def shared_jpeg_count(self) -> int:
+        """
+        The shared JPEG engine count
+        """
+        return self._attributes.shared_jpeg_count
+
+    @property
+    def shared_ofa_count(self) -> int:
+        """
+        The shared optical flow accelerator (OFA) engine count
+        """
+        return self._attributes.shared_ofa_count
+
+    @property
+    def gpu_instance_slice_count(self) -> int:
+        """
+        The GPU instance slice count
+        """
+        return self._attributes.gpu_instance_slice_count
+
+    @property
+    def compute_instance_slice_count(self) -> int:
+        """
+        The compute instance slice count
+        """
+        return self._attributes.compute_instance_slice_count
+
+    @property
+    def memory_size_mb(self) -> int:
+        """
+        Device memory size in MiB
+        """
+        return self._attributes.memory_size_mb
+
+
 cdef class FieldValue:
     """
     Represents the data from a single field value.
@@ -317,39 +389,56 @@ cdef class Device:
     about devices and their topology, as provided by the NVIDIA Management
     Library (NVML).  To use CUDA with a device, use :class:`cuda.core.Device`.
 
+    Creating a device instance causes NVML to initialize the target GPU.
+    NVML may initialize additional GPUs if the target GPU is an SLI slave.
+
     Parameters
     ----------
     index: int, optional
-        Integer representing the CUDA device index to get a handle to.
+        Integer representing the CUDA device index to get a handle to.  Valid
+        values are between ``0`` and ``cuda.core.system.get_num_devices() - 1``.
+
+        The order in which devices are enumerated has no guarantees of
+        consistency between reboots.  For that reason, it is recommended that
+        devices are looked up by their PCI ids or UUID.
+
     uuid: bytes or str, optional
         UUID of a CUDA device to get a handle to.
+
+    pci_bus_id: bytes or str, optional
+        PCI bus ID of a CUDA device to get a handle to.
 
     Raises
     ------
     ValueError
-        If neither `index` nor `uuid` are specified or if both are specified.
+        If anything other than a single `index`, `uuid` or `pci_bus_id` are specified.
     """
 
     cdef intptr_t _handle
 
-    def __init__(self, index: int | None = None, uuid: bytes | str | None = None):
+    def __init__(self, index: int | None = None, uuid: bytes | str | None = None, pci_bus_id: bytes | str | None = None):
         initialize()
 
-        if index is not None and uuid is not None:
-            raise ValueError("Handle requires only one of either device `index` or `uuid`.")
-        if index is None and uuid is None:
-            raise ValueError("Handle requires either a device `index` or `uuid`.")
+        args = [index, uuid, pci_bus_id]
+        arg_count = sum(x is not None for x in args)
+
+        if arg_count > 1:
+            raise ValueError("Handle requires only one of either device `index`, `uuid` or `pci_bus_id`.")
+        if arg_count == 0:
+            raise ValueError("Handle requires either a device `index`, `uuid` or `pci_bus_id`.")
 
         if index is not None:
             self._handle = nvml.device_get_handle_by_index_v2(index)
-        else:
+        elif uuid is not None:
             if isinstance(uuid, bytes):
                 uuid = uuid.decode("ascii")
             self._handle = nvml.device_get_handle_by_uuid(uuid)
-
-    @property
-    def handle(self) -> int:
-        return self._handle
+        elif pci_bus_id is not None:
+            if isinstance(pci_bus_id, bytes):
+                pci_bus_id = pci_bus_id.decode("ascii")
+            self._handle = nvml.device_get_handle_by_pci_bus_id_v2(pci_bus_id)
+        else:
+            raise ValueError("Error parsing arguments")
 
     @classmethod
     def get_all_devices(cls) -> Iterable[Device]:
@@ -428,6 +517,24 @@ cdef class Device:
         return nvml.device_get_name(self._handle)
 
     @property
+    def brand(self) -> BrandType:
+        """
+        Brand of the device
+        """
+        return BrandType(nvml.device_get_brand(self._handle))
+
+    @property
+    def index(self) -> int:
+        """
+        The NVML index of this device.
+
+        The order in which NVML enumerates devices has no guarantees of
+        consistency between reboots. For that reason it is recommended that
+        devices be looked up by their PCI ids or GPU UUID.
+        """
+        return nvml.device_get_index(self._handle)
+
+    @property
     def pci_info(self) -> PciInfo:
         """
         The PCI attributes of this device.
@@ -450,6 +557,39 @@ cdef class Device:
         board serial identifier.
         """
         return nvml.device_get_uuid(self._handle)
+
+    @property
+    def attributes(self) -> DeviceAttributes:
+        """
+        Get various device attributes.
+
+        For Ampere™ or newer fully supported devices.  Only available on Linux
+        systems.
+        """
+        return DeviceAttributes(nvml.device_get_attributes_v2(self._handle))
+
+    @property
+    def is_c2c_mode_enabled(self) -> bool:
+        """
+        Whether the C2C (Chip-to-Chip) mode is enabled for this device.
+        """
+        return bool(nvml.device_get_c2c_mode_info_v(self._handle).is_c2c_enabled)
+
+    @property
+    def persistence_mode_enabled(self) -> bool:
+        """
+        Whether persistence mode is enabled for this device.
+
+        For Linux only.
+        """
+        return nvml.device_get_persistence_mode(self._handle) == nvml.EnableState.FEATURE_ENABLED
+
+    @persistence_mode_enabled.setter
+    def persistence_mode_enabled(self, enabled: bool) -> None:
+        nvml.device_set_persistence_mode(
+            self._handle,
+            nvml.EnableState.FEATURE_ENABLED if enabled else nvml.EnableState.FEATURE_DISABLED
+        )
 
     def get_field_values(self, field_ids: list[int | tuple[int, int]]) -> FieldValues:
         """
@@ -494,8 +634,10 @@ cdef class Device:
 
 __all__ = [
     "BAR1MemoryInfo",
+    "BrandType",
     "Device",
     "DeviceArchitecture",
+    "DeviceAttributes",
     "FieldId",
     "FieldValue",
     "FieldValues",
--- a/cuda_core/docs/source/api.rst
+++ b/cuda_core/docs/source/api.rst
@@ -75,6 +75,7 @@ CUDA system information and NVIDIA Management Library (NVML)
 
    system.get_driver_version
    system.get_driver_version_full
+   system.get_driver_branch
    system.get_num_devices
    system.get_nvml_version
    system.get_process_name
@@ -82,14 +83,15 @@ CUDA system information and NVIDIA Management Library (NVML)
    :template: autosummary/cyclass.rst
 
    system.Device
+   system.BAR1MemoryInfo
+   system.BrandType
    system.DeviceArchitecture
+   system.DeviceAttributes
    system.FieldId
    system.FieldValue
    system.FieldValues
    system.MemoryInfo
-   system.BAR1MemoryInfo
    system.PciInfo
-
 
 .. module:: cuda.core.utils
 
--- a/cuda_core/tests/system/test_system_device.py
+++ b/cuda_core/tests/system/test_system_device.py
@@ -13,6 +13,7 @@ import os
 import re
 import sys
 
+import helpers
 import pytest
 from cuda.core import system
 
@@ -25,11 +26,6 @@ if system.CUDA_BINDINGS_NVML_IS_COMPATIBLE:
 def check_gpu_available():
     if not system.CUDA_BINDINGS_NVML_IS_COMPATIBLE or system.get_num_devices() == 0:
         pytest.skip("No GPUs available to run device tests", allow_module_level=True)
-
-
-def test_device_index_handle():
-    for device in system.Device.get_all_devices():
-        assert isinstance(device.handle, int)
 
 
 def test_device_architecture():
@@ -189,6 +185,77 @@ def test_unpack_bitmask(params):
 def test_unpack_bitmask_single_value():
     with pytest.raises(TypeError):
         _device._unpack_bitmask(1)
+
+
+def test_device_brand():
+    for device in system.Device.get_all_devices():
+        brand = device.brand
+        assert isinstance(brand, system.BrandType)
+        assert isinstance(brand.name, str)
+        assert isinstance(brand.value, int)
+
+
+def test_device_pci_bus_id():
+    for device in system.Device.get_all_devices():
+        pci_bus_id = device.pci_info.bus_id
+        assert isinstance(pci_bus_id, str)
+
+        new_device = system.Device(pci_bus_id=device.pci_info.bus_id)
+        assert new_device.index == device.index
+
+
+@pytest.mark.skipif(helpers.IS_WSL or helpers.IS_WINDOWS, reason="Device attributes not supported on WSL or Windows")
+def test_device_attributes():
+    skip_reasons = []
+
+    for device in system.Device.get_all_devices():
+        try:
+            attributes = device.attributes
+        except system.NotSupportedError:
+            skip_reasons.append(f"Device attributes not supported on '{device.name}'")
+            continue
+        assert isinstance(attributes, system.DeviceAttributes)
+
+        assert isinstance(attributes.multiprocessor_count, int)
+        assert attributes.multiprocessor_count > 0
+
+        assert isinstance(attributes.shared_copy_engine_count, int)
+        assert isinstance(attributes.shared_decoder_count, int)
+        assert isinstance(attributes.shared_encoder_count, int)
+        assert isinstance(attributes.shared_jpeg_count, int)
+        assert isinstance(attributes.shared_ofa_count, int)
+        assert isinstance(attributes.gpu_instance_slice_count, int)
+        assert isinstance(attributes.compute_instance_slice_count, int)
+        assert isinstance(attributes.memory_size_mb, int)
+        assert attributes.memory_size_mb > 0
+
+    if skip_reasons:
+        pytest.skip(" ; ".join(skip_reasons))
+
+
+def test_c2c_mode_enabled():
+    skip_reasons = set()
+    for device in system.Device.get_all_devices():
+        try:
+            is_enabled = device.is_c2c_mode_enabled
+        except nvml.NotSupportedError:
+            skip_reasons.add(f"C2C mode info not supported on {device}")
+        else:
+            assert isinstance(is_enabled, bool)
+    if skip_reasons:
+        pytest.skip(" ; ".join(skip_reasons))
+
+
+@pytest.mark.skipif(helpers.IS_WSL or helpers.IS_WINDOWS, reason="Persistence mode not supported on WSL or Windows")
+def test_persistence_mode_enabled():
+    for device in system.Device.get_all_devices():
+        is_enabled = device.persistence_mode_enabled
+        assert isinstance(is_enabled, bool)
+        try:
+            device.persistence_mode_enabled = False
+            assert device.persistence_mode_enabled is False
+        finally:
+            device.persistence_mode_enabled = is_enabled
 
 
 def test_field_values():


rwgk:
# setuptools-scm Shallow Clone Fallback Issue (exists already on `main`)

## Summary

During CI testing, we discovered that `cuda-pathfinder` is being built with fallback version `0.1.dev1+g1b2e4c088` instead of the expected `1.3.4.dev77+g1b2e4c088` in some Windows test runs. This occurs when setuptools-scm detects a shallow Git clone and silently falls back to the default version without calling our `git_describe_wrapper.py`.

## Root Cause

1. **Shallow Clone Detection**: setuptools-scm detects shallow clones early in its version detection process
2. **Early Fallback**: When a shallow clone is detected, setuptools-scm bypasses `git_describe_command` entirely and falls back to `0.0` or `0.1.x` versions
3. **Wrapper Never Called**: Our `git_describe_wrapper.py` is never executed because setuptools-scm decides to skip version detection before attempting to call it

## Evidence

### Test Results

The fallback version `0.1.dev1+g1b2e4c088` was detected in **18 Windows test files** (all Windows tests):

- `Test_win-64___py3.10__12.9.1__wheels__rtx2080__WDDM_.txt`
- `Test_win-64___py3.10__13.0.2__local__rtxpro6000__TCC_.txt`
- `Test_win-64___py3.10__13.1.0__local__rtxpro6000__TCC_.txt`
- `Test_win-64___py3.11__12.9.1__local__v100__MCDM_.txt`
- `Test_win-64___py3.11__13.0.2__wheels__rtx4090__WDDM_.txt`
- `Test_win-64___py3.11__13.1.0__wheels__rtx4090__WDDM_.txt`
- `Test_win-64___py3.12__12.9.1__wheels__l4__MCDM_.txt`
- `Test_win-64___py3.12__13.0.2__local__a100__TCC_.txt`
- `Test_win-64___py3.12__13.1.0__local__a100__TCC_.txt`
- `Test_win-64___py3.13__12.9.1__local__l4__TCC_.txt`
- `Test_win-64___py3.13__13.0.2__wheels__rtxpro6000__MCDM_.txt`
- `Test_win-64___py3.13__13.1.0__wheels__rtxpro6000__MCDM_.txt`
- `Test_win-64___py3.14__12.9.1__wheels__v100__TCC_.txt`
- `Test_win-64___py3.14__13.0.2__local__l4__MCDM_.txt`
- `Test_win-64___py3.14__13.1.0__local__l4__MCDM_.txt`
- `Test_win-64___py3.14t__12.9.1__local__l4__TCC_.txt`
- `Test_win-64___py3.14t__13.0.2__wheels__a100__MCDM_.txt`
- `Test_win-64___py3.14t__13.1.0__wheels__a100__MCDM_.txt`

**Notes**:

* All affected tests are Windows-based. Linux tests appear to have proper version numbers.

* We found `0.1.dev` also in the same 18 CI logs from the [latest run on `main`](https://github.com/NVIDIA/cuda-python/actions/runs/20871705999)

### Log Evidence

From `Test_win-64___py3.14t__12.9.1__local__l4__TCC_.txt`:

```
2026-01-11T03:06:35.1995098Z   C:\Windows\Temp\pip-build-env-qw2xma_m\overlay\Lib\site-packages\setuptools_scm\git.py:202: UserWarning: "C:\actions-runner\_work\cuda-python\cuda-python" is shallow and may cause errors
2026-01-11T03:06:35.1995894Z     warnings.warn(f'"{wd.path}" is shallow and may cause errors')
2026-01-11T03:06:36.0998424Z   Created wheel for cuda-pathfinder: filename=cuda_pathfinder-0.1.dev1+g1b2e4c088-py3-none-any.whl
2026-01-11T03:06:54.4353416Z cuda-bindings 12.9.6.dev2+g563cd83db requires cuda-pathfinder~=1.1, but you have cuda-pathfinder 0.1.dev1+g1b2e4c088 which is incompatible.
```

### Local Testing

We tested locally with a shallow clone and confirmed:

1. **Wrapper Not Called**: Even with `git_describe_command` configured, the wrapper script is never executed
2. **Silent Fallback**: setuptools-scm detects the shallow clone, warns about it, but proceeds with fallback version `0.0` or `0.1.x`
3. **No Error**: The build succeeds with the wrong version, causing dependency conflicts

## setuptools-scm Behavior

From setuptools-scm source code (`git.py`):

```python
def version_from_describe(...):
    if describe_command is not None:
        # Only called if setuptools-scm decides to attempt version detection
        describe_res = _run(describe_command, wd.path)
    else:
        describe_res = wd.default_describe()
    
    return describe_res.parse_success(parse=parse_describe)

# In _git_parse_inner:
version = version_from_describe(wd, config, describe_command)

if version is None:
    # Falls back to 0.0 or configured fallback_version
    tag = config.version_cls(config.fallback_version or "0.0")
    # ... creates version with fallback
```

**Key Finding**: When setuptools-scm detects a shallow clone early (via `is_shallow()`), it may skip calling `git_describe_command` entirely and go straight to the fallback.

## Impact

1. **CI Testing**: Windows CI runs are building `cuda-pathfinder` with incorrect versions
2. **Dependency Conflicts**: The fallback version `0.1.dev1` doesn't satisfy `cuda-bindings` requirement `~=1.1`
3. **SWQA Team Risk**: If SWQA uses shallow clones, they'll encounter the same issue
4. **Silent Failure**: The build succeeds but with wrong version, making it hard to detect

## Attempted Solutions

### 1. git_describe_wrapper.py Enhancement

We enhanced `scripts/git_describe_wrapper.py` to detect shallow clones proactively:

```python
# Check if repository is shallow
result = subprocess.run(
    ["git", "rev-parse", "--is-shallow-repository"],
    capture_output=True,
    text=True,
    timeout=5,
)
if result.returncode == 0 and result.stdout.strip() == "true":
    print("ERROR: Repository is a shallow clone.", file=sys.stderr)
    sys.exit(1)
```

**Result**: ❌ **Doesn't work** - The wrapper is never called when setuptools-scm detects a shallow clone.

### 2. setuptools-scm `fail_on_shallow` Option

setuptools-scm provides a `pre_parse = "fail_on_shallow"` option that should fail builds on shallow clones.

**Result**: ❌ **Didn't work in our tests** - May require different configuration or setuptools-scm version.

## References

- setuptools-scm source: https://github.com/pypa/setuptools-scm
- Git shallow clone: https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---depthltdepthgt

rwgk:
# CI Logs Analysis: PR #1454 vs Main Branch

## Summary

After carefully analyzing the CI logs from PR #1454 ([`/wrk/logs_20909188905`](https://github.com/NVIDIA/cuda-python/actions/runs/20909188905/job/60068643781?pr=1454)) and comparing them to the main branch logs ([`/wrk/main_logs_20871705999`](https://github.com/NVIDIA/cuda-python/actions/runs/20871705999)), I found that:

1. **0.1.dev versions still appear** in both PR and main branch logs (18 Windows test files in each)
2. **However, they are caught and replaced** before tests run
3. **All version tests pass** - no invalid versions reach the actual test execution
4. **The root cause is a workflow bug** in the Windows test workflow

## Detailed Findings

### 0.1.dev Versions Still Appear

**PR Logs**: Found `cuda-pathfinder-0.1.dev1+g233447b1b` in 18 Windows test files  
**Main Logs**: Found `cuda-pathfinder-0.1.dev1+gacc78f7c0` in 18 Windows test files

The same number of occurrences suggests this is a pre-existing issue, not something introduced by the PR.

### Where They Come From

The bad versions are built during the "Install cuda.pathfinder extra wheels for testing" step. The issue is in the Windows workflow:

**Windows workflow** (`.github/workflows/test-wheel-windows.yml` line 259):
```bash
pip install --only-binary=:all: -v . --group "test-cu${TEST_CUDA_MAJOR}"
```

**Linux workflow** (`.github/workflows/test-wheel-linux.yml` line 292):
```bash
pip install --only-binary=:all: -v ./*.whl --group "test-cu${TEST_CUDA_MAJOR}"
```

**The Problem**: When pip processes `.` (a directory), it needs to prepare metadata to determine dependencies, which triggers `setuptools-scm` to build from source (in a shallow clone, resulting in `0.1.dev1`). The `--only-binary=:all:` flag doesn't prevent metadata preparation - it only prevents installing from source distributions.

When pip processes `./*.whl`, it installs directly from the wheel file without needing to prepare metadata, avoiding the source build entirely.

### What Happens Next

The sequence in affected Windows tests:

1. **Good version installed first**: `cuda-pathfinder-1.3.4.dev85+g233447b1b` (from wheel artifact)
2. **Version verification passes**: The "Verify installed package versions" step successfully imports `cuda.pathfinder` and `cuda.bindings` (lines 15404-15405)
3. **Bad version built**: During "Install cuda.pathfinder extra wheels", pip builds `0.1.dev1` from source due to the `.` directory issue
4. **Bad version installed**: `Successfully installed cuda-pathfinder-0.1.dev1+g233447b1b` (line 15677)
5. **Bad version replaced**: `Uninstalling cuda-pathfinder-0.1.dev1+g233447b1b` then `Successfully installed cuda-pathfinder-1.3.4.dev85+g233447b1b` (lines 15758-15760)
6. **Tests run with good version**: All `test_version_number` tests pass

### Version Tests Status

**All version tests pass** in both PR and main branch logs:
- `test_bindings_version PASSED`
- `test_core_version PASSED`
- `test_pathfinder_version PASSED`

This confirms that:
- The import-time assertions work correctly (version verification step passes)
- The unit tests work correctly (all version tests pass)
- No invalid versions reach actual test execution

### Comparison: PR vs Main

| Metric | PR Logs | Main Logs |
|--------|---------|-----------|
| Files with `0.1.dev` versions | 18 (Windows) | 18 (Windows) |
| Version test failures | 0 | 0 |
| Import-time assertion failures | 0 | 0 |

**Key Difference**: The PR has the "Verify installed package versions" step that explicitly tests the import-time assertions, providing early detection. The main branch doesn't have this step, but also doesn't fail because the bad versions get replaced before tests run.

### Remaining Issue

While the bad versions are caught and replaced, they still cause:
1. **Dependency conflict warnings**: `cuda-bindings requires cuda-pathfinder~=1.1, but you have cuda-pathfinder 0.1.dev1+g233447b1b which is incompatible`
2. **Inefficiency**: Unnecessary source builds and reinstallations
3. **Risk**: If the replacement step were to fail, tests would run with the bad version

### Recommended Fix

Change the Windows workflow to match the Linux workflow:

```diff
- pip install --only-binary=:all: -v . --group "test-cu${TEST_CUDA_MAJOR}"
+ pip install --only-binary=:all: -v ./*.whl --group "test-cu${TEST_CUDA_MAJOR}"
```

This will prevent pip from building from source during metadata preparation, eliminating the temporary `0.1.dev` versions entirely.

## Conclusion

The PR's version validation mechanisms are working correctly:
- ✅ Import-time assertions catch invalid versions immediately
- ✅ Unit tests provide explicit coverage and pass
- ✅ CI workflow hardening provides early detection

The remaining `0.1.dev` versions are a pre-existing workflow bug (Windows using `.` instead of `./*.whl`) that causes temporary bad versions, but they are caught and replaced before tests run. The fix is straightforward: update the Windows workflow to use `./*.whl` instead of `.`.

rwgk:
# CI Logs Analysis: PR #1454 (post commit c574a94c2d1a91ab1d42ce33b033238f1fdb2e05)

**CI Run**: [`/wrk/logs_20910873985`](https://github.com/NVIDIA/cuda-python/actions/runs/20910873985/job/60073371591?pr=1454)

## Executive Summary

✅ **All checks passed successfully.** The fix to use `./*.whl` instead of `.` in the Windows workflow has eliminated the `0.1.dev` fallback versions that were previously appearing. All version validation mechanisms are working as intended.

## Key Findings

### 1. No `0.1.dev` Package Versions Detected

**Search Results**: Comprehensive grep across all log files found **zero instances** of `cuda-pathfinder`, `cuda-bindings`, or `cuda-core` packages with `0.1.dev` versions.

- All installed packages show valid versions:
  - `cuda-pathfinder-1.3.4.dev86+gc574a94c2`
  - `cuda-bindings-13.1.2.dev72+gc574a94c2`
  - `cuda-core-0.5.1.dev39+gc574a94c2`

### 2. Windows Workflow Fix Confirmed

The fix to change `pip install --only-binary=:all: -v .` to `pip install --only-binary=:all: -v ./*.whl` is working correctly:

**Example from `Test_win-64___py3.14t__13.1.0__wheels__a100__MCDM_.txt`**:
```
2026-01-12T07:27:30.4348704Z + pip install --only-binary=:all: -v ./cuda_pathfinder-1.3.4.dev86+gc574a94c2-py3-none-any.whl --group test-cu13
2026-01-12T07:27:31.1679583Z Processing c:\actions-runner\_work\cuda-python\cuda-python\cuda_pathfinder\cuda_pathfinder-1.3.4.dev86+gc574a94c2-py3-none-any.whl
```

Pip is now correctly processing the wheel file directly, preventing any source builds that would trigger `setuptools-scm` in shallow clones.

### 3. Version Validation Tests Passing

All version validation tests are passing across all platforms:

**Windows Examples**:
- `tests/test_version_number.py::test_pathfinder_version PASSED`
- `tests/test_version_number.py::test_bindings_version PASSED`
- `tests/test_version_number.py::test_core_version PASSED`

**Linux Examples**:
- `tests/test_version_number.py::test_pathfinder_version PASSED`
- `tests/test_version_number.py::test_bindings_version PASSED`
- `tests/test_version_number.py::test_core_version PASSED`

### 4. Import-Time Verification Steps Executing

The CI workflow hardening steps are executing successfully:

**Windows** (`Test_win-64___py3.14t__13.1.0__wheels__a100__MCDM_.txt`):
```
2026-01-12T07:27:30.1648632Z + python -c 'import cuda.pathfinder'
2026-01-12T07:27:30.2797131Z + python -c 'import cuda.bindings'
```

**Linux** (`Test_linux-64___py3.14t__13.1.0__local__l4.txt`):
```
2026-01-12T07:21:49.6369200Z + python -c 'import cuda.pathfinder'
2026-01-12T07:21:49.7363396Z + python -c 'import cuda.bindings'
```

These steps successfully trigger the import-time assertions in `__init__.py` files, providing immediate feedback if invalid versions are present.

### 5. No Dependency Conflicts

**Search Results**: No dependency conflict warnings related to version mismatches. The only "incompatible" matches found were from unrelated test names (`test_from_buffer_incompatible_dtype_and_itemsize`), which are expected test cases.

### 6. No Assertion Errors

**Search Results**: Zero instances of:
- `FATAL: invalid __version__`
- `AssertionError.*version`
- `Invalid version number detected`

This confirms that:
1. All packages have valid versions
2. Import-time assertions are not triggering
3. Unit tests are not encountering invalid versions

## Comparison with Previous Analysis

### Previous Issue (Logs `/wrk/logs_20909188905`)

In the previous CI run, we identified that Windows workflows were building `cuda-pathfinder` from source with `0.1.dev` versions due to pip treating `.` as a source distribution, even with `--only-binary=:all:`. The bad versions were then replaced by wheels before tests ran, masking the issue.

### Current State (Logs `/wrk/logs_20910873985`)

✅ **Issue Resolved**: The fix to use `./*.whl` instead of `.` ensures pip installs directly from wheel files, preventing any source builds in shallow clones.

## Build Workflow Analysis

### Build Jobs

All build jobs show `cuda-pathfinder` being built from source (expected behavior in build workflows):
```
Building wheels for collected packages: cuda-pathfinder
```

However, these builds occur in contexts where full git history is available (build jobs fetch tags), so `setuptools-scm` correctly generates versions like `1.3.4.dev86+gc574a94c2`.

### Test Jobs (Wheel Installation)

Test jobs that install from wheels show:
- `Processing c:\actions-runner\_work\cuda-python\cuda-python\cuda_pathfinder\cuda_pathfinder-1.3.4.dev86+gc574a94c2-py3-none-any.whl`
- No source builds triggered
- All packages installed with correct versions

## Two-Layer Defense Verification

### Layer 1: Import-Time Assertions ✅

The `assert tuple(int(_) for _ in __version__.split(".")[:2]) > (0, 1)` statements in `__init__.py` files are:
- Present in all three packages (`cuda.bindings`, `cuda.core`, `cuda.pathfinder`)
- Not triggering (no assertion errors in logs)
- Verified by the explicit import steps in CI workflows

### Layer 2: Unit Tests ✅

The `test_version_number.py` tests are:
- Running successfully across all platforms
- Passing for all three packages
- Providing granular reporting (separate tests for each package)

## Conclusion

The PR's version validation strategy is working as designed:

1. ✅ **Import-time assertions** provide fail-fast detection of invalid versions
2. ✅ **Unit tests** provide explicit coverage and granular reporting
3. ✅ **CI workflow hardening** ensures wheel-only installation in test workflows
4. ✅ **Windows workflow fix** prevents source builds that would trigger `setuptools-scm` fallback in shallow clones

**No issues detected.** The PR is ready for review.

## Recommendations

1. ✅ **Ready for merge**: All validation mechanisms are functioning correctly
2. ✅ **Monitoring**: Continue monitoring CI logs for any future `0.1.dev` versions (should not occur with current fixes)
3. ✅ **Documentation**: The PR description accurately reflects the implemented solution

cuda_bindings/cuda/bindings/__init__.py
```diff
@@ -3,3 +3,5 @@
 
 from cuda.bindings import utils
 from cuda.bindings._version import __version__
+
+assert tuple(int(_) for _ in __version__.split(".")[:2]) > (0, 1), "FATAL: invalid __version__"
```
kkraus14, line 7: I don't think we should do this outside of an opt-in debug mechanism. If someone grabbed the source code in some weird way that breaks the version resolution via setuptools_scm, we don't want it to error for them unnecessarily, where getting an "incorrect" version would be better than getting this assertion.

cuda_core/cuda/core/__init__.py
```diff
@@ -15,6 +15,8 @@
 
 import importlib
 
+assert tuple(int(_) for _ in __version__.split(".")[:2]) > (0, 1), "FATAL: invalid __version__"
```
kkraus14, line 18: +1

cuda_core/tests/helpers/__init__.py
```diff
@@ -18,9 +30,36 @@
...
 
 
-try:
-    from cuda_python_test_helpers import *  # noqa: F403
-except ImportError:
-    # Import shared platform helpers for tests across repos
-    sys.path.insert(0, str(pathlib.Path(__file__).resolve().parents[3] / "cuda_python_test_helpers"))
-    from cuda_python_test_helpers import *  # noqa: F403
+@functools.cache
+def supports_ipc_mempool(device_id: Union[int, object]) -> bool:
```
kkraus14, line 34: Do we need to move this helper in this PR? It would ideally be done in a separate PR.

cuda_pathfinder/cuda/pathfinder/__init__.py
```diff
@@ -14,6 +12,10 @@
 from cuda.pathfinder._headers.find_nvidia_headers import find_nvidia_header_directory as find_nvidia_header_directory
 from cuda.pathfinder._headers.supported_nvidia_headers import SUPPORTED_HEADERS_CTK as _SUPPORTED_HEADERS_CTK
 
+from cuda.pathfinder._version import __version__  # isort: skip
+
+assert tuple(int(_) for _ in __version__.split(".")[:2]) > (0, 1), "FATAL: invalid __version__"
```
kkraus14, line 17: +1

cuda_core/build_hooks.py
```diff
@@ -160,9 +161,80 @@ def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
...
+        return (False, False, None)
+
+    # Determine repo root (parent of cuda_core)
+    repo_root = Path(__file__).resolve().parent.parent
+
+    # Check if _version.py is under repo root (editable install)
+    version_file_path = Path(bv.__file__).resolve()
+    is_editable = repo_root in version_file_path.parents
+
+    # Extract major version from version tuple
```
mdboom: This is extracting from a string.
```suggestion
    # Extract major version from version string
```

.github/workflows/test-wheel-linux.yml
```diff
@@ -37,6 +37,8 @@ jobs:
     steps:
       - name: Checkout ${{ github.event.repository.name }}
         uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8  # v6.0.1
+        with:
+          fetch-depth: 1  # Intentionally shallow clone to test wheel installation without full git history
```
mdboom, line 41: `fetch-depth: 1` is already the default: https://github.com/actions/checkout/blob/main/README.md (and all the other instances of this below).

.github/workflows/test-wheel-linux.yml
```diff
@@ -269,17 +273,22 @@ jobs:
...
           else
-            pip install $(ls cuda_python*.whl)[all]
+            pip install --only-binary=:all: $(ls cuda_python*.whl)[all]
           fi
 
+      - name: Verify installed package versions
+        run: |
+          # Trigger __version__ assertions to guard against setuptools-scm fallback versions
+          python -c 'import cuda.pathfinder'
+          python -c 'import cuda.bindings'
```
mdboom, line 285: If we don't do the import-time checks (as @kkraus14 suggested) we should probably also remove this step.

cuda_core/build_hooks.py
```diff
@@ -160,9 +161,80 @@ def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
...
+            f"This mismatch could cause build or runtime errors.\n"
+            f"\n"
+            f"To fix:\n"
+            f"  1. Uninstall cuda-bindings: pip uninstall cuda-bindings\n"
+            f"  2. Or install from sources: pip install -e ./cuda_bindings\n"
+            f"  3. Or install matching version: pip install 'cuda-bindings=={cuda_major}.*'"
+        )
+
+    # Installed and version matches (or is editable), keep it
+    return ["cuda-bindings"]
```
mdboom, line 237: I think this check is probably fine to do, but if I understand what it's doing correctly, it won't run under normal circumstances with build isolation, because `cuda_bindings` is not a build dependency of `cuda_core`.  It is only without build isolation that this deeper check with `cuda_bindings` installed would even run.  Is it worth it for this complexity of creating code that might break (since it's rarely used...?)

cuda_core/tests/helpers/__init__.py
rwgk: Done: #1459

.github/workflows/test-wheel-linux.yml
rwgk: I really hope not. See my comment on the private repo.

When we talked on Thursday I thought there was immediate consensus that the silent fallback is harmful. I'm very surprised to get pushback here.

cuda_core/build_hooks.py
rwgk: Also see my comment on the private repo: if we don't have these guards, how can we be sure what the QA team is testing? This is like a seatbelt in a car: rarely used.

.github/workflows/test-wheel-linux.yml
rwgk: > is already the default

That's noted in commit 233447b1b8b9f2162104a5e1a68b8378c475d18b:

```
    - Set fetch-depth: 1 explicitly (although it is the default) with comment
      emphasizing that shallow cloning is intentional
```

Without this, someone later may not realize the intent and change it.

cuda_core/build_hooks.py
rwgk: Done: commit d889957c9aaad39ff3ff58479867ed831059687a

Thanks! (The comment was from an earlier version. Updating it was overlooked.)

kkraus14:
My biggest concern is that it's quite easy to download a commit from a repo without getting the git information, i.e. for any given commit / tag you can download an archive without the `.git` folder and information. Example from the latest commit as of this comment: https://github.com/NVIDIA/cuda-python/archive/abf7ff8a850ccdd79251c6e41ea28f0331de5b3d.zip

If someone downloads the archive in this way, I assume setuptools-scm wouldn't work as expected and so the version they'd get wouldn't be right and that assertion would trigger?

rwgk:
> My biggest concern is that it's quite easy to download a commit from a repo without getting the git information, i.e. for any given commit / tag you can download an archive without the `.git` folder and information. Example from the latest commit as of this comment: https://github.com/NVIDIA/cuda-python/archive/abf7ff8a850ccdd79251c6e41ea28f0331de5b3d.zip
> 
> If someone downloads the archive in this way, I assume setuptools-scm wouldn't work as expected and so the version they'd get wouldn't be right and that assertion would trigger?

I'll play with this, thanks!

rwgk:
> My biggest concern is that it's quite easy to download a commit from a repo without getting the git information, i.e. for any given commit / tag you can download an archive without the `.git` folder and information. Example from the latest commit as of this comment: https://github.com/NVIDIA/cuda-python/archive/abf7ff8a850ccdd79251c6e41ea28f0331de5b3d.zip
> 
> If someone downloads the archive in this way, I assume setuptools-scm wouldn't work as expected and so the version they'd get wouldn't be right and that assertion would trigger?

This doesn't work already on main:

```
    File "/tmp/rgrossekunst-tmp/pip-build-env-gtzug7_b/overlay/lib/python3.12/site-packages/setuptools_scm/_get_version_impl.py", line 188, in _version_missing
      raise LookupError(error_msg)
  LookupError: setuptools-scm was unable to detect version for /tmp/test_main_zip_artifact_740076/cuda-python-main.
```

I believe it's a direct consequence of PR #1411, although I didn't backtrack to the commit before the PR.

Cursor produced the attached reproducer (even though I wasn't directly asking for it!):

[test_main_zip_artifact.sh](https://github.com/user-attachments/files/24577168/test_main_zip_artifact.sh)

(On my colossus workstation it takes only 7 seconds from downloading the artifact to reproducing the full error.)

rwgk:
Here is a new version that let's you pass a commit hash (optional):

[test_cuda-python_zip_artifact.sh](https://github.com/user-attachments/files/24577251/test_cuda-python_zip_artifact.sh)

Using the commit right before PR #1418:

```
./test_main_zip_artifact.sh 51b9c6c4a6ba10ec123b702b39e36b4de0da19d9
```

That works:

```
Successfully installed cuda-pathfinder-1.3.4a0 iniconfig-2.3.0 packaging-25.0 pluggy-1.6.0 pygments-2.19.2 pytest-9.0.2

ERROR: Build succeeded unexpectedly!
```

I.e. this demonstrates that we lost that feature (Successfully installed cuda-pathfinder-...) already with PR #1411.

kkraus14:
> This doesn't work already on main

This is disappointing and we should probably address it before building additional validation on top of it. Seems like we need to explicitly add git archive support: https://setuptools-scm.readthedocs.io/en/latest/usage/#git-archives

mdboom:
> > This doesn't work already on main
> 
> This is disappointing and we should probably address it before building additional validation on top of it. Seems like we need to explicitly add git archive support: https://setuptools-scm.readthedocs.io/en/latest/usage/#git-archives

I see the requirements as:

1) A random person doing their own self-integration from source should be able to build something that installs and works, whether all the git info is there or not.

2) For our official releases, IMHO it's completely reasonable to require that it's a proper git checkout with tags.  This already works whether the clone is deep or shallow.  IIUC, the reason we ran into this is because the private fork of cuda-python didn't have the tags, but we can fix that by pushing them there one time and have this fixed going forward.

In scenario 1), it was already creating the wrong version (just a different wrong thing) prior to using `setuptools-scm` -- it would build 3.1.0+(25 extra commits) and happily call it 3.1.0.  We can either do the full [git archives]
(https://setuptools-scm.readthedocs.io/en/latest/usage/#git-archives) metadata thing described, or if we just want it to build but not care what the version number is, we can set a `fallback_version` in the config.  That doesn't have any ongoing maintenance cost.

rwgk:
Closing as obsolete.

I sent #1493 for review, which extracts a related but separate bug fix from this PR.

The init assertions were moved to PR #1470 already but did not get support from the team. @leofang summarized this concisely here:

https://github.com/NVIDIA/cuda-python/pull/1470#issuecomment-3747457156

I'll close that PR, too. We can figure something out later (new test cycle).

Issue #1455, opened 2026-01-11
[BUG]: Extremely slow `cuda_core/tests/test_helpers.py::test_patterngen_seeds` on Windows


rwgk:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Performance

### Component

cuda.core

### Describe the bug

Using the log archive from https://github.com/NVIDIA/cuda-python/actions/runs/20871705999 (main @ acc78f7c0e570b59fed7b07429dce5bfe143ffd8 ← #1368)

Maximum measured times per platform:
* `linux-64`: 14.93s
* `linux-aarch64`: 12.07s
* `win-64`: **945.29s** (15m 45s)

The `linux-64` maximum (14.93s) is from `Test_linux-64___py3.13__13.0.2__local__rtxpro6000.txt`.

The linux-aarch64 maximum (12.07s) is from `Test_linux-aarch64___py3.14t__13.1.0__local__a100.txt`.

The win-64 maximum (945.29s) is from `Test_win-64___py3.14__13.1.0__local__l4__MCDM_.txt`.

```
smc120-0004.ipp2a2.colossus.nvidia.com:/wrk/main_logs_20871705999 $ grep ' call.*test_helpers.py::test_patterngen_seeds' *.txt
Test_linux-64___py3.10__12.9.1__local__v100.txt:2026-01-10T03:19:17.6039085Z 14.63s call     tests/test_helpers.py::test_patterngen_seeds
Test_linux-64___py3.10__13.0.2__wheels__l4.txt:2026-01-10T03:08:24.5435099Z 12.83s call     tests/test_helpers.py::test_patterngen_seeds
Test_linux-64___py3.10__13.1.0__wheels__l4.txt:2026-01-10T03:08:21.9057669Z 13.24s call     tests/test_helpers.py::test_patterngen_seeds
Test_linux-64___py3.11__12.9.1__wheels__rtxpro6000.txt:2026-01-10T03:19:08.2634608Z 13.43s call     tests/test_helpers.py::test_patterngen_seeds
Test_linux-64___py3.11__13.0.2__local__l4.txt:2026-01-10T03:12:47.9856996Z 12.80s call     tests/test_helpers.py::test_patterngen_seeds
[...]

PR #1456, opened 2026-01-11
Add `skipif` `IS_WINDOWS` for `test_patterngen_seeds`

rwgk:
### Problem
The `test_patterngen_seeds` test runs extremely slowly on Windows platforms (up to ~945 seconds / ~15 minutes) compared to Linux (~12-15 seconds), significantly slowing down our CI and severely impeding local testing. See issue #1455 for detailed performance numbers.

### Solution
Add a `@pytest.mark.skipif` decorator to skip this test on Windows platforms. This restores CI and local testing speed while we work on investigating and fixing the root cause.

### Changes
- Added Windows skip condition to `test_patterngen_seeds` in `cuda_core/tests/test_helpers.py`
- Skip message references issue #1455 for tracking
- Test continues to run normally on Linux and other platforms

### DIFF
--- a/cuda_core/tests/test_helpers.py
+++ b/cuda_core/tests/test_helpers.py
@@ -48,6 +48,10 @@ def test_latchkernel():
 
 
 @pytest.mark.skipif(
+    IS_WINDOWS,
+    reason="Extremely slow on Windows (issue #1455).",
+)
+@pytest.mark.skipif(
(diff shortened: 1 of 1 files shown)


Andy-Jost (approved):
This looks like a good change.

It seems like the root cause is that UVM is much slower on Windows than Linux "just because." If so, that's really too bad.

PR #1464, opened 2026-01-12
Eliminate _CXX_API capsule for resource handle functions

### LINKED ISSUE #1452: Investigate eliminating _CXX_API capsule for resource handle functions
## Summary

PR #1368 uses a PyCapsule mechanism (`_CXX_API`) to share resource handle functions across Cython modules. This issue is to investigate whether this capsule can be eliminated entirely.

## Current Architecture

The `_CXX_API` capsule exports function pointers from `_resource_handles.so` to other Cython modules (`_buffer.pyx`, `_stream.pyx`, `_event.pyx`, etc.). The stated rationale is to avoid duplicate static/thread-local state if multiple modules linked the C++ code directly.

## Proposed Investigation

Investigate whether Cython modules can call directly into the resource handles code without the capsule indirection:

- If resource handle functions are regular `cdef` functions (non-inline) in `_resource_handles.pyx`, can other modules `cimport` and call them directly?
- Would this approach correctly share static state through `_resource_handles.so`?
- Are there performance implications for `nogil` code paths?

## Background

When module A `cimport`s a `cdef` function from module B, the call goes through `B.so` at runtime via Python's import mechanism. This should mean static state in B stays in one place, potentially making the capsule unnecessary for pure Cython-to-Cython usage.

## References

- PR #1368 - RAII resource handles implementation
- Issue #1450 - Related investigation for cuda.bindings driver API capsule

## Goal

Simplify the architecture by removing the capsule indirection if it's not required, reducing complexity while preserving correct behavior.

Andy-Jost:
## Summary
- Replace the PyCapsule-based function pointer table with direct Cython `cimport`
- Consumer modules now call resource handle functions directly through `_resource_handles.so`
- Simplifies the architecture while correctly sharing static/thread-local state

## Changes
- Remove `_CXX_API` capsule infrastructure (`resource_handles_cxx_api.hpp`, `_resource_handles_cxx_api.pxd`, `get_resource_handles_cxx_api_v1()`)
- Remove `_init_handles_table()` calls from all consumer modules
- Rename `create_event_handle(flags)` to `create_event_handle_noctx(flags)` to avoid C++ overload ambiguity for Cython binding
- Update `DESIGN.md` to reflect the simplified architecture
- Add clarifying comment in `build_hooks.py` for cpp file discovery

## Test Plan
- [x] Verified `resource_handles.cpp` symbols exist only in `_resource_handles.so` (confirmed via `nm -C`)
- [ ] CI tests pass

## Stats
**16 files changed, 179 insertions(+), 512 deletions(-)**

Net reduction: **333 lines**

Closes #1452

### DIFF
--- a/cuda_core/cuda/core/_resource_handles.pyx
+++ b/cuda_core/cuda/core/_resource_handles.pyx
@@ -10,9 +10,8 @@
 # The cdef extern from declarations below satisfy the .pxd declarations directly,
 # without needing separate wrapper functions.
 
-from cpython.pycapsule cimport PyCapsule_New
+from cpython.pycapsule cimport PyCapsule_GetName, PyCapsule_GetPointer
 from libc.stddef cimport size_t
-from libc.stdint cimport uint32_t, uint64_t, uintptr_t
 
 from cuda.bindings cimport cydriver
 
@@ -24,8 +23,7 @@ from ._resource_handles cimport (
     DevicePtrHandle,
 )
 
-import cython
-
+import cuda.bindings.cydriver as cydriver
 
 # =============================================================================
 # C++ function declarations (non-inline, implemented in resource_handles.cpp)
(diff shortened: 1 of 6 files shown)


cuda_core/cuda/core/_resource_handles.pyx
```diff
@@ -2,33 +2,105 @@
...
+
+cdef extern from "_cpp/resource_handles.hpp" namespace "cuda_core":
+    # Thread-local error handling
+    cydriver.CUresult get_last_error "cuda_core::get_last_error" () noexcept nogil
+    cydriver.CUresult peek_last_error "cuda_core::peek_last_error" () noexcept nogil
+    void clear_last_error "cuda_core::clear_last_error" () noexcept nogil
+
+    # Context handles
+    ContextHandle create_context_handle_ref "cuda_core::create_context_handle_ref" (
+        cydriver.CUcontext ctx) nogil
```
rwgk: Cursor discovered this:
___

## Minor Issue: Inconsistent `noexcept` Annotations

The `.pxd` file declares all handle functions as `noexcept nogil`:

```cython
# _resource_handles.pxd
cdef ContextHandle create_context_handle_ref(cydriver.CUcontext ctx) noexcept nogil
cdef StreamHandle create_stream_handle(...) noexcept nogil
cdef EventHandle create_event_handle(...) noexcept nogil
# etc.
```

But the `.pyx` file's `cdef extern from` declarations are missing `noexcept` on many of these same functions:

```cython
# _resource_handles.pyx
ContextHandle create_context_handle_ref "cuda_core::create_context_handle_ref" (
    cydriver.CUcontext ctx) nogil  # <-- missing noexcept
StreamHandle create_stream_handle "cuda_core::create_stream_handle" (
    ContextHandle h_ctx, unsigned int flags, int priority) nogil  # <-- missing noexcept
EventHandle create_event_handle "cuda_core::create_event_handle" (
    ContextHandle h_ctx, unsigned int flags) nogil  # <-- missing noexcept
```

Meanwhile, some functions *do* have `noexcept` in both places (e.g., `get_primary_context`, `get_current_context`, `get_legacy_stream`, `get_per_thread_stream`).

**Functions missing `noexcept` in `.pyx`:**
- `create_context_handle_ref`
- `create_stream_handle`
- `create_stream_handle_ref`
- `create_event_handle`
- `create_event_handle_noctx`
- `create_event_handle_ipc`
- `create_mempool_handle`
- `create_mempool_handle_ref`
- `create_mempool_handle_ipc`
- `deviceptr_alloc_from_pool`
- `deviceptr_alloc_async`
- `deviceptr_alloc`
- `deviceptr_alloc_host`
- `deviceptr_create_ref`
- `deviceptr_import_ipc`

**Impact**: The C++ functions don't throw exceptions, so `noexcept` is semantically correct for all of them. The inconsistency probably doesn't cause runtime issues, but it would be cleaner if the `.pyx` declarations matched the `.pxd` declarations exactly.

**Suggested fix**: Add `noexcept` to all the `cdef extern from` declarations in `_resource_handles.pyx` that have it in `_resource_handles.pxd`.

cpcloud:
If we have some expectations about visibility, can we roll the `nm` check into CI?

Andy-Jost:
> If we have some expectations about visibility, can we roll the `nm` check into CI?

I'm not opposed to this, and it would be easy. It seems unusual to verify build scripts in a unit test.

Andy-Jost:
Thanks @rwgk for pointing out the inconsistent use of `noexcept`. I did a pass to update everything to correctly use `noexcept` (C++ and Cython) and `nogil` (Cython only).

All of the C++ functions are marked `noexcept`. Callers are expected to use `get_last_error` and friends. We do not expect these functions to ever throw. Technically, `new` might throw, but that's not something we need to worry about, and the effect of `noexcept` calling `std::terminate` is fine in that case.

When bringing these functions into Cython, most functions are marked `noexcept` and `nogil`. The only exceptions are the `as_py` overloads, which interact with Python, require the GIL, and could raise exceptions.

Two functions manipulate Python objects: `create_stream_handle_with_owner` and `deviceptr_create_with_owner`. I updated those to be callable without the GIL (in case the owner is `None`) and to acquire the GIL, if it is needed. These can be called with or without the GIL and will function correctly in either case. There is an insignificant cost to checking the GIL if these are called with it held.

Issue #464, opened 2025-02-24
Turn all `cdef class` objects in the Python layer to `cdef public class`


leofang:
This allows us to share the class declarations with C/C++:
https://cython.readthedocs.io/en/latest/src/userguide/external_C_code.html#using-cython-declarations-from-c
which is needed after PR #463 set up a path for Python-C/C++/Cython interoperability, but without adding the `public` qualifier the interoperability was limited to Cython only.

leofang:
@vzhurba01 I hope this is a simple one-liner change to the codegen and the remaining task is on verifying the generated headers. Could you take a look?

leofang:
It is unclear to me if we still need this done if we want to go by the route of #564. @oleksandr-pavlyk @shwina WDYT?

shwina:
Hmm, if my understanding is correct, the two seem a bit orthogonal:

1. `cdef public` would allow e.g., third-party C code to use types defined in Cython (not sure if we know of use cases for this)
2. `get_native_hande()` would allow us to grab a pointer to the underlying native CUDA object wrapped by a Cython type

Am I missing something or can (1) be used in place of (2) somehow?

oleksandr-pavlyk:
Leo, I think it should be `cdef api` instead of `cdef public`.

PR #1466, opened 2026-01-12
Use __pyx_capi__ for CUDA driver function pointers

### LINKED ISSUE #1450: Investigate using __pyx_capi__ to simplify resource handle architecture
## Summary

PR #1368 introduces a PyCapsule-based architecture for resource handles. This issue is to investigate whether Cython's `__pyx_capi__` mechanism could simplify or replace this approach.

## Current Architecture

PR #1368 uses a PyCapsule mechanism **`_CUDA_DRIVER_API_V1`** to pass CUDA driver function pointers from `cuda.bindings` to the C++ code

## Proposed Investigation

Investigate whether `__pyx_capi__` (Cython's built-in mechanism for exposing C-level APIs) could replace this capsule:

- Can cuda.bindings expose driver functions via `__pyx_capi__` instead of requiring a custom capsule?

## References

- https://github.com/cython/cython/issues/3077 - Discussion of `__pyx_capi__` for cross-module C API access
- PR #1368 - RAII resource handles implementation

## Goal

Simplify the architecture if possible, reducing complexity and maintenance burden while preserving the core benefits of centralized resource lifetime management.

Andy-Jost:
## Summary

Replace the `_CUDA_DRIVER_API_V1` capsule with direct extraction of function pointers from `cuda.bindings.cydriver.__pyx_capi__` at module import time.

## Changes

- Remove `CudaDriverApiV1` struct and `_get_cuda_driver_api_v1_capsule()` from `_resource_handles.pyx`
- Remove `load_driver_api()`, `ensure_driver_loaded()`, and related machinery from `resource_handles.cpp`
- Add extern declarations for driver function pointers in `resource_handles.hpp`
- Populate function pointers at module import using `PyCapsule_GetPointer()` with `PyCapsule_GetName()` for signature lookup
- Update DESIGN.md to reflect the new `__pyx_capi__` approach

This simplifies the architecture by eliminating the custom capsule struct and its loading machinery. The driver function pointers are now populated directly from Cython's built-in cross-module API mechanism.

**Stats:** 4 files changed, 228 insertions(+), 392 deletions(-) (net -164 lines)

## Test Plan

- [x] Build succeeds
- [x] Import works on CPU-only machines (no GPU/driver required)
- [x] `test_stream.py`, `test_event.py`, `test_memory.py`, `test_device.py` pass
- [x] CI tests pass

Closes #1450

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -9,14 +9,25 @@ on:
     types: [closed, labeled]
     branches:
       - main
+  workflow_dispatch:
+    inputs:
+      backport-branch:
+        description: "Branch to backport commits onto"
+        required: false
+        type: string
+      pull-request:
+        description: "PR to backport"
+        required: true
+        type: number
+
 
 permissions:
   contents: write       # so it can comment
   pull-requests: write  # so it can create pull requests
 
 jobs:
-  backport:
-    name: Backport pull request
+  backport-from-pr:
+    name: Backport directly from a pull request
     if: ${{ github.repository_owner == 'nvidia' &&
             github.event.pull_request.merged == true &&
             contains( github.event.pull_request.labels.*.name, 'to-be-backported')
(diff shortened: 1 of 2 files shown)


kkraus14:
Instead of reaching for the `__pyx_capi__` which is an implementation detail of Cython, did we try using `cdef api` and `cdef public`? https://cython.readthedocs.io/en/latest/src/userguide/external_C_code.html#c-api-declarations

This generates header file(s) that we can use nicely as expected instead of reaching into internal structs generated by Cython.

Andy-Jost:
@kkraus14 No, I didn't consider `cdef api`.

We discussed `__pyx_capi__` in a review meeting and concluded it was stable based on [this discussion](https://github.com/cython/cython/issues/3077).

If we were to use `cdef api` instead, it would create a build-time dependency on cuda-bindings:
- We'd need to modify the cuda-bindings code generator templates to add the `api` keyword to the driver function definitions
- Cython would then generate `*_api.h` headers that we'd need to export and ship with cuda-bindings
- cuda-core would `#include` these headers and call an `import_*()` function at initialization

The tradeoff is that `cdef api` is an official Cython feature, whereas `__pyx_capi__` is technically internal (though stable in practice).

If you'd like me to look into using `cdef api`, please let me know and I will create a follow-up issue.

leofang (approved):
Wonderful! Thanks for looking into it @Andy-Jost.

leofang:
Sorry, for some reason I missed the `cdef api` discussion. That's something we'd like to explore (https://github.com/NVIDIA/cuda-python/issues/464) but I don't think we need to tie it to this safety feature.

Issue #965, opened 2025-09-11
CI: Automate backport process more


leofang:
We have a backport bot but it does not always work:
- Any PR touching files under `.github/` cannot be backported by the bot due to permission issues
- `cuda.bindings` most of the time needs to be regenerated, so a naive backport does not always work either

leofang:
> * Any PR touching files under `.github/` cannot be backported by the bot due to permission issues

To remind myself why it hasn't been fixed: This is a [typical error message](https://github.com/NVIDIA/cuda-python/actions/runs/18353007048/job/52277690368#step:4:67) from the backport bot
```
git push --set-upstream origin backport-1106-to-12.9.x
To https://github.com/NVIDIA/cuda-python
 ! [remote rejected] backport-1106-to-12.9.x -> backport-1106-to-12.9.x (refusing to allow a GitHub App to create or update workflow `.github/workflows/ci.yml` without `workflows` permission)
error: failed to push some refs to 'https://github.com/NVIDIA/cuda-python'
```
which means for the bot to work it needs the `workflows` permissions (specifically, I am guessing `workflows: write`). However, based on [the docs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#defining-access-for-the-github_token-scopes) we can't grant `GITHUB_TOKENS` such a permission. It seems a more complex solution is needed to work around the permission issue.

PR #1467, opened 2026-01-13
Add Nvfatbin Bindings

### LINKED ISSUE #156: Add nvfatbin bindings

isVoid:
## Description

closes #156 

## Checklist

### DIFF
--- a/cuda_bindings/docs/source/module/nvfatbin.rst
+++ b/cuda_bindings/docs/source/module/nvfatbin.rst
@@ -1,0 +1,89 @@
+.. SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+.. SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
+
+.. default-role:: cpp:any
+
+nvfatbin
+========
+
+Note
+----
+
+The nvfatbin bindings are not supported on nvFatbin installations <12.4. Ensure the installed CUDA toolkit's nvFatbin version is >=12.4.
+
+The Tile IR API (:func:`cuda.bindings.nvfatbin.add_tile_ir`) is only available in CUDA 13.1+.
+
+Functions
+---------
+
+NvFatbin defines the following functions for creating and populating fatbinaries.
+
+.. autofunction:: cuda.bindings.nvfatbin.create
+.. autofunction:: cuda.bindings.nvfatbin.destroy
+.. autofunction:: cuda.bindings.nvfatbin.add_ptx
+.. autofunction:: cuda.bindings.nvfatbin.add_cubin
+.. autofunction:: cuda.bindings.nvfatbin.add_ltoir
+.. autofunction:: cuda.bindings.nvfatbin.add_reloc
+.. autofunction:: cuda.bindings.nvfatbin.add_tile_ir
+.. autofunction:: cuda.bindings.nvfatbin.size
+.. autofunction:: cuda.bindings.nvfatbin.get
+.. autofunction:: cuda.bindings.nvfatbin.get_error_string
+.. autofunction:: cuda.bindings.nvfatbin.version
+
+Types
+---------
+.. autoclass:: cuda.bindings.nvfatbin.Result
+
+    .. autoattribute:: cuda.bindings.nvfatbin.Result.SUCCESS
+
+
+    .. autoattribute:: cuda.bindings.nvfatbin.Result.ERROR_INTERNAL
+
+
+    .. autoattribute:: cuda.bindings.nvfatbin.Result.ERROR_ELF_ARCH_MISMATCH
+
+
+    .. autoattribute:: cuda.bindings.nvfatbin.Result.ERROR_ELF_SIZE_MISMATCH
+
+
+    .. autoattribute:: cuda.bindings.nvfatbin.Result.ERROR_MISSING_PTX_VERSION
+
+
+    .. autoattribute:: cuda.bindings.nvfatbin.Result.ERROR_NULL_POINTER
+
+
+    .. autoattribute:: cuda.bindings.nvfatbin.Result.ERROR_COMPRESSION_FAILED
+
+
+    .. autoattribute:: cuda.bindings.nvfatbin.Result.ERROR_COMPRESSED_SIZE_EXCEEDED
+
+
+    .. autoattribute:: cuda.bindings.nvfatbin.Result.ERROR_UNRECOGNIZED_OPTION
+
+
+    .. autoattribute:: cuda.bindings.nvfatbin.Result.ERROR_INVALID_ARCH
+
+
+    .. autoattribute:: cuda.bindings.nvfatbin.Result.ERROR_INVALID_NVVM
+
+
+    .. autoattribute:: cuda.bindings.nvfatbin.Result.ERROR_EMPTY_INPUT
+
+
+    .. autoattribute:: cuda.bindings.nvfatbin.Result.ERROR_MISSING_PTX_ARCH
+
+
+    .. autoattribute:: cuda.bindings.nvfatbin.Result.ERROR_PTX_ARCH_MISMATCH
+
+
+    .. autoattribute:: cuda.bindings.nvfatbin.Result.ERROR_MISSING_FATBIN
+
+
+    .. autoattribute:: cuda.bindings.nvfatbin.Result.ERROR_INVALID_INDEX
+
+
+    .. autoattribute:: cuda.bindings.nvfatbin.Result.ERROR_IDENTIFIER_REUSE
+
+
+    .. autoattribute:: cuda.bindings.nvfatbin.Result.ERROR_INTERNAL_PTX_OPTION
+


cuda_bindings/cuda/bindings/_internal/nvfatbin.pxd
```diff
@@ -0,0 +1,27 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
+#
+# This code was automatically generated with version 13.0.0. Do not modify it directly.
```
kkraus14: I think we need to bump this to 13.1.0 and we should get the `nvFatbinAddTileIR` function as well.

cuda_bindings/cuda/bindings/_internal/nvfatbin.pxd
```diff
@@ -0,0 +1,27 @@
...
+
+from ..cynvfatbin cimport *
+
+
+###############################################################################
+# Wrapper functions
+###############################################################################
```
kkraus14, line 12: Seems like we're also missing `nvFatbinGetErrorString`

cuda_bindings/cuda/bindings/_internal/nvfatbin_linux.pyx
```diff
@@ -0,0 +1,307 @@
...
+cdef void* load_library() except* with gil:
+    cdef uintptr_t handle = load_nvidia_dynamic_lib("nvfatbin")._handle_uint
+    return <void*>handle
+
+
+cdef int _init_nvfatbin() except -1 nogil:
+    global __py_nvfatbin_init
+
+    cdef void* handle = NULL
+
```
rparolin, line 84: For double checked locking, I believe you missed the following here or else the init function always has to acquire the lock to check if the object was initialized.

```
        if __py_nvfatbin_init:
            return 0
```

cuda_bindings/cuda/bindings/_internal/nvfatbin_linux.pyx
```diff
@@ -0,0 +1,307 @@
...
+    with gil, __symbol_lock:
+        # Recheck the flag after obtaining the locks
+        if __py_nvfatbin_init:
+            return 0
+
+        # Load function
+        global __nvFatbinCreate
```
rparolin, line 98: Consider a data driving this to avoid repeating the same pattern.

```
handle = None

def resolve_symbol(sym_name):
    global handle

    sym = dlsym(RTLD_DEFAULT, sym_name)
    if sym is None:  # or == NULL depending on your binding
        if handle is None:
            handle = load_library()
        sym = dlsym(handle, sym_name)

    return sym  # optionally raise if still None


_SYMBOLS = [
    "nvFatbinCreate",
    "nvFatbinDestroy",
    "nvFatbinAddPTX",
    "nvFatbinAddCubin",
    "nvFatbinAddLTOIR",
    "nvFatbinAddReloc",
    "nvFatbinSize",
    "nvFatbinGet",
    "nvFatbinVersion",
]

for name in _SYMBOLS:
    globals()["__" + name] = resolve_symbol(name)
```

leofang:
FYI @rparolin all of these are generated code except for the tests. There is an internal MR targeting the codegen for this.

cuda_bindings/tests/test_nvfatbin.py
```diff
@@ -0,0 +1,215 @@
...
+@pytest.fixture
+def OBJECT(arch, tmpdir):
+    if arch == "sm_100":
+        pytest.skip("sm_100 is not supported on local system.")
+
+    empty_cplusplus_kernel = "__global__ void A() {} int main() { return 0; }"
+    with open(tmpdir / "object.cu", "w") as f:
+        f.write(empty_cplusplus_kernel)
+
+    subprocess.check_output(["nvcc", "-arch", arch, "-o", str(tmpdir / "object.o"), str(tmpdir / "object.cu")])
```
leofang: We probably need a check here and skip if NVCC is not usable (ex: there is no host compiler) or even available. For example, at the test stage we don't have access to NVCC in the CI (same in the numba-cuda repo).

cuda_bindings/cuda/bindings/_internal/nvfatbin.pxd
isVoid: Added in d962f61

cuda_bindings/cuda/bindings/_internal/nvfatbin.pxd
isVoid: Right this is now handled in fc4203e, I believe I was using my local 13.0 installation as a PoC run.

cuda_bindings/tests/test_nvfatbin.py
isVoid: I think this is handled in [bd868bd](https://github.com/NVIDIA/cuda-python/pull/1467/commits/bd868bdf7f523dbf6e52ff234966831844aa6032). When we don't have `nvcc`, tests that depend on this fixture are skipped.

cuda_bindings/cuda/bindings/_internal/nvfatbin_linux.pyx
isVoid: Right, as @leofang points out below - this is generated code. Reducing repetition in generated code is desired, but probably out of scope of this PR.

All *.pyx and *pxd files are cybind generated.

cuda_bindings/tests/test_nvfatbin.py
```diff
@@ -0,0 +1,288 @@
...
+}}
+"""
+
+CODE = """
+int __device__ inc(int x) {
+    return x + 1;
+}
+"""
+
+# This is the base64 encoded TileIR from sample VectorAddition program
```
kkraus14: Assuming this is generated from cuTile, it would be nice to capture instructions to recreate this if/when needed or even better to optionally generate it in some kind of way if cuTile is available.

Happy to defer this to a future PR though and would take a `#TODO` comment for now 😄

cuda_bindings/tests/test_nvfatbin.py
```diff
@@ -0,0 +1,288 @@
...
+
+@pytest.fixture
+def OBJECT(arch, tmpdir):
+    empty_cplusplus_kernel = "__global__ void A() {} int main() { return 0; }"
+    with open(tmpdir / "object.cu", "w") as f:
+        f.write(empty_cplusplus_kernel)
+
+    nvcc = shutil.which("nvcc")
+    if nvcc is None:
+        pytest.skip("nvcc not found on PATH")
```
kkraus14: Can we add a `#TODO` comment to use pathfinder to resolve this in the future instead? I imagine this wouldn't cover using the `nvidia-cuda-nvcc` wheel package.

cuda_bindings/tests/test_nvfatbin.py
leofang: Yeah would be nice to leave a comment on how this is generated for future selfs. Ralf did a nice job for similar needs (NVVM IR):
https://github.com/NVIDIA/cuda-python/blob/9711fe3fe355ab8738cce84c76362eed4a5be07c/cuda_bindings/tests/test_nvvm.py#L136-L140

cuda_bindings/tests/test_nvfatbin.py
leofang: Leaving a TODO is fine, though it's unlikely actionable in the near future. We do not have access to host compilers in the CI (at the test stage) so NVCC is useless.

cuda_bindings/tests/test_nvfatbin.py
```diff
@@ -0,0 +1,288 @@
...
+    return PTX_TEMPLATE.format(PTX_VERSION=ptx_version, ARCH=arch)
+
+
+@pytest.fixture
+def CUBIN(arch):
+    def CHECK_NVRTC(err):
+        if err != nvrtc.nvrtcResult.NVRTC_SUCCESS:
+            raise RuntimeError(repr(err))
+
+    err, program_handle = nvrtc.nvrtcCreateProgram(CODE.encode(), b"", 0, [], [])
```
leofang, line 129: Not a blocker for this PR, but I think we have some helper functions for NVRTC that can be used in cuda-bindings tests, so that we do not have to reimplement them in this file.

cuda_bindings/cuda/bindings/_internal/nvfatbin_linux.pyx
leofang: Rob, you made me think twice 😂  The first check is refactored out by Mike a few months ago to `_check_or_init_nvfatbin` which is declared inline. Then, the second check is done in `_init_nvfatbin` (this function) after grabbing the lock.

cuda_bindings/cuda/bindings/_internal/nvfatbin.pxd
```diff
@@ -0,0 +1,24 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
leofang: technically this module is new and we just need 2026 here

leofang:
@isVoid 
- The Windows builds fail with this error:
```
  cuda/bindings\nvfatbin.cpp(3718): error C2440: '=': cannot convert from 'const char *' to 'char *'
```
I suspect it's coming from `get_error_string` and we might need to cast away the constness, could you take a look?
- For test failures with local CTK, we need to append `libnvfatbin` to this list:
https://github.com/NVIDIA/cuda-python/blob/9711fe3fe355ab8738cce84c76362eed4a5be07c/.github/actions/fetch_ctk/action.yml#L17

leofang:
- For test failures with wheels, we need to update this dependency:
https://github.com/NVIDIA/cuda-python/blob/8725e96ea54f9294c70eeb3da61fb3c9cb6bb1b0/cuda_bindings/pyproject.toml#L37
- Please add docs:
   - Take a look at `cuda_bindings/docs/source/api.rst` to see how we add API references
   - Need to update the release notes for both `13.1.X` (we'll decide later if we'd do 13.1.2 or 13.2.0) and `12.9.6`

cuda_bindings/tests/test_nvfatbin.py
isVoid: Alternatively I can also embed an object file in this test script, just like the tileIR.

cuda_bindings/tests/test_nvfatbin.py
isVoid: Agreed, I saw this pattern showed up in test_nvjitlink as well. Perhaps a shared fixture that returns an nvrtc artifact.

cuda_bindings/tests/test_nvfatbin.py
leofang: Either way is fine as long as we know how to reproduce the embedded bytes

cuda_bindings/tests/test_nvfatbin.py
isVoid: Handled in 22bf2ce

cuda_bindings/cuda/bindings/_internal/nvfatbin.pxd
isVoid: [cda0e29](https://github.com/NVIDIA/cuda-python/pull/1467/commits/cda0e2981d00ddf26e2636c91d42f9a145353da6)

cuda_bindings/tests/test_nvfatbin.py
isVoid: I just looked up the size of an object file - with the small kernel here it's over 1.2M. Unlike tile IR this will clutter the test file. I'm going to leave this as a `TODO` for now.

leofang:
```
E   cuda.bindings._internal.utils.FunctionNotFoundError: function nvFatbinAddTileIR is not found
```
Need to skip the tile IR test for nvFatbin < 13.1

isVoid:
> stderr     = b'gcc: No such file or directory\nnvcc fatal   : Failed to preprocess host compiler properties.\n'

This shows in some CI environment there's nvcc installed but it's not properly setup due to missing host compiler.

leofang:
> > stderr     = b'gcc: No such file or directory\nnvcc fatal   : Failed to preprocess host compiler properties.\n'
> 
> This shows in some CI environment there's nvcc installed but it's not properly setup due to missing host compiler.

Yup! See https://github.com/NVIDIA/cuda-python/pull/1467#discussion_r2694763913.

cuda_bindings/pyproject.toml
```diff
@@ -34,7 +34,7 @@ dependencies = ["cuda-pathfinder ~=1.1"]
 
 [project.optional-dependencies]
 all = [
-    "cuda-toolkit[nvrtc,nvjitlink,nvvm]==13.*",
+    "cuda-toolkit[nvrtc,nvjitlink,nvvm,nvfatbin]==13.*",
```
leofang, line 37: @kkraus14 this minds me a discussion we had a few months ago -- One day installing `cuda-binding[all]` would bring in the whole CTK 😂

leofang:
LGTM! Thanks, @isVoid!

cuda_bindings/docs/source/api.rst
```diff
@@ -14,5 +14,6 @@ CUDA Python API Reference
    module/nvrtc
    module/nvjitlink
    module/nvvm
+   module/nvfatbin
```
leofang, line 17: @isVoid this file (`cuda_bindings/docs/source/module/nvfatbin.rst`) needs to be checked in.

isVoid: Yup, just added.

leofang:
@isVoid because of https://github.com/NVIDIA/cuda-python/issues/965 the backport bot does not work (a workflow file is touched). Could you manually cherry pick?

PR #1469, opened 2026-01-13
ci: allow backporting directly from the github ui [no-ci]

cpcloud:
Add the ability to backport from the GitHub UI using workflow dispatch events.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -9,14 +9,25 @@ on:
     types: [closed, labeled]
     branches:
       - main
+  workflow_dispatch:
+    inputs:
+      backport-branch:
+        description: "Branch to backport commits onto"
+        required: false
+        type: string
+      pull-request:
+        description: "PR to backport"
+        required: true
+        type: number
+
 
 permissions:
   contents: write       # so it can comment
   pull-requests: write  # so it can create pull requests
 
 jobs:
-  backport:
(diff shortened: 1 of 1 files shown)


.github/workflows/backport.yml
```diff
@@ -38,3 +49,28 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8  # v6.0.1
+
+      - name: Load branch from environment name
+        if: inputs.backport-branch == null
+        run: |
+          BACKPORT_BRANCH=$(yq '.backport_branch' ci/versions.yml)
+          echo "BACKPORT_BRANCH=${BRANCH}" >> $GITHUB_ENV
+
```
leofang, line 64:
```suggestion
```

leofang: oh, shouldn't we revert the two steps here?
- inputs.backport-branch != null: use the input
- otherwise, look up the backport branch from yaml

cpcloud: Not sure I follow. They are mutually exclusive so only one will run at a time.

The order that they are listed shouldn't matter.

leofang: I see it now, it's what I missed!

PR #1478, opened 2026-01-13
CI: Add doc-only support to skip tests and non-essential builds

### LINKED ISSUE #1350: CI: Allow only triggering the doc-build pipeline
Currently `/ok to test <commit>` fires up the full CI. We should allow a doc-only trigger which skips most of the build pipelines (but the one that is used by the doc CI) and all of the test pipelines.

Andy-Jost:
## Summary
- Adds `[doc-only]` tag support in PR titles to skip test jobs and non-essential builds
- Allows faster iteration on documentation-only changes

## Changes
- Modified `should-skip` job to detect `[doc-only]` in PR titles
- Updated build jobs (`build-linux-aarch64`, `build-windows`) to skip when `[doc-only]` is present
- Updated test jobs to skip when `[doc-only]` is present
- Updated `checks` job to not fail when tests are intentionally skipped

## Test plan
- [x] Verify test jobs are skipped with `[doc-only]` in title
- [x] Verify `build-linux-64` and `doc` jobs still run
- [x] Verify `checks` job passes despite skipped tests
- [x] Remove `[doc-only]` and verify normal CI behavior

Closes #1350


Andy-Jost:
## Testing

To test this, I made the draft PR title `[doc-only] CI: Add doc-only support...` and ran CI. The results were as expected:

- ✅ `build-linux-64` ran (needed for docs)
- ⏭️ `build-linux-aarch64` and `build-windows` were skipped
- ⏭️ All test jobs were skipped
- ✅ `doc` job ran and passed
- ✅ `checks` job passed despite skipped tests

**Cosmetic issue:** Skipped jobs display with unexpanded variables in the GitHub UI (e.g., `Build ${{ matrix.host-platform }}, CUDA ${{ needs.ci-vars.outputs.CUDA_BUILD_VER }}`). This is a known GitHub Actions limitation—when a job is skipped, it cannot evaluate expressions that reference outputs from other jobs. The CI logic is correct; only the display name is affected.

Then, I removed `[doc-only]` from the title and re-ran CI to verify normal behavior still works (all jobs run, tests execute, etc.).

.github/workflows/ci.yml
```diff
@@ -121,11 +127,12 @@ jobs:
         host-platform:
           - linux-64
     name: Test ${{ matrix.host-platform }}
-    if: ${{ github.repository_owner == 'nvidia' }}
+    if: ${{ github.repository_owner == 'nvidia' && !fromJSON(needs.should-skip.outputs.doc-only) }}
```
leofang, line 130: Q: If it's doc-only, we don't need to run tests, ~~so I think this change can be dropped.~~ We currently do not run doctest anyway (we probably should, but it is not easy to set up).

My main concern is that we ignore the warnings across this file that require jobs to be kept in sync. (We manually split them and violated DRY so as to gain parallelism in the CI.)

.github/workflows/ci.yml
```diff
@@ -219,11 +231,18 @@ jobs:
...
-                 needs.doc.result == 'cancelled' }}; then
+          #
+          # Note: When [doc-only] is in PR title, test jobs are intentionally
+          # skipped and should not cause failure.
+          doc_only=${{ needs.should-skip.outputs.doc-only }}
+          if ${{ needs.doc.result == 'cancelled' }}; then
             exit 1
-          else
-            exit 0
           fi
```
leofang, line 240: nit: can we move this to as an `else` block below?

.github/workflows/ci.yml
```diff
@@ -50,11 +51,16 @@ jobs:
...
+            skip="$(echo "${pr_title}" | grep -q '\[no-ci\]' && echo true || echo false)"
+            doc_only="$(echo "${pr_title}" | grep -q '\[doc-only\]' && echo true || echo false)"
           else
             skip=false
+            doc_only=false
           fi
           echo "skip=${skip}" >> "$GITHUB_OUTPUT"
+          echo "doc_only=${doc_only}" >> "$GITHUB_OUTPUT"
 
   # WARNING: make sure all of the build jobs are in sync
```
leofang, line 65: I guess this particular job needs to stay intact. Let's update this warning to explain why it's out of sync from other build jobs below.

Andy-Jost: Does "in sync" here mean the code blocks follow a parallel structure?

PR #1489, opened 2026-01-14
fix: remove incorrect noexcept from resource handle functions

Andy-Jost:
## Summary
- Remove `noexcept` from C++ resource handle functions that can throw
- Update Cython declarations to use `nogil except+` for proper exception translation
- Ensure all custom deleters are exception-safe

### 1. Remove incorrect `noexcept` from throwing functions

Functions that allocate memory (via `new`, `std::make_shared`, or container operations) can throw `std::bad_alloc`. Marking them `noexcept` causes `std::terminate` on OOM instead of allowing graceful error handling.

**Remove `noexcept` from 22 functions:**
- All `create_*` functions (context, stream, event, mempool, deviceptr)
- All `deviceptr_alloc*` functions
- `get_primary_context`, `get_current_context`, `get_device_mempool`
- `get_legacy_stream`, `get_per_thread_stream`

**Keep `noexcept` on 5 functions that truly cannot throw:**
- `get_last_error`, `peek_last_error`, `clear_last_error`
- `deallocation_stream`, `set_deallocation_stream`

### 2. Update Cython declarations

Change from `noexcept nogil` to `nogil except+` for all throwing functions, allowing Cython to translate C++ exceptions to Python exceptions.

### 3. Fix exception-unsafe deleters

Custom `shared_ptr` deleters must not throw (undefined behavior per C++ standard). Two deleters had potential issues:

- **`clear_mempool_peer_access`**: Uses `std::vector` which can throw `std::bad_alloc`. Wrapped in try-catch.
- **IPC cache deleter**: Uses `std::lock_guard` which can throw `std::system_error`. Wrapped cache operations in try-catch.

### DIFF
--- a/cuda_core/cuda/core/_cpp/resource_handles.cpp
+++ b/cuda_core/cuda/core/_cpp/resource_handles.cpp
@@ -171,7 +171,7 @@ struct ContextBox {
 };
 }  // namespace
 
-ContextHandle create_context_handle_ref(CUcontext ctx) noexcept {
+ContextHandle create_context_handle_ref(CUcontext ctx) {
     auto box = std::make_shared<const ContextBox>(ContextBox{ctx});
     return ContextHandle(box, &box->resource);
 }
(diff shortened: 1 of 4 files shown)


cpcloud:
Should we add some cpp linting now that we have a significant amount of code to maintain there?

Not in this PR of course :)

Andy-Jost:
> Should we add some cpp linting now that we have a significant amount of code to maintain there?
> 
> Not in this PR of course :)

yes!

PR #1465, opened 2026-01-12
cuda.core.system: More device-related APIs

mdboom:
This is sort of a grab bag of more APIs for `cuda.core.system`, mainly related to InfoROM inspection and more detailed information about PCI topology.

## Checklist

- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #1492, opened 2026-01-14
cuda.core.system: affinity, clock, fan, temperature and thermals

mdboom:
# Prerequisites:

- #1465


cuda_core/cuda/core/system/_temperature.pxi
```diff
@@ -0,0 +1,138 @@
...
+    int defaultMinTemp
+    int defaultMaxTemp
+    int currentTemp
+    int target
+
+
+cdef class ThermalSensor:
+    cdef:
+        _ThermalSensor *_ptr
+        object _owner
```
cpcloud, line 20: non-blocking: is it specified anywhere that this lifetime relationship needs to be maintained? IOW, would it be incorrect to just let the usual reference counting behavior handle this?

mdboom: I should add a comment.  `ptr` is a pointer somewhere in the middle of the memory of a Numpy array held by `_owner`.  So if `_owner` were to be decref'd, `_ptr` would be dangling.

PR #1498, opened 2026-01-15
Skip test_field_values where unsupported field values are used

mdboom:
Found by SWQA on hardware that isn't in our CI.

### DIFF
--- a/cuda_core/tests/system/test_system_device.py
+++ b/cuda_core/tests/system/test_system_device.py
@@ -262,6 +262,8 @@ def test_persistence_mode_enabled():
 
 
 def test_field_values():
+    skip_reasons = set()
+
     for device in system.Device.get_all_devices():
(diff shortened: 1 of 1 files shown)


mdboom:
Note, this doesn't need a backport -- it's `cuda.core`-only.

rwgk:
I'm concerned about the `skip_reasons` getting lost when something goes wrong in any of the iterations over the loop. I think we should fix that in a follow-on PR.

mdboom:
> I'm concerned about the `skip_reasons` getting lost when something goes wrong in any of the iterations over the loop. I think we should fix that in a follow-on PR.

Yes, though having multiple devices is extremely rare.  I'm starting to come around that keeping track of multiple skip reasons isn't worth the effort, and it's fine to just stop at the first occurrence.

PR #1500, opened 2026-01-15
release gated on docs build [no-ci]

### LINKED ISSUE #1499: release workflow fails without a fully specified CTK, but other parts of it still succeed, putting the release in an inconsistent state
The release workflow fails if you don't fully specify the CTK version: https://github.com/NVIDIA/cuda-python/actions/runs/21036962438/job/60487737202.

We should figure out a way to either build the docs without a CTK installed, or somehow prevent this from happening.

cpcloud:
Closes #1499

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -47,6 +47,12 @@ jobs:
       run:
         shell: bash -el {0}
     steps:
+      - name: validate build-ctk
+        run: |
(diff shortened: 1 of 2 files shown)


leofang (approved):
This is fine. I think `inputs.build-ctk-ver` can be removed now (and we won't need to type it during release), but we can do it in a separate PR.

PR #1502, opened 2026-01-15
Remove context dependency from __hash__ and __eq__

### LINKED ISSUE #1480: `__hash__`, `__eq__`, ... should not require a CUDA context
Example:
https://github.com/NVIDIA/cuda-python/blob/39e32931697e6c04f19e22efadd25d4f071b254e/cuda_core/cuda/core/_stream.pyx#L183-L186

@stiepan reported offline that this is problematic because at least for streams we do not always know the CUDA context associated with them (we could be wrapping a foreign stream), and triggering a lookup is expensive. Some of the objects are even independent of CUDA contexts.  

We decided to follow cccl-rt and do a simple pointer comparison for `__eq__`.  We should stick to this principle for `__hash__` as well.

Andy-Jost:
## Summary

- Simplify `Stream.__hash__` and `Stream.__eq__` to use pointer comparison only, removing expensive `Stream_ensure_ctx` calls
- Fix `Event.__hash__` to be consistent with `Event.__eq__` by removing context from hash

## Changes

- `_stream.pyx`: Remove `Stream_ensure_ctx` calls and context comparison from `__hash__` and `__eq__`
- `_event.pyx`: Remove `_h_context` from `__hash__` (was inconsistent with `__eq__` which only compares event handle)

## Motivation

As reported in #1480, requiring CUDA context for hash/equality is problematic because:
1. For foreign streams, we may not know the associated CUDA context
2. Triggering context lookup is expensive
3. Some objects are independent of CUDA contexts

This follows the cccl-rt principle of simple pointer comparison.

Closes #1480

### DIFF
--- a/cuda_core/cuda/core/_event.pyx
+++ b/cuda_core/cuda/core/_event.pyx
@@ -172,7 +172,7 @@ cdef class Event:
             raise RuntimeError(explanation)
 
     def __hash__(self) -> int:
-        return hash((type(self), as_intptr(self._h_context), as_intptr(self._h_event)))
+        return hash((type(self), as_intptr(self._h_event)))
 
     def __eq__(self, other) -> bool:
         # Note: using isinstance because `Event` can be subclassed.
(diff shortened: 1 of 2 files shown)


Andy-Jost:
The associated issue only mentions streams, but I applied the same fix to events as well.

PR #1503, opened 2026-01-15
ci: remove build-ctk-ver from doc builds

cpcloud:
Putting this up for review in case there is some case I am missing, but I do not see anything that indicates we need to care about the version, even in artifact download. AFAIS there is only a single wheel produce per cuda version for a given run-id.

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -7,9 +7,6 @@ name: "CI: Build and update docs"
 on:
   workflow_call:
     inputs:
-      build-ctk-ver:
-        type: string
-        required: true
       component:
         description: "Component(s) to build docs for"
         required: false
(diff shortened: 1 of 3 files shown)


.github/workflows/build-docs.yml
```diff
@@ -103,7 +87,7 @@ jobs:
           echo "CUDA_CORE_ARTIFACT_BASENAME=${CUDA_CORE_ARTIFACT_BASENAME}" >> $GITHUB_ENV
           echo "CUDA_CORE_ARTIFACT_NAME=${CUDA_CORE_ARTIFACT_BASENAME}-${FILE_HASH}" >> $GITHUB_ENV
           echo "CUDA_CORE_ARTIFACTS_DIR=$(realpath "$REPO_DIR/cuda_core/dist")" >> $GITHUB_ENV
-          CUDA_BINDINGS_ARTIFACT_BASENAME="cuda-bindings-python${PYTHON_VERSION_FORMATTED}-cuda${{ inputs.build-ctk-ver }}-linux-64"
+          CUDA_BINDINGS_ARTIFACT_BASENAME="cuda-bindings-python${PYTHON_VERSION_FORMATTED}-cuda*-linux-64"
```
leofang, line 90: Ah ha! So we needed the CTK ver also to infer the artifact name... hmmm... I will need to sleep over this.

cpcloud: ping @leofang, any thoughts after sleeping on it 😴

PR #1508, opened 2026-01-16
cuda.core.system: add conveniences to convert between device types

mdboom:
This adds two convenience methods to convert between `cuda.core.Device` and `cuda.core.system.Device`.

I have a few design questions for this one:

- Is the naming of `Device.to_system_device` and `Device.to_cuda_device` ok?
- This could be a new type accepted by the constructor, e.g. `cuda.core.Device(nvml_device)`, but that requires importing the other module from the constructor which could be a performance / cyclical import nightmare.

### DIFF
--- a/cuda_core/cuda/core/system/_device.pyx
+++ b/cuda_core/cuda/core/system/_device.pyx
@@ -1060,6 +1060,9 @@ cdef class Device:
         Retrieves the globally unique immutable UUID associated with this
         device, as a 5 part hexadecimal string, that augments the immutable,
         board serial identifier.
+
+        In the upstream NVML C++ API, the UUID includes a ``gpu-`` or ``mig-``
+        prefix.  That is not included in ``cuda.core.system``.
         """
         # NVML UUIDs have a `GPU-` or `MIG-` prefix.  We remove that here.
 


cuda_core/cuda/core/_device.pyx
```diff
@@ -1034,6 +1034,17 @@ class Device:
...
+    def to_system_device(self) -> 'cuda.core.system.Device':
+        """
+        Get the corresponding :class:`cuda.core.system.Device` (which is used
+        for NVIDIA Machine Library (NVML) access) for this
+        :class:`cuda.core.Device` (which is used for CUDA access).
+
+        The devices are mapped to one another by their UUID.
+        """
+        from cuda.core.system import Device as SystemDevice
+        return SystemDevice(uuid=self.uuid)
```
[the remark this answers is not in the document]
mdboom: Thanks, but testing shows this works without it.

cuda_core/cuda/core/system/_device.pyx
```diff
@@ -722,6 +722,36 @@ cdef class Device:
...
+        # this matching when it has a `GPU-` prefix, but for now we just strip
+        # it.  If a matching CUDA device can't be found, we will get a helpful
+        # exception, anyway, below.
+        uuid = self.uuid[4:]
+
+        for cuda_device in CudaDevice.get_all_devices():
+            if cuda_device.uuid == uuid:
+                return cuda_device
+
+        raise RuntimeError("No corresponding CUDA device found for this NVML device.")
```
cpcloud, line 747: More of a rant I guess than anything: I know this is pre-existing, but I really dislike that we raise `RuntimeError` everywhere.

cuda_core/cuda/core/system/_device.pyx
```diff
@@ -722,6 +722,36 @@ cdef class Device:
...
+        from cuda.core import Device as CudaDevice
+
+        # CUDA does not have an API to get a device by its UUID, so we just
+        # search all the devices for one with a matching UUID.
+
+        # NVML UUIDs have a `GPU-` or `MIG-` prefix.  Possibly we should only do
+        # this matching when it has a `GPU-` prefix, but for now we just strip
+        # it.  If a matching CUDA device can't be found, we will get a helpful
+        # exception, anyway, below.
+        uuid = self.uuid[4:]
```
cpcloud: Even though these are two different `Device` APIs, having their UUIDs not be identical feels like a design flaw. Can we fix that or is there a specific reason they must be this way?

Is the prefix actually providing unique information?

For example, is it ever the case where you have `GPU-$UUID` and `MIG-$UUID` and `$UUID` are equal?

Keeping the prefix also makes this not actually a UUID and therefore incompatible with any tool that consumes UUIDs directly. Every use has to be sliced out before being used.

cpcloud (approved):
Would really like to see if we can address the UUID inconsistency, but that isn't blocking and can be done in a follow-up.

cuda_core/cuda/core/system/_device.pyx
leofang: UUID is a complex mess, see nvbugs 4868877. Is there a way we can avoid using UUID while still allowing CUDA and NVML to handshake?

cuda_core/cuda/core/_device.pyx
```diff
@@ -1034,6 +1034,29 @@ class Device:
         total = system.get_num_devices()
         return tuple(cls(device_id) for device_id in range(total))
 
+    def to_system_device(self) -> 'cuda.core.system.Device':
```
leofang, line 1037: Throwing out some ideas:
1. We start a method with a verb if a new instance of some resource is returned (ex: `create_stream`)
2. We start a method with a noun if an 1:1 resource associated with self is returned (ex: `context`, `memory_resource`)

It seems 2 is closer to what we need here? Is it correct?

cuda_core/cuda/core/system/_device.pyx
```diff
@@ -722,6 +722,36 @@ cdef class Device:
                 pci_bus_id = pci_bus_id.decode("ascii")
             self._handle = nvml.device_get_handle_by_pci_bus_id_v2(pci_bus_id)
 
+    def to_cuda_device(self) -> "cuda.core.Device":
```
leofang, line 725: ditto

cuda_core/cuda/core/_device.pyx
cpcloud: IMO nouns:

- should be cheap to access
- suggest that the instance may contain them

A `to_*` method on the other hand suggests there might be some non-trivial overhead and indicates clearly that the instance doesn't contain the noun.

-1 for calling this just `system_device`.

cuda_core/cuda/core/system/_device.pyx
mdboom: What would be more appropriate here?  A custom exception along the lines of `DeviceNotFoundError`?

cuda_core/cuda/core/_device.pyx
mdboom: Yeah, I mostly agree with @cpcloud here.  I debated whether to use `to` or `as`.  `as` may be closer since there is the 1:1 relationship.

cuda_core/cuda/core/system/_device.pyx
mdboom: > For example, is it ever the case where you have GPU-$UUID and MIG-$UUID and $UUID are equal?

I don't /know/ but I really doubt it.  I'd be fine with truncating the GPU and MIG prefix in NVML (and adding another method to obtain it for users who might care).  That's probably better than requiring truncation by our users.

> Is there a way we can avoid using UUID while still allowing CUDA and NVML to handshake?

Using indices is clearly not the right choice -- there are envvars that control their assignment and a bunch of other things that make them unsuitable.

The NVML docs say:

```
The order in which NVML enumerates devices has no guarantees of consistency between reboots. For that reason it
is recommended that devices be looked up by their PCI ids or UUID. See
nvmlDeviceGetHandleByUUID() and nvmlDeviceGetHandleByPciBusId_v2().
```

I chose UUID because /technically speaking/ some really old devices and maybe some future devices may not actually be on a PCI bus.  And PCI bus ids are inconsistent between CUDA and NVML (in more problematic ways, such as CUDA only exposing the lower 16-bits of the domain).

cuda_core/cuda/core/_device.pyx
leofang: okay, resolving, thx

cuda_core/cuda/core/system/_device.pyx
leofang: resolving, thx

cuda_core/cuda/core/system/_device.pyx
```diff
@@ -1036,8 +1060,16 @@ cdef class Device:
...
+        In the upstream NVML C++ API, the UUID includes a ``gpu-`` or ``mig-``
+        prefix.  That is not included in ``cuda.core.system``.
         """
-        return nvml.device_get_uuid(self._handle)
+        # NVML UUIDs have a `GPU-` or `MIG-` prefix.  We remove that here.
+
+        # TODO: If the user cares about the prefix, we will expose that in the
+        # future using the MIG-related APIs in NVML.
+
+        return nvml.device_get_uuid(self._handle)[4:]
```
leofang, line 1072: I am thinking the system device should return the full UUID and we document the different expectations between cuda.core and cuda.core.system (or CUDA vs NVML). @mdboom thoughts?

mdboom: I totally can see it both ways.  My original implementation did was you suggested (following upstream NVML behavior).  But @cpcloud convinced me this is weird -- UUID has a well-defined meaning in our field that NVML deviates from.  I don't feel super strongly either way -- we just need to break the tie ;)

leofang: IIUC NVML is the only way for us to tell, from inside a running process, if we are using MIG instances or otherwise (bare-metal GPU, MPS, etc). CUDA purposely hides MIG from end users. So my thinking is if we don't follow NVML there is no other way for Python users to query. Could you check if my impression is correct?

mdboom: There is another API, `nvmlDeviceIsMigDeviceHandle`, that could be used to query whether it's MIG, and IMHO, that's better than the user needing to parse a string to get that info.

```
/**
 * Test if the given handle refers to a MIG device.
 *
 * A MIG device handle is an NVML abstraction which maps to a MIG compute instance.
 * These overloaded references can be used (with some restrictions) interchangeably
 * with a GPU device handle to execute queries at a per-compute instance granularity.
 *
 * For Ampere &tm; or newer fully supported devices.
 * Supported on Linux only.
 *
 * @param device                               NVML handle to test
 * @param isMigDevice                          True when handle refers to a MIG device
 */
nvmlReturn_t DECLDIR nvmlDeviceIsMigDeviceHandle(nvmlDevice_t device, unsigned int *isMigDevice);
```

cuda_core/cuda/core/system/_device.pyx
```diff
@@ -722,6 +722,30 @@ cdef class Device:
...
+        -------
+        cuda.core.Device
+            The corresponding CUDA device.
+        """
+        from cuda.core import Device as CudaDevice
+
+        # CUDA does not have an API to get a device by its UUID, so we just
+        # search all the devices for one with a matching UUID.
+
+        for cuda_device in CudaDevice.get_all_devices():
```
rparolin, line 743: Consider: Can we memoize this call so it only does the linear search once and caches the result?

mdboom: Probably in most cases that would be fine, but hot plugging devices is possible, so the device with the same UUID can change device handles after a replug.  Hot plugging a device during a running process that cares about it probably has all kinds of other problems, but since we can't predict how the application works, it's probably better putting caching in the hands of the user of this API.

PR #1509, opened 2026-01-16
cuda.bindings.nvml: Simplify the "#defines as enums" naming

mdboom:
cuda.bindings.nvml has a bunch of enums which are written by hand because they are `#defines` in the nvml header.  In hindsight, these didn't all need full names, but the common prefixes of the individual values can be removed.  (Since Python enums are a namespace, unlike C, the prefixes aren't necessary).


leofang:
Good catch!

Issue #716, opened 2025-06-18
[FEA]: Support for finding `libcudadevrt.a` through pathfinder


brandon-b-miller:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

cuda.bindings

### Is your feature request related to a problem? Please describe.

`numba-cuda` needs to [be able to find `libcudadevrt.a`](https://github.com/NVIDIA/numba-cuda/blob/e86aeab63d0aaa468abc0529f06b63550c42298a/numba_cuda/numba/cuda/codegen.py#L261-L264) to launch kernels involving cooperative groups, but needs to [find it itself via `get_cudalib`](https://github.com/NVIDIA/numba-cuda/blob/main/numba_cuda/numba/cuda/cudadrv/libs.py#L46-L61). Ideally the pathfinder could provide this for us. 

### Describe the solution you'd like

Something like `path_finder._get_nvidia_static_library("cudadevrt")`

### Describe alternatives you've considered

Numba's existing logic 

### Additional context

https://github.com/NVIDIA/numba-cuda/issues/302

leofang:
> `numba-cuda` needs to [be able to find `libcudadevrt.a`](https://github.com/NVIDIA/numba-cuda/blob/e86aeab63d0aaa468abc0529f06b63550c42298a/numba_cuda/numba/cuda/codegen.py#L261-L264) to launch kernels involving cooperative groups

@brandon-b-miller could you confirm if this is true? AFAIK cooperative launch does not require libcudadevrt. The only CUDA feature that needs it is CDP1 and CDP2 (CUDA Dynamic Parallelism).

brandon-b-miller:
In the relevant codepath, there's this note in the numba-cuda source code:

```
# We need to link against cudadevrt if grid sync is being used.
```

https://github.com/NVIDIA/numba-cuda/blob/main/numba_cuda/numba/cuda/dispatcher.py#L157-L161

leofang:
That's why I would like us to test it locally before determining the priority, because we have the same `grid.sync()` test in the CI and I am pretty sure libcudadevrt is not needed for CG, only for CDP1/2 (which I added to CuPy) 🙂 
[...]

PR #1512, opened 2026-01-16
Reorganize graph tests and add device-side launch support

### LINKED ISSUE #1331: CUDA graph phase N - Device-side graph launch
This task should be simple: 
- Allow users to opt-in and internally pass `cudaGraphInstantiateFlagDeviceLaunch` to `cudaGraphInstantiate()`
- Add code samples to showcase how this can done

The majority of work would happen on the JIT compiler side, aka numba-cuda; cuda-core will just set up the infrastructure.

Andy-Jost:
## Summary

This PR reorganizes the graph tests into a new `tests/graph/` subdirectory, adds a public `Graph.handle` property, and introduces tests for device-side graph launch.

Closes #1331

### Test Reorganization
- Create new `tests/graph/` subdirectory following the pattern of `tests/memory_ipc/`
- Break apart `test_graph.py` into logical groupings:
  - `test_basic.py`: basic graph construction and topology tests
  - `test_conditional.py`: conditional node tests (if, if-else, switch, while)
  - `test_advanced.py`: child graphs, update, stream lifetime
  - `test_options.py`: debug print, complete options, build mode
- Incorporate `test_graph_mem.py` (renamed to `test_capture_alloc.py`)
- Add shared kernel compilation helpers in `tests/helpers/graph_kernels.py`

### API Addition
- Add `Graph.handle` property to expose the underlying `CUgraphExec` handle
- Follows the existing pattern of `Stream.handle` and `Event.handle`
- Forward-compatible with future RAII handle refactoring

### Device-Side Graph Launch Tests
- Add `test_device_launch.py` with tests for device-side graph launch
- Uses `cudaGraphLaunch()` from device code via `cudaStreamGraphTailLaunch`
- Requires Hopper (sm_90+) architecture
- Uses `cuda.pathfinder.find_nvidia_header_directory()` to locate `libcudadevrt.a`

## Test Coverage

- 58 tests total in `tests/graph/`
- All tests pass on Hopper (H200)
- Device launch tests skip on pre-Hopper architectures


cuda_core/tests/graph/test_advanced.py
Andy-Jost, line 1: Moved from test_graph.py

cuda_core/tests/graph/test_basic.py
Andy-Jost, line 1: Moved from test_graph.py

cuda_core/tests/graph/test_conditional.py
Andy-Jost, line 1: Moved from test_graph.py

cuda_core/tests/graph/test_device_launch.py
Andy-Jost, line 1: New code

cuda_core/tests/graph/test_options.py
Andy-Jost, line 1: Moved from test_graph.py

cuda_core/tests/helpers/graph_kernels.py
Andy-Jost, line 1: Moved from test_graph.py

cuda_core/tests/graph/test_device_launch.py
```diff
@@ -0,0 +1,214 @@
...
+        path = os.path.join(base, "lib", "x64", "cudadevrt.lib")
+        if os.path.isfile(path):
+            return path
+    else:
+        # Linux: .so and .a are both in lib/
+        path = os.path.join(lib_dir, "libcudadevrt.a")
+        if os.path.isfile(path):
+            return path
+
+    return None
```
Andy-Jost: I'm not sure of the best way to locate the `cudadevrt` library (needed for device launch). I'm open to suggestions here.

rwgk: We have the existing requirement that `CUDA_HOME` needs to be set when running the cuda-core tests: did you consider finding the `.lib`/`.a` files relative to `CUDA_HOME`? I'm guessing that's probably pretty simple.

In any case, could you please reference #716 from the helper function, and maybe add a comment under 716 to point back to the helper function, after this PR is merged?

If you want to keep the current implementation:

```python
    try:
        loaded = load_nvidia_dynamic_lib("cudart")
    except DynamicLibNotFoundError:
        return None
```

That's far less likely to mask bugs.

cuda_core/cuda/core/_graph.py
```diff
@@ -746,6 +746,18 @@ def close(self):
         """Destroy the graph."""
         self._mnff.close()
 
+    @property
+    def handle(self) -> driver.CUgraphExec:
```
leofang, line 751: How do we decide if we want to expose `CUgraph` or `CUgraphExec`? Both can find their use cases.

Andy-Jost: It's a great question that I tried to sidestep in this change. (I'm working on another change to add resource handles to the graph module.)

My understanding is that `class GraphBuilder` corresponds to `CUgraph` and `class Graph` corresponds to `CUgraphExec`.

Following the existing pattern, after moving to resource handles, we would update the internal names to clarify:

- `GraphBuilder._mnff.graph` → `GraphBuilder._mnff._h_graph` (`CUgraph`)
- `Graph._mnff.graph` → `Graph._mnff._h_graph_exec` (`CUgraphExec`)

And the property types would be:

- `GraphBuilder.handle -> driver.CUgraph`
- `Graph.handle -> driver.CUgraphExec`


FYI, I added this property to expose `CUgraphExec` for the following code in `test_device_launch.py`:

```
inner_graph_handle = int(inner_graph.handle)
...
launch(gb_outer, LaunchConfig(grid=1, block=1), launcher_kernel, inner_graph_handle)
```

mdboom: I admit to not having a lot of context on the graph APIs specifically -- but my sense is that `cuda.core` exists largely to abstract away and hide raw handles and we should be discouraging end users to deal with them.  Since this only exists to support testing, maybe it should be `_handle`?

cuda_core/tests/graph/test_device_launch.py
```diff
@@ -0,0 +1,214 @@
...
+    if cudadevrt_path is None:
+        pytest.skip("libcudadevrt.a not found (set CONDA_PREFIX, CUDA_HOME, or CUDA_PATH)")
+
+    code = """
+    extern "C" __global__ void launch_graph_from_device(cudaGraphExec_t graph) {
+        if (threadIdx.x == 0 && blockIdx.x == 0) {
+            cudaGraphLaunch(graph, cudaStreamGraphTailLaunch);
+        }
+    }
+    """
```
Andy-Jost, line 93: FYI: CUDA device-side graph launch

cuda_core/cuda/core/_graph.py
Andy-Jost: Users pass this handle to kernels that perform device-side launch, so if they are using that feature they will need the handle; see https://github.com/NVIDIA/cuda-python/pull/1512/changes#r2708906879.

To the broader point, do we want to expose handles from `cuda.core` in general as an escape hatch (e.g., so users can invoke the Driver API directly)? So far, we've been doing this consistently for other classes such as `Stream`, `Event`, etc.

I like this approach because it means users are not limited by `cuda.core`, which it seems will always lag behind the driver in terms of functionality.

cuda_core/tests/graph/test_device_launch.py
Andy-Jost: Thanks for noting that `CUDA_HOME` is required -- I went with that option.

cuda_core/cuda/core/_graph.py
Andy-Jost: For ref, here is a summary of the `handle` properties available in Python. We can discuss whether this is the best design, but it is the established practice that this PR follows.

**CUDA Driver Resource Handles in cuda.core**

  | Python Class        | Property | CUDA Handle Type                | Notes                         |
  |---------------------|----------|---------------------------------|-------------------------------|
  | Context             | handle   | CUcontext                       | Primary/current context       |
  | Stream              | handle   | CUstream                        | Stream for async operations   |
  | Event               | handle   | CUevent                         | Timing/synchronization events |
  | Buffer              | handle   | CUdeviceptr                     | Device/pinned memory pointer  |
  | _MemPool            | handle   | CUmemoryPool                    | Memory pool for async alloc   |
  | Kernel              | handle  | CUkernel                        | Compiled kernel    |
  | ObjectCode          | handle   | CUlibrary                       | Loaded code module            |
  | Program             | handle   | nvrtcProgram or driver handle   | Backend-dependent             |
  | Linker              | handle   | CUlinkState or nvJitLink handle | Backend-dependent             |
  | IPCAllocationHandle | handle   | int                             | IPC handle (raw int)          |

Andy-Jost:
@mdboom Please take another look when you get a chance.

cuda_core/cuda/core/_graph.py
```diff
@@ -747,6 +747,18 @@ def close(self):
...
+        """Return the underlying ``CUgraphExec`` object.
+
+        .. caution::
+
+            This handle is a Python object. To get the memory address of the underlying C
```
mdboom, line 756: Who is this comment for?  People using this library from Cython?  It seems like Python programmers should just treat it like an opaque object and not ever call `int` on it.  Or am I missing something?

mdboom (approved):
LGTM -- other than my one question about the docstring.

PR #1514, opened 2026-01-20
Update copyright year with a precommit hook

mdboom:
It is hard to remember that the copyright year should be updated when there are changes to files.  This modifies our existing SPDX pre-commit hook to fix the latest copyright year.  

Locally, the pre-commit hook only runs on modified files, and we explicitly do not want to do a mass update of every file, especially ones that haven't changed in a while.

HOWEVER, since the pre-commit CI check runs on every file, it does require that every file is compliant with the rule, which is why there is an update of many files here.  It shows how many files we have forgotten to update in the past -- but once we do this mass update we should never have to do another one going forward since this rule will help us remember.

EDIT: There is actually a really hilarious cyclical problem with doing it that way.  You can ask it to update the copyright years based on the last known commit.  But that creates a commit in this year, so it will immediately want to update to this year.  I don't think that's the behavior we want, so I've modified this to only check the last year on /staged/ files.  It means the year check doesn't actually run as part of CI, but I think that's fine.  And it means no mass updating of files at this time -- just correctness going forward as things get naturally updated.

### DIFF
--- a/toolshed/check_spdx.py
+++ b/toolshed/check_spdx.py
@@ -37,24 +37,14 @@ COPYRIGHT_SUB = r"Copyright (c) {} \g<affiliation>"
 CURRENT_YEAR = str(datetime.date.today().year)
 
 
-def get_last_modified_year(filepath):
+def is_staged(filepath):
     # If the file is staged, we need to update it to the current year
     process = subprocess.run(  # noqa: S603
         ["git", "diff", "--staged", "--", filepath],  # noqa: S607
         capture_output=True,
         text=True,
     )
-    if process.stdout.strip():
-        return CURRENT_YEAR
-
-    # If the file is not staged, get the year of the last commit
-    process = subprocess.run(  # noqa: S603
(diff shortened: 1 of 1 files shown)


toolshed/check_spdx.py
```diff
@@ -26,29 +29,88 @@ def load_spdx_ignore():
...
+        match = re.search(COPYRIGHT_REGEX, blob)
+        if match is None:
+            print(f"MISSING valid copyright line in {filepath!r}")
+            good = False
+            continue
+
+        years = match.group("years").decode()
+        if "-" in years:
+            start_year, end_year = years.split("-", 1)
+            if int(start_year) > int(end_year):
```
rwgk, line 72: To catch typos, WDYT about adding

`or int(end_year) > int(CURRENT_YEAR)`

here?

I'd probably make this

```
        if int(start_year) >= int(end_year) or int(end_year) > int(CURRENT_YEAR):
```

to catch things like `2026-2026` as well (when someone maybe wanted to write `2025-2026` but got mixed up).

PR #1515, opened 2026-01-20
cuda.bindings.nvml: Correctly handle a test requiring root

mdboom:
The `cuda.bindings.nvml` tests were written with the assumption that they would be run by a non-root user (as is the case in our CI).  SWQA seemingly is running some tests as root, which then causes the test to fail.  This change makes the test robust to either situation.

### DIFF
--- a/cuda_bindings/tests/nvml/test_compute_mode.py
+++ b/cuda_bindings/tests/nvml/test_compute_mode.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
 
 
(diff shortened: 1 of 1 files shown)


mdboom:
This is proving to be a challenge.  On some of our CI runners, all signs indicate that the user is the root user (uid == 0, a member of gid == 0) etc., but the API call still fails as `NoPermissionError`.  I suspect this is because the user is root inside a container, and the actual thing (device handle, maybe) that this API is trying to access requires root outside of the container.  

We know, however, that SWQA has able to run as root and get this API call to pass (which then caused the original test to fail).

I'm thinking I'll just rewrite the test so that it will be robust to the call failing or passing regardless of the user's status, but that's a workaround.

Anyone else run into this issue and have a better solution?

kkraus14:
@mdboom since nvml does some pretty low level system interactions, things can be somewhat arbitrarily blocked off via things like seccomp policies even if we can detect we're in a container or anything else reasonable.

I think your approach of treating getting `nvml.NoPermissionError` as either a skip or a success sounds reasonable to me. I don't know if it makes sense for us to chase marking tests for it versus just having a blanket policy.

cuda_bindings/tests/nvml/test_compute_mode.py
```diff
@@ -25,8 +25,11 @@ def test_compute_mode_supported_nonroot(all_devices):
             continue
 
         for cm in COMPUTE_MODES:
-            with pytest.raises(nvml.NoPermissionError):
+            try:
                 nvml.device_set_compute_mode(device, cm)
+            except nvml.NoPermissionError:
+                skip_reasons.add(f"nvmlDeviceSetComputeMode requires root for device {device}")
```
leofang, line 31: I don't know how this got passed the CI, but it failed on main:
```
                except nvml.NoPermissionError:
>                   skip_reasons.add(f"nvmlDeviceSetComputeMode requires root for device {device}")
                    ^^^^^^^^^^^^
E                   NameError: name 'skip_reasons' is not defined
```
https://github.com/NVIDIA/cuda-python/actions/runs/21307037717/job/61337052365#step:26:510

PR #1516, opened 2026-01-20
fix: support zero sized allocation

cpcloud:
Fixed zero-sized allocations, which fail with the TCC windows driver mode.

### DIFF
--- a/cuda_core/tests/test_memory.py
+++ b/cuda_core/tests/test_memory.py
@@ -1255,7 +1255,7 @@ def test_memory_resource_alloc_zero_bytes(init_cuda, memory_resource_factory):
         mr = MR()
     elif MR is ManagedMemoryResource:
         skip_if_managed_memory_unsupported(device)
-        mr = MR(MROps(preferred_location=device.device_id))
+        mr = create_managed_memory_resource_or_skip(MROps(preferred_location=device.device_id))
     else:
         assert MR is DeviceMemoryResource
         mr = MR(device)


cuda_core/tests/test_device.py
```diff
@@ -47,6 +47,16 @@ def test_device_alloc(deinit_cuda):
     assert buffer.device_id == int(device)
 
 
+def test_device_alloc_zero_bytes(deinit_cuda):
```
kkraus14, line 50: Is there any way to extend these tests to all of the different memory resources?

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -455,7 +455,19 @@ void set_deallocation_stream(const DevicePtrHandle& h, StreamHandle h_stream) no
...
+            new DevicePtrBox{ptr, h_stream},
+            [h_pool](DevicePtrBox* b) {
+                GILReleaseGuard gil;
+                delete b;
+            }
+        );
+        return DevicePtrHandle(box, &box->resource);
+    }
+
     if (CUDA_SUCCESS != (err = p_cuMemAllocFromPoolAsync(&ptr, size, *h_pool, as_cu(h_stream)))) {
```
kkraus14, line 459: The docs don't indicate that allocating 0 bytes from this API (`cuMemAllocFromPoolAsync`, `cuMemAllocAsync`) isn't allowed and from a local test (on a WSL environment) it doesn't error, whereas `cuMemAlloc` does.

Are we sure we need to guard these?

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -509,7 +545,19 @@ DevicePtrHandle deviceptr_alloc(size_t size) {
...
+            new DevicePtrBox{reinterpret_cast<CUdeviceptr>(ptr), StreamHandle{}},
+            [](DevicePtrBox* b) {
+                GILReleaseGuard gil;
+                delete b;
+            }
+        );
+        return DevicePtrHandle(box, &box->resource);
+    }
+
     if (CUDA_SUCCESS != (err = p_cuMemAllocHost(&ptr, size))) {
```
kkraus14, line 513: Similarly, the docs don't indicate allocating size 0 is an error for this API and a local test on WSL returns successfully.

cuda_core/cuda/core/_cpp/resource_handles.cpp
cpcloud: I don't think we do.

cuda_core/tests/test_device.py
cpcloud: Done.

cuda_core/cuda/core/_cpp/resource_handles.cpp
cpcloud: Reverted the cpp changes.

cuda_core/cuda/core/_cpp/resource_handles.cpp
cpcloud: Done.

PR #1517, opened 2026-01-20
CI: run cuda.bindings examples on Linux and Windows

### LINKED ISSUE #697: `cuda-bindings` examples are not run as part of the CI
By inspecting @rwgk's test plan renewal I realized we missed this when we modernized the test and CI infrastructures. We need to follow `cuda-core` and make them tested through pytest:
https://github.com/NVIDIA/cuda-python/tree/main/cuda_core/tests/example_tests
Then we don't need to run them as a separate test job.

rwgk:
Closes #697, #1525

This PR is motivated by nvbug 5808967 / #1525 — currently there is no automatic testing for the cuda_bindings/examples at all, therefore it is possible that the QA team is side-tracked unnecessarily by failures that we can discover in the CI here automatically.

This PR enables running the cuda_bindings/examples in wheel-based test environments, which is both a gain and a simplification (see changes in `cuda_bindings/examples/common/common.py`).

Non-goal for this PR: structural changes to run the examples in various environments or in different ways.

Note that `scripts/run_tests.sh` runs the examples by default. This PR makes `scripts/run_tests.sh` succeed in local wheel-based environments (except for one unrelated failure in cuda_core, please ignore for the purpose of this PR):

* before this PR: [scripts_run_tests_main_log_2026-01-22+093402.txt](https://github.com/user-attachments/files/24804869/scripts_run_tests_main_log_2026-01-22%2B093402.txt)

```
$ grep 'FAILED examples' scripts_run_tests_main_log_2026-01-22+093402.txt
FAILED examples/2_Concepts_and_Techniques/streamOrderedAllocation_test.py::main
FAILED examples/3_CUDA_Features/simpleCudaGraphs_test.py::main - AssertionError
FAILED examples/0_Introduction/clock_nvrtc_test.py::main - AssertionError
FAILED examples/0_Introduction/simpleZeroCopy_test.py::main - AssertionError
FAILED examples/0_Introduction/simpleCubemapTexture_test.py::main - Assertion...
FAILED examples/0_Introduction/systemWideAtomics_test.py::main - AssertionError
FAILED examples/0_Introduction/vectorAddDrv_test.py::main - AssertionError
FAILED examples/0_Introduction/vectorAddMMAP_test.py::main - AssertionError
```

* with this PR: [scripts_run_tests_pr1517_log_2026-01-22+100143.txt](https://github.com/user-attachments/files/24804874/scripts_run_tests_pr1517_log_2026-01-22%2B100143.txt)

```
$ grep 'FAILED examples' scripts_run_tests_pr1517_log_2026-01-22+100143.txt
```
(no output)

___

For completeness: I was hoping that our CI reproduces the Python 3.14t failure reported under nvbug 5808967, but it does not. But at least we know now, which helps in the search for the root cause.

### DIFF
--- a/ci/test-matrix.yml
+++ b/ci/test-matrix.yml
@@ -17,62 +17,70 @@
 
 linux:
   pull-request:
+    # linux-64
     - { ARCH: 'amd64', PY_VER: '3.10',  CUDA_VER: '12.9.1', LOCAL_CTK: '1', GPU: 'v100',       GPU_COUNT: '1', DRIVER: 'latest' }
     - { ARCH: 'amd64', PY_VER: '3.10',  CUDA_VER: '13.0.2', LOCAL_CTK: '0', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest' }
-    - { ARCH: 'amd64', PY_VER: '3.10',  CUDA_VER: '13.1.0', LOCAL_CTK: '0', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest' }
+    - { ARCH: 'amd64', PY_VER: '3.10',  CUDA_VER: '13.1.1', LOCAL_CTK: '0', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest' }
     - { ARCH: 'amd64', PY_VER: '3.11',  CUDA_VER: '12.9.1', LOCAL_CTK: '0', GPU: 'rtxpro6000', GPU_COUNT: '1', DRIVER: 'latest' }
     - { ARCH: 'amd64', PY_VER: '3.11',  CUDA_VER: '13.0.2', LOCAL_CTK: '1', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest' }
-    - { ARCH: 'amd64', PY_VER: '3.11',  CUDA_VER: '13.1.0', LOCAL_CTK: '1', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest' }
+    - { ARCH: 'amd64', PY_VER: '3.11',  CUDA_VER: '13.1.1', LOCAL_CTK: '1', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest' }
     - { ARCH: 'amd64', PY_VER: '3.12',  CUDA_VER: '12.9.1', LOCAL_CTK: '1', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest' }
     - { ARCH: 'amd64', PY_VER: '3.12',  CUDA_VER: '13.0.2', LOCAL_CTK: '0', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest' }
-    - { ARCH: 'amd64', PY_VER: '3.12',  CUDA_VER: '13.1.0', LOCAL_CTK: '0', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest' }
+    - { ARCH: 'amd64', PY_VER: '3.12',  CUDA_VER: '13.1.1', LOCAL_CTK: '0', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest' }
     - { ARCH: 'amd64', PY_VER: '3.13',  CUDA_VER: '12.9.1', LOCAL_CTK: '0', GPU: 'v100',       GPU_COUNT: '1', DRIVER: 'latest' }
-    - { ARCH: 'amd64', PY_VER: '3.13',  CUDA_VER: '13.0.2', LOCAL_CTK: '1', GPU: 'H100',       GPU_COUNT: '1', DRIVER: 'latest' }
     - { ARCH: 'amd64', PY_VER: '3.13',  CUDA_VER: '13.0.2', LOCAL_CTK: '1', GPU: 'rtxpro6000', GPU_COUNT: '1', DRIVER: 'latest' }
-    - { ARCH: 'amd64', PY_VER: '3.13',  CUDA_VER: '13.1.0', LOCAL_CTK: '1', GPU: 'H100',       GPU_COUNT: '1', DRIVER: 'latest' }
-    - { ARCH: 'amd64', PY_VER: '3.13',  CUDA_VER: '13.1.0', LOCAL_CTK: '1', GPU: 'rtxpro6000', GPU_COUNT: '1', DRIVER: 'latest' }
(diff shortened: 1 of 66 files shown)


leofang:
> It would be preferable to do this at a higher level -- maybe it's possible to make a symlink from `cuda_bindings/tests/` to `cuda_bindings/examples` -- so that they will be included even for local development.

Let's follow what I set up in cuda-core. All cuda-core examples are run as part of the regular tests, either locally or in the CI.

leofang:
@rwgk for reproducing the bug, try replacing `3.13` here
https://github.com/NVIDIA/cuda-python/blob/78ade5c1fd437a1db960ed3433ae0efa2283e100/ci/test-matrix.yml#L33
with `3.14t` so that we use free-threading Python.

rwgk:
> Let's follow what I set up in cuda-core. All cuda-core examples are run as part of the regular tests, either locally or in the CI.

CC @rparolin since this is about priorities: This is turning into a bigger project than expected. What should I do?

I agree the cuda-core approach is much better than what we have now, but this PR provides immediate CI coverage without blocking that direction.

* I started this work expecting it'd be quick and help inform dealing with QA failures (nvbug 5808967).
* Now a very different direction is requested, even though this PR is very small and doesn't create difficulties for the different direction.
* It's an unknown to me how long it'll take to make the structural changes, especially at the moment: I have not gotten a complete CI run even for this easy PR, opened two days ago; currently rtxpro6000 jobs sit in the queue for hours.

My preference: Merge this PR and create a new issue for future proper prioritization: Rework organization of cuda_bindings/examples

rwgk:
> for reproducing the bug, try replacing `3.13` here
> 
> https://github.com/NVIDIA/cuda-python/blob/78ade5c1fd437a1db960ed3433ae0efa2283e100/ci/test-matrix.yml#L33
> 
> 
> with `3.14t` so that we use free-threading Python.

Done: commit dbfa3db586905272b6ed0bf93141b1818866568d

leofang:
Here is our reproducer 🙂 
https://github.com/NVIDIA/cuda-python/actions/runs/21267230884/job/61209868639?pr=1517#step:27:266

rwgk:
> Here is our reproducer 🙂 https://github.com/NVIDIA/cuda-python/actions/runs/21267230884/job/61209868639?pr=1517#step:27:266

Awesome! I'll work on taking care of that now, to get the nvbug out of limbo.

rwgk:
Wow, so many unrelated flakes! Five, I looked at all of them. The [one we're most interested in](https://github.com/NVIDIA/cuda-python/actions/runs/21269516981/job/61216929863?pr=1517) just says "This job failed" with a big X, it didn't even start up.

Currently there is no job running, only 34 queued. I'll cancel and rerun. Nothing lost, but hopefully we'll get more lucky with the infrastructure.

rwgk:
commit 7e03dd5a6f6daf8405b4d238fb32d865013d17a0 fixes nvbug 5808967 / #1525:

https://github.com/NVIDIA/cuda-python/actions/runs/21269516981/job/61302919967?pr=1517#step:27:129

```
examples/extra/isoFDModelling_test.py::main Checking for multiple GPUs...
CUDA-capable device count: 2

Checking GPU(s) for support of peer to peer memory access...
> Peer access from b'NVIDIA H100 NVL' (GPU0) -> b'NVIDIA H100 NVL' (GPU1) : Yes

> Peer access from b'NVIDIA H100 NVL' (GPU1) -> b'NVIDIA H100 NVL' (GPU0) : Yes

dt=  1.6666666666666667  delta=  25.0  nt=  600  freq max=  35.0
init object for device  0
init object for device  1
creating source on device  0
running create velocity on device  0
creating source on device  1
running create velocity on device  1
this code generates  90.05918964491383  GPoints/sec / device 
 nz=  384  nx=  768
Done
PASSED
examples/extra/jit_program_test.py::main 
PASSED
```

rwgk:
@seberg is there a chance that you could review this one file? Did I apply your idea (#1525) correctly?

https://github.com/NVIDIA/cuda-python/pull/1517/changes#diff-61778604948d1f94800d64d8c1092fe04ac2fb8ba7129d7686b4ba97802bc41d

cuda_bindings/examples/common/common.py
```diff
@@ -44,16 +26,16 @@ def pytest_skipif_compute_capability_too_low(devID, required_cc_major_minor):
 
 class KernelHelper:
     def __init__(self, code, devID):
-        prog = checkCudaErrors(nvrtc.nvrtcCreateProgram(str.encode(code), b"sourceCode.cu", 0, None, None))
+        include_dirs = []
+        for libname in ("cudart", "cccl"):
+            hdr_dir = pathfinder.find_nvidia_header_directory(libname)
+            if hdr_dir is None:
+                import pytest
```
leofang, line 33: I see why you said this is a "big project" now. This seems a bit awkward. Though we could use Copilot or other agents for this.

All examples should be runnable standalone. If an end user runs the example, and the headers are not found, then they'd hit this pytest import which may or may not fail right away, and even if it succeeds the next error message (coming from `skip`) would be confusing (they'd see "Skipped").
```python
>>> import pytest
>>> pytest.skip("ok")
  File "<python-input-1>", line 1, in <module>
    pytest.skip("ok")
    ~~~~~~~~~~~^^^^^^
  File "/local/home/leof/miniforge3/envs/py313_cu131/lib/python3.13/site-packages/_pytest/outcomes.py", line 139, in __call__
    raise Skipped(msg=reason, allow_module_level=allow_module_level)
Skipped: ok
```
We should avoid importing pytest in all examples in the next PR. @rwgk could you create an issue to track this?

rwgk: Done: #1530

cuda_bindings/examples/extra/isoFDModelling_test.py
kkraus14, line 1: I'm unclear on why this would fix the issue we were seeing in #1525. In each case, `args` is a list referencing other Python objects which do not go out of scope and are not overwritten or anything of the sort. I.E. here's a minimal example of what we're doing:
```
a = np.array([1,2,3])
args = [a]
args = [args[0].ctypes.data]
```

While it's not a good pattern and this PR is a net improvement, ultimately `a` is still alive and the above code wouldn't end up with a dangling pointer from the original `args` list getting garbage collected.

seberg: It's fine in many, but in some of them `args` is a list containing temporary objects: `args = [wavein + offset_sourceInject, src, np_it]`, `wavein + offset_sourceInject` is owned only by `args` there.

kkraus14: Ah, I completely missed those, good catch and that would indeed cause the issues we were seeing. Resolving this.

PR #1518, opened 2026-01-21
Adopt PEP 563 annotations and modern union syntax

Andy-Jost:
## Summary
- Use `from __future__ import annotations` consistently across cuda.core
- Replace `Union[X, Y]` with `X | Y` and `Optional[X]` with `X | None` in type annotations
- Remove redundant `TYPE_CHECKING` blocks where the imports are no longer needed
- Retain `TYPE_CHECKING` blocks only where needed for forward references (circular imports)

## Changes
- 15 files updated with modern type annotation syntax
- Net reduction of ~20 lines

## Test Coverage
- All existing tests pass
- No new tests needed (type annotation changes only)

### DIFF
--- a/cuda_core/cuda/core/_device.pyx
+++ b/cuda_core/cuda/core/_device.pyx
@@ -2,6 +2,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+from __future__ import annotations
+
 cimport cpython
 from libc.stdint cimport uintptr_t
 
@@ -9,7 +11,6 @@ from cuda.bindings cimport cydriver
 from cuda.core._utils.cuda_utils cimport HANDLE_RETURN
 
 import threading
-from typing import TYPE_CHECKING
 
 from cuda.core._context cimport Context
 from cuda.core._context import ContextOptions
@@ -33,9 +34,6 @@ from cuda.core._utils.cuda_utils import (
     runtime,
 )
 from cuda.core._stream cimport default_stream
-
-if TYPE_CHECKING:
-    from cuda.core._memory import Buffer, MemoryResource
 
 # TODO: I prefer to type these as "cdef object" and avoid accessing them from within Python,
 # but it seems it is very convenient to expose them for testing purposes...
--- a/cuda_core/cuda/core/_event.pyx
+++ b/cuda_core/cuda/core/_event.pyx
@@ -26,14 +26,11 @@ from cuda.core._utils.cuda_utils cimport (
 import cython
 from dataclasses import dataclass
 import multiprocessing
-from typing import TYPE_CHECKING, Optional
 
 from cuda.core._utils.cuda_utils import (
     CUDAError,
     check_multiprocessing_start_method,
 )
-if TYPE_CHECKING:
-    import cuda.bindings
 
 
 @dataclass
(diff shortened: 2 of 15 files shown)


rwgk:
I believe these are still needed:

[add_back_needed_TYPE_CHECKING_imports.txt](https://github.com/user-attachments/files/24780503/add_back_needed_TYPE_CHECKING_imports.txt)

That diff was generated by Cursor. I did not test.

Some of the added `import`s exist as `cimport`s. Cursor told me (I asked about `Context` specifically, to pick out one):

* `cimport Context` only tells Cython about the extension type; it’s not visible to Python type checkers or doc tools.
* With from `__future__ import annotations`, runtime doesn’t need the import, so removing it is fine for correctness/builds.
* If you want mypy/pyright/Sphinx to resolve `Context` in annotations, keep the `if TYPE_CHECKING: import`.

Does that reasoning hold water?

Andy-Jost:
@rwgk Thanks — I ran into several of these as well. This change introduces new dangling type references: annotations that don’t affect building, linting, or runtime correctness, but that cannot be resolved.

For example, this would fail when calling `get_type_hints`:

```
import typing
typing.get_type_hints(buffer.close)  # -> Exception: cannot resolve Stream
```

I started addressing these issues but quickly realized it led down a rabbit hole. There appear to be many similar problems beyond just this change.

If we need to support dynamic retrieval of type hints—for example, for documentation purposes—then we’ll likely need a more coordinated effort. In that case, we should probably add tests that apply `get_type_hints` across the codebase.

For this particular change, I chose to prioritize removing `TYPE_CHECKING` blocks even if it leaves dangling type references. Fixing those seems outside the scope of this update.

rwgk:
> For this particular change, I chose to prioritize removing `TYPE_CHECKING` blocks even if it leaves dangling type references. Fixing those seems outside the scope of this update.

Sounds good to me, but I'd definitely check in the sphinx-generated docs before/after this PR, to be sure the removals don't introduce a regression there. Did you do this already?

Andy-Jost:
Good point. Here are two links to the Buffer documentation: [before](https://nvidia.github.io/cuda-python/cuda-core/latest/generated/cuda.core.Buffer.html#cuda.core.Buffer) [after](https://nvidia.github.io/cuda-python/pr-preview/pr-1518/cuda-core/latest/generated/cuda.core.Buffer.html#cuda.core.Buffer)

This change made `DevicePointerT` a dangling type annotation in the `buffer` module. Looking at `Buffer.from_handle` as an example, there is no link in either case, so this change did not make it worse.

Taking one from your list, here is `Context` in the `device` module: [before](https://nvidia.github.io/cuda-python/cuda-core/latest/generated/cuda.core.Device.html) [after](https://nvidia.github.io/cuda-python/pr-preview/pr-1518/cuda-core/latest/generated/cuda.core.Device.html). Looking at `Device.set_current` the type annotation for `Context` is the same in both cases (no link).

The links are overall in a sorry state. For example:
- In `Buffer.copy_to`, no links for `Steam` and `GraphBuilder`
- In `Buffer.fill`, links for `Steam` and `GraphBuilder`

We have a lot to improve in the documentation, but I don't see any place this change make things worse.

rwgk (approved):
> We have a lot to improve in the documentation, but I don't see any place this change make things worse.

Confirmed for the cases you picked out. I see the `Buffer.copy_to` and `Buffer.fill` cases are identically weird before/after.

I'd lean towards not taking out the code that's there already, but without a pre-commit check in place, this does indeed not have a lot of value.

PR #1520, opened 2026-01-22
Cythonize _module.py

Andy-Jost:
## Summary

Converts `_module.py` to Cython (`_module.pyx`) for improved performance, adding RAII-based resource handle management for `CUkernel` and `CUlibrary` driver objects.

- Converts `Kernel`, `ObjectCode`, `KernelOccupancy`, and `KernelAttributes` to `cdef class`
- Adds `LibraryHandle` and `KernelHandle` to the resource_handles C++ infrastructure
- Replaces Python-level driver API calls with `cydriver` calls wrapped in `nogil` blocks
- Adds `.pxd` file for cross-module `cimport` support

## Changes

- `_module.py` → `_module.pyx` with `cdef class` definitions
- New `_module.pxd` with typed attribute and method declarations
- Extended `_cpp/resource_handles.{hpp,cpp}` with library/kernel handle types
- Updated `_resource_handles.{pxd,pyx}` with new handle functions
- Updated `_launcher.pyx` to directly access kernel handles via `cimport`
- Minor updates to `_linker.py`, `_program.py` to use new factory methods
- Test updates for API changes

## Test plan

- [x] All existing `test_module.py` tests pass
- [x] All existing `test_program.py` tests pass
- [x] CI passes

### DIFF
--- a/cuda_core/cuda/core/_context.pyx
+++ b/cuda_core/cuda/core/_context.pyx
@@ -46,7 +46,7 @@ cdef class Context:
         return as_intptr(self._h_context) == as_intptr(_other._h_context)
 
     def __hash__(self) -> int:
-        return hash((type(self), as_intptr(self._h_context)))
+        return hash(as_intptr(self._h_context))
 
 
 @dataclass
--- a/cuda_core/cuda/core/_event.pyx
+++ b/cuda_core/cuda/core/_event.pyx
@@ -169,7 +169,7 @@ cdef class Event:
             raise RuntimeError(explanation)
 
     def __hash__(self) -> int:
-        return hash((type(self), as_intptr(self._h_event)))
+        return hash(as_intptr(self._h_event))
 
     def __eq__(self, other) -> bool:
         # Note: using isinstance because `Event` can be subclassed.
--- a/cuda_core/cuda/core/_launch_config.pyx
+++ b/cuda_core/cuda/core/_launch_config.pyx
@@ -23,6 +23,9 @@ cdef bint _inited = False
 cdef bint _use_ex = False
 cdef object _lock = threading.Lock()
 
+# Attribute names for identity comparison and representation
+_LAUNCH_CONFIG_ATTRS = ('grid', 'cluster', 'block', 'shmem_size', 'cooperative_launch')
+
 
 cdef int _lazy_init() except?-1:
     global _inited, _use_ex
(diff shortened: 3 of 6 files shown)


cuda_core/cuda/core/_memory/_memory_pool.pxd
```diff
@@ -19,6 +19,16 @@ cdef class _MemPool(MemoryResource):
...
+cdef class _MemPoolAttributes:
+    cdef:
+        MemoryPoolHandle _h_pool
+
+    @staticmethod
+    cdef _MemPoolAttributes _init(MemoryPoolHandle h_pool)
+
+    cdef int _getattribute(self, cydriver.CUmemPool_attribute attr_enum, void* value) except? -1
+
+
```
Andy-Jost, line 31: The changes to `_MemPoolAttributes` follow a pattern set in the change to `KernelAttributes`. We no longer need to rely on weak references to share ownership with an "associated" class, as handle classes can be used instead.

cuda_core/cuda/core/_cpp/resource_handles.hpp
```diff
@@ -243,6 +285,14 @@ inline CUdeviceptr as_cu(const DevicePtrHandle& h) noexcept {
     return h ? *h : 0;
 }
 
+inline CUlibrary as_cu(const LibraryHandle& h) noexcept {
```
rparolin, line 288: Nit: Taking a LibraryHandle by const-ref feels wrong to me. My expectation is that a LibraryHandle should be map to cheap to copy type so pass-by-const-ref s
wouldn't be necessary or even be a regression as it would copy a larger pointer value.

Perhaps we could have a "handle type" be an index into a shared_ptr registry.  That would  enable you to do the cast operation to an object (eg. to_cu) from the handle type.

Andy-Jost: I’m not sure there’s much I can really do here. `LibraryHandle` **is** a `std::shared_ptr`. We went through a lot of design iteration on the resource handles and ended up with this implementation. It seems like the real concern is around what the term “Handle” (as opposed to something like “Holder”) implies to different people, but I don’t think we should make substantive implementation changes based on that alone.

The `const &` here avoids two unnecessary atomic operations, which is strictly better with no downsides. There’s no change from the `cuda.core` developer’s perspective (`as_cu` is used exactly the same way). This function is an internal implementation detail, invisible to end users, and something we can modify at any time — it doesn’t paint us into a corner.

Adding another level of indirection through a table of `std::shared_ptr`s seems like unnecessary complexity. How would we even modify that table? We’d just end up needing another level of locks. One of the nice things about `std::shared_ptr` is that it already handles everything cleanly; i.e., it works well right out of the box.

Andy-Jost:
> Looks good to me.
> 
> This change made me realize that the resource_handles design creates a dependency aggregation hot spot (all the dependencies of Stream, Context, etc. are forced together). But maybe that's OK in this case, assuming the number of cuStream etc. types is small-ish and will never expand by a lot? And they are probably highly dependent on each other already? — This is just to log the observation.

This is a really good observation, and it actually reflects an intentional design choice, even though we didn’t explicitly discuss it at the time. The cost you’re pointing out is real: these files do become dependency hot spots. The upside is that it becomes much easier to introduce changes that apply uniformly to all handles. For example, we could enable logging whenever a handle is created or destroyed, or track handle operations to generate resource-leak reports. Those kinds of features operate over handles in a generic way and can share machinery (classes, macros, etc.), so having all the handles in one place helps. Like any design decision, it comes with trade-offs, but in this case the benefits seemed to outweigh the costs.

PR #1524, opened 2026-01-22
cuda.bindings.nvml: Big wave of handwritten API

mdboom:
This contains changes to ~60 APIs in NVML that needed to have their outside APIs changed to be useful and/or memory safe from Python.  This is an important part of making `cuda.bindings._nvml` public, after which point it will be much harder to change the API.

@leofang mentioned elsewhere that in many cases, adding C shims to convert between what cybind handles and what NVML expects may be a more sustainable solution than hand-writing Cython as here.  I think it makes sense to do that as a follow-up, since there is time pressure to name `cuda.bindings.nvml` public, and this gets us to the point where these APIs are updated to where the APIs wouldn't have to change in future, only the implementation details.

These changes fit into the following general categories:

- Functions where the input or output argument is a struct containing only an API version and a single element, or sometimes even just a struct with a single element.  For speed-of-light and usability reasons, it is much more efficient to make this look like a single input or output argument, rather than forcing the Python side to pack and unpack a single element into a struct/class with a lot of overhead for no gain.  For these also, we can remove the `AUTO_LOWPP_CLASS` generated class, since it isn't needed on the Python side, reducing binary size.  This includes `get_device_performance_modes`, `get_dram_encryption_mode`, `set_dram_encrypytion_mode`, `system_get_conf_compute_key_rotation_threshold_info`, `system_set_conf_compute_key_rotation_threshold_info`, `device_set_power_management_limit_v2`, `device_get_pdi`, `device_set_nvlink_device_low_power_threshold`, `device_get_vgpu_heterogeneous_mode`, `device_set_vgpu_heterogeneous_mode`, `vgpu_instance_get_placement_id`, `vgpu_instance_get_runtime_state_size`, `vgpu_type_get_max_instances_per_gpu_instance`, `gpu_instance_get_vgpu_heterogeneous_mode`, `gpu_instance_set_vgpu_hetereogeneous_mode`, `device_get_unrepairable_memory_flag`, `device_set_rusd_settings_v1`.

- Structs containing a statically-sized array that also has a size member indicating the number of valid elements in that array.  For example:

```c
struct {
    int size
    Struct_t elements[MAX_NUMBER_OF_ELEMENTS]
```

By default, cybind would allow access to the uninitialized memory in `elements[idx >= size]`.  By declaring the size element as special, recent cybind's can do the right thing.  This affects the structs `nvmlBridgeChipHierarchy_t`, `nvmlClkMonStatus_t`, `nvmlConfComputeGpuAttestationReport_t`, `nvmlConfComputeGpuCertificate_t`, `nvmlGridLicenseableFeatures_t`, `nvmlNvlinkSupportedBwModes_t`.

- Functions where input and output are in the same struct, but the inputs are few.  Again, for speed-of-light, it is much better to let the binding handle allocating the struct and filling it with a small number of passed-in arguments.  This includes `device_get_vgpu_type_supported_placements`.

- Functions with structs where the input or output argument is a fixed-length string.  This includes `device_get_conf_compute_gpu_attestation_report`, `vgpu_type_get_license`, `device_read_write_prm_v1`

- Functions which take a reference to the value of an input argument.  For example, the input is an `int`, but the API takes an `int *`.  This is very rare (and unnecessary) for input arguments and is not supported by cybind.  Includes `device_set_temperature_threshold`.

- Functions that return a dynamically-sized array inside a struct (as a size and pointer pair).  Includes `device_get_vgpu_utilization`.

- Functions that take a versioned struct (a struct with a version number member), but the "base class" version is also a specific version.  This needs to be handwritten just because the cast is invalid (though otherwise cybind handles it and generates something that would have worked, had the compiler not complained).  Includes `device_get_gpu_instance_profile_info_v` and `device_get_gpu_instance_profile_info_by_id_v`

- `get_handle_by_uuid_v`: Input to the function is a struct (a variant type to be exact) containing a union to two differently-sized fixed-sized char arrays (string).  This isn't supported by `AUTO_LOWPP_CLASS`.

- A function that takes a struct which is solely a size/pointer pair.  Here, we should take as input a `AUTO_LOWPP_ARRAY` representing the array, not wrapped up in an additional `AUTO_LOWPP_CLASS` where the member array would be hard/impossible to initialize and manage the memory properly.  This includes `device_read_prm_counters_v1`.

- A few functions will be quite hard to support cleanly from Python.  They are not on the top of the list to support right now, so I think it best to just `SKIP` them and come back to them later.  These include `nvmlGpm*` and `nvmlDeviceWorkloadPowerProfile*`.

- Functions were already deprecated in 12.9 and should never have been wrapped.  (As a policy on this work, I have not wrapped anything that was deprecated in 12.9 or earlier).  Includes `device_set_power_management_limit`

### DIFF
--- a/cuda_bindings/cuda/bindings/_nvml.pyx
+++ b/cuda_bindings/cuda/bindings/_nvml.pyx
@@ -10833,6 +10833,8 @@ cdef class ConfComputeGpuCertificate:
     def cert_chain(self, val):
         if self._readonly:
             raise ValueError("This ConfComputeGpuCertificate instance is read-only")
+        if len(val) > 4096:
+            raise ValueError(f"Too many elements for field cert_chain, max is 4096, got {len(val)}")
         self._ptr[0].certChainSize = len(val)
         if len(val) == 0:
             return
@@ -10853,6 +10855,8 @@ cdef class ConfComputeGpuCertificate:
     def attestation_cert_chain(self, val):
         if self._readonly:
             raise ValueError("This ConfComputeGpuCertificate instance is read-only")
+        if len(val) > 5120:
+            raise ValueError(f"Too many elements for field attestation_cert_chain, max is 5120, got {len(val)}")
         self._ptr[0].attestationCertChainSize = len(val)
         if len(val) == 0:
             return
@@ -11015,6 +11019,8 @@ cdef class ConfComputeGpuAttestationReport:
     def attestation_report(self, val):
         if self._readonly:
             raise ValueError("This ConfComputeGpuAttestationReport instance is read-only")
+        if len(val) > 8192:
+            raise ValueError(f"Too many elements for field attestation_report, max is 8192, got {len(val)}")
         self._ptr[0].attestationReportSize = len(val)
         if len(val) == 0:
             return
@@ -11035,6 +11041,8 @@ cdef class ConfComputeGpuAttestationReport:
     def cec_attestation_report(self, val):
         if self._readonly:
             raise ValueError("This ConfComputeGpuAttestationReport instance is read-only")
+        if len(val) > 4096:
+            raise ValueError(f"Too many elements for field cec_attestation_report, max is 4096, got {len(val)}")
         self._ptr[0].cecAttestationReportSize = len(val)
         if len(val) == 0:
             return
@@ -11363,6 +11371,8 @@ cdef class NvlinkSupportedBwModes_v1:
     def bw_modes(self, val):
         if self._readonly:
             raise ValueError("This NvlinkSupportedBwModes_v1 instance is read-only")
+        if len(val) > 23:
+            raise ValueError(f"Too many elements for field bw_modes, max is 23, got {len(val)}")
         self._ptr[0].totalBwModes = len(val)
         if len(val) == 0:
             return


cuda_bindings/tests/nvml/test_device.py
```diff
@@ -0,0 +1,159 @@
...
+def test_get_nv_link_supported_bw_modes(all_devices):
+    for device in all_devices:
+        with unsupported_before(device, None):
+            modes = nvml.device_get_nvlink_supported_bw_modes(device)
+        assert isinstance(modes, nvml.NvLinkSupportedBWModes)
+        # #define NVML_NVLINK_TOTAL_SUPPORTED_BW_MODES 23
+        assert len(modes.supported_bw_modes) <= 23
+        assert not hasattr(modes, "total_bw_modes")
+
+        for mode in modes.bw_modes:
```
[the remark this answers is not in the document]
mdboom: Good catch, but you had it backwards, @Copilot.

cuda_bindings/tests/nvml/test_device.py
```diff
@@ -0,0 +1,159 @@
...
+    for device in all_devices:
+        status = nvml.device_get_unrepairable_memory_flag_v1(device)
+        assert isinstance(status, int)
+
+
+def test_device_vgpu_get_heterogeneous_mode(all_devices):
+    for device in all_devices:
+        with unsupported_before(device, None):
+            mode = nvml.device_get_vgpu_heterogeneous_mode(device)
+        assert isinstance(mode, int)
```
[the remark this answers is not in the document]
mdboom: Nope -- removal from the .pxd does not imply removal from the module.

cuda_bindings/tests/nvml/test_gpu.py
```diff
@@ -35,3 +36,47 @@ def test_gpu_get_platform_info(all_devices):
...
+    assert len(cert.attestation_cert_chain) == 0
+
+
+def test_conf_compute_gpu_certificate(all_devices):
+    for device in all_devices:
+        # Documentation says AMPERE or newer
+        with unsupported_before(device, None):
+            cert = nvml.device_get_conf_compute_gpu_certificate(device)
+
+        assert isinstance(cert, nvml.ComputeGpuCertificate)
```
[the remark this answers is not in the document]
mdboom: You got lucky here, and it's indeed a problem, but not for the reason you state.  Normally content of one function doesn't affect another like that.

cuda_bindings/tests/nvml/test_gpu.py
```diff
@@ -35,3 +36,47 @@ def test_gpu_get_platform_info(all_devices):
...
+    assert report.nonce.dtype == np.uint8
+
+
+def test_gpu_conf_compute_attestation_report(all_devices):
+    for device in all_devices:
+        # Documentation says AMPERE or newer
+        with unsupported_before(device, None):
+            report = nvml.device_get_conf_compute_gpu_attestation_report(device, nonce=b"12345678")
+
+        assert isinstance(report, nvml.ComputeGpuAttestationReport)
```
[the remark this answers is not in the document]
mdboom: Again, your mental model of how a procedural programming language actually works is astonishing.

mdboom:
@Copilot: You provided some interesting (if mostly incorrect) feedback about the tests.  Can you provide feedback about the actual implementation, especially in `_nvml.pyx`, or do you struggle with Cython?

cuda_bindings/cuda/bindings/_nvml.pyx
```diff
@@ -5642,6 +5511,8 @@ cdef class PlatformInfo_v1:
     @property
     def ib_guid(self):
         """~_numpy.uint8: (array of length 16).Infiniband GUID reported by platform (for Blackwell, ibGuid is 8 bytes so indices 8-15 are zero)"""
+        if 16 == 0:
+            return _numpy.array([])
```
kkraus14: What are the points of these? I imagine they would just get compiled out anyway?

cuda_bindings/cuda/bindings/_nvml.pyx
```diff
@@ -5650,13 +5521,17 @@ cdef class PlatformInfo_v1:
     def ib_guid(self, val):
         if self._readonly:
             raise ValueError("This PlatformInfo_v1 instance is read-only")
+        if len(val) != 16:
+            raise ValueError(f"Expected length { 16 } for field ib_guid, got {len(val)}")
```
kkraus14, line 5523: Since `val` isn't strongly typed here, I imagine the code generated for this check isn't the most efficient and goes through Python objects? Any way around this in using something like a memoryview as the type of `val`?

cuda_bindings/cuda/bindings/_nvml.pyx
```diff
@@ -26176,1596 +24143,1980 @@ cpdef unsigned int gpu_instance_get_compute_instance_remaining_capacity(intptr_t
...
-
-    .. seealso:: `nvmlDeviceIsMigDeviceHandle`
+        str: driver branch.
     """
-    cdef unsigned int is_mig_device
+    cdef nvmlSystemDriverBranchInfo_t info
+    # Calculation copied from the macro NVML_STRUCT_VERSION in nvml.h
+    # Needs to be updated if the version of the nvmlSystemDriverBranchInfo_t
+    # struct changes in the future.
+    info.version = sizeof(nvmlSystemDriverBranchInfo_v1_t) | (1 << 24)
```
kkraus14, line 24507: Is there a reason we copied this instead of externing it and using it directly?

cuda_bindings/cuda/bindings/_nvml.pyx
```diff
@@ -26176,1596 +24143,1980 @@ cpdef unsigned int gpu_instance_get_compute_instance_remaining_capacity(intptr_t
...
     """
-    cdef ComputeInstanceInfo info_py = ComputeInstanceInfo()
-    cdef nvmlComputeInstanceInfo_t *info = <nvmlComputeInstanceInfo_t *><intptr_t>(info_py._get_ptr())
+    cdef unsigned int[1] count = [0]
     with nogil:
-        __status__ = nvmlComputeInstanceGetInfo_v2(<ComputeInstance>compute_instance, info)
+        __status__ = nvmlSystemGetTopologyGpuSet(cpuNumber, <unsigned int*>count, NULL)
+    check_status_size(__status__)
+    if count[0] == 0:
+        return view.array(shape=(1,), itemsize=sizeof(intptr_t), format="P", mode="c")[:0]
```
kkraus14, line 24489: It's a bit funky that we're generally using numpy everywhere else and then returning memoryview objects here.

cuda_bindings/cuda/bindings/_nvml.pyx
```diff
@@ -25585,6 +23631,7 @@ cpdef gpu_instance_set_vgpu_scheduler_state(intptr_t gpu_instance, intptr_t p_sc
 
     .. seealso:: `nvmlGpuInstanceSetVgpuSchedulerState`
     """
+    (<nvmlVgpuSchedulerState_t*>p_scheduler).version = sizeof(nvmlVgpuSchedulerState_v1_t) | (1 << 24)
```
kkraus14, line 23623: I know the 1 << 24 was already an established pattern, but it would be good to make it a global with a descriptive name or something to indicate what's going on here.

cuda_bindings/cuda/bindings/_nvml.pyx
```diff
@@ -26176,1596 +24143,1980 @@ cpdef unsigned int gpu_instance_get_compute_instance_remaining_capacity(intptr_t
...
-    """Get device capabilities.
+cpdef device_set_hostname_v1(intptr_t device, str hostname):
+    """Set the hostname for the device.
 
     Args:
-        device (intptr_t): The identifier of the target device.
+        device (Device): The identifier of the target device.
+        hostname (str): The new hostname to set.
+    """
+    cdef bytes = cpython.PyUnicode_AsASCIIString(hostname)
```
kkraus14, line 24743: What happens if we encounter non-ascii characters in the hostname here? Should we use UTF-8 since that's the standard Python string encoding?

cuda_bindings/cuda/bindings/_nvml.pyx
```diff
@@ -26176,1596 +24143,1980 @@ cpdef unsigned int gpu_instance_get_compute_instance_remaining_capacity(intptr_t
...
+
+    .. seealso:: `nvmlDeviceGetGpuFabricInfoV`
     """
-    cdef bytes = cpython.PyUnicode_AsASCIIString(hostname)
-    if len(bytes) > 64:
-        raise ValueError("hostname must 64 characters or less")
+    cdef GpuFabricInfo_v3 gpu_fabric_info_v3_py
+    cdef GpuFabricInfo_v2 gpu_fabric_info_v2_py
+    cdef nvmlGpuFabricInfoV_t *gpu_fabric_info
+    if CUDA_VERSION >= 13000:
```
kkraus14, line 25304: Is this to make backporting to 12.9 easier? Otherwise we haven't guarded anything else that requires CUDA 13.0+ in `cuda.bindings` generally.

cuda_bindings/tests/nvml/test_device.py
```diff
@@ -0,0 +1,170 @@
...
+
+import cuda.bindings.runtime as cudart
+import pytest
+from cuda.bindings import _nvml as nvml
+
+from .conftest import unsupported_before
+
+
+def isSuccess(err):
+    return err == cudart.cudaError_t.cudaSuccess
```
kkraus14: should be using `nvmlReturn_t` instead of `cudaError_t` here

cuda_bindings/tests/nvml/test_device.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+def assertSuccess(err):
+    assert isSuccess(err)
+
+
+def driverVersionLessThan(target):
+    err, version = cudart.cudaDriverGetVersion()
```
kkraus14: Should use the appropriate nvml API for this instead of the cudart API: https://docs.nvidia.com/deploy/nvml-api/group__nvmlSystemQueries.html#group__nvmlSystemQueries

cuda_bindings/cuda/bindings/_nvml.pyx
mdboom: Yeah, they get compiled out.  I can fix up the generator so they don't even get generated -- it just comes from the fact that fixed-sized and dynamic-sized arrays share some code over there.

cuda_bindings/cuda/bindings/_nvml.pyx
mdboom: Perhaps.  It would just move the check into Cython's generated code, not remove it, though.

cuda_bindings/cuda/bindings/_nvml.pyx
mdboom: It's a macro in the header, and NVML is built without requiring the header.  Even if we did expose it, passing a struct to a macro that sizeof is then run on wouldn't work with Cython.

As follow up to this work, many of these hand-written Cython functions may get rewritten as C shims.  As part of that, we can copy the macro to generate these version numbers over to the shim which would be a little more maintainable than this.

cuda_bindings/cuda/bindings/_nvml.pyx
mdboom: Yes, but there is a plan to remove the numpy dependency for generated code in the future -- we just need to do it gradually.  In the meantime, I think this is preferable for new APIs.

cuda_bindings/cuda/bindings/_nvml.pyx
mdboom: Hostnames are by definition ascii, and if you need non-ascii characters, you need to punycode it (which is an ascii-only encoding).  This puts the onus to punycode it onto the user.

This will raise an exception that bubbles up to the user, which I think is the right thing, unless we wanted to implicitly punycode it for them.

cuda_bindings/cuda/bindings/_nvml.pyx
mdboom: Yes.  NVML has APIs where the /function/ is the same between CUDA versions, but the /struct/ passed to that function changes between 12.9 and 13.0.  AFAICT, we have facilities to define a function that only exists in 13.0 and have it be a runtime error when used with 12.9.  But the generator doesn't have a facility for this case where the contents of the struct change based on the CUDA version.

cuda_bindings/cuda/bindings/_nvml.pyx
mdboom: For backward compatibility, this needs to be pretty generic and ultimately go through `numpy.asarray` (which handles more things than Cython's cast to an array view).  While it would be nice to be more efficient, I think this is one of those cases where we are currently accepting a lot of things and would need to deprecate that carefully.

cuda_bindings/cuda/bindings/_nvml.pyx
mdboom: And to add a little more detail -- the places we use Numpy arrays are all in `AUTO_LOWPP_ARRAY` generated classes.  Functions that don't have/need those have always been simple memoryviews (which can be easily and efficiently turned into Numpy arrays by the user).

cuda_bindings/cuda/bindings/_nvml.pyx
mdboom: https://github.com/NVIDIA/cuda-python/issues/1535

cuda_bindings/cuda/bindings/_nvml.pyx
kkraus14: https://github.com/NVIDIA/cuda-python/issues/1535

cuda_bindings/cuda/bindings/_nvml.pyx
kkraus14: My hope is we can get to efficient cython generated cpp code here but I suspect this hits generic Python code for handling arbitrary objects.

PR #1526, opened 2026-01-22
Fix build warnings and optimize handle passing

Andy-Jost:
## Summary

- Fix 44 Cython warnings about `nogil` placement (should appear after `except+`; no functional difference, just clears a warning saying this will be an error in future version of Cython)
- Fix unreachable code warning in `ManagedMemoryResource`
- Optimize C++ handle-passing functions to use `const&` to avoid atomic ref count operations

## Changes

1. **Cython `nogil` placement** (`_resource_handles.pxd`, `_resource_handles.pyx`):
   - Changed from `func(...) nogil except+` to `func(...) except+ nogil`
   - This will be required in a future Cython version

2. **Dead code fix** (`_managed_memory_resource.pyx`):
   - Moved `super().__init__()` inside the `IF CUDA_CORE_BUILD_MAJOR >= 13` block
   - The `ELSE` branch raises an exception, making code after it unreachable

3. **C++ const& optimization** (`.hpp`, `.cpp`, `.pxd`, `.pyx`):
   - Functions that accept `shared_ptr` handles now take them by `const&`
   - Avoids atomic increment/decrement on function entry/exit
   - Affected functions: `create_stream_handle`, `create_event_handle`, `deviceptr_alloc_from_pool`, `deviceptr_alloc_async`, `deviceptr_import_ipc`, `set_deallocation_stream`

## Test plan

- [x] Build compiles cleanly with no warnings
- [x] CI tests pass

### DIFF
--- a/cuda_core/cuda/core/_resource_handles.pxd
+++ b/cuda_core/cuda/core/_resource_handles.pxd
@@ -56,41 +56,41 @@ cdef cydriver.CUresult peek_last_error() noexcept nogil
 cdef void clear_last_error() noexcept nogil
 
 # Context handles
-cdef ContextHandle create_context_handle_ref(cydriver.CUcontext ctx) nogil except+
-cdef ContextHandle get_primary_context(int device_id) nogil except+
-cdef ContextHandle get_current_context() nogil except+
+cdef ContextHandle create_context_handle_ref(cydriver.CUcontext ctx) except+ nogil
+cdef ContextHandle get_primary_context(int device_id) except+ nogil
+cdef ContextHandle get_current_context() except+ nogil
 
 # Stream handles
 cdef StreamHandle create_stream_handle(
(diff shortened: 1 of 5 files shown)


cuda_core/cuda/core/_resource_handles.pxd
```diff
@@ -56,41 +56,41 @@ cdef cydriver.CUresult peek_last_error() noexcept nogil
 cdef void clear_last_error() noexcept nogil
 
 # Context handles
-cdef ContextHandle create_context_handle_ref(cydriver.CUcontext ctx) nogil except+
-cdef ContextHandle get_primary_context(int device_id) nogil except+
-cdef ContextHandle get_current_context() nogil except+
+cdef ContextHandle create_context_handle_ref(cydriver.CUcontext ctx) except+ nogil
```
rparolin, line 59: Can you include in the PR description why `nogil except+` is different compared to `except+ nogil`?

PR #1528, opened 2026-01-23
 [cuda.pathfinder] Add support for finding CUDA binary utilities

### LINKED ISSUE #762: [FEA]: Support for finding binary utilities like `nvdisasm`
### Is this a duplicate?

### Area

cuda.pathfinder

(Marked "not sure" for the area because there's no pathfinder choice)

In Numba-CUDA we have https://github.com/NVIDIA/numba-cuda/pull/13 which attempts to improve the logic for finding `nvdisasm`. However, as we're leaning on `pathfinder` for all the path-finding logic now, it'd be nice to also be able to locate `nvdisasm` and other tools with `pathfinder`. Is this something that's planned / in-scope?

Something like

```python
nvdisasm_path = cuda.pathfinder.binary_utility('nvdisasm')
cuobjdump_path = cuda.pathfinder.binary_utility('cuobjdump')
# ... etc ...
```

to get the paths of the best choice for running these utilities.

Continuing to attempt to locate `nvdisasm` in Numba-CUDA. This is OK to do, but I wonder if this is a common need - I think I have seen people asking about doing similar things for other libraries / purposes when they need to produce SASS for user inspection.

### Additional context

_No response_

rparolin:
## Summary

Adds a new API to locate CUDA binary utilities (e.g., nvdisasm, cuobjdump, nvcc) following the same search strategy as the existing library and header finders.

## Motivation

Numba-CUDA and other projects need to locate CUDA binary utilities like nvdisasm for disassembly and debugging. This change enables using cuda.pathfinder for all CUDA resource discovery, providing a consistent interface across libraries, headers, and binaries.

Closes #762

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_binaries/supported_nvidia_binaries.py
+++ b/cuda_pathfinder/cuda/pathfinder/_binaries/supported_nvidia_binaries.py
@@ -4,32 +4,30 @@ import os
 
 # Site-packages bin directories where binaries might be found
 # Based on NVIDIA wheel layouts (same for Linux and Windows)
-CUDA_NVCC_BIN = os.path.join("nvidia", "cuda_nvcc", "bin")
-NSIGHT_SYSTEMS_BIN = os.path.join("nvidia", "nsight_systems", "bin")
-NSIGHT_COMPUTE_BIN = os.path.join("nvidia", "nsight_compute", "bin")
+_CUDA_NVCC_BIN = os.path.join("nvidia", "cuda_nvcc", "bin")
+_NSIGHT_SYSTEMS_BIN = os.path.join("nvidia", "nsight_systems", "bin")
+_NSIGHT_COMPUTE_BIN = os.path.join("nvidia", "nsight_compute", "bin")
 
 # Common CUDA binary utilities available on both Linux and Windows
 SITE_PACKAGES_BINDIRS = {
     # Core compilation tools
-    "nvcc": (CUDA_NVCC_BIN,),
-    "nvdisasm": (CUDA_NVCC_BIN,),
-    "cuobjdump": (CUDA_NVCC_BIN,),
-    "nvprune": (CUDA_NVCC_BIN,),
-    "fatbinary": (CUDA_NVCC_BIN,),
-    "bin2c": (CUDA_NVCC_BIN,),
-    "nvlink": (CUDA_NVCC_BIN,),
+    "nvcc": (_CUDA_NVCC_BIN,),
+    "nvdisasm": (_CUDA_NVCC_BIN,),
+    "cuobjdump": (_CUDA_NVCC_BIN,),
+    "nvprune": (_CUDA_NVCC_BIN,),
+    "fatbinary": (_CUDA_NVCC_BIN,),
+    "bin2c": (_CUDA_NVCC_BIN,),
+    "nvlink": (_CUDA_NVCC_BIN,),
     # Runtime/debugging tools
-    "cuda-gdb": (CUDA_NVCC_BIN,),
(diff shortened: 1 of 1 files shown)


cpcloud:
Why can't we use existing standard library functionality ([`shutil.which`](https://docs.python.org/3/library/shutil.html#shutil.which), it works on Windows too) for this?

```python
import shutil
import os

shutil.which(name_of_binary, path=os.pathsep.join(list_of_paths_in_order_of_preference))
```

cuda_pathfinder/cuda/pathfinder/_binaries/find_nvidia_binary_utility.py
```diff
@@ -0,0 +1,106 @@
...
+    candidate_dirs = supported_nvidia_binaries.SITE_PACKAGES_BINDIRS.get(utility_name, ())
+    dirs = []
+
+    for sub_dir in candidate_dirs:
+        dirs.extend(find_sub_dirs_all_sitepackages(sub_dir.split(os.sep)))
+
+    # 2. Search in Conda environment
+    if (conda_prefix := os.environ.get("CONDA_PREFIX")) is not None:
+        if IS_WINDOWS:
+            dirs.append(os.path.join(conda_prefix, "Library", "bin"))
```
rwgk, line 95: Is there an `else:` (and indentation for the append `"bin"`) missing here? (Maybe a copy-paste mishap, because below it's correct without the `else`?)

Or is the docstring above inconsistent with the implementation here?

cuda_pathfinder/cuda/pathfinder/_binaries/supported_nvidia_binaries.py
```diff
@@ -0,0 +1,35 @@
...
+    "nvprof": (CUDA_NVCC_BIN,),
+    "nsys": (NSIGHT_SYSTEMS_BIN,),
+    "nsight-sys": (NSIGHT_SYSTEMS_BIN,),
+    "ncu": (NSIGHT_COMPUTE_BIN,),
+    "nsight-compute": (NSIGHT_COMPUTE_BIN,),
+}
+
+SUPPORTED_BINARIES_ALL = SUPPORTED_BINARIES = tuple(SITE_PACKAGES_BINDIRS.keys())
+
+del CUDA_NVCC_BIN, NSIGHT_SYSTEMS_BIN, NSIGHT_COMPUTE_BIN, os
```
rwgk: This is very likely to get out of sync over time, b/o oversights. If you want to "hide" these symbols, prefixing with an underscore will make that more clear and self-maintaining. Another trick would be to introduce a helper function, but I think underscores will be just fine.

I seems very unusual to remove stdlib symbols (`os`), especially because this is not even a public module.

cuda_pathfinder/cuda/pathfinder/_binaries/find_nvidia_binary_utility.py
cpcloud: Docstring might be out of date, this was the original implementation that @rparolin had.

Happy to adjust. Should windows search `bin` as well as `Library/bin` or just the latter?

cuda_pathfinder/cuda/pathfinder/_binaries/supported_nvidia_binaries.py
cpcloud: Not sure I follow the bit about removing stdlib symbols, it's no different from removing any other symbol you don't want to expose.

In any event I will adjust this to just use underscores.

cuda_pathfinder/cuda/pathfinder/_binaries/find_nvidia_binary_utility.py
leofang: just the latter

PR #1532, opened 2026-01-24
Fix #1531 (remove stray skip_reasons code)

### LINKED ISSUE #1531: The main branch fails with `skip_reason` not defined
> I don't know how this got passed the CI, but it failed on main:
>```
>                except nvml.NoPermissionError:
> >                  skip_reasons.add(f"nvmlDeviceSetComputeMode requires root for device {device}")
>                    ^^^^^^^^^^^^
>E                   NameError: name 'skip_reasons' is not defined
>```
https://github.com/NVIDIA/cuda-python/actions/runs/21307037717/job/61337052365#step:26:510

_Originally posted by @leofang in https://github.com/NVIDIA/cuda-python/pull/1515#discussion_r2723538213_

rwgk:
Closes #1531

Deduplicate root-only devices while preserving order.

### DIFF
--- a/cuda_core/cuda/core/_cpp/resource_handles.cpp
+++ b/cuda_core/cuda/core/_cpp/resource_handles.cpp
@@ -234,7 +234,7 @@ struct StreamBox {
 };
 }  // namespace
 
-StreamHandle create_stream_handle(ContextHandle h_ctx, unsigned int flags, int priority) {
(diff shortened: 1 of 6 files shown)


mdboom:
Simplified this -- there is no need to try on multiple devices, since we don't test on systems with devices of different architecture generations, and if we ever come across a system like that, we will get a warning: https://github.com/NVIDIA/cuda-python/blob/main/cuda_bindings/tests/nvml/test_init.py#L20

PR #1534, opened 2026-01-25
Add __eq__ and __hash__ to Buffer, LaunchConfig, Kernel, ObjectCode

Andy-Jost:
## Summary

Complete hashability support for core API classes by adding `__eq__` and `__hash__` to Buffer, LaunchConfig, Kernel, and ObjectCode.

## Changes

- Add `__eq__` and `__hash__` to Buffer, LaunchConfig, Kernel, ObjectCode
- Remove `type(self)` from `__hash__` in Event, Context (and the new classes) - CUDA handles are already unique, and this maintains the hash/equality contract for subclasses
- Add `_LAUNCH_CONFIG_ATTRS` tuple for forward-compatible identity in LaunchConfig
- Update `__repr__` in LaunchConfig to use the same attribute list
- Add parameterized hash tests for all 8 core API classes

**Identity basis:**
| Class | Identity |
|-------|----------|
| Buffer | `(ptr, size)` |
| LaunchConfig | `(grid, cluster, block, shmem_size, cooperative_launch)` |
| Kernel | `handle` |
| ObjectCode | `handle` (triggers lazy load) |

Stream, Event, Context, Device already had `__eq__`/`__hash__`.

## Test Plan

- [x] New parameterized tests in `test_hashable.py` cover all 8 core classes
- [x] All existing hash/equality tests pass
- [x] Full test suite passes locally (1599 tests)


cuda_core/tests/test_hashable.py
```diff
@@ -12,24 +12,109 @@
...
+
+
+@pytest.fixture
+def sample_event(sample_device):
+    return sample_device.create_event()
+
+
+@pytest.fixture
+def sample_context(sample_device):
+    return sample_device.context
```
Andy-Jost, line 41: Once this change and the change adding `__weakref__` members are both merged, I plan to revisit these fixtures (which overlap with `test_weakref.py`). There is an opportunity to rewrite and simplify `test_hashable.py`, `test_comparable.py` and `test_weakref.py`, possibly combining them.

kkraus14:
> this maintains the hash/equality contract for subclasses

I think the reason to include `type` was to make it so subclasses don't hash / compare equally. That being said, I don't think there was a strong reason for that behavior and a subclass can always override the `__hash__` and `__eq__` methods if they wish to have that behavior.

cuda_core/cuda/core/_launch_config.pyx
```diff
@@ -131,11 +134,21 @@ cdef class LaunchConfig:
         if self.cooperative_launch and not Device().properties.cooperative_launch:
             raise CUDAError("cooperative kernels are not supported on this device")
 
+    def _identity(self):
+        return tuple(getattr(self, attr) for attr in _LAUNCH_CONFIG_ATTRS)
```
kkraus14, line 138: It may be worth caching this

Andy-Jost:
> > this maintains the hash/equality contract for subclasses
> 
> I think the reason to include `type` was to make it so subclasses don't hash / compare equally. That being said, I don't think there was a strong reason for that behavior and a subclass can always override the `__hash__` and `__eq__` methods if they wish to have that behavior.

I thought so as well, but then I found tests explicitly checking that subclasses compare equal to non-subclasses managing the same driver resource. This also came up in a recent issue where it was decided for streams to use `CUstream` (only) as the basis of identity.

cuda_core/cuda/core/_launch_config.pyx
Andy-Jost: Is `LaunchConfig` intended to be immutable? Currently it is not:

```
>>> config = LaunchConfig(grid=(1,2),  block=1)
>>> config
LaunchConfig(grid=(1, 2, 1), cluster=None, block=(1, 1, 1), shmem_size=0, cooperative_launch=False)
>>> config.grid = (2,2,2)
>>> config
LaunchConfig(grid=(2, 2, 2), cluster=None, block=(1, 1, 1), shmem_size=0, cooperative_launch=False)
```

kkraus14: In theory we could invalidate the cache on setter functions, but we can punt this to if it ever becomes a bottleneck.

PR #1538, opened 2026-01-28
Add __repr__ to core API classes and consolidate object protocol tests

Andy-Jost:
## Summary
- Add `__repr__` methods to Stream, Event, Context, Buffer, ObjectCode, and Kernel
- Create `test_object_protocols.py` consolidating tests from removed files
- Remove `test_hashable.py`, `test_comparable.py`, `test_weakref.py`

## Changes
- **`__repr__` implementations**: All core API classes now have informative repr output (e.g., `Stream(handle=0x...)`, `Buffer(ptr=0x..., size=64)`)
- **Consolidated test module** with 152 tests covering:
  - Weak references (weakref.ref, WeakValueDict, WeakKeyDict, WeakSet)
  - Hash consistency, distinctness across types, and small value guards
  - Equality (reflexive, cross-type, same-type distinct objects)
  - Collection usage (dict, set, weak collections)
  - repr format validation with regex patterns

## Test Coverage
- 152 parameterized tests (148 pass, 4 skip on single-GPU machines)
- All tests run locally and pass


rwgk:
Logging a Cursor prompt/response below. What do you think about the "middle ground" approach?

___

# `__repr__` vs `__str__` - Mainstream Wisdom vs LSC Experience

## Question

From my days working at Google on "Large Scale Changes", i.e. often touching thousands of files across a gigantic codebase, I developed the opinion: keep the default `__repr__`, but implement `__str__`. This is so that it's easy to see what type an object is when e.g. debug-printing a list or dict, or especially nested lists and dicts. What's the current mainstream wisdom in this regard?

## Response

That's an interesting perspective. The mainstream Python wisdom actually goes the opposite direction:

**The conventional advice** (from Python docs, Raymond Hettinger, etc.) is:

1. **Always implement `__repr__`** - it should be unambiguous and ideally `eval()`-able to recreate the object
2. **`__str__` is optional** - implement it only if you want a "pretty" user-facing string that differs from `__repr__`

The rationale is that `__repr__` is used in more contexts automatically (debuggers, logging, interactive REPL, and as the fallback for `__str__`), so it's the more important one to get right.

**However**, your concern is valid and points to a real tradeoff:

- Default `__repr__` gives you: `<cuda.core._stream.Stream object at 0x7f...>`
- Custom `__repr__` gives you: `Stream(handle=0x7f...)`

In nested structures, the default makes the *type* immediately obvious because of the full module path. The custom `__repr__` loses that context - if you're looking at a list of mixed objects, you might see `Stream(...)` but not immediately know it's `cuda.core.Stream` vs some other `Stream`.

**A middle ground** some projects use:

```python
def __repr__(self):
    return f"<{self.__class__.__module__}.{self.__class__.__name__} handle={...:#x}>"
```

This preserves the angle-bracket style and module path while adding useful state info.

That said, I think the mainstream has largely settled on "implement `__repr__`, let `__str__` fall back to it" - your Google LSC experience may have given you a perspective that's more informed by large-scale debugging scenarios than typical library design.

cuda_core/cuda/core/_module.pyx
```diff
@@ -864,3 +867,8 @@ cdef class ObjectCode:
         # Trigger lazy load to get the handle
         self._lazy_load_module()
         return hash(as_intptr(self._h_library))
+
+    def __repr__(self) -> str:
+        # Trigger lazy load to get the handle
+        self._lazy_load_module()
```
rwgk, line 873: Hm ... this could trigger "interesting" error scenarios.

Is the lazy loading really needed? How was the handle generated (if it exists)? Didn't that trigger the import already? If the handle does not exist, maybe we should just show `<not imported>` or similar in the `repr`, because that's what the actual state is?

cuda_core/tests/test_object_protocols.py
```diff
@@ -0,0 +1,315 @@
...
+    h = hash(obj)
+    assert abs(h) >= 10, f"hash {h} is suspiciously small"
+
+
+@pytest.mark.parametrize("a_name,b_name", SAME_TYPE_PAIRS)
+def test_hash_distinct_same_type(a_name, b_name, request):
+    """Distinct objects of the same type have different hashes."""
+    obj_a = request.getfixturevalue(a_name)
+    obj_b = request.getfixturevalue(b_name)
+    assert hash(obj_a) != hash(obj_b), f"{a_name} and {b_name} have same hash but different handles"
```
rwgk: In general hash collisions are allowed by the hash contract - only equal objects must have equal hashes, not the converse. Are you deliberately making the requirements stronger for the `cuda.core` types? If yes, could you please add a comment to explain?

cuda_core/tests/test_object_protocols.py
```diff
@@ -0,0 +1,315 @@
...
+    obj = request.getfixturevalue(fixture_name)
+    assert hash(obj) == hash(obj)
+
+
+@pytest.mark.parametrize("fixture_name", API_TYPES)
+def test_hash_not_small(fixture_name, request):
+    """Hash should not be a small number (guards against returning IDs or indices)."""
+    obj = request.getfixturevalue(fixture_name)
+    h = hash(obj)
+    assert abs(h) >= 10, f"hash {h} is suspiciously small"
```
rwgk: Hm, what if someone has 10+ GPUs and tests `Device(9)`, this would pass even if it were returning the device ID. Also, in general legitimate hash values can be small.

I'm thinking this test isn't really needed. I think it's better to limit testing to what we're sure must be true, e.g. the hash for objects that are equal must be equal, too, but I wouldn't assert anything on the hash value itself.

cuda_core/tests/test_object_protocols.py
```diff
@@ -0,0 +1,315 @@
...
+
+
+@pytest.mark.parametrize("a_name,b_name", SAME_TYPE_PAIRS)
+def test_same_type_inequality(a_name, b_name, request):
+    """Two distinct objects of the same type should not compare equal."""
+    obj_a = request.getfixturevalue(a_name)
+    obj_b = request.getfixturevalue(b_name)
+    assert obj_a is not obj_b, f"{a_name} and {b_name} are the same object"
+    assert obj_a != obj_b, f"{a_name} == {b_name} but they have different handles"
+
```
rwgk, line 264: Cursor suggested:

Missing test for equality of *same* objects created differently

The tests check that distinct objects are not equal, but don't verify that two wrapper objects pointing to the *same* underlying handle compare equal. For example:

```python
stream2 = Stream.from_handle(stream.handle)
assert stream == stream2  # This case isn't tested
```

cuda_core/tests/test_object_protocols.py
```diff
@@ -0,0 +1,315 @@
...
+@pytest.fixture
+def sample_event_alt(sample_device):
+    """An alternate Event object."""
+    return sample_device.create_event()
+
+
+@pytest.fixture
+def sample_context_alt(sample_device_alt):
+    """An alternate Context object (requires multi-GPU)."""
+    sample_device_alt.set_current()
```
rwgk: This calls `set_current()` which changes global state. The fixture doesn't restore the previous current device. Is this future-safe (what comes to mind: multiple GPUs and running tests in parallel)?

rwgk (commented):
Cursor noted:

Inconsistent `__repr__` styles

- `Device`: `<Device 0 (NVIDIA GeForce...)>` (angle brackets, no `handle=`)
- `Stream`: `Stream(handle=0x...)` (constructor-style)
- `LaunchConfig`: `LaunchConfig(grid=..., cluster=..., ...)` (constructor-style, all attrs)

Are these design choices? Could we make the fully, or maybe more, uniform?

cuda_core/cuda/core/_memory/_buffer.pyx
```diff
@@ -334,6 +334,9 @@ cdef class Buffer:
     def __hash__(self) -> int:
         return hash((as_intptr(self._h_ptr), self._size))
 
+    def __repr__(self) -> str:
+        return f"Buffer(ptr={as_intptr(self._h_ptr):#x}, size={self._size})"
```
kkraus14: Given we have a stream associated to the Buffer that we ultimately deallocate with, it may be nice to include that in the repr?

cuda_core/cuda/core/_module.pyx
kkraus14: My 2c: trigger lazy execution on `__repr__` is a relatively common pattern in Python where I think this makes sense.

cuda_core/cuda/core/_stream.pyx
```diff
@@ -221,6 +221,9 @@ cdef class Stream:
             return NotImplemented
         return as_intptr(self._h_stream) == as_intptr((<Stream>other)._h_stream)
 
+    def __repr__(self) -> str:
+        return f"Stream(handle={as_intptr(self._h_stream):#x})"
```
kkraus14: Because streams are tied to a specific context, does it make sense to capture the context here as well? Similar question for other types tied to a specific context.

cuda_core/cuda/core/_module.pyx
Andy-Jost: I also prefer keeping the lazy trigger here, rather than exposing a more complex object state to users.

cuda_core/tests/test_object_protocols.py
Andy-Jost: Moved to `sample_device_alt` and fixed.

cuda_core/tests/test_object_protocols.py
Andy-Jost: The hash contract is about correctness. Every object could return, say, `0` and meet the contract, but performance would suffer.

I put in this test as a sanity check. As a practical matter, we intend to use hashes for which collisions are extremely unlikely or even impossible.

cuda_core/tests/test_object_protocols.py
Andy-Jost: This is a heuristic test guarding against possible changes to return simple values likely to cause collisions.

cuda_core/cuda/core/_stream.pyx
Andy-Jost: Yes for streams; it will appear in the next upload. I considered doing this for `Event`s but decided against it because they can be imported via IPC, in which case there is no context.

cuda_core/tests/test_object_protocols.py
rwgk: Seems reasonable, but I'd definitely leave a comment in the code (one-liner).

> extremely unlikely

Without at least a hint, people will be scratching their heads.

cuda_core/tests/test_object_protocols.py
```diff
@@ -0,0 +1,315 @@
...
+
+@pytest.mark.parametrize("fixture_name", API_TYPES)
+def test_equality_basic(fixture_name, request):
+    """Object equality: reflexive, not equal to None or other types."""
+    obj = request.getfixturevalue(fixture_name)
+    assert obj == obj, "reflexive equality failed"
+    assert obj != None, "should not equal None"  # noqa: E711
+    assert obj != "string", "should not equal unrelated type"
+    if hasattr(obj, "handle"):
+        assert obj != obj.handle, "should not equal its own handle"
```
cpcloud: All of these `assert`s have explicit error messages, which will override pytest's pretty printing of failures. Can we just leave these off for now?

cuda_core/tests/test_object_protocols.py
cpcloud: I agree that this test is probably not adding much value.

cuda_core/tests/test_object_protocols.py
Andy-Jost: comments added

cuda_core/tests/test_object_protocols.py
Andy-Jost: Alright, removed

Andy-Jost:
/ok to test https://github.com/NVIDIA/cuda-python/commit/04d3275e96fb7e8b7104f6ff5e18d02600efc30c

Issue #659, opened 2025-05-27
[BUG]: CUDA API calls through `cuda-bindings` 3x slower than direct CUDA C++ API calls


lucifer1004:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Performance

### Component

cuda.bindings

### Describe the bug

I have a same CUDA kernel, and I am calling it using two different methods.

1. Compile the kernel to `.cubin`, load it in Python and call it using `cuda-bindings`. The TMA descriptors are also created via `cuda-bindings`.
2. Compile the kernel and relevant C++ host functions (including TMA desc, kernel attribute and launch config) to `.so`, load the `.so` in Python and call the C++ host function.

As revealed in the two timelines below, the first path is 3x slower than the second path. 
<img width="582" alt="Image" src="https://github.com/user-attachments/assets/b6962032-34ce-4180-aed7-09baeabffac2" />
<img width="803" alt="Image" src="https://github.com/user-attachments/assets/597bc426-2ccd-47a2-9a04-b75c66133dc6" />

### How to Reproduce

Follow the instructions in the reproduction repo https://github.com/lucifer1004/cuda-python-repro

### Expected behavior

CUDA Python's perf should be on par with CUDA C++.

### Operating System

Ubuntu Linux 24.04

### nvidia-smi output

+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 550.54.15              Driver Version: 550.54.15      CUDA Version: 12.9     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA H100 80GB HBM3          On  |   00000000:53:00.0 Off |                    0 |
| N/A   27C    P0             69W /  700W |       0MiB /  81559MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   1  NVIDIA H100 80GB HBM3          On  |   00000000:64:00.0 Off |                    0 |
[...]

PR #1543, opened 2026-01-29
Use faster method to get enum values

mdboom:
See [background discussion here](https://github.com/NVIDIA/cuda-python/issues/659#issuecomment-3820026706).

It turns out that instead of

```
cdef cydriver.CUtensorMapInterleave cyinterleave = interleave.value
```

doing this:

```
cdef cydriver.CUtensorMapInterleave cyinterleave = int(interleave)
```

reduces the function call overhead of `cuTensorMapEncodeTiled` by about 25%.  Not all API functions have that many enum arguments (5), so I wouldn't expect such a huge win across the board, but seems very worth doing nonetheless.

The reason this works is because `IntEnum` inherits from `int`, so there is a good fast path in C to grab its value quickly. `enum.value` has to do both a dictionary lookup for `value` and run the Python-written value property code.

### DIFF
--- a/cuda_bindings/cuda/bindings/driver.pyx.in
+++ b/cuda_bindings/cuda/bindings/driver.pyx.in
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2021-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2021-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
 
 # This code was automatically generated with version 13.1.0. Do not modify it directly.
@@ -8398,7 +8398,7 @@ cdef class CUstreamMemOpWaitValueParams_st:
         return _dict_CUstreamBatchMemOpType[self._pvt_ptr[0].waitValue.operation]
     @operation.setter
     def operation(self, operation not None : CUstreamBatchMemOpType):
-        self._pvt_ptr[0].waitValue.operation = operation.value
+        self._pvt_ptr[0].waitValue.operation = int(operation)
     {{endif}}
     {{if 'CUstreamBatchMemOpParams_union.waitValue.address' in found_struct}}
     @property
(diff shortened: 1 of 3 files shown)


kkraus14 (approved):

kkraus14:
@mdboom great find. Is this something where the Cython generated cpp code could be optimized in the currently slow case to avoid all of those Python calls?

mdboom:
> @mdboom great find. Is this something where the Cython generated cpp code could be optimized in the currently slow case to avoid all of those Python calls?

Currently cuda-bindings' enums are based on the stdlib `enum.IntEnum`, which is implemented in Python and therefore we generate Python classes (not `cdef` classes) to define our enums.  That means that all uses of them in Cython must go through Python, and additionally, stdlib `enum` is notoriously slow.  They could be replaced with a more efficient C extension type.  For the common case of just accessing the values in the enum, this would be quite easy to replace.  However stdlib `enum` has a bunch of other features (like iterating over the values) that our users might be relying on that we would have to recreate (or methodically decide which to deprecate etc.).  I think we should definitely do that at some point, as it's likely to be a big win (especially on import time).

jakirkham:
Cython does have a `cpdef enum`, which has a good mix of the C-level efficiency with the Python `Enum` behavior. Maybe worth considering

https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#structs-unions-enums

leofang:
@mdboom it'd be nice to regenerate and backport this PR to `12.9.x`.

@jakirkham Thanks for reminder. We avoided `cpdef enum` because of the namespace pollution, but it seems fixed since Cython 3.1.0. Would be a good idea to revisit this approach indeed. We've spent too much time hunting down Python `Enum`/`IntEnum` related perf issues (last occurrence: https://github.com/NVIDIA/cuda-python/issues/439#issuecomment-2773941585).

mdboom:
> Cython does have a `cpdef enum`, which has a good mix of the C-level efficiency with the Python `Enum` behavior. Maybe worth considering
> 
> https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#structs-unions-enums

Unfortunately, `cpdef enum` is just syntactic sugar for the Python stdlib `enum.Enum`, and generating a C-level enum at the same time.  We already have a C-level enum in our `cy*` layer, and the Python side is where all of the performance issues are, which Cython `cpdef enum` unfortunately doesn't address.  Measurement of [a prototype](https://github.com/NVIDIA/cuda-python/compare/main...mdboom:cuda-python:cpdef-enum?expand=1) shows this [actually regresses import times](https://gist.github.com/mdboom/78f61edcb7bd33f7b37e97e5df2b8461).

More info here: https://github.com/NVIDIA/cuda-python/issues/1557

jakirkham:
Thanks Michael and Leo! 🙏

Michael have you shared these issues with the Cython team?

mdboom:
> Michael have you shared these issues with the Cython team?

I wouldn't necessarily characterize them as Cython issues.  It's more that Cython is just a weird mishmash of performance pitfalls that I'm not sure could be resolved without changing fundamentally what it is ;)

PR #1544, opened 2026-01-29
Make cuda.bindings.nvml public

mdboom:
With all of the APIs vetted (at least from what they look like on the outside), I think we are ready to make this public.  :tada:


cuda_core/cuda/core/system/_system.pyx
```diff
@@ -19,13 +19,17 @@ else:
...
-    from cuda.bindings import _nvml as nvml
-    # TODO: We need to be even more specific than version numbers for development.
-    # This can be removed once we have a release including everything we need.
-    for member in ["FieldId", "ClocksEventReasons"]:
-        if not hasattr(nvml, member):
-            CUDA_BINDINGS_NVML_IS_COMPATIBLE = False
-            break
+    try:
+        from cuda.bindings import nvml
+    except ImportError:
```
rwgk:
```suggestion
    except ModuleNotFoundError:
```

rwgk (approved):
Awesome!

`ImportError` can mask bugs. `ModuleNotFoundError` seems safer here?

cuda_core/cuda/core/system/_system.pyx
mdboom: Good catch, thanks.

kkraus14:
It looks like every `12.9` job failed here. How / why was this merged?

PR #1547, opened 2026-01-30
Fix CI checks job to detect test failures

kkraus14:
## Summary

- Fix the `checks` job to detect test failures, not just cancellations
- The bug allowed PRs to merge even when tests failed because the checks job only verified `result == 'cancelled'` but not `result == 'failure'`

## Test plan

- [ ] Verify CI runs and the checks job correctly fails when tests fail


mdboom (approved):

.github/workflows/ci.yml
cpcloud, line 215: I can't remember if this necessary, but what about `${{ !failed() }}` here to avoid the extra conditionals?

kkraus14: We need to block if the job is cancelled as well, so it can't just be `!failed()`

leofang:
I wonder if we should discuss reverting https://github.com/NVIDIA/cuda-python/pull/1008 and getting back to the old way of gating the checks... This seems error prone and not scalable to me.

PR #1548, opened 2026-01-30
Redo of Make cuda.bindings.nvml public (#1544)

mdboom:


cuda_core/cuda/core/system/_system.pyx
kkraus14:
```suggestion
# This file needs to either use NVML exclusively, or when `cuda.bindings.nvml`
```

kkraus14 (approved):

Issue #1079, opened 2025-10-03
Cythonize `_linker.py`


leofang:
Ideally this should be P0 because numba-cuda uses `Linker` and this could be in a hot spot. However, by lowering to Cython it means `cuda.core` has to bump the lower bound for `cuda.bindings` from 12.6 to 12.8 (the first `cuda.bindings` release that has nvJitLink bindings included). We might want to wait for a few more months before pushing this?

Andy-Jost:
Update `Context.handle` and `Linker.handle` to return `0` (not `None`) when closed.

See https://github.com/NVIDIA/cuda-python/pull/1565#discussion_r2775466401

PR #1565, opened 2026-02-03
Cythonize _program.py

### LINKED ISSUE #1082: Cythonize `_program.py`

Andy-Jost:
## Summary

Converts `_program.py` to Cython (`_program.pyx`) with full cythonization of NVRTC and NVVM paths, RAII-based resource management, and `nogil` error handling.

### Cythonization
- `_program.py` → `_program.pyx` with `cdef class Program`
- New `_program.pxd` with typed attribute declarations
- `Program_init` cythonized with direct `cynvrtc`/`cynvvm` calls in `nogil` blocks
- `Program_compile_nvrtc` and `Program_compile_nvvm` use `nogil` for C API calls
- Long methods factored into `cdef inline` helpers

### Resource Handles
- Add `NvrtcProgramHandle` and `NvvmProgramHandle` (`std::shared_ptr`-based RAII)
- Add `create_nvrtc_program_handle()` and `create_nvvm_program_handle()`
- Remove `_MembersNeededForFinalize` / weakref.finalize pattern
- `as_py(NvvmProgramHandle)` returns Python `int` (matches cuda_bindings design)

### Error Handling
- Add `HANDLE_RETURN_NVRTC` for `nogil` NVRTC error handling with program log
- Add `HANDLE_RETURN_NVVM` for `nogil` NVVM error handling with program log
- Add `NVVMError` exception class
- Simplify `HANDLE_RETURN` to directly take `cydriver.CUresult`

### Validation & API
- Add `SUPPORTED_TARGETS` map for unified backend/target validation
- Add public `Program.driver_can_load_nvrtc_ptx_output()` API
- Consistent error message format for unsupported targets

### Code Quality
- Module docstring, `__all__`, type alias section added
- Docstrings updated to use `:class:` refs and public paths
- Type annotations added to public methods
- `Union` import removed in favor of `|` syntax
- Compiler options use `std::vector<const char*>` instead of `malloc`/`free`

### Tests
- Extend `test_object_protocols.py` with `Program` and `ObjectCode` variations
- Update `test_program.py`: test `close()` idempotency instead of handle state
- Update error message regex for unified format

## Test plan

- [x] All existing `test_program.py` tests pass
- [x] All existing `test_module.py` tests pass
- [x] `test_object_protocols.py` passes with new variations
- [x] `pre-commit` passes

Closes #1082

### DIFF
--- a/cuda_bindings/tests/nvml/test_cuda.py
+++ b/cuda_bindings/tests/nvml/test_cuda.py
@@ -1,6 +1,8 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
 
+import os
+
 import cuda.bindings.driver as cuda
 from cuda.bindings import nvml
 
@@ -54,4 +56,12 @@ def test_cuda_device_order():
     cuda_devices = get_cuda_device_names()
     nvml_devices = get_nvml_device_names()
 
-    assert cuda_devices == nvml_devices, "CUDA and NVML device lists do not match"
+    if "CUDA_VISIBLE_DEVICES" not in os.environ:
+        # If that environment variable isn't set, the device lists should match exactly
+        assert cuda_devices == nvml_devices, "CUDA and NVML device lists do not match"
+    else:
+        # If the environment variable is set, there may possibly be fewer CUDA devices,
+        # and each of them should still be found in NVML devices.
+        assert len(cuda_devices) <= len(nvml_devices)
+        for cuda_device in cuda_devices:
+            assert cuda_device in nvml_devices, f"CUDA device {cuda_device} not found in NVML device list"


cuda_core/tests/test_module.py
```diff
@@ -146,7 +146,7 @@ def test_object_code_load_ptx(get_saxpy_kernel_ptx):
     sym_map = mod.symbol_mapping
     mod_obj = ObjectCode.from_ptx(ptx, symbol_mapping=sym_map)
     assert mod.code == ptx
-    if not Program._can_load_generated_ptx():
+    if not Program.driver_can_load_nvrtc_ptx_output():
```
Andy-Jost: This is worthy of discussion. I changed this private API into a public one after refactoring because it seems somewhat useful to an end user. An alternative would be to continue using the private API if the value here is too low.

cuda_core/cuda/core/_program.pyx
```diff
@@ -0,0 +1,914 @@
...
+        weakref.finalize(program_obj, self.close)
+
+    def close(self):
+        if self.handle is not None:
+            if self.backend == "NVRTC":
+                handle_return(nvrtc.nvrtcDestroyProgram(self.handle))
+            elif self.backend == "NVVM":
+                nvvm = _get_nvvm_module()
+                nvvm.destroy_program(self.handle)
+            self.handle = None
```
Andy-Jost: I am keeping MNFF and the `weakref.finalize` cleanup for this PR. In a subsequent PR, I plan to extend the resource handles to `nvrtc` and `nvvm` and add `nogil` blocks.

cuda_core/cuda/core/_cpp/resource_handles.hpp
```diff
@@ -6,9 +6,16 @@
...
 #include <cuda.h>
+#include <nvrtc.h>
 #include <cstdint>
 #include <memory>
 
+// Forward declaration for NVVM - avoids nvvm.h dependency
+// Use void* to match cuda.bindings.cynvvm's typedef for compatibility
+#ifndef CYTHON_EXTERN_C
+typedef void *nvvmProgram;
+#endif
```
Andy-Jost: Since nvvm is an optional dependency, I opted to define nvvmProgram this way rather than include the header.

cuda_core/tests/test_module.py
rwgk: I'd definitely not make the API change in this PR, in part so that such a subtle change doesn't get drowned out by the many other changes here, but also to avoid entangled unintended side-effects from the disjoint changes (those tend to be hard to sort out, especially later, after this PR is merged already).

Andy-Jost: But this simply renames a private function, making it public. The implementation remains the same. Do you think there could be unintended side-effects from that?

The questions I'm interested in are whether we expect users to end up with systems that can generate unloadable PTX, and if so, can they benefit from a function to check it?

rwgk: That's a good question, but in my experience it's usually far better to discuss separately. E.g. yesterday I already got derailed here looking through the very large diffs. I saw there were a bunch of test failures, then I searched the code and saw new code still referencing `_can_load_generated_ptx`, then decided to suggest it's best to not convolute things.

What makes it beneficial to tag on the change & discussion here?

Andy-Jost: > What makes it beneficial to tag on the change & discussion here?

CI and review latency, TBH. If changes were much cheaper I would break things down much more.

rwgk: I'm not the best to judge if this should stay private or become public. If public, I'd consider `Program.nvrtc_generated_ptx_is_loadable()` as the API.

cuda_core/cuda/core/_module.pyx
```diff
@@ -816,7 +816,8 @@ cdef class ObjectCode:
         try:
             name = self._sym_map[name]
         except KeyError:
-            name = name.encode()
+            if isinstance(name, str):
+                name = name.encode()
```
Andy-Jost, line 820: Slipped in this small fix. `ObjectCode.get_kernel` failed when given bytes. New test in `test_module.py` 107-8

cuda_core/cuda/core/_cpp/resource_handles.hpp
```diff
@@ -6,9 +6,14 @@
 
 #include <Python.h>
 #include <cuda.h>
+#include <nvrtc.h>
 #include <cstdint>
 #include <memory>
 
+// Forward declaration for NVVM - avoids nvvm.h dependency
+// Use void* to match cuda.bindings.cynvvm's typedef
+using nvvmProgram = void*;
```
leofang, line 15: yes this is a nice outcome of using megaheader (cybind's feature) 👍

cuda_core/cuda/core/_module.pyx
leofang: Hmmm, I thought `name` was designed to be typed as `str`?

cuda_core/cuda/core/_program.pxd
```diff
@@ -0,0 +1,15 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
leofang:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

cuda_core/tests/test_module.py
```diff
@@ -104,6 +104,8 @@ def test_get_kernel(init_cuda):
     kernel = object_code.get_kernel("ABC")
     assert object_code.handle is not None
     assert kernel.handle is not None
+    # Also works with bytes
+    assert object_code.get_kernel(b"ABC").handle is not None
```
leofang, line 109: alright, so we make `name` support both str and bytes, not bad

cuda_core/tests/test_module.py
leofang: I would agree with Ralf here that it's best to split into a standalone (and small) PR, and we continue the discussion there. I am not 100% sure attaching it to `Program` is the best choice. Maybe making it as a free function under `cuda.core.utils` should be considered.

`cuda.bindings.utils` might also be worth considering. We already have a few PTX-related utilities there:
https://nvidia.github.io/cuda-python/cuda-bindings/latest/module/utils.html
Grep in the codebase and see how they are used in our tests.

> The questions I'm interested in are whether we expect users to end up with systems that can generate unloadable PTX, and if so, can they benefit from a function to check it?

PTX does not honor MVC (CUDA Minor Version Compatibility). More precisely, driver JIT does not honor MVC. A PTX generated by the CUDA compiler toolchain 13.x cannot be ~~loaded~~ JIT'd by the driver 13.y, for x > y. In other words, the driver must be at least as new as the CTK components that generate the PTX, in order to JIT and then load it. We build machineries in cuda.core so as to split JIT (PTXs) from loading (CUBINs).

cuda_core/tests/test_program.py
```diff
@@ -339,7 +336,6 @@ def test_ptx_program_with_various_options(init_cuda, ptx_code_object, options):
     assert program.backend == ("driver" if is_culink_backend else "nvJitLink")
     program.compile("cubin")
     program.close()
-    assert program.handle is None
```
leofang, line 342: Q: We don't purge `.handle` after close?

cuda_core/cuda/core/_utils/cuda_utils.pxd
```diff
@@ -22,7 +18,9 @@ ctypedef fused integer_t:
 cdef const cydriver.CUcontext CU_CONTEXT_INVALID = <cydriver.CUcontext>(-2)
 
 
-cdef int HANDLE_RETURN(supported_error_type err) except?-1 nogil
+cdef int HANDLE_RETURN(cydriver.CUresult err) except?-1 nogil
+cdef int HANDLE_RETURN_NVRTC(cynvrtc.nvrtcProgram prog, cynvrtc.nvrtcResult err) except?-1 nogil
+cdef int HANDLE_RETURN_NVVM(cynvvm.nvvmProgram prog, cynvvm.nvvmResult err) except?-1 nogil
```
leofang, line 23: Very big nitpick. Not a blocker.

FYI. The original design was to use `HANDLE_RETURN` to consume a fused type, which is a layman's way of writing template functions in Cython. Then we could consume all possible CUDA return types, and not thinking about which backend/library we are dealing with (driver/runtime/nvrtc/nvvm/...).

I can see the story is not fully satisfying here, in that compiler libs also need to take a handle in order to spit out error logs. I wonder if we would make it cleaner by having `HANDLE_RETURN_COMPILER` that takes a fused type over the return types of NVRTC/NVVM/....

cuda_core/cuda/core/_program.pxd
```diff
@@ -0,0 +1,15 @@
...
+
+from ._resource_handles cimport NvrtcProgramHandle, NvvmProgramHandle
+
+
+cdef class Program:
+    cdef:
+        NvrtcProgramHandle _h_nvrtc
+        NvvmProgramHandle _h_nvvm
+        str _backend
+        object _linker  # Linker
```
leofang, line 13: TODO: cythonize linker! :D

cuda_core/cuda/core/_program.pyx
leofang, line 1: I reviewed everything but this file, will finish tomorrow 🙏

cuda_core/cuda/core/_cpp/resource_handles.hpp
leofang: In fact including nvvm.h does not help and would break. The choice is correct.

cuda_core/cuda/core/_module.pyx
Andy-Jost: The docstring lists the type as "any" and I found other APIs that accept str or bytes, so I figured this was an improvement for the sake of consistency.

cuda_core/tests/test_program.py
Andy-Jost: The handle does indeed become invalid but I changed it to return `0`. I surveyed the `handle` properties and found almost all return `0` for invalid. See table below -- `Context` and `Linker` are the remaining exceptions and I plan to update them soon.

Here's a typical example:

```
@property
def handle(self) -> cuda.bindings.driver.CUstream:
    return as_py(self._h_stream)  # convert handle to int then CUstream
```

Both `0` and `None` are defensible, but the APIs should be consistent. I'm open to making another pass if we decide on `None` (separate discussion). If we go that route, I'd like to make a separate PR that updates all type annotations and docstrings at once.

Summary of closed handles:
```
  | Class           | Returns after close() | Mechanism                                                          |
  |-----------------|-----------------------|--------------------------------------------------------------------|
  | Context         | None                  | Explicit null check: if self._h_context.get() == NULL: return None |
  | Stream          | CUstream(0)           | as_py() wraps as_intptr() which returns 0 for reset shared_ptr     |
  | Event           | CUevent(0)            | Same as Stream                                                     |
  | MemoryPool      | CUmemoryPool(0)       | Same as Stream                                                     |
  | Buffer          | 0 (int)               | Uses as_intptr() directly, returns bare integer 0                  |
  | Kernel          | CUkernel(0)           | Same as Stream                                                     |
  | ObjectCode      | CUlibrary(0)          | Same as Stream                                                     |
  | Linker          | None                  | MNFF pattern: self.handle = None in close()                        |
  | Program (NVRTC) | nvrtcProgram(0)       | as_py() wraps 0 via make_py("cuda.bindings.nvrtc", ...)            |
  | Program (NVVM)  | 0 (int)               | PyLong_FromSsize_t(0) — NVVM has no wrapper class                  |
```

cuda_core/cuda/core/_utils/cuda_utils.pxd
Andy-Jost: I looked carefully into "overloading" `HANDLE_RETURN` but ran into difficulty, mainly with the library handle as you note.

With something like `HANDLE_RETURN_COMPILER`, the library handles would also need to be fused. But when creating programs, no handle exists yet, so we pass `NULL`. But Cython can't deduce the fused type from that so we need an explicit cast:

`HANDLE_RETURN_COMPILER(<cynvrtc.nvrtcProgram>NULL, cynvrtc.nvrtcCreateProgram(...))`

Since the target library is always known at the call site anyway, I decided it was simpler to use separate functions than to fight the fused type.

Andy-Jost: I also considered putting the program argument second in the call to `HANDLE_RETURN` and letting it default to `NULL`, but Cython does not allow arguments of fused type to be defaulted.

cuda_core/cuda/core/_program.pxd
Andy-Jost: Next up!

cuda_core/tests/test_module.py
Andy-Jost: Ok, then for this PR I'll avoid changing the public API. We'll end up with a private `cpdef` function used in testing and we can defer the other questions.

cuda_core/cuda/core/_module.pyx
stiepan, line 798: Would it be OK to call `launch(kernel, ...)` with the same `kernel` instance from multiple Python threads without locking? 

The context here is in-memory JIT caching. Let's say I have a single ObjectCode instance containing a kernel that I may need to repeatedly call from multiple threads. I don't want the threads to acquire locks every-time they launch the kernel (one time for a setup is fine). Is any of the following safe?
* There's a single `kernel=obj_code.get_kernel(...)` instance, shared by all threads and used in launch(...)  call without mutex.
* There's a single ObjectCode instance, each thread acquires mutex once and creates `local_kernel = obj_code.get_kernel()` kernel instance that it can later use without locking with `launch(...)`?

stiepan: Sorry, I know it may be a bit off-topic to this PR.

cuda_core/cuda/core/_utils/cuda_utils.pxd
leofang: > But Cython can't deduce the fused type from that so we need an explicit cast

Yes. This is a bit annoying. You can search how fused types are used in in cuda_bindings. We needed it there because all code are auto-generated and it's not too bad. Hand-writing them, OTOH, indeed is very verbose as you said and it defeats the purpose of having a type-erasing-ish wrapper. Resolving this thread.

cuda_core/tests/test_program.py
leofang: Thanks for the nice survey, Andy! Let's make an issue to track it and address it in a separate PR.

cuda_core/cuda/core/_module.pyx
Andy-Jost: Yes, both scenarios you describe should be safe. Multiple host threads can concurrently call `cuLaunchKernel` (or `cuLaunchKernelEx`) with the same `CUfunction`/`CUkernel` handle.

cuda_core/cuda/core/_utils/cuda_utils.pyx
```diff
@@ -32,6 +34,10 @@ class NVRTCError(CUDAError):
     pass
 
 
+class NVVMError(CUDAError):
```
leofang: Important: There is already a `NVVMError` in cuda-bindings (due to a different codegen that we use for all modules, other than driver/runtime/NVRTC). We should not re-invent it. It'd make downstream hard to decide if they want to catch the cuda-bindings one or cuda-core one.

cuda_core/cuda/core/_utils/cuda_utils.pyx
```diff
@@ -57,10 +63,72 @@ def _reduce_3_tuple(t: tuple):
...
+cdef int HANDLE_RETURN_NVRTC(cynvrtc.nvrtcProgram prog, cynvrtc.nvrtcResult err) except?-1 nogil:
+    """Handle NVRTC result codes, raising NVRTCError with program log on failure."""
+    if err == cynvrtc.nvrtcResult.NVRTC_SUCCESS:
+        return 0
+    # Get error string (can be called without GIL)
+    cdef const char* err_str = cynvrtc.nvrtcGetErrorString(err)
+    # Get program log size for additional context
+    cdef size_t logsize = 0
+    if prog != NULL:
+        cynvrtc.nvrtcGetProgramLogSize(prog, &logsize)
```
leofang: nit: can we just move these lines to `_raise_nvrtc_error`?

cuda_core/cuda/core/_utils/cuda_utils.pyx
```diff
@@ -57,10 +63,72 @@ def _reduce_3_tuple(t: tuple):
...
+cdef int HANDLE_RETURN_NVVM(cynvvm.nvvmProgram prog, cynvvm.nvvmResult err) except?-1 nogil:
+    """Handle NVVM result codes, raising NVVMError with program log on failure."""
+    if err == cynvvm.nvvmResult.NVVM_SUCCESS:
+        return 0
+    # Get error string (can be called without GIL)
+    cdef const char* err_str = cynvvm.nvvmGetErrorString(err)
+    # Get program log size for additional context
+    cdef size_t logsize = 0
+    if prog != NULL:
+        cynvvm.nvvmGetProgramLogSize(prog, &logsize)
```
leofang: ditto, move to `_raise_nvvm_error`?

mdboom (commented):
Has the performance improvement of this change been measured?  On first glance, this looks like deeply Pythonic code that wouldn't benefit much from Cythonizing.

Andy-Jost:
> Has the performance improvement of this change been measured? On first glance, this looks like deeply Pythonic code that wouldn't benefit much from Cythonizing.

No, it has not been measured. We are lacking performance tests AFAIK. It would be very helpful if we could develop a performance suite under this repo (or start a new repo) for collecting those numbers. I would expect the vast majority of time in the compilation flow to be spent waiting for the compiler. Now, at least, we release the GIL during that, so there could be an improvement in certain scenarios.

cuda_core/cuda/core/_utils/cuda_utils.pyx
Andy-Jost: It appears the error in cuda.bindings does not accept a string (it only reports the library error message):

Maybe cuda.core should define an exception that derives from the cuda.bindings one and adds the details?

```
# cuda/bindings/nvvm.pyx
38  class nvvmError(Exception):
  1
  2     def __init__(self, status):
  3         self.status = status
  4         s = Result(status)
  5         cdef str err = f"{s.name} ({s.value})"
  6         super(nvvmError, self).__init__(err)
  7
  8     def __reduce__(self):
  9         return (type(self), (self.status,))
```

cuda_core/cuda/core/_cpp/resource_handles.hpp
```diff
@@ -293,6 +349,14 @@ inline CUkernel as_cu(const KernelHandle& h) noexcept {
     return h ? *h : nullptr;
 }
 
+inline nvrtcProgram as_cu(const NvrtcProgramHandle& h) noexcept {
+    return h ? *h : nullptr;
```
rparolin, line 353: Consider: Because you are taking a std::shared_ptr by `const&` this clearly won't inc/dec any reference counts. However it doesn't prevent other threads of execution from doing so. Which leaves this function vulernable to a potentially changing std::shared_ptr state where an interleaving could mean the `operator bool` check reads an non-null address but it ends up being null when you attempt to dereference it via the `std::shared_ptr::operator*`.  Considering taking the parameter by value to resolve.

cuda_core/cuda/core/_utils/cuda_utils.pyx
```diff
@@ -20,6 +20,10 @@ except ImportError:
     from cuda import cudart as runtime
     from cuda import nvrtc
 
+from cuda.bindings.nvvm import nvvmError
+
+from cuda.bindings cimport cynvrtc, cynvvm
```
leofang, line 25: Sorry, just one last finding that I should have caught sooner.

The NVVM bindings were not added until cuda-bindings 12.9.0, while cuda-core support matrix is cuda-bindings 12.8.0+ as of v0.4.0 ([link](https://nvidia.github.io/cuda-python/cuda-core/latest/release/0.4.0-notes.html#breaking-changes)). We probably need a rel-note on this. 

In the long term, we might need to evolve more aggressively in the supported cuda-bindings versions, due to our usage at the Cython level (which cannot be worked around via try-except). Probably a good topic to discuss before going GA...

leofang:
> > Has the performance improvement of this change been measured? On first glance, this looks like deeply Pythonic code that wouldn't benefit much from Cythonizing.
> 
> No, it has not been measured. We are lacking performance tests AFAIK. It would be very helpful if we could develop a performance suite under this repo (or start a new repo) for collecting those numbers.

This is to be done in #1579. That said
1. I agree with @mdboom. Having a perf number to guide the lowering is always needed, because Cython occasionally can be surprising. 
   - For this case, I think making the `Program()` constructor take an empty kernel as input, as well as calling `.compile()`, could be two good timing proxies. [The snippets here](https://github.com/NVIDIA/cuda-python/pull/1390#issue-3736604729) can be reused (and eventually incorporated into the benchmark suite).
2. @stiepan has previously confirmed (though I dunno how - care to comment, Kamil? 🙂 ) that cythonizing the NVRTC path is important for perf. CuPy also has it cythonized, so in all likelihood this is a good step forward
3. nvmath-python needs this PR so that they can shake off some illegal code (the alternative is we merge the copy kernel #1367, but we veto'd it and I don't want us to keep rejecting their requests)

cuda_core/cuda/core/_utils/cuda_utils.pyx
leofang: ah... ok, interesting, so did we always raise `CUDAError` before this PR?

stiepan:
Hi @leofang 

Regarding the overhead, sorry, it's been some time ago, so I don't have compelling breakdown of the overhead I've seen with cuda-core. As I am testing with the changes here, using cuda-core and cuda-bindings for compilation and kernel launch, I no longer see a slowdown and we're gladly move to cuda-core in nvmath for jit compilation toolchain.

That being said, I think the crucial part is the kernel launch overhead. Two issues that I recall standing out, were overhead of kernel arguments preparation and missing gil releasing. The latter is fixed for quite sime time, thanks! 

For kernel arguments, I would like to check for you recommendations here. The kernel accepts (templated) struct instance comprising many simple types. I'd like to avoid breaking it down to the variadic number of simple types and building the args list in the runtime. Instead I can just prepare a struct instance on the host and pass void** ptr for driver to handle the rest. Is it doable with `ccx.launch`? Or is it something that cuda-core even would like to have (I can see how this may be too unpythonic and too easy to get wrong to land in cuda-core)? Currently, I just access Kernel.handle and pass it to the Python cuda-bindings driver API (so no compile-time deps) - driver.cuLaunchKernel seems to do just fine. Here, the question is whether Kernel.handle is here to stay and will be safe to use with cuKernelLaunch? AFAIK this property was added recently.

cuda_core/cuda/core/_utils/cuda_utils.pyx
Andy-Jost: Hmm, no, originally the exception was caught, annotated, and re-raised. I'll take a closer look.

```
    except Exception as e:
        ...
        # Starting Python 3.11 we could also use Exception.add_note() for the same purpose, but
        # unfortunately we are still supporting Python 3.10...
        e.args = (e.args[0] + (f"\nNVVM program log: {error_log}" if error_log else ""), *e.args[1:])
        raise e
```

cuda_core/cuda/core/_cpp/resource_handles.hpp
Andy-Jost: I don't think that's possible: the calling scope is required to own a reference so the refcount cannot drop to zero.

Andy-Jost:
Performance Results
=============
GPU: NVIDIA A10, Python 3.14, NVRTC path (C++ code type):

  | Metric                 | main (before) | cythonize-program (after) | Change        |
  |------------------------|---------------|---------------------------|---------------|
  | Program() constructor  | ~9.4 us       | ~5.3 us                   | ~44% faster   |
  | .compile() alone       | ~28,140 us    | ~27,700 us                | ~1.6% (noise) |
  | Program() + .compile() | ~28,250 us    | ~28,040 us                | ~0.7% (noise) |

# Summary
The constructor sees a meaningful ~44% reduction in Python overhead (9.4 -> 5.3 us). The .compile() time is dominated by the NVRTC compiler itself (~28 ms), so the Python-side improvement is negligible relative to that. The GIL-release benefit would only be observable in a multithreaded scenario.

# Benchmark Script
```
  import timeit
  from cuda.core import Device, Program
  dev = Device()
  dev.set_current()
  CODE = r"""
  extern "C" __global__ void empty_kernel(const float* a, const double* b) {
  }
  """
  # Warm up
  p = Program(CODE, code_type="c++")
  obj = p.compile(target_type="cubin")
  dev.sync()
  # Benchmark Program() constructor
  N_CTOR = 10000
  t_ctor = timeit.timeit(lambda: Program(CODE, code_type="c++"), number=N_CTOR)
  print(f"Program() constructor: {t_ctor/N_CTOR*1e6:.2f} us ({N_CTOR} iterations)")
  # Benchmark .compile()
  N_COMPILE = 100
  def compile_once():
      p = Program(CODE, code_type="c++")
      p.compile(target_type="cubin")
  t_compile = timeit.timeit(compile_once, number=N_COMPILE)
  t_compile_only = t_compile - (t_ctor / N_CTOR * N_COMPILE)
  print(f"Program() + compile(): {t_compile/N_COMPILE*1e6:.2f} us ({N_COMPILE} iterations)")
  print(f".compile() alone (est): {t_compile_only/N_COMPILE*1e6:.2f} us")
```

cuda_core/tests/test_program.py
Andy-Jost: Added to https://github.com/NVIDIA/cuda-python/issues/1079#issuecomment-3880047423

cuda_core/cuda/core/_utils/cuda_utils.pyx
Andy-Jost: Fixed in the latest upload. The original behavior is restored.

cuda_core/cuda/core/_utils/cuda_utils.pyx
Andy-Jost: Added the release note. Please make sure I accurately captured the issue.

cuda_core/cuda/core/_utils/cuda_utils.pyx
leofang: Thanks for reminding me about this trick. I implemented and then forgot about it 😄

PR #1568, opened 2026-02-03
Fix to use resident device stream in IPC peer access tests

rwgk:
## Description

This resolves failures in QA environments (nvbug 5821337; internal PR 242).

### DIFF
--- a/cuda_core/tests/memory_ipc/test_peer_access.py
+++ b/cuda_core/tests/memory_ipc/test_peer_access.py
@@ -94,20 +94,21 @@ class TestBufferPeerAccessAfterImport:
         # Test 1: Buffer accessible from resident device (dev1) - should always work
         dev1 = Device(1)
         dev1.set_current()
+        dev1_stream = dev1.create_stream()
         PatternGen(dev1, NBYTES).verify_buffer(buffer, seed=False)
 
         # Test 2: Buffer NOT accessible from dev0 initially (peer access not preserved)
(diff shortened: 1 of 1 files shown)


cuda_core/tests/memory_ipc/test_peer_access.py
```diff
@@ -94,28 +94,29 @@ def child_main(self, mr, buffer):
         # Test 1: Buffer accessible from resident device (dev1) - should always work
         dev1 = Device(1)
         dev1.set_current()
+        dev1_stream = dev1.create_stream()
```
kkraus14, line 97: This stream is used for test2, can we move the creation underneath the test 2 comment?

cuda_core/tests/memory_ipc/test_peer_access.py
```diff
@@ -94,28 +94,29 @@ def child_main(self, mr, buffer):
...
         PatternGen(dev1, NBYTES).verify_buffer(buffer, seed=False)
```
kkraus14, line 98: Just a note that PatternGen uses unified / managed memory underneath it where I suspect this may give us some pain / troubles on Windows?

rwgk: Side remark: I'm working on getting rid of that distraction under #1565 

In the meantime we discovered that the root cause is much deeper.

I saw @testcupy closed the internal PR. Closing this copy, too.

PR #1570, opened 2026-02-03
NVML: Change Inforom-related APIs from str to bytes

mdboom:
Based on a discussion elsewhere with the NVML team, it seems that these APIs should really be `bytes` (containing embedded NULL characters), not `str`.


mdboom:
Converting to draft.  The discussion still continues (elsewhere) about what the correct solution is.

mdboom:
> LGTM, where are the specific sizes coming from?

The come from the header -- our generator tool parses these values in the header automatically.

mdboom:
Closing as WONTFIX.  This is not actually a bug at our end -- there is something else wrong with the hardware producing an invalid result and failing our test.

PR #1574, opened 2026-02-04
ci: setup nightly repeated test runs using pytest-repeat

cpcloud:
Setup nightly runs of tests repeated 100 times for each of cuda-core, cuda-bindings, and cuda-pathfinder.


Andy-Jost (approved):
LGTM

PR #1581, opened 2026-02-05
Faster enum implementation (Pure Python version)

mdboom:
This adds an implementation of integer enums that is significantly faster than the one in the stdlib (`enum.IntEnum`).

There was an alternate implementation of this that used a handwritten C extension, but the performance benefit was minor.  This is both much faster than the stdlib, but also small and easy-to-maintain.

Additionally, this lets us assign docstrings to the enumeration values, which is a feature the stdlib enum doesn't have.  Then all of the generated code has been updated to include docstrings, auto-derived from the headers.

```
+------------------------------+----------+-----------------------+
| Benchmark                    | baseline | fast_enum_py_docs     |
+==============================+==========+=======================+
| enum from int                | 300 ns   | 170 ns: 1.77x faster  |
+------------------------------+----------+-----------------------+
| int from enum                | 55.7 ns  | 57.5 ns: 1.03x slower |
+------------------------------+----------+-----------------------+
| enum value                   | 203 ns   | 79.9 ns: 2.54x faster |
+------------------------------+----------+-----------------------+
| cuda.bindings.driver import  | 38.3 ms  | 32.1 ms: 1.19x faster |
+------------------------------+----------+-----------------------+
| cuda.bindings.runtime import | 73.4 ms  | 61.4 ms: 1.20x faster |
+------------------------------+----------+-----------------------+
| cuda.bindings.nvrtc import   | 62.1 ms  | 55.2 ms: 1.12x faster |
+------------------------------+----------+-----------------------+
```

### DIFF
--- a/cuda_bindings/cuda/bindings/_internal/utils.pyx
+++ b/cuda_bindings/cuda/bindings/_internal/utils.pyx
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
 
@@ -6,6 +6,28 @@ cimport cpython
 from libc.stdint cimport intptr_t
 from libcpp.utility cimport move
 from cython.operator cimport dereference as deref
+
+
+cdef extern from *:
+    """
+    #if defined(__clang__)
+        #define _COMPILER_VERSION ("Clang " __clang_version__)
+    #elif defined(__GNUC__) || defined(__GNUG__)
+        #define _COMPILER_VERSION ("GCC " __VERSION__)
+    #elif defined(_MSC_VER)
+        #define _COMPILER_VERSION ("MSVC " Py_STRINGIFY(_MSC_VER))
+    #else
+        #define _COMPILER_VERSION ("Unknown Compiler")
+    #endif
+    """
+    cdef char *_COMPILER_VERSION
+
+
+cpdef str get_c_compiler():
+    """
+    Returns a string describing the C compiler used to build cuda.bindings
+    """
+    return _COMPILER_VERSION.decode()
 
 
 cdef bint is_nested_sequence(data):
--- a/cuda_bindings/tests/nvml/test_device.py
+++ b/cuda_bindings/tests/nvml/test_device.py
@@ -71,7 +71,7 @@ def test_get_nv_link_supported_bw_modes(all_devices):
     for device in all_devices:
         with unsupported_before(device, None):
             modes = nvml.device_get_nvlink_supported_bw_modes(device)
-        assert isinstance(modes, nvml.NvLinkSupportedBWModes)
+        assert isinstance(modes, nvml.NvLinkSupportedBWModes_v1)
         # #define NVML_NVLINK_TOTAL_SUPPORTED_BW_MODES 23
         assert len(modes.bw_modes) <= 23
         assert not hasattr(modes, "total_bw_modes")
--- a/cuda_bindings/tests/test_utils.py
+++ b/cuda_bindings/tests/test_utils.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
 
 import importlib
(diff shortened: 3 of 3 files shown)


cuda_bindings/cuda/bindings/_internal/_fast_enum.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+"""
+This is a replacement for the stdlib enum.IntEnum.
+
+Notably, it has much better import time performance, since it doesn't need to
+generate and evaluate Python code.
+
+It supports the most important subset of the IntEnum API.  See `test_enum` in
+`cuda_bindings/tests/test_basics.py` for details.
```
kkraus14, line 15: Based on this comment I assume this means that we're only supporting a subset of an `IntEnum` and given we're returning these that would probably constitute an API break?

I.e. because we were previously returning an `IntEnum` as error codes, if someone has code that does something like `isinstance(value, enum.IntEnum)`, that would then start returning `False`?

leofang: Yes. But moving away from Enum/IntEnum is our only route to gain performance on the error handling side. I think this ducktyping is good enough and the break due to type checking seems to be theoretical. Users most likely would just write `isinstance(err, CUresult)` (which is nonbreaking) instead of `isinstance(err, Enum)`.

mdboom: Yeah -- I'm probably being overly cautious with the word `subset` here.  I carefully looked at everything `IntEnum` exposes and replicated it here.  The test even tests `stdlib` and our implementation using the same test code.  You are right that the biggest visible change is the `isinstance(value, enum.IntEnum)` thing, but I agree with @leofang, that seems unlikely.

In any event, I'll add a CHANGELOG entry here.

kkraus14: If we inherited from `enum.IntEnum` instead of `int` would that nullify some of the performance gains?

Could we use `__instancecheck__` to make `isinstance(enum.IntEnum)` and `isinstance(enum.Enum)` work as they did previously?

Either of these would likely limit the breakage to if someone was using anti-patterns like using `type(...)` or something like that.

mdboom: > If we inherited from enum.IntEnum instead of int would that nullify some of the performance gains?

Yeah, unfortunately you can't bring in the value type (`IntEnum`) without bringing in the metaclass and all of its startup cost:

```
TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```

> Could we use __instancecheck__ to make isinstance(enum.IntEnum) and isinstance(enum.Enum) work as they did previously?

Unfortunately not.  `__instancecheck__` is a method on the type/metaclass that you are testing for.  In other words, to make `isinstance(FastEnum.VALUE, enum.IntEnum)` work, we would have to monkeypatch an `__instancecheck__` to `enum.IntEnum`.

[This GitHub code search with a regex](https://github.com/search?q=%2Fisinstance%5C%28.*%2C+IntEnum%5C%29%2F+cuda&type=code&ref=advsearch) (for `isinstance\(.*, IntEnum\)` in a file that also contains the word `cuda` throws up a few possible problematic things.  I guess we need to decide whether the performance matters enough that it's worth the pain to help our users fix those things in a few places.

mdboom: Monkeypatching `IntEnum`'s `__instancecheck__` does work, with the loss of some performance:

```
+------------------------------+-------------------+-----------------------+
| isinstance(enum, IntEnum)    | 48.1 ns           | 80.6 ns: 1.68x slower |
+------------------------------+-------------------+-----------------------+
```

If performance matters to our users, we can suggest updating their code to `isinstance(enum, FastEnum)`, which has no penalty from the monkeypatching.

Maybe this is an ok solution for backward compatibility.

mdboom: @leofang, @kkraus14: So I think we just need to do decide:

1) We merge this as-is where `isinstance(enum, IntEnum)` returns `False`.
2) We include a monkeypatch to make `isinstance(enum, IntEnum)` work.  This makes that isinstance check about 68% slower (but that check is exclusive to frameworks doing things in general, not really in user code), but we wouldn't break dependent libraries.  As with all monkeypatches, unintended consequences are a little hard to reason about.  I /think/ this is safe, but I wouldn't bet my life on it.
3) We don't do this whole thing right now.  (And, as a reminder to myself, to revert the generator changes if that's what we decide).

This is what (2) looks like:

```python
def __instancecheck__(cls, instance) -> bool:
    if isinstance(instance, FastEnum):
        return True
    return issubclass(type(instance), IntEnum)


type(IntEnum).__instancecheck__ = __instancecheck__
```

leofang: I vote for 1 and if there is any bug report we do 2 in a patch release.

rparolin: +1.  I'd vote for 1 and then fallback to 2 if required based on user feedback.

kkraus14: I'm good with that, lets ship it

PR #1584, opened 2026-02-06
Add code coverage support for Windows and linux platform

rluo8:
This is to add code coverage support for Windows platform.

Here are the main steps:
1. Run coverage on Linux runner - linux-amd64-gpu-a100-latest-1 , and upload coverage data and source to artifact. 
2. Run coverage on Windows runner - windows-amd64-gpu-a100-latest-1, and upload coverage to artifact.
3. Download the coverage data from artifact, and then combine the coverage data. Generate coverage report using the combined data of Linux and Windows. Then upload the report to github pages.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -98,7 +98,7 @@ jobs:
 
       - name: Build cuda-pathfinder
         run: |
-          .venv/bin/pip install -v ./cuda_pathfinder
+          .venv/bin/pip install -v ./cuda_pathfinder --group test
 
       - name: Build cuda-bindings
         run: |
@@ -306,6 +306,7 @@ jobs:
         run: |
           .venv/Scripts/pip install -v ./cuda_python_test_helpers
           .venv/Scripts/pip install coverage pytest-cov Cython
+          .venv/Scripts/pip install --group ./cuda_pathfinder/pyproject.toml:test
           .venv/Scripts/pip install --group ./cuda_bindings/pyproject.toml:test
           .venv/Scripts/pip install --group ./cuda_core/pyproject.toml:test
 


rluo8:
Hi @mdboom ,
Could you please take a look?
Thanks!

mdboom:
In order to test this before we merge it, I

- pushed rluo8/main to upstream/windows-coverage
- Manually triggered the coverage action on that branch

The job is running here: https://github.com/NVIDIA/cuda-python/actions/runs/21754529128

.github/workflows/coverage.yml
```diff
@@ -50,6 +50,8 @@ jobs:
 
       - name: Checkout ${{ github.event.repository.name }}
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+        with:
+          fetch-depth: 0
```
mdboom, line 54: Do we need to fetch the whole history?

rluo8: Yes, since cuda_bindings and cuda_core both use setuptools_scm to get their version from git history, fetch-depth: 0 is required. Without full history, setuptools_scm may compute an incorrect version, which causes cuda.core.__init__.py's import-time check to fail with ImportError: cuda.bindings 12.x or 13.x must be installed.

Error like this without fetch-depth:0:

+ cd /__w/cuda-python/cuda-python/.venv/lib/python3.14/site-packages
+ /__w/cuda-python/cuda-python/.venv/bin/pytest -v --cov=./cuda --cov-append --cov-context=test --cov-config=/__w/cuda-python/cuda-python/.coveragerc /__w/cuda-python/cuda-python/cuda_core/tests
ImportError while loading conftest '/__w/cuda-python/cuda-python/cuda_core/tests/conftest.py'.
../../../../cuda_core/tests/conftest.py:7: in <module>
    import helpers
../../../../cuda_core/tests/helpers/__init__.py:10: in <module>
    from cuda.core._utils.cuda_utils import handle_return
cuda/core/__init__.py:14: in <module>
    raise ImportError("cuda.bindings 12.x or 13.x must be installed")
E   ImportError: cuda.bindings 12.x or 13.x must be installed
Error: Process completed with exit code 4.

.github/workflows/coverage.yml
```diff
@@ -90,6 +92,26 @@ jobs:
           host-platform: ${{ env.HOST_PLATFORM }}
           cuda-version: ${{ env.CUDA_VER }}
 
+      - name: Patch cuda_bindings for Cython coverage
```
leofang: Q: What is the reason that we need to patch the build configs?

mdboom: Cython's coverage support assumes the generated files sit alongside the source files.  We very recently changed this to reduce our wheel size, but for this, we need to go back to the old way.  I didn't think it was worth reverting that otherwise good change to the non-common-case of code coverage.

leofang: I see, thanks Mike. I'll think about this in the background. Resolving.

.github/workflows/coverage.yml
```diff
@@ -114,41 +136,394 @@ jobs:
...
+          echo ""
+          echo "[SUCCESS] Coverage reports generated successfully"
 
-      - name: Archive code coverage results
+      - name: Archive combined coverage results
         uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f  # v6.0.0
         with:
-          name: coverage
+          name: coverage-combined
           path: docs/coverage/
```
leofang, line 479: Q: Does each PR (and that built from main) overwrite with each other?

rluo8: Yes, they would overwrite with each other and only keep the latest report in docs/coverage. 
The coverage workflow needs to be manually triggered for PR. And for main branch, it was scheduled to be nightly triggered. Current github page report was from this PR because the nightly main coverage run failed and couldn't generate a report. 
Historical coverage data is available as artifacts uploaded by each coverage run, with a retention period of 7 days (configurable via retention-days in the workflow).
Do we need to keep historical coverage reports in github pages?

.github/workflows/coverage.yml
```diff
@@ -90,6 +92,26 @@ jobs:
...
+
+      - name: Patch cuda_core for Cython coverage
+        run: |
+          # Patch cuda_core build_hooks.py
+          sed -i 's/build_dir=\"build\/cython\",/build_dir=\".\",/' cuda_core/build_hooks.py
+
+          # Patch cuda_core pyproject.toml - add after cuda.core._cpp line
+          sed -i '/\"cuda\.core\._cpp\" = \[\".*\"\]/a \"*\" = [\"*.cpp\"]' cuda_core/pyproject.toml
+
+          echo "Applied coverage patches to cuda_core"
```
kkraus14: Rather than doing these patches I think it would be better to add build commands into the projects that need it and pass the options through via `pip install --config-settings` or using an environment variable.

kkraus14:
Also, should we consider using something like https://codecov.io instead of manually managing all of the coverage reports?

.github/workflows/coverage.yml
rluo8: Yes, using an environment variable would be better and more reliable. There is already a CUDA_PYTHON_COVERAGE variable in the project which enables Cython linetrace. However, due to a recent change in wheel packaging, extra changes are needed in the project setup to properly support the coverage build. @mdboom , would it be all right to make changes to the project setup files for coverage support?

rluo8:
> Also, should we consider using something like https://codecov.io instead of manually managing all of the coverage reports?

Codecov could be a good option. It's a third-party service that might need to be enabled for the project first if using it.

.github/workflows/coverage.yml
mdboom: Yes, it's fine to update them for what we need.  As Keith suggested, if there is a way to just pass in configuration options rather than patching it, that would be preferable.

rluo8: @kkraus14 @mdboom , I have updated the project setup files to use environment variable CUDA_PYTHON_COVERAGE to control the cython files location. And the files included in the wheel were not affected when CUDA_PYTHON_COVERAGE set to 0 or no CUDA_PYTHON_COVERAGE.

PR #1589, opened 2026-02-09
Display C compiler version in logs

mdboom:
It is sometimes useful to know the C compiler that was used to build our wheels.  This just adds it to the build logs.

### DIFF
--- a/.github/workflows/build-wheel.yml
+++ b/.github/workflows/build-wheel.yml
@@ -106,6 +106,16 @@ jobs:
         run: |
           env
 
+      - name: Display compiler version
+        if: ${{ !startsWith(inputs.host-platform, 'win') }}
+        run: |
+          cc --version
+
+      - name: Display compiler version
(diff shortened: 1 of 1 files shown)


leofang:
Q: Do we want a Windows-only or generic solution? Right now it's the former.

At build time, we use cibuildwheels which
- pulls a container on Linux, so we can just reverse-look up the gcc version from which manylinux standard that we pick, there is a list in the manylinux repo
- overlays files on Windows, so we can just reverse-look up the msvc version from which VM image that we pick, there is a list in the GHA image repo

kkraus14 (approved):

mdboom:
> Q: Do we want a Windows-only or generic solution? Right now it's the former.

This has a path for Unix-like and Windows separately.

Though now that I look at this, I think this isn't going to do the right thing at all because this is outside the `cibuildwheel` container.

PR #1591, opened 2026-02-10
Add function to get C compiler used to build cuda.bindings

mdboom:
This sort of metadata is helpful when debugging performance issues.

### DIFF
--- a/cuda_bindings/cuda/bindings/_internal/utils.pyx
+++ b/cuda_bindings/cuda/bindings/_internal/utils.pyx
@@ -15,7 +15,7 @@ cdef extern from *:
     #elif defined(__GNUC__) || defined(__GNUG__)
         #define _COMPILER_VERSION ("GCC " __VERSION__)
     #elif defined(_MSC_VER)
-        #define _COMPILER_VERSION ("MSVC " _MSC_VER)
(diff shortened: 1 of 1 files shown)


cuda_bindings/cuda/bindings/_internal/utils.pyx
```diff
@@ -8,6 +8,28 @@ from libcpp.utility cimport move
...
+cdef extern from *:
+    """
+    #if defined(__clang__)
+        #define _COMPILER_VERSION ("Clang " __clang_version__)
+    #elif defined(__GNUC__) || defined(__GNUG__)
+        #define _COMPILER_VERSION ("GCC " __VERSION__)
+    #elif defined(_MSC_VER)
+        #define _COMPILER_VERSION ("MSVC " _MSC_VER)
+    #else
+        #define _COMPILER_VERSION ("Unknown Compiler")
```
cpcloud, line 20: Should this `#error` instead? Do we support other compilers?

mdboom: I don't think we explicitly support other compilers.  I was just thinking let's not break anyone who is using something else, since this isn't really a "feature" just a debugging tool for performance nerds.

PR #1593, opened 2026-02-10
refactor(pathfinder): replace info_summary_append fixture with Python logging

cpcloud:
## What

Replaces the custom `info_summary_append` pytest fixture with standard Python `logging`, writing test diagnostic info to a file uploaded as a GHA artifact instead of printing a terminal summary.

## Why

- The terminal INFO summary was verbose and noisy in CI output
- Grep-based verification of the summary (`grep '^INFO test_'` in `run-tests`) was brittle
- A file artifact is easier to inspect on-demand without cluttering logs

## Changes

- **`conftest.py`** -- `info_summary_append` fixture and `pytest_terminal_summary` hook replaced by two fixtures: a session-scoped `_info_summary_handler` (manages a `logging.FileHandler`) and a function-scoped `info_log` (yields a `LoggerAdapter` with the test node name). Log file is keyed off `CUDA_PATHFINDER_TEST_LOAD_NVIDIA_DYNAMIC_LIB_STRICTNESS` env var so both CI runs produce separate files.
- **3 test files** -- `info_summary_append(...)` calls become `info_log.info(...)`.
- **`ci/tools/run-tests`** -- Removed `tee` pipe and `grep` verification for pathfinder.
- **Both test-wheel workflows** -- Added `upload-artifact` step (pinned v6.0.0) after the last pathfinder test run with `if-no-files-found: error`.

### DIFF
--- a/cuda_pathfinder/tests/test_driver_lib_loading.py
+++ b/cuda_pathfinder/tests/test_driver_lib_loading.py
@@ -127,7 +127,7 @@ def test_load_lib_no_cache_does_not_dispatch_ctk_lib_to_driver_path(mocker):
 
 
 @pytest.mark.parametrize("libname", sorted(_DRIVER_ONLY_LIBNAMES))
-def test_real_load_driver_lib(info_summary_append, libname):
+def test_real_load_driver_lib(info_log, libname):
     """Load a real driver library in a dedicated subprocess.
 
     This complements the mock tests above: it exercises the actual OS
@@ -151,9 +151,9 @@ def test_real_load_driver_lib(info_summary_append, libname):
         skip_if_missing_libnvcudla_so(libname, timeout=timeout)
         if STRICTNESS == "all_must_work":
             raise_child_process_failed()
-        info_summary_append(f"Not found: {libname=!r}")
+        info_log.info(f"Not found: {libname=!r}")
     else:
         abs_path = payload.abs_path
         assert abs_path is not None
-        info_summary_append(f"abs_path={quote_for_shell(abs_path)}")
+        info_log.info(f"abs_path={quote_for_shell(abs_path)}")
         assert os.path.isfile(abs_path)
--- a/cuda_pathfinder/tests/test_find_bitcode_lib.py
+++ b/cuda_pathfinder/tests/test_find_bitcode_lib.py
@@ -60,17 +60,17 @@ def _located_bitcode_lib_asserts(located_bitcode_lib):
 
 @pytest.mark.usefixtures("clear_find_bitcode_lib_cache")
 @pytest.mark.parametrize("libname", SUPPORTED_BITCODE_LIBS)
-def test_locate_bitcode_lib(info_summary_append, libname):
+def test_locate_bitcode_lib(info_log, libname):
     try:
         located_lib = locate_bitcode_lib(libname)
         lib_path = find_bitcode_lib(libname)
     except BitcodeLibNotFoundError:
         if STRICTNESS == "all_must_work":
             raise
-        info_summary_append(f"{libname}: not found")
+        info_log.info(f"{libname}: not found")
         return
 
-    info_summary_append(f"{lib_path=!r}")
+    info_log.info(f"{lib_path=!r}")
     _located_bitcode_lib_asserts(located_lib)
     assert os.path.isfile(lib_path)
     assert lib_path == located_lib.abs_path
--- a/cuda_pathfinder/tests/test_find_static_lib.py
+++ b/cuda_pathfinder/tests/test_find_static_lib.py
@@ -57,17 +57,17 @@ def _located_static_lib_asserts(located_static_lib):
 
 @pytest.mark.usefixtures("clear_find_static_lib_cache")
 @pytest.mark.parametrize("libname", SUPPORTED_STATIC_LIBS)
-def test_locate_static_lib(info_summary_append, libname):
+def test_locate_static_lib(info_log, libname):
     try:
         located_lib = locate_static_lib(libname)
         lib_path = find_static_lib(libname)
     except StaticLibNotFoundError:
         if STRICTNESS == "all_must_work":
             raise
-        info_summary_append(f"{libname}: not found")
+        info_log.info(f"{libname}: not found")
         return
 
-    info_summary_append(f"abs_path={quote_for_shell(lib_path)}")
+    info_log.info(f"abs_path={quote_for_shell(lib_path)}")
     _located_static_lib_asserts(located_lib)
     assert os.path.isfile(lib_path)
     assert lib_path == located_lib.abs_path


ci/tools/run-tests
```diff
@@ -32,10 +32,7 @@ if [[ "${test_module}" == "pathfinder" ]]; then
   echo "Running pathfinder tests with " \
       "LD:${CUDA_PATHFINDER_TEST_LOAD_NVIDIA_DYNAMIC_LIB_STRICTNESS} " \
       "FH:${CUDA_PATHFINDER_TEST_FIND_NVIDIA_HEADERS_STRICTNESS}"
-  pytest -ra -s -v --durations=0 tests/ |& tee /tmp/pathfinder_test_log.txt
-  # Fail if no "INFO test_" lines are found; capture line count otherwise
-  line_count=$(grep '^INFO test_' /tmp/pathfinder_test_log.txt | wc -l)
```
rwgk: What I wanted to achieve here originally (in the early pathfinder days):

* fail if somehow this didn't run (e.g. empty lists):

https://github.com/NVIDIA/cuda-python/blob/ea45bbfbacd9583f3a9fc485b4f3b135fd581a1d/cuda_pathfinder/tests/test_load_nvidia_dynamic_lib.py#L92

* make it easy to harvest the `Number of "INFO test_ lines"` run simply by `grep`ping the CI logs.

With this PR that's lost, but pathfinder is more stable now, so that should be an OK loss.

The change delta of this PR is net positive (b/o the upload steps) and IIUC introduces a layer of indirection to get to the information?

I think I can get used to the INFO lines being written to `cuda_pathfinder/test-info-summary-default.log`, but could we just `cat ./test-info-summary-default.log` here? Then the information will be as easily accessible as before, and we don't even need the changes in the test-wheel-*.yml files. It'd be nice to add

```
echo "Number of \"INFO test_\" lines: $(wc -l test-info-summary-default.log)"
```

because then I could still do a simple `grep` over the CI logs as before, to quickly check if everything is as expected.

cuda_pathfinder/tests/conftest.py
```diff
@@ -2,30 +2,35 @@
...
 import pytest
 
+_LOGGER_NAME = "cuda_pathfinder.test_info"
 
-def pytest_configure(config):
-    config.custom_info = []
 
+@pytest.fixture(scope="session")
+def _info_summary_handler(request):
+    strictness = os.environ.get("CUDA_PATHFINDER_TEST_LOAD_NVIDIA_DYNAMIC_LIB_STRICTNESS", "default")
```
rwgk: I think it's better and simpler to hardwire the filename. The strictness level is only for controlling hard asserts. Also, maybe make it easy to know what the file is about even when it's pulled into another context, e.g. by adding in `pathfinder`:

```
log_path = request.config.rootpath / f"pathfinder-test-info-summary-log.txt"
```

Naming it `.txt` makes for the smoothest experience viewing, uploading, etc. Other extensions tend to confuse one tool or another.

cuda_pathfinder/tests/conftest.py
```diff
@@ -2,30 +2,35 @@
...
+    log_path = request.config.rootpath / f"test-info-summary-{strictness}.log"
```
rwgk: We want to be sure we're not accidentally using a file from a previous run (e.g. if there is a segfault but some follow-on logic still looks for that file). Cursor generated the code below. If we hard-wire the filename as suggested, we don't even need the `glob`.

```diff
diff --git a/cuda_pathfinder/tests/conftest.py b/cuda_pathfinder/tests/conftest.py
index 2a82cca0c..a10278f2d 100644
--- a/cuda_pathfinder/tests/conftest.py
+++ b/cuda_pathfinder/tests/conftest.py
@@ -7,9 +7,15 @@ import os

 import pytest

+_INFO_LOG_GLOB = "test-info-summary-*.log"
 _LOGGER_NAME = "cuda_pathfinder.test_info"


+def pytest_configure(config):
+    for log_path in config.rootpath.glob(_INFO_LOG_GLOB):
+        log_path.unlink(missing_ok=True)
+
+
 @pytest.fixture(scope="session")
 def _info_summary_handler(request):
     strictness = os.environ.get("CUDA_PATHFINDER_TEST_LOAD_NVIDIA_DYNAMIC_LIB_STRICTNESS", "default")
```

cuda_pathfinder/tests/conftest.py
```diff
@@ -2,30 +2,35 @@
...
+    handler = logging.FileHandler(log_path, mode="w")
+    handler.setFormatter(logging.Formatter("%(test_node)s: %(message)s"))
 
-def pytest_terminal_summary(terminalreporter, exitstatus, config):  # noqa: ARG001
-    if not config.getoption("verbose"):
```
rwgk, line 13: Hm, what about deleting only the `verbose` condition, but keeping the other two? Otherwise the file will be overwritten 99 times or so?

However, we should unconditionally remove the file when the `_info_summary_handler` fixture runs. (I've been really confused a few times in the past by stale output, correlating it wrongly to a more recent action.)

rwgk (commented):
I'm OK with it, although I want to note that breaking currently certainly coherent output into pieces has pitfalls. Definitely, we need to clean up at the beginning. We should also add the new log file to `.gitignore`.

cuda_pathfinder/tests/conftest.py
cpcloud: Is there ever a case where the pathfinder tests are run twice CI? Once with `see_what_works` and once with the other setting?

ci/tools/run-tests
cpcloud: I am actively trying to avoid spewing the log in CI. Every time I've had to look at CI logs for a pathfinder failure, I always spend 50% of that time scrolling through these logs.

cpcloud: Happy to print the number of lines.

cuda_pathfinder/tests/conftest.py
cpcloud: What's getting overwritten 99 times?

cuda_pathfinder/tests/conftest.py
cpcloud: CI runs pathfinder tests twice per job -- once with `see_what_works` (`test-wheel-linux.yml:250`, `test-wheel-windows.yml:223`) and once with `all_must_work` (`test-wheel-linux.yml:295`, `test-wheel-windows.yml:272`). The strictness level is embedded in the filename so the second run does not silently overwrite the first run's log. `pytest_configure` only deletes the file matching the current strictness level, so both logs coexist for artifact upload.

cuda_pathfinder/tests/conftest.py
cpcloud: See https://github.com/NVIDIA/cuda-python/pull/1593#discussion_r2798557243, it's also related to your comment here.

cuda_pathfinder/tests/conftest.py
cpcloud: The `_info_summary_handler` fixture is `scope="session"`, so it's created exactly once per pytest invocation regardless of how many times `pytest-repeat` re-runs the tests. The `FileHandler` opens the file once with `mode="w"` at session start and appends to it for the duration -- repeated tests just add more lines to the same open handle.

ci/tools/run-tests
cpcloud: Done, now printing the number of lines.

.github/workflows/test-wheel-linux.yml
```diff
@@ -297,3 +297,11 @@ jobs:
           CUDA_PATHFINDER_TEST_LOAD_NVIDIA_DYNAMIC_LIB_STRICTNESS: all_must_work
           CUDA_PATHFINDER_TEST_FIND_NVIDIA_HEADERS_STRICTNESS: all_must_work
         run: run-tests pathfinder
+
+      - name: Upload cuda.pathfinder test info summary
```
rwgk, line 302: I posted #1621.

Introducing these artifacts makes something simple weirdly complex. The INFO lines are a tiny fraction of the total log (see PR #1621 description for numbers).

cuda_pathfinder/tests/conftest.py
rwgk: Hm, that's basically spamming the filesystem.

.github/workflows/test-wheel-linux.yml
cpcloud: It's still too noisy, and the mechanism that we are using to track this information lacks any sort of real structure. It's very ad-hoc and can't really ever be integrated into or compose well with another system.

rwgk: Can you point to a concrete, measurable problem (similar to the numbers I showed under #1621)?

> The fraction of `INFO` lines in a typical CI log file is around 2.5%

I'm using those `INFO` lines regularly when working on pathfinder, or if random issues pop up. Hiding that information in disconnected artifacts would force me to puzzle together information from pieces that was perviously available at a glance.

cpcloud: It's not a problem in CI anymore it seems. It's still a big code smell though. Is that measurable? Not exactly. Does it have to be to be a valid concern? Definitely not. Closing the PR.

Issue #1038, opened 2025-09-26
[FEA]: Explore pathfinder "scoped search" feature


rwgk:
Problem to solve
-----------------

The searches implemented in `load_nvidia_dynamic_lib()` and `find_nvidia_header_directory()` are completely independent from each other. For example, `load_nvidia_dynamic_lib("nvrtc")` might resolve to a site-packages path, while `load_nvidia_dynamic_lib("nvJitLink")` finds a Conda path, and `find_nvidia_header_directory("nvvm")` locates headers under `/usr/local/cuda`. This can lead to surprising inconsistencies and confusing behavior deep in the call stack.

Potential approach (with low maintenance as a goal in mind)
------------------------------------------------------------

Core idea: Introduce the concept of a **pivot library** that defines the scope for all subsequent searches.

Here, **package system** means, for example:
* `/usr/local/cuda` (or another "standard" CTK installation)
* Conda (including pixi)
* site-packages (wheels, PyPI)

How it would work in practice:

* A client code author decides which library is most pivotal and loads it first.

* `search_scope = cuda.pathfinder.determine_search_scope(pivot_libname="nvrtc")` — Loads the pivot library and returns it in a wrapper object.

* `loaded_dl = search_scope.load_nvidia_dynamic_lib("nvJitLink")` — Only succeeds if the target library exists in the same package system, otherwise fails with a helpful error.

* `hdr_dir = search_scope.find_nvidia_header_directory("nvvm")` — Similarly constrained to the same package system.

This design shifts responsibility for consistency to package managers, which already provide version alignment via pinning. We avoid duplicating those mechanisms and instead ensure our tooling surfaces clear, consistent results.

rwgk:
xref to https://github.com/NVIDIA/cuda-python/pull/1090#discussion_r2421147082

Comment by @kkraus14 there:

> If one library has a binary dependency on another library, I don't think compatibility guarantees are made outside of driver version. I.e. `lib2` depends on `lib1`, the only guarantee we currently have is that it works if both `lib2` and `lib1` are the same patch version. So from the perspective of `test_load_nvidia_dynamic_libs.py` we could end up in an environment with different patch versions and someone blows up at load time because we load an incompatible dependency (since the dependencies aren't captured in python package dependencies to support using a system CTK).

Cross-referencing here since this might be something we could guard against (e.g. helpful error message) as part of the scoped search feature.

rwgk:
Tracking prompt and response:

___

[...]

PR #1595, opened 2026-02-10
feat(pathfinder): add CTK root canary probe for non-standard-path libs

cpcloud:
## Problem

`libnvvm.so` lives under `$CTK_ROOT/nvvm/lib64/` (or `nvvm/bin` on Windows), which is not on the default loader path. On bare system CTK installs, `dlopen("libnvvm.so.4")` can fail when `CUDA_HOME`/`CUDA_PATH` is unset even though nvvm is installed.

## Solution

Keep the CTK-root canary strategy, but run the canary load in a subprocess:

1. Spawn `python -m cuda.pathfinder._dynamic_libs.canary_probe_subprocess cudart`
2. Child does `load_with_system_search("cudart")` and prints a JSON payload with the resolved absolute path (or `null`)
3. Parent derives CTK root from that path and searches for the target lib relative to that root

This avoids polluting loader state in the caller process while preserving the existing fallback behavior.

## Why JSON for child-parent payload?

- Encodes both outcomes explicitly: path string or `null`
- Avoids brittle ad-hoc stdout parsing and escaping pitfalls (especially for Windows paths)
- Gives a stable wire format that can be extended later without changing parsing rules

## Search order

`site-packages -> conda -> already-loaded -> system search -> CUDA_HOME -> subprocess canary probe`

The canary still runs only after `CUDA_HOME` to keep existing precedence.

## Tests

- Updated `tests/test_ctk_root_discovery.py` to mock subprocess canary resolution
- Added assertion that parent `load_with_system_search()` is not called during canary probing
- Updated ordering tests so `CUDA_HOME` still wins over canary
- Verified with `pixi run test-pathfinder` (`129 passed, 1 skipped`)

### DIFF
--- a/cuda_pathfinder/tests/test_ctk_root_discovery.py
+++ b/cuda_pathfinder/tests/test_ctk_root_discovery.py
@@ -90,6 +90,17 @@ def test_derive_ctk_root_linux_root_level():
     assert _derive_ctk_root_linux("/lib64/libcudart.so.13") == "/"
 
 
+def test_derive_ctk_root_linux_targets_lib64():
+    assert (
+        _derive_ctk_root_linux("/usr/local/cuda-13.1/targets/x86_64-linux/lib64/libcudart.so.13")
+        == "/usr/local/cuda-13.1"
+    )
+
+
+def test_derive_ctk_root_linux_targets_lib():
+    assert _derive_ctk_root_linux("/opt/cuda/targets/sbsa-linux/lib/libcudart.so.12") == "/opt/cuda"
+
+
 def test_derive_ctk_root_windows_ctk13():
     path = r"C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v13.0\bin\x64\cudart64_13.dll"
     assert _derive_ctk_root_windows(path) == r"C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v13.0"
--- a/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/find_nvidia_dynamic_lib.py
+++ b/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/find_nvidia_dynamic_lib.py
@@ -157,13 +157,20 @@ def _derive_ctk_root_linux(resolved_lib_path: str) -> str | None:
 
     Standard system CTK layout: ``$CTK_ROOT/lib64/libfoo.so.XX``
     (some installs use ``lib`` instead of ``lib64``).
+    Also handles target-specific layouts:
+    ``$CTK_ROOT/targets/<triple>/lib64/libfoo.so.XX`` (or ``lib``).
 
     Returns None if the path doesn't match a recognized layout.
     """
     lib_dir = os.path.dirname(resolved_lib_path)
     basename = os.path.basename(lib_dir)
     if basename in ("lib64", "lib"):
-        return os.path.dirname(lib_dir)
+        parent = os.path.dirname(lib_dir)
+        grandparent = os.path.dirname(parent)
+        if os.path.basename(grandparent) == "targets":
+            # This corresponds to /.../targets/<triple>/lib{,64}
+            return os.path.dirname(grandparent)
+        return parent
     return None
 
 


cuda_pathfinder/tests/test_ctk_root_discovery.py
```diff
@@ -0,0 +1,245 @@
...
+def test_derive_ctk_root_linux_unrecognized():
+    assert _derive_ctk_root_linux("/some/weird/path/libcudart.so.13") is None
+
+
+def test_derive_ctk_root_linux_root_level():
+    assert _derive_ctk_root_linux("/lib64/libcudart.so.13") == "/"
+
+
+def test_derive_ctk_root_windows_ctk13():
+    path = r"C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v13.0\bin\x64\cudart64_13.dll"
```
cpcloud, line 105: This works cross-platform due to explicit use of `ntpath` in `_derive_ctk_root_windows`. Given that the code won't look much different using the platform specific version versus not, it seems somewhat useful to have these around instead of having to skip a bunch of tests based on platform.

cpcloud:
I think if there's opposition to this approach, which I believe we all discussed on the same call, then if we still care about solving the problem (I think we do), then I'd kindly request a counter-PR implementing an alternative proposal. Otherwise, we're going to get bogged down in reviewing waiting for the perfect solution.

AFAICT, there is no perfect way to solve this problem: it's _just_ picking the least worst option.

cpcloud:
The subprocess approach is not my favorite, but if we're prioritizing isolation, I don't see a less complex option for getting the canary search path.

cpcloud:
One argument in favor of this approach is that the search is a last ditch effort after everything has failed, and it's backward compatible.

You might argue that canary searches are a form of system search, but I actually kept the existing priority in order to reduce the foot-bazooka of this whole thing by keeping existing installations behaving exactly the same.

rwgk:
> The subprocess approach is not my favorite, but if we're prioritizing isolation, I don't see a less complex option for getting the canary search path.

I think the isolation is important. If we use caching of the `abs_path`, the runtime hit will only be once per process.

I agree with everything else you wrote above.

cpcloud:
I added the subprocess isolation.

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py
```diff
@@ -22,6 +27,73 @@
...
+    child process instead of polluting the current process.
+    """
+    cmd = [
+        sys.executable,
+        "-m",
+        "cuda.pathfinder._dynamic_libs.canary_probe_subprocess",
+        libname,
+    ]
+    try:
+        result = subprocess.run(  # noqa: S603
```
rwgk: While working on the initial version of pathfinder, using `subprocess` was frowned upon, for general security reasons I think, so I implemented cuda_pathfinder/tests/spawned_process_runner.py. I was hoping you'd use that here.

The rest looks good to me, although I still need to spend some time carefully looking at the tests.

rwgk: spawned_process_runner.py came from here (from before we had Cursor):

https://chatgpt.com/share/681914ce-f274-8008-9e9f-4538716b4ed7

There I started with:

> I need to run Python code in a separate process, but I cannot use suprocess.run because that uses fork, which is incompatible with my requirements.

So I think spawned_process_runner.py will give us better isolation (and predictability), it's not just the secops concern.

cpcloud: I'm not sure what you mean. Can you make a patch suggestion instead of the indirection of a link to code that I then have to bring into the PR?

cpcloud: The goal is isolation, why doesn't subprocess run achieve that goal? The chat history is quite long, so if you want to summarize that here that'd be nice.

cpcloud: Also, I feel like we don't need the isolation in this PR, I will happily follow up with whatever approach makes you comfortable. Let's not get hung up on this detail. It's important, but I don't think it affects the overall design and approach.

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py
```diff
@@ -50,11 +122,21 @@ def _load_lib_no_cache(libname: str) -> LoadedDL:
...
+            # Canary probe: if the direct system search and CUDA_HOME both
+            # failed (e.g. nvvm isn't on the linker path and CUDA_HOME is
+            # unset), try to discover the CTK root by loading a well-known CTK
+            # lib in a subprocess, then look for the target lib relative to
+            # that root.
+            abs_path = _try_ctk_root_canary(finder)
+            if abs_path is not None:
+                found_via = "system-ctk-root"
+            else:
+                finder.raise_not_found_error()
```
leofang: Sorry I have been trapped in other fires and was unable to provide feedbacks timely 😢 

Loading `cudart` in a subprocess is safer than loading it in the main process. That said, my recollection from the Feb-05 meeting was that we'd use other anchor points such as nvJitLink (basically, #1038). Does this mean we changed our mind a bit and decided to use cudart instead?

cpcloud: nvjitlink is fine to start with as well. I'm honestly not sure why one would be preferable over another. That said, as long as we're not choosing something super niche, it doesn't seem like it's worth spending too much time on and can be changed in a follow-up.

leofang: Thanks, Phillip. I do not have any objection.

rwgk: > my recollection from the Feb-05 meeting was that we'd use other anchor points such as nvJitLink (basically, https://github.com/NVIDIA/cuda-python/issues/1038). Does this mean we changed our mind a bit and decided to use cudart instead?

The main idea behind 1038 is that the pivot library has two roles: 1. provide an anchor point to find other libraries from (nvvm is the only case that needs it I think), 2. more importantly actually, limit the scope of future searches (via an object that remembers the pivot library).

This PR doesn't have the concept of a search scope, it's essentially only a trick to find nvvm independently, like any other independently found library.

I was a bit surprised when Phillip sent this PR, but I think it's useful, because I believe that we'll have the independent and scoped searches side by side indefinitely. This PR will make 1038 less important, mainly for safety / internal consistency, and to guide users to a consistent setup via helpful error messages.

Re cudart vs nvjitlink:

For the scoped search, the choice of the pivot library is up to the user.

For the independent search (in an isolated process), cudart is the better choice, because it's much smaller than nvjitlink:

740 vs 96964 KiB

```
smc120-0009.ipp2a2.colossus.nvidia.com:/usr/local/cuda-13.1/lib64 $ (for so in *.so; do ls -s $(realpath $so); done) | sort -n
32 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libOpenCL.so.1.0.0
40 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libnvtx3interop.so.1.1.0
48 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libcufile_rdma.so.1.16.1
400 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libcuobjclient.so.1.0.0
724 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libnvblas.so.13.2.1.1
740 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libcudart.so.13.1.80
972 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libcufftw.so.12.1.0.78
1636 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libnppc.so.13.0.3.3
1640 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libnppisu.so.13.0.3.3
2424 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libnvfatbin.so.13.1.115
3380 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libcufile.so.1.16.1
4236 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libnppitc.so.13.0.3.3
4284 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libnvrtc-builtins.so.13.1.115
5784 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libnvjpeg.so.13.0.3.75
6592 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libnppim.so.13.0.3.3
6836 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libnppicc.so.13.0.3.3
8456 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libnppidei.so.13.0.3.3
10060 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libnpps.so.13.0.3.3
13112 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libnppial.so.13.0.3.3
25848 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libnppig.so.13.0.3.3
26332 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libnppist.so.13.0.3.3
52924 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libcublas.so.13.2.1.1
58192 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libnppif.so.13.0.3.3
96964 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libnvJitLink.so.13.1.115
101472 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libcusolverMg.so.12.0.9.81
111500 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libnvrtc.so.13.1.115
129928 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libcurand.so.10.4.1.81
139108 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libcusolver.so.12.0.9.81
165772 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libcusparse.so.12.7.3.1
292796 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libcufft.so.12.1.0.78
490832 /usr/local/cuda-13.1/targets/x86_64-linux/lib/libcublasLt.so.13.2.1.1
```

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py
rwgk: I believe the isolation is 1. crucial, and 2. cannot be implemented via fork.

> The chat history is quite long

The most important aspect is right at the top, the original prompt (two trivial original typos corrected here):

> I need to run Python code in a separate process, but I cannot use subprocess.run because that uses fork, which is incompatible with my requirements.

The key point: What if cudart was loaded already?

A forked process will still have that in memory. You'll get a **guaranteed conflict-free** `found_via` `system-search` only in a spawned process.

I've been scratching my head the past few days, because TBH it wasn't immediately obvious to me that your `canary_probe_subprocess.py` is kind of genius (it's coming from a subtly different angle than my implementation last year). You're directly forcing a system search. Basically you're saying: I know I've exhausted all other options, I didn't find nvvm, now I want a fully independent system search, because that's the only remaining option I still have to find nvvm, and I'm **not allowing any other way** of finding the canary lib.

Two suggestions:

* Rename to `canary_probe_process.py` (i.e. drop the `sub`) to not suggest a misleading connection to the `subprocess` module.
* Use `spawned_process_runner` to achieve full isolation.

A third suggestion would be: set up a real environment in which the search only works with the new CTK root canary probe feature, but that's probably at least half a day of battling with our CI scripts. I'm OK to leave that for later and to purely rely on mocks for now.

cpcloud:
@leofang @rwgk Other than the conflicts, is this ready to go?

rwgk:
> @leofang @rwgk Other than the conflicts, is this ready to go?

Almost, but I wouldn't want to release this without switching from `fork` to `spawn`:

https://github.com/NVIDIA/cuda-python/pull/1595#discussion_r2820328394

cpcloud:
I don't think we should bolt on any notion of scope yet. I have a refactor that we should get in (hopefully) before adding anything like that. It adds an additional concept to reason about as a user and finding nvidia libraries is already a fraught landscape full of landmines.

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py
cpcloud: Implemented in c3130d10c.

I switched the canary probe from `subprocess.run(...)` to `run_in_spawned_child_process(...)`, which uses `multiprocessing.get_context(\"spawn\")` for full process isolation (no inherited loaded-lib state from forked children).

I also moved the spawned-process runner into `cuda.pathfinder._utils` and made the test helper re-export that shared implementation so we only have one runner.

Validation: `pixi run pytest tests/test_ctk_root_discovery.py tests/test_spawned_process_runner.py` (22 passed).

cpcloud:
@rwgk @leofang This PR is ready to go.

rwgk:
Hi @cpcloud , I worked pretty intensely with Cursor (GPT-5.3 Codex High) to add eight commits. That's much easier and more efficient than the traditional back-and-forth via PR comments. Could you please review from your side?

I'll update the branch and trigger the CI. Assuming CI passes, this PR looks merge-ready to me.

rwgk:
Oof ... I think this doesn't actually work:

```
(PathfinderVenv) smc120-0009.ipp2a2.colossus.nvidia.com:/wrk/forked/cuda-python $ python -c 'from cuda import pathfinder; print(pathfinder.__version__); print(pathfinder.load_n
vidia_dynamic_lib("cudart"))'
1.3.6.dev20+g47011d71b
LoadedDL(abs_path='/usr/local/cuda-13.1/targets/x86_64-linux/lib/libcudart.so.13', was_already_loaded_from_elsewhere=False, _handle_uint=782303744, found_via='system-search')

(PathfinderVenv) smc120-0009.ipp2a2.colossus.nvidia.com:/wrk/forked/cuda-python $ python -c 'from cuda import pathfinder; print(pathfinder.__version__); print(pathfinder.load_n
vidia_dynamic_lib("nvvm"))'
1.3.6.dev20+g47011d71b
  File "/wrk/forked/cuda-python/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py", line 283, in load_nvidia_dynamic_lib
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/wrk/forked/cuda-python/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py", line 179, in _load_lib_no_cache
    finder.raise_not_found_error()
  File "/wrk/forked/cuda-python/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/find_nvidia_dynamic_lib.py", line 270, in raise_not_found_error
cuda.pathfinder._dynamic_libs.load_dl_common.DynamicLibNotFoundError: Failure finding "libnvvm.so": No such file: libnvvm.so*, No such file: libnvvm.so*

(PathfinderVenv) smc120-0009.ipp2a2.colossus.nvidia.com:/wrk/forked/cuda-python $ realpath /usr/local/cuda-13.1/nvvm/lib64/libnvvm.so
/usr/local/cuda-13.1/nvvm/lib64/libnvvm.so.4.0.0
```

Maybe the "walking up" part isn't working as intended? I'm looking into it.

rwgk:
I pushed a simple fix (92bf6c9). This works now:

```
(PathfinderVenv) smc120-0009.ipp2a2.colossus.nvidia.com:/wrk/forked/cuda-python $ python -c 'from cuda import pathfinder; print(pathfinder.__version__); print(pathfinder.load_n
vidia_dynamic_lib("nvvm"))'
1.3.6.dev20+g47011d71b.d20260224
LoadedDL(abs_path='/usr/local/cuda-13.1/nvvm/lib64/libnvvm.so', was_already_loaded_from_elsewhere=False, _handle_uint=50529072, found_via='system-ctk-root')
```

I think it's fully sufficient as a solution, but I want to think a bit how we can have at least a minimal real tests.

cpcloud:
> Hi @cpcloud , I worked pretty intensely with Cursor (GPT-5.3 Codex High) to add eight commits. That's much easier and more efficient than the traditional back-and-forth via PR comments. Could you please review from your side?
> 
> I'll update the branch and trigger the CI. Assuming CI passes, this PR looks merge-ready to me.

It's not actually easier. It makes the PR more difficult to review. There's no motivation describing what was wrong that motivated the additional commits before making the commits. So, we still have to go back and forth because I'm going to ask why you made a given change.

There's one fix that I can see, and seven commits that cna easily be done as follow-ups.

Can we please merge this? It doesn't need to be perfect, and if you want to fix things up afterward, you're free to do that.

cpcloud:
You can use cursor to review PRs instead of pushing commits. At least then I can understand its rational before taking any action, and we can discuss.

cpcloud:
Personally I really dislike people pushing to my PRs with zero discussion about what they are committing.

rwgk:
I will not push to your PRs in the future, sorry.

I was careful about leaving meaningful commit messages; a couple are hand-edited. (I'm generally prompting Cursor to not generate long-ish commit messages, to not cross into AI slop territory.)

Sometimes it takes me a few iterations of playing with the code to really "see" what it's doing. E.g. I didn't immediately see that the code changed with commit ef3a763 was prone to masking bugs. I wouldn't want to merge code with that behavior.

The rest of the commits are more of the basic housekeeping / polishing kind, except the last two:
* 60861ec: the relatively complex new feature was hit for any libname, even though we're sure that all supported libnames except `nvvm` live in the usual location.
* 92bf6c9: the most important use case (Linux, CUDA_HOME/CUDA_PATH not set) didn't actually work.

The exclusive reliance on mocks is a big worry point for me. Going to the other extreme, no mocks, is unfortunately not very practical, but we have to find a sensible middle ground, to maintain a reasonable level of code quality. I'm OK merging without any real tests if you prefer (I'll approve and retrigger the CI), but usually such things slip, and ultimately erode the quality of the code base.
 
Cursor is helping greatly in turning guesses / vague suggestions into testable, proven changes in minutes, or weed out ideas that are simply going in the wrong direction (e.g. I was trying to avoid the json encoding/decoding, but Cursor convinced me it's best to keep). An alternative way to leverage the power of the LLMs would be to work on a parallel branch, from which you can cherry-pick what would otherwise be copy-pasted suggestions.

cpcloud:
>  I was trying to avoid the json encoding/decoding, but Cursor convinced me it's best to keep

This is a good example of wasted effort that we could have resolved by discussing, because I was skeptical of that as well.

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/supported_nvidia_libs.py
```diff
@@ -358,6 +358,13 @@
 
 LIBNAMES_REQUIRING_RTLD_DEEPBIND = ("cufftMp",)
 
+# CTK root canary probe config:
+# - anchor libs: expected on the standard system loader path and used to derive
+#   CTK root in an isolated child process.
+# - discoverable libs: libs that are allowed to use the CTK-root canary fallback.
+_CTK_ROOT_CANARY_ANCHOR_LIBNAMES = ("cudart",)
+_CTK_ROOT_CANARY_DISCOVERABLE_LIBNAMES = ("nvvm",)
```
cpcloud, line 366: What is the motivation behind limiting the discovery to `nvvm`? It's unnecessarily restrictive for a use case that we don't have solid information about. What's the worst that could happen here? Nothing is discoverable from any other path, and so everything is discovered using the canary? That seems like the **exact** use case we want to support: someone doesn't have any of the traditional installs, so we discover everything from the canary anchor.

From the commit message, it just looks like this change was made to satisfy a testing constraint, which artificially kneecaps the functionality.

rwgk: > What is the motivation behind limiting the discovery to nvvm?

All the other supported libnames are colocated with cudart. If they are not found when we reach `_try_ctk_root_canary`, cudart will not be found, too, or the layout is completely unexpected and it wouldn't be meaningful to proceed.

> From the commit message, it just looks like this change was made to satisfy a testing constraint,

Copying the commit message here for easy reference:
> This avoids unnecessary canary subprocess work for other libraries and locks the behavior with a focused regression test.

Sorry I decided to not edit the second part, although I noticed that it's worded in an unfortunate way.

The first part is the correct explanation. (There is no kneecapping; the feature is meaningful only for nvvm.)

A better way to write this would have been:

This avoids unnecessary canary subprocess work for other libraries. The new behavior is locked with a focused regression test.

This PR looks good to me as is, it gets rid of a very annoying stumbling block that was probably the main motivation behind the nvvm-related work under https://github.com/NVIDIA/numba-cuda/pull/308. My only worry is that we don't have real test coverage, i.e. we'll notice some regressions only through manual testing, or bug reports. Unfortunately, to fix that we'd have to go into the CI scripts. A separate PR for that sounds good.

rwgk:
> > I was trying to avoid the json encoding/decoding, but Cursor convinced me it's best to keep
> 
> This is a good example of wasted effort that we could have resolved by discussing, because I was skeptical of that as well.

On my side, the effort was worth about 5 minutes, asking Cursor twice about it (at the very beginning, and then again after commit ef3a763 was done).

I just asked Cursor to clearly lay out our options. Below is what it gave me. I think "keep what we have" (Option 1) is most appropriate for our purposes, even long term, but Option 3 sounds good, too, although I believe at a very low priority, compared to many other things we could do with our time.

___

## Child-parent payload options

### Option 1: keep JSON (current PR)

- Transport stays `stdout` from child to parent.
- Child prints JSON payload:
  - found: `"/abs/path/..."`
  - not found: `null`
- Parent parses the final non-empty line with `json.loads(...)`.

Pros:
- Explicit and unambiguous for `str | None`.
- Robust if Windows paths or escaping ever get tricky.
- Easy to extend later (add fields) without redesigning framing.

Cons:
- Slight encode/decode overhead.
- A bit more ceremony for a tiny payload.


### Option 2: plain text sentinel (custom mini-protocol)

- Transport stays `stdout`, but no JSON.
- Child prints one final line in a custom format, e.g.:
  - not found: `NONE`
  - found: `PATH:/abs/path/...`
- Parent parses by prefix/sentinel:
  - `NONE` -> `None`
  - `PATH:` -> path suffix
  - anything else -> error

Pros:
- Very small and simple implementation.
- No JSON dependency/parsing.

Cons:
- Custom protocol to maintain.
- More brittle for future evolution than JSON.
- Any payload growth usually requires protocol redesign.

Safer Option-2 variant:
- Use short tagged records like:
  - `N`
  - `P\t<path>`
- This reduces ambiguity vs free-form sentinel strings.


### Option 3: return Python value directly from spawned runner

- Child function returns `str | None` via process IPC/queue.
- No `print(...)`, no `stdout` parsing, no JSON/sentinel format.
- Parent receives typed value directly.

Pros:
- Cleanest interface.
- Strongest type-level clarity.
- Avoids all wire-format parsing concerns.

Cons:
- Requires a small refactor of spawned runner contract.
- Slightly larger change surface than Options 1/2.
- Return values must be IPC-serializable (effectively pickleable with the
  current multiprocessing/queue approach), which is fine for `str | None` but
  can be a real constraint for more general payloads.


## Current status

- Option 1 is what the PR currently implements.
- Option 3 is likely a small-ish follow-up refactor if we want the cleanest long-term shape.

PR #1597, opened 2026-02-11
Add Device context manager for temporary device switching

### LINKED ISSUE #1586: Feature Request: Device context manager for temporary device switching
## Problem

There is no context manager for temporarily switching the active CUDA device in `cuda.core`. Users who need to perform work on a specific device and then restore the previous device must manually manage this state:

```python
from cuda.core import Device

# Save current state, switch, do work, restore
dev0 = Device(0)
dev0.set_current()
# ... do work on device 0 ...

dev1 = Device(1)
dev1.set_current()
# ... do work on device 1 ...

dev0.set_current()  # manually restore -- easy to forget, not exception-safe
```

This pattern is error-prone (the restore call is not exception-safe) and verbose compared to the idiomatic Python `with` statement. The pattern appears in real-world code such as [dask-cuda](https://github.com/rapidsai/dask-cuda/pull/1625/changes), which implemented a workaround for this missing feature.

## Proposed Design

Add `__enter__` and `__exit__` methods to `Device`, making it usable as a context manager that temporarily activates the device's primary context and restores the previous state on exit.

### API

```python
from cuda.core import Device

dev0 = Device(0)
dev0.set_current()
# ... do work on device 0 ...

with Device(1) as device:
    # device 1 is now current
    # ...

# device 0 is automatically restored here
```

### Semantics

On `__enter__`:
1. Query the current CUDA context via `cuCtxGetCurrent` and save it on the context manager instance.
2. Call `self.set_current()` (which uses the primary context for this device via `cuCtxSetCurrent`).
3. Return `self`.

On `__exit__`:
1. Restore the saved context via `cuCtxSetCurrent`. If the saved context was `NULL`, set `NULL` (no active context).
2. Do NOT suppress exceptions (return `False`).

### Key design properties

**Stateless restoration (no Python-side stack).** Each `__enter__` call queries the actual CUDA driver state for the current context rather than maintaining a Python-side stack. On `__exit__`, it restores exactly what was saved. This is the critical lesson from CuPy's experience ([cupy/cupy#6965](https://github.com/cupy/cupy/issues/6965), [cupy/cupy#7427](https://github.com/cupy/cupy/pull/7427)): libraries that maintain their own stack of previous devices break interoperability with libraries that use the CUDA API directly to check device state. By always querying and restoring the driver-level state, we interoperate correctly with PyTorch, CuPy, and any other library that uses `cudaGetDevice`/`cudaSetDevice` or `cuCtxGetCurrent`/`cuCtxSetCurrent`.

**Reentrant and reusable.** Because `Device` is a thread-local singleton and the saved-context state is stored per-`__enter__` invocation (not on the `Device` object itself), the context manager is both reusable and reentrant:

```python
dev0 = Device(0)
dev1 = Device(1)

with dev0:
    with dev1:
        with dev0:  # reentrant -- works correctly
            ...
        # dev1 restored
    # dev0 restored
# original state restored
```

To achieve reentrancy, the saved context must NOT be stored on `self` (the `Device` singleton). Instead, use a thread-local stack or return a helper object from `__enter__` that holds the saved state. The simplest correct approach: store a per-thread stack of saved contexts on the `Device` class (or module-level), pushing on `__enter__` and popping on `__exit__`.

**Implementation sketch** (in `_device.pyx`):

```python
# Module-level: add a per-thread stack for saved contexts
# (reuse existing _tls threading.local())

def __enter__(self):
    # Query actual CUDA state -- do NOT use a Python-side device cache
    prev_ctx = handle_return(cuCtxGetCurrent())
    # Store on a per-thread stack so nested `with` works
    if not hasattr(_tls, '_ctx_stack'):
        _tls._ctx_stack = []
    _tls._ctx_stack.append(prev_ctx)
    self.set_current()
    return self

def __exit__(self, exc_type, exc_val, exc_tb):
    prev_ctx = _tls._ctx_stack.pop()
    handle_return(cuCtxSetCurrent(prev_ctx))
    return False
```

Note: The stack here is NOT a device stack -- it is a stack of saved `CUcontext` values that the `__exit__` restores. Each entry corresponds to exactly one `__enter__` call. This is fundamentally different from CuPy's old broken approach which tracked a stack of device IDs and queried that stack instead of the CUDA API.

**Interoperability with other libraries.** Because we use `cuCtxSetCurrent` (driver API), and both PyTorch and CuPy use `cudaGetDevice`/`cudaSetDevice` (runtime API) which queries the same underlying driver state, cross-library nesting works:

```python
with torch.cuda.device(1):
    with Device(2):
        # Both torch and cuda.core see device 2
        ...
    # torch sees device 1 again (cuda.core restored the context)
```

Note that correct cross-library nesting depends on each library querying the CUDA API for the current device on context exit rather than relying on a cached value. Libraries that follow this pattern (including CuPy v12+ and the CUDA runtime API) will interoperate correctly.

### 1. Separate `Device.activate()` method returning a context manager

```python
with Device(1).activate():
    ...
```

This avoids adding `__enter__`/`__exit__` to the singleton `Device` object. However, it adds API surface for no practical benefit -- the saved context state can be stored on a thread-local stack rather than on the `Device` instance, making `Device` itself safe to use directly as a reentrant context manager. The `with Device(1):` syntax is also more natural and matches PyTorch's `with torch.cuda.device(1):` pattern.

**Rejected** because it adds unnecessary indirection.

### 2. Do nothing -- recommend `set_current()` only

Per CuPy's internal policy, context managers for device switching are banned in CuPy's own codebase because they are footguns for library developers. The argument is that `set_current()` is explicit and unambiguous.

However, `cuda.core` targets end users (not just library internals), and the context manager pattern is:
- Exception-safe by default
- Idiomatic Python
- Already provided by PyTorch and CuPy (for end users)
- Requested by downstream users (dask-cuda)

**Rejected** as the sole approach, but `set_current()` remains the recommended approach for library code that needs precise control.

### 3. Use `cuCtxPushCurrent`/`cuCtxPopCurrent` instead of `cuCtxSetCurrent`

The CUDA driver provides an explicit context stack via push/pop. Using this would make nesting trivially correct. However, `Device.set_current()` currently uses `cuCtxSetCurrent` for primary contexts (not push/pop), and mixing the two models is fragile. The push/pop model also does not interoperate with libraries using `cudaSetDevice` (runtime API). The current approach of save-via-query/restore-via-set is correct and interoperable.

**Rejected** because it would diverge from the runtime API model that other libraries use.

## Open Questions

1. **Should `__enter__` call `set_current()` even if this device is already current?** Calling `cuCtxSetCurrent` with the already-current context is cheap (no-op at the driver level) and keeps the implementation simple. The alternative (check-and-skip) adds complexity for negligible performance gain. Recommendation: always call `set_current()`.

2. **What should `__enter__` do if `set_current()` has never been called on this device?** Currently, many `Device` properties require `set_current()` to have been called first (`_check_context_initialized`). The context manager should unconditionally call `set_current()`, initializing the device if needed. This is the natural expectation: `with Device(1):` should make device 1 ready for use.

3. **Should we document cross-library interop expectations?** We should document that `with Device(N):` works correctly for `cuda.core` code, and that cross-library nesting works as long as the other library's context manager correctly queries CUDA state on exit rather than relying on a cached value.

## Test Plan

- Basic usage: `with Device(0):` sets device 0 as current, restores on exit.
- Exception safety: device is restored even when an exception is raised inside the `with` block.
- Nesting (same device): `with dev0: with dev0:` works without error.
- Nesting (different devices): `with dev0: with dev1:` correctly restores dev0 on exit of inner block.
- Deep nesting / reentrancy: `with dev0: with dev1: with dev0: with dev1:` restores correctly at each level.
- `Device` remains usable after context manager exit (singleton not corrupted).
- Multi-GPU: requires 2+ GPUs. Verify `cudaGetDevice()` (runtime API) reflects the device set by the context manager.
- Thread safety: context manager state is per-thread (uses thread-local storage), so concurrent threads using different devices should not interfere.

Andy-Jost:
## Summary

Closes #1586. Adds `__enter__`/`__exit__` to `Device` so it can be used as a context manager that temporarily activates a device and restores the previous CUDA context on exit.

```python
from cuda.core import Device

dev0 = Device(0)
dev0.set_current()
# ... do work on device 0 ...

with Device(1) as dev1:
    # device 1 is now current
    buf = dev1.allocate(1024)

# device 0 is automatically restored here
```

## Changes

- **`cuda/core/_device.pyx`**: Added `__enter__` and `__exit__` methods to `Device`. On enter, queries the current context via `cuCtxGetCurrent` and saves it on a per-thread stack (`_tls._ctx_stack`), then calls `set_current()`. On exit, restores the saved context via `cuCtxSetCurrent`. Uses peek-then-pop ordering so the stack is not corrupted if `cuCtxSetCurrent` raises.
- **`tests/test_device.py`**: Added 12 tests covering basic usage, context restoration, exception safety, same-device nesting, deep nesting, multi-GPU nesting, `set_current()` inside a `with` block, device usability after exit, device initialization, and thread safety (3 threads on 3 GPUs).
- **`tests/conftest.py`**: Added teardown to `mempool_device_x2` and `mempool_device_x3` fixtures to clean up residual contexts between tests.

## Design

- **Stateless restoration**: Each `__enter__` queries the actual CUDA driver state rather than maintaining a Python-side device cache. This ensures correct interoperability with other libraries (PyTorch, CuPy) that use `cudaSetDevice`/`cuCtxSetCurrent`.
- **Reentrant**: Saved contexts are stored on a per-thread stack (not on the `Device` singleton), so nested and reentrant usage works correctly.
- **Uses `cuCtxGetCurrent`/`cuCtxSetCurrent`**: Consistent with `set_current()` and the runtime API model. Does not use `cuCtxPushCurrent`/`cuCtxPopCurrent`.

## Test Coverage

All tests pass locally on single-GPU (L40) and multi-GPU (3x RTX PRO 6000 Blackwell) machines. Stress-tested with 20 randomized iterations via `pytest-repeat` + `pytest-randomly` with no ordering issues.

### DIFF
--- a/cuda_core/cuda/core/_device.pyx
+++ b/cuda_core/cuda/core/_device.pyx
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # SPDX-License-Identifier: Apache-2.0
 
@@ -1187,6 +1187,49 @@ class Device:
 
     def __reduce__(self):
         return Device, (self.device_id,)
+
+    def __enter__(self):
+        """Set this device as current for the duration of the ``with`` block.
+
+        On exit, the previously current device is restored automatically.
+        Nested ``with`` blocks are supported and restore correctly at each
+        level.
+
+        Returns
+        -------
+        Device
+            This device instance.
+
+        Examples
+        --------
+        >>> from cuda.core import Device
+        >>> with Device(0) as dev0:
+        ...     buf = dev0.allocate(1024)
+
+        See Also
+        --------
+        set_current : Non-context-manager entry point.
+        """
+        cdef cydriver.CUcontext prev_ctx
+        with nogil:
+            HANDLE_RETURN(cydriver.cuCtxGetCurrent(&prev_ctx))
+        if not hasattr(_tls, '_ctx_stack'):
+            _tls._ctx_stack = []
+        _tls._ctx_stack.append(<uintptr_t><void*>prev_ctx)
+        self.set_current()
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        """Restore the previously current device upon exiting the ``with`` block.
+
+        Exceptions are not suppressed.
+        """
+        cdef uintptr_t prev_ctx_ptr = _tls._ctx_stack[-1]
+        cdef cydriver.CUcontext prev_ctx = <cydriver.CUcontext><void*>prev_ctx_ptr
+        with nogil:
+            HANDLE_RETURN(cydriver.cuCtxSetCurrent(prev_ctx))
+        _tls._ctx_stack.pop()
+        return False
 
     def set_current(self, ctx: Context = None) -> Context | None:
         """Set device to be used for GPU executions.
--- a/cuda_core/tests/test_device.py
+++ b/cuda_core/tests/test_device.py
@@ -436,3 +436,179 @@ def test_device_set_membership(init_cuda):
     # Same device_id should not add duplicate
     device_set.add(dev0_b)
     assert len(device_set) == 1, "Should not add duplicate device"
+
+
+# ============================================================================
+# Device Context Manager Tests
+# ============================================================================
+
+
+def _get_current_context():
+    """Return the current CUcontext as an int (0 means NULL / no context)."""
+    return int(handle_return(driver.cuCtxGetCurrent()))
+
+
+def test_context_manager_basic(deinit_cuda):
+    """with Device(0) sets the device as current and restores on exit."""
+    assert _get_current_context() == 0, "Should start with no active context"
+
+    with Device(0):
+        assert _get_current_context() != 0, "Device should be current inside the with block"
+
+    assert _get_current_context() == 0, "No context should be current after exiting"
+
+
+def test_context_manager_restores_previous(deinit_cuda):
+    """Context manager restores the previously active context, not NULL."""
+    dev0 = Device(0)
+    dev0.set_current()
+    ctx_before = _get_current_context()
+    assert ctx_before != 0
+
+    with Device(0):
+        pass
+
+    assert _get_current_context() == ctx_before, "Should restore the previous context"
+
+
+def test_context_manager_exception_safety(deinit_cuda):
+    """Device context is restored even when an exception is raised."""
+    # Start with no active context so restoration is distinguishable
+    assert _get_current_context() == 0
+
+    with pytest.raises(RuntimeError, match="test error"), Device(0):
+        assert _get_current_context() != 0, "Device should be active inside the block"
+        raise RuntimeError("test error")
+
+    assert _get_current_context() == 0, "Must restore NULL context after exception"
+
+
+def test_context_manager_returns_device(deinit_cuda):
+    """__enter__ returns the Device instance for use in 'as' clause."""
+    device = Device(0)
+    with device as dev:
+        assert dev is device
+
+    assert _get_current_context() == 0
+
+
+def test_context_manager_nesting_same_device(deinit_cuda):
+    """Nested with-blocks on the same device work correctly."""
+    dev0 = Device(0)
+
+    with dev0:
+        ctx_outer = _get_current_context()
+        with dev0:
+            ctx_inner = _get_current_context()
+            assert ctx_inner == ctx_outer, "Same device should yield same context"
+        assert _get_current_context() == ctx_outer, "Outer context restored after inner exit"
+
+    assert _get_current_context() == 0
+
+
+def test_context_manager_deep_nesting(deinit_cuda):
+    """Deep nesting and reentrancy restore correctly at each level."""
+    dev0 = Device(0)
+
+    with dev0:
+        ctx_level1 = _get_current_context()
+        with dev0:
+            ctx_level2 = _get_current_context()
+            with dev0:
+                assert _get_current_context() != 0
+            assert _get_current_context() == ctx_level2
+        assert _get_current_context() == ctx_level1
+
+    assert _get_current_context() == 0
+
+
+def test_context_manager_nesting_different_devices(mempool_device_x2):
+    """Nested with-blocks on different devices restore correctly."""
+    dev0, dev1 = mempool_device_x2
+    ctx_dev0 = _get_current_context()
+
+    with dev1:
+        ctx_inside = _get_current_context()
+        assert ctx_inside != ctx_dev0, "Different device should have different context"
+
+    assert _get_current_context() == ctx_dev0, "Original device context should be restored"
+
+
+def test_context_manager_deep_nesting_multi_gpu(mempool_device_x2):
+    """Deep nesting across multiple devices restores correctly at each level."""
+    dev0, dev1 = mempool_device_x2
+
+    with dev0:
+        ctx_level0 = _get_current_context()
+        with dev1:
+            ctx_level1 = _get_current_context()
+            assert ctx_level1 != ctx_level0
+            with dev0:
+                assert _get_current_context() == ctx_level0, "Same device should yield same primary context"
+                with dev1:
+                    assert _get_current_context() == ctx_level1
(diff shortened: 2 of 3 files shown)


cuda_core/cuda/core/_device.pyx
```diff
@@ -1188,6 +1188,49 @@ class Device:
...
+        """Restore the previously current device upon exiting the ``with`` block.
+
+        Exceptions are not suppressed.
+        """
+        cdef uintptr_t prev_ctx_ptr = _tls._ctx_stack[-1]
+        cdef cydriver.CUcontext prev_ctx = <cydriver.CUcontext><void*>prev_ctx_ptr
+        with nogil:
+            HANDLE_RETURN(cydriver.cuCtxSetCurrent(prev_ctx))
+        _tls._ctx_stack.pop()
+        return False
```
cpcloud, line 1232: If `cuCtxSetCurrent` raises here while an exception is already propagating from the `with` block, the CUDA error replaces the user's original exception. The original ends up in `e.__context__` but is easy to miss.

Consider guarding:

```python
def __exit__(self, exc_type, exc_val, exc_tb):
    cdef uintptr_t prev_ctx_ptr = _tls._ctx_stack.pop()
    cdef cydriver.CUcontext prev_ctx = <cydriver.CUcontext><void*>prev_ctx_ptr
    try:
        with nogil:
            HANDLE_RETURN(cydriver.cuCtxSetCurrent(prev_ctx))
    except Exception:
        if exc_type is None:
            raise  # no active exception, surface the CUDA error
        # else: swallow the restore failure to preserve the original exception;
        # the stack entry is already popped so the next __exit__ won't retry.
    return False
```

This also simplifies the peek-then-pop dance — just pop eagerly, since a failed `cuCtxSetCurrent` with a context obtained from `cuCtxGetCurrent` moments earlier is essentially unrecoverable anyway.

(`_graphics.pyx` has the same pattern, so this is a pre-existing issue there too, but worth getting right here.)

cuda_core/cuda/core/_device.pyx
```diff
@@ -1188,6 +1188,49 @@ class Device:
...
+        ...     buf = dev0.allocate(1024)
+
+        See Also
+        --------
+        set_current : Non-context-manager entry point.
+        """
+        cdef cydriver.CUcontext prev_ctx
+        with nogil:
+            HANDLE_RETURN(cydriver.cuCtxGetCurrent(&prev_ctx))
+        if not hasattr(_tls, '_ctx_stack'):
```
cpcloud, line 1216: Nit: the `try`/`except AttributeError` pattern (EAFP) is marginally faster after the first call per thread, and is the more common pattern elsewhere in `_device.pyx` (see `Device_ensure_tls_devices`):

```python
try:
    _tls._ctx_stack
except AttributeError:
    _tls._ctx_stack = []
```

Not blocking — just a consistency suggestion.

cuda_core/tests/test_device.py
```diff
@@ -436,3 +436,179 @@ def test_device_set_membership(init_cuda):
     # Same device_id should not add duplicate
     device_set.add(dev0_b)
     assert len(device_set) == 1, "Should not add duplicate device"
+
+
+# ============================================================================
+# Device Context Manager Tests
+# ============================================================================
```
cpcloud, line 443: Nit: consider putting `_get_current_context()` in `conftest.py` — it's generally useful for any context-related test, and other test files may want it as multi-device / context-switching tests grow.

cpcloud (changes requested):
Solid implementation — the design choices (stateless restoration via `cuCtxGetCurrent`/`cuCtxSetCurrent`, per-thread stack, peek-then-pop in `__exit__`) are all correct and well-reasoned. Tests are thorough. Two items to address before merge:

1. **`__exit__` can mask the caller's exception** — if `cuCtxSetCurrent` raises during unwinding from a user exception, the CUDA error replaces the original (it's still in `__context__`, but most users won't look there). See inline comment for a suggested fix.
2. **Missing docs / release notes** — this adds a new public API entry point on `Device` but the PR doesn't update `interoperability.rst` (the "Current device/context" section should mention the context manager alongside `set_current()`), `getting-started.rst`, or `0.7.x-notes.rst`. Even for a draft, having these roughed in makes the feature discoverable.

leofang (changes requested):
Blocking accidental merge of this PR before a design review/survey happens, as discussed offline and described in the linked issue.

Andy-Jost:
Closing due to lack of activity.

Issue #241, opened 2024-11-13
Sharing buffers with display renderers (like OpenGL, Vulkan, etc.)


jakirkham:
A common use case is to do some computation on the GPU and then render the result in a display ( https://github.com/cupy/cupy/issues/5711 ) ( https://github.com/vispy/vispy/issues/1985 ) ( https://github.com/vispy/vispy/issues/1986 ) ( https://github.com/napari/napari/issues/2243 ). This could be an image that is shown or a plot or something else. Given the registration/unregistration and mapping/unmappaing  of memory, this can generate quite a bit of boilerplate in Python ( https://github.com/vispy/vispy/pull/2391 ), which in turn hinders adoption. Ideally some standard objections/functions could be supplied (ideally in CUDA Python) that users could leverage to prep data for rendering. Preferably they could use one function call to covert any `__cuda_array_interface__`, `__dlpack__`, etc. supporting object into one that can be used for rendering (and hang onto the lifetimes of any supporting buffers via RAII-style semantics)

leofang:
Adding @rparolin to evaluate how the APIs would look like to make comptue-graphics interop easy and performant

rparolin:
@leofang : Is there a use case that we are looking to enable that can help focus the conversation?  Is there a graphics API that the majority of the community is using? For example, pygame? opengl?

leofang:
I think OpenGL should be the primary target. John has linked to various discussions and works in the issue description that some of us contributed to in the past few years, which should be good starting points especially those involving CuPy <-> vispy inter-op.

jakirkham:
My understanding is users would like an object that represents a GL buffer (analogous to `StridedMemoryView`) and has RAII style behavior so cleanup happens during Python GC

Having a function to facilitate that conversion (`StridedMemoryView` to `GLStridedMemoryView`?) would also help

This comment captures this: https://github.com/vispy/vispy/issues/1986#issuecomment-1307609294

mohawk2:
Just adding this to cross-link onto the DLPack issue I opened: https://github.com/dmlc/dlpack/issues/183

PR #1608, opened 2026-02-12
Add CUDA-Graphics (OpenGL) interop support to cuda.core

rparolin:
## Summary

Adds CUDA-Graphics (OpenGL) interoperability support to `cuda.core`, enabling zero-copy sharing of GPU memory between CUDA compute kernels and OpenGL graphics renderers.

- **`GraphicsResource`** — Cython extension class for registering and managing OpenGL buffers/images with CUDA. Supports `from_gl_buffer()` and `from_gl_image()` factory methods, `map()`/`unmap()` with context manager support, and automatic resource cleanup via `close()`.
- **`GraphicsRegisterFlags`** — IntEnum for controlling registration behavior (`NONE`, `READ_ONLY`, `WRITE_DISCARD`, `SURFACE_LOAD_STORE`, `TEXTURE_GATHER`).
- **C++ handle layer** — `GraphicsResourceHandle` (shared_ptr with custom deleter) integrated into the existing resource handle infrastructure.
- **Example** — `gl_interop_plasma.py`: real-time plasma effect rendered by a CUDA kernel directly into an OpenGL PBO, displayed via pyglet.
- **Tests** — Comprehensive test suite covering registration, map/unmap, context manager, error handling, idempotent close, and GC behavior.

### New Public API

```python
import ctypes
from OpenGL import GL as gl
from cuda.core import GraphicsResource, GraphicsRegisterFlags

# Create an OpenGL Pixel Buffer Object (PBO) — GPU memory owned by OpenGL
pbo = ctypes.c_uint(0)
gl.glGenBuffers(1, ctypes.byref(pbo))
gl.glBindBuffer(gl.GL_PIXEL_UNPACK_BUFFER, pbo.value)
nbytes = width * height * 4  # RGBA, 1 byte per channel
gl.glBufferData(gl.GL_PIXEL_UNPACK_BUFFER, nbytes, None, gl.GL_DYNAMIC_DRAW)
pbo_id = pbo.value

# Register the PBO with CUDA for zero-copy access
resource = GraphicsResource.from_gl_buffer(pbo_id, flags=GraphicsRegisterFlags.WRITE_DISCARD)

# Map for CUDA kernel access (context manager auto-unmaps)
with resource.map(stream=stream) as buf:
    launch(stream, config, kernel, buf.handle, width, height, time)

# Cleanup
resource.close()
```

### Files Changed

| File | Description |
|------|-------------|
| `_graphics.pyx` / `_graphics.pxd` | New Cython module with `GraphicsResource` and `GraphicsRegisterFlags` |
| `_cpp/resource_handles.cpp` / `.hpp` | `GraphicsResourceHandle` creation + `cuGraphicsUnregisterResource` function pointer |
| `_resource_handles.pxd` / `.pyx` | Type alias and function pointer initialization for graphics handles |
| `__init__.py` | Export `GraphicsResource` and `GraphicsRegisterFlags` |
| `tests/test_graphics.py` | Test suite (~350 lines) |
| `examples/gl_interop_plasma.py` | Full working example (~410 lines) |
| `pixi.toml` / `pixi.lock` | Added `pyglet` test dependency |

### Screenshot

image

### DIFF
--- a/cuda_core/cuda/core/_graphics.pyx
+++ b/cuda_core/cuda/core/_graphics.pyx
@@ -13,29 +13,34 @@ from cuda.core._resource_handles cimport (
 from cuda.core._stream cimport Stream, Stream_accept
 from cuda.core._utils.cuda_utils cimport HANDLE_RETURN
 
-from enum import IntEnum
-
 from cuda.core._memory import Buffer
 
-__all__ = ['GraphicsResource', 'GraphicsRegisterFlags']
+__all__ = ['GraphicsResource']
+
+_REGISTER_FLAGS = {
+    "none": cydriver.CU_GRAPHICS_REGISTER_FLAGS_NONE,
+    "read_only": cydriver.CU_GRAPHICS_REGISTER_FLAGS_READ_ONLY,
+    "write_discard": cydriver.CU_GRAPHICS_REGISTER_FLAGS_WRITE_DISCARD,
+    "surface_load_store": cydriver.CU_GRAPHICS_REGISTER_FLAGS_SURFACE_LDST,
+    "texture_gather": cydriver.CU_GRAPHICS_REGISTER_FLAGS_TEXTURE_GATHER,
+}
 
 
-class GraphicsRegisterFlags(IntEnum):
-    """Flags for registering a graphics resource with CUDA.
-
-    These flags specify the intended usage when registering a graphics
-    resource (e.g., an OpenGL buffer) for CUDA access.
-    """
-    NONE = cydriver.CU_GRAPHICS_REGISTER_FLAGS_NONE
-    """No hints about how this resource will be used. CUDA may read and write."""
-    READ_ONLY = cydriver.CU_GRAPHICS_REGISTER_FLAGS_READ_ONLY
-    """CUDA will not write to this resource."""
-    WRITE_DISCARD = cydriver.CU_GRAPHICS_REGISTER_FLAGS_WRITE_DISCARD
-    """CUDA will not read from this resource and will write over the entire contents."""
-    SURFACE_LOAD_STORE = cydriver.CU_GRAPHICS_REGISTER_FLAGS_SURFACE_LDST
-    """CUDA will bind this resource to a surface reference."""
-    TEXTURE_GATHER = cydriver.CU_GRAPHICS_REGISTER_FLAGS_TEXTURE_GATHER
-    """CUDA will perform texture gather operations on this resource."""
+def _parse_register_flags(flags):
+    if flags is None:
+        return 0
+    if isinstance(flags, str):
+        flags = (flags,)
+    result = 0
+    for f in flags:
+        try:
+            result |= _REGISTER_FLAGS[f]
+        except KeyError:
+            raise ValueError(
+                f"Unknown register flag {f!r}. "
+                f"Valid flags: {', '.join(sorted(_REGISTER_FLAGS))}"
+            ) from None
+    return result
 
 
 class _MappedBufferContext:
@@ -114,16 +119,20 @@ cdef class GraphicsResource:
         )
 
     @classmethod
-    def from_gl_buffer(cls, int gl_buffer, *, int flags=0) -> GraphicsResource:
+    def from_gl_buffer(cls, int gl_buffer, *, flags=None) -> GraphicsResource:
         """Register an OpenGL buffer object for CUDA access.
 
         Parameters
         ----------
         gl_buffer : int
             The OpenGL buffer name (``GLuint``) to register.
-        flags : int, optional
-            Registration flags from :class:`GraphicsRegisterFlags`.
-            Defaults to :attr:`GraphicsRegisterFlags.NONE`.
+        flags : str or sequence of str, optional
+            Registration flags specifying intended usage. Accepted values:
+            ``"none"``, ``"read_only"``, ``"write_discard"``,
+            ``"surface_load_store"``, ``"texture_gather"``.
+            Multiple flags can be combined by passing a sequence
+            (e.g., ``("surface_load_store", "read_only")``).
+            Defaults to ``None`` (no flags).
 
         Returns
         -------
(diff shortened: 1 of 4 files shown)


cuda_core/examples/gl_interop_plasma.py
```diff
@@ -0,0 +1,410 @@
...
+                stream, config, kernel,
+                buf.handle,           # pointer to PBO memory (on GPU)
+                np.int32(WIDTH),
+                np.int32(HEIGHT),
+                np.float32(t),        # animation time
+            )
+        # (c) Unmap happens automatically when the `with` block exits.
+        #     The PBO now belongs to OpenGL again.
+
+        stream.sync()
```
rparolin: Should this sync be inside the above context to avoid a potential race with the unmap call making the memory available to OpenGL and triggering the barrier that ensures all the PBO values have been written.

cuda_core/cuda/core/_graphics.pyx
```diff
@@ -0,0 +1,319 @@
...
+        if stream is not None:
+            s_obj = Stream_accept(stream)
+            cy_stream = as_cu(s_obj._h_stream)
+
+        cdef cydriver.CUdeviceptr dev_ptr = 0
+        cdef size_t size = 0
+        with nogil:
+            HANDLE_RETURN(
+                cydriver.cuGraphicsMapResources(1, &raw, cy_stream)
+            )
```
Andy-Jost, line 263: optional: you could make the unmap more reliable (and avoid the duplicated calls to `cuGraphicsUnmapResources`) by making this into a resource, similar to the handle.

cuda_core/examples/gl_interop_plasma.py
rparolin: stream.sync() shouldn't be required as the unmap call has an implicit stream sync guaranteeing all cuda work is complete before handing the memory back to OpenGL.

rparolin: https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__INTEROP.html#group__CUDART__INTEROP

leofang:
@jakirkham as the original requester of #241, would it be possible for you to review and provide us your feedbacks here? 🙂

cuda_core/cuda/core/_graphics.pyx
```diff
@@ -0,0 +1,318 @@
...
+    def from_gl_buffer(cls, int gl_buffer, *, int flags=0) -> GraphicsResource:
+        """Register an OpenGL buffer object for CUDA access.
+
+        Parameters
+        ----------
+        gl_buffer : int
+            The OpenGL buffer name (``GLuint``) to register.
+        flags : int, optional
+            Registration flags from :class:`GraphicsRegisterFlags`.
+            Defaults to :attr:`GraphicsRegisterFlags.NONE`.
```
leofang: Drive-by comment: In Python we tend not to use enums. I think in this case accepting a str like `register_option="readonly"` is already good enough.

rparolin: @leofang updated.

cuda_core/cuda/core/_graphics.pyx
```diff
@@ -0,0 +1,337 @@
...
+            mapped = resource.map(stream=s)
+            buf = mapped._buffer  # or use mapped.handle, mapped.size
+            # ... do work ...
+            resource.unmap(stream=s)
+
+        Parameters
+        ----------
+        stream : :class:`~cuda.core.Stream`, optional
+            The CUDA stream on which to perform the mapping. If ``None``,
+            the default stream (``0``) is used.
```
leofang, line 230: @rparolin Sorry I missed this when I skimmed over the PR. In cuda.core we discourage users from letting `stream=None` because of the ambiguity. Everywhere else where a `stream` is accepted, we raise an exception if `stream=None`.

cuda_core/cuda/core/_graphics.pyx
```diff
@@ -0,0 +1,337 @@
...
+        """Unmap this graphics resource, releasing it back to the graphics API.
+
+        After unmapping, the :class:`~cuda.core.Buffer` previously returned
+        by :meth:`map` must not be used.
+
+        Parameters
+        ----------
+        stream : :class:`~cuda.core.Stream`, optional
+            The CUDA stream on which to perform the unmapping. If ``None``,
+            the default stream (``0``) is used.
```
leofang, line 281: ditto regarding `stream=None`

cuda_core/cuda/core/_graphics.pyx
```diff
@@ -0,0 +1,337 @@
...
+
+        Parameters
+        ----------
+        stream : :class:`~cuda.core.Stream`, optional
+            The CUDA stream on which to perform the mapping. If ``None``,
+            the default stream (``0``) is used.
+
+        Returns
+        -------
+        _MappedBufferContext
```
leofang, line 234: Assuming we want `_MappedBufferContext` to be a private API, we don't have to document the return type.

cuda_core/cuda/core/_graphics.pyx
```diff
@@ -0,0 +1,337 @@
...
+            with resource.map(stream=s) as buf:
+                # use buf.handle, buf.size, etc.
+            # automatically unmapped here
+
+        Or called directly for explicit control::
+
+            mapped = resource.map(stream=s)
+            buf = mapped._buffer  # or use mapped.handle, mapped.size
+            # ... do work ...
+            resource.unmap(stream=s)
```
leofang, line 224: I've been pondering about this because the fact that we teach users about 
1. there is an undocumented return type (`mapped`)
2. the buffer needs to be accessed via a private attribute (`mapped._buffer`)

is concerning (recall the `_handle` incident before cutting the 0.6.0 release, https://github.com/NVIDIA/cuda-python/pull/1660). We should avoid teaching poor practices.

@Andy-Jost @rparolin I have been thinking, maybe `GraphicsResource` should just inherit from `Buffer`:
```cython
cdef class GraphicsResource(Buffer):
    def __enter__(...): ...
    def __exit__(...): ...
    def from_gl_buffer(...): ...
    def from_gl_image(...): ...
```
so that we can write cleaner code like this
```python
with GraphicsResource.from_gl_buffer(vbo, stream=s) as buf:
    view = StridedMemoryView.from_buffer(buf, shape=(256,), dtype=np.float32)
```
or this
```python
buf = GraphicsResource.from_gl_buffer(vbo, stream=s)
# ... do work ...
buf.close()  # unmap on s since Buffer can internally hold a deallocation stream
```
This also matches the built-in `mmap` closer: https://docs.python.org/3/library/mmap.html. WDYT?

rparolin: I 💯 agree with the `_buffer` feedback -- that was included in this PR by mistake.

rparolin: @leofang Do we need to inherit from Buffer or should we just implement the context manager enter/exit functions?

leofang: Oh! Right, probably no inheritance is fine as long as 
- `GraphicsResource.from_gl_buffer` can be used as a context manager
- `GraphicsResource.from_gl_buffer` returns a `Buffer`

leofang:
Follow-up: https://github.com/NVIDIA/cuda-python/pull/1701

PR #1609, opened 2026-02-12
[REL] Add git archival support for setuptools-scm

mdboom:
This allows setuptools-scm to work when our projects are built from a `git archive`, which
includes how our official conda packages are built.

Useful references for existing `git archive` usage is:

- [Where our git archives are created](https://github.com/NVIDIA/cuda-python/blob/main/.github/workflows/release-upload.yml#L54)
- [Where out git archives are downloaded as part of conda feedstock](https://github.com/conda-forge/cuda-pathfinder-feedstock/blob/main/recipe/recipe.yaml#L12)

There are some important differences here from how the [setuptools-scm docs](https://setuptools-scm.readthedocs.io/en/latest/usage/#git-archives) suggest:

- Since we are a monorepo, we need a `.git_archival.txt` file at the root of each of our subprojects.  We can't just copy this file there, because `git archive` will only perform one `git describe` substitution per repo to avoid a denial-of-service attack.  Instead, there is a single `.git_archival.txt` file at the root of this repository, and symlinks to it from each of the subprojects.  This may break if creating git archives on Windows, which we don't currently do.

- Using the suggested `describe-name: $Format:%(describe:tags=true,match=*[0-9]*)$` doesn't work for our tagging scheme where there may be a project prefix, e.g. `cuda-pathfinder-v1.3.4`. The bug exists in `setuptools-scm` where it assumes a version in a `.git_archival.txt` file doesn't have a prefix and is /exactly/ what should be used as the version in the package.  Instead, we use the `ref-names` approach, which includes more detailed information about the available tags, and `setuptools-scm` falls back to the more customizable form of version parsing.  The docs warn that this may cause different checksums in the `git-archive` from the same tag.  I don't know if this will be a problem for us in practice, since we do it once in CI.

With this change, the workaround in our Conda feedstock of `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_CUDA_PATHFINDER=${{ version }}` should no longer be necessary, though I think it is harmless to leave it that way.

### DIFF
--- a/cuda_bindings/cuda/bindings/_internal/_fast_enum.py
+++ b/cuda_bindings/cuda/bindings/_internal/_fast_enum.py
@@ -1,0 +1,76 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
+
+
+# This code was automatically generated across versions from 12.9.1 to 13.1.1. Do not modify it directly.
+
+
+"""
+This is a replacement for the stdlib enum.IntEnum.
+
+Notably, it has much better import time performance, since it doesn't generate
+and evaluate Python code at startup time.
+
+It supports the most important subset of the IntEnum API.  See `test_enum` in
+`cuda_bindings/tests/test_basics.py` for details.
+"""
+
+from typing import Any, Iterator
+
+
+class FastEnumMetaclass(type):
+    def __init__(cls, name, bases, namespace):
+        super().__init__(name, bases, namespace)
+
+        cls.__singletons__ = {}
+        cls.__members__ = {}
+        for name, value in cls.__dict__.items():
+            if name.startswith("__") and name.endswith("__"):
+                continue
+
+            if isinstance(value, tuple):
+                value, doc = value
+            elif isinstance(value, int):
+                doc = None
+            else:
+                continue
+
+            singleton = int.__new__(cls, value)
+            singleton.__doc__ = doc
+            singleton._name = name
+            cls.__singletons__[value] = singleton
+            cls.__members__[name] = singleton
+
+        for name, member in cls.__members__.items():
+            setattr(cls, name, member)
+
+    def __repr__(cls) -> str:
+        return f"<enum '{cls.__name__}'>"
+
+    def __len__(cls) -> int:
+        return len(cls.__members__)
+
+    def __iter__(cls) -> Iterator["FastEnum"]:
+        return iter(cls.__members__.values())
+
+    def __contains__(cls, item: Any) -> bool:
+        return item in cls.__singletons__
+
+
+class FastEnum(int, metaclass=FastEnumMetaclass):
+    def __new__(cls, value: int) -> "FastEnum":
+        singleton: FastEnum = cls.__singletons__.get(value)
+        if singleton is None:
+            raise ValueError(f"{value} is not a valid {cls.__name__}")
+        return singleton
+
+    def __repr__(self) -> str:
+        return f"<{self.__class__.__name__}.{self._name}: {int(self)}>"
+
+    @property
+    def name(self) -> str:
+        return self._name
+
+    @property
+    def value(self) -> int:
+        return int(self)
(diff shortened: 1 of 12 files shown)


.gitattributes
```diff
@@ -1,5 +1,3 @@
-cuda/_version.py export-subst
```
mdboom, line 1: This is a file that hasn't existed in the repo for a long time.  Just removing it for cleanup -- not strictly necessary.

.pre-commit-config.yaml
```diff
@@ -29,7 +29,7 @@ repos:
         language: python
         additional_dependencies:
           - https://files.pythonhosted.org/packages/cc/20/ff623b09d963f88bfde16306a54e12ee5ea43e9b597108672ff3a408aad6/pathspec-0.12.1-py3-none-any.whl
-        exclude: '.*pixi\.lock'
+        exclude: '(.*pixi\.lock)|(\.git_archival\.txt)'
```
mdboom, line 32: We can't put any "extra" content in this file, including a copyright header, so we need to have the SPDX precommit ignore it.

rwgk:
> This may break if creating git archives on Windows, which we don't currently do.

I had run-ins with softlinks (similar experience to getting that mosquito buzzing around your ears in the middle of the night).

We are solidly vested in pre-commit:

Could we make it so that pre-commit enforces that the **copies** are in sync? ("One Cursor prompt and it's done?")

.gitattributes
```diff
@@ -16,3 +14,5 @@ cuda_core/cuda/core/_cpp/*.hpp -binary text diff
 *.svg binary
 # SCM syntax highlighting & preventing 3-way merges
 pixi.lock merge=binary linguist-language=YAML linguist-generated=true
+
+.git_archival.txt export-subst
```
rwgk, line 22: There is no immediately obvious connection from `.git_archival.txt` to `setuptools-scm`. If you know it, good, if you don't you have to go through a couple hoops to discover what this is for. Easy idea to help at least a little bit:

```
# See PR #1609
.git_archival.txt export-subst
```

Maybe there is a better way to make the connection more discoverable?

.git_archival.txt
```diff
@@ -0,0 +1,3 @@
+node: $Format:%H$
```
rwgk, line 1: To document this here: I asked Cursor(GPT-5.3 Codex Extra High):

```
Do you know if we can add comments at the top of the file? Without, there is no immediately obvious connection to setuptools-scm.
```

The response:

```
  Short answer: not as normal comment lines.
  • setuptools-scm parses .git_archival.txt as RFC822-style headers (key: value), not as a general text file.
  • A top line like # this is for setuptools-scm is a non-header line, which stops header parsing, so node, node-date, etc. may not be read.
  • Safe alternative inside the file: add an extra header field, e.g. x-setuptools-scm: git archive metadata (unknown keys are ignored).
  • Best place for human-readable explanation is usually .gitattributes, e.g. a comment above .git_archival.txt export-subst.
```

I saw Cursor writing code to actually try it out locally, so I believe this is a reliable/conclusive finding.

mdboom:
> Could we make it so that pre-commit enforces that the **copies** are in sync? ("One Cursor prompt and it's done?")

How this works is that any file with the `export-subst` attribute will get parsed and have the values filled in when `git archive` builds the archive.  However, [according to the git docs](https://git-scm.com/docs/gitattributes), "However, only one %(describe) placeholder is expanded per archive to avoid denial-of-service attacks."  So if you have multiple files, all but the first one doesn't get filled in.  So any solution with multiple copies of the file will fail.

.gitattributes
mdboom: I've added a comment here.

rwgk:
FYI - sharing observations

Cursor is telling me that the next conda pathfinder release will work, based on this PR, even with the `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_CUDA_PATHFINDER` override removed again.

However, installing from a github zip file still does not work. Concretely, this reproducer is still failing:

https://github.com/NVIDIA/cuda-python/pull/1454#issuecomment-3741379903

I tried to find a fix, but it seems to be somewhere between complicated and impractical. Is this actually important for us?

leofang:
@mdboom could you please kindly backport this PR?

PR #1610, opened 2026-02-12
[REL]: Add a workflow to tag a release

mdboom:
This is a suggested solution to workaround the issue we ran into with the recent `cuda_pathfinder` release.

(Initial revision created with Claude -- be extra cautious about its errors).

This adds a tag to the repo by first creating a new empty commit and then tagging that.  This will force `CI` to rerun with a tagged commit and create wheels and tarballs that have a proper tagged version, which then should be handled correctly when the `CI: Release` workflow is used to upload to PyPI etc.

This also acts as a starting point to put more automation steps for things we want to do when we tag a release.

An important downside of this approach is that you can't specify a specific commit to tag for release -- it needs to be the tip of main right now.  (Because we need to create a new commit, and we need to put the new commit somewhere).  This is a subtle "race condition" because things could be merged during the moment the user is trying to tag a release.  I thought I would put this up to create some discussion around this.  One nice solution would be to create a branch, then create the new commit and tag on that branch, which leaves a nice branch around to create bug fix releases from in the future.  But that would no longer put tags on main as we currently do.

This also updates the `CI: Release` task so it takes a component and a version, rather than a git tag, to be inline with this new workflow and reduce the ability to create an invalid tag.  I thought having these two workflows be different would cause a lot of confusion.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,7 +40,7 @@ on:
           - pypi
 
 env:
-  TAG: ${{ fromJSON('{"cuda_bindings":"v","cuda_core":"cuda-core-v","cuda_pathfinder":"cuda-pathfinder-v","cuda_python":"cuda-python-v","all":"all-v"}')[inputs.component] }}${{ inputs.version }}
+  TAG: ${{ fromJSON('{"cuda_bindings":"v","cuda_core":"cuda-core-v","cuda_pathfinder":"cuda-pathfinder-v","cuda_python":"v","all":"v"}')[inputs.component] }}${{ inputs.version }}
 
 defaults:
   run:


mdboom:
Test run of this on my own fork here: https://github.com/mdboom/cuda-python/actions/runs/21956196329/job/63420782798

rwgk:
Hm, did you see the PR #1606 that I posted yesterday (Cursor-generated)?

That adds a tag-push-trigger, so we don't need an empty commit.

Do we **want** an empty commit?

(Note that #1606 also adds extra safety, ci/tools/validate-release-wheels; without that it'd just be a small surgical PR.)

rwgk:
> An important downside of this approach is that you can't specify a specific commit to tag for release -- it needs to be the tip of main right now. 

The tag-push-trigger approach doesn't have that downside. According to what Cursor explained to me, we can even tag cuda-pathfinder, cuda-bindings, cuda-core concurrently and it'll all work.

mdboom:
Sorry, @rwgk, I didn't see #1606.  Let me look...

.github/workflows/tag-release.yml
leofang, line 1: @mdboom @rwgk I like this file. Can we keep it so that no members on the team would ever need to manually push a tag to this repo?

leofang: I think this file alone (without the change in the other file) is composable with #1606. 
- We run this new workflow to generate a tag
- The tag creation triggers the CI

rwgk: That works for me. It's a bit on the heavy-handed side just for the tagging, but if we fold in some easy safety features I can see good value:

* pull-down for: cuda-pathfinder, cuda-bindings, cuda-core
* pull-down for: bump patch, bump minor, bump major
* **compute** the version number
* validate that X.Y.Z-notes.rst exists
* validate that corresponding the nv-versions.json entry exits

That would prevent oversights like we had in the release of cuda-bindings 13.1.0 (missing release notes under cuda-python).

mdboom: @leofang: At a minimum we need to update this so it doesn't create an empty commit (which is no longer needed after #1606).  And if we do that, we also have ways to fix the race condition problem.  

The race condition is: I wanted to create a release, so I go to the page to trigger this workflow, and then someone else merges a PR while the workflow is waiting to run.  Now we've made a release with unintended content.

If we require the user of this workflow to specify a commit hash to tag, we avoid the race condition.

FWIW, even those the GitHub UI for creating a release tag doesn't do everything we need, has the ability to specify a commit in their design since it's a real issue:

image

And then the other thing to test is:

> - The tag creation triggers the CI

There are a bunch of rules about GitHub workflows triggering other workflows (to prevent infinite loops etc).  This may not work -- we will need to test.

mdboom: As I think about this further, another downside is that this workflow could only ever /validate/ the kinds of things @rwgk is suggesting.  (If it needs to add a commit to add content to the repo, we are back to the problems outlined in the OP).  Is it really useful to have a workflow just to validate these things when instead we could write a local script that could actually /do/ those things and then add the tag? (and the push would correctly and helpfully fail if main had been updated in the meantime)?

leofang: > > * The tag creation triggers the CI
> 
> There are a bunch of rules about GitHub workflows triggering other workflows (to prevent infinite loops etc). This may not work -- we will need to test.

Isn't this just handled by #1606?

mdboom: #1606 is triggered by adding a tag either locally or from the GitHub UI.  Having a workflow create the tag may not work, because of the infinite loop avoidance rules.  It's a little hard to tell from the docs -- we will just have to test it.

cpcloud:
Not a huge fan of release steps being split. It makes debugging releases more difficult because you have to follow two workflows (or maybe three if we split docs). Not to mention that we now have another github actions feature to understand: workflows triggering other workflows.

Why can't we do everything in a single workflow? Tag -> draft release -> test pypi -> pypi -> docs -> undraft release?

cpcloud (changes requested):
## Review: Tag Release Workflow

There are several bugs that need fixing before this can merge.

### Critical

**1. `vars.TAG` vs `env.TAG` in `release.yml` (two locations)**

The PR defines `TAG` as a workflow-level `env` variable but then references it as `vars.TAG` in the `doc` and `upload-archive` job `with:` blocks. `vars.TAG` refers to GitHub repository-level variables (Settings > Variables), not the workflow `env`. These will resolve to empty/undefined, silently breaking both downstream workflows.

Should be `${{ env.TAG }}` in both places.

**2. `env.TAG` JSON map keys don't match `component` input options in `release.yml`**

The new `env.TAG` definition uses underscore keys:
```
{"cuda_bindings":"v","cuda_core":"cuda-core-v","cuda_pathfinder":"cuda-pathfinder-v","cuda_python":"v","all":"v"}
```
But the `component` input options use hyphens: `cuda-core`, `cuda-bindings`, `cuda-pathfinder`, `cuda-python`. The `fromJSON(...)[inputs.component]` lookup will fail for every individual component, producing a TAG that's just the bare version number with no prefix (e.g. `1.2.3` instead of `v1.2.3`).

Either the JSON keys need to use hyphens to match the existing input options, or the input options need to change to underscores (which would be a broader breaking change).

**3. `validate-release-wheels` still references the old `inputs.git-tag` (line 170)**

The input was renamed from `git-tag` to `version`, but the `validate-release-wheels` invocation in the `publish-wheels` job was not updated. `${{ inputs.git-tag }}` will be empty after the rename, so wheel validation will pass or fail incorrectly. This should be `${{ env.TAG }}` to stay consistent with the rest of the workflow.

### Medium

**4. No version format validation**

Neither workflow validates the version string format. If someone enters `v1.2.3` instead of `1.2.3`, the resulting tag would be `vv1.2.3`. A validation step (e.g. a regex check that the version matches `^\d+\.\d+\.\d+`) would prevent this class of error.

**5. `cuda_bindings` implies `cuda_python` but only one tag is created**

The `tag-release.yml` description says "cuda_bindings implies cuda_python as well," but the workflow only creates a single tag. If a dual-tag is intended, it's missing. If not, the description should be updated to avoid confusion.

### Low

**6. Direct push to `main` may conflict with branch protections**

`tag-release.yml` does `git push origin main`. If branch protection rules require PRs or status checks, this will fail at runtime. The PR description acknowledges this design tension — the branch-based approach mentioned there would avoid this.

mdboom:
Aside from the little bugs highlighted, I think we need to step back to the overall design for a release and ensure this still fits in with that.

mdboom:
@cpcloud: Since you're deep in release automation right now, do you just want to take this over?  If not, I can take it, but I've definitely swapped all of the context out of my head on this atm ;)

Issue #1247, opened 2025-11-17
Update `pyproject.toml` for `uv`


leofang:
Capturing feedbacks provided by @xiakun-lu offline.

The NCCL team noticed that `uv sync` complains `nccl4py[cu12]` and `nccl4py[cu13]` are incompatible (`uv venv && uv pip install -e .` works out of box):
```toml
[project.optional-dependencies]
cu12 = [
    "nvidia-nccl-cu12",
    "cuda-bindings~=12.0",
]
cu13 = [
    "nvidia-nccl-cu13",
    "cuda-bindings~=13.0",
]
```
With CPython 3.14.0:
```
Creating virtual environment at: .venv
  × No solution found when resolving dependencies for split (markers: python_full_version >= '3.13'):
  ╰─:arrow_forward: Because nccl4py[cu13] depends on cuda-bindings>=13.0,<14.dev0 and nccl4py[cu12] depends on cuda-bindings>=12.0,<13.dev0, we can conclude that nccl4py[cu12] and nccl4py[cu13] are incompatible.
      And because your project requires nccl4py[cu12] and nccl4py[cu13], we can conclude that your project's requirements are unsatisfiable.
```

It can be fixed by adding this configuration:
```toml
[tool.uv]
conflicts = [
    [
        { extra = "cu12" },
        { extra = "cu13" },
    ]
]
```

leofang:
xref: https://docs.astral.sh/uv/reference/settings/#conflicts

leofang:
cc @kkraus14 for vis (since you're playing with uv 😉)

PR #1615, opened 2026-02-12
Switch CI to uv and use build[uv] frontend for cibuildwheel

kkraus14:
## Summary

Replaces `pip` with `uv` across all CI workflows and switches cibuildwheel to the `build[uv]` build frontend.

### CI workflow changes

- **build-wheel.yml**: `uv build --wheel` for pathfinder/cuda-python, `uv tool run twine` for wheel checks, `uv venv` + `uv pip install --group` for Cython test deps.
- **test-wheel-linux.yml / test-wheel-windows.yml**: `setup-uv` with `activate-environment`, `uv pip install` for wheels + `--group` for dependency groups.
- **build-docs.yml**: `uv pip install --system` into conda env.
- **coverage.yml**: `setup-uv` with `activate-environment`, `uv pip install` + `--group` for test deps.
- **ci/tools/run-tests**: `uv pip install` for wheels + `--group` for dependency groups.
- Uses `astral-sh/setup-uv@v7.3.0` (pinned to SHA) with `activate-environment: true` in test/coverage workflows.
- Uses `uv run --no-project --with wheel` for the cuda.core wheel merge step.

### cibuildwheel

- Switches to `build[uv]` build frontend for `cuda_bindings` and `cuda_core` on all platforms.
- Downgrades cibuildwheel to v3.2.1 to avoid a PATH mangling bug in v3.3.x on Windows ([pypa/cibuildwheel#2663](https://github.com/pypa/cibuildwheel/issues/2663)). Upgrade back once a fixed release is available.

### Package metadata

- Adds `[dependency-groups]` with test dependencies to `cuda_bindings` and `cuda_core`.
- Adds `pytest-randomly` to `cuda_bindings` test dependency group.
- Replaces `cuda-core[test]` self-reference in `cuda_core`'s `test-cu*` groups with `{include-group = "test"}` ([PEP 735](https://peps.python.org/pep-0735/)) to avoid triggering the build backend in CI.
- Adds `[tool.uv]` cu12/cu13 extra conflicts to `cuda_core`.

## Key patterns

- **Dependency groups via `--group`**: `uv pip install ./artifact.whl --group test` installs a wheel and its test dependency group in one call. Groups must not self-reference the project (use `{include-group = "..."}` instead) to avoid triggering the build backend.
- **Pre-built wheels**: `uv pip install path/to/*.whl` installs CI-built wheel artifacts into the venv.
- **One-shot tools**: `uv tool run twine check --strict *.whl` and `uv run --no-project --with wheel python script.py` avoid polluting the environment.

## Test plan

- [ ] All build jobs pass (linux-64, linux-aarch64, win-64, all Python versions including py3.14t)
- [ ] All test jobs pass (linux + windows, local + wheels, all GPU types)
- [ ] Docs build passes
- [ ] Coverage workflow passes


cuda_python/pyproject.toml
```diff
@@ -32,15 +32,24 @@ classifiers = [
     "Environment :: GPU :: NVIDIA CUDA :: 12",
     "Environment :: GPU :: NVIDIA CUDA :: 13",
 ]
-dynamic = ["version", "dependencies", "optional-dependencies"]
+dynamic = ["version"]
 requires-python = ">=3.10"
+dependencies = [
+    "cuda-bindings",
```
kkraus14: This is losing the CUDA major version

.github/workflows/test-wheel-windows.yml
```diff
@@ -208,6 +208,9 @@ jobs:
         with:
           python-version: ${{ matrix.PY_VER }}
 
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
```
kkraus14: Use SHA instead of tagged version here

.github/workflows/test-wheel-linux.yml
```diff
@@ -224,6 +224,9 @@ jobs:
           # we use self-hosted runners on which setup-python behaves weirdly (Python include can't be found)...
           AGENT_TOOLSDIRECTORY: "/opt/hostedtoolcache"
 
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
```
kkraus14: +1 SHA

.github/workflows/coverage.yml
```diff
@@ -82,6 +82,9 @@ jobs:
           # we use self-hosted runners on which setup-python behaves weirdly...
           AGENT_TOOLSDIRECTORY: "/opt/hostedtoolcache"
 
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
```
kkraus14: +1

.github/workflows/build-wheel.yml
```diff
@@ -106,15 +106,18 @@ jobs:
         run: |
           env
 
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
```
kkraus14: +1

.github/workflows/build-docs.yml
```diff
@@ -77,6 +77,9 @@ jobs:
           conda config --show-sources
           conda config --show
 
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
```
kkraus14: +1

.github/workflows/build-wheel.yml
```diff
@@ -267,8 +270,8 @@ jobs:
         if: ${{ strategy.job-index == 0 && inputs.host-platform == 'linux-64' }}
         run: |
           pushd cuda_python
-          pip wheel -v --no-deps .
-          twine check --strict *.whl
+          uv build --wheel
+          twine check --strict dist/*.whl
```
kkraus14: This may not be needed anymore in using `uv`?

pyproject.toml
```diff
@@ -0,0 +1,124 @@
...
+        { package = "cuda-bindings", extra = "all" },
+        { package = "cuda-core", group = "test-cu12-ft" },
+    ],
+    [
+        { package = "cuda-bindings", extra = "all" },
+        { package = "cuda-pathfinder", group = "test-cu12" },
+    ],
+    [
+        { package = "cuda-bindings", extra = "all" },
+        { package = "cuda-pathfinder", group = "cu12" },
```
kkraus14: These are incorrect since they assume `cuda-bindings==13.*`, but `cuda-bindings==12.*` is also supported by pathfinder

.github/workflows/build-wheel.yml
```diff
@@ -330,7 +321,8 @@ jobs:
 
       - name: Build cuda.core Cython tests
         run: |
-          pip install ${{ env.CUDA_CORE_ARTIFACTS_DIR }}/"cu${BUILD_CUDA_MAJOR}"/*.whl --group ./cuda_core/pyproject.toml:test
+          uv pip install ${{ env.CUDA_CORE_ARTIFACTS_DIR }}/"cu${BUILD_CUDA_MAJOR}"/*.whl
+          uv pip install "pytest>=6.2.4" pytest-randomly pytest-repeat "cython>=3.2,<3.3" setuptools
```
mdboom: We should find a solution that doesn't duplicate the test requirements already specified in `pyproject.toml`.

kkraus14: Done

pyproject.toml
kkraus14: Since `main` only supports `cuda-bindings==13.*` currently this was actually correct.

cuda_core/pyproject.toml
```diff
@@ -90,7 +94,15 @@ version_file = "cuda/core/_version.py"
 tag_regex = "^cuda-core-(?P<version>v\\d+\\.\\d+\\.\\d+)"
 git_describe_command = ["git", "describe", "--dirty", "--tags", "--long", "--match", "cuda-core-v*[0-9]*"]
 
+[tool.uv]
+conflicts = [[{ extra = "cu12" }, { extra = "cu13" }]]
```
leofang, line 95: Ah, it's already here, so this PR will close #1247

.github/workflows/build-wheel.yml
```diff
@@ -312,11 +322,12 @@ jobs:
...
       - name: Install cuda.pathfinder (required for next step)
         run: |
-          pip install cuda_pathfinder/*.whl
+          uv pip install --system cuda_pathfinder/dist/*.whl
 
       - name: Build cuda.bindings Cython tests
         run: |
-          pip install ${{ env.CUDA_BINDINGS_ARTIFACTS_DIR }}/*.whl --group ./cuda_bindings/pyproject.toml:test
+          uv pip install --system ${{ env.CUDA_BINDINGS_ARTIFACTS_DIR }}/*.whl
+          uv export --frozen --package cuda-bindings --only-group test --no-emit-project --no-emit-workspace --no-hashes | uv pip install --system -r -
```
leofang: This incantation is scary 😱 Can we add a short comment here?

.github/workflows/build-wheel.yml
```diff
@@ -330,7 +341,8 @@ jobs:
 
       - name: Build cuda.core Cython tests
         run: |
-          pip install ${{ env.CUDA_CORE_ARTIFACTS_DIR }}/"cu${BUILD_CUDA_MAJOR}"/*.whl --group ./cuda_core/pyproject.toml:test
+          uv pip install --system ${{ env.CUDA_CORE_ARTIFACTS_DIR }}/"cu${BUILD_CUDA_MAJOR}"/*.whl
+          uv export --frozen --package cuda-core --only-group test --no-emit-project --no-emit-workspace --no-hashes | uv pip install --system -r -
```
leofang: ditto

.github/workflows/test-wheel-linux.yml
```diff
@@ -279,17 +282,18 @@ jobs:
...
           fi
 
       - name: Install cuda.pathfinder extra wheels for testing
         run: |
           set -euo pipefail
           pushd cuda_pathfinder
-          pip install --only-binary=:all: -v ./*.whl --group "test-cu${TEST_CUDA_MAJOR}"
-          pip list
+          uv pip install --system -v ./*.whl
+          uv export --frozen --package cuda-pathfinder --only-group "test-cu${TEST_CUDA_MAJOR}" --no-emit-project --no-emit-workspace --no-hashes | uv pip install --system -r -
```
leofang: 😱

cuda_python/pyproject.toml
```diff
@@ -41,6 +41,10 @@ documentation = "https://nvidia.github.io/cuda-python/"
...
-# The [tool.setuptools_scm] section is handled in setup.py since we need to
-# dynamically set the dependency to cuda_bindings based on the dynamically
-# determinded version
+# Version from git tags; setup.py still overrides install_requires at wheel build
+# for exact version pinning (~= for release, == for dev).
+[tool.setuptools_scm]
+root = ".."
+version_file = "cuda_python/_version.py"
+tag_regex = "^(?P<version>v\\d+\\.\\d+\\.\\d+)"
+git_describe_command = ["git", "describe", "--dirty", "--tags", "--long", "--match", "v*[0-9]*"]
```
leofang: @mdboom to review this change

pyproject.toml
```diff
@@ -0,0 +1,126 @@
...
+[tool.uv]
+package = false
+# cuda-pathfinder must appear as a direct constraint so that the resolver
+# recognises it as a workspace member when it surfaces as a *transitive*
+# build-dependency (cuda-core build -> cuda-bindings -> cuda-pathfinder).
+constraint-dependencies = ["cuda-pathfinder"]
+
+# Workspace-level conflicts: cuda-toolkit can only be one major version at a
+# time, so every cu12 extra/group is incompatible with every cu13 one.
+conflicts = [
```
leofang: Q: It seems we have to list all possible combinations of conflicts? Is this the only way at the workspace level? Maybe uv could honor the conflicts in each individual pyproject.toml? We currently only add a conflict for cuda-core.

kkraus14: This is on my list to get cursor to see if it can simplify 😄. I agree this is quite annoying.

.github/workflows/build-wheel.yml
kkraus14: I think a lot of these are gross because we aren't using a venv and so some of the normal commands don't work. Just trying to get everything working first and then I'll clean it up.

pyproject.toml
kkraus14: @leofang see the updated comment. Unfortunately uv doesn't automatically propagate the information upwards so we need to keep this here in addition to defining it at the package level.

I'll open an issue upstream and link it here.

cpcloud:
There are build _frontends_?

.github/workflows/build-docs.yml
```diff
@@ -77,6 +77,9 @@ jobs:
           conda config --show-sources
           conda config --show
 
+      - name: Install uv
+        uses: astral-sh/setup-uv@d4b2f3b6ecc6e67c4457f6d3e41ec42d3d0fcb86  # v5
```
kkraus14: We may want to just install uv via the conda package here

.github/workflows/build-docs.yml
```diff
@@ -181,18 +184,18 @@ jobs:
       - name: Install all packages
         run: |
           pushd cuda_pathfinder
-          pip install *.whl
+          uv pip install --system *.whl
```
kkraus14, line 187: If we don't install `uv` via the conda package do we need to explicitly target the conda python as opposed to just using `--system`?

.github/workflows/test-wheel-linux.yml
```diff
@@ -224,6 +224,12 @@ jobs:
           # we use self-hosted runners on which setup-python behaves weirdly (Python include can't be found)...
           AGENT_TOOLSDIRECTORY: "/opt/hostedtoolcache"
 
+      - name: Install uv
+        uses: astral-sh/setup-uv@d4b2f3b6ecc6e67c4457f6d3e41ec42d3d0fcb86  # v5
+
+      - name: Create venv
+        run: uv venv
```
kkraus14: I think the `setup-uv` action has the ability to automatically create and activate a venv. We should use that.

.github/workflows/build-docs.yml
kkraus14: This and all other usages should use a newer version if possible. v7.3.0 is available for example.

.github/workflows/build-wheel.yml
```diff
@@ -106,15 +106,28 @@ jobs:
         run: |
           env
 
-      - name: Install twine
+      - name: Install uv
+        uses: astral-sh/setup-uv@d4b2f3b6ecc6e67c4457f6d3e41ec42d3d0fcb86  # v5
```
kkraus14: If Python isn't being managed by uv we need to follow the instructions here: https://docs.astral.sh/uv/guides/integration/github/#setting-up-python

.github/workflows/build-docs.yml
kkraus14: Yes — `--system` targets whichever `python` is first on PATH, and `setup-miniconda` with `activate-environment` puts the conda Python there, so `uv pip install --system` installs directly into the conda env's site-packages.

.github/workflows/build-wheel.yml
```diff
@@ -444,15 +469,15 @@ jobs:
 
       - name: Merge cuda.core wheels
         run: |
-          pip install wheel
+          uv pip install --system wheel
           python ci/tools/merge_cuda_core_wheels.py \
```
kkraus14: Can run this via uv with the needed `wheel` and other dependencies instead of installing `wheel` into the system install

.github/workflows/build-wheel.yml
```diff
@@ -106,15 +106,25 @@ jobs:
...
+      # stay within cmd.exe's ~8KB environment limit (needed for py3.14t).
+      - name: Shorten TEMP path (Windows)
+        if: ${{ startsWith(inputs.host-platform, 'win') }}
+        shell: pwsh
         run: |
-          pip install twine
+          $short = "C:\t"
+          New-Item -ItemType Directory -Force -Path $short | Out-Null
+          "TEMP=$short" | Out-File -FilePath $env:GITHUB_ENV -Append
+          "TMP=$short" | Out-File -FilePath $env:GITHUB_ENV -Append
```
kkraus14: This was all quite surprising to need to do. We should check if we can remove it and if we absolutely can't we should try to make the temp directory at least a little bit better of a name.

.github/workflows/build-wheel.yml
```diff
@@ -310,13 +325,23 @@ jobs:
...
+
+      - name: Install test deps and wheels for Cython tests
         run: |
-          pip install cuda_pathfinder/*.whl
+          uv pip install cuda_pathfinder/dist/*.whl
+          uv pip install ${{ env.CUDA_BINDINGS_ARTIFACTS_DIR }}/*.whl
+          uv pip install ${{ env.CUDA_CORE_ARTIFACTS_DIR }}/"cu${BUILD_CUDA_MAJOR}"/*.whl
+          # --inexact preserves the wheels installed above
+          uv sync --frozen --inexact --package cuda-bindings --only-group test --no-install-project
+          uv sync --frozen --inexact --package cuda-core --only-group test --no-install-project
```
kkraus14: I'm not sure how worthwhile it is doing this versus just using `uv pip install` with dependency groups and not dealing with the uv lock file for now.

.github/workflows/build-wheel.yml
```diff
@@ -330,7 +355,7 @@ jobs:
 
       - name: Build cuda.core Cython tests
         run: |
-          pip install ${{ env.CUDA_CORE_ARTIFACTS_DIR }}/"cu${BUILD_CUDA_MAJOR}"/*.whl --group ./cuda_core/pyproject.toml:test
+          export PATH="${GITHUB_WORKSPACE}/.venv/bin:${GITHUB_WORKSPACE}/.venv/Scripts:${PATH}"
```
kkraus14: This is a bit surprising. Why is this needed if the venv is activated?

.github/workflows/build-wheel.yml
```diff
@@ -192,9 +202,10 @@ jobs:
           $CHOWN -R $(whoami) ${{ env.CUDA_BINDINGS_ARTIFACTS_DIR }}
           ls -lahR ${{ env.CUDA_BINDINGS_ARTIFACTS_DIR }}
 
+
```
kkraus14:
```suggestion
```

cuda_python/pyproject.toml
mdboom: Makes sense to me.  If I'm following, this was just missing from `cuda_python` and this is copy-pasted from `cuda_bindings`.

ci/tools/run-tests
```diff
@@ -74,16 +74,12 @@ elif [[ "${test_module}" == "core" ]]; then
 
   pushd ./cuda_core
   if [[ "${LOCAL_CTK}" == 1 ]]; then
-    # We already installed cuda-bindings, and all CTK components exist locally,
-    # so just install the test dependencies.
-    pip install "${CUDA_CORE_ARTIFACTS_DIR}"/*.whl --group "test-cu${TEST_CUDA_MAJOR}${FREE_THREADING}"
+    uv pip install "${CUDA_CORE_ARTIFACTS_DIR}"/*.whl --group "test-cu${TEST_CUDA_MAJOR}${FREE_THREADING}"
```
cpcloud, line 77: Blocking: in the LOCAL_CTK==1 core install path, the previous minor-version constraint on cuda-toolkit appears removed. Without pinning to the local CTK minor, dependency resolution can pull a newer runtime than the locally installed toolkit and reintroduce environment skew/flakiness.

cuda_bindings/pyproject.toml
```diff
@@ -46,6 +46,7 @@ test = [
     "numpy>=1.21.1",
     "pytest>=6.2.4",
     "pytest-benchmark>=3.4.1",
+    "pytest-randomly",
```
cpcloud, line 49: Nit (non-blocking): this adds a duplicate pytest-randomly entry in the test dependency group.

cpcloud (changes requested):
Requesting changes for one blocking issue: please restore a local-CTK minor pin in ci/tools/run-tests for the LOCAL_CTK==1 core install path (to avoid toolkit/runtime skew). The duplicate pytest-randomly entry I noted is just a non-blocking nit.

Issue #294, opened 2024-12-13
CI: Enable nightly tests


leofang:
Steps:
1. Add another workflow triggered every night
2. Reuse the latest artifacts generated from the main branch (as seen at the time of tests)
3. Only run tests and skip the build stage (as we already have the artifacts)

Need to cover more driver versions (#293), GPU species, and number of GPUs. If we have any slow tests (maybe benchmarks from #109?), they should be run at this stage.

leofang:
To add:
- benchmarks (#407)

leofang:
Update: The infrastructure as described in the OP is set up in #1987. Keeping the issue open until most sub-tasks are done and closed.

PR #1623, opened 2026-02-13
Allow `StridedMemoryView` to be constructed from dlpacks type code

isVoid:
## Description

closes https://github.com/NVIDIA/cuda-python/issues/1624

Patched dtype handler in `_memoryview.pyx` to handle dlpack type code of kDLBfloat. Added tests to construct from an external object (currently, cupy) can communicate via dlpack bfloat16 type code.

`ml_dtypes` is added as an optional dependency, when `ml_dtypes` isn't present, this code path should gracefully raise with readable error.

Note: this still depends on ~~an unreleased~~ now released build of cupy (14.0.0). https://github.com/cupy/cupy/pull/9670

## Checklist

### DIFF
--- a/cuda_core/tests/test_utils.py
+++ b/cuda_core/tests/test_utils.py
@@ -611,6 +611,8 @@ def no_ml_dtypes(monkeypatch):
         param(StridedMemoryView.from_any_interface, id="from_any_interface"),
     ],
 )
+@pytest.mark.skipif(cp is None, reason="CuPy is not installed")
+@pytest.mark.skipif(cp is not None and _get_cupy_version_major() < 14, reason="CuPy version is less than 14.0.0")
 def test_ml_dtypes_bfloat16_dlpack_requires_ml_dtypes(init_cuda, no_ml_dtypes, api):
     a = cp.array([1, 2, 3], dtype="bfloat16")
     smv = api(a, stream_ptr=0)
--- a/cuda_core/pyproject.toml
+++ b/cuda_core/pyproject.toml
@@ -61,8 +61,8 @@ test-cu12 = [ {include-group = "ml-dtypes" }, "cuda-core[test]", "cupy-cuda12x; 
 test-cu13 = [ {include-group = "ml-dtypes" }, "cuda-core[test]", "cupy-cuda13x; python_version < '3.14'", "cuda-toolkit[cudart]==13.*"]  # runtime headers needed by CuPy
 # free threaded build, cupy doesn't support free-threaded builds yet, so avoid installing it for now
 # TODO: cupy should support free threaded builds
-test-cu12-ft = ["cuda-core[test]", "cuda-toolkit[cudart]==12.*"]
-test-cu13-ft = ["cuda-core[test]", "cuda-toolkit[cudart]==13.*"]
+test-cu12-ft = [ {include-group = "ml-dtypes" }, "cuda-core[test]", "cuda-toolkit[cudart]==12.*"]
+test-cu13-ft = [ {include-group = "ml-dtypes" }, "cuda-core[test]", "cuda-toolkit[cudart]==13.*"]
 
 [project.urls]
 homepage = "https://nvidia.github.io/cuda-python/"


cuda_core/tests/test_utils.py
```diff
@@ -524,3 +529,111 @@ def test_from_array_interface_unsupported_strides(init_cuda):
...
+
+@pytest.mark.parametrize(
+    "slices",
+    [
+        param((slice(None), slice(None)), id="contiguous"),
+        param((slice(None, None, 2), slice(1, None, 2)), id="strided"),
+    ],
+)
+@pytest.mark.skipif(torch is None, reason="PyTorch is not installed")
+def test_ml_dtypes_bfloat16_torch_dlpack(init_cuda, slices):
```
isVoid, line 579: This tests interop between pytorch tensor, it passes for me locally. But would require adding torch into test environment. Should we do that?

cuda_core/tests/test_utils.py
```diff
@@ -12,7 +12,12 @@
     from numba import cuda as numba_cuda
 except ImportError:
     numba_cuda = None
+try:
+    import torch
+except ImportError:
+    torch = None
```
cpcloud, line 19: This should really use `pytest.importorskip` inside your test. The current pattern is definitely an anti-pattern.

leofang: I think this one is fine, since it matches the style in this file (and a few other places).

cuda_core/tests/test_utils.py
leofang: No. It's too big, and needs careful design (PyTorch is very painful to test against). We should defer it to when we start working on nightly CI (https://github.com/NVIDIA/cuda-python/issues/294).

cuda_core/tests/test_utils.py
```diff
@@ -524,3 +529,111 @@ def test_from_array_interface_unsupported_strides(init_cuda):
...
+    if a.flags["C_CONTIGUOUS"]:
+        assert smv.strides in (None, strides_in_counts)
+    else:
+        assert smv.strides == strides_in_counts
+
+
+def test_ml_dtypes_bfloat16_from_any_interface_prefers_dlpack(init_cuda):
+    a = cp.array([1, 2, 3, 4, 5, 6], dtype="bfloat16")
+    wrapped = _DLPackOnlyArray(a)
+    smv = StridedMemoryView.from_any_interface(wrapped, stream_ptr=0)
```
leofang: Q: Do we really need `_DLPackOnlyArray`? I think Phillip added a `from_dlpack` constructor that would allow the same effect.

cuda_core/tests/test_utils.py
```diff
@@ -524,3 +529,111 @@ def test_from_array_interface_unsupported_strides(init_cuda):
...
+    assert smv.readonly is a.__cuda_array_interface__["data"][1]
+
+    strides_in_counts = convert_strides_to_counts(a.strides, a.dtype.itemsize)
+    if a.flags["C_CONTIGUOUS"]:
+        assert smv.strides in (None, strides_in_counts)
+    else:
+        assert smv.strides == strides_in_counts
+
+
+def test_ml_dtypes_bfloat16_from_any_interface_prefers_dlpack(init_cuda):
```
leofang: Ah OK. I think this test is unnecessary. IIRC we have a similar test somewhere that checks the preferred order, and it is dtype-agnostic by design (because we can't check the dtype if the handshake fails).

cuda_core/tests/test_utils.py
```diff
@@ -524,3 +529,111 @@ def test_from_array_interface_unsupported_strides(init_cuda):
...
+
+
+@pytest.mark.parametrize(
+    "slices",
+    [
+        param((slice(None), slice(None)), id="contiguous"),
+        param((slice(None, None, 2), slice(1, None, 2)), id="strided"),
+    ],
+)
+def test_ml_dtypes_bfloat16_dlpack(init_cuda, slices):
```
leofang, line 550: This test should add a `skipif` if CuPy<14.0.0 is in use

cuda_core/tests/test_utils.py
```diff
@@ -524,3 +529,111 @@ def test_from_array_interface_unsupported_strides(init_cuda):
...
+    strides_in_counts = convert_strides_to_counts(strides_in_bytes, a.element_size())
+    if a.is_contiguous():
+        assert smv.strides in (None, strides_in_counts)
+    else:
+        assert smv.strides == strides_in_counts
+
+
+@pytest.fixture
+def no_ml_dtypes(monkeypatch):
+    monkeypatch.setattr("cuda.core._memoryview.bfloat16", None)
```
leofang, line 603: lol... ok I did not know monkeypatching a Cython module could work... interesting

leofang:
@isVoid CuPy v14.0.0 is out and we are preparing to release cuda.core v0.6.0. Let's try to squeeze this in!

cuda_core/tests/test_utils.py
isVoid: Can probably leave a TODO here for a general refactoring.

cuda_core/tests/test_utils.py
isVoid: Makes sense. This PR should only test if bfloat16 type code works with dlpack. Not to test `from_any_interface` feature.

cuda_core/tests/test_utils.py
isVoid: Done

cuda_core/tests/test_utils.py
isVoid: It's now removed per the comment below.

cuda_core/tests/test_utils.py
cpcloud: Definitely an anti pattern to address. We shouldn't just keep accumulating them. Not blocking.

isVoid:
UPDATE: Done.

Just one more thing - I need to push some documentation addition.

PR #1625, opened 2026-02-13
Fix Buffer.from_handle(mr=...) not calling mr.deallocate()

### LINKED ISSUE #1619: [BUG]: Buffer.from_handle does not respect custom MemoryResource deallocation
### Is this a duplicate?

### Type of Bug

Silent Failure

### Component

cuda.core

### Describe the bug

Custom MemoryResource.deallocate is not called when Buffer instances created with `Buffer.from_handle(mr=mr : MemoryResource)` are destroyed.

Up to (incl.) the most recent 0.5.* release, Buffer instances created from_handle with memory resource instance specified, took ownership of the resource. I.e. when the Buffer instance was destroyed, the MemoryResource got a corresponding deallocate call. This no longer seem to be a case with the cuda-core built from the top-of main branch.

### How to Reproduce

```
from cuda.core import MemoryResource, Buffer
import numpy as np


class DummyResource(MemoryResource):
    def __init__(self):
        self.active_allocs = {}

    def allocate(self, size, stream=None, logger=None) -> Buffer:
        a = np.zeros(size, dtype=np.int8)
        ptr = a.ctypes.data
        self.active_allocs[ptr] = a
        return Buffer.from_handle(ptr=ptr, size=size, mr=self)

    def deallocate(self, ptr, size, stream=None):
        del self.active_allocs[ptr]

    @property
    def is_device_accessible(self) -> bool:
        return True

    @property
    def is_host_accessible(self) -> bool:
        return False

    @property
    def device_id(self) -> int:
        raise RuntimeError("the resource is not bound to any GPU")


resource = DummyResource()
buffer = resource.allocate(10)
assert int(buffer.handle) in resource.active_allocs
assert len(resource.active_allocs) == 1
del buffer
assert len(resource.active_allocs) == 0  # in cuda-core <=0.5.* this calls resource.deallocate, but not top-of-tree cuda-core
```

### Expected behavior

Expected behaviour is to keep what cuda-core did so far - allow user specify custom deallcation via MemoryResource deallocate method. That ownership semantics established some usage pattern for custom memory resources built on top of cuda-core. Changing Buffer.from_handle to always act as non-owining wrapper can lead to hidden bugs/leaks in code that relies on the pattern.

### Operating System

_No response_

### nvidia-smi output

_No response_

Andy-Jost:
## Summary

- Fixes #1619: `Buffer.from_handle(ptr, size, mr=mr)` now calls `mr.deallocate(ptr, size, stream)` when the buffer is closed or garbage collected, restoring the pre-RAII behavior.
- Adds `deviceptr_create_with_mr()` to the resource handle infrastructure, whose shared_ptr deleter invokes a registered callback with the pointer, size, and deallocation stream.
- The callback is implemented in Cython and converts the raw stream handle to a `cuda.core.Stream` before calling `mr.deallocate()`.
- Removes the unused `stream` parameter from `Buffer._init`.

Closes #1619

## Changes

- `resource_handles.hpp/.cpp`: Add `MRDeallocCallback` type, `register_mr_dealloc_callback()`, and `deviceptr_create_with_mr()`.
- `_resource_handles.pxd/.pyx`: Expose the new types and functions to Cython.
- `_buffer.pyx`: Register the deallocation callback at module init; use `deviceptr_create_with_mr` when `mr` is provided in `Buffer._init`; update docstrings.
- `_legacy.py`: Remove unused `stream` argument from `Buffer._init` calls.
- `tests/helpers/buffers.py`: Add `TrackingMR` test helper.
- `tests/test_memory.py`: Add `test_mr_deallocate_called_on_close` and `test_mr_deallocate_called_on_gc`.

## Test Plan

- [x] New tests verify `mr.deallocate()` is called on explicit `close()` and on GC
- [x] All existing memory and object protocol tests pass (338 passed, 15 skipped)
- [x] CI

### DIFF
--- a/ci/tools/run-tests
+++ b/ci/tools/run-tests
@@ -33,8 +33,9 @@ if [[ "${test_module}" == "pathfinder" ]]; then
       "LD:${CUDA_PATHFINDER_TEST_LOAD_NVIDIA_DYNAMIC_LIB_STRICTNESS} " \
       "FH:${CUDA_PATHFINDER_TEST_FIND_NVIDIA_HEADERS_STRICTNESS}"
   pytest -ra -s -v --durations=0 tests/ |& tee /tmp/pathfinder_test_log.txt
-  # Fail if no "INFO test_" lines are found; capture line count otherwise
-  line_count=$(grep '^INFO test_' /tmp/pathfinder_test_log.txt | wc -l)
+  # Report the number of "INFO test_" lines (including zero)
+  # to support quick validations based on GHA log archives.
+  line_count=$(awk '/^INFO test_/ {count++} END {print count+0}' /tmp/pathfinder_test_log.txt)
   echo "Number of \"INFO test_\" lines: $line_count"
   popd
 elif [[ "${test_module}" == "bindings" ]]; then
@@ -46,9 +47,9 @@ elif [[ "${test_module}" == "bindings" ]]; then
     pip install $(ls "${CUDA_BINDINGS_ARTIFACTS_DIR}"/*.whl)[all] --group test
   fi
   echo "Running bindings tests"
-  ${SANITIZER_CMD} pytest -rxXs -v --durations=0 tests/
+  ${SANITIZER_CMD} pytest -rxXs -v --durations=0 --randomly-dont-reorganize tests/
   if [[ "${SKIP_CYTHON_TEST}" == 0 ]]; then
-    ${SANITIZER_CMD} pytest -rxXs -v --durations=0 tests/cython
+    ${SANITIZER_CMD} pytest -rxXs -v --durations=0 --randomly-dont-reorganize tests/cython
   fi
   popd
 elif [[ "${test_module}" == "core" ]]; then
@@ -79,11 +80,11 @@ elif [[ "${test_module}" == "core" ]]; then
     pip install $(ls "${CUDA_CORE_ARTIFACTS_DIR}"/*.whl)["cu${TEST_CUDA_MAJOR}"] --group "test-cu${TEST_CUDA_MAJOR}${FREE_THREADING}"
   fi
   echo "Running core tests"
-  ${SANITIZER_CMD} pytest -rxXs -v --durations=0 tests/
+  ${SANITIZER_CMD} pytest -rxXs -v --durations=0 --randomly-dont-reorganize tests/
   # Currently our CI always installs the latest bindings (from either major version).
   # This is not compatible with the test requirements.
   if [[ "${SKIP_CYTHON_TEST}" == 0 ]]; then
-    ${SANITIZER_CMD} pytest -rxXs -v --durations=0 tests/cython
+    ${SANITIZER_CMD} pytest -rxXs -v --durations=0 --randomly-dont-reorganize tests/cython
   fi
   popd
 fi
--- a/cuda_bindings/cuda/bindings/_lib/utils.pxd.in
+++ b/cuda_bindings/cuda/bindings/_lib/utils.pxd.in
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2021-2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2021-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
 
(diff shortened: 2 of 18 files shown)


cuda_core/cuda/core/_memory/_buffer.pyx
```diff
@@ -73,27 +98,30 @@ cdef class Buffer:
...
         owner : object | None = None
     ):
-        """Legacy init for compatibility - creates a non-owning ref handle.
+        """Create a Buffer from a raw pointer.
 
-        Note: The stream parameter is accepted for API compatibility but is
-        ignored since non-owning refs are never freed by the handle.
+        When ``mr`` is provided, the buffer takes ownership: ``mr.deallocate()``
+        is called when the buffer is closed or garbage collected.  When ``owner``
+        is provided, the owner is kept alive but no deallocation is performed.
```
Andy-Jost, line 109: I removed the stream parameter because it was unused and this is an internal API (I think). I see the comment noting "API compatibility," but I assume this is either inaccurate or refers to the internal interface. If `Buffer._init` is a *public* API, then I wil restore it.

cuda_core/cuda/core/_memory/_buffer.pyx
```diff
@@ -37,6 +39,29 @@ from cuda.core._dlpack import DLDeviceType, make_py_capsule
...
+    const StreamHandle& h_stream,
+) noexcept:
+    """Called by the C++ deleter to deallocate via MemoryResource.deallocate."""
+    try:
+        stream = None
+        if h_stream:
+            stream = Stream._from_handle(Stream, h_stream)
+        mr.deallocate(int(ptr), size, stream)
+    except Exception:
+        pass  # Cannot propagate exceptions from a C++ destructor
```
cpcloud:
```suggestion
        import sys
        print(f"Warning: mr.deallocate() failed during Buffer destruction: {exc}",
            file=sys.stderr)
```

Just so that something is shown. I think this would mirror the Python interpreter's behavior.

Not a requirement, but would really help debugging I suspect.

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -565,6 +565,44 @@ DevicePtrHandle deviceptr_create_with_owner(CUdeviceptr ptr, PyObject* owner) {
...
+        [mr, size](DevicePtrBox* b) {
+            GILAcquireGuard gil;
+            if (gil.acquired() && mr_dealloc_cb) {
+                mr_dealloc_cb(mr, b->resource, size, b->h_stream);
+            }
+            // Decref mr even if callback was skipped (prevent leak when
+            // callback is not registered -- should not happen in practice).
+            if (gil.acquired()) {
+                Py_DECREF(mr);
+            }
```
cpcloud: Seems like these branches should be merged into something like:


```suggestion
            if (gil.acquired()) {
                if (mr_dealloc_cb) {
                    mr_dealloc_cb(mr, b->resource, size, b->h_stream);
                }
                
                // Decref mr regardless of whether the callback ran, as long as
                // we have the GIL. Leak is acceptable during interpreter finalization and shouldn't happen in practice.
                Py_DECREF(mr);
            }
            delete b;
```

cuda_core/tests/test_memory.py
```diff
@@ -471,6 +471,31 @@ def test_buffer_external_managed(change_device):
...
+    device = Device()
+    device.set_current()
+    mr = TrackingMR()
+    buf = mr.allocate(1024)
+    assert len(mr.active) == 1
+    del buf
+    gc.collect()
+    assert len(mr.active) == 0
+
+
```
cpcloud, line 498: The two new tests cover `close()` (no stream) and GC, but neither exercises the path where `close(stream=s)` forwards the stream through the C++ deleter to `mr.deallocate(ptr, size, stream)`. A third test that verifies the stream argument arrives at `mr.deallocate()` would close that gap, e.g.:

```python
def test_mr_deallocate_receives_stream():
    device = Device()
    device.set_current()
    received = {}
    class StreamCaptureMR(TrackingMR):
        def deallocate(self, ptr, size, stream=None):
            received['stream'] = stream
            super().deallocate(ptr, size, stream)
    mr = StreamCaptureMR()
    buf = mr.allocate(1024)
    buf.close(stream)
    assert received['stream'] is not None
```

cuda_core/cuda/core/_memory/_buffer.pyx
cpcloud: Probably a hallucination or I was reviewing older code, thanks!

PR #1626, opened 2026-02-13
Add one-click release workflow for cuda-pathfinder

cpcloud:
Replaces the manual release checklist for cuda-pathfinder with a single
`workflow_dispatch` that takes a version number.

## Pipeline

`prepare` → `docs` → `publish-testpypi` → `verify-testpypi` → `publish-pypi` → `verify-pypi` → `finalize`

`upload-assets` runs in parallel with the publish chain (attached to the draft
release with `--clobber`).

The verify jobs install the package in a fresh venv and assert `__version__`
matches -- replacing the manual `pip install` + import step. If TestPyPI
verification fails, PyPI publish never runs.

## Notable choices

- Refuses to run from non-default branches (prevents accidental tags from feature branches)
- Fails on duplicate PyPI uploads rather than silently skipping (unexpected state should be investigated)
- Uses existing `ci/tools/lookup-run-id` and `ci/tools/download-wheels` unmodified
- All `workflow_dispatch` inputs flow through `env:` to prevent shell injection

### DIFF
--- a/.github/workflows/release-cuda-pathfinder.yml
+++ b/.github/workflows/release-cuda-pathfinder.yml
@@ -1,0 +1,441 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# One-click release workflow for cuda-pathfinder.
+#
+# Provide a version number and commit SHA.  The workflow automatically finds
+# the CI run and creates the git tag, creates a draft GitHub release with the standard
+# body, builds versioned docs, uploads source archive + wheels to the
+# release, publishes to TestPyPI, verifies the install, publishes to PyPI,
+# verifies again, and finally marks the release as published.
+
+name: "Release: cuda-pathfinder"
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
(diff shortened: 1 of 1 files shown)


cpcloud:
Good point. I'll add that as a required input for now. I wonder if there's a way to make this work. We can of course alter our dev workflow, which might be the only way to reasonably achieve full automation here.

cpcloud:
@mdboom Thanks for the callout - this concern was valid.

I updated the release workflow to require an explicit `commit` input and pin the release tag to that exact SHA. The workflow now validates that the SHA is a full commit hash, exists, and is reachable from the default branch. It also fails if the target tag already exists at a different commit.

So merges to `main` while the workflow is running no longer change what commit is being released.

cpcloud:
I think if we're going for `a`/`b` releases, we should just publish them as prereleases on regular PyPI. This feels like a thing that a subsequent PR can tackle, if that's alright.

Would be nice to try this out for a vanilla non-alpha/beta release before adding that.

cpcloud:
> To help the LLMs, it'd be good to add comments in release.yml and the new file, e.g.
> 
> ```
> # ATTENTION: If you make changes in this file, also review <the other>.
> ```

This kind of thing is 100% the kind of thing that goes in a well-crafted AGENTS.md. I'd rather put that there, since most agents except for Claude are guaranteed to read it on startup.

PR #1628, opened 2026-02-14
refactor(tests): replace exec() with importlib in example runner

OfficialSerge:
## Description

This PR is in response to the following TODO item in `cuda_core/tests/example_tests/utils.py`.
```python
exec(script, env if env else {})
```

## Key Changes
- refactor `run_example` in `utils.py` using `importlib`, adding support for *both* `main()` and `exec()` scripts, tracebacks should now provide real line number feedback, thus paving the way for us to progressively enhance the remaining non-module tests and examples.
- updated `test_basic_example.py` to use `pathlib.Path` instead of `os.path.join()`. 
- removed redundant manual `Device(0)` and `set_current()` calls in test class, instead relying on exiting `deinit_cuda()` fixture for consistent teardowns (see `conftest.py`)

## Next Steps
- go through the remaining examples still without `main()` and refactor them into Python modules, some examples are currently refactored while others are still plain executables so it's a mix.


cuda_core/tests/example_tests/test_basic_examples.py
```diff
@@ -3,21 +3,22 @@
...
 
 import pytest
-from cuda.core import Device
 
 from .utils import run_example
 
-samples_path = os.path.join(os.path.dirname(__file__), "..", "..", "examples")
-sample_files = glob.glob(samples_path + "**/*.py", recursive=True)
+# not dividing, but navigating into the "examples" directory.
+EXAMPLES_DIR = Path(__file__).resolve().parent.parent.parent / "examples"
```
mdboom:
```suggestion
EXAMPLES_DIR = Path(__file__).resolve().parents[2] / "examples"
```

cuda_core/tests/example_tests/utils.py
```diff
@@ -12,24 +13,38 @@ class SampleTestError(Exception):
...
-        sys.path.append(samples_path)
-        # TODO: Refactor the examples to give them a common callable `main()` to avoid needing to use exec here?
-        exec(script, env if env else {})  # noqa: S102
+        sys.path.append(parent_dir)
+        sys.argv = [str(fullpath)]
+
+        # Collect metadata for file 'module_name' located at 'fullpath'.
+        # CASE: file does not exist -> spec is none.
+        # CASE: file is not .py -> spec is none.
+        # CASE: file does not have proper loader (module.spec.__loader__) -> spec.loader is none.
```
mdboom: I don't think we necessarily need to duplicate information from the Python stdlib docs here, but don't feel strongly either way.

mdboom (approved):

cuda_core/tests/example_tests/utils.py
OfficialSerge: I see, my ruff linter was flagging it as a potential point of runtime failure, the if-none check fixed these issues though I figured I'd leave an explanation to help people understand the need for the check, in case there are any future revisions.

cuda_core/tests/example_tests/test_basic_examples.py
```diff
@@ -3,21 +3,22 @@
...
+SAMPLE_FILES = sorted([str(p.relative_to(EXAMPLES_DIR)) for p in EXAMPLES_DIR.glob("**/*.py")])
 
-@pytest.mark.parametrize("example", sample_files)
+
+@pytest.mark.parametrize("example_rel_path", SAMPLE_FILES)
 class TestExamples:
-    def test_example(self, example, deinit_cuda):
-        run_example(samples_path, example)
-        if Device().device_id != 0:
-            Device(0).set_current()
```
leofang: We should kick off CI and check if removing this line is OK. IIRC on a multi-GPU system it would fail without this.

cuda_core/tests/example_tests/utils.py
leofang: This seems a bit more complicated than necessary...? Assuming we always have main (#1664), can we simplify the logic here after merging the other PR?

OfficialSerge: Yes, technically all we need once the examples are refactored to Python modules is the following
```python
spec = importlib.util.spec_from_file_location(module_name, fullpath)
module = importlib.util.module_from_spec(spec)
sys.modules[module_name] = module

spec.loader.exec_module(module)
module.main()
```
### There are 2 possible ways example(s) code may be invoked

*via a user:* In this case, `__name__` does equal `__main__` and we execute the code in the if statement and the subsequent example.

*via a script:* In this case `__name__` is equal to the module stem i.e `vector_add`, thus what is in the if statement doesn't run, calling `module.main()` here is what saves us and actually runs the example code.

cuda_core/tests/example_tests/test_basic_examples.py
OfficialSerge: Okay, I think I get this, let me know if I mess up:

- each thread gets only a single reference to any one Device (hence thread-local singleton pattern)
- a thread can reference multiple Devices
- a Device can have multiple CUDA Contexts but a Context can only belong to a single GPU Device
- Contexts on the same Device are mutually exclusive
- the driver manages the context stack for a given thread

So, in a multi-GPU example, the driver recruits n devices using a given thread, then runs the kernel and calls `deinit_cuda()`, popping the context off the context stack. 

### The Problem
The driver doesn't update the `current_device` to 0 when popping multiple shared (cudaDeviceEnablePeerAccess) Device contexts, thus when a program asks for a new Device the driver returns the nth device instead of the 0th.

### A Possible Solution
Redundantly set `Device(0)` as the current device prior to running the example, if prior example was multi-GPU, we are now back to Device 0, otherwise the redundant call does nothing.
```python
def test_example(self, example_rel_path: str, deinit_cuda) -> None:
    from cuda.core import Device

    Device(0).set_current()
    run_example(str(EXAMPLES_DIR), example_rel_path)
```

leofang:
@OfficialSerge Sorry we dropped the ball. I suspect a recent PR has made this PR obsolete (and caused merge conflicts)? Could you check?

OfficialSerge:
Yes, it seems like the refactoring of the tests is already reflected in main, that specific bit is now redundant (I've closed it), to reiterate it wrapped examples in `main()` functions and thus allowed for them to be imported and ran as modules instead of executables.

OfficialSerge:
@leofang consider using the importlib implementation over the subprocess-based approach? I believe it to be faster and less resource hungry than spawning a sub-process for each and every test. I can spearhead a case for importlib if you'd like.

leofang:
Thank you @OfficialSerge for your contribution. I think as noted earlier the recent move to subprocess has made this PR obsolete. I personally also prefer to run examples in the same test process, but I don't plan to revert the team decision, also I don't think `importlib` is technically better. Would it be OK if we just close this PR?

OfficialSerge:
@leofang that works fine for me, sorry I couldn't be more helpful I was newly hired as a SWE whilst in the middle of working on this. I will follow the repo to see what more I can do at some point.

mdboom:
Looking at this again with fresh eyes -- the big change here is that we used to run samples in a `subprocess` and now run them in process.  I think that's the wrong move for examples, which are designed to run in their own process.  It's fine for them to have side effects and be non-thread-safe.  We also want to test that the work from a cold process startup.

So given all that, I think I'm just going to close this.

PR #1632, opened 2026-02-17
Fix bare except in Stream_accept that swallows all exceptions

### LINKED ISSUE #1631: Stream_accept uses bare except that swallows all exceptions
## Description

`Stream_accept` in `_stream.pyx` uses a bare `except: pass` when probing whether a foreign object supports the `__cuda_stream__` protocol. This catches all exceptions, including `KeyboardInterrupt`, `SystemExit`, and `RuntimeError` from malformed protocol implementations.

## Current behavior

```python
try:
    stream = Stream._init(arg)
except:
    pass
```

Any exception raised during `Stream._init` is silently swallowed, including:
- `RuntimeError` from a buggy `__cuda_stream__` implementation (e.g., wrong return format)
- `TypeError` from internal failures within `__cuda_stream__()`
- `KeyboardInterrupt`, `SystemExit`

## Expected behavior

Only the specific `TypeError` indicating the object lacks `__cuda_stream__` should be caught. All other exceptions should propagate so that protocol implementation bugs are surfaced.

## Proposed fix

Narrow the catch to `except TypeError as e` with a string check for `"__cuda_stream__"` in the message, which matches the specific `TypeError` raised by `_handle_from_stream_protocol` when the attribute is missing.

Andy-Jost:
## Summary

- Narrows bare `except: pass` in `Stream_accept` to `except TypeError` with a string check for `"__cuda_stream__"`, so only the "protocol not supported" case is silently handled
- Buggy `__cuda_stream__` implementations, `RuntimeError`, `KeyboardInterrupt`, `SystemExit`, etc. now propagate correctly

Closes #1631

## Changes

- `_stream.pyx`: Replace `except: pass` with `except TypeError as e: if "__cuda_stream__" not in str(e): raise`

## Test Coverage

- Existing `Stream_accept` and stream protocol tests cover the happy path
- The change is a strict narrowing of the catch clause; no new behavior is introduced for well-formed objects


Andy-Jost:
I found this issue as a follow-on to [this comment](https://github.com/NVIDIA/cuda-python/pull/1625#discussion_r2817844494) while searching for other places where `try...except` was used to swallow exceptions too broadly.

cuda_core/cuda/core/_stream.pyx
```diff
@@ -509,8 +509,9 @@ cdef Stream Stream_accept(arg, bint allow_stream_protocol=False):
     elif allow_stream_protocol:
         try:
             stream = Stream._init(arg)
-        except:
-            pass
+        except TypeError as e:
+            if "__cuda_stream__" not in str(e):
+                raise
```
Andy-Jost: cf. https://github.com/NVIDIA/cuda-python/blob/a1614ed1f35116a3cbfe4790c78262b74b08c3a1/cuda_core/cuda/core/_stream.pyx#L474-L477

leofang:
@Andy-Jost Just to confirm, this PR can wait after we release v0.6.0, or do you want this to be part of v0.6.0 as well?

Andy-Jost:
This is not a critical fix, so I put it under v0.7.0. But I think it is very safe, so if it gets approved in time I will include it with v0.6.0.

cuda_core/cuda/core/_stream.pyx
cpcloud: Why are we swallowing an `AttributeError`, just to tell the user "The thing doesn't have this attribute"? This seems super roundabout, and adding a string contains check to this oddity doesn't seem like the right approach.

Andy-Jost: I think the overall result is correct. If someone passes an arbitrary object to the stream protocol and that object lacks a `__cuda_stream__` attribute, then raising `TypeError` is the correct response. This matches standard protocols such as `iter()`, `len()`, and `hash()`.

That said, I think you're right that it would be arguably cleaner to use `hasattr` at line 474 rather than catch the `AttributeError`, though that would mean checking `hasattr` and then calling the equivalent of `getattr` separately.

The problem is that Python does not provide a clean method of detecting when an object lacks a specific protocol, and adding something custom here, like `ProtocolNotSupportedError`, would be even more out of the norm.

I don't see a better option than catching the `TypeError` and narrowing it with this heuristic to make sure it looks like the one we expect.

Andy-Jost: Actually, now that I say that, perhaps `Stream_accept` near line 512 could just use `hasattr` to check for `__cuda_stream__`. But that raises its own complications, since `Stream._init` has many paths besides the stream protocol one.

Andy-Jost: I think the latest upload is probably what you had in mind.

Andy-Jost:
@cpcloud Does the latest upload address your concerns?

PR #1633, opened 2026-02-17
[doc-only] Update LICENSE.md

jollylili:
Call out cuda.pathfinder as Apache 2.0


rwgk (approved):

rwgk:
Thanks @jollylili, I'll go ahead with merging this change; I think this doesn't need Keith's attention.

PR #1636, opened 2026-02-18
WIP: Add `cuda.core.utils.make_aligned_dtype`

### LINKED ISSUE #734: Add `cuda.core.utils.get_aligned_dtype`
Many CUDA Python projects require the ability to declare a custom alignment for a give dtype. While NumPy builtin/structure dtypes work in most cases, this is one of the few issues that it cannot fulfill today. @seberg has figured out a prototype to address this issue (https://github.com/numpy/numpy/issues/28318#issuecomment-2678883667), and the plan is for cuda.core to provide this infrastructure, let other teams start using it, gather feedbacks, and eventually upstream this back to NumPy.

leofang:
## Description

closes #734.

Credit goes entirely to @seberg (see #734 and https://github.com/cupy/cupy/pull/9650) 🙂 

DO NOT REVIEW yet. Need to address some details.

## Checklist

### DIFF
--- a/cuda_core/cuda/core/_utils/dtype_utils.pyx
+++ b/cuda_core/cuda/core/_utils/dtype_utils.pyx
@@ -1,0 +1,183 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import numpy
+
+
+cdef Py_ssize_t _get_cuda_scalar_alignment(object descr):
+    """Return the CUDA alignment for a scalar (non-structured) NumPy dtype.
+
+    Uses standard C/CUDA alignment rules: alignment equals the itemsize
+    for types up to 16 bytes, capped at 16.  The result is always a
+    power of two.
+    """
+    cdef Py_ssize_t size = descr.itemsize
+    # Cap at 16 (e.g. complex128 is 16 bytes, no CUDA type exceeds 16-byte alignment)
+    if size > 16:
+        return 16
+    # Round down to nearest power of 2
+    cdef Py_ssize_t alignment = 1
+    while alignment * 2 <= size:
+        alignment *= 2
+    return alignment
+
+
+def make_aligned_dtype(dtype, *, int alignment=-1, int recurse=False):
+    """Create a new structured dtype with sufficient alignment for GPU use.
+
+    Many CUDA kernels require structure members to be naturally aligned
+    (each field's offset is a multiple of its own alignment) and the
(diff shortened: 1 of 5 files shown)


cuda_core/cuda/core/_utils/dtype_utils.pyx
```diff
@@ -0,0 +1,183 @@
...
+
+            if subdtype.names is None:
+                subalignment = _get_cuda_scalar_alignment(subdtype)
+            elif not recurse:
+                subalignment = subdtype.alignment
+                if subdtype.metadata:
+                    subalignment = subdtype.metadata.get(
+                        "__cuda_alignment__", subalignment)
+            else:
+                subdtype = make_aligned_dtype(subdtype, recurse=recurse)
```
cpcloud, line 138: When `recurse=True`, we recurse into nested structured dtypes before honoring existing `__cuda_alignment__` metadata. That can reject already-prepared nested dtypes (with intentionally padded `itemsize`) via `Input descriptor had larger itemsize than inferred.`

Could we mirror the non-recursive path and first respect `__cuda_alignment__` if present, then recurse only when metadata is absent?

cuda_core/cuda/core/_utils/dtype_utils.pyx
```diff
@@ -0,0 +1,183 @@
...
+            offset = field_info[1]
+
+            if offset < min_offset:
+                raise ValueError(
+                    "make_aligned_dtype() only supports well-behaved "
+                    "in-order fields (it ignores field offsets).")
+
+            min_offset = offset + subdtype.itemsize
+
+            if subdtype.names is None:
```
cpcloud, line 130: `subdtype.names is None` also covers subarray / unstructured-void fields. `_get_cuda_scalar_alignment()` then infers alignment from total `itemsize`, which can over-align (e.g. `('f4', 3)` has size 12 but element alignment 4).

That can silently change field offsets and ABI layout. Should we detect subarrays (`subdtype.subdtype`) and align from the base dtype (or explicitly reject subarrays until supported)?

cuda_core/cuda/core/_utils/dtype_utils.pyx
seberg: Yeah, I changed this pattern slightly in the cupy PR (I suspect that addresses this).

cuda_core/cuda/core/_utils/dtype_utils.pyx
seberg: I'll fix this in the cupy PR.  (I may have forgotten about it, because the rest of CuPy in that PR doesn't support subarrays.)

leofang:
Pushing this out...

PR #1647, opened 2026-02-18
Fix #1638: Faster returning of results

### LINKED ISSUE #1638: [PERF]: Improve speed of returning the result pair
In the driver, runtime and nvrtc modules, returning the result of an API call usually follows this pattern:

```python
    if err != cydriver.CUDA_SUCCESS:
        return (CUresult(err), None) 
    return (CUresult(err), result)
```

`CUresult(err)` actually performs two dictionary lookups in large dictionaries.  One is to find `CUresult` in the global dictionary of the module (`cdef classes` are actually Python objects so don't have a direct reference), and then to look up the enum value/name pair by the integer within the `CUresult` enum itself.

We can store `CUresult` in a `cdef` variable to skip the first lookup.  We can also store `CUresult.CUDA_SUCCESS` in a `cdef` variable to make the "happy path" faster.

Preliminary measurements show this removes about 20% of the overhead (300us on my machine) on the #659 benchmark.

# Non-backward compatible change

This approach still has the overhead of packing and unpacking the tuple everytime.   We could instead do:

```python
    if err != cydriver.CUDA_SUCCESS:
        raise Exception(...)
    return result
```

This would reduce the happy path to a simple C integer comparison and incref'ing the result.  Unfortunately, this isn't backward compatible, and means our users would have to change their code from:

```python
    err, result = driver.cuMyApiCall(...)
```

to:

```python
    result = driver.cuMyApiCall(...)

    try:
        result = driver.cuMyApiCall(...)
    except Exception as e:
        ...
```

We already follow this pattern in libraries other than driver, runtime and nvrtc, so it may be worth doing this for consistency reasons as well as performance reasons.

mdboom:
See #1638 for details as to why this works.

On my machine with the #659 benchmark, I see a reduction from 3.02us to 2.54us per iteration.

### DIFF
--- a/cuda_bindings/cuda/bindings/_lib/utils.pxd.in
+++ b/cuda_bindings/cuda/bindings/_lib/utils.pxd.in
@@ -5,6 +5,7 @@ cimport cuda.bindings.driver as driver
 cimport cuda.bindings.cydriver as cydriver
 cimport cuda.bindings.cyruntime as cyruntime
 from libcpp.vector cimport vector
+from cpython.buffer cimport PyBuffer_Release, Py_buffer
 
 cdef class _HelperKernelParams:
     cdef Py_buffer _pybuffer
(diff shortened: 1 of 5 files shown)


leofang:
@mdboom sorry for noticing this late, but I am confused by the diff -- it seems like this PR just fixes a recently-introduced (I think) regression? We've fixed this enum issue long time ago by a fast dict lookup (#546), but that fix seems to be gone (and replaced by a presumably even-faster memoization in this PR).

leofang:
> it seems like this PR just fixes a recently-introduced (I think) regression?

Ah, I see. The "regression" was introduced in the fast enum refactoring (https://github.com/NVIDIA/cuda-python/pull/1581). So with this PR it means regardless of how fast the enum implementation (builtin or custom) is the memoization is always needed... 🙂

PR #1648, opened 2026-02-18
[PERF]: Use __new__ to directly instantiate cdef classes

### LINKED ISSUE #1643: [PERF] Construction of known `cdef classes` can go through `__new__`
In driver, runtime and nvrtc, when an instance of a `cdef class` is contructed (usually as a return value), it just uses the obvious Python-level constructor, for example:

```
cdef CUtensorMap tensorMap = CUtensorMap()
```

This makes a slow Python-like call (the ironically named `PyObject_FastCall`).

Instead, following Cythons [fast instantiation](https://cython.readthedocs.io/en/latest/src/userguide/extension_types.html#fast-instantiation) advice, we can use `__new__`, which will make a faster C-level call to `__cinit__`.

```
cdef CUtensorMap tensorMap = CUtensorMap.__new__(CUtensorMap)
```

mdboom:
Fixes #1643.  This follows the advice in [the Cython docs](https://cython.readthedocs.io/en/latest/src/userguide/extension_types.html#fast-instantiation) for directly instantiating a `cdef class` Python object.  This avoids a dictionary lookup and a Python-calling-convention call to the constructor.

### DIFF
--- a/cuda_bindings/cuda/bindings/driver.pyx.in
+++ b/cuda_bindings/cuda/bindings/driver.pyx.in
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2021-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
 
-# This code was automatically generated with version 13.1.0. Do not modify it directly.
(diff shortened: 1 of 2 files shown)


cpcloud (approved):
... it's so ugly :(

leofang:
Q: We don't want this anymore? Seems fine to me.

leofang:
Ah, I see https://github.com/NVIDIA/cuda-python/issues/1643#issuecomment-3923076958.

PR #1652, opened 2026-02-19
Refactor cuda-bindings setup.py to defer CUDA_HOME check to build time

### LINKED ISSUE #1635: Refactor cuda-bindings setup.py to defer CUDA_HOME check to build time
## Problem

`cuda_bindings/setup.py` requires `CUDA_HOME`/`CUDA_PATH` at module import time (line 29-31) and imports Cython/pyclibrary at the top level (lines 16-17). This means any tool that needs to read package metadata — including `uv lock`, `uv sync`, and `pip install --dry-run` — must have the CUDA Toolkit installed and `CUDA_HOME` set, even when no actual build is happening.

This creates friction with modern Python tooling (uv workspaces, lockfile generation, metadata introspection) and prevents patterns like `uv sync --package cuda-bindings --group test` from working without a full CUDA environment.

## Proposed solution

Convert `cuda_bindings` to use a custom PEP 517 build backend (similar to what `cuda_core` already does with `build_hooks.py`):

1. **Defer `CUDA_HOME` check** — Move it from module-level into the actual build functions (`build_wheel`, `build_editable`). `get_requires_for_build_wheel` should not need CUDA.
2. **Defer heavy imports** — `Cython`, `pyclibrary`, and other build-only imports should happen inside build functions, not at module import.
3. **Static metadata in `pyproject.toml`** — Anything that can be declared statically (package name, deps, classifiers) should stay in `pyproject.toml`.
4. **Custom build backend** — A `build_hooks.py` (or similar) that wraps `setuptools.build_meta` and only invokes CUDA-dependent logic when actually building.

## Context

This came up during the uv migration (#1615). Currently `uv lock` requires `CUDA_HOME=/usr/local/cuda` as a workaround. The `cuda_core` package already follows the deferred pattern via its `build_hooks.py`.

## References

- `cuda_core/build_hooks.py` — existing example of the deferred pattern
- `cuda_bindings/setup.py` — current implementation (~400 lines)
- #1615 — uv workspace migration PR where this friction was encountered

kkraus14:
## Summary

- Introduces `cuda_bindings/build_hooks.py` as a custom PEP 517 build backend that defers all CUDA-dependent logic (header parsing, code generation, cythonization) to `build_wheel`/`build_editable`
- Rewrites `cuda_bindings/setup.py` as a thin shim (matching `cuda_core/setup.py`) that only provides the `ext_modules` bridge and parallel `build_ext`
- Changes `pyproject.toml` build-backend from `setuptools.build_meta` to `build_hooks`

Metadata queries (`uv lock`, `uv sync`, `pip install --dry-run`) no longer require `CUDA_HOME` or `CUDA_PATH` to be set.

Follows the same pattern already used by `cuda_core/build_hooks.py`.

## Test plan

- [ ] Verify `python -c "import build_hooks"` succeeds without `CUDA_HOME` (confirmed locally)
- [ ] Verify `uv lock` / `uv pip compile` works without `CUDA_HOME`
- [ ] Verify `CUDA_HOME=/usr/local/cuda uv build --wheel` produces a working wheel
- [ ] Verify `pip install -e .` with `CUDA_HOME` set works for editable installs
- [ ] CI wheel builds pass on Linux and Windows

Closes #1635

### DIFF
--- a/cuda_bindings/setup.py
+++ b/cuda_bindings/setup.py
@@ -40,12 +40,10 @@ class build_ext(_build_ext):
     def build_extensions(self):
         if nthreads > 0:
             self.parallel = nthreads
-        super().build_extensions()
-
-    def build_extension(self, ext):
         if _is_clang(self.compiler):
-            ext.extra_compile_args = [a for a in ext.extra_compile_args if a != "-fno-var-tracking-assignments"]
-        super().build_extension(ext)
+            for ext in self.extensions:
+                ext.extra_compile_args = [a for a in ext.extra_compile_args if a != "-fno-var-tracking-assignments"]
+        super().build_extensions()
 
 
 setup(


cuda_bindings/build_hooks.py
```diff
@@ -0,0 +1,433 @@
...
+        (["cuda/bindings/_internal/utils.pyx"], None),
+        *(([f], None) for f in dst_files if f.endswith(".pyx")),
+    ]
+
+    for sources, libraries in sources_list:
+        extensions += _prep_extensions(sources, libraries, include_dirs, library_dirs, extra_compile_args)
+
+    if strip and sys.platform == "linux" and "--debug" not in sys.argv:
+        for ext in extensions:
+            ext.extra_link_args.append("-Wl,--strip-all")
```
mdboom: It would probably be less confusing to handle `extra_link_args` in parallel with `extra_compile_args`, (there is already a "non debug build" path above) rather than handling it separately here.

cuda_bindings/build_hooks.py
```diff
@@ -0,0 +1,433 @@
...
+
+    # Cythonize
+    compiler_directives = dict(language_level=3, embedsignature=True, binding=True, freethreading_compatible=True)
```
mdboom: This name was less confusing in its old location.  But now since it's in a function doing a lot more things, maybe change it to `cython_directives`?

cuda_bindings/build_hooks.py
kkraus14: Addressed in https://github.com/NVIDIA/cuda-python/pull/1652/commits/be4dedc422108400f414968dbb4d018c80714383

cuda_bindings/build_hooks.py
kkraus14: Addressed in https://github.com/NVIDIA/cuda-python/pull/1652/commits/be4dedc422108400f414968dbb4d018c80714383

cuda_bindings/setup.py
```diff
@@ -406,42 +33,23 @@ def _is_clang(compiler_cxx):
...
-class ParallelBuildExtensions(build_ext):
-    def initialize_options(self):
-        super().initialize_options()
+class build_ext(_build_ext):
+    def build_extensions(self):
         if nthreads > 0:
             self.parallel = nthreads
+        if _is_clang(self.compiler):
+            for ext in self.extensions:
+                ext.extra_compile_args = [a for a in ext.extra_compile_args if a != "-fno-var-tracking-assignments"]
```
kkraus14, line 45: @mdboom I moved this so we only check this once instead of per extension. Would be great if you could take a look in case I have the wrong idea.

mdboom: Yes, this should be fine.  It would be nicer to do it where we set the flags in the first place, but we don't know what the compiler is yet at that point.

Issue #1634, opened 2026-02-18
cuda.core v0.6.0


leofang:
### Tasks for cuda-core and cuda-bindings patch release

- [x] File an internal nvbug to communicate test plan & release schedule with QA
- [x] Ensure all pending PRs are reviewed, tested, and merged
- [x] Check (or update if needed) the dependency requirements
- [ ] ~~Create a public rc tag~~
- [ ] ~~Point QA to fetch public artifacts (wheels) from the GHA run ID, example: `gh run download 12323257563 -p "cuda-core*" -R NVIDIA/cuda-python`~~
- [ ] ~~Wait for QA reports and fix any issues found~~
- [x] Finalize the doc update, including release notes ("Note: Touching docstrings/type annotations in code is OK during code freeze, apply your best judgement!")
- [x] Update the docs for the new version
- [x] Create a public release tag
- [x] If any code change happens, rebuild the wheels from the new tag
- [x] Update the conda recipe & release conda packages
- [ ] ~~Upload conda packages to nvidia channel~~
- [x] Upload wheels to PyPI
- [ ] Post-release QA
- [x] Finalize the announcement update
- [x] Send out the announcement internally
- [x] Send out the announcement externally (GitHub Release -> Announcement)

Andy-Jost:
Tracking bug: 5910741

Andy-Jost:
@leofang I think this is complete, except for post-release QA and the external announcement, which I don't know how to handle. Can you help disposition this issue?

rparolin:
@Andy-Jost Can we close this out? So I can close out the 0.6 release milestone?

leofang:
@rparolin I unchecked the last three items and stroked out the last one (external announcement, which we haven't done for quite some time). For drafting/sending the internal announcement, please work with @jollylili.

PR #1660, opened 2026-02-19
Fix issues discovered during v0.6.0 integration testing

Andy-Jost:
## Summary

- Restores `_handle` as a property on `Context` and `Kernel` that delegates to the public `.handle` property, providing backward compatibility for downstream projects (e.g., [numba-cuda](https://github.com/NVIDIA/numba-cuda/pull/794)) that relied on the private attribute removed during the resource handles refactoring
- Fixes pickle support for Cython `cdef class` types (`ObjectCode`, `IPCEventDescriptor`, `IPCBufferDescriptor`, `Buffer`) where `@classmethod` callables in `__reduce__` fail pickle's identity check
- Disables `cuLibraryUnload` in `LibraryBox` deleters to prevent potential segfaults when the owning context is destroyed before the library (to be re-enabled once `LibraryBox` tracks its owning context)
- Adds `0.6.0` to the docs version switcher (`nv-versions.json`)
- Guards `ml_dtypes` import in `test_utils.py` as an optional dependency

## Changes

- `_context.pyx`: Add `_handle` property that returns `self.handle`
- `_module.pyx`: Add `_handle` property on `Kernel` that returns `self.handle`; convert `ObjectCode._reduce_helper` from `@classmethod` to `@staticmethod`
- `_event.pyx`: Convert `IPCEventDescriptor._init` from `@classmethod` to `@staticmethod`
- `_memory/_buffer.pyx`: Add `@staticmethod _reduce_helper` for `Buffer.__reduce__`
- `_memory/_ipc.pyx`: Convert `IPCBufferDescriptor._init` from `@classmethod` to `@staticmethod`
- `_cpp/resource_handles.cpp`: Comment out `cuLibraryUnload` calls in `LibraryBox` deleters
- `docs/nv-versions.json`: Add `0.6.0` entry
- `tests/test_utils.py`: Guard `ml_dtypes` import with try/except and add skip markers

## Test Coverage

- Pickle round-trip for `ObjectCode` verified manually
- `_handle` property delegates to existing `.handle` which is already tested
- numba-cuda's existing tests exercise the `_handle` path

## Related Work

- numba-cuda PR: https://github.com/NVIDIA/numba-cuda/pull/794
- Part of v0.6.0 release: #1634

### DIFF
--- a/cuda_core/cuda/core/_cpp/resource_handles.cpp
+++ b/cuda_core/cuda/core/_cpp/resource_handles.cpp
@@ -750,7 +750,8 @@ LibraryHandle create_library_handle_from_file(const char* path) {
         new LibraryBox{library},
         [](const LibraryBox* b) {
             GILReleaseGuard gil;
-            p_cuLibraryUnload(b->resource);
+            // TODO: re-enable once LibraryBox tracks its owning context
+            // p_cuLibraryUnload(b->resource);
             delete b;
         }
     );
@@ -768,7 +769,8 @@ LibraryHandle create_library_handle_from_data(const void* data) {
         new LibraryBox{library},
         [](const LibraryBox* b) {
             GILReleaseGuard gil;
-            p_cuLibraryUnload(b->resource);
+            // TODO: re-enable once LibraryBox tracks its owning context
+            // p_cuLibraryUnload(b->resource);
             delete b;
         }
     );
--- a/cuda_core/cuda/core/_event.pyx
+++ b/cuda_core/cuda/core/_event.pyx
@@ -295,9 +295,9 @@ cdef class IPCEventDescriptor:
     def __init__(self, *arg, **kwargs):
         raise RuntimeError("IPCEventDescriptor objects cannot be instantiated directly. Please use Event APIs.")
 
-    @classmethod
-    def _init(cls, reserved: bytes, busy_waited: cython.bint):
-        cdef IPCEventDescriptor self = IPCEventDescriptor.__new__(cls)
+    @staticmethod
+    def _init(reserved: bytes, busy_waited: cython.bint):
+        cdef IPCEventDescriptor self = IPCEventDescriptor.__new__(IPCEventDescriptor)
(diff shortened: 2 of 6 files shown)


Andy-Jost:
cc @brandon-b-miller — this restores the private `_handle` attributes so numba-cuda doesn't need version-gated access.

cuda_core/cuda/core/_context.pyx
```diff
@@ -39,6 +39,10 @@ cdef class Context:
             return None
         return as_py(self._h_context)
 
+    @property
+    def _handle(self):
```
mdboom, line 43: Does `_handle = handle` work here?  (It might not in Cython).

Andy-Jost:
@mdboom Tried it — Cython rejects `_handle = handle` in a `cdef` class (compile error). Properties in `cdef` classes are implemented as C-level `PyGetSetDef` entries, not regular Python descriptors, so the alias trick doesn't work.

cuda_core/cuda/core/_context.pyx
Andy-Jost: Cython does not support this (see https://github.com/NVIDIA/cuda-python/pull/1660#issuecomment-3929323546)

leofang:
Brandon will be verifying with numba-cuda locally.

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -750,7 +750,8 @@ LibraryHandle create_library_handle_from_file(const char* path) {
         new LibraryBox{library},
         [](const LibraryBox* b) {
             GILReleaseGuard gil;
-            p_cuLibraryUnload(b->resource);
+            // TODO: re-enable once LibraryBox tracks its owning context
+            // p_cuLibraryUnload(b->resource);
```
leofang, line 754: Note for posterity: This restores the old behavior since cuda.core 0.1.0 (this was changed when module/library was cythonized).

PR #1661, opened 2026-02-19
Update NVML bindings to use cuda_pathfinder

mdboom:


kkraus14:
@mdboom do we need to bump the minimum version of `cuda.pathfinder` to take advantage of this?

cpcloud:
@mdboom Can fix conflicts so we can merge this?

PR #1664, opened 2026-02-20
refactor(examples): wrap logic in main() functions

OfficialSerge:
## Description

This PR is the second half of the response to the following TODO item in `cuda_core/tests/example_tests/utils.py`.

see [part 1](https://github.com/NVIDIA/cuda-python/pull/1628)
```python
exec(script, env if env else {})
```

## Key Changes
- refactor any remaining non-module examples into Python modules.
- `os.path.join(cuda_path, "include")` refactor to `Path(cuda_path) / "include"` in `thread_block_cluster.py`.
- `os.path.isdir(cuda_include)` refactor to `cuda_include.is_dir()` in `thread_block_cluster.py`.
- some indenting of the `code` blocks, they now have leading white space technically, this shouldn't break anything.
- tried keeping everything else _the same._

## Next Steps
- consider using `contextlib.closing` for Stream and Buffer objects, current failed assertions will halt the execution of the example before `.close()` is called, thus cleanup is now the garbage collector's problem, this is non-deterministic though; since the Python's garbage collector is lazy, _cleanup isn't immediate,_ not the end of the world but Streams or Buffers linger in the GPU drivers. I imagine many failed tests due to assertion errors may cause unneeded strain on GPU resources.

Let me know if this sounds good (seems okay but not critical imo).

### DIFF
--- a/cuda_core/examples/jit_lto_fractal.py
+++ b/cuda_core/examples/jit_lto_fractal.py
@@ -294,8 +294,8 @@ def main():
             axs.set_title(title)
             axs.axis("off")
         plt.show()
+    print("done!")
 
 
 if __name__ == "__main__":
     main()
-    print("done!")
--- a/cuda_core/examples/memory_ops.py
+++ b/cuda_core/examples/memory_ops.py
@@ -25,111 +25,115 @@ from cuda.core import (
     launch,
 )
 
-if np.__version__ < "2.1.0":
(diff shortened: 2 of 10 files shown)


cpcloud:
Nit (non-blocking): in cuda_core/examples/simple_multi_gpu_example.py, vector_sub iterates with i but writes C[tid]. For this sample launch shape they coincide in practice, but C[i] might be clearer and safer if this kernel is copied into a different grid/block setup.

PR #1665, opened 2026-02-20
Fix get_nested_resource_ptr to accept both str and bytes inputs

### LINKED ISSUE #1354: Make `get_nested_resource_ptr` support a nested sequence of bytes
Copilot picked up one thing that is interesting.
https://github.com/NVIDIA/cuda-python/blob/4886636b572e80d74a40fc0a352c9fb762767592/cuda_bindings/cuda/bindings/_internal/utils.pyx#L100-L101
Here, this block handles a nested sequence and the expectation is that the inner objects `obj_i` are strings. But, in such cases it makes sense to also accept bytes, something like
```python
obj_i_type = type(obj_i)
if obj_i_type is str:
    obj_i_bytes = obj_i.encode()
else:
    assert obj_i_type is bytes
    obj_i_bytes = obj_i
```

rparolin:
## Summary
- Fix `get_nested_resource_ptr` in `utils.pyx` to accept both `str` and `bytes` inputs for char resources, instead of only `str` via a Cython `<str?>` cast
- Raise a clear `TypeError` when neither `str` nor `bytes` is provided
- Add tests for bytes-encoded options in nvjitlink and nvvm

Closes #1354

### DIFF
--- a/cuda_bindings/cuda/bindings/_internal/utils.pyx
+++ b/cuda_bindings/cuda/bindings/_internal/utils.pyx
@@ -122,7 +122,7 @@ cdef int get_nested_resource_ptr(nested_resource[ResT] &in_out_ptr, object obj, 
             if ResT is char:
                 obj_i_type = type(obj_i)
                 if obj_i_type is str:
-                    obj_i_bytes = obj_i.encode()
+                    obj_i_bytes = obj_i.encode("utf-8")
                 elif obj_i_type is bytes:
                     obj_i_bytes = obj_i
                 else:


cuda_bindings/cuda/bindings/_internal/utils.pyx
```diff
@@ -120,7 +120,14 @@ cdef int get_nested_resource_ptr(nested_resource[ResT] &in_out_ptr, object obj,
         nested_ptr.reset(nested_vec, True)
         for i, obj_i in enumerate(obj):
             if ResT is char:
-                obj_i_bytes = (<str?>(obj_i)).encode()
+                obj_i_type = type(obj_i)
+                if obj_i_type is str:
+                    obj_i_bytes = obj_i.encode()
```
mdboom: We should specify an encoding here, since the default is system/installation specific prior to Python 3.15.  Probably `utf-8` is the right choice, unless the options aren't unicode-safe.  The `nvvm.h` header doesn't really say...

mdboom (changes requested):

Andy-Jost (commented):
LGTM!

The only thing I see is a `type(obj) == str` check rather than `isinstance`, but I don't think that's that big issue.

PR #1668, opened 2026-02-20
ci: require explicit tag input for cuda-pathfinder release workflow

cpcloud:
## Summary
- replace `version` + `commit` workflow inputs with a single required `tag` input (`cuda-pathfinder-vX.Y.Z`)
- validate that the provided tag exists, matches expected format, and points to a commit reachable from the default branch
- remove in-workflow tag creation so releases only operate on pre-existing tags

## Test plan
- [x] `nix run nixpkgs#actionlint -- .github/workflows/release-cuda-pathfinder.yml`
- [ ] trigger workflow_dispatch in a staging repo with a valid existing `cuda-pathfinder-vX.Y.Z` tag
- [ ] trigger workflow_dispatch with invalid/missing tags to confirm validation failures

### DIFF
--- a/.github/workflows/release-cuda-pathfinder.yml
+++ b/.github/workflows/release-cuda-pathfinder.yml
@@ -4,7 +4,7 @@
 
 # One-click release workflow for cuda-pathfinder.
 #
-# Provide a release tag. The workflow validates the tag, finds
+# Provide a release tag. The workflow finds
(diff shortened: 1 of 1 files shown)


cpcloud:
definite feels overkill to run the whole build and test suite here since this will never be involved in any of that.

Issue #200, opened 2024-10-30
Design the TensorMap object


leofang:
I am not sure if we even need a class full of states, attributes, and methods to wrap `CUtensorMap`. I suspect we just need a lightweight class object + a free, factory function that takes a variety of configs and returns the object (i.e. no class constructor).

Unanswered question: Is `CUtensorMap` bound to a CUDA context?

leofang:
> Is `CUtensorMap` bound to a CUDA context?

Asked internally and the answer is yes. It means the TensorMap object would need to offer a reverse lookup to device/context, as done in other objects.

leofang:
Deprioritizing this to P1 and moving out of beta 3.

Issue #199, opened 2024-10-30
EPIC: Support TMA descriptor


leofang:
Initializing a TMA descriptor through the driver APIs
https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TENSOR__MEMORY.html
is really tedious and error prone. We need a way to abstract it out, which aligns well with the mission of `cuda.core`. This also allows JIT compilers to easier consume and incorporate into the compilation pipelines.

In my understanding there are two (implicit?) requirements for this to be useful:
1. Creating/initializing a TMA object on host
2. Passing the object to the `cuda.core.launch()` API as a kernel arg

leofang:
(Tentatively set to beta 3.)

leofang:
Deprioritizing this to P1 and moving out of beta 3.

leofang:
@fbusato is working on DLPack -> TensorMap conversion on the C++ side: https://github.com/NVIDIA/cccl/pull/6237. I envision that once it's done, we can easily expose this functionality through `StridedMemoryView`, which already supports DLPack.

PR #1387, opened 2025-12-16
WIP: Support passing SMV as a kernel argument of type `cuda::std::mdspan`

leofang:
## Description

Accompanying PR: https://github.com/cupy/cupy/pull/9525

Still WIP, lots to sort out.

closes #748 

## Checklist

- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #1687, opened 2026-02-24
Add TMA TensorMapDescriptor support

rparolin:
## Summary

- Add `TensorMapDescriptor` Cython class wrapping the CUDA driver's `CUtensorMap` for Hopper+ TMA (Tensor Memory Accelerator) bulk data movement
- Support tiled and im2col descriptor creation via `from_tiled()` and `from_im2col()` class methods, with automatic dtype inference, stride computation, and validation
- Integrate `TensorMapDescriptor` as a first-class kernel argument in `_kernel_arg_handler.pyx`
- Add comprehensive tests (`test_tensor_map.py`) and an example (`tma_tensor_map.py`)

Closes https://github.com/NVIDIA/cuda-python/issues/199
Closes https://github.com/NVIDIA/cuda-python/issues/200

### DIFF
--- a/cuda_core/tests/test_tensor_map.py
+++ b/cuda_core/tests/test_tensor_map.py
@@ -376,8 +376,7 @@ class TestTensorMapReplaceAddress:
 
         dev0.set_current()
         buf0 = dev0.allocate(1024 * 4)
-        desc = TensorMapDescriptor.from_tiled(
-            buf0,
+        desc = _as_view(buf0).as_tensor_map(
             box_dim=(64,),
             data_type=TensorMapDataType.FLOAT32,
         )
@@ -399,8 +398,7 @@ class TestTensorMapReplaceAddress:
         skip_if_managed_memory_unsupported(dev1)
 
         dev1.set_current()
-        desc = TensorMapDescriptor.from_tiled(
-            dev1.allocate(1024 * 4),
+        desc = _as_view(dev1.allocate(1024 * 4)).as_tensor_map(
             box_dim=(64,),
             data_type=TensorMapDataType.FLOAT32,
         )
@@ -427,10 +425,9 @@ class TestTensorMapMultiDeviceValidation:
 
         with pytest.raises(
             ValueError,
-            match=r"TensorMapDescriptor\.from_tiled expects tensor on device 0, got 1",
+            match=r"TensorMapDescriptor\._from_tiled expects tensor on device 0, got 1",
         ):
-            TensorMapDescriptor.from_tiled(
-                buf1,
+            _as_view(buf1).as_tensor_map(
                 box_dim=(64,),
                 data_type=TensorMapDataType.FLOAT32,
             )
@@ -448,8 +445,7 @@ class TestTensorMapMultiDeviceValidation:
         mr = create_managed_memory_resource_or_skip(ManagedMemoryResourceOptions(preferred_location=dev1.device_id))
         managed_buf = mr.allocate(1024 * 4)
 
-        desc = TensorMapDescriptor.from_tiled(
-            managed_buf,
+        desc = _as_view(managed_buf).as_tensor_map(
             box_dim=(64,),
             data_type=TensorMapDataType.FLOAT32,
         )


leofang (changes requested):
There is a coordinated effort between C++ and Python: https://github.com/NVIDIA/cuda-python/issues/199#issuecomment-3438556979. Can we please look into reusing the C++ implementation (mainly because @fbusato is a TMA expert) and avoid re-implementing it if possible?

fbusato:
Fighting with poor documentation and bugs don't make me an expert :). 
Anyway, we provide two main functionalities in this direction:

- [make_tma_descriptor()](https://nvidia.github.io/cccl/unstable/libcudacxx/extended_api/tma/make_tma_descriptor.html#libcudacxx-extended-api-tma-make-tma-descriptor) `DLPack` to `cuTensorMap`.
- [mdspan to DLPack](https://nvidia.github.io/cccl/unstable/libcudacxx/extended_api/mdspan/mdspan_to_dlpack.html#libcudacxx-extended-api-mdspan-mdspan-to-dlpack) and [DLPack to mdspan](https://nvidia.github.io/cccl/unstable/libcudacxx/extended_api/mdspan/dlpack_to_mdspan.html#libcudacxx-extended-api-mdspan-dlpack-to-mdspan).

The implementation of `make_tma_descriptor` is here https://github.com/NVIDIA/cccl/blob/main/libcudacxx/include/cuda/__tma/make_tma_descriptor.h. Please let me know if there are functionalities that need to be isolated for reuse.

cuda_core/cuda/core/_kernel_arg_handler.pyx
```diff
@@ -322,6 +350,9 @@ cdef class ParamHolder:
                 elif isinstance(arg, driver.CUgraphConditionalHandle):
                     prepare_arg[cydriver.CUgraphConditionalHandle](self.data, self.data_addresses, arg, i)
                     continue
+                elif isinstance(arg, tensor_map_descriptor_type):
+                    prepare_tensor_map_arg(self.data, self.data_addresses, <TensorMapDescriptor>arg, i)
+                    continue
```
leofang: I don't remember off top of my head why we had to repeat the checks in this block, I believe it has to do with backward compatibility. Since TMA support is new, we don't need to add it here.

.gitignore
```diff
@@ -14,6 +14,7 @@ __pycache__/
 !*_impl.cpp
 !cuda_bindings/cuda/bindings/_lib/param_packer.cpp
 !cuda_bindings/cuda/bindings/_bindings/loader.cpp
+!cuda_core/cuda/core/_cpp/*.cpp
```
leofang: why do we ignore them...?

cuda_core/cuda/core/__init__.py
```diff
@@ -68,3 +68,12 @@
...
     StreamOptions,
 )
+from cuda.core._tensor_map import (
+    TensorMapDataType,
+    TensorMapDescriptor,
+    TensorMapIm2ColWideMode,
+    TensorMapInterleave,
+    TensorMapL2Promotion,
+    TensorMapOOBFill,
+    TensorMapSwizzle,
```
leofang: Critical:
1. Only expose `TensorMapDescriptor` to under `cuda.core`
2. Since `StridedMemoryView` already parses DLPack content, we expose `StridedMemoryView.as_tensor_map` (parallel to `as_mdspan` in #1387) and return a `TensorMapDescriptor`
3. I don't think we should offer alternative constructors, at least for now, because otherwise we shoot ourselves in the foot when DevTechs come and complain about perf like in #659. If we really, really want to do it, we must provide a `TensorMapOptions` encapsulating `Interleave`, `Swizzle`, etc, and expose `Device.create_tensor_map(options: TensorMapOptions)`. Right now, this does not follow our general design pattern, and makes `cuda.core` namespace crowded.
4. Make sure `TensorMapDescriptor` tracks the current device/context like streams and events. It is CUDA context-dependent. (Hence it makes sense to offer `Device.create_tensor_map`.)

cuda_core/examples/tma_replace_address.py
leofang, line 1: 👍

cuda_core/examples/tma_tensor_map.py
```diff
@@ -0,0 +1,175 @@
...
+}
+"""
+
+# ---------------------------------------------------------------------------
+# Compile the kernel
+# ---------------------------------------------------------------------------
+prog = Program(
+    code,
+    code_type="c++",
+    options=ProgramOptions(std="c++17", arch=f"sm_{arch_str}"),
```
leofang: nit: this is simpler
```suggestion
    options=ProgramOptions(std="c++17", arch=f"sm_{dev.arch}"),
```

cuda_core/examples/tma_tensor_map.py
leofang, line 1: 👍

cuda_core/cuda/core/_kernel_arg_handler.pyx
cpcloud: Addressed in 892ee605e. I removed the redundant fallback `isinstance(..., tensor_map_descriptor_type)` branch from the slow path and kept TensorMap handling only in the direct type-dispatch path for this new feature. I also added a context-compatibility check before packing a TensorMap kernel argument.

.gitignore
cpcloud: Done in 892ee605e. I removed the `.gitignore` exception for `cuda_core/cuda/core/_cpp/*.cpp`. Those sources are now treated as normal tracked C++ sources.

cuda_core/cuda/core/__init__.py
cpcloud: Addressed in 892ee605e. For this review set I applied: (1) `cuda.core` now exports only `TensorMapDescriptor`, (2) added `StridedMemoryView.as_tensor_map(...)` and switched the TMA examples/tests to this path, (3) moved TensorMap option enums off the `cuda.core` top-level to reduce namespace crowding, and (4) `TensorMapDescriptor` now records creation context/device and enforces compatibility during `replace_address` and kernel-arg packing. I kept existing constructors for compatibility in this PR, but the primary user path is now through `StridedMemoryView.as_tensor_map`.

cuda_core/examples/tma_replace_address.py
cpcloud: Thanks! No further change needed on this thread.

cuda_core/examples/tma_tensor_map.py
cpcloud: Applied in 892ee605e. I updated the example to use `ProgramOptions(std="c++17", arch=f"sm_{dev.arch}")`.

cuda_core/examples/tma_tensor_map.py
cpcloud: Thanks! No additional changes needed here.

cpcloud:
@leofang Please review. This is now blocked on your review.

cuda_core/cuda/core/_cpp/tensor_map_cccl.h
```diff
@@ -0,0 +1,43 @@
...
+
+#ifndef CUDA_CORE_TENSOR_MAP_CCCL_H_
+#define CUDA_CORE_TENSOR_MAP_CCCL_H_
+
+#include <stddef.h>
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
```
leofang: nit:
```suggestion
#ifdef __cplusplus
#include <cstddef>
#include <cstdint>
extern "C" {
#else
#include <stddef.h>
#include <stdint.h>
#endif
```

cuda_core/cuda/core/_cpp/tensor_map.cpp
```diff
@@ -0,0 +1,149 @@
...
+#    define CUDA_CORE_HAS_CUDA_TMA 1
+#  else
+#    define CUDA_CORE_HAS_CUDA_TMA 0
+#  endif
+#  if __has_include(<dlpack/dlpack.h>)
+#    include <dlpack/dlpack.h>
+#    define CUDA_CORE_HAS_DLPACK_H 1
+#  else
+#    define CUDA_CORE_HAS_DLPACK_H 0
+#  endif
```
leofang: Q: we guarantee to have a `dlpack.h` during build time, but it's not accessible via `<dlpack/dlpack.h>`, so does it mean we end up with `CUDA_CORE_HAS_DLPACK_H` == 0?

cuda_core/cuda/core/_cpp/tensor_map.cpp
```diff
@@ -0,0 +1,149 @@
...
+#include "tensor_map_cccl.h"
+
+#include <string.h>
+
+#include <algorithm>
+#include <exception>
+
+#if defined(__has_include)
+#  if __has_include(<cuda/tma>)
+#    include <cuda/tma>
```
leofang, line 16: Q: I am confused -- The TMA header was added fairly recently. We build cuda.core against both CUDA 12 & 13 and merge the resulting wheels. Without vendoring the CCCL header, how did we manage to build and make CI green? 🤔

cuda_core/cuda/core/_kernel_arg_handler.pyx
```diff
@@ -124,6 +130,26 @@ cdef inline int prepare_arg(
     return 0
 
 
+cdef inline int prepare_tensor_map_arg(
+        vector.vector[void*]& data,
+        vector.vector[void*]& data_addresses,
+        TensorMapDescriptor arg,
+        const size_t idx) except -1:
+    arg._check_context_compat()
```
leofang: I am torn on this. I understand the rational but this call will be very slow since it involves multiple driver API calls. 

We should check this when a TMA is constructed in Python, memoize the device/context attributes, and then move on. For example, we don't do pointer attribute check at launch time either. It adds just too much overhead.

cuda_core/cuda/core/_kernel_arg_handler.pyx
```diff
@@ -124,6 +130,26 @@ cdef inline int prepare_arg(
...
+    # Allocate a temporary buffer for the 128-byte CUtensorMap struct.
+    # We copy rather than pointing directly at arg._tensor_map for lifetime
+    # safety: ParamHolder owns and frees its argument buffers independently.
+    cdef void* ptr = PyMem_Malloc(sizeof(cydriver.CUtensorMap))
+    if ptr is NULL:
+        raise MemoryError("Failed to allocate memory for CUtensorMap")
+    memcpy(ptr, arg._get_data_ptr(), sizeof(cydriver.CUtensorMap))
```
leofang: This is unnecessary because the driver will copy again. Just pack the pointer and pass it to cuLaunchKernel, and the driver will copy it over.

cuda_core/cuda/core/_kernel_arg_handler.pyx
```diff
@@ -273,6 +299,9 @@ cdef class ParamHolder:
                     # it's a CUdeviceptr:
                     self.data_addresses[i] = <void*><intptr_t>(arg.handle.getPtr())
                 continue
+            elif arg_type is tensor_map_descriptor_type:
+                prepare_tensor_map_arg(self.data, self.data_addresses, <TensorMapDescriptor>arg, i)
+                continue
```
leofang: Let's move this down, after `arg_type is complex`. At least after `arg_type is int`, which is our fast path.

cuda_core/cuda/core/_memoryview.pyx
```diff
@@ -316,6 +316,39 @@ cdef class StridedMemoryView:
         view_buffer_strided(view, self.get_buffer(), layout, dtype, self.readonly)
         return view
 
+    def as_tensor_map(
```
leofang, line 290: 👍

cuda_core/cuda/core/_memoryview.pyx
```diff
@@ -316,6 +316,39 @@ cdef class StridedMemoryView:
...
+        self,
+        box_dim,
+        *,
+        element_strides=None,
+        data_type=None,
+        interleave=None,
+        swizzle=None,
+        l2_promotion=None,
+        oob_fill=None,
```
leofang: I see. I understand now this is unavoidable. Let's pack these into a `TensorMapDescriptorOptions` dataclass, and write a docstring there:
```python
class TensorMapDescriptorOptions:
    box_dim: int
    element_strides: Optional[tuple[int, ...]] = None
    data_type: Optional[np.dtype] = None
    ...
```
then the kwarg checking can be moved to its `__post_init__`.

cuda_core/cuda/core/_tensor_map.pyx
```diff
@@ -0,0 +1,936 @@
...
+    numpy.dtype(numpy.int32): TensorMapDataType.INT32,
+    numpy.dtype(numpy.uint64): TensorMapDataType.UINT64,
+    numpy.dtype(numpy.int64): TensorMapDataType.INT64,
+    numpy.dtype(numpy.float16): TensorMapDataType.FLOAT16,
+    numpy.dtype(numpy.float32): TensorMapDataType.FLOAT32,
+    numpy.dtype(numpy.float64): TensorMapDataType.FLOAT64,
+}
+
+if ml_bfloat16 is not None:
+    _NUMPY_DTYPE_TO_TMA[numpy.dtype(ml_bfloat16)] = TensorMapDataType.BFLOAT16
```
leofang: 👍

cuda_core/cuda/core/_tensor_map.pyx
```diff
@@ -0,0 +1,936 @@
...
+        size_t err_cap) nogil
+
+
+try:
+    from ml_dtypes import bfloat16 as ml_bfloat16
+except ImportError:
+    ml_bfloat16 = None
+
+
+class TensorMapDataType(enum.IntEnum):
```
leofang, line 47: 1. Users should just give us numpy/ml_dtype types. This does not need to go public (as part of the suggested `TensorMapDescriptorOptions`).
2. Internally, we can just use `cydriver` C enums. Going through `TensorMapDataType` is slow (due to Python object access, and `IntEnum` has a big overhead).

cuda_core/cuda/core/_tensor_map.pxd
```diff
@@ -0,0 +1,18 @@
...
+from cuda.bindings cimport cydriver
+from libc.stdint cimport intptr_t
+
+
+cdef class TensorMapDescriptor:
+    cdef cydriver.CUtensorMap _tensor_map
+    cdef int _device_id
+    cdef intptr_t _context
+    cdef object _source_ref
+    cdef object _view_ref
```
leofang: Q: I think this can be typed?
```suggestion
    cdef StridedMemoryView _view_ref
```

cuda_core/cuda/core/_tensor_map.pyx
```diff
@@ -0,0 +1,936 @@
...
+        cdef int device_type
+        cdef int c_device_id
+        cdef int dl_device_type
+        cdef int dl_device_id
+        cdef int c_cccl_interleave_int
+        cdef int c_cccl_swizzle_int
+        cdef int c_cccl_l2_promotion_int
+        cdef int c_cccl_oob_fill_int
+        cdef int rc
+        if _tma_dtype_to_dlpack(tma_dt, &dl_code, &dl_bits, &dl_lanes):
```
leofang, line 636: I am not sure why do we need this function...? If a StridedMemoryView is provided, and it was created via DLPack, it'd hold a DLTensor that we can just access. The amount of parameter packing/unpacking/checking here seems to be too excessive to defeat the purpose of code reusing (and delegate to CCCL to do checking).

cuda_core/cuda/core/_cpp/tensor_map.cpp
leofang: It's only available via CTK until CUDA 13.2, meaning before this week it was not there. 
https://nvidia.github.io/cccl/unstable/libcudacxx/extended_api/tma/make_tma_descriptor.html

cuda_core/cuda/core/_tensor_map.pyx
```diff
@@ -0,0 +1,936 @@
...
+            HANDLE_RETURN(cydriver.cuCtxGetDevice(&current_dev))
+        current_dev_id = <int>current_dev
+        if self._device_id >= 0 and current_dev_id != self._device_id:
+            raise RuntimeError(
+                f"TensorMapDescriptor belongs to device {self._device_id}, "
+                f"but current device is {current_dev_id}")
+        return 0
+
+    @classmethod
+    def from_tiled(cls, tensor, box_dim, *,
```
leofang: To speed up merging of this PR & align with CUDA C++, I suggest that we make all methods private (by prefixing an underscore) for now and punt the alternative constructors to a future PR. We only rely on DLPack for now.

cuda_core/cuda/core/_tensor_map.pyx
```diff
@@ -0,0 +1,936 @@
...
+            "method": "tiled",
+            "rank": rank,
+            "data_type": tma_dt,
+            "swizzle": swizzle,
+        }
+
+        return desc
+
+    @classmethod
+    def from_im2col(cls, tensor, pixel_box_lower_corner, pixel_box_upper_corner,
```
leofang: ditto, make it private

cuda_core/cuda/core/_tensor_map.pyx
```diff
@@ -0,0 +1,936 @@
...
+            "method": "im2col",
+            "rank": rank,
+            "data_type": tma_dt,
+            "swizzle": swizzle,
+        }
+
+        return desc
+
+    @classmethod
+    def from_im2col_wide(cls, tensor, pixel_box_lower_corner_width, pixel_box_upper_corner_width,
```
leofang: ditto, make it private

cuda_core/cuda/core/_tensor_map.pyx
```diff
@@ -0,0 +1,936 @@
...
+            desc._repr_info = {
+                "method": "im2col_wide",
+                "rank": rank,
+                "data_type": tma_dt,
+                "swizzle": swizzle,
+            }
+
+            return desc
+
+    def replace_address(self, tensor):
```
leofang, line 1055: Q: Maybe we should take a `StridedMemovyView` here instead of a tensor? Then, we sweep the responsibility to users (to create a view), and we simplify the code here (by just accessing `view.ptr` and `view.device_id`).

cuda_core/cuda/core/_tensor_map.pyx
```diff
@@ -0,0 +1,936 @@
...
+    """Validate or default element_strides to all-ones."""
+    if element_strides is not None:
+        if len(element_strides) != rank:
+            raise ValueError(
+                f"element_strides must have {rank} elements, got {len(element_strides)}")
+        return element_strides
+    return (1,) * rank
+
+
+cdef class TensorMapDescriptor:
```
leofang, line 459: TODO: Expose `TensorMapDescriptor.device(self) -> Device` following stream and event.

cuda_core/examples/tma_replace_address.py
```diff
@@ -0,0 +1,189 @@
...
+    // ---- Wait for the TMA load to complete ----
+    if (tid == 0)
+    {
+        asm volatile(
+            "{ .reg .pred P;                                           \n"
+            "WAIT:                                                     \n"
+            "  mbarrier.try_wait.parity.shared.b64 P, [%0], 0;         \n"
+            "  @!P bra WAIT;                                           \n"
+            "}                                                         \n"
+            :: "r"((unsigned)__cvta_generic_to_shared(&mbar)));
```
leofang: Let's not teach users to use raw PTX instructions. I am sure libcudc++ has nice wrappers for them. 
https://nvidia.github.io/cccl/unstable/libcudacxx/ptx/instructions/generated/mbarrier_try_wait_parity.html

cuda_core/examples/tma_replace_address.py
```diff
@@ -0,0 +1,189 @@
...
+            :: "r"((unsigned)__cvta_generic_to_shared(smem)),
+               "l"(&tensor_map),
+               "r"(tile_start),
+               "r"((unsigned)__cvta_generic_to_shared(&mbar)));
+
+        // Tell the mbarrier how many bytes the TMA will deliver.
+        asm volatile(
+            "mbarrier.arrive.expect_tx.shared.b64 _, [%0], %1;"
+            :: "r"((unsigned)__cvta_generic_to_shared(&mbar)),
+               "r"((unsigned)(TILE_SIZE * sizeof(float))));
```
leofang: no raw PTX (see my comment below/above)

cuda_core/examples/tma_replace_address.py
```diff
@@ -0,0 +1,189 @@
...
+}
+"""
+
+# ---------------------------------------------------------------------------
+# Compile the kernel
+# ---------------------------------------------------------------------------
+prog = Program(
+    code,
+    code_type="c++",
+    options=ProgramOptions(std="c++17", arch=f"sm_{arch_str}"),
```
leofang: nit:
```suggestion
    options=ProgramOptions(std="c++17", arch=f"sm_{dev.arch}"),
```

cuda_core/examples/tma_tensor_map.py
```diff
@@ -0,0 +1,172 @@
...
+            :: "r"((unsigned)__cvta_generic_to_shared(smem)),
+               "l"(&tensor_map),
+               "r"(tile_start),
+               "r"((unsigned)__cvta_generic_to_shared(&mbar)));
+
+        // Tell the mbarrier how many bytes the TMA will deliver.
+        asm volatile(
+            "mbarrier.arrive.expect_tx.shared.b64 _, [%0], %1;"
+            :: "r"((unsigned)__cvta_generic_to_shared(&mbar)),
+               "r"((unsigned)(TILE_SIZE * sizeof(float))));
```
leofang: ditto

cuda_core/examples/tma_tensor_map.py
```diff
@@ -0,0 +1,172 @@
...
+    // ---- Wait for the TMA load to complete ----
+    if (tid == 0)
+    {
+        asm volatile(
+            "{ .reg .pred P;                                           \n"
+            "WAIT:                                                     \n"
+            "  mbarrier.try_wait.parity.shared.b64 P, [%0], 0;         \n"
+            "  @!P bra WAIT;                                           \n"
+            "}                                                         \n"
+            :: "r"((unsigned)__cvta_generic_to_shared(&mbar)));
```
leofang: ditto

cuda_core/examples/tma_replace_address.py
leofang, line 1: I now think this example should just be combined with `tma_tensor_map.py`, since we have lots of code repetition here.

cuda_core/cuda/core/__init__.py
```diff
@@ -68,3 +68,4 @@
     Stream,
     StreamOptions,
 )
+from cuda.core._tensor_map import TensorMapDescriptor
```
leofang: see comments below/above
```suggestion
from cuda.core._tensor_map import TensorMapDescriptor, TensorMapDescriptorOptions
```

cuda_core/cuda/core/_cpp/tensor_map.cpp
rparolin: Yes, this is generally good practice in C++ code so you can avoid having to check if a pre-processor macro is defined before checking its value.

```
#if defined(CUDA_CORE_HAS_DLPACK_H) && CUDA_CORE_HAS_DLPACK_H == 0
  // ...
#endif
```

cuda_core/cuda/core/_cpp/tensor_map.cpp
rparolin: The dependent code is only compiled in when the header is available: `CUDA_CORE_HAS_CUDA_TMA`.

cpcloud:
Latest update on this branch:

- Pushed a final follow-up commit (`9478eca7e8`) to finish the remaining TensorMap review items.
- Added `TensorMapDescriptorOptions` for the tiled path and wired `StridedMemoryView.as_tensor_map(..., options=...)` to accept either the dataclass or a mapping.
- Kept dtype-like inputs (`numpy` / `ml_dtypes`) on the public API while using raw driver enum values internally, so the hot path no longer needs to carry Python `IntEnum` objects around.
- Added `cuda_core/cuda/core/_memoryview.pxd` and typed `_view_ref` as `StridedMemoryView`, which lets the descriptor retain DLPack-backed views with explicit Cython typing.
- Folded the last one-off validated-view helper back into `replace_address()` so the TensorMap code path is a bit smaller/clearer.
- All previously open review threads on the PR are now resolved.

Verification on the current branch:

- `pixi run -e cu13 pytest tests/test_tensor_map.py -q` -> `8 passed, 40 skipped`
- `pixi run -e cu13 pytest tests -q` -> `1783 passed, 188 skipped, 2 xfailed, 22 warnings`

The warnings in the broader run were the existing `pytest.mark.flaky` registration warnings from the memory IPC tests.

cuda_core/cuda/core/_tensor_map.pyx
```diff
@@ -0,0 +1,679 @@
...
+        ------
+        ValueError
+            If the tensor rank is outside [1, 5], the pointer is not
+            16-byte aligned, or dimension/stride constraints are violated.
+        """
+        cdef TensorMapDescriptor desc = cls.__new__(cls)
+
+        view = _get_validated_view(tensor)
+        desc._source_ref = tensor
+
```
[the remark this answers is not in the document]
cpcloud: Handled by keeping both the exporting object and the validated `StridedMemoryView` on the descriptor (`_source_ref` + `_view_ref`). That keeps the DLPack capsule/view alive for the descriptor lifetime instead of only holding the original tensor.

cuda_core/cuda/core/_cpp/tensor_map.cpp
cpcloud: Handled by checking for vendored `"dlpack.h"` first and only falling back to `<dlpack/dlpack.h>`. `CUDA_CORE_HAS_DLPACK_H` now stays true in the build layout we actually use.

cuda_core/cuda/core/_cpp/tensor_map.cpp
cpcloud: Handled by gating the CCCL helper on `__has_include(<cuda/tma>)` / `CUDA_CORE_HAS_CUDA_TMA` and falling back to the direct driver path when the header is unavailable. That keeps CUDA 12 builds green without vendoring the new TMA header.

cuda_core/cuda/core/_kernel_arg_handler.pyx
cpcloud: Handled by moving the compatibility state to construction time on `TensorMapDescriptor` and dropping the launch-time context/device check from arg packing. `prepare_tensor_map_arg()` now just forwards the descriptor pointer.

cuda_core/cuda/core/_kernel_arg_handler.pyx
cpcloud: Handled by removing the host-side `CUtensorMap` allocation/copy. `prepare_tensor_map_arg()` now passes `arg._get_data_ptr()` directly and relies on the driver's own kernel-argument copy.

cuda_core/cuda/core/_kernel_arg_handler.pyx
cpcloud: Handled by moving the `tensor_map_descriptor_type` exact-type dispatch below the scalar fast paths (`bool` / `int` / `float` / `complex`). That keeps the common launch path hot and isolates TMA handling to its own branch.

cuda_core/cuda/core/_memoryview.pyx
cpcloud: Kept this shape in the final API: `StridedMemoryView.as_tensor_map()` is now the public tiled-constructor entry point, and the example/test call sites were moved over to it.

cuda_core/cuda/core/_memoryview.pyx
cpcloud: Handled by adding `TensorMapDescriptorOptions` with the docstring plus `__post_init__` validation/normalization, and teaching `StridedMemoryView.as_tensor_map(..., options=...)` to forward that bundled config.

cuda_core/cuda/core/_tensor_map.pyx
cpcloud: Kept this cleanup in the follow-up dtype/options refactor. The tiled-constructor path now carries lightweight internal TMA dtype ints on the hot path and only wraps back to `TensorMapDataType(...)` for repr/debug metadata.

cuda_core/cuda/core/_tensor_map.pyx
cpcloud: Handled by normalizing user input to NumPy/`ml_dtypes` dtypes in `TensorMapDescriptorOptions` / `_resolve_data_type()` and switching the internal hot path to raw `cydriver` integer values (`_TMA_DT_*`). We only wrap back to `TensorMapDataType(...)` for repr metadata.

cuda_core/cuda/core/_tensor_map.pxd
cpcloud: Handled by adding `cuda_core/cuda/core/_memoryview.pxd` and typing `_view_ref` as `StridedMemoryView` in `_tensor_map.pxd`, so the retained view is strongly typed across the Cython modules.

cuda_core/cuda/core/_tensor_map.pyx
cpcloud: Handled by trimming the helper surface rather than adding more layers: `_get_validated_view()` now only normalizes arbitrary tensor inputs to a validated `StridedMemoryView`, while the descriptor builders reuse the view metadata and still delegate tiled validation/encoding to CCCL where possible.

cuda_core/cuda/core/_tensor_map.pyx
cpcloud: Handled by making the alternative constructors private (`_from_tiled`, `_from_im2col`, `_from_im2col_wide`) and steering the public user path through `StridedMemoryView.as_tensor_map()`.

cuda_core/cuda/core/_tensor_map.pyx
cpcloud: Handled together with the same API cleanup: `from_im2col()` stays internal as `_from_im2col()` for now.

cuda_core/cuda/core/_tensor_map.pyx
cpcloud: Handled together with the same API cleanup: `from_im2col_wide()` is now `_from_im2col_wide()` and remains internal while the public surface settles.

cuda_core/cuda/core/_tensor_map.pyx
cpcloud: I did not narrow the signature all the way to `StridedMemoryView`, but I did consolidate the implementation around a validated view: `replace_address()` now normalizes through `_get_validated_view()`, uses the view for pointer/device checks, and stores it in `_view_ref` to keep the export alive.

cuda_core/cuda/core/_tensor_map.pyx
cpcloud: Handled by adding `TensorMapDescriptor.device`, which returns the associated `Device` the same way stream/event objects expose their device affiliation.

cuda_core/examples/tma_replace_address.py
cpcloud: Handled by removing the raw PTX path and switching the example kernel to libcudacxx `cuda::barrier` plus `cuda::device::experimental::cp_async_bulk_tensor_1d_global_to_shared(...)` / `barrier_arrive_tx(...)` wrappers.

cuda_core/examples/tma_replace_address.py
cpcloud: Same fix here: the example no longer embeds raw PTX, it uses the libcudacxx barrier/TMA wrappers instead.

cuda_core/examples/tma_tensor_map.py
cpcloud: Updated here too: the surviving `tma_tensor_map.py` example now uses the libcudacxx barrier/TMA wrappers instead of raw PTX.

cuda_core/examples/tma_tensor_map.py
cpcloud: Also cleaned up the unified example path here: `tma_tensor_map.py` now carries the direct `ProgramOptions(...)` call and the explicit CUDA/CCCL include-path handling needed for `<cuda/barrier>`.

cuda_core/examples/tma_replace_address.py
cpcloud: Handled by deleting `tma_replace_address.py` and folding that flow into `tma_tensor_map.py`, so the single example now covers both descriptor creation and `replace_address()` reuse.

cpcloud:
@leofang Addressed all your comments. I think this is good to go.

PR #1688, opened 2026-02-24
[pathfinder] Add `DynamicLibUnknownError`, `DynamicLibNotAvailableError` and prepare for v1.4.0 release

### LINKED ISSUE #1684: CuPy noticed the error type change for `load_nvidia_dynamic_lib("nccl")` (windows)
Probably better to open an issue:  I don't have a very strong opinion on this, but it would be nice to know where to go (i.e. if to catch both/all errors in CuPy or just blocklist the one version in CI).

CuPy notice this change in CI (if probably just `cupy.show_runtime()` breaking on windows). (Checking for `nccl`.)

I.e. the CuPy code currently assumed that `load_nvidia_dynamic_lib` always raises `DynamicLibNotFoundError`.  It could be fixed on either side, but maybe it makes sense to just keep raising `DynamicLibNotFoundError` (with the new message), so that downstream try/except is unaffected?

(Although, I am not sure if it may be nice to be able to distinguish the two errors.)

CC @leofang

_Originally posted by @seberg in https://github.com/NVIDIA/cuda-python/pull/1602#discussion_r2846123326_

rwgk:
Close #1684

Also prepare for cuda-pathfinder v1.4.0 release.

### DIFF
--- a/cuda_pathfinder/docs/source/release/1.3.5-notes.rst
+++ b/cuda_pathfinder/docs/source/release/1.3.5-notes.rst
@@ -10,12 +10,11 @@ Released on Feb 23, 2026
 
 .. warning::
 
-   **YANK NOTICE (PyPI):** ``cuda-pathfinder==1.3.5`` is planned to be yanked
-   because it introduced a short-lived backward-incompatible exception behavior
-   in ``load_nvidia_dynamic_lib()`` (unsupported library names raising
-   ``ValueError``).
+   ``cuda-pathfinder==1.3.5`` introduced a short-lived backward-incompatible
+   exception regression in ``load_nvidia_dynamic_lib()`` (unsupported
+   library names incorrectly raising ``ValueError``).
 
-   Please use ``cuda-pathfinder>=1.4.0``, which restores backward-compatible
+   This bug was corrected in ``cuda-pathfinder>=1.4.0``, restoring backward-compatible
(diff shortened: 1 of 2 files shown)


rwgk:
~~cuda-pathfinder 1.3.5 yank playbook~~ removed

See commit 99ad65060836e3d5ce48c3ce73cf94515654287c

cuda_pathfinder/docs/source/release/1.3.5-notes.rst
```diff
@@ -8,12 +8,27 @@
 
 Released on Feb 23, 2026
 
+.. warning::
+
+   ``cuda-pathfinder==1.3.5`` introduced a short-lived backward-incompatible
+   exception regression in ``load_nvidia_dynamic_lib()`` (unsupported
+   library names incorrectly raising ``ValueError``).
+
+   This bug was corrected in ``cuda-pathfinder>=1.4.0``, restoring backward-compatible
```
cpcloud, line 17: This reads awkward to me. Bugs are corrected in specific releases, and by implication that means `>=`. Combining those two notions just makes this strange to read.

Not blocking though.

rwgk: I agree, this should just be 1.4.0, but it's not misleading, I don't want to run the CI again for this, and not use [doc-only] in the PR description because that could be confusing when we look at it later. I'll go ahead and merge this as-is, then make the release.

PR #1694, opened 2026-02-26
Standardize error handling and output across all examples

### LINKED ISSUE #1678: Examples: Standardize error handling and output
Parent issue: #1673

### Goal

Apply consistent error handling, early-exit patterns, and output messages across all examples.

### Current state

**Error handling:**
- Some files use `sys.exit(0)` for unsupported configurations; others silently skip.
- NumPy version checking uses two different APIs (`np.__version__ < "2.1.0"` vs `np.lib.NumpyVersion()`).
- Validation is a mix of `assert` statements and `print` + `sys.exit`.

**Output messages:**
- Completion messages vary: `"done!"`, `"done"` (no exclamation), `"Example completed successfully!"`, `"All tests passed successfully!"`, `"Memory management example completed!"`, or no message at all.
- Some examples print intermediate progress, others are silent until the end.
- Some use `sys.stderr` for warnings/info, others use `stdout`.

### Decisions needed

1. What is the standard early-exit pattern for unsupported configurations?
2. Should examples use `assert` or `print` + `sys.exit` for validation?
3. What should the standard success message be (or should there be one)?
4. Should warnings/skip messages go to stderr or stdout?

### Files affected

All 25 example files listed in #1673.

cpcloud:
## Summary

Closes #1678.

- Remove success/completion fluff messages ("done!", "passed", "completed successfully", etc.) — no news is good news
- Use `sys.exit(1)` instead of `sys.exit(0)` for unsupported configurations (non-zero = did not run)
- Send all skip/warning/error messages to `stderr`
- Replace user-facing `assert` with proper `if`/`print`/`sys.exit(1)` checks
- Standardize `sys.exit(-1)` to `sys.exit(1)`
- Use `pytest.skip()` for unsupported-config checks in `_test.py` files (run under pytest)

Applies to all 23 example files across `cuda_core/examples/` and `cuda_bindings/examples/`, plus `common/common.py`.

### DIFF
--- a/cuda_bindings/examples/0_Introduction/simpleP2P_test.py
+++ b/cuda_bindings/examples/0_Introduction/simpleP2P_test.py
@@ -24,23 +24,19 @@ __global__ void SimpleKernel(float *src, float *dst)
 
 
 def main():
-    print("Starting...")
+    import pytest
 
     if platform.system() == "Darwin":
-        print("simpleP2P is not supported on Mac OSX - waiving sample")
-        return
+        pytest.skip("simpleP2P is not supported on Mac OSX")
 
     if platform.machine() == "armv7l":
-        print("simpleP2P is not supported on ARMv7 - waiving sample")
-        return
+        pytest.skip("simpleP2P is not supported on ARMv7")
 
     if platform.machine() == "aarch64":
-        print("simpleP2P is not supported on aarch64 - waiving sample")
-        return
+        pytest.skip("simpleP2P is not supported on aarch64")
 
     if platform.machine() == "sbsa":
-        print("simpleP2P is not supported on sbsa - waiving sample")
-        return
+        pytest.skip("simpleP2P is not supported on sbsa")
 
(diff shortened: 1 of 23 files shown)


cuda_bindings/examples/0_Introduction/simpleP2P_test.py
```diff
@@ -239,9 +232,8 @@ def main():
         checkCudaErrors(cudart.cudaSetDevice(i))
 
     if error_count != 0:
-        print("Test failed!")
-        sys.exit(-1)
```
rparolin, line 243: We aren't breaking behaviour for higher systems that would be explicitly checking for a -1 return code?

cpcloud: It's a test for an example. No one should be depending on this to not break at any time. Depending on the specific return code of invoking this program would be analogous to depending on a specific object being available in a project's test suite.

PR #1696, opened 2026-02-26
chore: fix pixi toml

cpcloud:


cpcloud:
@kkraus14 Can you admin merge this? We're not actually running pixi in CI anywhere (yet).

PR #1698, opened 2026-02-27
Fix get_nested_resource_ptr to accept both str and bytes inputs

rparolin:
Reverts NVIDIA/cuda-python#1697

### DIFF
--- a/cuda_bindings/tests/test_nvvm.py
+++ b/cuda_bindings/tests/test_nvvm.py
@@ -127,7 +127,7 @@ def test_get_buffer_empty(get_size, get_buffer):
 
 
 @pytest.mark.parametrize("options", [[], ["-opt=0"], ["-opt=3", "-g"], [b"-opt=0"]])
-def test_compile_program_with_minimal_nvvm_ir(minimal_nvvmir, options):  # noqa: F401, F811
+def test_compile_program_with_minimal_nvvm_ir(minimal_nvvmir, options):
     with nvvm_program() as prog:
         nvvm.add_module_to_program(prog, minimal_nvvmir, len(minimal_nvvmir), "FileNameHere.ll")
         try:


cuda_bindings/cuda/bindings/_internal/utils.pyx
```diff
@@ -120,10 +121,17 @@ cdef int get_nested_resource_ptr(nested_resource[ResT] &in_out_ptr, object obj,
...
+                    obj_i_bytes = obj_i.encode("utf-8")
+                elif obj_i_type is bytes:
+                    obj_i_bytes = obj_i
+                else:
+                    raise TypeError(
+                        f"Expected str or bytes, got {obj_i_type.__name__}")
                 str_len = <size_t>(len(obj_i_bytes)) + 1  # including null termination
                 deref(nested_res_vec)[i].resize(str_len)
-                obj_i_ptr = <char*>(obj_i_bytes)
+                obj_i_ptr = PyBytes_AsString(obj_i_bytes)
```
leofang: In general we should let Cython figure out the correct C API to use, and not intervene with manual C API calls unless we can prove 
1. Cython emits wrong code, or
2. We can do it with an equally safe but more performant way

cuda_bindings/tests/test_nvjitlink.py
```diff
@@ -99,6 +99,13 @@ def test_create_and_destroy(option):
     nvjitlink.destroy(handle)
 
 
+@pytest.mark.parametrize("option", ARCHITECTURES)
```
leofang: no need to parametrize it, just one `bytes` instance is fine

cuda_bindings/tests/test_nvvm.py
```diff
@@ -115,7 +115,9 @@ def test_get_buffer_empty(get_size, get_buffer):
         assert buffer == b"\x00"
 
 
-@pytest.mark.parametrize("options", [[], ["-opt=0"], ["-opt=3", "-g"]])
+@pytest.mark.parametrize(
+    "options", [[], ["-opt=0"], ["-opt=3", "-g"], [b"-opt=0"], [b"-opt=3", b"-g"], ["-opt=3", b"-g"]]
+)
```
leofang: ditto

cuda_bindings/tests/test_nvvm.py
```diff
@@ -135,7 +137,9 @@ def test_compile_program_with_minimal_nvvm_ir(minimal_nvvmir, options):  # noqa:
         assert ".visible .entry kernel()" in buffer.decode()
 
 
-@pytest.mark.parametrize("options", [[], ["-opt=0"], ["-opt=3", "-g"]])
+@pytest.mark.parametrize(
+    "options", [[], ["-opt=0"], ["-opt=3", "-g"], [b"-opt=0"], [b"-opt=3", b"-g"], ["-opt=3", b"-g"]]
+)
```
leofang: ditto; also, if this is already tested above, no need to test again

cpcloud:
@rparolin Looks like there are conflicts.

cpcloud:
Nit (non-blocking): in get_nested_resource_ptr() for the char path, would isinstance(obj_i, (str, bytes)) be preferable to exact type(...) is ... checks? That is a bit more permissive for subclasses while keeping behavior the same for normal inputs.

PR #1699, opened 2026-02-27
Fix PinnedMemoryResource IPC NUMA ID derivation

### LINKED ISSUE #1603: [BUG]: IPC-enabled pinned pool uses a fixed host NUMA node
### Is this a duplicate?

### Type of Bug

Runtime Error

### Component

cuda.core

### Describe the bug

`PinnedMemoryResource(ipc_enabled=True)` currently routes pinned IPC pools to host NUMA location ID `0` instead of the NUMA node closest to the active CUDA device. On multi-NUMA systems (especially when the current device is attached to a non-zero host NUMA node), this can cause pinned IPC pool creation/allocation failures or incorrect NUMA placement. The expected behavior is to select the current device's host NUMA ID (with a safe fallback only when the attribute/context is unavailable).

(Generated by Cursor(GPT-5.3 Codex Extra High) based on the production code changes under #1575)

### How to Reproduce

See nvbug 5823243

### Expected behavior

See test under #1575

### Operating System

_No response_

### nvidia-smi output

_No response_

### LINKED ISSUE #1710: Add preferred_location_type option and query property to ManagedMemoryResource
Managed memory pools support a "preferred location" that hints where allocations should reside (device, host, or a specific host NUMA node). The current `ManagedMemoryResourceOptions.preferred_location` field only accepts a device ordinal, `-1` for host, or `None` — there is no way to specify a NUMA node, and no way to query the resolved location after construction.

Prompted by [this review comment](https://github.com/NVIDIA/cuda-python/pull/1699#discussion_r2862921044) on PR #1699, which identified that `ManagedMemoryResource.device_id` was insufficient for querying the preferred location.

### Changes

Extend `ManagedMemoryResourceOptions` with a `preferred_location_type` field and add a `ManagedMemoryResource.preferred_location` query property. The default type, `None`, matches the previous behavior. Valid inputs are shown below:

| `preferred_location_type` | `preferred_location` | Driver location type | Meaning |
|---|---|---|---|
| `None` | `None` | `NONE` | No preference (legacy) |
| `None` | `-1` | `HOST` | Host (legacy) |
| `None` | `0` | `DEVICE` | Device 0 (legacy) |
| `"device"` | `0` | `DEVICE` | Device 0 |
| `"host"` | `None` | `HOST` | Host |
| `"host_numa"` | `3` | `HOST_NUMA` | NUMA node 3 |
| `"host_numa"` | `None` | `HOST_NUMA` | Current device's NUMA node (auto-resolved) |

### LINKED ISSUE #1720: [BUG]: DeviceMemoryResource should query driver for peer access state on non-owned pools
### Is this a duplicate?

### Type of Bug

Runtime Error

### Component

cuda.core

### Describe the bug

When `DeviceMemoryResource` is created without options (e.g., `DeviceMemoryResource(dev)`), it wraps the default device memory pool. This is a **non-owned** pool that is shared across all such references.

Currently, `cuda.core` initializes the internal `_peer_accessible_by` tracking variable to `()` (empty tuple), assuming no peer access. However, the actual driver-side peer access state of the default pool may differ if:

1. Other code has modified peer access on the shared default pool
2. Previous operations in the same process modified peer access and didn't clean up
3. The Python tests use the shared pool and leave it in a modified state

This causes a mismatch between `cuda.core`'s tracked state and the actual driver state, leading to:
- Incorrect `peer_accessible_by` property values
- Unexpected behavior when setting peer access (no-op if we think we're already in the target state)
- Test failures that depend on a clean initial peer access state

### How to Reproduce

```python
from cuda.core import Device, DeviceMemoryResource

dev = Device(0)

# Create a DMR with the default pool and modify peer access
dmr1 = DeviceMemoryResource(dev)
dmr1.peer_accessible_by = (1,)  # Enable peer access for device 1

# Create another DMR with the same default pool
dmr2 = DeviceMemoryResource(dev)

# dmr2._peer_accessible_by is (), but actual driver state has peer access for device 1
print(dmr2.peer_accessible_by)  # Returns () -- WRONG! Should reflect actual state
```

### Expected behavior

When wrapping a non-owned pool (the default device memory pool), `DeviceMemoryResource` should query the CUDA driver to determine the actual peer access state and initialize `_peer_accessible_by` accordingly.

This could be done using `cuMemPoolGetAccess` to query each peer device's access permissions on pool initialization.

### Workaround

Use owned pools by specifying options:
```python
from cuda.core import Device, DeviceMemoryResource, DeviceMemoryResourceOptions

dmr = DeviceMemoryResource(dev, DeviceMemoryResourceOptions())
```

This creates an owned pool with a known clean initial state.

### Operating System

N/A (affects all platforms)

### nvidia-smi output

N/A

Andy-Jost:
## Summary

- Fixes #1603: `PinnedMemoryResource(ipc_enabled=True)` hardcoded host NUMA ID `0`, causing failures on multi-NUMA systems when the active device is on a different NUMA node
- Closes #1710: Adds `preferred_location_type` option to `ManagedMemoryResourceOptions` and `preferred_location` query property to `ManagedMemoryResource`
- Fixes #1720: `DeviceMemoryResource.peer_accessible_by` returned stale results when wrapping a non-owned (default) memory pool
- Adds `numa_id` option to `PinnedMemoryResourceOptions` for explicit NUMA node selection
- Refactors `_MemPool` hierarchy to separate shared pool machinery from device-specific concerns

## Changes

**Bugfix (`_pinned_memory_resource.pyx`)**:
- When `ipc_enabled=True` and `numa_id` is not specified, derives the NUMA node from the current device's `host_numa_id` attribute (requires active CUDA context)
- Adds `numa_id: int | None = None` to `PinnedMemoryResourceOptions`
- Adds `numa_id` property to `PinnedMemoryResource`
- Removes `_check_numa_nodes` warning machinery in favor of proper NUMA node selection

**`ManagedMemoryResource` preferred location (`_managed_memory_resource.pyx`)**:
- Adds `preferred_location_type: str | None` field to `ManagedMemoryResourceOptions` with values `"device"`, `"host"`, `"host_numa"`, or `None` (legacy behavior)
- Adds `preferred_location` query property to `ManagedMemoryResource`
- Adds `_resolve_preferred_location` cdef function with CUDA 13 compile-time guard
- `device_id` property restored to backwards-compatible behavior (returns preferred device ordinal or `-1`)

**Refactor (`_MemPool` hierarchy)**:
- Moves `_dev_id`, `device_id`, and `peer_accessible_by` from `_MemPool` into `DeviceMemoryResource`
- Eliminates `_MemPoolOptions`; pool initialization refactored into freestanding `cdef` functions (`MP_init_create_pool`, `MP_init_current_pool`, `MP_raise_release_threshold`)
- Extracts `__init__` bodies into inline `cdef` helpers (`_DMR_init`, `_PMR_init`, `_MMR_init`)
- `ManagedMemoryResource.device_id` returns preferred device ordinal when location is a device, otherwise `-1`
- `PinnedMemoryResource.device_id` always returns `-1`

**Bugfix (`_device_memory_resource.pyx`)**:
- For non-owned pools, `peer_accessible_by` now always queries the CUDA driver instead of returning a cached value, ensuring multiple wrappers around the same shared pool see consistent state

## Test Coverage

- 4 new tests covering the `numa_id` behavior matrix: default without IPC, default with IPC, explicit NUMA ID, negative NUMA ID error
- 6 new tests for `ManagedMemoryResource` preferred location: default, device, host, host_numa, validation errors, NUMA auto-resolve failure
- 1 new test for `DeviceMemoryResource.peer_accessible_by` driver query on shared pools
- `skip_if_managed_memory_unsupported` updated to handle CUDA 12 builds
- Updated existing IPC tests to assert `numa_id` values
- All existing memory and IPC tests pass (pinned, device, managed, peer access, IPC)


cuda_core/cuda/core/_memory/_managed_memory_resource.pyx
```diff
@@ -64,40 +64,12 @@ cdef class ManagedMemoryResource(_MemPool):
...
-            super().__init__(device_id, opts_base)
-            _check_concurrent_managed_access()
-        ELSE:
-            raise RuntimeError("ManagedMemoryResource requires CUDA 13.0 or later")
+        _MMR_init(self, options)
+
+    @property
+    def device_id(self) -> int:
+        """Return -1. Managed memory migrates automatically and is not tied to a specific device."""
+        return -1
```
leofang, line 97: Q: Wouldn't this be a breaking change? In the old implementation, `device_id` was used to initialize `_MemPool. _dev_id`, which is used to back `.device_id`, but the new implementation returns `.device_id = -1` unconditionally. I understand we meant to say the pages are migratable (not pinned), but maybe there is a better way to restore the capability of querying the preferred location?

Andy-Jost: I didn't mean for this to be reviewed yet.  Now marked as Draft.

In the old API, `device_id` has an overloaded meaning. For `DeviceMemoryResource`, it refers to the device owning memory. In `ManagedMemoryResource`, it indicates the preferred location if that location is a device. That is not expressive enough because there are four possible types of preferred location:

1. A device
2. The host
3. A host NUMA node
4. None

I'm still considering options, but perhaps we should add a new property, `preferred_location`, for managed pools. There is, unfortunately, no uniform way to refer to all preferred locations, so we might need to resort to something like the following:

```
ManagedMemoryResource.preferred_location ->
    ('device_id', 0)  # device 0
    ('host', None)    # host preferred
    ('numa_id', 0)    # NUMA node 0
    None              # no preferred location
```

I'm not in a position to say whether we need to retain backwards compatibility for this attribute. Do you think so? If yes, then I can retain the original meaning, possibly with a deprecation warning.

Andy-Jost: I created #1710 to track this as a separate issue, but I'm including it in this PR.

Andy-Jost: I restored the old behavior where `MMR.device_id` returns the preferred location if it is a device and -1 otherwise. I'd prefer to unconditionally return -1 if the breaking change is tolerable. A middle ground is to issue a deprecation warning. What do you think @leofang?

cuda_core/cuda/core/_memory/_device_memory_resource.pyx
```diff
@@ -122,27 +128,26 @@ cdef class DeviceMemoryResource(_MemPool):
...
+        self._dev_id = cydriver.CU_DEVICE_INVALID
+        self._peer_accessible_by = ()
 
-        super().__init__(dev_id, opts_base)
+    def __init__(self, device_id: Device | int, options=None):
+        _DMR_init(self, device_id, options)
+
+    def __dealloc__(self):
+        try:
+            self.close()
```
cpcloud: You're not supposed to call methods in `__dealloc__`: https://cython.readthedocs.io/en/stable/src/userguide/special_methods.html#finalization-methods-dealloc-and-del

> In particular, don’t call any other methods of the object or do anything which might cause the object to be resurrected. It’s best if you stick to just deallocating C data.

- `_peer_accessible_by` is a Python object
- `_DMR_set_peer_accessible_by` looks like it's modifying the object based on the way it's being invoked.

cuda_core/cuda/core/_memory/_device_memory_resource.pyx
```diff
@@ -226,6 +262,82 @@ cdef class DeviceMemoryResource(_MemPool):
         return False
 
 
+cdef inline _DMR_set_peer_accessible_by(DeviceMemoryResource self, devices):
```
cpcloud, line 281: This seems like something that should almost certainly never be invoked in `__dealloc__`. It's reading from _and_ writing to the Python object whose `__dealloc__` method is being called.

cuda_core/cuda/core/_memory/_device_memory_resource.pyx
mdboom: Sounds like a job for `__del__` and also adding a context manager.

cuda_core/cuda/core/_memory/_device_memory_resource.pyx
Andy-Jost: Thanks, yeah this was a mess. I removed all peer-access-clearing logic from Python. The real problem was that certain tests were using the default memory pool, which is persistent and should keep its peer access settings. Those tests incorrectly assumed the beginning state should be "no peer access." I updated those to simply use fresh pools.

Andy-Jost:
@cpcloud Please take another look

cuda_core/cuda/core/_memory/_device_memory_resource.pyx
Andy-Jost: This was to work around a driver bug (5698116). When a mempool is destroyed and a new one allocated, if the new pool happens to reuse the previous memory address, it incorrectly inherits the peer access attributes of the destroyed pool. This code is all about clearing the peer access attributes on destruction to avoid that.

The real fix is in C++ under `resource_handles.*` and should only be applied to owned pools created by cuda.core. The changes here, in Python, were a misadventure. This was resetting peer access for the *shared* default pool, which is not being destroyed but just released.

PR #1701, opened 2026-02-27
Refactor graphics interop around RAII-mapped buffers

rparolin:
## Summary

- Reworks graphics interop around the existing handle/RAII pattern instead of making `GraphicsResource` inherit from `Buffer`
- Keeps `GraphicsResource` responsible for registration/unregistration only; `map()` now returns a mapped `Buffer`
- Adds a mapped-graphics device pointer handle whose deleter captures both the `GraphicsResourceHandle` and `StreamHandle` and calls `cuGraphicsUnmapResources(...)` on release
- Preserves `with resource.map(stream=s) as buf:` by giving `Buffer` context-manager support
- Supports both `with GraphicsResource.from_gl_buffer(vbo) as resource:` for automatic resource cleanup and `with GraphicsResource.from_gl_buffer(vbo, stream=s) as buf:` for register-and-map-in-one-step usage
- Restores default-stream compatibility for `map()` / `unmap()` and keeps `GraphicsResource.handle` as the raw `CUgraphicsResource` handle

## Motivation

Review feedback pointed toward Andy Jost's `resource_handles` direction rather than turning `GraphicsResource` into a mutable `Buffer`.

This version follows the same ownership model used elsewhere in `cuda_core`: the mapped pointer owns unmap lifetime, and the resource registration owns unregister lifetime. That keeps stream-ordered cleanup in the handle layer, avoids changing `GraphicsResource.handle` semantics based on mapping state, and removes the need to treat an unmapped resource as a partially valid `Buffer`.

## Key changes

- **Handle-based mapped buffers**: `map()` creates a real `Buffer` from a new mapped-graphics device pointer handle. The handle deleter captures the graphics resource and mapping stream, and unmaps on release.
- **GraphicsResource stays a resource owner**: `GraphicsResource` no longer inherits from `Buffer`; it tracks registration state, forwards `unmap()` / `close()` to the active mapped buffer, and keeps a strong reference to that buffer so `is_mapped`, `unmap()`, and double-map protection remain correct even if callers drop their own Python reference.
- **Context-manager behavior**:
  - `with resource.map(stream=s) as buf:` still works and automatically unmaps on exit.
  - `with GraphicsResource.from_gl_buffer(vbo, stream=s) as buf:` registers, maps on entry, and unregisters on exit.
  - `with GraphicsResource.from_gl_buffer(vbo) as resource:` scopes the registration lifetime and closes automatically on exit, avoiding a manual `resource.close()`.
- **Compatibility cleanup**: `map()` / `unmap()` again accept `stream=None` and use the default stream, and `GraphicsResource.handle` once again exposes the raw graphics-resource handle.

### DIFF
--- a/cuda_core/tests/test_graphics.py
+++ b/cuda_core/tests/test_graphics.py
@@ -318,6 +318,16 @@ class TestMapUnmap:
                 assert isinstance(buf, Buffer)
                 assert buf.handle != 0
 
+    def test_chained_map_context_manager_unmaps(self):
+        with _gl_context_and_buffer(nbytes=4096) as (gl_buf, _):
+            stream = _create_stream()
+            with GraphicsResource.from_gl_buffer(gl_buf, flags="write_discard").map(stream=stream) as buf:
+                assert isinstance(buf, Buffer)
+                assert buf.handle != 0
+                assert buf.size > 0
+            assert buf.handle == 0
+            assert buf.size == 0
+
     def test_map_with_stream(self):
         with _gl_context_and_buffer(nbytes=4096) as (gl_buf, nbytes):
             stream = _create_stream()


cuda_core/cuda/core/_graphics.pyx
```diff
@@ -292,42 +264,63 @@ cdef class GraphicsResource:
...
+            Accepted for compatibility with :meth:`Buffer.close` but not
+            used for the graphics unmap/unregister operations.
         """
         cdef cydriver.CUgraphicsResource raw
         cdef cydriver.CUstream cy_stream
         if not self._handle:
             return
         if self._mapped:
-            # Best-effort unmap before unregister
+            # Best-effort unmap before unregister (use stream 0 as fallback)
```
cpcloud: Potential stream-ordering bug here: `map()`/`unmap()` now require an explicit stream, but `close()` always unmaps on stream 0 and ignores both `_map_stream` and the optional `close(stream=...)` arg.

If a resource was mapped on a non-default stream and then closed while mapped, unmap can be issued on the wrong stream, which may break ordering guarantees with in-flight work. Could we unmap using `_map_stream` (or the passed `stream`, with a clear fallback policy) instead of hard-coding stream 0?

cuda_core/cuda/core/_graphics.pyx
```diff
@@ -292,42 +264,63 @@ cdef class GraphicsResource:
...
                 cydriver.cuGraphicsUnmapResources(1, &raw, cy_stream)
             )
         self._mapped = False
+        # Clear Buffer fields
+        self._h_ptr.reset()
+        self._size = 0
+        self._map_stream = None
+
+    def __enter__(self):
+        return self
```
cpcloud: `__enter__` currently returns `self` even when the resource is not mapped. That allows patterns like `with GraphicsResource.from_gl_buffer(vbo) as buf:` (without `stream=`), where `buf.handle`/`buf.size` are not valid.

Could we either (a) raise in `__enter__` when `not self._mapped`, or (b) require/perform mapping for the context-manager path to avoid this silent footgun?

cpcloud:
ping @rparolin. happy to fix this up if you want.

leofang:
> Thanks for the refactor here — the `GraphicsResource(Buffer)` direction looks good overall.

Yeah and thinking about it more, I suspect this is a must:
- If GL buffer registration is context-independent: This PR makes the UX nicer
- If GL buffer registration is locked to the current CUDA context: We need to track the underlying device/context, so we either inherit from `Buffer` (this PR) or add `Buffer.from_gl_buffer()` constructor (and keep `GraphicsResource` internal). 
   - From the viewpoint of staying consistent across cuda.core, the latter might be worth considering too, but one challenge is that `Buffer` today does not have the `map`/`unmap` concept. 
   - I would be happy for you @rparolin @Andy-Jost @cpcloud to brainstorm and I can review later 🙂 

I will ask internally if it's context-independent or not, but it's not a blocker.

Andy-Jost:
Thanks for the discussion here. I'd like to propose an alternative approach that I think addresses both of Phillip's concerns and Leo's observation about `Buffer` not having `map`/`unmap`.

**Proposal: Use the `resource_handles` module instead of modifying `Buffer`**

We already have a C++ handle layer (`cuda/core/_cpp/resource_handles.hpp`) that provides RAII-based lifetime management with `std::shared_ptr` custom deleters. This pattern already solves the exact problems raised here:

1. **Stream capture**: `DevicePtrHandle` already captures the allocation stream in its deleter and uses it for `cuMemFreeAsync`. We can do the same for graphics resources - capture the map stream and use it for unmap.

2. **Dependency tracking**: Handles can hold references to other handles, preventing out-of-order destruction. A mapped buffer handle can hold a reference to the `GraphicsResourceHandle`, ensuring the graphics resource outlives the mapped pointer.

3. **Arbitrary release actions**: The shared_ptr deleter can chain actions - e.g., unmap then unregister.

**Concrete design:**

- `GraphicsResource.map(stream)` returns a `Buffer` (or just the existing `DevicePtrHandle` under the hood)
- The returned handle's deleter captures:
  - The `GraphicsResourceHandle` (keeps it alive)
  - The map `StreamHandle` (for stream-ordered unmap)
- When the mapped buffer handle is released, the deleter calls `cuGraphicsUnmapResources` on the captured stream
- `GraphicsResource` itself doesn't need to track `_mapped` state - the existence of mapped buffer handles implies mapped state

**Benefits:**

- `Buffer` stays unchanged (no `map`/`unmap` concept needed)
- Stream ordering is correct by construction (cpcloud's concern 1)
- No invalid `__enter__` state possible (cpcloud's concern 2) - you either have a valid mapped handle or you don't
- Dependency tracking prevents use-after-free
- Consistent with how we handle `DeviceMemoryResource` allocations

This is the same pattern we use for memory pool allocations, where the pointer handle captures the pool handle and stream for proper cleanup ordering.

Thoughts?

cpcloud:
+1 to @Andy-Jost's design. Looks solid.

leofang:
FYI
> I will ask internally if it's context-independent or not

it is context-dependent.

> **Proposal: Use the `resource_handles` module instead of modifying `Buffer`**

Modifying `Buffer` in what way? Sorry it's not clear to me.

Can we sketch a bit how the user code would look like with this proposal? The changes in this PR were motivated by the sketch in here: https://github.com/NVIDIA/cuda-python/pull/1608#discussion_r2862666297. Rob asked the right question there. Inheritance is probably not necessary, as long as the same UX can be delivered.

cuda_core/cuda/core/_graphics.pyx
cpcloud: `close(stream=...)` still hard-codes stream 0 here, so a resource mapped on another stream can be unmapped out of order. Please unmap on `_map_stream` or the explicit `stream` argument instead of ignoring both.

cuda_core/cuda/core/_graphics.pyx
cpcloud: `__enter__` still returns `self` even when the resource was never mapped, so `with GraphicsResource.from_gl_buffer(vbo) as buf:` silently yields a "Buffer" whose `handle`/`size` are invalid. Please raise when `not self._mapped` (or make the context-manager path perform the mapping).

cpcloud:
@rparolin Do you want to fire off an agent to implement @Andy-Jost's design? Probably could make pretty short work of it.

rparolin:
@cpcloud made the requested changes and aligned with @Andy-Jost recommendations. Please re-review. Thanks!

rparolin:
> Re-reviewed the latest graphics follow-up changes. The earlier context-manager and stream-ordering concerns look addressed, and passes in my environment.

Thanks!

Issue #1707, opened 2026-03-02
[BUG]: CuPy fails to auto-detect CUDA when CUDA_PATH is unset in containers


jcfr:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Runtime Error

### Component

cuda.pathfinder

### Describe the bug

CuPy fails with the following error when `CUDA_PATH` (and `CUDA_HOME`) are unset, even when CUDA Toolkit is installed at the default location `/usr/local/cuda`:

```
[CUPY_DEBUG_LIBRARY_LOAD] CUDA headers not found via cuda-pathfinder
RuntimeError: Failed to auto-detect CUDA root directory. Please specify `CUDA_PATH` environment variable if you are using CUDA versions not yet supported by CuPy.
```

This was discovered in one of the Holoscan containers (based on the TensorRT image) and occurs in containerized environments where CUDA is installed at `/usr/local/cuda` but environment variables are not set. CuPy's own `_get_cuda_path()` in `cupy/_environment.py` ([source](https://github.com/cupy/cupy/blob/2ba99ee2c2a13f64378d7349fe670a5b1d4b52c0/cupy/_environment.py#L185-L187)) includes a fallback to `/usr/local/cuda`:

```python
# Use typical path
if os.path.exists('/usr/local/cuda'):
    return '/usr/local/cuda'
```

The CuPy header detection path uses `cuda-pathfinder`'s `find_nvidia_header_directory('cudart')` (called from `_get_include_dir_from_conda_or_wheel`), which does **not** have this fallback.

**Root cause:** `cuda-pathfinder`'s `find_nvidia_header_directory` (used by CuPy in `_get_include_dir_from_conda_or_wheel`) only checks:
1. NVIDIA Python wheels (site-packages)
2. Conda layout (`CONDA_PREFIX`)
3. `CUDA_HOME` / `CUDA_PATH` env vars

It does **not** fall back to `/usr/local/cuda`, unlike CuPy's `get_cuda_path()`.

### How to Reproduce

[...]

PR #956, opened 2025-09-09
`cuda.pathfinder._find_nvidia_header_directory()`: add support for CTK libs

rwgk:
Closes #832, #954

Make the `cuda.pathfinder._find_nvidia_header_directory` API public (by removing the leading underscore) and extend the function to also support CTK library headers.

See new `find_nvidia_header_directory()` docstring for details:

https://nvidia.github.io/cuda-python/cuda-pathfinder/1.2.3/generated/cuda.pathfinder.find_nvidia_header_directory.html

PR #1708, opened 2026-03-02
fix: add /usr/local/cuda fallback for CTK header discovery

jcfr:
When `CUDA_PATH`/`CUDA_HOME` are unset, fall back to `/usr/local/cuda` on non-Windows platforms. Aligns with CuPy's `get_cuda_path()` behavior and fixes CuPy header detection in containers (e.g. Holoscan).

Follow-up of PR:
* #956

## Description

closes https://github.com/NVIDIA/cuda-python/issues/1707

Add a `/usr/local/cuda` fallback in `find_nvidia_header_directory` (used by CuPy's `_get_include_dir_from_conda_or_wheel`) when `CUDA_PATH` and `CUDA_HOME` are unset. This matches the behavior of CuPy's `_get_cuda_path()` in `cupy/_environment.py`, which already falls back to `/usr/local/cuda` for the typical Linux container layout.

**Problem:** CuPy fails with "Failed to auto-detect CUDA root directory" when running `RawKernel` in containers where CUDA is at `/usr/local/cuda` but env vars are not set. The header detection path (via cuda-pathfinder) only checked site-packages, conda, and env vars—not the default install path.

**Solution:** On non-Windows platforms, if headers are not found via existing checks, try `/usr/local/cuda/include` when the directory exists.

## Checklist

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_headers/find_nvidia_headers.py
+++ b/cuda_pathfinder/cuda/pathfinder/_headers/find_nvidia_headers.py
@@ -108,9 +108,12 @@ def _find_ctk_header_directory(libname: str) -> LocatedHeaderDir | None:
             return LocatedHeaderDir(abs_path=result, found_via="CUDA_HOME")
 
     # Fallback: typical system install path (matches CuPy's get_cuda_path())
-    if not IS_WINDOWS and os.path.exists('/usr/local/cuda'):
-        if result := _locate_based_on_ctk_layout(libname, h_basename, '/usr/local/cuda'):
-            return LocatedHeaderDir(abs_path=result, found_via="system_default")
+    if (
+        not IS_WINDOWS
+        and os.path.exists("/usr/local/cuda")
+        and (result := _locate_based_on_ctk_layout(libname, h_basename, "/usr/local/cuda"))
+    ):
+        return LocatedHeaderDir(abs_path=result, found_via="system_default")
 
     return None
 


leofang (changes requested):
As noted in https://github.com/NVIDIA/cuda-python/issues/1707#issuecomment-3981720201 we need more triage; from the pathfinder perspective we purposely do not want to assume there's a default location to fall back to, because it'd cover up a class of bugs but open up another class. 

Blocking the review/merge of this PR until we finish triaging on the CuPy side.

rwgk:
> As noted in [#1707 (comment)](https://github.com/NVIDIA/cuda-python/issues/1707#issuecomment-3981720201) we need more triage; from the pathfinder perspective we purposely do not want to assume there's a default location to fall back to, because it'd cover up a class of bugs but open up another class.

I agree this is likely to cause at least as much trouble as it hopes to avoid.

@jcfr as of PR #1595 we have another mechanism that we could leverage, without introducing a new extremely specific fallback: we could use the CTK root canary probe to anchor the fallback search for headers. This will work as long as `libcudart.so` is found via the dynamic loader system search; usually that's through `/etc/ld.so.conf.d/cuda-*.conf`, which is created by the CTK installer.

cpcloud:
Can we close this PR?

leofang:
I converted it to draft, please give us some time. We are short in bandwidth for CuPy this week 🥲

cpcloud:
I put up #1731, which should address the issue using the same approach that the analogous library search problem uses.

I think we can close this PR out.

rwgk:
@jcfr We'll release #1731 with cuda-pathfinder v1.4.1 (planned for today**). Could you please try it out once it's up? Does it resolve the issue in your environments?

Closing this PR.

**EDIT: https://github.com/NVIDIA/cuda-python/releases/tag/cuda-pathfinder-v1.4.1

PR #1729, opened 2026-03-06
Add explicit CUDA graph construction API

Andy-Jost:
## Summary

- Implements explicit (non-stream-capture) CUDA graph construction via `GraphDef` and a `Node` class hierarchy with 14 concrete node types: `EmptyNode`, `KernelNode`, `AllocNode`, `FreeNode`, `MemsetNode`, `MemcpyNode`, `ChildGraphNode`, `HostCallbackNode`, `EventRecordNode`, `EventWaitNode`, `IfNode`, `IfElseNode`, `WhileNode`, `SwitchNode`
- RAII lifetime management for graph resources (events, kernels) via CUDA User Objects, ensuring resources survive graph cloning and GC
- Reverse-lookup registries for Event and Kernel handles (`HandleRegistry` template), enabling owning handle recovery and full metadata round-tripping on reconstruction
- Event metadata (timing, IPC, busy-wait, device, context) stored in C++ `EventBox` with accessor functions, eliminating metadata loss on handle reconstruction
- Kernel `from_handle` simplified: registry-first lookup for cuda.core-created kernels, `_keepalive` reference for foreign kernels

## Changes

- `_graphdef.pyx` / `_graphdef.pxd`: Full explicit graph API — `GraphDef`, `Node` hierarchy, `Condition`, fluent builder methods, `__repr__`, `__eq__`/`__hash__`/`__weakref__` protocols, CUDA User Object attachment for event/kernel lifetime
- `resource_handles.{hpp,cpp}`: `GraphHandle` with owning/non-owning variants and parent-capture for child graphs; `GraphNodeHandle` (RAII node handle with parent graph reference); `HandleRegistry<Key, Handle, Hash>` template for thread-safe reverse lookup; Event/Kernel registries with `weak_ptr` tracking; `EventBox` metadata fields and `get_event_*` accessors; `get_kernel_library` accessor; IPC cache refactored to use `HandleRegistry`
- `_resource_handles.{pxd,pyx}`: Cython declarations for graph node handles, event/kernel registries, and accessor functions
- `_event.{pxd,pyx}`: Consolidated factory methods (`_init`, `_from_handle`); metadata properties delegate to `EventBox` via C++ accessors
- `_module.{pxd,pyx}`: `Kernel.from_handle` with registry lookup and `_keepalive` for foreign kernels; renamed `_from_obj` to `_from_handle`

## Test Coverage

- `test_explicit.py`: 176 unit tests (topology, node types, attributes, execution)
- `test_explicit_integration.py`: 6 integration tests (heat diffusion, bisection, switch dispatch) covering all 14 node types
- `test_explicit_lifetime.py`: Lifetime tests for child graphs, conditional body graphs, events, kernels, registry handle recovery, and graph node reconstruction
- `test_explicit_errors.py`: Error handling and edge-case tests
- `test_module.py`: Library mismatch warning and foreign kernel `from_handle` tests
- `test_object_protocols.py`: Protocol tests for new types
- Conditional node tests require CC >= 9.0 (skipped on older hardware)


cuda_core/cuda/core/_memory/_buffer.pyx
Andy-Jost, line 251: `Buffer.fill` was reworked to extract the input processing, so memset nodes could share that.

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -318,47 +364,98 @@ StreamHandle get_per_thread_stream() {
 namespace {
 struct EventBox {
     CUevent resource;
+    bool timing_disabled;
+    bool busy_waited;
+    bool ipc_enabled;
+    int device_id;
+    ContextHandle h_context;
```
Andy-Jost, line 371: These properties are set at event creation time and cannot be queried through the driver API. Moreover, graph-attached events are returned from the driver as plain `CUevent` handles, and reconstructing the Cython Event object from one of those would lose this information.

The solution is to move the property metadata into C++ and set up a reverse look-up so that the driver-returned `CUevent` can be used to retrieve the managing `shared_ptr`, which holds this `EventBox`.

Graph-attached kernels are handled similarly.

cuda_core/cuda/core/_graph/_graphdef.pyx
```diff
@@ -0,0 +1,2081 @@
...
+            cydriver.CU_USER_OBJECT_NO_DESTRUCTOR_SYNC)
+        if ret == cydriver.CUDA_SUCCESS:
+            ret = cydriver.cuGraphRetainUserObject(
+                graph, user_obj, 1, cydriver.CU_GRAPH_USER_OBJECT_MOVE)
+            if ret != cydriver.CUDA_SUCCESS:
+                cydriver.cuUserObjectRelease(user_obj, 1)
+    if ret != cydriver.CUDA_SUCCESS:
+        if user_obj == NULL:
+            destroy(ptr)
+        HANDLE_RETURN(ret)
```
Andy-Jost, line 150: Resources like host functions, event handles, and kernel handles are placed into CUDA user objects, which are like capsules. This way, the lifetimes are properly tied to graphs they appear in, and they follow graphs through cloning steps.

cuda_core/cuda/core/_graph/_graphdef.pyx
```diff
@@ -0,0 +1,2081 @@
...
+    Notes
+    -----
+    - IPC (inter-process communication) is not supported for graph
+      memory allocation nodes per CUDA documentation.
+    - The allocation uses the device's default memory pool.
+    """
+
+    device: int | Device | None = None
+    memory_type: str = "device"
+    peer_access: list | None = None
```
Andy-Jost, line 275: I made this options class to match the way memory resources are created. But other graph operations just take arguments, and there are only three arguments wrapped here. I'm not sure having this this is better than just adding these arguments to the `alloc()` function.

cuda_core/cuda/core/_graph/_graphdef.pyx
```diff
@@ -0,0 +1,2081 @@
...
+        cdef Node n = Node.__new__(Node)
+        n._h_node = create_graph_node_handle(<cydriver.CUgraphNode>NULL, self._h_graph)
+        return n
+
+    def alloc(self, size_t size, options: GraphAllocOptions | None = None) -> AllocNode:
+        """Add an entry-point memory allocation node (no dependencies).
+
+        See :meth:`Node.alloc` for full documentation.
+        """
+        return self._entry.alloc(size, options)
```
Andy-Jost, line 323: I would rather use *args, **kwargs in these forwarding functions, but I have a feeling someone would complain about IDE integration.

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -160,6 +163,49 @@ class GILAcquireGuard {
...
+// Maps raw CUDA handles (CUevent, CUkernel, etc.) back to their owning
+// shared_ptr so that _ref constructors can recover full metadata.
+// Uses weak_ptr to avoid preventing destruction.
+// ============================================================================
+
+template<typename Key, typename Handle, typename Hash = std::hash<Key>>
+class HandleRegistry {
+public:
+    void register_handle(const Key& key, const Handle& h) {
+        std::lock_guard<std::mutex> lock(mutex_);
```
rparolin, line 178: You can write `std::lock_guard lock(mutex_);` CTAD was included in C++17. Removing a code ripple if you want to change the type of the mutex later.

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -160,6 +163,49 @@ class GILAcquireGuard {
...
+    }
+
+    void unregister_handle(const Key& key) noexcept {
+        try {
+            std::lock_guard<std::mutex> lock(mutex_);
+            auto it = map_.find(key);
+            if (it != map_.end() && it->second.expired()) {
+                map_.erase(it);
+            }
+        } catch (...) {}
```
rparolin, line 189: We should probably provide some feedback to the user that unregister_handle operation failed.

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -160,6 +163,49 @@ class GILAcquireGuard {
...
+    }
+
+    Handle lookup(const Key& key) {
+        std::lock_guard<std::mutex> lock(mutex_);
+        auto it = map_.find(key);
+        if (it != map_.end()) {
+            if (auto h = it->second.lock()) {
+                return h;
+            }
+            map_.erase(it);
```
rparolin, line 199: I don't think you want to erase the found element in your `lookup` function.

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -160,6 +163,49 @@ class GILAcquireGuard {
...
+        }
+        return {};
+    }
+
+private:
+    std::mutex mutex_;
+    std::unordered_map<Key, std::weak_ptr<typename Handle::element_type>, Hash> map_;
```
rparolin, line 206: Consider giving this `std::unordered_map` a name so that it can reused in dependent code.

cuda_core/cuda/core/_cpp/resource_handles.cpp
rparolin, line 371: Consider memory layout density -- order members primarily by size and alignment.

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -798,25 +877,96 @@ LibraryHandle create_library_handle_ref(CUlibrary library) {
...
+    auto box = std::make_shared<const KernelBox>(KernelBox{kernel, h_library});
+    KernelHandle h(box, &box->resource);
+    kernel_registry.register_handle(kernel, h);
+    return h;
 }
 
-KernelHandle create_kernel_handle_ref(CUkernel kernel, const LibraryHandle& h_library) {
-    auto box = std::make_shared<const KernelBox>(KernelBox{kernel, h_library});
+KernelHandle create_kernel_handle_ref(CUkernel kernel) {
+    if (auto h = kernel_registry.lookup(kernel)) {
```
rparolin, line 907: `auto` is deducing a `Handle` which we are implicitly converting to bool for the if-statement check. Is this intentional? Or do you need to call a functoin to check of the handle is valid?

Consider using trailing new if-statement:
`
if (auto h = kernel_registry.lookup(kernel); h->isValid()) { .... }
`

rparolin:
@Andy-Jost I'd definitely start planning how you want to break this PR up into smaller pieces to make it easier/faster to review.

cuda_core/cuda/core/_cpp/resource_handles.cpp
Andy-Jost: Yes. This would be an ideal place to use a logger. Do we have plans to add one to cuda.core? I'm being cautious here because this is typically called from a destructor.

cuda_core/cuda/core/_cpp/resource_handles.cpp
Andy-Jost: The found element is an expired `std::weak_ptr`

cuda_core/cuda/core/_cpp/resource_handles.cpp
Andy-Jost: I think `operator bool()` is a common way to check a `shared_ptr` being non-null.

Andy-Jost:
Closing, as this was split into two smaller PRs.

PR #1733, opened 2026-03-06
Improve optional CUDA module import handling for NVVM and nvJitLink

### LINKED ISSUE #980: [ENH]: Robust imports of optional dependencies (nvvm, nvJitLink)

cpcloud:
Fixes #980

## Summary
- add a public `cuda.pathfinder.optional_cuda_import()` helper that treats only truly optional-module failures as unavailable while re-raising nested `ModuleNotFoundError`
- switch `cuda.core` NVVM and nvJitLink availability checks to use the shared helper, including dynamic library probing behavior
- make `cuda-pathfinder` a direct dependency of `cuda-core`, and add targeted tests/docs/release-note updates in both subpackages

## Test plan
- [x] `cd cuda_pathfinder && pixi run -e default pytest tests/test_optional_cuda_import.py -q`
- [x] `cd cuda_core && pixi run -e cu13 pytest tests/test_optional_dependency_imports.py -q`

### DIFF
--- a/cuda_core/cuda/core/_linker.pyx
+++ b/cuda_core/cuda/core/_linker.pyx
@@ -662,9 +662,8 @@ def _decide_nvjitlink_or_driver() -> bool:
         if _nvjitlink_has_version_symbol(nvjitlink):
             _use_nvjitlink_backend = True
             return False  # Use nvjitlink
-        warn_detail = "too old (<12.3)"
         warn_txt = (
-            f"{'nvJitLink*.dll' if sys.platform == 'win32' else 'libnvJitLink.so*'} is {warn_detail}."
+            f"{'nvJitLink*.dll' if sys.platform == 'win32' else 'libnvJitLink.so*'} is too old (<12.3)."
             f" Therefore cuda.bindings.nvjitlink is not usable and {warn_txt_common} nvJitLink."
         )
 
--- a/cuda_core/pyproject.toml
+++ b/cuda_core/pyproject.toml
@@ -47,7 +47,7 @@ classifiers = [
     "Environment :: GPU :: NVIDIA CUDA :: 13",
 ]
 dependencies = [
-    "cuda-pathfinder >=1.1",
+    "cuda-pathfinder >=1.4.2",
     "numpy",
 ]
 
--- a/cuda_pathfinder/docs/source/release/1.4.1-notes.rst
+++ b/cuda_pathfinder/docs/source/release/1.4.1-notes.rst
@@ -33,11 +33,6 @@ Highlights
   libraries: site-packages, Conda, then CUDA_HOME/CUDA_PATH.
   (`PR #1690 <https://github.com/NVIDIA/cuda-python/pull/1690>`_)
 
-* Add ``optional_cuda_import()`` to support robust optional imports of CUDA
-  Python modules. It returns ``None`` when the requested module is absent or a
-  probe hits ``DynamicLibNotFoundError``, while still re-raising unrelated
-  ``ModuleNotFoundError`` exceptions (for missing transitive dependencies).
-
 * Fix site-packages search order for virtual environments created with
   ``--system-site-packages``. The search now correctly prioritizes the venv's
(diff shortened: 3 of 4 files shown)


cuda_pathfinder/docs/source/release/1.4.1-notes.rst
```diff
@@ -33,6 +33,11 @@ Highlights
   libraries: site-packages, Conda, then CUDA_HOME/CUDA_PATH.
   (`PR #1690 <https://github.com/NVIDIA/cuda-python/pull/1690>`_)
 
+* Add ``optional_cuda_import()`` to support robust optional imports of CUDA
```
rwgk: Ah, did you want to add this into the 1.4.1 release?

You could start the 1.4.2 release notes in this PR, since you have this paragraph already.

cuda_core/pyproject.toml
```diff
@@ -47,6 +47,7 @@ classifiers = [
     "Environment :: GPU :: NVIDIA CUDA :: 13",
 ]
 dependencies = [
+    "cuda-pathfinder >=1.1",
```
rwgk: I think this needs to be `>=1.4.2`? Because this PR makes cuda_core depending on the optional_cuda_import feature, which will only be in 1.4.2.

cuda_core/cuda/core/_linker.pyx
```diff
@@ -649,21 +650,19 @@ def _decide_nvjitlink_or_driver() -> bool:
...
 
-        try:
-            if _nvjitlink_has_version_symbol(nvjitlink):
-                _use_nvjitlink_backend = True
-                return False  # Use nvjitlink
-        except RuntimeError:
-            warn_detail = "not available"
-        else:
-            warn_detail = "too old (<12.3)"
+        if _nvjitlink_has_version_symbol(nvjitlink):
```
rwgk, line 662: Did you remove the `try`-`except RuntimeError` around this call intentionally?

If yes, then `warn_detail` below could be eliminated (i.e. simply inline the literal in the `warn_txt` assignment).

cuda_core/pyproject.toml
cpcloud: Kind of a chicken-and-egg problem, because the solver will fail to solve until 1.4.2 is released. So I think this needs to remain as is for now.

cuda_core/cuda/core/_program.pyx
```diff
@@ -461,8 +462,8 @@ class ProgramOptions:
 # =============================================================================
 
 # Module-level state for NVVM lazy loading
-cdef object_nvvm_module = None
-cdef bint _nvvm_import_attempted = False
+_nvvm_module = None
+_nvvm_import_attempted = False
```
leofang, line 466: Making them `cdef` helps us hide the objects from curious Python users.

cuda_pathfinder/cuda/pathfinder/_optional_cuda_import.py
```diff
@@ -0,0 +1,43 @@
...
+from __future__ import annotations
+
+import importlib
+from collections.abc import Callable
+from types import ModuleType
+
+from cuda.pathfinder._dynamic_libs.load_dl_common import DynamicLibNotFoundError
+
+
+def optional_cuda_import(
```
leofang, line 13: Q: Does this have to be a public API? Can we keep it internal instead?

Issue #1580, opened 2026-02-05
Python latency testing & benchmarking


leofang:
Add Python latency benchmarks that ensure we are achieving our target of < 1us overhead when comparing against the equivalent C++ implementation.

leofang:
@danielfrg will take over the learnings from https://github.com/NVIDIA/cuda-python/issues/1579#issuecomment-3880678588 and work out a benchmark suite that can target 
- C++
- cuda-bindings
- cuda-core

to cover a range of most common tasks.

rparolin:
@danielfrg This issue assigned to cuda.core-0.7.0 currently which is schedule to ship on April 6th.  Are you still confident you can complete for this deadline?

danielfrg:
I opened up https://github.com/NVIDIA/cuda-python/pull/1736 and its ready with the the first benchmark but most importantly the template i think we can follow. If we get that reviewed this week i am pretty confident I can add/migrate other benchmarks easily

leofang:
Push out to the milestone since this is not critical/blocking.

PR #1736, opened 2026-03-06
cuda.bindings latency benchmarks

danielfrg:
## Description

closes https://github.com/NVIDIA/cuda-python/issues/1580

## Description

closes https://github.com/NVIDIA/cuda-python/issues/1580

@leofang @mdboom I migrated one benchmark from the pytest suite to use pyperf and added a C++ equivalent.

- Added a small benchmark discovery to find `bench_*.py` files with `bench_*()` functions
- Uses `bench_time_func`
- C++ benchmarks output pyperf-compatible JSON so both sides can be analyzed with the same `pyperf stats` / `pyperf hist` commands.
- The readme explain how to run it on the different envs using pixi

The benchmark is `cuPointerGetAttribute`, both Python and C++ call the same driver API with error checking.

These are one set of results for Python and C++ in my system, so we are ok under the <1us. They dont run the same warmup and runs for each, i still need to finish that but just to give you an idea.

```
# Python (pyperf bench_time_func)
bindings.pointer_attributes.pointer_get_attribute: Mean +- std dev: 603 ns +- 25 ns

# C++ (driver API baseline)
cpp.pointer_attributes.pointer_get_attribute: Mean +- std dev: 29 ns +- 1 ns
```

I still need to work on matching params for all the benchmarks and so on and so on but wanted to get feedback first if this looks fine to keep going.

## Checklist


cuda_bindings/benchmarks/benchmarks/bench_pointer_attributes.py
```diff
@@ -0,0 +1,17 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
mdboom, line 1: Maybe move this file into a `benchmarks` subdirectory so as not to clutter the top-level of the `cuda_bindings/benchmarks` directory.

cuda_bindings/benchmarks/benchmarks/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+
+if(NOT CUDA_DRIVER_INCLUDE_DIR)
+    message(FATAL_ERROR "Could not find cuda.h. Ensure CUDA_HOME is set or install cuda-crt-dev.")
+endif()
+
+if(NOT CUDA_DRIVER_LIBRARY)
+    message(FATAL_ERROR "Could not find libcuda. Ensure the NVIDIA driver is installed.")
+endif()
+
+add_executable(bench_pointer_attributes_cpp bench_pointer_attributes.cpp)
```
mdboom, line 46: Did you forget to commit `bench_pointer_attributes.cpp`?

cuda_bindings/benchmarks/runner/main.py
```diff
@@ -0,0 +1,113 @@
...
+    return f"bindings.{module_suffix}.{suffix}"
+
+
+def make_time_func(fn: Callable[[], None]) -> Callable[[int], float]:
+    """Wrap a bench_ function into a pyperf bench_time_func compatible callable."""
+
+    def time_func(loops: int) -> float:
+        t0 = time.perf_counter()
+        for _ in range(loops):
+            fn()
```
mdboom: I appreciate the decorator approach here, but this means we will be measuring the overhead of this Python function call, in addition to the actual cuda_bindings function call we are measuring.

Even though it's less convenient, I think we need to manually inline this timing benchmark into the function itself and not use this wrapper in order to get accurate timings.

cuda_bindings/benchmarks/README.md
```diff
@@ -0,0 +1,42 @@
...
+There are a couple of environments defined based on how `cuda.bindings` is installed:
+
+- `wheel`: Installs from conda packages
+- `source`: Installs from source
+
+There are a couple of tasks defined:
+
+- `bench`: Runs the Python benchmarks
+- `bench-cpp`: Runs the C++ benchmarks
+
```
mdboom, line 16: Maybe mention [pyperf system tune](https://pyperf.readthedocs.io/en/latest/system.html#system) here?

cuda_bindings/benchmarks/bench_pointer_attributes.py
```diff
@@ -0,0 +1,17 @@
...
+from runner.runtime import alloc_persistent, assert_drv
+
+
+# Allocate memory used by the tests
+PTR = alloc_persistent(1 << 18)
+ATTRIBUTE = cuda.CUpointer_attribute.CU_POINTER_ATTRIBUTE_MEMORY_TYPE
+
+
+def bench_pointer_get_attribute() -> None:
+    err, _ = cuda.cuPointerGetAttribute(ATTRIBUTE, PTR)
```
mdboom: When this is refactored to do its own timing measurement, the `PTR` and `ATTRIBUTE` vars should also be moved here (but outside of the loop) so the Python compiler will use fast local variable lookups rather than global lookups.

danielfrg:
Thanks for the comments! I dont think we **need** to merge now. I'll address the comments and once we are happy with a template we have here we can commit and then in another PR i can just add more benchmarks.

cuda_bindings/benchmarks/benchmarks/cpp/CMakeLists.txt
danielfrg: Ah yes, it was ignored on the repo gitignore, fixed now.

cuda_bindings/benchmarks/benchmarks/bench_pointer_attributes.py
danielfrg: done

cuda_bindings/benchmarks/runner/main.py
danielfrg: That makes sense, removed.

danielfrg:
Addressed the comments and i relaxed one of the deps in pixi so i think you should be able to try again.

cpcloud:
Are these going to run in CI or in any sort of regular way? I'm not sure we should have this much additional code that is going to go stale immediately.

Can we run one iteration of the benchmarks in CI so they don't go stale?

danielfrg:
Yes, the idea is for those to run in CI but I dont think we have decided the infra where thats going to happen, i remember having a discussion a while back but i dont remember what the problem is with the infra right now. 

It sounds like a good idea to at least run one iteration for now. If thats what we want I will add it here.

cuda_bindings/benchmarks/benchmarks/cpp/bench_support.hpp
```diff
@@ -0,0 +1,218 @@
...
+        std::exit(2);
+    }
+    return options;
+}
+
+inline std::string iso_now() {
+    const auto now = std::chrono::system_clock::now();
+    const std::time_t t = std::chrono::system_clock::to_time_t(now);
+    std::tm tm{};
+    gmtime_r(&t, &tm);
```
rparolin, line 89: Are we testing this implementation on Windows? In implementations in the past `gmtime_r` was not available on Windows.
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/gmtime-s-gmtime32-s-gmtime64-s?view=msvc-170

```
struct tm result;
gmtime_r(&time, &result);
```

```
struct tm result;
gmtime_s(&result, &time);
```

cuda_bindings/benchmarks/benchmarks/cpp/bench_support.hpp
```diff
@@ -0,0 +1,218 @@
...
+
+inline void print_summary(const std::string& name, const std::vector<RunResult>& results) {
+    // Collect all timed values
+    std::vector<double> all_values;
+    for (const auto& run : results) {
+        for (double v : run.values) {
+            all_values.push_back(v);
+        }
+    }
+    if (all_values.empty()) return;
```
rparolin: Nit: Does this fit style?

```if (all_values.empty()) 
       return;    // return expression on the next line to make the early return obvious.
```

cuda_bindings/benchmarks/benchmarks/cpp/bench_support.hpp
```diff
@@ -0,0 +1,218 @@
...
+
+    double sum = 0;
+    for (double v : all_values) sum += v;
```
rparolin: Nit: Ditto. for-loop body should be on the next line?

cuda_bindings/benchmarks/benchmarks/cpp/bench_support.hpp
danielfrg: No, i am not testing this on windows at the moment.

danielfrg: Added an ifdef for windows

danielfrg:
@cpcloud I just added a small smoke test running one loop in CI

danielfrg:
I'll merge in a couple of hours and then add other benchmarks in a following PR unless any objections.

danielfrg:
Whats the "Check job status" CI?

rwgk:
> Whats the "Check job status" CI?

It'll kick in after all other GitHub Actions jobs are finished, to determine if the PR can be merged.

Give me a moment to take a look.

rwgk:
I just triggered the CI. Assuming it passes, this should auto-merge.

rwgk:
@danielfrg are you around to resolve the merge conflict? — We'll have to rerun the CI.

danielfrg:
Resolved the conflict

rwgk:
@danielfrg I added 4f8eef0 and merged main again — hopefully it'll go through now.

PR #1738, opened 2026-03-07
refactor(pathfinder): unify header search architecture with library loading pattern

cpcloud:
## Summary

- Introduces a `HeaderDescriptorSpec` descriptor catalog for headers, mirroring the `DescriptorSpec` catalog pattern already used for dynamic library loading. All per-header metadata (basename, site-packages dirs, anchor layout, platform availability, system install dirs) is now authored once in a single catalog tuple, rather than spread across several manually maintained dicts.
- `supported_nvidia_headers.py` now derives its legacy table exports from the catalog (same approach as `supported_nvidia_libs.py`), and `find_nvidia_headers.py` operates on descriptors directly — eliminating the name-based `if libname == "nvvm"` / `if libname == "cccl"` branching in favor of descriptor fields (`anchor_include_rel_dirs`, `include_subdirs`, `include_subdirs_windows`).
- Public API (`LocatedHeaderDir`, `find_nvidia_header_directory`, `locate_nvidia_header_directory`, `SUPPORTED_HEADERS_CTK`) is unchanged.

**Intentional behavioral note:** The old code raised `RuntimeError` when calling `locate_nvidia_header_directory` with a header name not available on the current platform (e.g., `"cufile"` on Windows). The new code returns `None` instead, since the descriptor exists but the search simply finds nothing. This is more consistent with how callers handle "not found" and mirrors the library side's distinction between unknown vs. unavailable.

## Test plan

- [x] All existing header tests pass (`test_find_nvidia_headers.py`)
- [x] All descriptor catalog tests pass (`test_descriptor_catalog.py`)
- [x] Full test suite passes (882 passed, 1 skipped)
- [x] Verified all derived legacy tables produce identical values to the old hardcoded dicts

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_headers/find_nvidia_headers.py
+++ b/cuda_pathfinder/cuda/pathfinder/_headers/find_nvidia_headers.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
 from __future__ import annotations
@@ -6,16 +6,28 @@ from __future__ import annotations
 import functools
 import glob
 import os
+from collections.abc import Callable
 from dataclasses import dataclass
+from typing import TYPE_CHECKING
 
 from cuda.pathfinder._dynamic_libs.load_nvidia_dynamic_lib import (
     _resolve_system_loaded_abs_path_in_subprocess,
 )
 from cuda.pathfinder._dynamic_libs.search_steps import derive_ctk_root
-from cuda.pathfinder._headers import supported_nvidia_headers
+from cuda.pathfinder._headers.header_descriptor import (
+    HEADER_DESCRIPTORS,
+    platform_include_subdirs,
+    resolve_conda_anchor,
+)
 from cuda.pathfinder._utils.env_vars import get_cuda_home_or_path
 from cuda.pathfinder._utils.find_sub_dirs import find_sub_dirs_all_sitepackages
-from cuda.pathfinder._utils.platform_aware import IS_WINDOWS
+
+if TYPE_CHECKING:
+    from cuda.pathfinder._headers.header_descriptor import HeaderDescriptor
+
+# ---------------------------------------------------------------------------
+# Data types
+# ---------------------------------------------------------------------------
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


cuda_pathfinder/cuda/pathfinder/_headers/find_nvidia_headers.py
```diff
@@ -150,51 +187,17 @@ def locate_nvidia_header_directory(libname: str) -> LocatedHeaderDir | None:
...
-
-           - Probe a system-loaded ``cudart`` in a spawned child process,
-             derive the CTK root from the resolved library path, then search
-             CTK include layout under that root.
+        1. **NVIDIA Python wheels** — site-packages directories from the descriptor.
+        2. **Conda environments** — platform-specific conda include layouts.
+        3. **CUDA Toolkit environment variables** — ``CUDA_HOME`` / ``CUDA_PATH``.
+        4. **CTK root canary probe** — subprocess canary (descriptors with
+           ``use_ctk_root_canary=True`` only).
+        5. **System install directories** — glob patterns from the descriptor.
```
leofang, line 195: Q: This seems like an extra thing that did not exist before (also `find_nvidia_header_directory` does not have)?

cpcloud: Not new behavior — the old code already searched `SUPPORTED_INSTALL_DIRS_NON_CTK` glob paths (e.g. `/usr/include/nvshmem_*`) for non-CTK headers, it just wasn't documented in the docstring. The new docstring is more complete. Both `locate_nvidia_header_directory` and `find_nvidia_header_directory` now list all 5 steps.

cuda_pathfinder/cuda/pathfinder/_headers/supported_nvidia_headers.py
```diff
@@ -1,71 +1,68 @@
...
-    "npp": "npp.h",
-    "nvcc": "fatbinary_section.h",
-    "nvfatbin": "nvFatbin.h",
-    "nvjitlink": "nvJitLink.h",
-    "nvjpeg": "nvjpeg.h",
-    "nvrtc": "nvrtc.h",
-    "nvvm": "nvvm.h",
-}
+_CTK_DESCRIPTORS = tuple(desc for desc in HEADER_DESCRIPTOR_CATALOG if desc.packaged_with == "ctk")
+_NON_CTK_DESCRIPTORS = tuple(desc for desc in HEADER_DESCRIPTOR_CATALOG if desc.packaged_with == "other")
```
Andy-Jost, line 19: nit: is it worth asserting `len(_CTK_DESCRIPTORS) + len(_NON_CTK_DESCRIPTORS) == len(HEAD_DESCRIPTOR_CATALOG)`?

cpcloud: only if it's possible to construct a descriptor whose `packaged_with` value is neither `"ctk"` nor `"other"`. let me check whether that's being validated.

cpcloud: This is actually caught by mypy during pre-commit. Leaving as is.

PR #1739, opened 2026-03-07
tests: skip local NVML runtime mismatches while preserving CI failures

cpcloud:
## Summary
- Centralize NVML runtime gating through `require_nvml_runtime_or_skip_local` fixtures in `cuda_bindings` and `cuda_core` tests.
- Treat NVML init/load failures (including driver/library mismatch and missing NVML shared library) as local skips, but continue to fail in CI by re-raising.
- Cover the behavior with regression tests and apply fixture-based gating across NVML-dependent test modules/fixtures.
- Rationale: after a driver upgrade without a reboot (a common local developer state), NVML can report a temporary driver/library mismatch; local runs should skip NVML-dependent tests instead of failing collection, while CI should still fail fast for real infra regressions.

## Test plan
- [x] `pixi run --manifest-path cuda_bindings pytest cuda_bindings/tests --override-ini norecursedirs=examples -k "not test_cufile"`
- [x] `CI=1 pixi run --manifest-path cuda_bindings pytest cuda_bindings/tests/nvml/test_init.py::test_init_ref_count` (expected error on NVML mismatch in CI mode)
- [x] `pixi run --manifest-path cuda_core test` (currently blocked in this workspace by unrelated import mismatch: `cuda.core._resource_handles does not export expected C function create_culink_handle`)
- [x] `CI=1 pixi run --manifest-path cuda_core pytest cuda_core/tests/system/test_system_system.py::test_num_devices` (same unrelated import mismatch blocker)

### DIFF
--- a/cuda_core/tests/system/test_system_system.py
+++ b/cuda_core/tests/system/test_system_system.py
@@ -18,8 +18,6 @@ from cuda.core._utils.cuda_utils import handle_return
 
 from .conftest import skip_if_nvml_unsupported
 
-pytestmark = skip_if_nvml_unsupported
-
 
 def test_driver_version():
     driver_version = system.get_driver_version()
--- a/cuda_core/tests/test_memory.py
+++ b/cuda_core/tests/test_memory.py
@@ -365,7 +365,6 @@ def test_buffer_external_host():
 
 
 @pytest.mark.parametrize("change_device", [True, False])
-@pytest.mark.usefixtures("require_nvml_runtime_or_skip_local")
 def test_buffer_external_device(change_device):
     n = ccx_system.get_num_devices()
     if n < 1:
@@ -390,7 +389,6 @@ def test_buffer_external_device(change_device):
 
 
 @pytest.mark.parametrize("change_device", [True, False])
-@pytest.mark.usefixtures("require_nvml_runtime_or_skip_local")
 def test_buffer_external_pinned_alloc(change_device):
     n = ccx_system.get_num_devices()
     if n < 1:
@@ -416,7 +414,6 @@ def test_buffer_external_pinned_alloc(change_device):
 
 
 @pytest.mark.parametrize("change_device", [True, False])
-@pytest.mark.usefixtures("require_nvml_runtime_or_skip_local")
 def test_buffer_external_pinned_registered(change_device):
     n = ccx_system.get_num_devices()
     if n < 1:
@@ -450,7 +447,6 @@ def test_buffer_external_pinned_registered(change_device):
 
 
 @pytest.mark.parametrize("change_device", [True, False])
-@pytest.mark.usefixtures("require_nvml_runtime_or_skip_local")
 def test_buffer_external_managed(change_device):
     n = ccx_system.get_num_devices()
     if n < 1:


cuda_core/tests/test_memory.py
```diff
@@ -365,6 +365,7 @@ def test_buffer_external_host():
 
 
 @pytest.mark.parametrize("change_device", [True, False])
+@pytest.mark.usefixtures("require_nvml_runtime_or_skip_local")
```
mdboom: Why do the tests here now require a working NVML?  These tests predate NVML in cuda_bindings...  What's the root cause?

cuda_core/tests/system/test_system_system.py
```diff
@@ -18,6 +18,8 @@
 
 from .conftest import skip_if_nvml_unsupported
 
+pytestmark = skip_if_nvml_unsupported
```
mdboom: Most of the tests in this file are expected to run, other than test_gpu_driver_version, even without an NVML available.

cpcloud: I'll look into why this ends up being required.

cuda_core/tests/test_memory.py
cpcloud: I'll look into why this ends up being required.

cpcloud: `get_num_devices` will use NVML if it's available. So, yeah, they're pre-existing tests, but they're hitting other APIs now.

Now that these route through NVML when it's available, they're a place where we need to skip if nvml is available but fails in an expected way.

> What's the root cause?

The root cause is that I upgraded the driver without rebooting. Since NVML is a driver library, I can no longer use it without a reboot.

I don't want to reboot to keep working in the repo, especially if I'm working on something unrelated to any of this code.

cuda_core/tests/system/test_system_system.py
cpcloud: Not required

cuda_core/tests/test_memory.py
mdboom: Thanks for the clarification.  That makes sense.

However, I'm not sure we want to pave over broken installations like this (though I feel your pain about wanting to continue working with a partially-broken install).  The test suite can't know whether things are broken because of local installation issues or changes to the code that have broken things, and this would pave over the latter.  I know this would eventually get caught in CI, but I'm not a fan of having local and remote testing behave differently.

cpcloud:
Not worth the review effort/delay. By the time anyone else chimes in, everyone will have forgotten why this was done and it'll remain in limbo.

cuda_core/tests/test_memory.py
leofang: @cpcloud there is a way to update driver without rebooting. Just need to unload and load the kernel modules with `rmmod` and/or `modprobe`, see, e.g. 
https://forums.developer.nvidia.com/t/reset-driver-without-rebooting-on-linux/40625/2
Some of our internal infra use this, for example, to achieve <1 min driver refreshing at node allocation time.

leofang:
Agreed with Mike. The system needs to be in a sane state. Running `nvidia-smi` prior to any CUDA program, as in our CI, would cache such issues.

PR #1751, opened 2026-03-11
refactor(workflows): remove release build-ctk-ver input

cpcloud:
## Summary
- derive the docs CTK build version from `ci/versions.yml` in the checked-out ref instead of requiring a manual `build-ctk-ver` input
- remove redundant `build-ctk-ver` plumbing from the shared docs callers in `release.yml`, `release-cuda-pathfinder.yml`, and `ci.yml`
- update the core release instructions to match the simplified release workflow inputs


Andy-Jost (approved):
No issues

PR #1752, opened 2026-03-11
ci: consolidate release publishing flow

cpcloud:
## Summary
- remove the manual wheel-index selector from `CI: Release`
- publish the selected wheels to TestPyPI and then PyPI in a single workflow run
- update the release docs to match the streamlined release flow and clean up shell quoting in `release.yml`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -52,15 +52,10 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
-          if [[ -n "${{ inputs.run-id }}" ]]; then
-            echo "Using provided run ID: ${{ inputs.run-id }}"
-            echo "run-id=${{ inputs.run-id }}" >> "$GITHUB_OUTPUT"
-          else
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -169,14 +158,37 @@ jobs:
...
+      - determine-run-id
+      - publish-testpypi
+    environment:
+      name: pypi
+      url: https://pypi.org/${{ inputs.component != 'all' && format('p/{0}/', inputs.component) || '' }}
+    permissions:
+      id-token: write
+    steps:
+      - name: Checkout Source
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
```
rparolin, line 179: Consider: Hoisting all the explicit SHA to the top and give them a variable name.  So all tunables can be changed in a single place.

cpcloud: That'd be nice, except there's still repetition across files.

Also, I don't think it matters because dependabot is updating these automatically for us. It handles repetition with aplomb.

PR #1757, opened 2026-03-12
[no-ci] Claim context7 repo

danielfrg:
This is just to claim that we own the cuda-python repo on context7.

Context7 is a popular MCP server for getting code doc into agents and we should keep it updated, right now the last automatic built they did was 6 months ago. With this we can control it more.


danielfrg:
I look at their docs but unfortunately i dont think we can add a comment and the file needs to be in the root of the repo.

cpcloud:
Let's make a decision here. JSON famously doesn't support comments. We have `git log context7.json`, let's just make sure the commit message is thorough enough for our future selves.

rwgk:
I'm fine with relying on `git log`, but there is actually a [documented `"description"` field](https://context7.com/schema/context7.json) we could use, e.g.:

```json
{
    "$schema: "https://context7.com/schema/context7.json",
    "description": "NVIDIA cuda-python ownership claim for Context7 MCP documentation",
    "url": "https://context7.com/nvidia/cuda-python",
    "public_key": "pk_gupaHhsdvsuT1j3BZpb7i"
}
```

danielfrg:
I would just rely on the commit message, that description field its for describing what the project does for coding agents.

rwgk:
> I would just rely on the commit message, that description field its for describing what the project does for coding agents.

That seems useful; what I suggested in my previous comment would seem to fit that expectation, and it's a trivial effort to make the change?

danielfrg:
Yeah I am not opposed to adding a description, just that I would not add a line mentioning ownership, Context 7 or MCP. Instead it would be like: CUDA Python documentation, libraries include cuda.bindings, cuda.core. or something like that that describes what we provide there.

rwgk:
@danielfrg I added a description. Good?

@mdboom Could you please look again, to unblock merging this PR?

danielfrg:
Do we need to run CI on this one too even with the `[no-ci]` tag in the title?

rwgk:
> Do we need to run CI on this one too even with the `[no-ci]` tag in the title?

Yes, but it will be very quick: all builds and tests are skipped with the `[no-ci]` tag.

The bot doesn't seem to trigger, probably because it thinks the CI ran already (I cancelled it manually yesterday, to not waste resources). I'll update the branch again, that should do the trick hopefully.

PR #1768, opened 2026-03-16
fix(pathfinder): keep canary probes script-safe

### LINKED ISSUE #1767: [BUG]: `run_in_spawned_child_process` fails if used as script
### Is this a duplicate?

### Type of Bug

Runtime Error

### Component

cuda.pathfinder

### Describe the bug

Forwwarding from https://github.com/cupy/cupy/issues/9803. If everything else fails for e.g. header discovery, `cuda-pathfinder` falls back to using:
* `_resolve_system_loaded_abs_path_in_subprocess`
* and: `run_in_spawned_child_process`

However, if you use this inside a script (i.e. missing `if __name__ == "__main__"`), the process `spawn` is problematic, as it leads to the script itself being executed during bootstrapping.
I.e. there are (apparently) two possible outcomes here:
* Python raises an error pointing out that `if __name == "__main__"` seems missing (I got this if I just call `_resolve_system_loaded_abs_path_in_subprocess()` directly.
* In the OP the script spawning process just triggered a second run of the script with the first run hitting the 10 second time-out if the script runs more than 10 seconds.
  (I am not sure why the nested call succeeds spawning, but I doubt it matters.)

Not sure if this is a priority, because I am not sure how likely it is for cuda-pathfinder to reach this path.  The work-around of setting `CUDA_PATH=` is also straight-forward. But it limits the reliability of this fallback.

### How to Reproduce

I am not sure what likely needs to be weird about the environment for `cuda-pathfinder` to go to such lengths to find the right paths.
However, calling:
```
cuda.pathfinder._headers.find_nvidia_headers._find_ctk_header_directory_via_canary("nvrtc", "nvrtc.h")
```
in a Python script reproduces the first error for me (an older cuda-pathfinder version, `main` requires a different signature there, I think.).

### Expected behavior

Maybe rather than spawning, this needs to use a subprocess to avoid issues with scripts?
I.e. execute something like:
```
subprocess.check_output([sys.executable, "-c", f"from cuda.pathfinder import ...; probe_canary_abs_path_and_print_json('{libname}')"])
```
or maybe nicer and safer, via:
```
[sys.executable, "-m", cuda.pathfinder._something, libname]
```
(I am not sure if there are subtleties around `subprocess`, but I guess `spawn` must be similar already?).

### Operating System

_No response_

### nvidia-smi output

_No response_

cpcloud:
Closes #1767.

## Summary
- Run the CTK canary probe via `python -m cuda.pathfinder._dynamic_libs.canary_probe_subprocess` instead of `multiprocessing` `spawn`, and add a regression test showing a plain caller script is executed only once.
- Preserve the original isolation requirement behind PR #1595: the canary still runs in a fresh Python interpreter with independent loader state, but it no longer bootstraps through the caller's `__main__`, so script entrypoints without an `if __name__ == \"__main__\"` guard do not recurse or fail during multiprocessing startup.
- Anchor the canary subprocess to the parent package's import root so the child resolves the same `cuda.pathfinder` distribution as the parent. This avoids wheel-based CI importing the checkout tree copy, which can be missing generated files such as `cuda.pathfinder._version`, and keeps source-vs-wheel behavior consistent.
- Update the header-discovery/release-note wording so the canary path is described as a dedicated subprocess while leaving the shared `spawned_process_runner` in place for the callable-based isolation paths and real-loading tests that still need it.

## Reviewer context
- PR #1595 originally introduced the canary probe as a dedicated `python -m ...` subprocess.
- During review it was switched to `multiprocessing.get_context(\"spawn\")` to avoid a forked child inheriting preloaded CUDA libraries from the parent and to keep the probe isolated/predictable.
- This change keeps that same guarantee because the canary still runs in a dedicated module entrypoint in a fresh interpreter; the difference is only that it no longer re-imports the user's script as `__main__`, which is the failure mode reported in #1767.
- While validating the change in wheel CI, the child `python -m ...` process was found to be resolving `cuda.pathfinder` from the checkout tree because `run-tests pathfinder` runs from `./cuda_pathfinder`. In that environment the checkout copy can lack generated files such as `cuda.pathfinder._version`, so the child fails before probing.
- Setting the child `cwd` to the already-imported parent package's import root makes the subprocess import the same distribution root as the parent, whether the parent came from source or from an installed wheel.

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_headers/find_nvidia_headers.py
+++ b/cuda_pathfinder/cuda/pathfinder/_headers/find_nvidia_headers.py
@@ -115,7 +115,7 @@ def find_via_ctk_root_canary(desc: HeaderDescriptor) -> LocatedHeaderDir | None:
     """Try CTK header lookup via CTK-root canary probing.
 
     Skips immediately if the descriptor does not opt in (``use_ctk_root_canary``).
-    Otherwise, system-loads ``cudart`` in a dedicated Python subprocess, derives
+    Otherwise, system-loads ``cudart`` in a fully isolated Python subprocess, derives
     CTK root from the resolved library path, and searches the expected include
     layout under that root.
     """
--- a/cuda_pathfinder/docs/source/release/1.4.0-notes.rst
+++ b/cuda_pathfinder/docs/source/release/1.4.0-notes.rst
@@ -12,8 +12,8 @@ Highlights
 ----------
 
 * Add CTK root canary probing for non-standard-path libraries in
-  ``load_nvidia_dynamic_lib()`` (notably ``nvvm``), including dedicated
-  subprocess isolation for the canary probe.
+  ``load_nvidia_dynamic_lib()`` (notably ``nvvm``), including spawned child
+  process isolation for the canary probe.
   (`PR #1595 <https://github.com/NVIDIA/cuda-python/pull/1595>`_)
 
 * Restore backward-compatible exception behavior for
--- a/cuda_pathfinder/docs/source/release/1.4.3-notes.rst
+++ b/cuda_pathfinder/docs/source/release/1.4.3-notes.rst
@@ -1,0 +1,22 @@
+.. SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+.. SPDX-License-Identifier: Apache-2.0
+
+.. py:currentmodule:: cuda.pathfinder
+
+``cuda-pathfinder`` 1.4.3 Release notes
+=======================================
+
+Highlights
+----------
+
+* Fix CTK canary probing when it is reached from plain Python scripts that do
+  not use an ``if __name__ == "__main__"`` guard. The canary now runs in a
+  fully isolated Python subprocess without re-entering the caller's script
+  during child-process startup.
+  (`PR #1768 <https://github.com/NVIDIA/cuda-python/pull/1768>`_)
+
+* Make the canary subprocess resolve ``cuda.pathfinder`` from the same import
+  root as the parent process. This avoids mixed source-tree versus wheel imports
+  in wheel-based test environments and keeps the probe behavior consistent
+  across source and installed-package workflows.
+  (`PR #1768 <https://github.com/NVIDIA/cuda-python/pull/1768>`_)


cuda_pathfinder/cuda/pathfinder/_headers/find_nvidia_headers.py
```diff
@@ -115,7 +115,7 @@ def find_via_ctk_root_canary(desc: HeaderDescriptor) -> LocatedHeaderDir | None:
     """Try CTK header lookup via CTK-root canary probing.
 
     Skips immediately if the descriptor does not opt in (``use_ctk_root_canary``).
-    Otherwise, system-loads ``cudart`` in a spawned child process, derives
+    Otherwise, system-loads ``cudart`` in a dedicated Python subprocess, derives
```
rwgk:
```suggestion
    Otherwise, system-loads ``cudart`` in a fully isolated Python subprocess, derives
```

cuda_pathfinder/docs/source/release/1.4.0-notes.rst
```diff
@@ -12,8 +12,8 @@ Highlights
 ----------
 
 * Add CTK root canary probing for non-standard-path libraries in
-  ``load_nvidia_dynamic_lib()`` (notably ``nvvm``), including spawned child
-  process isolation for the canary probe.
+  ``load_nvidia_dynamic_lib()`` (notably ``nvvm``), including dedicated
```
rwgk: Was this changed by accident? I think it's better to not touch old release notes, except maybe for adding clearly marked edits pointing to related fixes.

Maybe add 1.4.3 release notes in this PR? Then we could cut a new release straightaway (I think it's worth it).

cpcloud: yep

cuda_pathfinder/cuda/pathfinder/_headers/find_nvidia_headers.py
cpcloud: Updated the wording to "fully isolated Python subprocess" in `find_nvidia_headers.py`.

cuda_pathfinder/docs/source/release/1.4.0-notes.rst
cpcloud: Reverted the `1.4.0` note text and added new `1.4.3` release notes for this fix set in `cuda_pathfinder/docs/source/release/1.4.3-notes.rst`.

seberg:
That was quick, thanks!

PR #1762, opened 2026-03-12
Infrastructure changes preparing for explicit graph construction

Andy-Jost:
## Summary

Groundwork for explicit CUDA graph construction (#1317).

## Changes

- **HandleRegistry template**: Maps raw CUDA handles (CUevent, CUkernel) back to
  their owning `shared_ptr` via `weak_ptr`, enabling reconstruction of Python
  objects from driver-returned handles.
- **EventBox metadata**: Event properties (timing_disabled, busy_waited,
  ipc_enabled, device_id, context) stored in C++ alongside the CUevent handle,
  accessed via `get_box()` pointer arithmetic. Replaces cached Python-level fields.
- **Event/kernel reverse-lookup registries**: HandleRegistry instantiations for
  events and kernels, with automatic registration/cleanup.
- **Event.from_handle()**: Public API for creating non-owning Event objects from
  foreign CUevent handles.
- **Kernel reverse-lookup**: `Kernel.from_handle` now uses the kernel registry
  with library-mismatch warning.
- **IPC cache refactor**: Migrated to use HandleRegistry.
- **Package conversion**: `_graph.py` → `_graph/__init__.py` (rename only).

## Test Coverage

- New tests in `test_module.py` for `Kernel.from_handle` library-mismatch
  warning and foreign kernel handle wrapping.
- All existing tests pass unchanged.

## Related Work

- Phase 2 (GraphDef API + tests) will follow in a separate PR.
- Parent issue: #1317

PR #1772, opened 2026-03-16
Add explicit CUDA graph construction API (GraphDef, GraphNode)

### LINKED ISSUE #1317: CUDA graph phase N - explicit graph construction
Instead of relying on stream capturing, which is considered an implementation detail (that in the future we could allow users to opt in or out), our graph builder APIs were designed to allow an alternative implementation based on explicitly constructing a graph by adding nodes. 

The use cases are unclear atm, so just keeping this in the backlog for now.

Andy-Jost:
## Summary

- Introduces `GraphDef` and `GraphNode` types for explicit CUDA graph construction, complementing the existing stream-capture path via `GraphBuilder`.
- Adds a full node hierarchy (kernel, memset, memcpy, alloc, free, child graph, event record/wait, host callback, and conditional nodes) with a fluent builder API.
- Extracts graph instantiation logic into a shared helper so both `GraphBuilder.complete()` and `GraphDef.instantiate()` support `GraphCompleteOptions`, and removes the pre-CUDA 12.0 `cuGraphInstantiateWithFlags` fallback.

Depends on #1762 (merged).

Closes #1317.

## Changes

- **New files**: `_graphdef.pxd` / `_graphdef.pyx` containing `GraphDef`, `GraphNode`, and all node subclasses
- **`_graph/__init__.py`**: Extracted `_instantiate_graph()` helper; updated exports for new types
- **`resource_handles.cpp/.hpp`**: Added graph and graph-node RAII handles, `HandleRegistry` for reverse handle lookup
- **`_resource_handles.pxd/.pyx`**: Cython declarations for graph handle types
- **`_memory/_buffer.pyx`**: Graph memory allocation support
- **`_utils/cuda_utils.pxd/.pyx`**: Utility additions for graph construction
- **Tests**: Comprehensive test suite across 4 new test files (unit, error, integration, lifetime) plus object protocol tests; instantiation and execution tests parametrized over `GraphCompleteOptions` variants

## Test plan

- [ ] Existing graph capture tests pass (no regressions from instantiation refactor)
- [ ] New explicit graph tests pass: `test_explicit.py`, `test_explicit_errors.py`, `test_explicit_integration.py`, `test_explicit_lifetime.py`
- [ ] Object protocol tests pass: `test_object_protocols.py`

### DIFF
--- a/cuda_core/tests/graph/test_explicit.py
+++ b/cuda_core/tests/graph/test_explicit.py
@@ -41,6 +41,12 @@ def _skip_if_no_mempool():
         pytest.skip("Device does not support mempool operations")
 
 
+def _skip_if_no_managed_mempool():
+    _skip_if_no_mempool()
+    if not Device(0).properties.concurrent_managed_access:
+        pytest.skip("Device does not support managed memory pool operations")
+
+
 def _driver_has_node_get_params():
     from cuda.bindings import driver as drv
 
@@ -266,6 +272,7 @@ def _build_alloc_node(g):
 
 
 def _build_alloc_managed_node(g):
+    _skip_if_no_managed_mempool()
     device_id = Device().device_id
     options = GraphAllocOptions(memory_type="managed")
     entry = g.alloc(ALLOC_SIZE)


cuda_core/cuda/core/_graph/_graphdef.pyx
```diff
@@ -0,0 +1,2111 @@
...
+    and specifying dependencies. Once construction is complete, call
+    instantiate() to obtain an executable Graph.
+    """
+
+    def __init__(self):
+        """Create a new empty graph definition."""
+        cdef cydriver.CUgraph graph = NULL
+        with nogil:
+            HANDLE_RETURN(cydriver.cuGraphCreate(&graph, 0))
+        self._h_graph = create_graph_handle(graph)
```
cpcloud, line 291: `GraphDef()` is the new top-level entry point for explicit construction, but right now it calls `cuGraphCreate` before any shared init path has run. In a clean `pixi run --manifest-path cuda_core python -c "from cuda.core._graph._graphdef import GraphDef; GraphDef()"` session this immediately raises `CUDA_ERROR_NOT_INITIALIZED`. Other high-level `cuda.core` entry points hide that detail, so this one should too.

```suggestion
    def __init__(self):
        """Create a new empty graph definition."""
        Device()
        cdef cydriver.CUgraph graph = NULL
        with nogil:
            HANDLE_RETURN(cydriver.cuGraphCreate(&graph, 0))
        self._h_graph = create_graph_handle(graph)
```

cuda_core/cuda/core/_graph/_graphdef.pyx
```diff
@@ -0,0 +1,2111 @@
...
+        return f"<Node handle=0x{<uintptr_t>node:x}>"
+
+    def __eq__(self, other) -> bool:
+        if not isinstance(other, Node):
+            return NotImplemented
+        cdef Node o = <Node>other
+        return as_intptr(self._h_node) == as_intptr(o._h_node)
+
+    def __hash__(self) -> int:
+        return hash(as_intptr(self._h_node))
```
cpcloud: `Node.__eq__` / `__hash__` only look at the raw `CUgraphNode` handle. That works for real nodes, but the synthetic entry node is always `NULL`, so `GraphDef()._entry == other_graph._entry` and both hash to `0`. The new object-protocol coverage in `tests/test_object_protocols.py` already exercises `sample_root_node` vs `sample_root_node_alt`, and this implementation will collapse distinct graphs in sets/dicts.

I think node identity needs to include the owning graph as well, not just the node handle.

```suggestion
    def __eq__(self, other) -> bool:
        if not isinstance(other, Node):
            return NotImplemented
        cdef Node o = <Node>other
        return (
            as_intptr(graph_node_get_graph(self._h_node)),
            as_intptr(self._h_node),
        ) == (
            as_intptr(graph_node_get_graph(o._h_node)),
            as_intptr(o._h_node),
        )

    def __hash__(self) -> int:
        return hash((
            as_intptr(graph_node_get_graph(self._h_node)),
            as_intptr(self._h_node),
        ))
```

cuda_core/cuda/core/_graph/_graphdef.pyx
```diff
@@ -0,0 +1,2111 @@
...
+                    c_user_data = malloc(len(buf))
+                    if c_user_data == NULL:
+                        raise MemoryError(
+                            "failed to allocate user_data buffer")
+                    c_memcpy(c_user_data, <const char*>buf, len(buf))
+                    _attach_user_object(
+                        as_cu(h_graph), c_user_data,
+                        <cydriver.CUhostFn>free)
+
+            node_params.userData = c_user_data
```
cpcloud, line 1304: This branch keeps Python callables alive, and it keeps bytes-backed `user_data` alive, but it does **not** keep the `ctypes.CFUNCTYPE` wrapper alive. If the caller drops `raw_fn` before launch/clone, ctypes can GC the callback trampoline while the graph still holds the raw function pointer, which turns into a use-after-free crash when CUDA calls back into it.

I'd retain the ctypes wrapper through the same graph user-object mechanism used for Python callables, and add a regression test that `del raw_fn; gc.collect(); graph.launch(...)` still works.

```suggestion
        if isinstance(fn, ct._CFuncPtr):
            Py_INCREF(fn)
            fn_pyobj = <void*>fn
            _attach_user_object(
                as_cu(h_graph), fn_pyobj,
                <cydriver.CUhostFn>_py_host_destructor)

            node_params.fn = <cydriver.CUhostFn><uintptr_t>ct.cast(
                fn, ct.c_void_p).value

            if user_data is not None:
                if isinstance(user_data, int):
                    c_user_data = <void*><uintptr_t>user_data
                else:
                    buf = bytes(user_data)
                    c_user_data = malloc(len(buf))
                    if c_user_data == NULL:
                        raise MemoryError(
                            "failed to allocate user_data buffer")
                    c_memcpy(c_user_data, <const char*>buf, len(buf))
                    _attach_user_object(
                        as_cu(h_graph), c_user_data,
                        <cydriver.CUhostFn>free)

            node_params.userData = c_user_data
```

cuda_core/cuda/core/_graph/_graphdef.pyx
Andy-Jost: Isn't this the intended design, that users are responsible for calling `Device.set_current`?

To pick an example, I see this for stream creation:

```
>>> import cuda.core as cc
>>> cc.Device(0).create_stream()
  File "<python-input-1>", line 1, in <module>
    cc.Device(0).create_stream()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "cuda/core/_device.pyx", line 1300, in cuda.core._device.Device.create_stream
  File "cuda/core/_device.pyx", line 975, in cuda.core._device.Device._check_context_initialized
cuda.core._utils.cuda_utils.CUDAError: Device 0 is not yet initialized, perhaps you forgot to call .set_current() first?
```

cpcloud: Fair enough. If it is, then I rescind my comment!

cuda_core/cuda/core/_graph/_graphdef.pyx
Andy-Jost: Good catch. Fixed it and added tests for callback lifetimes (all missing before).

Andy-Jost:
@cpcloud The latest upload fixes the issues you pointed out, except the CUDA init problem, which needs clarification.

cpcloud:
@Andy-Jost Since your PR is not a draft you shouldn't need ok-to-test anymore.

PR #1773, opened 2026-03-16
fix(core): unit-test peer access planning without multi-GPU hardware

cpcloud:
## Summary
- extract `DeviceMemoryResource.peer_accessible_by` transition planning into a helper so the normalization and delta logic can be validated without multiple GPUs
- add single-GPU tests for peer-access normalization, inaccessible-peer validation, and transition planning while keeping the existing x2/x3 integration tests
- preserve the hardware-backed peer-access and IPC integration checks so multi-GPU machines still exercise the end-to-end behavior

## Test plan
- [x] `pixi run -e default pytest tests/test_memory_peer_access_utils.py tests/test_memory_peer_access.py -q`
- [x] `pixi run -e default pytest tests/memory_ipc/test_peer_access.py -q`

### DIFF
--- a/cuda_core/cuda/core/_memory/_device_memory_resource.pyx
+++ b/cuda_core/cuda/core/_memory/_device_memory_resource.pyx
@@ -25,6 +25,7 @@ import multiprocessing
 import platform  # no-cython-lint
 import uuid
 
+from ._peer_access_utils import plan_peer_access_update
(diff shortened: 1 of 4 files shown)


leofang:
Q: What does this PR fix? I don't follow. Can we please always file an issue first, before sending a PR?

PR #1777, opened 2026-03-17
refactor(pathfinder): replace spawned child runner with subprocess entrypoint

cpcloud:
## Summary
- replace the remaining `multiprocessing`-based spawned child runner in `cuda_pathfinder` with a dedicated `python -m` subprocess entrypoint
- move the real-loading tests onto that subprocess entrypoint so the test isolation model matches the canary probe path
- add focused subprocess helper coverage and remove the old runner module plus its standalone test

## Test plan
- [x] `cd cuda_pathfinder && pixi run pytest tests/`
- [x] `pixi run ruff check cuda_pathfinder/cuda cuda_pathfinder/tests`

### DIFF
--- a/cuda_pathfinder/tests/child_load_nvidia_dynamic_lib_helper.py
+++ b/cuda_pathfinder/tests/child_load_nvidia_dynamic_lib_helper.py
@@ -5,12 +5,16 @@ from __future__ import annotations
 
 import subprocess
 import sys
+import tempfile
 from pathlib import Path
 
 from cuda.pathfinder._testing.load_nvidia_dynamic_lib_subprocess import DYNAMIC_LIB_NOT_FOUND_MARKER
 
 LOAD_NVIDIA_DYNAMIC_LIB_SUBPROCESS_MODULE = "cuda.pathfinder._testing.load_nvidia_dynamic_lib_subprocess"
-LOAD_NVIDIA_DYNAMIC_LIB_SUBPROCESS_CWD = Path(__file__).resolve().parents[1]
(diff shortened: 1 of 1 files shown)


cuda_pathfinder/cuda/pathfinder/_testing/load_nvidia_dynamic_lib_subprocess.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env python
```
rwgk, line 1: Currently wondering: do we need this new subdirectory?

I'll play with this for a few minutes, hoping that we don't have to move the test-only code here.

rwgk: Gist of my exploration with Cursor:

We'll merge this PR now and plan a follow-on PR to unify the production and testing subprocess code paths.

## Why unify later

- Reduce duplication between the canary probe subprocess and the test-only
  load subprocess entrypoint.
- Share a single JSON response protocol for success, not-found, and errors.
- Centralize subprocess invocation, timeout handling, and stderr formatting.
- Keep production and tests aligned as future behavior evolves.
- Make it easier to reason about, test, and maintain subprocess behavior.

rwgk: Here is the result of continuing the exploration with Cursor: #1779

PR #1779, opened 2026-03-17
refactor(pathfinder): unify dynamic-lib subprocess probing (production, test code)

### LINKED ISSUE #1771: refactor(pathfinder): evaluate replacing spawned_process_runner with dedicated subprocess entrypoints
## Context

In PR #1768 we switched the CTK canary probe away from `multiprocessing.get_context(\"spawn\")` and back to a dedicated `python -m ...` subprocess entrypoint, while keeping `cuda.pathfinder._utils.spawned_process_runner` in place for the callable-based isolation paths used by tests.

During review, a follow-on idea came up: could `cuda_pathfinder` use the same fully isolated subprocess pattern for those tests as well, and potentially eliminate `spawned_process_runner.py` entirely?

## Motivation

Right now `cuda_pathfinder` has two different subprocess-isolation mechanisms:

- dedicated module subprocesses for the canary probe
- `multiprocessing` `spawn` via `spawned_process_runner.py` for test helpers / callable execution

This works today, but it means we have two different models to reason about when debugging subprocess behavior, import resolution, loader-state isolation, and script-entrypoint behavior.

## Question to investigate

Would it be better to replace the remaining `spawned_process_runner.py` usage in `cuda_pathfinder` with the same dedicated-subprocess approach used by the canary probe?

Things to evaluate:

- whether the test helpers can be expressed cleanly as module/file entrypoints instead of spawned Python callables
- whether the fully isolated subprocess approach preserves the same loader-state guarantees that motivated `spawn` in the first place
- whether this would simplify maintenance by making production and test isolation use one mechanism
- whether there are places where callable-based `spawn` is still meaningfully better and should be kept

## Scope

This is a follow-on refactor / design issue, not a blocker for #1768.

rwgk:
Closes #1771

Follow-on to PR #1777 

- Consolidate canary and test subprocess entrypoints behind a single module with a shared JSON payload protocol.
- Update load/test helpers to use the unified subprocess command and payload parsing.
- Remove the old canary/test-only subprocess modules and adjust lint config.

Special attention was paid to:
- Canary remains system-search only (no change to discovery semantics).
- JSON payload parsing stays strict and errors remain explicit.
- Subprocess cwd stays neutral to avoid source-tree import assumptions.

___

Suggested review path that minimizes context switching:

Production flow first
- cuda_pathfinder/cuda/pathfinder/_dynamic_libs/subprocess_protocol.py
  Focus: payload schema (status, abs_path), parsing errors, command builder.
- cuda_pathfinder/cuda/pathfinder/_dynamic_libs/dynamic_lib_subprocess.py
  Focus: canary vs load mode behavior, JSON output, abs_path validation.
- cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py
  Focus: canary subprocess invocation, timeout/error handling, JSON parsing.

Test and helper flow
- cuda_pathfinder/tests/child_load_nvidia_dynamic_lib_helper.py
  Focus: command invocation, neutral cwd, JSON parsing.
- cuda_pathfinder/tests/test_load_nvidia_dynamic_lib_subprocess.py
  Focus: subprocess command args and payload expectations.
- cuda_pathfinder/tests/test_ctk_root_discovery.py
  Focus: updated canary probe payload/command expectations.
- cuda_pathfinder/tests/test_load_nvidia_dynamic_lib.py
  Focus: not-found handling and parsed abs_path.
- cuda_pathfinder/tests/test_driver_lib_loading.py
  Focus: not-found handling and parsed abs_path.
- ruff.toml
  Focus: lint allowlist path updated.

### DIFF
--- a/cuda_core/cuda/core/_cpp/resource_handles.cpp
+++ b/cuda_core/cuda/core/_cpp/resource_handles.cpp
@@ -56,6 +56,9 @@ decltype(&cuLibraryLoadData) p_cuLibraryLoadData = nullptr;
 decltype(&cuLibraryUnload) p_cuLibraryUnload = nullptr;
 decltype(&cuLibraryGetKernel) p_cuLibraryGetKernel = nullptr;
 
+// Graph
+decltype(&cuGraphDestroy) p_cuGraphDestroy = nullptr;
+
 // Linker
 decltype(&cuLinkDestroy) p_cuLinkDestroy = nullptr;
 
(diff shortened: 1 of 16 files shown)


cuda_pathfinder/cuda/pathfinder/_dynamic_libs/dynamic_lib_subprocess.py
```diff
@@ -0,0 +1,116 @@
...
+    return abs_path
+
+
+def _validate_abs_path(abs_path: str) -> None:
+    assert abs_path, f"empty path: {abs_path=!r}"
+    assert os.path.isabs(abs_path), f"not absolute: {abs_path=!r}"
+    assert os.path.isfile(abs_path), f"not a file: {abs_path=!r}"
+
+
+def _load_nvidia_dynamic_lib_for_test(libname: str) -> str:
```
cpcloud, line 51: This feels super weird to me to have in the actual library code.

cpcloud:
Non-blocking nit: normalizing load-mode `DynamicLibNotFoundError` to a bare `{"status": "not-found", "abs_path": null}` payload keeps the protocol clean, but it also drops the child-process detail we used to get in not-found cases, which may make CI/debugging a little harder when a subprocess miss is caused by something subtler than the environment simply lacking the library.

Also not a blocker, but I could see a small follow-up either preserving a bit more diagnostic detail for load-mode not-found results, or keeping the test-only load verifier behind a `_testing` wrapper while leaving the shared protocol/CLI in `_dynamic_libs`.

PR #1782, opened 2026-03-17
docs(bindings): add pixi docs environment

cpcloud:
## Summary
- add a `docs` feature and `docs` environment to `cuda_bindings/pixi.toml`, mirroring the bindings docs dependencies used in CI
- add a Linux `build-docs` task so the bindings docs can be built locally through `pixi`
- update `cuda_bindings/pixi.lock` with the packages required by the new docs environment

## Test plan
- [x] `pixi run --manifest-path cuda_bindings -e docs python -c "import sphinx, myst_nb, sphinx_copybutton, nvidia_sphinx_theme"`
- [x] `pixi run --manifest-path cuda_bindings -e docs build-docs`

### DIFF
--- a/cuda_bindings/pixi.toml
+++ b/cuda_bindings/pixi.toml
@@ -23,6 +23,30 @@ pytest-repeat = "*"
 pyglet = ">=2.1.9"
 numpy = "*"
 
+[feature.docs.dependencies]
+# Keep this list aligned with cuda_python/docs/environment-docs.yml.
+python = "3.12.*"
+cython = ">=3.2,<3.3"
+myst-parser = "*"
+numpy = "*"
+numpydoc = "*"
+pip = "*"
+pydata-sphinx-theme = "*"
+pytest = "*"
+scipy = "*"
+sphinx = "<8.2.0"
+sphinx-copybutton = "*"
+myst-nb = "*"
+enum_tools = "*"
+sphinx-toolbox = "*"
+pyclibrary = ">=0.1.7"
+
+[feature.docs.pypi-dependencies]
+nvidia-sphinx-theme = "*"
+
+[feature.docs.target.linux.dependencies]
+make = "*"
+
 [feature.cython-tests.dependencies]
 cython = ">=3.2,<3.3" # for tests that exercise APIs from cython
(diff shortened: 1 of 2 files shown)


cuda_bindings/pixi.toml
```diff
@@ -23,6 +23,30 @@ pytest-repeat = "*"
 pyglet = ">=2.1.9"
 numpy = "*"
 
+[feature.docs.dependencies]
+# Keep this list aligned with cuda_python/docs/environment-docs.yml.
```
mdboom, line 27: There is another list in `toolshed/setup-docs-env.sh`.  I think we should try to reduce these.  Probably `toolshed/setup-docs-env.sh` should go.  (And the reference to it removed from `cuda_python/docs/environment-docs.yml` -- replace it with a reference to this `pixi.toml` file).

cuda_bindings/pixi.toml
```diff
@@ -23,6 +23,30 @@ pytest-repeat = "*"
...
+python = "3.12.*"
+cython = ">=3.2,<3.3"
+myst-parser = "*"
+numpy = "*"
+numpydoc = "*"
+pip = "*"
+pydata-sphinx-theme = "*"
```
mdboom, line 34: Maybe this is a follow-on to this PR to reduce unnecessary direct dependencies here. 

1) It seems weird to install two themes.  The one we use directly is `nvidia-sphinx-theme`.  Sometimes one theme depends on another, but we should let dependency resolution handle that, if so.

2) Why an explicit dependency on `pip`?

cpcloud: 1. Agreed.
2. Not sure, the agent put that in there. I'll poke around and see if it's necessary.

cuda_bindings/pixi.toml
rwgk: > Probably toolshed/setup-docs-env.sh should go.

SGTM. It was useful to me at the time, when there was nothing else.

cpcloud:
Superseded by #1870, which added the bindings pixi docs environment (along with the other packages) and has been merged. Closing.

Issue #1681, opened 2026-02-20
Examples: Circle back and add examples for all recently added features / APIs


rparolin:

cpcloud:
Seems very AI-taskable.

PR #1783, opened 2026-03-17
[Example] Add a simple NVML example

mdboom:
This is one example as part of #1681


cuda_bindings/examples/4_CUDA_Libraries/nvidia_smi.py
```diff
@@ -0,0 +1,241 @@
...
+
+# Utilities to help format the output table. See below for NVML usage.
+
+
+def format_size(bytes_val: int) -> str:
+    """Formats bytes to MiB."""
+    return f"{bytes_val / (1024 * 1024):.0f}MiB"
+
+
+LINES = [[[4, 27, 6], [18, 3], [20]], [[4, 6, 13, 13], [22], [9, 10]]]
```
cpcloud, line 29: Why not just dump out a TSV, PSV, or CSV and pipe to `column -t` instead of devoting 40+ lines to a table formatter that's irrelevant for NVML APIs?

Not gonna block, but I do feel like this hides the actual useful bits.

cpcloud (approved):
Ship it!

cuda_bindings/examples/4_CUDA_Libraries/nvidia_smi.py
mdboom: Yeah, a little bit.  It started as an exercise to replicate `nvidia-smi` (which I thought many CUDA programmers might already be familiar with) in Python.  But I agree it's a bit ridiculous.

PR #1795, opened 2026-03-20
[doc-only] pathfinder: clarify Windows DLL search semantics

### LINKED ISSUE #1781: [BUG][pathfinder] Windows system search fails to find any CUDA DLL when CUDA_HOME/CUDA_PATH are not set

rwgk:
Fixes #1781

Document that Python 3.8+ excludes PATH from the native Windows DLL search used by cuda.pathfinder, so CTK installs are typically found via CUDA_HOME/CUDA_PATH or other explicit locations instead.

Made-with: Cursor

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
+++ b/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
@@ -156,10 +156,10 @@ def _load_lib(desc: LibDescriptor, filename: str) -> ctypes.CDLL:
 
 
 def load_with_system_search(desc: LibDescriptor) -> LoadedDL | None:
-    """Try to load a library using system search paths.
+    """Try to load a library using the native Linux dynamic-loader search path.
 
     Args:
-        libname: The name of the library to load
+        desc: Descriptor for the library to load
 
     Returns:
         A LoadedDL object if successful, None if the library cannot be loaded
--- a/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_windows.py
+++ b/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_windows.py
@@ -116,10 +116,16 @@ def check_if_already_loaded_from_elsewhere(desc: LibDescriptor, have_abs_path: b
 
 
 def load_with_system_search(desc: LibDescriptor) -> LoadedDL | None:
-    """Try to load a DLL using system search paths.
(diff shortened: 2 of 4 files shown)


rwgk:
@cpcloud @mdboom Issue #1781 turned out to be a real rabbit hole, mainly because I started out with the wrong expectation (that `PATH` is meant to searched). See the comment here, and the comment right after:

* https://github.com/NVIDIA/cuda-python/pull/1506#issuecomment-4094235691

I believe simply updating the documentation is best, but please let me know if you have a different opinion.

mdboom:
This makes sense.  I think our users probably expect if they install a CTK and then install cuda_bindings that the cuda_bindings examples will "just work", and that seems to be the case here, so I'm fine with just updating docs.

PR #1800, opened 2026-03-20
Test cuda_core examples

### LINKED ISSUE #1798: Examples: Test cuda-core examples by calling main()
The cuda_core examples are currently running in CI by `exec`'ing code.  Now that they have all been refactored to have a `main()` function in #1674, we should be able to do that instead, as [this comment](https://github.com/NVIDIA/cuda-python/blob/main/cuda_core/tests/example_tests/utils.py#L31) suggests:

```
```

mdboom:
Prior to this PR, `cuda.core` examples were being parsed and imported, but not actually run, as part of CI.  This updates now actually runs the ones that we can in CI.  (Fixes #1798).

This also adds PEP 723 metadata to the examples so they can be run directly with `uv run` or `pipx run`.  This metadata is also used to determine if a particular example can run as part of CI.

### DIFF
--- a/cuda_core/cuda/core/_cpp/resource_handles.cpp
+++ b/cuda_core/cuda/core/_cpp/resource_handles.cpp
@@ -9,8 +9,13 @@
 #include <cstdint>
 #include <cstring>
 #include <mutex>
+#include <stdexcept>
 #include <unordered_map>
 #include <vector>
+
+#ifndef _WIN32
+#include <unistd.h>
+#endif
 
 namespace cuda_core {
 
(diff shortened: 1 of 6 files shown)


cuda_core/tests/example_tests/test_basic_examples.py
```diff
@@ -1,24 +1,109 @@
...
+
+
+def has_package_requirements_or_skip(example):
+    with open(example, encoding="utf-8") as f:
+        content = f.read()
+
+    # The canonical regex as defined in PEP 723
+    pep723 = re.search(r"(?m)^# /// (?P<type>[a-zA-Z0-9-]+)$\s(?P<content>(^#(| .*)$\s)+)^# ///$", content)
+    if not pep723:
+        return
```
rwgk: Could/should we enforce here that the PEP 723 information is available?

cuda_core/tests/example_tests/test_basic_examples.py
```diff
@@ -1,24 +1,109 @@
...
+        metadata[key] = value
+
+    if "dependencies" in metadata:
+        dependencies = eval(metadata["dependencies"])  # noqa: S307
+        for dependency in dependencies:
+            name = re.match("[a-zA-Z0-9_-]+", dependency)
+            try:
+                importlib.metadata.distribution(name.string)
+            except importlib.metadata.PackageNotFoundError:
+                pytest.skip(f"Skipping {example} due to missing package requirement: {name}")
```
rwgk: To avoid multiple roundtrips adding missing packages: Maybe collect all names here and move the `pytest.skip` after the loop, showing all names together?

cuda_core/tests/example_tests/test_basic_examples.py
```diff
@@ -1,24 +1,109 @@
...
+    has_package_requirements_or_skip(example_path)
+
+    system_requirement = SYSTEM_REQUIREMENTS.get(example, lambda: True)
+    if not system_requirement():
+        pytest.skip(f"Skipping {example} due to unmet system requirement")
+
+    process = subprocess.run([sys.executable, example_path], capture_output=True)  # noqa: S603
+    if process.returncode != 0:
+        if process.stdout:
+            print(process.stdout.decode())
```
rwgk: `.decode(error="replace")` to minimize distraction from the primary error

also below

rwgk (approved):
I triggered a rerun, it looks like there was some general issue starting containers or similar. Let's see if it's cleared up.

PR #1519, opened 2026-01-22
[FEA] Centralize CUDA environment variable handling and prioritize CUDA_PATH over CUDA_HOME

rparolin:
This PR introduces centralized CUDA environment variable handling across cuda-python packages with improved consistency and better handling of conflicting environment variable configurations.

 The previous implicit priority of CUDA_HOME over CUDA_PATH was inconsistent with:
-   NVIDIA's own CUDA Toolkit installers (which set CUDA_PATH by default)
-   CMake's FindCUDAToolkit module (prioritizes CUDA_PATH)

### Breaking Change - Environment Variable Priority

Changed the search order priority from `CUDA_HOME > CUDA_PATH` to `CUDA_PATH > CUDA_HOME` to align with NVIDIA tooling standards and industry conventions. This affects all packages that query CUDA Toolkit locations (cuda.pathfinder, cuda.core, cuda.bindings).

### New Module: `cuda.pathfinder._utils.env_var_constants`

- Defines canonical search order `CUDA_ENV_VARS_ORDERED = ("CUDA_PATH", "CUDA_HOME")`
- Zero-dependency module safe for bootstrap scenarios
- Single source of truth for environment variable ordering

### Enhanced `cuda.pathfinder._utils.env_vars`

- Improved `get_cuda_home_or_path()` with robust path comparison using `os.path.samefile()`
- Handles symlinks, junctions, trailing slashes, and case-insensitivity correctly
- Warns users when multiple environment variables are set but point to different locations
- Result is cached for process lifetime for consistent behavior
- Preserves empty strings when explicitly set

Closes #1433.

PR #1801, opened 2026-03-20
feat(pathfinder): centralize CUDA env var handling, prioritize `CUDA_PATH` over `CUDA_HOME`

### LINKED ISSUE #1433: [FEA]: Improve handling of CUDA_HOME and CUDA_PATH in build across projects
### Is this a duplicate?

### Area

cuda.pathfinder

Refer to comment: https://github.com/NVIDIA/cuda-python/pull/1395#issuecomment-3716120282

Implement the changes in the linked comment, ideally in pathfinder.

_No response_

### Additional context

_No response_

rwgk:
Closes #1433

Continuation of #1519 (squash-merged and rebased onto `main`; @rparolin as co-author).

## Summary

Centralizes `CUDA_PATH`/`CUDA_HOME` handling in `cuda-pathfinder` and establishes a consistent priority: **`CUDA_PATH` takes precedence over `CUDA_HOME`** when both are set.

- **New public API**: `cuda.pathfinder.get_cuda_path_or_home()` returns the highest-priority CUDA root directory. Warns when both vars are set and differ.
- **`CUDA_PATH` > `CUDA_HOME`**: All consumers now go through the centralized function. The previous ad-hoc `CUDA_HOME > CUDA_PATH` ordering in some files is removed.
- **Empty strings treated as undefined**: `CUDA_PATH=""` no longer shadows a valid `CUDA_HOME`.
- **`found_via` provenance label**: Changed from `"CUDA_HOME"` to `"CUDA_PATH"` in `LoadedDL`, `LocatedHeaderDir`, `LocatedStaticLib`, and `LocatedBitcodeLib` to match the new priority.
- **Multi-path splitting dropped**: The `os.pathsep`-based splitting in `build_hooks.py` is removed. — We are unaware of real-world usage. This could be added back later if needed. At this stage we believe the extra complexity isn't warranted.
- **Consumers migrated**: `conftest.py`, `cuda_core` tests/examples, `cuda_bindings` examples all use `get_cuda_path_or_home()`.
- **Documentation**: pathfinder 1.5.0 release notes added, `cuda_bindings` environment variable docs updated, API docs updated.

## Build hooks and `cuda.pathfinder` at build time

The `build_hooks.py` files in `cuda_bindings` and `cuda_core` currently use `os.environ.get("CUDA_PATH", os.environ.get("CUDA_HOME"))` directly instead of `get_cuda_path_or_home()`. This is necessary because in PEP 517 isolated build environments, `backend-path=["."]` causes the `cuda` namespace package to resolve to only the project's `cuda/` directory, shadowing the installed `cuda-pathfinder` in the build-env's site-packages.

A proven workaround exists: replace the `_NamespacePath` with a plain `list` that includes the site-packages `cuda/` directory, allowing `import cuda.pathfinder` to succeed. This is demonstrated and tested in PR #1803. After `cuda-pathfinder >= 1.5` is released on PyPI, a follow-up PR will switch the build hooks to use `get_cuda_path_or_home()`.

### DIFF
--- a/cuda_core/tests/test_object_protocols.py
+++ b/cuda_core/tests/test_object_protocols.py
@@ -1,10 +1,11 @@
 # SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 """
-Tests for Python object protocols (__eq__, __hash__, __weakref__, __repr__).
+Tests for Python object protocols (__eq__, __hash__, __weakref__, __repr__, pickle).
 
 This module tests that core cuda.core classes properly implement standard Python
-object protocols for identity, hashing, weak references, and string representation.
+object protocols for identity, hashing, weak references, string representation,
+and serialization.
 """
 
 import itertools
@@ -15,7 +16,17 @@ import pytest
 from helpers.graph_kernels import compile_common_kernels
 from helpers.misc import try_create_condition
 
-from cuda.core import Buffer, Device, Kernel, LaunchConfig, Program, Stream, system
+from cuda.core import (
+    Buffer,
+    Device,
+    DeviceMemoryResource,
+    DeviceMemoryResourceOptions,
+    Kernel,
+    LaunchConfig,
+    Program,
+    Stream,
+    system,
+)
 from cuda.core._graph._graphdef import GraphDef
 from cuda.core._program import _can_load_generated_ptx
 
@@ -206,6 +217,30 @@ def sample_object_code_alt(init_cuda):
 def sample_kernel_alt(sample_object_code_alt):
     """An alternate Kernel object."""
     return sample_object_code_alt.get_kernel("test_kernel_alt")
+
+
+# =============================================================================
+# Fixtures - IPC samples (for pickle tests)
+# =============================================================================
+
+POOL_SIZE = 2097152
+
+
+@pytest.fixture
+def sample_ipc_buffer_descriptor(ipc_device):
+    """An IPCBufferDescriptor."""
+    options = DeviceMemoryResourceOptions(max_size=POOL_SIZE, ipc_enabled=True)
+    mr = DeviceMemoryResource(ipc_device, options=options)
+    buf = mr.allocate(64)
+    return buf.get_ipc_descriptor()
+
+
+@pytest.fixture
+def sample_ipc_event_descriptor(ipc_device):
+    """An IPCEventDescriptor."""
+    stream = ipc_device.create_stream()
+    e = stream.record(options={"ipc_enabled": True})
+    return e.get_ipc_descriptor()
 
 
 # =============================================================================
@@ -606,6 +641,20 @@ FROM_HANDLE_COPIES = [
     ("sample_kernel", lambda k: Kernel.from_handle(int(k.handle))),
 ]
 
+# Types with __reduce__ support (pickle/cloudpickle).
+# Event, Buffer, and memory resources are excluded: Event only supports
+# IPC-based serialization via multiprocessing reduction; Buffer and memory
+# resource __reduce__ use a cross-process registry that doesn't support
+# same-process roundtrips.
+PICKLE_TYPES = [
+    "sample_device",
+    "sample_object_code_cubin",
+    "sample_ipc_buffer_descriptor",
+    "sample_ipc_event_descriptor",
+]
+
+PICKLE_MODULES = ["pickle", "cloudpickle"]
+
 # Derived type groupings for collection tests
 DICT_KEY_TYPES = sorted(set(HASH_TYPES) & set(EQ_TYPES))
 WEAK_KEY_TYPES = sorted(set(HASH_TYPES) & set(EQ_TYPES) & set(WEAKREF_TYPES))
@@ -796,3 +845,18 @@ def test_repr_format(fixture_name, pattern, request):
     obj = request.getfixturevalue(fixture_name)
     result = repr(obj)
     assert re.fullmatch(pattern, result)
+
+
+# =============================================================================
+# Pickle tests
+# =============================================================================
+
+
+@pytest.mark.parametrize("pickle_module", PICKLE_MODULES)
+@pytest.mark.parametrize("fixture_name", PICKLE_TYPES)
+def test_pickle_roundtrip(fixture_name, pickle_module, request):
+    """Object survives a pickle/cloudpickle roundtrip."""
+    mod = pytest.importorskip(pickle_module)
+    obj = request.getfixturevalue(fixture_name)
+    result = mod.loads(mod.dumps(obj))
+    assert type(result) is type(obj)


rwgk:
**Decision: drop `os.pathsep` splitting of `CUDA_HOME`/`CUDA_PATH`**

Both `cuda_bindings/build_hooks.py` and `cuda_core/build_hooks.py` currently split the env var value on `os.pathsep` and iterate over a list of CUDA roots. After reviewing the codebase and the broader ecosystem, I'm removing this pattern. Rationale:

- **No real-world usage.** The NVIDIA CTK installer (Windows), conda environments, pixi configs, and CMake's `FindCUDAToolkit` all set `CUDA_PATH`/`CUDA_HOME` to a single directory. There is no documented convention for pathsep-separated multi-path values.
- **Never tested.** The two `build_hooks.py` files even disagreed on env var priority (`CUDA_HOME > CUDA_PATH` in bindings, `CUDA_PATH > CUDA_HOME` in core) — if anyone had used multi-path, that inconsistency would have surfaced.
- **Avoids premature complexity.** Adding multi-path support to the new centralized `get_cuda_home_or_path()` (especially combined with the conflict-warning logic) would add design surface for a feature nobody uses today.

Both `_get_cuda_paths()` functions will be replaced with a plain `get_cuda_home_or_path()` call returning a single path. If a real multi-path use case surfaces later, we can add it with proper design at that point.

___

Related commit: 2164c33aeab77aabe33e0591a75a23604cd00da3

rwgk:
**Decision: treat empty `CUDA_PATH`/`CUDA_HOME` as undefined**

Rob's original implementation preserved empty strings (`CUDA_PATH=""`) as valid return values from `get_cuda_home_or_path()`. I'm changing this so that empty strings are equivalent to the variable not being set. Rationale:

- **No valid use case.** An empty string is not a valid filesystem path. Nobody intentionally sets `CUDA_PATH=""` to mean "the CUDA path is empty."
- **Fixes a subtle shadowing bug.** With the old behavior, `CUDA_PATH=""` + `CUDA_HOME=/usr/local/cuda` would return `""` (CUDA_PATH wins by priority), silently hiding the valid CUDA_HOME. Treating empty as undefined lets it correctly fall through.
- **Simplifies callers.** All downstream consumers already do `if not cuda_path: raise ...`. Returning `""` just pushes the falsy check to every call site.
- **Matches convention.** The standard `os.environ.get("CUDA_PATH")` idiom throughout the codebase has always treated empty and unset interchangeably (via `if not value`). This aligns with that expectation.

This addresses my review comment on Rob's PR: https://github.com/NVIDIA/cuda-python/pull/1519#discussion_r2729884304

rwgk:
Dump of findings from independent review with Cursor GPT-5.4 Extra High:

```
  Findings
  • High cuda_core and cuda_bindings now import get_cuda_path_or_home() in their build hooks (cuda_core/build_hooks.py:33, cuda_bindings/build_hooks.py:38), but their metadata still allows
     older cuda-pathfinder releases in both build and runtime deps (cuda_core/pyproject.toml:10, cuda_core/pyproject.toml:51, cuda_bindings/pyproject.toml:9,
    cuda_bindings/pyproject.toml:35). That means a resolver can legally choose a 1.4.x release that does not export this symbol, and isolated source builds will fail with ImportError
    before the actual CUDA-path logic runs.
  • Medium The new CUDA_PATH-first behavior is not reflected in public provenance fields. Winner selection now goes through get_cuda_path_or_home()
    (cuda_pathfinder/cuda/pathfinder/_utils/env_vars.py:77, cuda_pathfinder/cuda/pathfinder/_utils/env_vars.py:105), but env-based results still hard-code found_via="CUDA_HOME" in
    headers/libs lookup (cuda_pathfinder/cuda/pathfinder/_headers/find_nvidia_headers.py:110, cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_steps.py:198,
    cuda_pathfinder/cuda/pathfinder/_static_libs/find_static_lib.py:152, cuda_pathfinder/cuda/pathfinder/_static_libs/find_bitcode_lib.py:148). So callers can be told the path came from
    CUDA_HOME even when CUDA_PATH won.
  • Medium The header-dependent core tests now skip based on “an env var resolved” rather than “headers actually exist.” skipif_need_cuda_headers only checks the helper result in
    cuda_core/tests/conftest.py:257, while the shared helpers still require a real <cuda_root>/include directory in cuda_core/tests/helpers/__init__.py:13 and
    cuda_core/tests/helpers/__init__.py:17. With CUDA_PATH=/bad/path, these tests will run and fail for the wrong reason instead of skipping cleanly.
  • Low The docs version switcher points multiple releases at the wrong docs set. cuda_pathfinder/docs/nv-versions.json:7, cuda_pathfinder/docs/nv-versions.json:8,
    cuda_pathfinder/docs/nv-versions.json:11, cuda_pathfinder/docs/nv-versions.json:12, cuda_pathfinder/docs/nv-versions.json:15, and cuda_pathfinder/docs/nv-versions.json:16 map 1.5.0,
    1.4.3, and 1.4.2 to the 1.4.1 URL.

  Open questions / assumptions
  • I did not count the removed os.pathsep multi-root handling in the build hooks as a formal finding. It looks intentional, but it is still a compatibility change worth documenting if
    anyone relies on path-list values in CUDA_PATH or CUDA_HOME.
  • I could not run trustworthy local commands/tests in this environment because command execution was returning empty success responses, so this is a static review against main plus the
    PR/issue context rather than an executed test pass.

  Compared with main, the overall CUDA_PATH precedence cleanup looks directionally right, but I would not merge without fixing the dependency bounds first; that one is a real build-breaker.
```

rwgk:
Responses to https://github.com/NVIDIA/cuda-python/pull/1801#issuecomment-4101654463

>   • High cuda_core and cuda_bindings now import get_cuda_path_or_home() in their build hooks (cuda_core/build_hooks.py:33, cuda_bindings/build_hooks.py:38), but their metadata still allows
>      older cuda-pathfinder releases in both build and runtime deps (cuda_core/pyproject.toml:10, cuda_core/pyproject.toml:51, cuda_bindings/pyproject.toml:9,
>     cuda_bindings/pyproject.toml:35). That means a resolver can legally choose a 1.4.x release that does not export this symbol, and isolated source builds will fail with ImportError
>     before the actual CUDA-path logic runs.

This is understood: we cannot immediately pin cuda-bindings and cuda-core to cuda-pathfinder>=1.5 because that version is not released yet. We will resolve this chicken-and-egg situation by releasing cuda-pathfinder 1.5.0 (shortly after this PR is merged), then adding the pins before any new cuda-bindings or cuda-core release is made.

>   • Medium The new CUDA_PATH-first behavior is not reflected in public provenance fields. Winner selection now goes through get_cuda_path_or_home()
>     (cuda_pathfinder/cuda/pathfinder/_utils/env_vars.py:77, cuda_pathfinder/cuda/pathfinder/_utils/env_vars.py:105), but env-based results still hard-code found_via="CUDA_HOME" in
>     headers/libs lookup (cuda_pathfinder/cuda/pathfinder/_headers/find_nvidia_headers.py:110, cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_steps.py:198,
>     cuda_pathfinder/cuda/pathfinder/_static_libs/find_static_lib.py:152, cuda_pathfinder/cuda/pathfinder/_static_libs/find_bitcode_lib.py:148). So callers can be told the path came from
>     CUDA_HOME even when CUDA_PATH won.

It was an oversight that `CUDA_HOME` was still assigned to `found_via`. We decided to fix this with commit 8d3ed030597d3638e466bcc118d84346144f2527, which includes additions to the documentation to call out that the provenance label aligns with the new CUDA_PATH-first priority, not necessarily which variable supplying the value (in keeping with the status quo).
 
>   • Medium The header-dependent core tests now skip based on “an env var resolved” rather than “headers actually exist.” skipif_need_cuda_headers only checks the helper result in
>     cuda_core/tests/conftest.py:257, while the shared helpers still require a real <cuda_root>/include directory in cuda_core/tests/helpers/__init__.py:13 and
>     cuda_core/tests/helpers/__init__.py:17. With CUDA_PATH=/bad/path, these tests will run and fail for the wrong reason instead of skipping cleanly.

This was intentional. The rationale discussed before (with Opus 4.6 1M Thinking) previously was: simple is best here: it's OK if there are downstream failures if `CUDA_PATH` unexpectedly points to a location that does not have the `include` subdir.

>   • Low The docs version switcher points multiple releases at the wrong docs set. cuda_pathfinder/docs/nv-versions.json:7, cuda_pathfinder/docs/nv-versions.json:8,
>     cuda_pathfinder/docs/nv-versions.json:11, cuda_pathfinder/docs/nv-versions.json:12, cuda_pathfinder/docs/nv-versions.json:15, and cuda_pathfinder/docs/nv-versions.json:16 map 1.5.0,
>     1.4.3, and 1.4.2 to the 1.4.1 URL.

Fixed with commit 6d065e90cba17de4302016ef145a9790493c3001

>   Open questions / assumptions
>   • I did not count the removed os.pathsep multi-root handling in the build hooks as a formal finding. It looks intentional, but it is still a compatibility change worth documenting if
>     anyone relies on path-list values in CUDA_PATH or CUDA_HOME.

Yes this was intentional: https://github.com/NVIDIA/cuda-python/pull/1801#issuecomment-4100499549

Since this affects the build hooks only, we believe it'd be more distracting than helpful to mention it in the release notes or other documentation.

rwgk:
Successful full CI: https://github.com/NVIDIA/cuda-python/actions/runs/23447214012?pr=1801

rwgk:
Commit 1eff0f5a5caed229fb479c026f56f6c48fe73faf only changed the release notes, therefore I canceled the CI, to not waste our resources. See URL in previous comment for a link to a successful full run.

conftest.py
```diff
@@ -1,13 +1,14 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-import os
 
 import pytest
 
+from cuda.pathfinder import get_cuda_path_or_home
```
cpcloud, line 9: To keep the repo-level core Cython gate accurate after the env-var change, this file needs `os` again if you preserve the `include/` existence check below.

```suggestion
import os

import pytest

from cuda.pathfinder import get_cuda_path_or_home
```

conftest.py
```diff
@@ -1,13 +1,14 @@
...
+
 
 def pytest_collection_modifyitems(config, items):  # noqa: ARG001
-    cuda_home = os.environ.get("CUDA_HOME")
+    cuda_home = get_cuda_path_or_home()
```
cpcloud: Blocking: this now treats any non-empty `CUDA_PATH`/`CUDA_HOME` as if headers are available. If the env var points at a stale or incomplete root, the core Cython tests stop skipping and fail later in compilation instead. Please keep the `include/` existence check here.

```suggestion
    cuda_root = get_cuda_path_or_home()
    cuda_home = cuda_root if cuda_root is not None and os.path.isdir(os.path.join(cuda_root, "include")) else None
```

cuda_core/tests/conftest.py
```diff
@@ -253,6 +255,6 @@ def test_something(memory_resource_factory):
 
 
 skipif_need_cuda_headers = pytest.mark.skipif(
-    not os.path.isdir(os.path.join(os.environ.get("CUDA_PATH", ""), "include")),
+    get_cuda_path_or_home() is None,
     reason="need CUDA header",
 )
```
cpcloud, line 275: Same regression here: any non-empty `CUDA_PATH`/`CUDA_HOME` now marks headers as available. `test_cooperative_launch()` includes `<cooperative_groups.h>`, so a stale toolkit root stops skipping and fails inside `Program.compile()` instead. Please keep the include-dir check at this marker.

```suggestion
skipif_need_cuda_headers = pytest.mark.skipif(
    (cuda_root := get_cuda_path_or_home()) is None or not os.path.isdir(os.path.join(cuda_root, "include")),
    reason="need CUDA header",
)
```

cpcloud (changes requested):
Two regressions make the new `CUDA_PATH` flow harder to use in practice:

1. `conftest.py` now imports `cuda.pathfinder` during collection, which breaks repo-root `pytest` runs unless `cuda-pathfinder` is already installed. The old file only depended on stdlib + pytest, so this is a new collection-time dependency.
2. The new header gates in `conftest.py` and `cuda_core/tests/conftest.py` only check whether a CUDA env var is set. If `CUDA_PATH`/`CUDA_HOME` points at a stale or incomplete root, we stop skipping and fail later in compilation instead.

I left inline suggestions on the affected lines.

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_steps.py
```diff
@@ -183,20 +183,23 @@ def find_in_conda(ctx: SearchContext) -> FindResult | None:
 
 
 def find_in_cuda_home(ctx: SearchContext) -> FindResult | None:
```
cpcloud: Non-blocking architecture note: after the recent descriptor/search-step refactors, this helper now represents the shared CUDA env-var anchor rather than specifically `CUDA_HOME`. Renaming this step (and the analogous header step/tests) to something like `find_in_cuda_env()` would keep the step layer aligned with the centralized `get_cuda_path_or_home()` helper and the new `found_via="CUDA_PATH"` provenance.

```suggestion
def find_in_cuda_env(ctx: SearchContext) -> FindResult | None:
```

conftest.py
rwgk: Done: commit 9c13237ba898793c85939f7e08f414a6a4e5280a

I believe skipping if `CUDA_PATH` is set, but there is no `include/`, is likely to mask oversights. My commit introduces a hard, obvious failure in that situation. I believe that'll be more helpful than skipping, which is likely to get overlooked.

cuda_core/tests/conftest.py
rwgk: This was deliberate ([see comment above](https://github.com/NVIDIA/cuda-python/pull/1801#issuecomment-4102220707)), but commit 9c13237ba898793c85939f7e08f414a6a4e5280a resolves this, and your other feedback above, in a safer way:

* We skip if `CUDA_PATH` / `CUDA_HOME` is not set.
* If set, we fail hard if the `include/` directory doesn't exist.

(I believe in practice that amounts to "hey, your `CUDA_PATH` is pointing to the wrong directory"; which will be obvious.)

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_steps.py
rwgk: Done: commit 025508397bef0a9442ef3fd644e6f61a26605c22

I decided to simply make this `find_in_cuda_path`. `cuda_env` is too ambiguous (no obvious connection to `CUDA_PATH` or `CUDA_HOME`).

Thanks for catching this; I overlooked it.

PR #1813, opened 2026-03-24
Fix typo in nvbug5880275

mdboom:


leofang:
I wonder why our linter did not catch this typo...

mdboom:
> I wonder why our linter did not catch this typo...

Our linter can't see into Cython code.  (See the discussion about including `.pyi` files).

PR #1825, opened 2026-03-25
Standardize internal version checks in cuda.core

Andy-Jost:
## Summary

Centralizes all internal CUDA version checking into a new `cuda/core/_utils/version.{pyx,pxd}` module, replacing ad-hoc patterns scattered across multiple files. Both Python (`binding_version`, `driver_version`) and Cython (`cy_binding_version`, `cy_driver_version`) accessors are provided, each returning cached `(major, minor, patch)` tuples.

## Changes

- **New module** `cuda/core/_utils/version.{pyx,pxd}` with:
  - `binding_version()` / `cy_binding_version()`: returns `(major, minor, patch)` from `cuda-bindings` metadata, cached via `@functools.cache` (Python) and module-level globals (Cython)
  - `driver_version()` / `cy_driver_version()`: returns `(major, minor, patch)` converted from `cuDriverGetVersion`, same caching strategy
- **Updated all internal version comparisons** to use the new functions and tuple format (e.g., `(12, 3, 0)` instead of `12030` or `(12, 3)`)
- **Removed dead code** in `_module.pyx`, `_launcher.pyx`, and `_launch_config.pyx`:
  - `_lazy_init` / `_inited` / `_lock` / `_use_ex` lazy initialization machinery
  - `_kernel_ctypes` / `_get_kernel_ctypes` / `_is_cukernel_get_library_supported` unused helpers
  - Pre-CUDA 12.0 version gates (CUDA 12.0 is the minimum supported version)
- **Standardized error messages** to format versions as `'.'.join(map(str, version_tuple))`
- **Removed** `get_binding_version` and `get_driver_version` from `cuda_utils.pyx`
- **Updated all test files** to use the new version module
- **Left untouched**: the public NVML-based `system.get_driver_version` API, which serves a different purpose (user-facing system info vs. internal feature gating)

## Test Coverage

- Updated `test_device.py`, `test_module.py`, `test_program.py`, `test_cuda_utils.py`, `test_optional_dependency_imports.py`, and `graph/test_explicit.py` to use new version functions
- `_program.pyx` and `_linker.pyx` use the Python `def` versions (`binding_version`, `driver_version`) so that `test_optional_dependency_imports.py` can continue to monkeypatch them for controlled testing

### DIFF
--- a/cuda_core/cuda/core/_utils/version.pyx
+++ b/cuda_core/cuda/core/_utils/version.pyx
@@ -9,7 +9,7 @@ from cuda.core._utils.cuda_utils import driver, handle_return
 
 
 @functools.cache
-def binding_version():
+def binding_version() -> tuple[int, int, int]:
     """Return the cuda-bindings version as a (major, minor, patch) triple."""
     try:
         parts = importlib.metadata.version("cuda-bindings").split(".")[:3]
@@ -19,10 +19,10 @@ def binding_version():
 
 
 @functools.cache
-def driver_version():
+def driver_version() -> tuple[int, int, int]:
     """Return the CUDA driver version as a (major, minor, patch) triple."""
     cdef int ver = handle_return(driver.cuDriverGetVersion())
-    return (ver // 1000, (ver % 1000) // 10, ver % 10)
+    return (ver // 1000, (ver // 10) % 100, ver % 10)
 
 
 cdef tuple _cached_binding_version = None


cuda_core/cuda/core/_utils/version.pyx
```diff
@@ -0,0 +1,43 @@
...
+from cuda.core._utils.cuda_utils import driver, handle_return
+
+
+@functools.cache
+def binding_version() -> tuple[int, int, int]:
+    """Return the cuda-bindings version as a (major, minor, patch) triple."""
+    try:
+        parts = importlib.metadata.version("cuda-bindings").split(".")[:3]
+    except importlib.metadata.PackageNotFoundError:
+        parts = importlib.metadata.version("cuda-python").split(".")[:3]
```
rwgk, line 17: I'd put a simple sanity check:

```
    assert len(parts) == 3
```

* Cheap
* Should never fire
* But if it does, it can prevent a lot of confusion

(I'd deliberately not make this any more sophisticated.)

cuda_core/cuda/core/_utils/version.pyx
```diff
@@ -0,0 +1,43 @@
...
+    return tuple(int(v) for v in parts)
+
+
+@functools.cache
+def driver_version() -> tuple[int, int, int]:
+    """Return the CUDA driver version as a (major, minor, patch) triple."""
+    cdef int ver = handle_return(driver.cuDriverGetVersion())
+    return (ver // 1000, (ver // 10) % 100, ver % 10)
```
rwgk, line 25: According to the below (ChatGPT Pro, instant), this is "technically correct but semantically misleading." I'd go with Option B.

___

## User Prompt

I'm reviewing this cython code (NVIDIA/cuda-python repo):

```
@functools.cache
def driver_version() -> tuple[int, int, int]:
    """Return the CUDA driver version as a (major, minor, patch) triple."""
    cdef int ver = handle_return(driver.cuDriverGetVersion())
    return (ver // 1000, (ver // 10) % 100, ver % 10)
```

___

On the CTK side I see this (the last digit is always zero, irrespective of the patch version):

```
$ grep '#define CUDA_VERSION' $cudah
./12.0.1_525.85.12/cuda_cudart/targets/x86_64-linux/include/cuda.h:#define CUDA_VERSION 12000
./12.1.1_530.30.02/cuda_cudart/targets/x86_64-linux/include/cuda.h:#define CUDA_VERSION 12010
./12.2.2_535.104.05/cuda_cudart/targets/x86_64-linux/include/cuda.h:#define CUDA_VERSION 12020
./12.3.2_545.23.08/cuda_cudart/targets/x86_64-linux/include/cuda.h:#define CUDA_VERSION 12030
./12.4.1_550.54.15/cuda_cudart/targets/x86_64-linux/include/cuda.h:#define CUDA_VERSION 12040
./12.5.1_555.42.06/cuda_cudart/targets/x86_64-linux/include/cuda.h:#define CUDA_VERSION 12050
./12.6.3_560.35.05/cuda_cudart/targets/x86_64-linux/include/cuda.h:#define CUDA_VERSION 12060
./12.8.1_570.124.06/cuda_cudart/targets/x86_64-linux/include/cuda.h:#define CUDA_VERSION 12080
./12.9.1_575.57.08/cuda_cudart/targets/x86_64-linux/include/cuda.h:#define CUDA_VERSION 12090
./13.0.2_580.95.05/cuda_cudart/targets/x86_64-linux/include/cuda.h:#define CUDA_VERSION 13000
./13.1.1_590.48.01/cuda_cudart/targets/x86_64-linux/include/cuda.h:#define CUDA_VERSION 13010
./13.2.0_595.45.04/cuda_cudart/targets/x86_64-linux/include/cuda.h:#define CUDA_VERSION 13020
```

___

But what about the driver? Does the last digit there have a significance? Is it ever different from 0?


---

## Assistant Response

Short answer: your instinct is right to question it — the **“patch” digit in `cuDriverGetVersion()` is not meaningful in practice and is effectively always 0**.

### 1. What `cuDriverGetVersion()` actually returns

From the CUDA driver API, `cuDriverGetVersion()` returns an integer encoding:

```
version = 1000 * major + 10 * minor
```

That’s it. There is **no patch component encoded**.

Examples:
- CUDA 12.6 → `12060`
- CUDA 13.1 → `13010`

If decoded as:
```
(ver // 1000, (ver // 10) % 100, ver % 10)
```

you get `(12, 6, 0)` — but that last `0` is just an artifact.

---

### 2. Why your CTK observation matches

`CUDA_VERSION` uses the same encoding:

```
#define CUDA_VERSION 12060
```

So:
- Toolkit patch versions (e.g. 12.6.3) are **not represented**
- The last digit is always `0`

---

### 3. Driver vs Toolkit versioning

**CUDA Toolkit**
- Has real patch versions (e.g. 12.6.3)
- `CUDA_VERSION` does not encode them

**CUDA Driver**
- Version tied to NVIDIA driver (e.g. 560.35.05)
- `cuDriverGetVersion()` reports only **CUDA API compatibility level**

---

### 4. Is the last digit ever non-zero?

No — it is structurally always zero.

---

### 5. Practical implication

Your code is technically correct but semantically misleading:

```
return (ver // 1000, (ver // 10) % 100, ver % 10)
```

It suggests a meaningful patch version, which does not exist.

---

### 6. Recommendation

**Option A (preferred):**
```
def driver_version() -> tuple[int, int]:
    return (ver // 1000, (ver // 10) % 100)
```

**Option B (if API stability matters):**
Keep the triple but document that patch is always zero.
```
"""Return (major, minor, patch=0). Patch is always 0 for driver."""
```
---

### Bottom line

- Patch digit is always zero
- No semantic meaning
- Treat as (major, minor)

---

### Bonus

To get the actual driver version (e.g. `560.35.05`), use NVML:
- `nvmlSystemGetDriverVersion()`

cuda_core/cuda/core/_program.pyx
```diff
@@ -520,10 +520,10 @@ def _get_nvvm_module():
     _nvvm_import_attempted = True
 
     try:
-        version = get_binding_version()
-        if version < (12, 9):
+        version = binding_version()
+        if version < (12, 9, 0):
             raise RuntimeError(
-                f"NVVM bindings require cuda-bindings >= 12.9.0, but found {version[0]}.{version[1]}.x. "
```
Andy-Jost, line 526: Here is the old code using a three-part version for cuda-bindings.

cuda_core/cuda/core/_utils/version.pyx
Andy-Jost: I was going to ask whether to make bindings a pair or triple until I found the three-part format in the existing code [here](https://github.com/NVIDIA/cuda-python/pull/1825/changes#r2991948196).

The danger with mixing these is that comparisons between two- and three-part tuples will give inconsistent and surprising results (see below). Rather than trying to remember which version uses which format, it's better to follow a uniform rule: make every version a three-tuple and always compare against a three-tuple. That's a simple rule everyone can remember, and it makes inconsistencies easy to spot.

```
>>> (12, 3) <= (12, 3, 0)
True
>>> (12, 3) >= (12, 3, 0)
False
```

cuda_core/cuda/core/_utils/version.pyx
Andy-Jost: This is covered by the trailing `[:3]` at lines 15, 17.

PR #1826, opened 2026-03-26
Remove workaround related to `LongPathsEnabled` on Windows

### LINKED ISSUE #1820: [BUG]: Windows CI runners need LongPathsEnabled for free-threaded Python
## Problem

The `nvidia-cutlass` 4.2.0.0 PyPI wheel contains files with very long paths. When installed under free-threaded Python on Windows, one path reaches **266 characters**, exceeding the default Windows MAX_PATH limit of 260:

```
C:\actions-runner\_work\_tool\Python\3.14.3\x64-freethreaded\Lib\site-packages\cutlass_library\source\examples\68_hopper_fp8_warp_specialized_grouped_gemm_with_blockwise_scaling\68_hopper_fp8_warp_specialized_grouped_gemm_with_blockwise_scaling_with_sparse_groups.cu
```

The `x64-freethreaded` directory name is 12 characters longer than the `x64` used by regular CPython, which is just enough to push this path over the limit. Regular (non-free-threaded) Python installs the same wheel without issue.

The `pip install` step fails with:

```
ERROR: Could not install packages due to an OSError.
HINT: This error might have occurred since this system does not have Windows Long Path support enabled.
  ...
FileNotFoundError: [Errno 2] No such file or directory: 'C:\actions-runner\_work\_tool\Python\3.14.3\x64-freethreaded\Lib\site-packages\cutlass_library\source\examples\68_hopper_fp8_...'
```

## Affected jobs

All `win-64` / `py3.14t` test jobs that run `pip install --group test-cu{12,13}` (which pulls in `nvidia-cutlass`):

- Test win-64 / py3.14t, 12.9.1, local, l4 (TCC)
- Test win-64 / py3.14t, 13.0.2, wheels, a100 (MCDM)
- Test win-64 / py3.14t, 13.2.0, wheels, a100 (MCDM)

## Current workaround

PR #1816 skips the `pip install --group` and the subsequent `all_must_work` pathfinder test run for free-threaded Python on Windows. The `see_what_works` first pass still runs, so free-threaded builds are not completely untested.

## Fix

Enable the Windows Long Path registry key on the CI runners:

```
HKLM\SYSTEM\CurrentControlSet\Control\FileSystem\LongPathsEnabled = 1
```

Once enabled, the workaround in the CI workflow can be removed.

## References

- pip documentation: https://pip.pypa.io/warnings/enable-long-paths
- Microsoft documentation: https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation

rwgk:
Closes #1820

The CI runner images now have `LongPathsEnabled=1` (nv-gha-runners/vm-images#241), so the `nvidia-cutlass` wheel installs successfully under free-threaded Python on Windows even though the longest path (266 characters) exceeds the default MAX_PATH limit.

## Changes

- Remove the `if: !endsWith(matrix.PY_VER, 't')` guards that skipped `pip install --group` and the `all_must_work` pathfinder test for free-threaded Python on Windows
- Add a `Verify LongPathsEnabled` step that fails early with a clear error referencing #1820 if a runner lacks the registry setting

### DIFF
--- a/.github/workflows/build-wheel.yml
+++ b/.github/workflows/build-wheel.yml
@@ -28,12 +28,11 @@ jobs:
       fail-fast: false
       matrix:
         python-version:
-          # TEMPORARY: only building 3.14t to verify LongPathsEnabled
-          # - "3.10"
-          # - "3.11"
-          # - "3.12"
-          # - "3.13"
-          # - "3.14"
+          - "3.10"
+          - "3.11"
+          - "3.12"
+          - "3.13"
+          - "3.14"
           - "3.14t"
     name: py${{ matrix.python-version }}
     runs-on: ${{ (inputs.host-platform == 'linux-64' && 'linux-amd64-cpu8') ||
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,7 +106,7 @@ jobs:
         host-platform:
           - linux-aarch64
     name: Build ${{ matrix.host-platform }}, CUDA ${{ needs.ci-vars.outputs.CUDA_BUILD_VER }}
-    if: github.event_name == 'TEMPORARY_DISABLED'  # TEMPORARY: only testing Windows 3.14t
+    if: ${{ github.repository_owner == 'nvidia' && !fromJSON(needs.should-skip.outputs.skip) && !fromJSON(needs.should-skip.outputs.doc-only) }}
     secrets: inherit
     uses: ./.github/workflows/build-wheel.yml
     with:
@@ -145,7 +145,7 @@ jobs:
         host-platform:
           - linux-64
     name: Test ${{ matrix.host-platform }}
-    if: github.event_name == 'TEMPORARY_DISABLED'  # TEMPORARY: only testing Windows 3.14t
+    if: ${{ github.repository_owner == 'nvidia' && !fromJSON(needs.should-skip.outputs.doc-only) }}
     permissions:
       contents: read  # This is required for actions/checkout
     needs:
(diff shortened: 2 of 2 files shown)


.github/workflows/test-wheel-windows.yml
```diff
@@ -208,6 +208,15 @@ jobs:
         with:
           python-version: ${{ matrix.PY_VER }}
 
+      - name: Verify LongPathsEnabled
+        run: |
+          $val = (Get-ItemProperty -Path 'HKLM:\SYSTEM\CurrentControlSet\Control\FileSystem' -Name 'LongPathsEnabled').LongPathsEnabled
+          echo "LongPathsEnabled = $val"
+          if ($val -ne 1) {
+            echo "::error::LongPathsEnabled is not set to 1 (see issue #1820)"
```
cpcloud, line 216:
```suggestion
          if ($val -ne 1) {
            echo "::error::LongPathsEnabled is $val but should be set to 1 (see issue #1820)"
```

No reason to print the value if it's 1, nor is there a reason to be so cryptic in the error message: just combine saying what it is with what is expected.

rwgk: This is for a CI log, having a confirmation about `LongPathsEnabled` is useful (especially for LLMs to build context). Why do you not want to show that one line?

The output with the code as-is:

```
LongPathsEnabled = 0
::error::LongPathsEnabled is not set to 1 (see issue #1820)
```

Seems very clear to me:
* Problem: ::error::LongPathsEnabled is not set to 1
* Solution: see issue #1820

The solution needs some explanation (unless you know about it already), so the pointer to the issue is a very efficient way to helpful.

cpcloud: I don't follow the LLM argument here. Is an LLM concerned about non failure case? Why does it need to have context in the boring, successful case. The context is there from the error message.

I personally despise output fluff like `"build done"` or `"things went well"`, or just echoing information just because maybe it'll be useful someday.

Unless it's notable, actionable information, like a failure or actionable warning, then there should be no output. I'm not saying there are no exceptions to that, but this to me is a clear case of don't add any more spam to CI logs than necessary.

PR #1835, opened 2026-03-30
Improve error message for ManagedMemoryResource() on unsupported platforms

### LINKED ISSUE #1617: ManagedMemoryResource() without options fails when the default pool does not support managed allocations
`ManagedMemoryResource()` (no options) attempts to retrieve the current memory pool via `cuMemGetMemPool` with `CU_MEM_ALLOCATION_TYPE_MANAGED`. On platforms where the default pool does not support managed allocations, this fails with a cryptic error:

```
CUDAError: CUDA_ERROR_NOT_SUPPORTED: This error indicates that the attempted operation is not supported
on the current system or device.
```

Meanwhile, `ManagedMemoryResource(options=ManagedMemoryResourceOptions(preferred_location=0))` works on the same platform because it creates a new managed pool via `cuMemPoolCreate`.

Improve the error message so users understand what went wrong, and investigate whether the API should handle this case more gracefully (e.g., by falling back to pool creation).

**Reproducer**:

```python
from cuda.core import Device, ManagedMemoryResource, ManagedMemoryResourceOptions

device = Device(0)
device.set_current()

# Fails with CUDA_ERROR_NOT_SUPPORTED
mr = ManagedMemoryResource()

# Workaround: succeeds
mr = ManagedMemoryResource(options=ManagedMemoryResourceOptions(preferred_location=0))
```

**Test results**:

On WSL2 x86_64 (`concurrent_managed_access=False`, `memory_pools_supported=True`):
```
ManagedMemoryResource() ...
  failed: CUDA_ERROR_NOT_SUPPORTED: This error indicates that the attempted
  operation is not supported on the current system or device.

ManagedMemoryResource(options=ManagedMemoryResourceOptions(preferred_location=0)) ...
  succeeded (device_id=0)
```

On Linux x86_64 (`concurrent_managed_access=True`, `memory_pools_supported=True`):
```
ManagedMemoryResource() ...
  succeeded (device_id=-1)

ManagedMemoryResource(options=ManagedMemoryResourceOptions(preferred_location=0)) ...
  succeeded (device_id=0)
```

Andy-Jost:
Closes #1617

## Summary

`ManagedMemoryResource()` (no options) calls `cuMemGetMemPool` to retrieve the default managed memory pool, but on platforms without concurrent managed access (e.g. WSL2), this fails with a cryptic `CUDA_ERROR_NOT_SUPPORTED`. Meanwhile, explicitly creating a pool via `ManagedMemoryResource(options=ManagedMemoryResourceOptions(...))` works fine on the same platform.

This PR catches the error and re-raises it as a `RuntimeError` with an actionable message pointing users to the explicit options path. The improved message is only emitted when concurrent managed access is confirmed to be unavailable; otherwise the original `CUDAError` propagates unchanged.

The error is identified via string match on the `CUDAError` message rather than inspecting a structured error code because (1) we prefer not to change the `CUDAError` class or the `MP_init_current_pool` API for this, and (2) this is not a hot path.

## Changes
- `_managed_memory_resource.pyx` — catch `CUDAError` from `MP_init_current_pool` in the `opts is None` path; check `concurrent_managed_access` via device properties and raise a clear `RuntimeError` when applicable
- `_memory_pool.pyx` — improve the CUDA < 13 fallback error message in `MP_init_current_pool` to describe the unsupported operation
- `test_managed_memory_warning.py` — add `test_default_pool_error_without_concurrent_access` using the existing `device_without_concurrent_managed_access` fixture

## Test Plan
- [x] Reproduced on WSL2 (RTX 3500 Ada, `concurrent_managed_access=False`)
- [x] Verified fix produces the improved error message
- [x] CI

### DIFF
--- a/cuda_core/cuda/core/_memory/_memory_pool.pyx
+++ b/cuda_core/cuda/core/_memory/_memory_pool.pyx
@@ -257,7 +257,9 @@ cdef int MP_init_current_pool(
         self._h_pool = create_mempool_handle_ref(pool)
         self._mempool_owned = False
     ELSE:
-        raise RuntimeError("not supported")
+        raise RuntimeError(
+            "Getting the current memory pool requires CUDA 13.0 or later"
+        )
     return 0
 
 
--- a/cuda_core/cuda/core/_memory/_managed_memory_resource.pyx
+++ b/cuda_core/cuda/core/_memory/_managed_memory_resource.pyx
@@ -11,6 +11,7 @@ from cuda.core._utils.cuda_utils cimport (
     HANDLE_RETURN,
     check_or_create_options,
 )
+from cuda.core._utils.cuda_utils import CUDAError
 
 from dataclasses import dataclass
 import threading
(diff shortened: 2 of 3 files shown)


cuda_core/cuda/core/_memory/_memory_pool.pyx
```diff
@@ -257,7 +257,10 @@ cdef int MP_init_current_pool(
         self._h_pool = create_mempool_handle_ref(pool)
         self._mempool_owned = False
     ELSE:
-        raise RuntimeError("not supported")
+        raise RuntimeError(
+            "Getting the current memory pool for a memory location and "
+            "allocation type requires CUDA 13.0 or later"
+        )
```
Andy-Jost, line 262: Unrelated fix, but a needed improvement.

PR #1836, opened 2026-03-31
[no-ci] fix(ci): pr-metadata-check — blocked-label bug, exact match, multi-word labels

rwgk:
## Summary

Fixes the PR metadata workflow’s blocked-label false positive and incorrect handling of multi-word GitHub labels; tightens blocked-label detection to **exact** name matches (case-insensitive); error text for blocked labels uses **GitHub’s spelling** of the label name.

## Fixes

1. **Blocked-label false positive** — `BLOCKED_PATTERNS` was unquoted in a `for pattern in $BLOCKED_PATTERNS` loop, so the shell split the string into tokens (`blocked`, `DO`, `NOT`, …). Each token was passed to `grep` against the full label name, so `DO` matched the substring `do` in **`documentation`**, incorrectly reporting a blocked label. The workflow now derives blocked matches from the JSON payload with `jq`; error lines use each label’s **original** spelling from GitHub, not a lowercased form.

2. **Module / type label iteration** — `LABEL_NAMES` was filled with `jq -r '.[].name'` but then iterated with `for label in $LABEL_NAMES`, so **multi-word** label names were split on spaces and never matched the allowed module/type lists. Module and type checks now read label names **line-by-line** from `jq` (and the success summary builds the label list from the same `jq` stream).

## Behavior change (intentional)

**Blocked labels** — Detection no longer uses substring `grep` on label text. It matches only labels whose **name** is exactly `blocked` or `do not merge`, **case-insensitive** (via `jq`). Custom labels that merely *contain* those words (e.g. `blocked-by-review`) are **not** treated as blocked; that is stricter than substring matching and matches reviewer feedback.

___

### Workflow logs reviewed while working on this PR

| Run | Labels | Expected failures | Blocked false positive? |
|---|---|---|---|
| [23748478947](https://github.com/NVIDIA/cuda-python/actions/runs/23748478947/job/69206957035?pr=1822) (PR #1822) | `documentation`, `cuda.bindings` | assignee, milestone | Yes -- `documentation` flagged |
| [23748648922](https://github.com/NVIDIA/cuda-python/actions/runs/23748648922/job/69184145394?pr=1830) (PR #1830) | `enhancement`, `feature`, `cuda.bindings` | assignee | No |
| [23757944518](https://github.com/NVIDIA/cuda-python/actions/runs/23757944518/job/69217560355?pr=1793) (PR #1793) | (none) | module label, type label, milestone | No |


.github/workflows/pr-metadata-check.yml
```diff
@@ -84,15 +84,14 @@ jobs:
...
-            done
-          done
+          # Read labels line-by-line (jq outputs one per line) to handle
+          # multi-word label names; match with a single regex to avoid
+          # word-splitting "DO NOT MERGE" into individual grep patterns.
+          while IFS= read -r label; do
+            if echo "$label" | grep -qiE "blocked|do not merge"; then
+              ERRORS="${ERRORS}- **Blocked label detected**: label \`$label\` prevents merging. Remove it when the PR is ready.\n"
+            fi
+          done < <(echo "$LABELS" | jq -r '.[].name')
```
cpcloud: Nice catch on the word-splitting bug. Could we take this one step further and do an exact, case-insensitive `jq` check from `LABELS` instead of another `grep` pattern?

```suggestion
          # Match blocked labels exactly (case-insensitively) using the PR
          # payload we already have, instead of doing substring matches.
          BLOCKED_LABELS=$(jq -r '
            [.[].name | ascii_downcase] as $pr_labels
            | ["blocked", "do not merge"] as $blocking
            | $pr_labels - ($pr_labels - $blocking)
            | .[]
          ' <<<"$LABELS")
          while IFS= read -r label; do
            [ -n "$label" ] || continue
            ERRORS="${ERRORS}- **Blocked label detected**: label \`$label\` prevents merging. Remove it when the PR is ready.\n"
          done <<<"$BLOCKED_LABELS"
```

rwgk:
Archiving a Cursor-generated review below. I'll try to get it to fix the "Minor UX detail" and `LABEL_NAMES` issue it uncovered.

___

## Net change @ c6b46b5

**Scope:** `.github/workflows/pr-metadata-check.yml` — the “blocked label” block only (replaces the nested `for` / `grep` loop with a `jq` pipeline + `while read` loop).

### Bug fix

The previous code put `BLOCKED_PATTERNS="blocked DO NOT MERGE do not merge"` in an unquoted `for pattern in $BLOCKED_PATTERNS` loop. The shell word-split that string into tokens (`blocked`, `DO`, `NOT`, `MERGE`, …). Each token was passed to `grep -qi` against the **full** label string, so the token `DO` matched the substring `do` in **`documentation`**, producing a false “Blocked label detected” result.

### Follow-up behavior change (reviewer suggestion)

The fix is not only “stop splitting patterns.” The shipped logic also changes **substring** matching (`grep` on label text) to **exact label name** matching (case-insensitive): only labels whose GitHub name is exactly `blocked` or `do not merge` (after lowercasing in `jq`) are treated as blocked.

So compared to the old intent of “grep for blocked-ish text,” this is **stricter**: a label like `blocked-by-security` (or any name that merely **contains** `blocked` but is not exactly `blocked`) would have been caught by substring `grep` but will **not** be caught by the `jq` intersection. That tightening is intentional per review; it should be explicit in the PR description.

### Minor UX detail

Error messages use the lowercased names emitted by `jq`, not necessarily the exact casing shown in the GitHub UI.

### Unchanged in this PR

`LABEL_NAMES` is still built with `jq -r '.[].name'` and iterated with `for label in $LABEL_NAMES`, which can still mishandle **multi-word labels** for the module/type label checks. The blocked-label path correctly avoids that by reading from `LABELS` via `jq` instead of reusing `$LABEL_NAMES` for matching.

rwgk:
@cpcloud this is ready for review again. I replaced the PR title and rewrote the description (there are now three changes in this PR).

Issue #1457, opened 2026-01-12
Expose NVVM version detection helper as public API


leofang:
This
https://github.com/NVIDIA/cuda-python/blob/da7eb1f5a97aa21d8f78098e13e7c4edad013530/cuda_core/tests/test_program.py#L58-L132
has a real need (ex: https://github.com/NVIDIA/numba-cuda/pull/681). We should expose it to `cuda.bindings.utils`, similar to the PTX version helpers.

abhilash1910:
@leofang  I  an add  this for  NVVM .

PR #1837, opened 2026-03-31
[NVVM] Expose nvvm version detection in cuda.bindings.utils.

abhilash1910:
## Description

Fixes issue #1457 . Provides an api call to `cuda.bindings.utils` to check the version of nvvm. 
@leofang @rwgk pinging for review.


abhilash1910:
pre-commit.ci autofix

cuda_bindings/cuda/bindings/utils/_nvvm_utils.py
```diff
@@ -0,0 +1,93 @@
...
+
+!nvvm.annotations = !{{!0}}
+!0 = !{{void ()* @dummy_kernel, !"kernel", i32 1}}
+
+!nvvmir.version = !{{!1}}
+!1 = !{{i32 {major}, i32 {minor}, i32 {debug_major}, i32 {debug_minor}}}
+"""
+
+
+def check_nvvm_options(options: Sequence[bytes]) -> bool:
```
rwgk: For a public Python API, `bytes` seems unusual here.

Most pythonic would be `Sequence[str]`, but `Sequence[str | bytes]` would seem fine, too.

The implementation below actually converts `bytes` to `str`:

```
        options_list = [opt.decode("utf-8") if isinstance(opt, bytes) else opt for opt in options]
```

So requiring `bytes` in the API is especially surprising. My recommendation is to simply use `Sequence[str]`; then the `options_list` line can be removed completely.

I also recommend using `check_nvvm_compiler_options` as the function name, for clarity, especially in the call sites.

cuda_bindings/cuda/bindings/utils/_nvvm_utils.py
```diff
@@ -0,0 +1,93 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026-2027 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
rwgk: Remove `-2027`

cuda_bindings/cuda/bindings/utils/_nvvm_utils.py
```diff
@@ -0,0 +1,93 @@
...
+    >>> check_nvvm_options([b"-arch=compute_90", b"-numba-debug"])
+    True  # if -numba-debug is supported by the installed libNVVM
+    """
+    try:
+        from cuda.bindings import nvvm
+        from cuda.bindings._internal.nvvm import _inspect_function_pointer
+
+        if _inspect_function_pointer("__nvvmCreateProgram") == 0:
+            return False
+    except Exception:
```
rwgk: This is far too broad. It's very likely to mask bugs. Also the other `except Exception` further below. Could you please try this in Cursor:

> Could you please narrow down the exception catching as much as possible? Please use `ModuleNotFoundError` for the `nvvm` import; we also want to check `exc.name == "nvvm"`. If that import works, we don't want to guard the `_inspect_function_pointer` import, or the `_inspect_function_pointer()` call. If those don't work, that's a bug we want to surface.

cuda_bindings/cuda/bindings/utils/_nvvm_utils.py
```diff
@@ -0,0 +1,93 @@
...
+            program,
+            precheck_ir_bytes,
+            len(precheck_ir_bytes),
+            "precheck.ll",
+        )
+
+        options_list = [opt.decode("utf-8") if isinstance(opt, bytes) else opt for opt in options]
+        nvvm.verify_program(program, len(options_list), options_list)
+        nvvm.compile_program(program, len(options_list), options_list)
+    except Exception:
```
rwgk: If you're compiling anyway, do you actually need the `verify_program()` call?

I'm thinking the `try` should just be around this one line:

```python
        try:
            nvvm.compile_program(prog, len(options), options)
        except nvvm.nvvmError as e:
            # can we add something here to ensure we're not masking errors other than invalid options?
```

I believe it's really important to take great care that we're not masking actual errors; e.g. the hard-wired `_PRECHECK_NVVM_IR` might need tweaks for future GPU generations. If we're simply reporting any error as "invalid compiler option", it'll potentially take someone downstream a long time to drill down all the way back here.

cuda_bindings/cuda/bindings/utils/_nvvm_utils.py
```diff
@@ -0,0 +1,93 @@
...
+    bool
+        True if the options are supported, False otherwise.
+
+    Examples
+    --------
+    >>> from cuda.bindings.utils import check_nvvm_options
+    >>> check_nvvm_options([b"-arch=compute_90", b"-g"])
+    True
+    >>> check_nvvm_options([b"-arch=compute_90", b"-numba-debug"])
+    True  # if -numba-debug is supported by the installed libNVVM
```
rwgk: Could we use a non-numba example here?

cuda_core/tests/test_program.py
```diff
@@ -70,81 +70,14 @@ def _has_nvrtc_pch_apis_for_tests():
...
-            ptx_version = get_ptx_ver(ptx_str)
-            cuda_version = get_minimal_required_cuda_ver_from_ptx_ver(ptx_version)
-            _libnvvm_version = cuda_version
-            return _libnvvm_version
-        finally:
-            nvvm.destroy_program(program)
+        from cuda.bindings.utils import check_nvvm_options
 
+        return check_nvvm_options([f"-arch={arch}".encode()])
     except Exception:
```
rwgk: It's only a test, therefore not nearly as critical as in the production code, but we may miss regressions if this isn't handled with similar care.

cuda_bindings/cuda/bindings/utils/_nvvm_utils.py
abhilash1910: Yes I will update this, I initially thought this to be `Sequence[bytes |  str]` .

cuda_bindings/cuda/bindings/utils/_nvvm_utils.py
abhilash1910: Addressed this part.

cuda_bindings/cuda/bindings/utils/_nvvm_utils.py
abhilash1910: While I do agree `verify_program()` is not needed, the location of  proposed `try` could eventually throw as then the `destroy_program()` call will never get triggered. 

What I was thinking is that :

```python
program = nvvm.create_program()
try:
  <..rest of code>
  nvvm.compile_program()
except nvvm.nvvmError as e:
 <exception raising>
finally:
  destroy_program()
```

abhilash1910:
pre-commit.ci autofix

cuda_bindings/cuda/bindings/utils/__init__.py
```diff
@@ -1,7 +1,8 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
rwgk, line 1: nit: 2025-2026

cuda_bindings/cuda/bindings/utils/_nvvm_utils.py
rwgk: Yes, we should definitely keep the existing `finally` code.

cuda_bindings/cuda/bindings/utils/_nvvm_utils.py
rwgk: To track the offline suggestion here: The second example doesn't add anything, it's very similar to the first one. I'd delete the second one, to avoid a random reference to numba (which could give the wrong impression that cuda-bindings is somehow closely related to numba).

abhilash1910:
pre-commit.ci autofix

cuda_bindings/cuda/bindings/utils/_nvvm_utils.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+        precheck_ir_bytes = precheck_ir.encode("utf-8")
+        nvvm.add_module_to_program(
+            program,
+            precheck_ir_bytes,
+            len(precheck_ir_bytes),
+            "precheck.ll",
+        )
+
+        nvvm.compile_program(program, len(options), options)
```
rwgk: This looks great, the only change I'd still make is to have a narrower, nested `try` (untested):

```python
        try:
            nvvm.compile_program(program, len(options), options)
        except nvvm.nvvmError as e:
            if e.status == nvvm.Result.ERROR_INVALID_OPTION:
                return False
            raise
```

I.e. almost what you have already, but indented by one level and a `try` on top.

Then you have the outer, broad `try` for the `finally`, and the inner, narrow `try` for the exact exception you want to detect.

cuda_bindings/cuda/bindings/utils/_nvvm_utils.py
```diff
@@ -0,0 +1,89 @@
...
+"""
+
+
+def check_nvvm_compiler_options(options: Sequence[str]) -> bool:
+    """
+    Abstracted from https://github.com/NVIDIA/numba-cuda/pull/681
+
+    Check if the specified options are supported by the current libNVVM version.
+
+    The options are a list of bytes, each representing a compiler option.
```
rwgk: Ah, I was to fast triggering the CI. I just canceled the run, after seeing that the docstring here is out-of-date:

Here: list of str (not bytes)

Further below: remove `b` from `b"..."` literals in Parameters and Examples sections.

abhilash1910: Thanks @rwgk

rwgk (approved):

rwgk:
This doesn't work with the cuda-bindings from the 12.9.x backport branch, e.g.:

https://github.com/NVIDIA/cuda-python/actions/runs/24091305075/job/70279241852?pr=1837

```
tests/test_program.py:76: in _check_nvvm_arch
    from cuda.bindings.utils import check_nvvm_compiler_options
E   ImportError: cannot import name 'check_nvvm_compiler_options' from 'cuda.bindings.utils' (/opt/hostedtoolcache/Python/3.10.20/x64/lib/python3.10/site-packages/cuda/bindings/utils/__init__.py)
```

@leofang wrote under #1457:

> We should expose it to cuda.bindings.utils, similar to the PTX version helpers.

That makes using it from `cuda_core/tests/` a little tricky: it'll only be in future cuda-bindings releases.

@leofang how would you handle this situation?

rwgk:
@abhilash1910 — Offline we decided (roughly): let's move the (or some) tests to `cuda_bindings/`, and if we want to keep certain tests in `cuda_core/`, skip them if `cuda.bindings.utils.check_nvvm_compiler_options` is not available. Could you please try how that shakes out? Please ping me on chat if you have questions.

abhilash1910:
pre-commit.ci autofix

cuda_bindings/cuda/bindings/utils/_nvvm_utils.py
```diff
@@ -0,0 +1,89 @@
...
+    If the test program fails to compile, the options are not supported and False
+    is returned.
+
+    If the test program compiles successfully, True is returned.
+
+    cuda.bindings.nvvm returns exceptions instead of return codes.
+
+    Parameters
+    ----------
+    options : Sequence[bytes]
```
rwgk: This is still `Sequence[bytes]` (should be `str`)

cuda_bindings/tests/test_utils.py
```diff
@@ -11,10 +11,28 @@
...
+    check_nvvm_compiler_options,
+    get_cuda_native_handle,
+    get_minimal_required_cuda_ver_from_ptx_ver,
+    get_ptx_ver,
+)
 
 have_cufile = importlib.util.find_spec("cuda.bindings.cufile") is not None
 
+
+def _is_libnvvm_available() -> bool:
```
rwgk, line 24: The `except Exception` here is overly broad and could mask bugs (all kinds of runtime errors).

Since `cuda.bindings._internal.nvvm` is internal to cuda_bindings, the import should be unconditional — if it fails, that's a bug we want to surface.

The only expected exception is `DynamicLibNotFoundError` (or subclasses) when libNVVM isn't installed.

Suggested change:

```python
def _is_libnvvm_available() -> bool:
    from cuda.bindings._internal.nvvm import _inspect_function_pointer
    from cuda.pathfinder import DynamicLibNotFoundError

    try:
        return _inspect_function_pointer("__nvvmCreateProgram") != 0
    except DynamicLibNotFoundError:
        return False
```

cuda_core/tests/test_program.py
```diff
@@ -70,81 +70,26 @@ def _has_nvrtc_pch_apis_for_tests():
...
-
-!nvvmir.version = !{{!1}}
-!1 = !{{i32 {major}, i32 {minor}, i32 {debug_major}, i32 {debug_minor}}}
-"""
-
-
-def _get_libnvvm_version_for_tests():
-    """
-    Detect libNVVM version by compiling dummy IR and analyzing the PTX output.
+def _has_check_nvvm_compiler_options():
```
rwgk, line 73: This could be made slightly more explicit and tighter by separating the module import from the attribute check:

```python
def _has_check_nvvm_compiler_options():
    try:
        import cuda.bindings.utils as utils
    except ModuleNotFoundError:
        return False
    return hasattr(utils, "check_nvvm_compiler_options")
```

This makes the intent clearer: first check if the module exists, then check if the function is available. The `ModuleNotFoundError` guard handles very old cuda-bindings, while `hasattr` handles bindings that have the module but predate this new function.

cuda_bindings/cuda/bindings/utils/_nvvm_utils.py
```diff
@@ -0,0 +1,89 @@
...
+        List of compiler options as strings (e.g., ["-arch=compute_90", "-g"]).
+
+    Returns
+    -------
+    bool
+        True if the options are supported, False otherwise.
+
+    Examples
+    --------
+    >>> from cuda.bindings.utils import check_nvvm_options
```
rwgk: I overlooked before: this should be `check_nvvm_compiler_options` (import and call).

rwgk (commented):
A few pretty minor suggestions (docstring fixes and test code only).

I verified that all existing test coverage is preserved.

cuda_core/tests/test_program.py
abhilash1910: Agree and addressed, thanks

abhilash1910:
pre-commit.ci autofix

rwgk (approved):

Issue #1330, opened 2025-12-06
CUDA graph phase N - graph updates


leofang:
This task should cover updating both whole graphs and individual graph nodes.

Andy-Jost:
**Implementation plan for graph updates**

Breaking this into five PRs:

1. **Whole-graph update tests** — Add tests for the existing `Graph.update()` API (`cuGraphExecUpdate`), which replaces an entire executable graph from a new graph definition. Refactoring: test file reorganization and breaking down of `_graphdef.pyx` into a submodule.

2. **Edge mutation on GraphDef** — Support adding and removing edges between nodes in a graph definition, wrapping `cuGraphAddDependencies` / `cuGraphRemoveDependencies`.

3. **GraphBuilder interop with GraphDef** — Add a method `GraphBuilder.graph_def` to get the explicit representation of a graph built by stream capture. This involves changing `GraphBuilder` to use an RAII graph handle internally so that ownership can be shared.

4. **Node property setters (definition-level)** — Add property setters to all `GraphNode` subclasses (`KernelNode`, `MemsetNode`, `MemcpyNode`, etc.) that call the corresponding `cuGraph*NodeSetParams` APIs to modify nodes in a graph definition.

5. **Exec-level node updates** — Enable modifying nodes in an instantiated graph using `graph[node].property = new_value`, so users can update kernel parameters, memset values, etc. without rebuilding the graph.

leofang:
Q: Is this P0 or P1?

leofang:
Assuming this is about either a new feature or a fix that is backward-compatible, tag this as P1.

Andy-Jost:
The remaining work involves feature additions only. Nice-to-have for v1.0.0.

Andy-Jost:
Wait, @rparolin is this a P0 for v1.0.0? Items 1-2 are merged, and I have a PR posted for item 3 (#2008), but I don't see 4-5 making it into the release.

PR #1850, opened 2026-04-02
Add edge mutation and MutableSet interface for graph nodes

Andy-Jost:
## Summary

Implements step 2 of #1330 (graph updates): edge mutation on `GraphDef`. Graph node adjacencies (`pred`/`succ`) are now exposed as `AdjacencySetProxy`, a `collections.abc.MutableSet` proxy backed by `cuGraphAddDependencies`/`cuGraphRemoveDependencies`. Nodes can be removed via `GraphNode.destroy()` (wraps `cuGraphDestroyNode`), and property setters on `pred`/`succ` support bulk edge replacement.

## Examples

```python
# Inspect adjacencies
b in node.pred     # predecessors
b in node.succ     # successors

# Mutate edges
node.succ.add(b)           # add a single edge
node.succ |= {b, c}        # bulk add
node.succ.update({b, c})   # same thing
node.pred.discard(a)       # remove one predecessor
node.pred -= {a, b}        # bulk remove
node.succ = {x, y}         # replace all successors
node.pred.clear()          # remove all predecessors

# Remove a node and all its edges from a graph
node.destroy()
```

## Changes

- **`AdjacencySetProxy`**: full `MutableSet` implementation with bulk-efficient `clear()`, `__isub__()`, `__ior__()`, `update()`, and `remove_edges()`. Includes `_from_iterable` override for binary ops and duplicate-add guard to avoid `CUDA_ERROR_INVALID_VALUE`.
- **`GraphNode.destroy()`**: removes a node and its edges from the graph definition via `cuGraphDestroyNode`, then zeroes the underlying handle to prevent stale memory access. Safe to call on already-destroyed nodes. Properties (`type`, `pred`, `succ`, `handle`) degrade gracefully to `None` or empty sets.
- **`GraphNode.is_valid`**: returns `False` after `destroy()` has been called
- **`GraphNode.pred`/`succ` setters**: clear existing edges and add new ones in bulk
- **`GraphDef.nodes()`/`edges()`**: return `set` instead of `tuple` for consistency
- **Test helpers**: `compile_parallel_kernels()` in `graph_kernels.py`, `assert_mutable_set_interface()` in `collection_interface_testers.py`

## Test Coverage

- `TestMutateYRig` class: baseline, node destroy (single/multiple), edge insertion, join-node removal with rewiring
- `test_destroyed_node`: verifies handle invalidation, graceful property degradation, edge-add failure on destroyed nodes, and double-destroy safety
- `test_adjacency_set_interface`: exercises every `MutableSet` method against a reference `set`
- `test_adjacency_set_pred_direction`: verifies bidirectional edge visibility
- `test_adjacency_set_property_setter`: verifies bulk edge replacement via setters
- `test_add_wrong_type`, `test_cross_graph_edge`, `test_self_edge`: error-path coverage
- `test_convert_linear_to_fan_in`: rewires a sequential graph into parallel fan-in topology

## Related Work

- Part of #1330 (step 2 of 4)
- Builds on the `_graph_def/` subpackage refactor from `graph-updates` branch

### DIFF
--- a/cuda_core/cuda/core/_cpp/resource_handles.hpp
+++ b/cuda_core/cuda/core/_cpp/resource_handles.hpp
@@ -415,6 +415,9 @@ GraphNodeHandle create_graph_node_handle(CUgraphNode node, const GraphHandle& h_
 // Extract the owning graph handle from a node handle.
 GraphHandle graph_node_get_graph(const GraphNodeHandle& h) noexcept;
 
+// Zero the CUgraphNode resource inside the handle, marking it invalid.
+void invalidate_graph_node_handle(const GraphNodeHandle& h) noexcept;
+
 // ============================================================================
 // Graphics resource handle functions
 // ============================================================================
--- a/cuda_core/tests/graph/test_graphdef_mutation.py
+++ b/cuda_core/tests/graph/test_graphdef_mutation.py
@@ -9,7 +9,6 @@ from helpers.collection_interface_testers import assert_mutable_set_interface
 from helpers.graph_kernels import compile_parallel_kernels
 from helpers.marks import requires_module
 
-from cuda.bindings import driver
 from cuda.core import Device, LaunchConfig, LegacyPinnedMemoryResource
 from cuda.core._graph._graph_def import GraphDef, KernelNode, MemsetNode
 from cuda.core._utils.cuda_utils import CUDAError
@@ -153,11 +152,11 @@ class TestMutateYRig:
         assert rig.nodes == rig.initial_nodes
         rig.close()
 
-    def test_discard_a1(self, init_cuda):
-        """Discard a1 (creates a race on arm a). Arm b yields the expected
+    def test_destroy_a1(self, init_cuda):
+        """Destroy a1 (creates a race on arm a). Arm b yields the expected
         value, and the final step is correctly ordered after b completes."""
         rig = YRig()
-        rig.a[1].discard()
+        rig.a[1].destroy()
         rig.run()
         _, b_exp, _ = rig.expected_output
         assert rig.B_out == b_exp
@@ -167,10 +166,10 @@ class TestMutateYRig:
         assert rig.nodes == rig.initial_nodes - {a1}
         rig.close()
 
-    def test_discard_a2(self, init_cuda):
-        """Discard a2, connect a1--r"""
+    def test_destroy_a2(self, init_cuda):
+        """Destroy a2, connect a1--r"""
         rig = YRig()
-        rig.a[2].discard()
+        rig.a[2].destroy()
         rig.a[1].succ.add(rig.r)
         rig.A_OPS.pop()
         rig.run()
@@ -180,11 +179,11 @@ class TestMutateYRig:
         assert rig.nodes == rig.initial_nodes - {a2}
         rig.close()
 
-    def test_discard_joint(self, init_cuda):
+    def test_destroy_joint(self, init_cuda):
         """Remove the joining node and instead add edges directly to r."""
         rig = YRig()
         _, _, a2, _, b1, j, r = rig.a + rig.b + [rig.j, rig.r]
-        j.discard()
+        j.destroy()
         r.pred = {a2, b1}
         rig.run()
         assert rig.output == rig.expected_output
@@ -266,8 +265,8 @@ def test_adjacency_set_property_setter(init_cuda):
     assert hub.pred == set()
 
 
-def test_discarded_node(init_cuda):
-    """Test uses of discarded nodes."""
+def test_destroyed_node(init_cuda):
+    """Test that destroy() invalidates a node."""
     mr = LegacyPinnedMemoryResource()
     buf = mr.allocate(4)
     arr = np.from_dlpack(buf).view(np.int32)
@@ -278,25 +277,34 @@ def test_discarded_node(init_cuda):
     a = g.memset(ptr, 0, 4)
     b = a.memset(ptr, 42, 4)
 
+    assert a.is_valid
+    assert b.is_valid
     assert b in g.nodes()
     assert (a, b) in g.edges()
 
-    b.discard()
+    b.destroy()
 
-    # b is removed from the graph but still usable
+    assert not b.is_valid
     assert b not in g.nodes()
     assert (a, b) not in g.edges()
+
+    # Python object is invalid but using it does not crash.
     assert isinstance(b, MemsetNode)
-    assert b.type == driver.CUgraphNodeType.CU_GRAPH_NODE_TYPE_KERNEL
+    assert b.type is None
     assert b.pred == set()
     assert b.succ == set()
-    assert b.handle != 0
-    assert b.dptr == ptr
-    assert b.value == 42
-    assert b.width == 4
+    assert b.handle is None
+    assert b.dptr == ptr  # tolerable
+    assert b.value == 42  # tolerable
+    assert b.width == 4  # tolerable
 
-    # Repeated discard succeeds quietly.
-    b.discard()
+    # Adding an edge to a destroyed node fails.
+    with pytest.raises(CUDAError):
+        a.succ.add(b)
+
+    # Repeated destroy succeeds quietly.
+    b.destroy()
+    assert not b.is_valid
 
 
 def test_add_wrong_type(init_cuda):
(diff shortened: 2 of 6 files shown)


cuda_core/cuda/core/_cpp/resource_handles.hpp
```diff
@@ -415,6 +415,9 @@ GraphNodeHandle create_graph_node_handle(CUgraphNode node, const GraphHandle& h_
 // Extract the owning graph handle from a node handle.
 GraphHandle graph_node_get_graph(const GraphNodeHandle& h) noexcept;
 
+// Zero the CUgraphNode resource inside the handle, marking it invalid.
+void invalidate_graph_node_handle(const GraphNodeHandle& h) noexcept;
```
rparolin, line 419: Could we make the `invalidate_graph_node_handle` take a mutable reference? Rather than internally mutating through a const& handle?

Andy-Jost: I don't think it will improve anything. `GraphNodeHandle` is a `std::shared_ptr`, and this "invalidate" function relies on `T& std::shared_ptr::operator*() const noexcept` to get a non-`const` reference to the target object (note the function is marked `const` but the return type is a non-`const` `T&`; [doc](https://en.cppreference.com/w/cpp/memory/shared_ptr/operator%252A.html)). A`const` qualifier on `GraphNodeHandle` only refers to whether the pointer can be reset or rebound to a new target through something like assignment/move, not whether the target itself is mutable. The "invalidate" function marks the target as having been invalidated by a call to `cuGraphNodeDestroy` but does not change the target location, so a `const` handle is correct.

Removing `const` from this function would not have the effect you might expect. Consider these three handles to two targets:

```
GraphNodeHandle h1 = ...;      // a graph node handle
const GraphNodeHandle h2 = h1; // a second handle to the same graph node
GraphNodeHandle h3 = ...;      // a handle to an unrelated graph node
```

Marking `h2` with `const` protects the handle, not the `GraphNode` it points to. In this case, `h2 = h3` would be disallowed. If we changed the declaration as you suggest, `invalidate_graph_node_handle(h2)` would be disallowed; however, `invalidate_graph_node_handle(h1)` would still be allowed and would invalidate the shared graph node.

In general, handle targets are always marked `const` because resource handles by design only permit attach, detach, and dereference operations. The `GraphNodeHandle` specifically and oddly needs invalidation because of the strange (I'd say broken) semantics of `cuGraphNodeDestroy`.

One way to support invalidation would be removing `const` from this declaration:

```using GraphNodeHandle = std::shared_ptr<const CUgraphNode>```

Then one could write `*h2 = NULL` to invalidate. That's not a good solution because it allows mutation everywhere.

The better approach (what's in this PR) is to keep the current definition, consistent with other handle types, while marking `CUgraphNode` as `mutable` in `GraphNodeBox` and providing an "invalidate" function to invalidate the handle when needed. This keeps the `GraphNodeHandle` similar to all other handles, while supporting this odd destroy behavior that is needed because of `cuGraphNodeDestroy`.

File: cuda_core/cuda/core/_graph/_graph_def/_adjacency_set_proxy.pyx [hunk 2]
```diff
@@ -0,0 +1,246 @@
...
+    @classmethod
+    def _from_iterable(cls, it):
+        return set(it)
+
+    # --- abstract methods required by MutableSet ---
+
+    def __contains__(self, x):
+        if not isinstance(x, GraphNode):
+            return False
+        return x in (<_AdjacencySetCore>self._core).query()
```
mdboom, line 42: This may be premature optimization, but is it required to instantiate an entire list that is a copy of the elements on the C++ side just to test for `in`?  We'd have to check that this works in Cython, but having `query` return an iterator comprehension rather than a list comprehension may be enough.

cuda_core/cuda/core/_cpp/resource_handles.hpp
Andy-Jost: This function should probably be called `invalidate_graph_node`

cuda_core/cuda/core/_graph/_graph_def/_adjacency_set_proxy.pyx
mdboom: Maybe this?

```diff
--- a/cuda_core/cuda/core/_graph/_graph_def/_adjacency_set_proxy.pyx
+++ b/cuda_core/cuda/core/_graph/_graph_def/_adjacency_set_proxy.pyx
@@ -68,7 +68,7 @@ class AdjacencySetProxy(MutableSet):
         """Remove all edges in a single driver call."""
         members = (<_AdjacencySetCore>self._core).query()
         if members:
-            (<_AdjacencySetCore>self._core).remove_edges(members)
+            (<_AdjacencySetCore>self._core).remove_edges(list(members))

     def __isub__(self, it):
         """Remove edges to all nodes in *it* in a single driver call."""
@@ -139,7 +139,7 @@ cdef class _AdjacencySetCore:
             c_from[0] = as_cu(other._h_node)
             c_to[0] = as_cu(self._h_node)

-    cdef list query(self):
+    cdef query(self):
         cdef cydriver.CUgraphNode c_node = as_cu(self._h_node)
         if c_node == NULL:
             return []
@@ -153,8 +153,8 @@ cdef class _AdjacencySetCore:
         with nogil:
             HANDLE_RETURN(self._query_fn(
                 c_node, nodes_vec.data(), &count))
-        return [GraphNode._create(self._h_graph, nodes_vec[i])
-                for i in range(count)]
+        return (GraphNode._create(self._h_graph, nodes_vec[i])
+                for i in range(count))

     cdef Py_ssize_t count(self):
         cdef cydriver.CUgraphNode c_node = as_cu(self._h_node)
```

cuda_core/cuda/core/_graph/_graph_def/_adjacency_set_proxy.pyx
```diff
@@ -0,0 +1,246 @@
...
+        """Add edges to multiple nodes at once."""
+        nodes = []
+        for other in others:
+            if isinstance(other, GraphNode):
+                nodes.append(other)
+            else:
+                nodes.extend(other)
+        if not nodes:
+            return
+        for n in nodes:
```
mdboom, line 93: Could move this check next to the `nodes.extend(other)` above, since we already know it to be true if `isinstance(other, GraphNode)` above.

cuda_core/tests/graph/test_graphdef_mutation.py
```diff
@@ -0,0 +1,394 @@
...
+        assert rig.nodes == rig.initial_nodes - {j}
+        rig.close()
+
+    def test_insert_b(self, init_cuda):
+        """Insert a node into arm b."""
+        rig = YRig()
+        coeffs = 5, 3
+        b_new = rig.graph_def.launch(rig.config, rig.affine, rig.ptr_b, *coeffs)
+        b0, b1 = rig.b
+        b0.succ.discard(b1)
```
mdboom, line 200: `discard` is defined to silently do nothing if the thing being removed doesn't exist.  And it definitely does this by experimentation, but it's puzzling to me /why/ it does, since the underlying call to `cuGraphRemoveDependencies` says:

"Specifying a non-existing dependency will return an error."

I'm probably just missing some step in the logic here that is (correctly) preventing the exception from bubbling up.  But in the event that this is the API not doing what it says it would (and therefore might do so in the future), it's probably worth doing this twice to test that removing an already-moved element doesn't fail.

mdboom (approved):

cuda_core/cuda/core/_graph/_graph_def/_adjacency_set_proxy.pyx
Andy-Jost: I was thinking about this, too. Definitely a good idea to see whether query can return a generator. I was thinking of seeing whether I can implement contains() directly on `AdjacencySetCore`. We have to query the driver in batch, but we don't need to reconstruct the Python objects at all just to check containment.

cuda_core/tests/graph/test_graphdef_mutation.py
Andy-Jost: I had to put this containment check in `discard` for exactly this reason.

```python
    def discard(self, value):
        if not isinstance(value, GraphNode):
            return
        if value not in self:  # <<<
            return
```

Also, the situation you mention is covered in ` assert_mutable_set_interface`:

```python
    # discard non-member is a no-op
    subject.discard(d)
    assert subject == ref
```

mdboom: Awesome.  That's exactly the thing I missed.

PR #1853, opened 2026-04-02
Add GraphNode identity cache for stable object round-trips

Andy-Jost:
## Summary

Ensures that `GraphNode` objects round-tripped through `GraphDef.nodes()`, `edges()`, or `pred`/`succ` traversal return the **same Python object** (preserving `is` identity). This avoids surprising behavior and enables idiomatic Python.

Depends on #1850.

## Example

After this change, the following will pass:

```python
# For nodes a, b belonging to graph g
assert any(x is a for x in g.nodes())

a.succ = {b}    # set successors of a to {b}
b2, = a.succ    # get successors of a
assert b2 is b  # the set contains b
```

## Changes

- **C++ HandleRegistry for CUgraphNode** (`resource_handles.cpp`): Added a `graph_node_registry` (matching the existing `kernel_registry` pattern) that deduplicates `GraphNodeHandle`s so the same `CUgraphNode` always maps to the same `shared_ptr`.
- **Cython WeakValueDictionary** (`_graph_node.pyx`): Added a module-level `_node_registry` and a `_registered()` helper. All node creation paths — `GN_create` (for driver round-trips) and every builder function (`GN_launch`, `GN_join`, `GN_alloc`, etc.) — register their results.
- **Fix `HandleRegistry::unregister_handle`** (`resource_handles.cpp`): Removed the `expired()` guard that prevented registry cleanup when the `shared_ptr` was still alive. After `destroy()`, the Python object still holds the handle, so `expired()` returned `false` and the stale entry remained. If the driver reused the `CUgraphNode` pointer value for a new node, the registry returned the invalidated handle (`handle=0x0`), causing `CUDA_ERROR_INVALID_VALUE`. This was confirmed by CI failures on CUDA 12.9.1 runners where pointer reuse was more aggressive.
- **Rename `invalidate_graph_node_handle` → `invalidate_graph_node`**: For clarity.
- **Identity assertions** (`test_graphdef.py`): `test_node_type_preserved_by_nodes` and `test_node_type_preserved_by_pred_succ` now assert `is` identity, not just equality.
- **`is not` in test** (`test_graphdef_mutation.py`): `test_convert_linear_to_fan_in` now uses `node is not reduce_node` instead of `node != reduce_node`.

## Test Coverage

- `test_node_type_preserved_by_nodes` — parameterized over all node types, checks `is` identity through `nodes()`.
- `test_node_type_preserved_by_pred_succ` — checks `is` identity through `pred`/`succ` traversal.
- `test_convert_linear_to_fan_in` — exercises `is not` in a real graph rewiring scenario.

## Related Work

- Part of #1330 (CUDA graph updates)
- Depends on #1850 (edge mutation)

### DIFF
--- a/cuda_core/cuda/core/_cpp/resource_handles.cpp
+++ b/cuda_core/cuda/core/_cpp/resource_handles.cpp
@@ -174,13 +174,8 @@ public:
     }
 
     void unregister_handle(const Key& key) noexcept {
-        try {
-            std::lock_guard<std::mutex> lock(mutex_);
-            auto it = map_.find(key);
-            if (it != map_.end() && it->second.expired()) {
-                map_.erase(it);
-            }
-        } catch (...) {}
+        std::lock_guard<std::mutex> lock(mutex_);
+        map_.erase(key);
     }
 
     Handle lookup(const Key& key) {
@@ -989,7 +984,7 @@ GraphHandle graph_node_get_graph(const GraphNodeHandle& h) noexcept {
     return h ? get_box(h)->h_graph : GraphHandle{};
 }
 
-void invalidate_graph_node_handle(const GraphNodeHandle& h) noexcept {
+void invalidate_graph_node(const GraphNodeHandle& h) noexcept {
     if (h) {
         CUgraphNode node = get_box(h)->resource;
         if (node) {
--- a/cuda_core/cuda/core/_cpp/resource_handles.hpp
+++ b/cuda_core/cuda/core/_cpp/resource_handles.hpp
@@ -416,7 +416,7 @@ GraphNodeHandle create_graph_node_handle(CUgraphNode node, const GraphHandle& h_
 GraphHandle graph_node_get_graph(const GraphNodeHandle& h) noexcept;
 
 // Zero the CUgraphNode resource inside the handle, marking it invalid.
-void invalidate_graph_node_handle(const GraphNodeHandle& h) noexcept;
+void invalidate_graph_node(const GraphNodeHandle& h) noexcept;
 
 // ============================================================================
 // Graphics resource handle functions
--- a/cuda_core/cuda/core/_graph/_graph_def/_graph_node.pyx
+++ b/cuda_core/cuda/core/_graph/_graph_def/_graph_node.pyx
@@ -48,7 +48,7 @@ from cuda.core._resource_handles cimport (
     create_graph_handle_ref,
     create_graph_node_handle,
     graph_node_get_graph,
-    invalidate_graph_node_handle,
+    invalidate_graph_node,
 )
 from cuda.core._utils.cuda_utils cimport HANDLE_RETURN, _parse_fill_value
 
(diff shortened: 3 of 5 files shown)


cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -174,13 +174,8 @@ class HandleRegistry {
...
     void unregister_handle(const Key& key) noexcept {
-        try {
-            std::lock_guard<std::mutex> lock(mutex_);
-            auto it = map_.find(key);
-            if (it != map_.end() && it->second.expired()) {
-                map_.erase(it);
-            }
-        } catch (...) {}
+        std::lock_guard<std::mutex> lock(mutex_);
+        map_.erase(key);
```
Andy-Jost, line 178: The `expired()` check was always unnecessary. For `GraphNodes` it is harmful. By not purging invalidated `GraphNode` handles, the driver reusing an old pointer could cause a false hit.

cuda_core/cuda/core/_cpp/resource_handles.hpp
```diff
@@ -416,7 +416,7 @@ GraphNodeHandle create_graph_node_handle(CUgraphNode node, const GraphHandle& h_
 GraphHandle graph_node_get_graph(const GraphNodeHandle& h) noexcept;
 
 // Zero the CUgraphNode resource inside the handle, marking it invalid.
-void invalidate_graph_node_handle(const GraphNodeHandle& h) noexcept;
+void invalidate_graph_node(const GraphNodeHandle& h) noexcept;
```
Andy-Jost, line 419: Following on [this discussion](https://github.com/NVIDIA/cuda-python/pull/1850#discussion_r3034089808), this is a better name.

cuda_core/tests/graph/test_graphdef.py
```diff
@@ -697,6 +699,31 @@ def test_node_attrs_preserved_by_nodes(node_spec):
...
+    assert b2 is b
+
+    (a2,) = b.pred
+    assert a2 is a
+
+    # edges()
+    ((a2, b2),) = g.edges()
+    assert a2 is a
+    assert b2 is b
+
```
mdboom, line 725: Since a potential failure mode here might be unintentional leaking of references, maybe it's also worth testing something like:

```
del a, a2, b, b2
```

and then asserting that the registry is empty.

cuda_core/cuda/core/_graph/_graph_def/_graph_node.pyx
```diff
@@ -57,10 +57,19 @@ from cuda.core._graph._utils cimport (
     _attach_user_object,
 )
 
+import weakref
+
 from cuda.core import Device
 from cuda.core._graph._graph_def._adjacency_set_proxy import AdjacencySetProxy
 from cuda.core._utils.cuda_utils import driver, handle_return
 
+_node_registry = weakref.WeakValueDictionary()
```
mdboom, line 66: I think you explained this live in one of the PR review meetings, but since it's subtle enough, it might be worth documenting why there is a registry on both the C++ and Python sides and how they interact (if at all).

PR #1854, opened 2026-04-03
[doc-only] docs(pathfinder): prepare 1.5.1 release notes

rwgk:
Add cuda-pathfinder 1.5.1 release notes and register 1.5.1 in nv-versions so the published docs include the new version entry.

Made-with: Cursor


rwgk:
Thanks @mdboom!

PR #1859, opened 2026-04-03
Graph node follow-ups: repr, containment, empty(), registry docs

Andy-Jost:
## Summary

Follow-on changes addressing review feedback from #1850 and #1853.

## Changes

- **`__repr__` improvements** (`_subclasses.pyx`): All `GraphNode` subclass `__repr__` methods now start with `handle=0x...` followed by static, type-specific identity data, enhancing their usefulness for debugging.
- **Cheap containment test** (`_adjacency_set_proxy.pyx`): Added `_AdjacencySetCore.contains()` that checks membership by comparing raw `CUgraphNode` handles at the C level, avoiding Python object construction. Uses a 16-element stack buffer for a single driver call in the common case.
- **Stack-buffer optimization in `query()`** (`_adjacency_set_proxy.pyx`): Same 16-element stack buffer approach applied to `query()`, eliminating the count-then-fetch two-call pattern for small adjacency sets.
- **Single-call optimization in `nodes()` and `edges()`** (`_graph_def.pyx`): Pre-allocate vectors to 128 entries and pass them on the first driver call. Only fall back to a second call if the graph exceeds 128 nodes/edges.
- **Early type check in `update()`** (`_adjacency_set_proxy.pyx`): Type validation moved inline to fail fast on invalid input, per review feedback.
- **`GraphDef.empty()`** (`_graph_def.pyx`): New method for creating entry-point empty nodes without the awkward `g.join()` spelling.
- **Registry cleanup test** (`test_graphdef.py`): `test_registry_cleanup` exercises `destroy()`, graph deletion, and weak-reference cleanup of the node registry.
- **Registry design documentation** (`_cpp/REGISTRY_DESIGN.md`): Documents the two-level identity map (C++ `HandleRegistry` and Cython `_node_registry`) with cross-references at each instantiation site.

## Test Coverage

- Existing graph tests validate `empty()` (replacing prior `g.join()` calls).
- `test_registry_cleanup` covers explicit destroy, graph teardown, and GC-driven cleanup.
- `__repr__` patterns validated by `test_object_protocols.py`.

### DIFF
--- a/cuda_core/cuda/core/_graph/_graph_def/_graph_def.pyx
+++ b/cuda_core/cuda/core/_graph/_graph_def/_graph_def.pyx
@@ -332,18 +332,20 @@ cdef class GraphDef:
         set of GraphNode
             All nodes in the graph.
         """
-        cdef size_t num_nodes = 0
+        cdef vector[cydriver.CUgraphNode] nodes_vec
+        nodes_vec.resize(128)
+        cdef size_t num_nodes = 128
 
         with nogil:
-            HANDLE_RETURN(cydriver.cuGraphGetNodes(as_cu(self._h_graph), NULL, &num_nodes))
+            HANDLE_RETURN(cydriver.cuGraphGetNodes(as_cu(self._h_graph), nodes_vec.data(), &num_nodes))
 
         if num_nodes == 0:
             return set()
 
-        cdef vector[cydriver.CUgraphNode] nodes_vec
-        nodes_vec.resize(num_nodes)
-        with nogil:
-            HANDLE_RETURN(cydriver.cuGraphGetNodes(as_cu(self._h_graph), nodes_vec.data(), &num_nodes))
+        if num_nodes > 128:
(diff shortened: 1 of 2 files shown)


cuda_core/cuda/core/_graph/_graph_def/_adjacency_set_proxy.pyx
```diff
@@ -156,6 +159,35 @@ cdef class _AdjacencySetCore:
         return [GraphNode._create(self._h_graph, nodes_vec[i])
                 for i in range(count)]
```
Andy-Jost, line 160: I wasn't able to return a generator as suggested [here](https://github.com/NVIDIA/cuda-python/pull/1850#discussion_r3034126702) because `nodes_vec` is a stack object and it would lead to a use-after-free unless we define a cdef class for the iterator. I added an optimized `contains` method that avoids reconstructing Python objects for containment checks.

PR #1805, opened 2026-03-23
Move enum explanations and health checks from cuda_core to cuda_bindings

rwgk:
~Closes 1712~ — **This PR closed in favor of PR #1860**

The `DRIVER_CU_RESULT_EXPLANATIONS` and `RUNTIME_CUDA_ERROR_EXPLANATIONS` dicts are fundamentally tied to the cuda-bindings release (they must match the enums shipped in that release). Having them live exclusively in cuda_core meant the health-check tests failed whenever cuda_core was tested against a different version of cuda-bindings (nvbug 5932944).

### Changes

- **Move the dicts to `cuda_bindings/cuda/bindings/_utils/`** as the single authoritative source (renamed to `_EXPLANATIONS` with a `_CTK_MAJOR_MINOR_PATCH` version tag).
- **Delete the copies from cuda_core.** `cuda_utils.pyx` now imports directly from `cuda.bindings._utils`, with a `ModuleNotFoundError` fallback to an empty dict.
- **Move the exhaustive health-check tests** to `cuda_bindings/tests/test_enum_explanations.py`, where they belong alongside the dicts they verify.

### Impact on error messages for cuda-core users

When cuda-core raises a `CUDAError`, it tries to include a human-readable explanation of the error code (e.g. "This indicates that one or more of the parameters passed to the API call is not within an acceptable range of values").

With this change:

- **cuda-bindings >= this PR:** Error messages continue to include explanations, exactly as before.
- **cuda-bindings < this PR** (older releases that don't ship `_utils`): Error messages fall back to the driver/runtime error name and description string obtained from `cuGetErrorString` / `cudaGetErrorString`. The explanations are a nice-to-have supplement, and the error name + description are still informative. Upgrading to a current cuda-bindings release restores the full explanations.

PR #1860, opened 2026-04-05
cuda_core: derive error enum explanations from bindings docstrings

### LINKED ISSUE #1712: [Bug]: Move enum explanations and health checks from cuda_core to cuda_bindings
Related: nvbug 5932944

**Situation**

Currently (`main` @ eabaf4bf3333a32ac9eb9d183ce98ab35cc79eb5) we have these enum explanations and related health checks :

* https://github.com/NVIDIA/cuda-python/blob/main/cuda_core/cuda/core/_utils/driver_cu_result_explanations.py

* https://github.com/NVIDIA/cuda-python/blob/main/cuda_core/cuda/core/_utils/runtime_cuda_error_explanations.py

* https://github.com/NVIDIA/cuda-python/blob/eabaf4bf3333a32ac9eb9d183ce98ab35cc79eb5/cuda_core/tests/test_cuda_utils.py#L11-L40

**Problem**

The tests succeed only for the matching cuda-bindings release.

If cuda_core is tests are run against an older or newer version of cuda-bindings, they will fail (that's nvbug 5932944).

**Solution**

The explanations dicts are fundamentally directly tied to the bindings, therefore they should be moved to cuda_bindings (maybe a new `cuda_bindings/cuda/bindings/_utils/` directory). Accordingly, the tests should be moved, too.

**Backward compatibility**

When the explanations dicts are moved, the cuda_core code using them needs to be made tolerant to not finding them in existing releases of cuda-bindings. I.e. the explanations will be used if available in cuda_bindings, or otherwise simply not be added when generating error messages at runtime. This is considered to be a minor loss that will become a non-issue over time as old cuda-bindings releases fall out of use.

rwgk:
Closes #1712

This PR grows out of the validation work under #1805. That work showed that driver/runtime error enums expose usable per-member `__doc__` text once a small amount of normalization is applied. The enum docstrings are usable starting with `cuda-bindings >= 13.2.0` on the main branch, and `cuda-bindings 12.9.6` on the 12.9.x backport branch. See the [comment below](https://github.com/NVIDIA/cuda-python/pull/1860#issuecomment-4189310640) for a summary of a manual review of the cleaned enum docstrings.

## Summary
- use cleaned enum `__doc__` strings for driver/runtime error explanations in `cuda_core` when `12.9.6 <= cuda-bindings < 13.0.0` or `cuda-bindings >= 13.2.0`
- keep the CTK 13.1.1 explanation tables frozen as fallback for bindings versions that do not expose usable enum-member `__doc__`
- centralize the version gate + cleanup logic in `enum_explanations_helpers.py`
- update `cuda_utils` coverage to share the same version predicate and verify representative enums expose `__doc__` and that explanation text is attached to raised `CUDAError`s
- remove the obsolete `toolshed/reformat_cuda_enums_as_py.py`

### DIFF
--- a/cuda_core/tests/test_cuda_utils.py
+++ b/cuda_core/tests/test_cuda_utils.py
@@ -19,6 +19,15 @@ def _skip_if_bindings_pre_enum_docstrings():
         pytest.skip("cuda-bindings version does not expose usable enum __doc__ strings")
 
 
+def _assert_cleanup_example_matches_or_xfail(actual, expected):
+    # Pin a few real cleanup-sensitive enum docs. If one starts failing, review
+    # the raw ``__doc__`` and today's cleaned output: either update the expected
+    # text to match an acceptable upstream change, or fix the cleanup logic.
+    if actual != expected:
+        pytest.xfail("please review this failure")
+    assert actual == expected
+
+
 def test_check_driver_error():
     num_unexpected = 0
     for error in driver.CUresult:
@@ -71,6 +80,69 @@ def test_runtime_error_enum_has_non_empty_docstring():
     doc = runtime.cudaError_t.cudaErrorInvalidValue.__doc__
     assert doc is not None
     assert doc.strip() != ""
+
+
+# These use real enum members rather than synthetic strings, to pin a few
+# representative cleanup-sensitive docs end to end. Together with the helper
+# unit tests, this gives a harder assurance that today's live bindings output
+# is rendered into the user-facing text we expect. Unexpected changes are
+# marked as xfail so they prompt manual review of the drift, without causing
+# a hard test failure.
+@pytest.mark.parametrize(
+    ("explanations", "error", "expected"),
+    [
+        pytest.param(
+            cuda_utils.DRIVER_CU_RESULT_EXPLANATIONS,
+            driver.CUresult.CUDA_ERROR_NOT_INITIALIZED,
+            "This indicates that the CUDA driver has not been initialized with cuInit() or that initialization has failed.",
+            id="driver_not_initialized_role_cleanup",
+        ),
+        pytest.param(
+            cuda_utils.DRIVER_CU_RESULT_EXPLANATIONS,
+            driver.CUresult.CUDA_ERROR_INVALID_CONTEXT,
+            (
+                "This most frequently indicates that there is no context bound to the current thread. "
+                "This can also be returned if the context passed to an API call is not a valid handle "
+                "(such as a context that has had cuCtxDestroy() invoked on it). This can also be "
+                "returned if a user mixes different API versions (i.e. 3010 context with 3020 API calls). "
+                "See cuCtxGetApiVersion() for more details. This can also be returned if the green "
+                "context passed to an API call was not converted to a CUcontext using cuCtxFromGreenCtx API."
+            ),
+            id="driver_invalid_context_multiple_roles",
+        ),
+        pytest.param(
+            cuda_utils.RUNTIME_CUDA_ERROR_EXPLANATIONS,
+            runtime.cudaError_t.cudaErrorLaunchTimeout,
+            (
+                "This indicates that the device kernel took too long to execute. This can only occur "
+                "if timeouts are enabled - see the device attribute cudaDevAttrKernelExecTimeout for "
+                "more information. This leaves the process in an inconsistent state and any further "
+                "CUDA work will return the same error. To continue using CUDA, the process must be "
+                "terminated and relaunched."
+            ),
+            id="runtime_launch_timeout_role_cleanup",
+        ),
+        pytest.param(
+            cuda_utils.RUNTIME_CUDA_ERROR_EXPLANATIONS,
+            runtime.cudaError_t.cudaErrorIncompatibleDriverContext,
+            (
+                "This indicates that the current context is not compatible with this the CUDA Runtime. "
+                "This can only occur if you are using CUDA Runtime/Driver interoperability and have "
+                "created an existing Driver context using the driver API. The Driver context may be "
+                "incompatible either because the Driver context was created using an older version of "
+                "the API, because the Runtime API call expects a primary driver context and the Driver "
+                "context is not primary, or because the Driver context has been destroyed. Please see "
+                '"Interactions with the CUDA Driver API" for more information.'
+            ),
+            id="runtime_incompatible_driver_context_codegen_bug",
+        ),
+    ],
+)
+def test_enum_doc_cleanup_examples_are_reviewed_on_change(explanations, error, expected):
+    _skip_if_bindings_pre_enum_docstrings()
+
+    actual = explanations.get(int(error))
+    _assert_cleanup_example_matches_or_xfail(actual, expected)
 
 
 def test_check_driver_error_attaches_explanation():


rwgk:
I manually reviewed the cleaned driver/runtime enum-doc output for both `cuda-bindings 13.2.0` and `cuda-bindings 12.9.6` using this local helper script:

```python

import cuda.bindings
from cuda.bindings import driver, runtime

from cuda.core._utils.enum_explanations_helpers import clean_enum_member_docstring


def dump_enum_docs(label, enum_type):
    print(f"=== {label} ===")
    for member in enum_type:
        cleaned = clean_enum_member_docstring(member.__doc__)
        print(f"{member.name} ({int(member)})")
        print(cleaned if cleaned is not None else "<None>")
        print()


def main():
    print(f"cuda.bindings.__version__ = {cuda.bindings.__version__}")
    print()
    dump_enum_docs("driver.CUresult", driver.CUresult)
    dump_enum_docs("runtime.cudaError_t", runtime.cudaError_t)


if __name__ == "__main__":
    main()
```

High-level result from the manual review:

- the cleaned output looks presentable in both versions
- I did not see leftover `~`, `::`, or Sphinx-role markup in the cleaned strings
- this gave me enough confidence to use the docstring-backed path for `12.9.6 <= cuda-bindings < 13.0.0` as well as `cuda-bindings >= 13.2.0`
- the only remaining oddities I noticed look upstream/generated rather than cleanup-related: `cudaErrorIncompatibleDriverContext` still says `this the CUDA Runtime`, and `cudaErrorApiFailureBase` has no docstring (which still falls back to `cudaGetErrorString()` in `cuda_core`)

cuda_core/cuda/core/_utils/driver_cu_result_explanations.py
```diff
@@ -356,3 +351,5 @@
     ),
     999: "This indicates that an unknown internal error has occurred.",
 }
+
+DRIVER_CU_RESULT_EXPLANATIONS = get_best_available_explanations(driver.CUresult, _FALLBACK_EXPLANATIONS)
```
mdboom: It would be great to move this one level up so this module full of pre-written results doesn't have to be imported at all if we have a new enough `cuda-bindings`.

cuda_core/cuda/core/_utils/enum_explanations_helpers.py
```diff
@@ -0,0 +1,121 @@
...
+    s = re.sub(
+        r":(?:py:)?(?:obj|func|meth|class|mod|data|const|exc):`([^`]+)`",
+        lambda m: re.sub(r"^~?\.", "", m.group(1)),
+        s,
+    )
+    s = re.sub(r"\*\*([^*]+)\*\*", r"\1", s)
+    s = re.sub(r"\*([^*]+)\*", r"\1", s)
+    s = re.sub(r"\s+", " ", s).strip()
+    s = _fix_hyphenation_wordwrap_spacing(s)
+    return s
```
mdboom, line 89: This seems really brittle, but I guess I don't see a way around it.  I would prefer to just depend on `rst2txt` or something, that but seems unmaintained.

cuda_core/cuda/core/_utils/enum_explanations_helpers.py
```diff
@@ -0,0 +1,121 @@
...
+    for error messages.
+    """
+    if doc is None:
+        return None
+    s = doc
+    # Known codegen bug on cudaErrorIncompatibleDriverContext. Remove once fixed
+    # in cuda-bindings code generation.
+    s = s.replace("\n:py:obj:`~.Interactions`", ' "Interactions ')
+    s = re.sub(
+        r":(?:py:)?(?:obj|func|meth|class|mod|data|const|exc):`([^`]+)`",
```
mdboom: This regex seems too specific.  Could we just use one that would capture any rst directive?

cuda_core/cuda/core/_utils/enum_explanations_helpers.py
```diff
@@ -0,0 +1,121 @@
...
+    """Remove spaces around hyphens introduced by line wrapping in generated ``__doc__`` text.
+
+    This is a narrow workaround for wrapped forms such as ``non- linear`` that
+    would otherwise look awkward in user-facing messages.
+    """
+    prev = None
+    while prev != s:
+        prev = s
+        s = re.sub(r"([a-z])- ([a-z])", r"\1-\2", s)
+        s = re.sub(r"([a-z]) -([a-z])", r"\1-\2", s)
```
mdboom: Seems too narrow.  Why only perform this on lower-case words?

cuda_core/cuda/core/_utils/enum_explanations_helpers.py
mdboom: Something like (untested): 

```
r":(?:[a-z]+:)?(?:[a-z]):`([^`]+)`"
```

mdboom (changes requested):

rwgk:
**My Cursor prompt:**

> Regarding: importing `_FALLBACK_EXPLANATIONS` only when needed.

> Brainstorming: I guess we could achieve this by leaving `cuda_utils.pyx` untouched, but maybe it's better (especially long-term) if we also consider changing `cuda_utils.pyx`? Could you please think through the alternatives, how this would shake out either way? What's your recommendation?

**Cursor response:**

I checked the current import path:

- `cuda_core/cuda/core/_utils/cuda_utils.pyx:29` imports `DRIVER_CU_RESULT_EXPLANATIONS` and `RUNTIME_CUDA_ERROR_EXPLANATIONS` at module import time.
- `cuda_core/cuda/core/_utils/driver_cu_result_explanations.py:355` and `cuda_core/cuda/core/_utils/runtime_cuda_error_explanations.py:543` each define a huge `_FALLBACK_EXPLANATIONS` dict and then immediately call `get_best_available_explanations(...)`.

So today, even on `13.2.x` and `12.9.6`, importing `cuda_utils` still imports both big fallback-table modules.

**Option 1**
- Leave `cuda_utils.pyx` untouched.
- Move the large dicts into new fallback-only modules, for example `driver_cu_result_explanations_fallback.py` and `runtime_cuda_error_explanations_fallback.py`.
- Keep the current `driver_cu_result_explanations.py` and `runtime_cuda_error_explanations.py` as thin selector modules that either:
  - return `DocstringBackedExplanations(...)`, or
  - import the fallback module only on old bindings.

How it shakes out:
- `cuda_utils.pyx` stays exactly as-is.
- No Cython rebuild.
- Public/import shape stays stable.
- Existing tests in `cuda_core/tests/test_cuda_utils.py` mostly continue to work unchanged.
- If we want to avoid duplicating the version gate twice, `enum_explanations_helpers.py` can grow a small loader-based helper, e.g. “pick docstrings or call this fallback loader”.

Tradeoff:
- adds two small new modules
- a little bit of plumbing in pure Python

**Option 2**
- Move the version-selection logic up into `cuda_utils.pyx`.
- `cuda_utils.pyx` would either:
  - import fallback modules only when old bindings are detected, or
  - lazily initialize the explanation source on first use.

How it shakes out:
- this is the most literal reading of “move it one level up”
- could avoid importing even the selector modules until needed
- but it pushes import orchestration into Cython

Tradeoff:
- requires rebuilding `cuda_core`
- more invasive diff
- more logic in compiled code for something that is really import/selection policy
- less nice to unit-test in isolation than the pure-Python helper path
- you would likely need to touch `cuda_core/tests/test_cuda_utils.py` more, and maybe lose some of the clean separation you currently have in `enum_explanations_helpers.py`

**Option 3**
- Hybrid: keep `cuda_utils.pyx` mostly unchanged, but have it import lightweight getter functions instead of precomputed objects.
- That still means touching `cuda_utils.pyx`, but not much.

Tradeoff:
- more churn than Option 1
- not obviously better unless you specifically want first-use lazy initialization inside `cuda_utils.pyx`

**Recommendation**
- I would keep `cuda_utils.pyx` untouched for this PR.
- I’d implement the lazy fallback import entirely in Python by splitting the big tables into fallback-only modules and making the current driver/runtime modules thin selectors.
- If we want the selector code to stay tidy, I’d slightly generalize `get_best_available_explanations()` so it can accept a fallback loader instead of an already-imported dict.

Why I think this is the best fit:
- it satisfies the reviewer’s actual performance/import concern
- it avoids a Cython rebuild for an import-policy change
- it keeps the existing `.get(int(error))` shape intact everywhere
- it preserves the current separation of concerns: `cuda_utils.pyx` consumes explanations, while `_utils` decides where they come from

cuda_core/cuda/core/_utils/driver_cu_result_explanations.py
rwgk: Done: commit 70258546444afe54b2632a39d68e7ec7f601c996

See also: [comment here laying out our options for how to do this](https://github.com/NVIDIA/cuda-python/pull/1860#issuecomment-4194368360)

I went with Option 1.

cuda_core/cuda/core/_utils/enum_explanations_helpers.py
rwgk: As discussed offline: what we have here is sufficiently simple, extensively unit tested, and in the worst case we have a cosmetic issue (unwanted sphinx markup in the error messages).

To address the "brittle" concern, I added commit 27aae00d7e4529b1aa372d53b57fe698d58bfb7c

For easy reference, copy-pasting from the commit message:

> Add a small set of real enum-doc cleanup examples that assert today's exact cleaned output for representative live bindings cases. Mark unexpected drift as xfail so future upstream doc changes trigger manual review without causing a hard test failure.

I believe in combination with the rest of the extensive unit tests, there is very little wiggle room for major issues to go undetected.

cuda_core/cuda/core/_utils/enum_explanations_helpers.py
rwgk: Done: commit f70c09963e1c4e866d7392eee66e14f17fcd35b1

Cursor explained:

Updated to use a generic inline-role matcher instead of enumerating specific py roles.

Concretely, the cleanup now accepts both `:role:...` and `:domain:role:...` forms, so it handles things like `:term:`, `:c:func:`, and the existing `:py:...:` cases uniformly. I also expanded the focused tests to cover both domain-qualified and non-domain roles.

I manually re-ran show_all_enum_doc.py after this change; the current 13.2.x output stayed unchanged, which is what I wanted. The broader regex just makes the cleanup less brittle for future/generated variants.

cuda_core/cuda/core/_utils/enum_explanations_helpers.py
rwgk: Done: commit f70c09963e1c4e866d7392eee66e14f17fcd35b1

Cursor explained:

Broadened this so it is no longer limited to lowercase-only cases.

The cleanup now targets asymmetric word-wrap artifacts whenever the hyphen is attached to a word character on one side and separated by a space on the other, so it also covers cases like `GPU- Direct`, `peer -GPU`, and `L2- cache`. At the same time, it still avoids rewriting intentional separators like `a - b`.

I added focused tests for the uppercase/digit cases plus a negative case for preserving a real dash separator, and re-ran show_all_enum_doc.py to confirm the current 13.2.x output remains unchanged.

rwgk:
@cpcloud for visibility (ctk-next references the toolshed/reformat_cuda_enums_as_py.py script removed in this PR)

Issue #1839, opened 2026-03-31
Examples: Fix or remove conjugateGradientMultiBlockCG_test.py


mdboom:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

cuda.bindings

### Is your feature request related to a problem? Please describe.

This [example has been effectively disabled since 2021](https://github.com/NVIDIA/cuda-python/blob/main/cuda_bindings/examples/4_CUDA_Libraries/conjugateGradientMultiBlockCG_test.py#L212) "due to a bug in NVRTC".  

### Describe the solution you'd like

If it can be easily fixed (or the bug in NVRTC no longer exists), let's re-enable it.  If it's not easy, let's just remove the example, since it's not providing any value.

mdboom:
Fixed by #1922

PR #1861, opened 2026-04-05
[fix] Replace pytest.skip with requirement_not_met in conjugateGradientMultiBlockCG example (#1530)

### LINKED ISSUE #1530: [FEA]: Avoid importing pytest in all cuda_bindings/examples
### Is this a duplicate?

### Area

cuda.bindings

xref: https://github.com/NVIDIA/cuda-python/pull/1517#discussion_r2722857419

All examples should be runnable standalone. If an end user runs the example, and the headers are not found, then they'd hit this pytest import which may or may not fail right away, and even if it succeeds the next error message (coming from skip) would be confusing (they'd see "Skipped").

_No response_

### Additional context

_No response_

edenfunf:
## What

Replace all `pytest.skip()` calls with `requirement_not_met()` in
`cuda_bindings/examples/4_CUDA_Libraries/conjugateGradientMultiBlockCG_test.py`,
and remove the `import pytest` that was placed inside `main()`.

**Before:**
```python
def main():
    tol = 1e-5

    import pytest

    # WAIVE: Due to bug in NVRTC
    return

    if platform.system() == "Darwin":
        pytest.skip("conjugateGradientMultiBlockCG is not supported on Mac OSX")
    ...
    if not device_prop.managedMemory:
        pytest.skip("Unified Memory not supported on this device")
```

**After:**
```python
def main():
    tol = 1e-5

    # WAIVE: Due to bug in NVRTC
    requirement_not_met("conjugateGradientMultiBlockCG is currently waived due to a known NVRTC issue")

    if platform.system() == "Darwin":
        requirement_not_met("conjugateGradientMultiBlockCG is not supported on Mac OSX")
    ...
    if not device_prop.managedMemory:
        requirement_not_met("Unified Memory not supported on this device")
```

## Why

Issue #1530 tracks that pytest must not be imported in standalone examples.
After auditing all examples in `cuda_bindings/examples/`, this is the only
file that still imports pytest. Every other example already uses
`requirement_not_met()` from `cuda.bindings._example_helpers` to signal
unsupported configurations and exit cleanly via `sys.exit`.

Using `pytest.skip()` in a standalone script is incorrect — it raises
`pytest.skip.Exception` which is only meaningful when running under a pytest
test session. Outside of pytest, the exception propagates uncaught.

## How

- Added `requirement_not_met` to the `_example_helpers` import at the top of
  the file (it is already exported from `cuda.bindings._example_helpers.__all__`).
- Removed the `import pytest` statement that was placed inside `main()`.
- Replaced the silent `return` waive (for the known NVRTC issue) with a
  proper `requirement_not_met()` call that prints a message to stderr and
  exits with a non-zero code (or zero if `CUDA_BINDINGS_SKIP_EXAMPLE=0`).
- Replaced all remaining `pytest.skip(...)` calls with `requirement_not_met(...)`,
  matching the pattern used in every other example file.

## Test

- `ruff check` and `ruff format --check` pass on the modified file.
- `pre-commit run --files cuda_bindings/examples/4_CUDA_Libraries/conjugateGradientMultiBlockCG_test.py`
  passes all hooks (ruff, SPDX headers, end-of-file, trailing whitespace, etc.).
- The file has no remaining references to `pytest` (verified with grep).
- Behavior is unchanged: the example exits early due to the NVRTC waive,
  just now through the standard `requirement_not_met()` path instead of
  a silent `return`.

Fixes #1530
Relates to #1839

### DIFF
--- a/cuda_bindings/examples/4_CUDA_Libraries/conjugateGradientMultiBlockCG_test.py
+++ b/cuda_bindings/examples/4_CUDA_Libraries/conjugateGradientMultiBlockCG_test.py
@@ -26,6 +26,7 @@ from cuda.bindings._example_helpers import (
     KernelHelper,
     check_cuda_errors,
     find_cuda_device,
+    requirement_not_met,
 )
 
 conjugate_gradient_multi_block_cg = """\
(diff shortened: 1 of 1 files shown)


edenfunf:
Hi, could someone help review and trigger CI for this PR? Thanks!

rwgk:
Sorry we cannot take PRs against cuda-bindings, which has `SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE` (only cuda-core and cuda-pathfinder, which have `SPDX-License-Identifier: Apache-2.0`). Unfortunately I don't have a good idea for making that more obvious; I'll ask around.

edenfunf:
> Sorry we cannot take PRs against cuda-bindings, which has `SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE` (only cuda-core and cuda-pathfinder, which have `SPDX-License-Identifier: Apache-2.0`). Unfortunately I don't have a good idea for making that more obvious; I'll ask around.

Thanks for the clarification!

Got it — I didn’t realize that cuda-bindings is not open for external contributions due to the license. Appreciate the explanation.

I’ll focus on cuda-core / cuda-pathfinder or other open components instead.

Thanks again!

PR #1863, opened 2026-04-06
Fix managed memory misclassified as kDLCUDAHost in DLPack device mapping

rparolin:
## Summary
- `_smv_get_dl_device()` classified all device+host-accessible buffers as `kDLCUDAHost`, including managed (unified) memory which should be `kDLCUDAManaged`
- CCCL's `make_tma_descriptor` rejects `kDLCUDAHost` with *"Device type must be kDLCUDA or kDLCUDAManaged"*, breaking TMA descriptor creation from managed buffers
- Preserve the `is_managed` flag (already queried via `CU_POINTER_ATTRIBUTE_IS_MANAGED` in `_query_memory_attrs()`) in `_MemAttrs`, expose it on `Buffer`, and use it in `_smv_get_dl_device()` to return `kDLCUDAManaged`

Fixes: https://nvbugspro.nvidia.com/bug/6044342

### DIFF
--- a/cuda_core/tests/test_memory.py
+++ b/cuda_core/tests/test_memory.py
@@ -556,7 +556,7 @@ def test_buffer_dunder_dlpack():
     [
         (DummyDeviceMemoryResource, (DLDeviceType.kDLCUDA, 0)),
         (DummyHostMemoryResource, (DLDeviceType.kDLCPU, 0)),
-        (DummyUnifiedMemoryResource, (DLDeviceType.kDLCUDAHost, 0)),
+        (DummyUnifiedMemoryResource, (DLDeviceType.kDLCUDAManaged, 0)),
         (DummyPinnedMemoryResource, (DLDeviceType.kDLCUDAHost, 0)),
     ],
 )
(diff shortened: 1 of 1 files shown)


cuda_core/cuda/core/_memory/_buffer.pyx
```diff
@@ -396,6 +396,12 @@ cdef class Buffer:
         _init_mem_attrs(self)
         return self._mem_attrs.is_host_accessible
 
+    @property
+    def is_managed(self) -> bool:
```
cpcloud, line 391: This exposes the signal we need, but the managed-buffer DLPack export is still unchanged. `StridedMemoryView.from_any_interface(buffer)` on a `Buffer` goes through `Buffer.__dlpack_device__()` and `_dlpack.setup_dl_tensor_device()` first, and both of those still map `d && h` to `kDLCUDAHost`. That means managed `Buffer`s are still misclassified on the public DLPack path even after this patch.

Could we centralize the buffer-side DLPack classification and reuse it from both call sites? For example:

```suggestion
    @property
    def is_managed(self) -> bool:
        """Return True if this buffer is CUDA managed (unified) memory, otherwise False."""
        _init_mem_attrs(self)
        return self._mem_attrs.is_managed

    def _dlpack_device(self) -> tuple[int, int]:
        d = self.is_device_accessible
        h = self.is_host_accessible
        if d and (not h):
            return (DLDeviceType.kDLCUDA, self.device_id)
        if d and h:
            return (DLDeviceType.kDLCUDAManaged if self.is_managed else DLDeviceType.kDLCUDAHost, 0)
        if (not d) and h:
            return (DLDeviceType.kDLCPU, 0)
        raise BufferError("buffer is neither device-accessible nor host-accessible")
```

Then `Buffer.__dlpack_device__()` and `setup_dl_tensor_device()` can share the same mapping instead of drifting again.

cuda_core/cuda/core/_memoryview.pyx
```diff
@@ -607,8 +607,10 @@ cdef inline int _smv_get_dl_device(
             device_type = _kDLCUDA
             device_id = buf.device_id
         elif d and h:
-            # We do not currently differentiate pinned vs managed here.
-            device_type = _kDLCUDAHost
+            if buf.is_managed:
```
cpcloud: This branch only runs for views that are not already backed by an incoming DLPack capsule. For `StridedMemoryView.from_any_interface(managed_buf)`, `view.dl_tensor != NULL`, so we return the device type that `Buffer.__dlpack__()` encoded before we reach this code. In other words, the tensor-map regression still survives on the `Buffer` -> DLPack -> `StridedMemoryView` path.

At minimum, I'd add a sync note here so the buffer-side export is updated in the same patch:

```suggestion
            # Keep this mapping in sync with Buffer.__dlpack_device__() and _dlpack.setup_dl_tensor_device().
            if buf.is_managed:
                device_type = _kDLCUDAManaged
            else:
                device_type = _kDLCUDAHost
```

rparolin:
@cpcloud please re-review

PR #1866, opened 2026-04-06
docs: audit example coverage in published docs

### LINKED ISSUE #1680: Examples: Audit example coverage in documentation
Parent issue: #1673

### Goal

Ensure all examples are referenced or linked from the published documentation.

### Current state

- Some examples are referenced in `getting-started.rst`, others are not.
- Neither `cuda_core/examples/` nor `cuda_bindings/examples/` is comprehensively documented.
- Users browsing the docs may not discover examples that aren't linked.

### Tasks

1. Inventory which examples are currently referenced in the docs and where.
2. Identify examples that have no documentation references.
3. Add links/references for undiscovered examples in appropriate doc pages.

### Files affected

Documentation files in `cuda_core/docs/` and `cuda_bindings/docs/`, plus all 25 example files listed in #1673.

cpcloud:
## Summary
- add dedicated `examples` pages for `cuda.core` and `cuda.bindings` so every shipped example is linked from the published docs
- update the existing doc entry points to link the new example indexes, fix the stale `strided_memory_view.py` reference, and follow the snake_case bindings example rename from #1847
- make the new GitHub example links version-aware so release docs point at the matching package tag while `latest` and preview builds continue to point at `main`

## Test plan
- [x] `bash -n cuda_core/docs/build_docs.sh`
- [x] `bash -n cuda_bindings/docs/build_docs.sh`
- [x] `git diff --check`
- [x] Verified the new docs pages reference the current `cuda_core/examples/*.py` and `cuda_bindings/examples/**/*.py` files after the #1847 rename

Closes #1680.

### DIFF
--- a/cuda_bindings/pixi.lock
+++ b/cuda_bindings/pixi.lock
@@ -1081,21 +1081,21 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/cairo-1.18.4-h3394656_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/colorama-0.4.6-pyhd8ed1ab_1.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/conda-gcc-specs-15.2.0-h53410ce_16.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-cccl_linux-64-12.9.27-ha770c72_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-crt-dev_linux-64-12.9.86-ha770c72_2.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-cudart-12.9.79-h5888daf_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-cudart-dev-12.9.79-h5888daf_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-cudart-dev_linux-64-12.9.79-h3f2d84a_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-cudart-static-12.9.79-h5888daf_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-cudart-static_linux-64-12.9.79-h3f2d84a_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-cudart_linux-64-12.9.79-h3f2d84a_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-nvrtc-12.9.86-hecca717_1.conda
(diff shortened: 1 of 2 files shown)


cuda_bindings/docs/source/examples.rst
```diff
@@ -0,0 +1,68 @@
...
+
+This page links to the ``cuda.bindings`` examples shipped in the
+`cuda-python repository <https://github.com/NVIDIA/cuda-python/tree/main/cuda_bindings/examples>`_.
+Use it as a quick index when you want a runnable sample for a specific API area
+or CUDA feature.
+
+Introduction
+------------
+
+- `clock_nvrtc_test.py <https://github.com/NVIDIA/cuda-python/blob/main/cuda_bindings/examples/0_Introduction/clock_nvrtc_test.py>`_
```
mdboom: Do we want to link to main on github?  I worry the examples could get out-of-sync with the docs or the API.  We could:

### Link to a specific tag

Since we already have a `release` config value, we can automate this like:

```
- `clock_nvrtc_test.py <https://github.com/NVIDIA/cuda-python/blob/|release|
/cuda_bindings/examples/0_Introduction/clock_nvrtc_test.py>`_
```

-- though we probably want to do something special in the `conf.py` to map a "latest" doc build to point to the "main" branch.

### Include the files verbatim as pages in the docs

This could be mostly automated.

### Ensure examples work across a range of versions

We should fix/augment our CI to make sure all the examples are tested against the range of supported versions.  We are kind of there, but not quite.

mdboom: An alternative approach would be to include the example files in the doc build, and link to those rather than to github, which also could be automated.

cuda_bindings/pixi.toml
```diff
@@ -58,10 +58,37 @@ cuda-version = "12.*"
 [feature.cu13.dependencies]
 cuda-version = "13.2.*"
 
+[feature.docs.dependencies]
+cython = "*"
+myst-parser = "*"
+numpy = "*"
+numpydoc = "*"
+pip = "*"
```
mdboom: As I think I mentioned in the earlier PR -- this seems weird / potentially hazardous.

cpcloud:
Rebased onto current `upstream/main`, so this branch now includes the #1847 example rename. The follow-up updates in this PR refresh the bindings example references to the new snake_case filenames, make the new example links version-aware, and drop the overlapping `pixi` workflow changes from this PR.

cuda_bindings/docs/source/examples.rst
cpcloud: Updated this to use package-aware refs instead of hardcoding `main`. `latest` and preview builds now point at `main`, while versioned docs point at the matching package tag (`cuda-core-v...` for core and `v...` for bindings). I also refreshed the bindings filenames after rebasing onto #1847.

cuda_bindings/pixi.toml
cpcloud: Dropped the `pixi` docs env/task changes from this PR so it stays scoped to the docs coverage audit. The overlapping docs-workflow work can continue separately.

PR #1871, opened 2026-04-07
[no-ci] pr-metadata-check: restrict cuda_bindings changes to NVIDIA organization members

rwgk:
**Closed: Superseded by #1878**

Related issue: #1139

- Extend `pr-metadata-check` to read the PR author's association and changed files from the `pull_request_target` payload/API.
- Block PRs that touch `cuda_bindings/` unless the author association is `MEMBER` or `OWNER`, including rename cases.
- Update the workflow messaging to reflect the broader requirements check and point contributors to `cuda_bindings/LICENSE`.

**Note:** We don't have a practical way to end-to-end test this workflow before merging. If issues show up after merge, we can roll back quickly with a `[no-ci]` revert PR.

PR #1874, opened 2026-04-07
[no-ci] Fix race condition in PR metadata check workflow

leofang:
## Summary

The `pr-metadata-check` workflow reads assignees, labels, and milestone from `github.event.pull_request`, which is a snapshot taken at event trigger time. When a PR is opened, the `opened` event fires before labels/milestone/assignee are fully applied (e.g. by automation or rapid manual edits), causing the check to fail with stale data. Re-running the job doesn't help because it re-uses the same stale event payload.

This PR fixes the race by fetching live PR data via `gh api` instead of relying on the event payload. The downstream validation logic is unchanged.

## What changed

- Replaced `github.event.pull_request.{assignees,labels,milestone}` env vars with a single `gh api` call that fetches current PR state
- Uses `--jq` to filter the API response to only the 3 fields needed (assignees, labels, milestone)
- Added `PR_NUMBER`, `GH_REPO`, and `GH_TOKEN` env vars to support the API call
- All existing validation logic (module labels, type labels, blocked labels, milestone, assignee) remains identical

## Test plan

- Open a test PR and verify the check passes even if labels/milestone are added a few seconds after creation
- Verify the check still fails correctly when metadata is genuinely missing
- Verify bot/draft PRs are still skipped

-- Leo's bot

### DIFF
--- a/.github/workflows/pr-metadata-check.yml
+++ b/.github/workflows/pr-metadata-check.yml
@@ -39,7 +39,8 @@ jobs:
 
           # Fetch live PR data to avoid stale event payload (race condition
           # when labels/milestone are added shortly after PR creation).
-          PR_JSON=$(gh api "repos/${GH_REPO}/pulls/${PR_NUMBER}" \
+          PR_JSON=$(gh pr view "${PR_NUMBER}" --repo "${GH_REPO}" \
+            --json assignees,labels,milestone \
             --jq '{assignees: .assignees, labels: .labels, milestone: (.milestone.title // empty)}')
           ASSIGNEES=$(echo "$PR_JSON" | jq '.assignees')
           LABELS=$(echo "$PR_JSON" | jq '.labels')


.github/workflows/pr-metadata-check.yml
```diff
@@ -37,6 +37,13 @@ jobs:
             exit 0
           fi
 
+          # Fetch live PR data to avoid stale event payload (race condition
+          # when labels/milestone are added shortly after PR creation).
+          PR_JSON=$(gh api "repos/${GH_REPO}/pulls/${PR_NUMBER}")
```
cpcloud: You should pull out the subset that you need using `--jq '{assignees: .assignees, labels: .labels, milestone: .milestone.title // empty}'` and then further refine it in the subsequent subshell invocations. This avoids wasting cycles on huge payloads that you don't use.

leofang: updated in commit ee3e7a8

.github/workflows/pr-metadata-check.yml
```diff
@@ -37,6 +37,14 @@ jobs:
             exit 0
           fi
 
+          # Fetch live PR data to avoid stale event payload (race condition
+          # when labels/milestone are added shortly after PR creation).
+          PR_JSON=$(gh api "repos/${GH_REPO}/pulls/${PR_NUMBER}" \
```
cpcloud: why not `gh pr view "${PR_NUMBER}"` --json ... --jq`?

leofang: Good call — updated in commit f1850e0 to use `gh pr view "${PR_NUMBER}" --repo "${GH_REPO}" --json assignees,labels,milestone --jq ...`.

-- Leo's bot

rparolin (approved):

leofang:
As discussed in the meeting, I'll merge this after #1871.

leofang:
> As discussed in the meeting, I'll merge this after #1871.

Since we are dropping #1871 in favor of #1878, there is no merge conflict and we can merge this right away.

PR #1888, opened 2026-04-09
Move cuda_bindings to a src layout

mdboom:
This moves cuda_bindings to use a [src layout](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/), which is the recommended way to avoid a number of issues when the local checkout environment is getting inadvertently pulled in.  While this may in fact be a bug in `pip` that it isn't sufficiently isolating the build environment, it's still good practice to use `src` layout, so we may want to consider it.

Of course, a big change like this will have far-reaching merge implications and on other workflows etc., so I'm not suggesting we merge this as-is without some discussion.


rwgk:
This PR makes it so that we have to use

```
cd cuda_bindings/src/cuda/bindings
```

to jump to our sources. One extra level, four characters more, I guess it's only a tiny extra inconvenience overall.

But I'm wondering, could it also be:
```
cd src/cuda/bindings
cd src/cuda/core
cd src/cuda/pathfinder
```

I guess not? — Or is there some place for the README, pyproject.toml etc. files with that layout?

leofang:
No. Different projects need to have their own src. 

I am supportive of this move, though my explicit complaint for the whole situation is: Python needs to do better in distinguishing local scripts/modules vs out-of-tree builds. Forcing everyone to move to `src/` to avoid issues is very opinionated and unaesthetic.

rwgk:
> I am supportive of this move,

Sounds good to me. Is there already a plan for moving cuda_core and cuda_pathfinder? (Doing all asap would seem best to me.)

mdboom:
> I am supportive of this move, though my explicit complaint for the whole situation is: Python needs to do better in distinguishing local scripts/modules vs out-of-tree builds. Forcing everyone to move to src/ to avoid issues is very opinionated and unaesthetic.

Yes, I agree.  I think this is just an historical artifact that the language started as "run stuff out of the current directory" long before packages and virtual environments etc. were ever dreamed of.  It still hasn't fully moved into the future.

> Sounds good to me. Is there already a plan for moving cuda_core and cuda_pathfinder? (Doing all asap would seem best to me.)

I think we should do them all at the same time -- I just wanted to have a bot build a PoC first and prove it's possible.  Mainly I wanted to confirm that none of the GHA would need to change, since that would indicate things our /users/ might have to change.

There will be a little bit of fallout for anything that automates putting cython-gen/cybind code into this repo since the destination directory changes.

PR #1890, opened 2026-04-10
Fix debug builds of cuda_bindings and cuda_core

mdboom:
This builds on top of #1883 and is an alternative to it.

There are three problems this is trying to solve:

- Since the move to `build_hooks.py`, debug builds have been broken, since there is no way to pass `--debug` in `sys.argv` to the underlying build.  This updates it to use a PEP 517 configuration flag instead, which can be passed through `pip` or `uv` etc.

- Tying debug builds directly to editable installs seems like the wrong approach.  Personally, I find editable installs to be more trouble than they are worth for compiled projects (such as this), and I like to install into multiple virtual environments with different versions of Python simultaneously, which you can't do with editable installs.  This retains the behavior that editable installs are always debuggable, but also makes it possible to build non-editable installs as debuggable.

- This ties debugging flags to stripping.  IMHO, there is never a need to build with `-O0 -g` and then strip symbols, so this links them together under a single config.

### DIFF
--- a/cuda_bindings/README.md
+++ b/cuda_bindings/README.md
@@ -12,7 +12,8 @@ This subpackage adheres to the developing practices described in the parent meta
 
 ## Debugging
 
-Pass the `pip` / `uv` configuration option `-C="debug=True"` explicitly to build debuggable binaries.
+Pass the `pip` / `uv` configuration option `-C="debug=True"` or
+`--config-settings="debug=True"` to explicitly to build debuggable binaries.
 Debuggable binaries are built by default for editable builds.
 
 Debuggable builds are not supported on Windows.
--- a/cuda_bindings/README.md
+++ b/cuda_bindings/README.md
@@ -12,12 +12,8 @@ This subpackage adheres to the developing practices described in the parent meta
 
 ## Debugging
 
-Editable installs have debuggable binaries by default.  To build a non-editable
-debug build, pass the `debug=True` configuration option to `pip` or `uv`:
-
-```
-pip install -v ./cuda_bindings -C="debug=True"
-```
+Pass the `pip` / `uv` configuration option `-C="debug=True"` explicitly to build debuggable binaries.
+Debuggable binaries are built by default for editable builds.
 
 Debuggable builds are not supported on Windows.
 
--- a/cuda_bindings/README.md
+++ b/cuda_bindings/README.md
@@ -10,6 +10,15 @@ Please refer to the [Installation page](https://nvidia.github.io/cuda-python/cud
 
 This subpackage adheres to the developing practices described in the parent metapackage [CONTRIBUTING.md](https://github.com/NVIDIA/cuda-python/blob/main/CONTRIBUTING.md).
 
+## Debugging
+
+Editable installs have debuggable binaries by default.  To build a non-editable
+debug build, pass the `debug=True` configuration option to `pip` or `uv`:
+
+```
+pip install -v ./cuda_bindings -C="debug=True"
+```
+
 ## Testing
 
 Testing dependencies can be installed using the `[test]` optional dependency identifier. For example, `pip install -v -e .[test]`.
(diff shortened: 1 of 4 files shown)


cuda_bindings/build_hooks.py
```diff
@@ -457,10 +428,11 @@ def _cleanup_dst_files():
...
 def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
-    _build_cuda_bindings(strip=True)
+    debug = config_settings.get("debug", False) if config_settings else False
+    _build_cuda_bindings(debug=debug)
     return _build_meta.build_wheel(wheel_directory, config_settings, metadata_directory)
 
 
 def build_editable(wheel_directory, config_settings=None, metadata_directory=None):
-    _build_cuda_bindings(strip=False)
+    _build_cuda_bindings(debug=True)
```
rwgk: Could we add this here?
```
    debug = config_settings.get("debug", True) if config_settings else True
```

That would make things a little more obvious (at least to me):
* Passing `-C="debug=False"` or `-C="debug=True"` works universally (better than silently ignoring in editable builds)
* For editable builds `debug=True` is the default
* For non-editable builds `debug=False` is the default

cuda_bindings/build_hooks.py
rwgk: It'd be great to this catch here:
```
        if sys.platform == "win32":
            if debug:
                raise RuntimeError("Debuggable builds are not supported on Windows.")
        else:
```

cuda_core/README.md
```diff
@@ -10,6 +10,17 @@ Please refer to the [Installation page](https://nvidia.github.io/cuda-python/cud
 
 This subpackage adheres to the developing practices described in the parent metapackage [CONTRIBUTING.md](https://github.com/NVIDIA/cuda-python/blob/main/CONTRIBUTING.md).
 
+## Debugging
+
+Editable installs have debuggable binaries by default.  To build a non-editable
+build, pass the `debug=True` configuration option to `pip` or `uv`:
```
rwgk: Pass the `pip` / `uv` configuration option `-C="debug=True"` explicitly to request debuggable binaries. However, this is only needed for non-editable builds; for editable builds `debug=True` is the default.

cuda_bindings/build_hooks.py
rwgk: Oh, wait: we can only set to True if `sys.platform != "win32"`:
```
debug_default = sys.platform != "win32"
debug = config_settings.get("debug", debug_default) if config_settings else debug_default
```

mdboom: It theoretically should still work on Windows, it will just have no effect.

cuda_bindings/README.md
```diff
@@ -10,6 +10,13 @@ Please refer to the [Installation page](https://nvidia.github.io/cuda-python/cud
 
 This subpackage adheres to the developing practices described in the parent metapackage [CONTRIBUTING.md](https://github.com/NVIDIA/cuda-python/blob/main/CONTRIBUTING.md).
 
+## Debugging
+
+Pass the `pip` / `uv` configuration option `-C="debug=True"` explicitly to build debuggable binaries.
```
leofang: nit: can we mention both `--config-settings` and `-C` just for searchability?

cuda_bindings/build_hooks.py
rwgk: My idea was to avoid surprises: someone assuming it has an effect, then "troubleshooting" only to eventually find ah it's silently ignored.

leofang (approved):
Thanks for taking over, Mike! I have been hoping to get meaningful support for `--config-settings`. This is a good start (and solves multiple issues, size inflation and debug build)!

leofang:
Since https://github.com/NVIDIA/cuda-python/commit/398b28dadb107480444c999e6ab50060c88ddb89 was green and the latest change was doc-only, let me admin-merge this.

PR #1893, opened 2026-04-10
Pin commit of msvc-dev-cmd

mdboom:
To prevent sidechain attacks.


leofang (approved):

leofang:
Confirmed our org admin allows this commit.

PR #1894, opened 2026-04-10
Add torch.Tensor fast path for StridedMemoryView via AOTI tensor bridge

### LINKED ISSUE #749: Provide a fast path for constructing a `StridedMemoryView` from a `torch.tensor`

leofang:
## Summary

Add a fast path for constructing `StridedMemoryView` from `torch.Tensor` objects using PyTorch's AOT Inductor (AOTI) stable C ABI, bypassing the DLPack/CAI protocol overhead.

### How it works

When a `torch.Tensor` is passed to any `from_*` classmethod (`from_dlpack`, `from_cuda_array_interface`, `from_array_interface`, or `from_any_interface`), the tensor metadata (data pointer, shape, strides, dtype, device) is read directly from the underlying C struct via AOTI function pointers, instead of going through the Python-level `__dlpack__()` or `__cuda_array_interface__` protocols.

The key technique (`pyobj_to_aten_handle`) extracts the `AtenTensorHandle` by offsetting past `PyObject_HEAD` in the `THPVariable` struct — pure C pointer arithmetic with zero Python API calls. The AOTI functions (`aoti_torch_get_data_ptr`, `aoti_torch_get_sizes`, etc.) then read tensor metadata through PyTorch's stable C ABI.

**PyTorch is NOT a build-time or runtime dependency.** The AOTI symbols are resolved lazily at runtime from `torch._C` (loaded with `RTLD_GLOBAL`) only when the user actually passes a `torch.Tensor`. The `_tensor_bridge` module is never imported at `cuda.core` load time.

### Performance

Benchmarked with `%timeit` (Python 3.12, PyTorch 2.11, NVIDIA RTX 6000 Ada):

```
                          v0.7.0 (DLPack)   tensor-bridge (AOTI)   speedup
GPU tensor stream_ptr=-1:    4.41 us             565 ns              ~8x
GPU tensor stream_ptr=0:    13.20 us             624 ns             ~21x
CPU tensor stream_ptr=-1:    5.08 us             552 ns              ~9x
```

At the C level (no Python overhead), AOTI extracts all 7 metadata fields in ~14 ns — ~4x faster than the DLPack C exchange API (~60 ns) for the same metadata.

### Stream ordering

- **AOTI fast path**: When `stream_ptr != -1`, establishes stream ordering between PyTorch's current CUDA stream (queried via `aoti_torch_get_current_cuda_stream`) and the consumer stream, using the same event-based pattern as the existing CAI path.
- **CAI fallback safety**: PyTorch's `__cuda_array_interface__` reports version 2 with no `stream` field, making the standard CAI sync path a no-op. We detect torch tensors in the CAI path and apply AOTI-based stream ordering to fix this safety gap.

### Version compatibility

- **Minimum**: PyTorch 2.3 (when the AOTI functions we use were introduced)
- **Maximum**: PyTorch 2.11 (latest tested; the `THPVariable` struct layout and `AtenTensorHandle == at::Tensor*` identity are undocumented internals)
- Unknown versions gracefully fall back to DLPack/CAI paths

### Files changed

- **`cuda/core/_tensor_bridge.pyx`** (new): AOTI tensor bridge — `pyobj_to_aten_handle`, `view_as_torch_tensor`, `sync_torch_stream`, dtype/itemsize mapping
- **`cuda/core/_include/aoti_shim.h`** (new): Vendored subset of PyTorch's AOTI stable C ABI declarations
- **`cuda/core/_memoryview.pyx`**: Torch detection (`_is_torch_tensor` with type cache + version bounds), lazy bridge loading, fast path in all `from_*` classmethods, CAI stream safety fix, lazy dtype resolution
- **`tests/test_utils.py`**: 12 new test cases (dtypes, shapes, slicing, devices, decorator)
- **`docs/source/release/1.0.0-notes.rst`**: Release notes entry

Closes #749

Co-Authored-By: Emilio Castillo <ecastillo@nvidia.com>

## Test plan

- [x] All existing tests pass (117 passed, 38 skipped)
- [x] 12 new torch tensor bridge tests pass (CPU + GPU)
- [x] Stream ordering verified for both AOTI and CAI paths
- [x] Graceful fallback for torch < 2.3 (returns False, uses DLPack)
- [ ] CI validation


cuda_core/cuda/core/_tensor_bridge.pyx
```diff
@@ -0,0 +1,327 @@
...
+cdef inline AtenTensorHandle pyobj_to_aten_handle(object obj):
+    """Extract AtenTensorHandle by offsetting past PyObject_HEAD.
+
+    In PyTorch 2.3–2.9 the first field after PyObject_HEAD is
+    ``c10::MaybeOwned<at::Tensor> cdata``; from 2.10 onward it is
+    ``at::Tensor cdata``.  In both cases the address of ``cdata``
+    is usable as an ``AtenTensorHandle`` (``at::Tensor*``) for the
+    AOTI stable C ABI functions.
+    """
+    return <AtenTensorHandle>(<char*><PyObject*>obj + sizeof(PyObject))
```
leofang: Note: I have filed a feature request to discuss if this API can be formalized in AOTI directly, so that we can relax upper bound safely and be forward compatible: https://github.com/pytorch/pytorch/issues/180107.

cuda_core/cuda/core/_include/aoti_shim.h
```diff
@@ -0,0 +1,83 @@
+/*
+ * Vendored subset of PyTorch's AOT Inductor (AOTI) stable C ABI.
+ * Original: torch/csrc/inductor/aoti_torch/c/shim.h
+ * PyTorch is licensed under the BSD 3-Clause License.
+ * See https://github.com/pytorch/pytorch/blob/main/LICENSE
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
```
leofang: xref: nvbugs 6077493

leofang:
/ok to test [c7331a9](https://github.com/NVIDIA/cuda-python/pull/1894/commits/c7331a9530f450641ab096a5acdc1e55087e2946)

leofang:
/ok to test [37fce1a](https://github.com/NVIDIA/cuda-python/pull/1894/commits/37fce1a0fcc5c23a505b0373d18b0ef134d4ebae)

leofang:
/ok to test [f6a3032](https://github.com/NVIDIA/cuda-python/pull/1894/commits/f6a303254bbedec015ccdd51cd0bf29b7a11ea76)

leofang:
/ok to test [2748a52](https://github.com/NVIDIA/cuda-python/pull/1894/commits/2748a52fbdcf9ea007e8e207b53bfd15627a8f66)

leofang:
/ok to test [d543be1](https://github.com/NVIDIA/cuda-python/pull/1894/commits/d543be141f6407190f053f99450128b21ea312bf)

leofang:
pre-commit.ci autofix

leofang:
pre-commit.ci autofix

cuda_core/cuda/core/_memoryview.pyx
```diff
@@ -29,6 +31,73 @@ from cuda.core._utils.cuda_utils cimport HANDLE_RETURN
...
+    """Bootstrap AOTI symbols, then import _tensor_bridge on first use."""
+    global _tensor_bridge
+    if _tensor_bridge is not None:
+        return _tensor_bridge
+    torch_C = sys.modules.get("torch._C")
+    if torch_C is None:
+        raise RuntimeError(
+            "torch._C is not loaded; cannot initialise the tensor bridge. "
+            "Make sure PyTorch is imported before passing a torch.Tensor.")
+    ctypes.CDLL(torch_C.__file__, mode=ctypes.RTLD_GLOBAL)
```
NaderAlAwar, line 95: question: if Windows is supported here, should this path handle Windows explicitly instead of always using mode=ctypes.`RTLD_GLOBAL`? There isn’t a direct Windows equivalent for RTLD_GLOBAL

NaderAlAwar (approved):

cuda_core/cuda/core/_memoryview.pyx
leofang: As per the ctypes docs, on Windows `mode` is ignored.

leofang:
/ok to test [615f984](https://github.com/NVIDIA/cuda-python/pull/1894/commits/615f984ce1b823cce1a6760f23a715e0900e7558)

cuda_core/cuda/core/_include/aoti_shim.h
```diff
@@ -0,0 +1,106 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef int32_t AOTITorchError;
+
+/* Opaque tensor handle -- corresponds to at::Tensor on the C++ side. */
+struct AtenTensorOpaque;
+typedef struct AtenTensorOpaque* AtenTensorHandle;
+
```
rwgk, line 52: To help future maintainers and agents:

```
/*
 * IMPORTANT: Keep the AOTI_SHIM_API declaration list below in sync with
 * aoti_shim.def. On Windows, build_hooks.py turns that .def file into the
 * stub import library that MSVC needs to link _tensor_bridge without making
 * PyTorch a build-time dependency. If you add, remove, or rename an imported
 * AOTI symbol here, update aoti_shim.def in the same change.
 */
```

cuda_core/cuda/core/_include/aoti_shim.def
```diff
@@ -0,0 +1,29 @@
+; Stub import library definition for PyTorch's AOTI stable C ABI symbols.
+; Used on Windows only: 'lib /DEF:aoti_shim.def /OUT:aoti_shim.lib /MACHINE:X64'
+; generates a minimal import library that satisfies the MSVC linker.
+; At runtime the symbols resolve from torch_cpu.dll (loaded by 'import torch').
```
rwgk, line 4: Similar to the suggested comment in aoti_shim.h:

```
; IMPORTANT: Keep this export list in sync with the AOTI_SHIM_API declarations
; in aoti_shim.h. build_hooks.py turns this file into the stub import library
; that MSVC uses to link _tensor_bridge, so any added/removed/renamed AOTI
; symbol must be updated in both files.
```

rwgk (commented):

I manually checked the `_tensor_bridge.pyx` code referenced below: the findings seem correct.

___

- **High** `cuda_core/cuda/core/_memoryview.pyx:221` routes every recognized torch tensor into `view_as_torch_tensor()`, but `cuda_core/cuda/core/_tensor_bridge.pyx:318` only reads raw metadata and never reenacts PyTorch's export guardrails. That means tensors PyTorch intentionally rejects at the protocol boundary, notably `requires_grad`, conjugated, non-strided/sparse, and wrong-current-device CUDA tensors, no longer fail the same way here. `cuda_core/cuda/core/_tensor_bridge.pyx:343` even says those cases are rejected, but there is no code that does it.

- **Medium** `cuda_core/cuda/core/_memoryview.pyx:146`, `cuda_core/cuda/core/_memoryview.pyx:1028`, and `cuda_core/cuda/core/_memoryview.pyx:1036` make `-1` the explicit opt-out for synchronization and treat `stream_ptr=None` as ambiguous. The new torch path changes that: `cuda_core/cuda/core/_tensor_bridge.pyx:359` treats `None` as "do nothing", so `StridedMemoryView.from_dlpack(torch_cuda_tensor)` now returns an unsynchronized view instead of forcing the caller to choose a stream policy.

- **Medium** The AOTI dtype tables at `cuda_core/cuda/core/_tensor_bridge.pyx:194` and `cuda_core/cuda/core/_tensor_bridge.pyx:227` only cover `uint8` on the unsigned side, while the existing DLPack conversion starting at `cuda_core/cuda/core/_memoryview.pyx:1108` supports `uint16`, `uint32`, and `uint64`. Because all torch tensors now short-circuit before the DLPack path at `cuda_core/cuda/core/_memoryview.pyx:221`, those torch dtypes regress from "works via DLPack" to `TypeError("Unsupported AOTI dtype code")`. The new dtype tests at `cuda_core/tests/test_utils.py:1000` only cover the narrower happy-path set.

rwgk:
I forgot to add: I spent a few minutes to answer: Why did we not need a `.def` file in Boost.Python? — Conclusion: because we were using `__declspec(dllexport)` and linking against `boost_python*.lib`. For the purposes of the tensor bridge in this PR, it's a hard requirement that we don't link against a pytorch lib, therefore the `.def` file is unavoidable.

cuda_core/cuda/core/_include/aoti_shim.def
leofang: done in 6b0dffec20597559e697ffaba7d81bc9add7b497

cuda_core/cuda/core/_include/aoti_shim.h
leofang: done in 6b0dffec20597559e697ffaba7d81bc9add7b497

leofang:
> * **High** `cuda_core/cuda/core/_memoryview.pyx:221` routes every recognized torch tensor into `view_as_torch_tensor()`, but `cuda_core/cuda/core/_tensor_bridge.pyx:318` only reads raw metadata and never reenacts PyTorch's export guardrails. That means tensors PyTorch intentionally rejects at the protocol boundary, notably `requires_grad`, conjugated, non-strided/sparse, and wrong-current-device CUDA tensors, no longer fail the same way here. `cuda_core/cuda/core/_tensor_bridge.pyx:343` even says those cases are rejected, but there is no code that does it.

Discussed offline. We would acknowledge this as a breaking change and move forward, for two reasons:
1. The existence of SMV is, in a sense, to allow “unsafe” access patterns that other CUDA libraries (cccl, nvmath, cutile, …) could tap into. The fact that we have to get the pointer to a torch tensor, check for contiguity, and start working, is already deviating from a DL framework’s expectation. The checks don’t make it safer for interop at the autograd level.
2. PyTorch’s own C-level DLPack `__dlpack_c_exchange_api__` implementation does not do any of these checks either. So this deficiency already exists if we lower DLPack from Python to C. 

> * **Medium** `cuda_core/cuda/core/_memoryview.pyx:146`, `cuda_core/cuda/core/_memoryview.pyx:1028`, and `cuda_core/cuda/core/_memoryview.pyx:1036` make `-1` the explicit opt-out for synchronization and treat `stream_ptr=None` as ambiguous. The new torch path changes that: `cuda_core/cuda/core/_tensor_bridge.pyx:359` treats `None` as "do nothing", so `StridedMemoryView.from_dlpack(torch_cuda_tensor)` now returns an unsynchronized view instead of forcing the caller to choose a stream policy.

Fixed in commit 6b0dffe.

> * **Medium** The AOTI dtype tables at `cuda_core/cuda/core/_tensor_bridge.pyx:194` and `cuda_core/cuda/core/_tensor_bridge.pyx:227` only cover `uint8` on the unsigned side, while the existing DLPack conversion starting at `cuda_core/cuda/core/_memoryview.pyx:1108` supports `uint16`, `uint32`, and `uint64`. Because all torch tensors now short-circuit before the DLPack path at `cuda_core/cuda/core/_memoryview.pyx:221`, those torch dtypes regress from "works via DLPack" to `TypeError("Unsupported AOTI dtype code")`. The new dtype tests at `cuda_core/tests/test_utils.py:1000` only cover the narrower happy-path set.

Fixed in commit 6b0dffe.

PR #1895, opened 2026-04-12
[no-ci][DO NOT MERGE] DUMMY CHANGE: cuda_bindings/pyproject.toml

rwgkunplugged:
DUMMY CHANGE for testing new workflow as external user:

https://github.com/NVIDIA/cuda-python/blob/33efe60c447cbc41b0511badaf190c28f2dd4124/.github/workflows/restricted-paths-guard.yml


leofang:
(Let me re-test with this PR.)

leofang:
@rwgk I see the label is re-applied (I removed it before re-opening the PR). I see the action summary correctly points it out: https://github.com/NVIDIA/cuda-python/actions/runs/24728813508?pr=1895. But the CI is not failed. I assume we haven't wired up the CI status check and this is expected?

rwgk:
> I assume we haven't wired up the CI status check and this is expected?

Yes.

There is #1896 already, which adds the merge gate (maybe I should call it "merge block"?), but I'm intentionally keeping it in draft mode until we have more confidence that the label assignment workflow is sufficiently reliable.

leofang:
Sounds good, thanks Ralf!

PR #1896, opened 2026-04-12
[no-ci] CI: add dedicated merge gate for restricted-paths review

rwgk:
## Summary

Follow-on to PR #1878

Add a dedicated GitHub Actions workflow, `restricted-paths-review-gate.yml`, that fails when the `Needs-Restricted-Paths-Review` label is present on a PR.

This keeps the merge-blocking behavior narrowly scoped to the restricted-paths policy, instead of making the broader checks in `pr-metadata-check.yml` required.

## Why a separate workflow

- `pr-metadata-check.yml` is not currently merge-blocking.
- Reusing `pr-metadata-check.yml` would mix two different policies:
  PR metadata hygiene and restricted-paths review.
- A dedicated gate gives us one required check with one clear meaning:
  if `Needs-Restricted-Paths-Review` is present, merging is blocked until a maintainer removes the label.

## Behavior

- The existing `restricted-paths-guard.yml` workflow continues to assign `Needs-Restricted-Paths-Review` when an untrusted author touches `cuda_bindings/` or `cuda_python/`.
- The new `restricted-paths-review-gate.yml` workflow reads the live PR labels and:
  - fails if `Needs-Restricted-Paths-Review` is present
  - passes otherwise
- This is implemented as a separate `pull_request_target` workflow so it can be used as a required status check in branch protection / rulesets.

## Important rollout detail

This PR adds the workflow only. The check does not become merge-blocking until a repo admin adds the new check to the repository ruleset after this PR is merged.

## Post-merge admin steps

1. Open or update a PR so the new workflow runs at least once from the base branch, and GitHub records the check context.
2. Go to `Settings` -> `Rules` -> `Rulesets` -> `Prevent committing without PR`.
3. Edit `Required status checks`.
4. Add the check named `Restricted paths review gate`.
5. Save the ruleset.
6. Verify on a PR with `Needs-Restricted-Paths-Review`: merging should be blocked while the label is present, and unblocked after a maintainer removes it.

## Branch-scope note

The existing `Prevent committing without PR` ruleset applies to:

- `main`
- `12.9.x`
- `11.8.x`
- `release/**/*`

If the new gate should apply to all of those branches, editing that existing ruleset is the right approach.

If the new gate should apply only to `main`, the better follow-up is to create a separate ruleset scoped only to `refs/heads/main` and require `Restricted paths review gate` there.

___

Made with Cursor GPT-5.4 Extra High Fast

### DIFF
--- a/.github/workflows/restricted-paths-review-gate.yml
+++ b/.github/workflows/restricted-paths-review-gate.yml
@@ -4,9 +4,9 @@
 name: "CI: Restricted Paths Review Gate"
 
 on:
-  # Temporary for PR-level testing before this workflow exists on the base
-  # branch. Switch back to pull_request_target before merging.
-  pull_request:
+  # Keep this separate from pr-metadata-check.yml so only the
+  # Needs-Restricted-Paths-Review policy becomes merge-blocking.
+  pull_request_target:
     types:
       - opened
       - synchronize
--- a/.github/workflows/restricted-paths-review-gate.yml
+++ b/.github/workflows/restricted-paths-review-gate.yml
@@ -1,0 +1,81 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: "CI: Restricted Paths Review Gate"
+
+on:
+  # Keep this separate from pr-metadata-check.yml so only the
+  # Needs-Restricted-Paths-Review policy becomes merge-blocking.
+  pull_request_target:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - ready_for_review
+      - labeled
+      - unlabeled
+
+jobs:
+  restricted-paths-review-gate:
+    name: Restricted paths review gate
+    if: github.repository_owner == 'NVIDIA'
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read
+    steps:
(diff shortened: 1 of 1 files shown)


rwgk:
Smoke testing, as described in the message of commit 2b94e27a90b16daeee1354f87353829de05844b8:

**Manually add Needs-Restricted-Paths-Review label**

Screenshot 2026-04-12 at 15 29 42

___

**Summary of failed (as expected) workflow**

Screenshot 2026-04-12 at 15 30 29

___

**Manually remove Needs-Restricted-Paths-Review label**

Screenshot 2026-04-12 at 15 31 38

___

**Summary of successful workflow**

Screenshot 2026-04-12 at 15 32 09

rwgk:
@aterrel I think the chance of something still slipping through the cracks is very small:
- The Needs-Restricted-Paths-Review label is sticky: it is only added automatically, never removed automatically.
- The automatic exemption is narrowly defined: the label is _not_ added only when the PR author is a `COLLABORATOR`, `MEMBER`, or `OWNER`, and those trusted associations are tightly controlled by repository/organization administration.
- Only someone with repository triage-or-higher access can remove the label before merging, which is a fairly small group of people.

rwgk:
Looks like I still need to refine the `restricted-paths-guard.yml` workflow before we merge this PR:

https://github.com/NVIDIA/cuda-python/pull/1878#issuecomment-4237778198

rwgk:
Relicensing under PR #2293 has made this PR obsolete.

PR #1901, opened 2026-04-14
[no-ci] CI: drive restricted paths guard from user-in-org-check-bot

rwgk:
Related issue: #1139

Depends on the paired `github-apps` PR 78 that adds and deploys `user-in-org-check-bot`.

**TODO before merging:**

- switch `.github/workflows/restricted-paths-guard.yml` back from `pull_request` to `pull_request_target`
- remove the temporary `# XXX DUMMY CHANGE XXX` line from `cuda_bindings/pyproject.toml`

## Summary

This PR is the `cuda-python` half of the rollout for the new `user-in-org-check-bot`.

It adds repository config for the new bot and rewrites `.github/workflows/restricted-paths-guard.yml` to consume explicit bot-managed PR-author membership labels instead of relying on `author_association`.

## What changes

- add `.github/user-in-org-check-bot.yaml` to configure `user-in-org-check-bot` for `NVIDIA/cuda-python`
- update `restricted-paths-guard.yml` to recognize these terminal membership labels:
  - `PR-Author-In-NVIDIA-Org`
  - `PR-Author-Outside-NVIDIA-Org`
  - `PR-Author-Org-Check-Failure`
- keep `Needs-Restricted-Paths-Review` as the policy label for confirmed non-member PRs that touch `cuda_bindings/` or `cuda_python/`
- fail the workflow on operational problems (to cleanly avoid guessing)
- only re-evaluate on `labeled` / `unlabeled` events when one of the membership-bot labels changes

## Workflow behavior

For PRs that do not touch restricted paths, the guard exits green.

For PRs that touch restricted paths, the guard now works as follows:

- if `Needs-Restricted-Paths-Review` is already present, exit quickly and leave it sticky for manual removal
- if `PR-Author-In-NVIDIA-Org` is already present, exit quickly and pass
- if `PR-Author-Outside-NVIDIA-Org` is already present, ensure `Needs-Restricted-Paths-Review` is present and finish green
- if `PR-Author-Org-Check-Failure` is already present, fail with guidance to rerun later or apply a manual override label
- otherwise, poll live PR labels for up to about 60 seconds waiting for a terminal bot result
- if no terminal result appears within that window, fail
- if both `PR-Author-In-NVIDIA-Org` and `PR-Author-Outside-NVIDIA-Org` are present at the same time, fail explicitly because the state is inconsistent

## Why this shape

- It separates the authoritative org-membership question from the repo-specific policy decision.
- It avoids false positives from `author_association`.
- It keeps the rescue path obvious in pathological cases:
  - maintainers can manually apply `PR-Author-In-NVIDIA-Org`
  - maintainers can manually apply `PR-Author-Outside-NVIDIA-Org`
  - maintainers can still manage `Needs-Restricted-Paths-Review` manually if needed
- It avoids catch-22 behavior by keeping manual intervention possible when the bot or workflow is unavailable.

## Scope boundary

This PR does **not** make `Needs-Restricted-Paths-Review` merge-blocking by itself.

That remains intentionally separate from the guard logic. The follow-on gate PR, `#1896`, handles the merge-blocking status check so the policy and enforcement pieces can be reviewed independently.

## Testing note

Because Probot repo config is read from the default branch, the full end-to-end behavior with the deployed app cannot be exercised from this PR alone before merge.

To make the workflow logic testable in this PR:

- the trigger is temporarily changed from `pull_request_target` to `pull_request`
- `cuda_bindings/pyproject.toml` includes a temporary dummy line so the PR definitely touches a restricted path
- PR labels can be manipulated manually to simulate the bot outcomes

## Test plan

- Open a PR from this branch and confirm the temporary `pull_request` trigger runs.
- With no terminal bot label present, confirm the guard times out after about 60 seconds and fails.
- Add `PR-Author-In-NVIDIA-Org` and confirm the guard succeeds quickly without adding `Needs-Restricted-Paths-Review`.
- Add `PR-Author-Outside-NVIDIA-Org` and confirm the guard adds `Needs-Restricted-Paths-Review` and succeeds.
- Add `PR-Author-Org-Check-Failure` and confirm the guard fails with rerun/manual-override guidance.
- Add both `PR-Author-In-NVIDIA-Org` and `PR-Author-Outside-NVIDIA-Org` and confirm the guard fails with an explicit conflicting-label error.
- Before merge, remove the dummy change and switch the workflow trigger back to `pull_request_target`.

### DIFF
--- a/.github/user-in-org-check-bot.yaml
+++ b/.github/user-in-org-check-bot.yaml
@@ -1,0 +1,13 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# Configuration file for `user-in-org-check-bot` GitHub App.
+
+enabled: true
+org: NVIDIA
+member_label: PR-Author-In-NVIDIA-Org
+# Make non-member outcomes explicit so the restricted-paths workflow can
+# distinguish them from missing/delayed bot results.
+non_member_label: PR-Author-Outside-NVIDIA-Org
+# Fail closed if the bot cannot determine organization membership cleanly.
+error_label: PR-Author-Org-Check-Failure
--- a/.github/workflows/restricted-paths-guard.yml
+++ b/.github/workflows/restricted-paths-guard.yml
@@ -5,8 +5,9 @@ name: "CI: Restricted Paths Guard"
 
 on:
   # Run on drafts too so maintainers get early awareness on WIP PRs.
-  # Label updates on fork PRs require pull_request_target permissions.
-  pull_request_target:
+  # Temporary testing mode: use pull_request so this branch's workflow
+  # definition can be exercised before switching back to pull_request_target.
+  pull_request:
     types:
       - opened
       - synchronize
--- a/.github/workflows/restricted-paths-guard.yml
+++ b/.github/workflows/restricted-paths-guard.yml
@@ -12,6 +12,8 @@ on:
       - synchronize
       - reopened
       - ready_for_review
+      - labeled
+      - unlabeled
 
 jobs:
   restricted-paths-guard:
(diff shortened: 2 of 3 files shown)


rwgk:
Closing: github-apps PR 78 did not receive direct feedback, an alternative PR 79 is on track to provide a different solution.

PR #1905, opened 2026-04-14
Fix IPC memory pool leak in nightly repeated test runs

cpcloud:
## Summary

- IPC test fixtures and test methods create `DeviceMemoryResource`/`PinnedMemoryResource` instances that are never closed
- Under single-run CI (`--count=1`) this is harmless, but nightly scheduled CI runs with `pytest-repeat --count=100` (introduced in PR #1574), accumulating ~200+ leaked CUDA mempools and POSIX fd exports per test file
- Around iteration 83, pool creation segfaults inside `ipc_memory_resource` fixture at `conftest.py:179` — this is why **every** nightly test matrix job has been red since 2026-02-05
- Convert `ipc_memory_resource` fixture from `return` to `yield` + `mr.close()` teardown
- Close locally-created memory resources in `test_errors`, `test_send_buffers`, and `test_workerpool` after child processes join

## Test plan

- [ ] Nightly CI scheduled run passes (all matrix cells that previously segfaulted in `memory_ipc/test_errors.py::TestDanglingBuffer`)
- [ ] Push CI passes (single-run, verifies no behavioral regression from adding `close()` calls)

### DIFF
--- a/cuda_core/tests/conftest.py
+++ b/cuda_core/tests/conftest.py
@@ -183,7 +183,8 @@ def ipc_memory_resource(request, ipc_device):
         mr = PinnedMemoryResource(options=options)
 
     assert mr.is_ipc_enabled
-    return mr
+    yield mr
+    mr.close()
 
 
 @pytest.fixture
--- a/cuda_core/tests/memory_ipc/test_errors.py
+++ b/cuda_core/tests/memory_ipc/test_errors.py
@@ -26,6 +26,7 @@ class ChildErrorHarness:
         # from PARENT_ACTION.
         self.device = ipc_device
         self.mr = ipc_memory_resource
+        self._extra_mrs = []
 
         # Start a child process to generate error info.
         pipe = [multiprocessing.Queue() for _ in range(2)]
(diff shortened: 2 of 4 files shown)


rwgk (approved):
I reviewed with Cursor GPT-5.4 Extra High Fast

The only finding is the one below. Not sure it's worth the trouble: are we repeating the un-happy path?

- Low: the new explicit mr.close() calls at cuda_core/tests/memory_ipc/test_errors.py:47, cuda_core/tests/memory_ipc/test_send_buffers.py:48, cuda_core/tests/memory_ipc/test_workerpool.py:45, cuda_core/tests/memory_ipc/test_workerpool.py:87, and cuda_core/tests/memory_ipc/test_workerpool.py:131 only run on the happy path. If a Queue.get(), join(), or verification assert fails before those tails, that attempt still leaks the same IPC mempool/FD resources this PR is trying to clean up. I’d prefer those bodies wrapped in try/finally or the extra MRs owned by fixtures/context managers so cleanup still happens on failing attempts.

PR #1908, opened 2026-04-14
CI: Detect changed paths and gate cuda.bindings tests

### LINKED ISSUE #299: CI: Detect if build/test is needed
Not all PRs would require full build/test pipelines (as we currently do today). For example, if a PR only touches code in `cuda.core`, then 
1. we don't need to rebuild `cuda.bindings`
   -  The build artifacts can be reused from the latest main branch
3. we don't need to re-test `cuda.bindings`

We should add detection logics to inform the CI workflows so that we can save CI time (and thereby reducing the costs). The same logic can be applied to nightly tests too (#294).

cpcloud:
## Summary

- Adds a `detect-changes` job to `ci.yml` using native `git merge-base` + `git diff --name-only` to classify which top-level modules changed in a PR
- Composes gating outputs respecting the dependency graph: `cuda_pathfinder → cuda_bindings → cuda_core`
- Threads `skip-bindings-test` input through `test-wheel-linux.yml` and `test-wheel-windows.yml` so cuda.bindings tests are skipped when only unrelated modules changed
- Splits the overloaded `SKIP_CUDA_BINDINGS_TEST` env var into two orthogonal flags: `USE_BACKPORT_BINDINGS` (artifact fetch for CTK major mismatch) and `SKIP_CUDA_BINDINGS_TEST` (test-time gate)
- Non-PR events (push to main, tag, schedule) unconditionally run everything
- Build-side gating outputs are emitted but not yet consumed (follow-up PR)

Closes #299

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -379,6 +379,7 @@ jobs:
     runs-on: ubuntu-latest
     needs:
       - should-skip
+      - detect-changes
       - test-linux-64
       - test-linux-aarch64
       - test-windows
@@ -405,7 +406,16 @@ jobs:
           #
           # Note: When [doc-only] is in PR title, test jobs are intentionally
           # skipped and should not cause failure.
+          #
+          # detect-changes gates whether heavy test matrices run at all; if it
+          # does not succeed, downstream test jobs are skipped rather than
+          # failed, which would otherwise go unnoticed here. Require its
+          # success explicitly so a broken gating step cannot masquerade as a
+          # green CI run.
           doc_only=${{ needs.should-skip.outputs.doc-only }}
+          if ${{ needs.detect-changes.result != 'success' }}; then
+            exit 1
+          fi
           if ${{ needs.doc.result == 'cancelled' || needs.doc.result == 'failure' }}; then
             exit 1
           fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,38 +105,31 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+        with:
+          fetch-depth: 0
 
-      - name: Filter changed paths
+      - name: Detect changed paths
         id: filter
         if: ${{ startsWith(github.ref_name, 'pull-request/') }}
-        uses: dorny/paths-filter@d1c1ffe0248fe513906c8e24db8ea791d46f8590  # v3.0.3
-        with:
-          # dorny/paths-filter v3 on push to a non-default branch defaults to
-          # diffing against the repository default branch. No explicit base is
-          # needed; omitting it lets backport PRs targeting release branches
-          # diff against their actual target rather than always against main.
-          filters: |
-            bindings:
-              - 'cuda_bindings/**'
-            core:
-              - 'cuda_core/**'
-            pathfinder:
-              - 'cuda_pathfinder/**'
-            python_meta:
-              - 'cuda_python/**'
-            test_helpers:
-              - 'cuda_python_test_helpers/**'
-            shared:
-              - '.github/**'
-              - 'ci/**'
-              - 'scripts/**'
-              - 'toolshed/**'
-              - 'conftest.py'
-              - 'pyproject.toml'
-              - 'pixi.toml'
-              - 'pixi.lock'
-              - 'pytest.ini'
-              - 'ruff.toml'
+        run: |
+          # Diff against the merge base with the default branch. Uses
+          # merge-base so diverged branches only show files changed on
+          # the PR side, not upstream commits.
+          base=$(git merge-base HEAD origin/main)
+          changed=$(git diff --name-only "$base"...HEAD)
+
+          has_match() {
+            echo "$changed" | grep -qE "$1" && echo true || echo false
+          }
+
+          {
+            echo "bindings=$(has_match '^cuda_bindings/')"
+            echo "core=$(has_match '^cuda_core/')"
+            echo "pathfinder=$(has_match '^cuda_pathfinder/')"
+            echo "python_meta=$(has_match '^cuda_python/')"
+            echo "test_helpers=$(has_match '^cuda_python_test_helpers/')"
+            echo "shared=$(has_match '^(\.github/|ci/|scripts/|toolshed/|conftest\.py$|pyproject\.toml$|pixi\.(toml|lock)$|pytest\.ini$|ruff\.toml$)')"
+          } >> "$GITHUB_OUTPUT"
 
       - name: Compose gating outputs
         id: compose
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,7 +119,7 @@ jobs:
           changed=$(git diff --name-only "$base"...HEAD)
 
           has_match() {
-            echo "$changed" | grep -qE "$1" && echo true || echo false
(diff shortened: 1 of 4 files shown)


rwgk (commented):

___

Reviewed the final tree for PR 1908 at commit d2059559818249696edaec702344a297ba1f67d6

`actionlint` passed on the touched workflows, so the notes below are about CI behavior / control flow, not YAML syntax.

## 1. High: `detect-changes` can fail without clearly failing the final required status

### Evidence

- All three test jobs now depend on `detect-changes`:
  - `.github/workflows/ci.yml:284`
  - `.github/workflows/ci.yml:309`
  - `.github/workflows/ci.yml:334`
- The terminal `checks` job does not depend on `detect-changes` directly and only inspects:
  - `doc`
  - `test-linux-64`
  - `test-linux-aarch64`
  - `test-windows`
  at `.github/workflows/ci.yml:365`
- The shell logic in `checks` only treats `failure` and `cancelled` as errors:
  - `.github/workflows/ci.yml:394`
  - `.github/workflows/ci.yml:398`

### Why this matters

- If `detect-changes` fails, the downstream test jobs can become `skipped` because a required dependency did not complete successfully.
- In that state, `checks` may still exit 0 because it does not treat those `skipped` results as an error.
- Result: a broken path-detection step could suppress or reduce CI coverage without producing a clearly failing final status.

### Agent hint

- Re-check GitHub Actions `needs` semantics for `always()` plus an upstream failed prerequisite leading to downstream `skipped`.
- The most direct fix is probably one of:
  - add `detect-changes` to `checks.needs` and explicitly require `needs.detect-changes.result == 'success'`
  - or treat unexpected `skipped` test jobs as fatal when `doc_only != true`
- Preserve the intended `[doc-only]` skip behavior while making accidental skips fail closed.

## 2. Medium: diff baseline is hardcoded to `origin/main`

### Evidence

- `detect-changes` computes the merge base against `origin/main` at `.github/workflows/ci.yml:118`.
- The repo also maintains a release/backport branch in `ci/versions.yml:4`:
  - `backport_branch: "12.9.x"`

### Why this matters

- For PR 1908, whose base is `main`, this is fine.
- If the same `pull-request/*` CI path is ever used for PRs targeting `12.9.x` or another non-`main` branch, changed-path classification will be wrong.
- Wrong classification can suppress outputs such as `test_bindings`, which means under-testing on backport / release PRs.

### Agent hint

- First confirm whether `pull-request/*` branches are ever created for non-`main` PR bases in this repo.
- If yes, derive the true PR base branch before `git merge-base` instead of assuming `main`.
- Because this workflow runs on pushes to `pull-request/*`, `github.base_ref` may not be available; one workable approach is to resolve the PR number from `github.ref_name` and query `gh pr view --json baseRefName`.
- If non-`main` PR bases are impossible by construction, add a short comment documenting that invariant so future readers do not "fix" it incorrectly.
- This area already had churn within the PR history, so the next agent should validate the actual branch-generation semantics before changing behavior.

## Bottom line

- I did not find a concrete bug in the `USE_BACKPORT_BINDINGS` / `SKIP_CUDA_BINDINGS_TEST` split itself.
- The two findings above are the only items that looked important enough to hand back to the PR author.

rwgk:
Cursor GPT-5.4 Extra High Fast generated:

___

Reviewed the current final tree for PR 1908 at commit c11f901a0c.

## Findings

### 1. Medium: `detect-changes` can under-classify cross-package renames

#### Evidence

- `.github/workflows/ci.yml:134` builds the changed-path list with:
  - `git diff --name-only "$base"...HEAD`
- `.github/workflows/ci.yml:141` through `.github/workflows/ci.yml:146` classify changed modules only from those reported paths.
- The resulting `test_bindings` output then drives:
  - `.github/workflows/ci.yml:308`
  - `.github/workflows/ci.yml:333`
  - `.github/workflows/ci.yml:358`

#### Why this matters

- In a pure rename or move, `git diff --name-only` can report only the destination path.
- That means a refactor like `cuda_bindings/... -> cuda_core/...` can produce `bindings=false` even though the bindings package lost code.
- In that case, `skip-bindings-test` may become true and reduce CI coverage for the PR.
- The same shape of issue can apply to moves across `shared` prefixes.

#### Agent hints

- The simplest fail-closed fix is probably:
  - use `git diff --no-renames --name-only "$base"...HEAD`
  - this reports both old and new paths for renames/moves, which is better for path-based gating
- A more explicit alternative is:
  - use `git diff --name-status -M "$base"...HEAD`
  - when status is `R*`, inspect both the old path and the new path
- If the PR author prefers to keep the current implementation for now, at least add a short comment documenting the rename blind spot so future readers do not assume the gating is exact.
- A focused regression check would be valuable here, even if it is just a small shell-based test or script fixture for the `detect-changes` truth table.

## Notes

- The two earlier findings from my first review look addressed in the current branch.
- Base-branch resolution now uses PR metadata via `nv-gha-runners/get-pr-info` in `.github/workflows/ci.yml:115` through `.github/workflows/ci.yml:124`, which looks appropriate for `pull-request/<N>` refs that may target non-`main` branches.
- The final `checks` job now includes `detect-changes` in `needs` at `.github/workflows/ci.yml:382` and explicitly requires it to succeed at `.github/workflows/ci.yml:416`, so a broken gating step should no longer hide behind skipped downstream jobs.
- I did not find a new issue in the `USE_BACKPORT_BINDINGS` / `SKIP_CUDA_BINDINGS_TEST` split; that part still looks coherent.
- Residual caution: this is still a static CI review. I do not see a dedicated automated check for the `detect-changes` truth table, so future refactors could regress rename/path handling fairly easily.

## Bottom line

- I see one remaining medium-confidence CI edge case around rename handling.
- Otherwise, the previously reported issues appear fixed.

rwgk (approved):
I think this is fine to merge-as-is and work on the

`detect-changes` can under-classify cross-package renames

finding in a follow-on PR. I'm fine either way.

cpcloud:
Thanks @rwgk — both findings addressed:

**Finding 1 (High) — `detect-changes` could silently cascade into skipped tests.** Added `detect-changes` to `checks.needs` with an explicit `needs.detect-changes.result == 'success'` guard, so a gating-step failure now fails the final status rather than producing a green run with reduced coverage. Commit [`c11f901`](https://github.com/NVIDIA/cuda-python/pull/1908/commits/c11f901a0c043302813070a6742772deb86dfc21). The legitimate `[doc-only]` skip path is unaffected (detect-changes stays `success` in that case).

**Finding 2 (Medium) — diff baseline hardcoded to `origin/main`.** Confirmed via `copy-pr-bot` config that `pull-request/<N>` branches are populated regardless of the PR's base (including `12.9.x`), so the hardcode was indeed wrong for backport PRs. Resolved the real base at runtime via `nv-gha-runners/get-pr-info@main` (already used elsewhere in this repo) and pass it to `git merge-base`. Fails loudly if the base can't be resolved rather than silently falling back. Commit [`f410ba6`](https://github.com/NVIDIA/cuda-python/pull/1908/commits/f410ba657fa9ba19e63875440abfd4c7fad13c97).

leofang:
Thanks, @cpcloud @rwgk! I noticed this fails on the main branch, could you take a look?
https://github.com/NVIDIA/cuda-python/actions/runs/24566662170/job/71828554003

rwgk:
It's failing in the last line here:

```
      - name: Detect changed paths
        id: filter
        if: ${{ startsWith(github.ref_name, 'pull-request/') }}
        env:
          BASE_REF: ${{ fromJSON(steps.pr-info.outputs.pr-info).base.ref }}
```

Maybe the `env` is evaluated even if the `if` evaluates to false?

PR #1910, opened 2026-04-14
Make Linker.backend a classmethod

### LINKED ISSUE #714: [FEA]: Make `Linker.backend` a `classmethod`
### Is this a duplicate?

### Area

cuda.core

`Linker` fully abstracts the linking details away from the user, but in certain cases the user might need to query the backend. For example, a user linking static code might prefer to pass `ltoir` rather than `ptx` if they know `nvjitlink` is available. `Linker.backend` exposes this property but requires instantiating a `Linker` instance to query. We should make it possible to get the info without doing so. 

I'd like `Linker.backend` to return `driver` or `nvjitlink` without instantiating the class

Instantiating a LinkerOptions in a function that catches the warning

### Additional context

https://github.com/NVIDIA/numba-cuda/pull/133

cpcloud:
## Summary

- Converts `Linker.backend` from an instance property to a classmethod so callers can query the linking backend without constructing a `Linker`
- Classmethod calls `_decide_nvjitlink_or_driver()` (existing memoised probe) and maps the return value to `"nvJitLink"` or `"driver"`
- Updates the one in-repo call site in `_program.pyx` and existing test assertion to use parens
- Adds 6 GPU-free tests via monkeypatch (no `Device()` / `cuInit` required)
- Adds `0.8.0-notes.rst` documenting the breaking change

**Breaking change:** `linker.backend` (attribute access) now returns a bound method, not a string. All call sites must use `Linker.backend()`. Only 2 in-repo call sites affected; numba-cuda (the requester) will adopt `Linker.backend()` from day one.

## Test plan

- [x] 6 GPU-free tests pass locally (monkeypatch on module globals)
- [x] CI: existing GPU test `test_linker_init` passes with updated paren call
- [x] CI: Cython compile succeeds

Closes #714


cuda_core/docs/source/release/1.0.0-notes.rst
leofang, line 1: I believe we are shooting for 1.0.0, unless @mdboom concludes that a 0.8.0 is a must for RAPIDS adoption.

cuda_core/tests/test_linker_backend.py
leofang, line 1: nit: let's move the tests to the existing `test_linker.py`. There is no need for separate test files.

cuda_core/tests/test_linker_backend.py
```diff
@@ -0,0 +1,57 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""GPU-free tests for Linker.backend classmethod.
+
+These live in a separate file from test_linker.py because that module calls
+Device() at import time, which requires a GPU. These tests use monkeypatch
+to set module-level flags and never touch CUDA devices.
```
leofang: Does this make sense? Would the driver `cuLink*` APIs be functional when there is no GPU present? I don't think this is ever tested. 

NVRTC/nvJitLink work without GPU or CUDA, but it is not something we want to guarantee to users without the compiler team onboarding. This is not documented AFAIK.

cuda_core/docs/source/release/1.0.0-notes.rst
```diff
@@ -0,0 +1,17 @@
...
+
+
+Breaking Changes
+----------------
+
+- :meth:`Linker.backend` is now a classmethod instead of an instance property.
+  Call sites must use ``Linker.backend()`` (with parentheses) instead of
+  ``linker.backend``. This allows querying the linking backend without
+  constructing a ``Linker`` instance — for example, to choose between PTX and
+  LTOIR input before linking.
```
leofang: 👍

leofang (changes requested):
Thanks, Phillip! Implementation-wise LGTM. Left a few suggestions.

cuda_core/docs/source/release/1.0.0-notes.rst
cpcloud: Renamed to `1.0.0-notes.rst`.

cuda_core/tests/test_linker_backend.py
cpcloud: Moved into `test_linker.py` and deleted `test_linker_backend.py`.

cuda_core/tests/test_linker_backend.py
cpcloud: Good point — `test_linker.py` already calls `Device()` at module import (line 11), so the GPU-free framing was moot. Dropped the rationale and merged the tests in; they just sit alongside the existing ones now.

rparolin (approved):

leofang (commented):
Given that we just recently went to the opposite direction (converting some methods to properties, #1945 & #1986), it'd be better to have @mdboom or @Andy-Jost to chime in from the design consistency perspective. It might be possible that we need to re-evaluate if #714 is really necessary.

cuda_core/cuda/core/_linker.pyx
```diff
@@ -167,10 +167,22 @@ cdef class Linker:
...
+
+        .. note::
+
+            Prefer letting :class:`Linker` decide. Query ``backend()``
+            only when you need to dispatch based on input format (for
+            example: choose PTX vs. LTOIR before constructing a
+            ``Linker``). The returned string names an implementation
+            detail whose support matrix may shift across CTK releases.
+        """
+        return "driver" if _decide_nvjitlink_or_driver() else "nvJitLink"
```
Andy-Jost: We did just complete a sweep to standardize the property/method distinction. It would be great to keep that as clean as possible.

One option is to roll our own classproperty:

```python
class classproperty:
    def __init__(self, func):
        self.func = func
    def __get__(self, instance, owner):
        return self.func(owner)

class Linker:
    @classproperty
    def backend(cls):
        . . .
```

Related: python/cpython#89519

There are caveats and hazards in complex cases, but they wouldn't come into play here.

Another option would be to keep the property while adding a module-level function `get_backend()`.

Is there any risk that the backend selection might become dependent on constructor args? If so, it would be better to use a freestanding function IMO.

leofang:
We discussed this PR in the team sync today, but I forgot the conclusion... Maybe Mike or Ralf remembers? 😅

cuda_core/cuda/core/_linker.pyx
```diff
@@ -167,10 +167,22 @@ cdef class Linker:
         else:
             return as_py(self._culink_handle)
 
-    @property
-    def backend(self) -> str:
-        """Return this Linker instance's underlying backend."""
-        return "nvJitLink" if self._use_nvjitlink else "driver"
+    @classmethod
+    def backend(cls) -> str:
```
rwgk: > We discussed this PR in the team sync today, but I forgot the conclusion... Maybe Mike or Ralf remembers? 😅

In the meeting there was a discussion about making this a class property. It sounded like that's not entirely straightforward. I just asked here:

https://chatgpt.com/share/69f9637b-9280-83ea-ae35-12c7a18f9337

It seems indeed not straightforward.

My pitch in the meeting was: let's keep this simple and name this `@classmethod` `which_backend` or `get_backend` (I'd pick `which_backend` because it seems more intuitive to me; like `bash` `which cat`).

Apart from this pure naming question, this PR looks good and final to me.

leofang: Done in commit 792d99f.

cuda_core/cuda/core/_linker.pyx
leofang: Resolving, see the decision below: https://github.com/NVIDIA/cuda-python/pull/1910#discussion_r3185883468

leofang (approved):

PR #1911, opened 2026-04-14
Rework Linker dispatching for cross-major nvJitLink/driver skew

### LINKED ISSUE #712: Re-work on `Linker` dispatching logic
CUDA 12.0 is the first release where nvJitLink debuts. So far we have been operating under the assumption that nvJitLink is always available and can always generate a PTX (or CUBIN) that the driver can load. However, this assumption partially relies on another assumption that both nvJitLink and the driver are within the same major version 12.x.

By the time the next major release is out, there exists a possibility that the driver version (13.x) is newer than nvJitLink version (12.x). In this case, unless we need to link LTO IRs (which the driver cannot handle) we should prefer for the driver to handle the linking, instead of using nvJitLink. For linking LTO IRs, we must raise an exception. 

We should also raise an exception if the user tries to link LTO IRs from different major versions. I believe it is not a supported use case. However, we currently do not have any introspection capability to check the LTO IR version.

cpcloud:
## Summary

Reworks the `Linker` backend selection and fixes three latent driver-linker resource-lifetime bugs that the rework surfaced on cross-major CTK + driver CI runners.

### Per-instance dispatch (`refactor(core): per-instance Linker backend dispatch`)

- Replaces module-level "decide once" backend selection with per-`Linker`-instance dispatch at `__init__` time
- Factors the decision into a pure `_choose_backend()` helper for GPU-free unit testing
- Handles nvJitLink/driver major-version mismatches: falls back to the driver linker for non-LTO linking, raises `RuntimeError` for LTO when the backends are incompatible
- Probes `driver_version()` lazily; only `CUDAError` from `cuDriverGetVersion` is treated as "driver unknown" (other exceptions propagate)
- `_probe_nvjitlink()` is cached and warns at most once when nvJitLink is absent
- Validates every input is an `ObjectCode` before the `code_type` pre-scan so malformed inputs raise `TypeError` instead of a dispatch `RuntimeError`

**Breaking change:** `options.link_time_optimization=True` with nvJitLink absent now raises `RuntimeError` instead of silently passing `CU_JIT_LTO` to the driver (which was not real LTO linking).

#### Decision matrix

| driver | nvJitLink | ltoir input | lto/ptx | result |
|--------|-----------|-------------|---------|--------|
| any | None | no | no | driver |
| any | None | yes/lto | — | raise |
| M | (M,\*) | any | any | nvJitLink |
| D≠N | (N,\*) | no | no | driver fallback |
| D≠N | (N,\*) | yes/lto | — | raise |
| None | available | any | any | nvJitLink |

### Driver-linker lifetime fixes

With per-instance dispatch, CTK 12.9 + driver 13.0 runners now hit the driver linker for cross-major linking, which was never exercised before. That exposed three independent latent bugs:

- **`fix(core): retain driver log buffers for CUlinkState lifetime`** — `Linker_link` was nulling `_drv_log_bufs` right after `cuLinkComplete`, leaving `cuLinkDestroy` (run later in `tp_dealloc`) with dangling pointers to the info/error log bytearrays. Retain the bytearrays until the `cdef class` is deallocated; `.pxd` declaration order ensures `_culink_handle` is destroyed before the buffers.
- **`fix(core): retain cuLinkCreate optionValues array for CUlinkState lifetime`** — `c_jit_keys`/`c_jit_values` were stack-local `vector`s destroyed on `Linker_init` return, but CUDA requires `optionValues` to stay valid for the life of the `CUlinkState` (the driver writes log-fill sizes back into its slots). Promote them to `cdef class` fields declared after `_culink_handle` so `tp_dealloc` destroys them after `cuLinkDestroy` runs.
- **`fix(core): release driver-backend buffers in Linker.close()`** — `close()` now caches decoded logs into `_info_log`/`_error_log` before dropping `_drv_log_bufs`, so `get_error_log()`/`get_info_log()` still work after `close()` on both success and failed-link paths. The option vectors are swapped with empty locals (not `.clear()`ed) so the backing allocation is actually freed rather than retaining capacity.

## Test plan

- [x] GPU-free parameterized tests for the full `_choose_backend()` decision matrix, including the `driver_major=None` (build-container) path (`tests/test_linker_dispatch.py`)
- [x] Test helpers handle driver-version failure gracefully
- [x] Driver-backend regression test: force the driver backend, link + close + drop a `Linker`, then do a full compile + link cycle (`test_driver_linker_lifetime_no_heap_corruption`)
- [x] Failed-link + `close()` → `get_error_log()` regression test (`test_driver_linker_get_error_log_after_close_on_failed_link`)
- [x] CI: existing GPU tests pass with per-instance dispatch
- [x] CI: cross-major behavior verified (CTK 12.9.1 + driver 13.0 runners exercise the driver-fallback path that surfaced the lifetime bugs)

Closes #712


cuda_core/cuda/core/_linker.pxd
```diff
@@ -2,15 +2,27 @@
...
 cdef class Linker:
     cdef:
         NvJitLinkHandle _nvjitlink_handle
         CuLinkHandle _culink_handle
+        # _drv_jit_keys/_drv_jit_values are the C arrays handed to cuLinkCreate.
+        # The driver retains a reference to the optionValues array for the life
+        # of the CUlinkState (it writes back log-size outputs into its slots),
+        # so these must live past cuLinkCreate and outlive cuLinkDestroy.
+        # Declared after _culink_handle so their C++ destructors run AFTER
+        # cuLinkDestroy executes during tp_dealloc.
```
rparolin: Its easy to miss the subtly here that reordering the fields breaks the assumptions of object lifetimes and can cause memory issues.  Since I don't have any better ideas on how to guard against this... I'd make it VERY obvious that people should not be reordering the members here unless they know what they are doing.  Something to the effect...

`# WARNING: Here be dragons... do not reorder data members... <reasons>`

rparolin:
```
python
def close(self):
    if self._culink_handle is not None:
        cuLinkDestroy(self._culink_handle)
        self._culink_handle = None
    self._drv_jit_keys.clear()
    self._drv_jit_values.clear()

def __dealloc__(self):
    self.close()  # ensures correct order even without 'with' block
```

I'm not convinced this is much better... but I offer it as food for thought. We could specify the correct destruction order of the members instead in a member function and manually called that when the object is destroyed by the GC.  The best I can say about this is it less subtle that we are attempting to handle complext lifetime issues.

rparolin: cc: @leofang  @mdboom

cuda_core/cuda/core/_linker.pyx
```diff
@@ -170,21 +191,34 @@ cdef class Linker:
...
+            driver_major = driver_version()[0]
+        except CUDAError:
+            driver_major = None
+        backend = _choose_backend(
+            driver_major,
+            _probe_nvjitlink(),
+            inputs_have_ltoir=False,
+            lto_requested=False,
+        )
+        return CompilerBackendType.NVJITLINK if backend == "nvjitlink" else CompilerBackendType.DRIVER
```
isVoid, line 232: The previous backend decision tree doesn't consider LTO input or request LTO output (non-per instance). In the context of this PR, this function has gone stale, the better alternative is to directly query `.backend` attribute. But since it's a public API, we should mark this as deprecation first. WDYT?

cuda_core/cuda/core/_linker.pxd
isVoid: I like the idea. Marking teardown explicit is better than relying on implicit declaration order of cython compiler behavior.

One thing is that `__dealloc__` has strong assumptions on what can happen inside (no-raise, no fancy memory operations). `close()` method is a python public facing method which already needs to handle log buffering etc. So I'd suggest the following split:

```
cdef _close_noexcept(self) noexcept:
    # Close handle, clear drv_jit_keys and values

def close(self):
    # log work
    self._close_noexcept()

def __dealloc__(self):
    self._close_noexcept()
```

Notice that `_close_noexcept` should leave the handle with an empty shared_ptr, the vectors all with size 0. `__dealloc__` serves as a backstop regardless whether user explicitly calls `close()`.

isVoid: [0ca1aff](https://github.com/NVIDIA/cuda-python/pull/1911/commits/0ca1affc46c99f7e83e1309ca6711ec5ea41f8f1)

leofang:
Superseded by #2320.

PR #1912, opened 2026-04-14
Add persistent program cache for Program.compile

### LINKED ISSUE #176: Implement compiler caches
We will add opt-in cache classes for library developers and end users to cache the compiler artifacts.

### LINKED ISSUE #177: Add `cuda.core.utils.InMemoryProgramCache`

### LINKED ISSUE #178: Add `cuda.core.utils.PersistentProgramCache`
Two kinds of on-disk caches:
- one for sqlite3 based (suitable for single processes)
- one for file-stream based (suitable for concurrent processes, thanks to atomic I/O)

It is unclear to me (yet) if they can be easily combined under the same class `PersistentProgramCache`. We'll need some research.

### LINKED ISSUE #179: Add `cuda.core.utils.ProgramCacheResource`
This is the ABC class that all concrete cache classes should inherit/overwrite:
```python
class ProgramCacheResource(abc.ABC):

    @abc.abstractmethod
    def __getitem__(self, key: Any) -> Module:
        """Retrieve the cached Module instance. key should be a hashable
        object.
        """
        ...

    @abc.abstractmethod
    def __setitem__(self, key: Any, val: Module):
        """Cache a Module instance. key should be a hashable object.
        """
        ...
```

cpcloud:
## Summary

Closes #176.
Closes #177.
Closes #178.
Closes #179.

Adds a persistent on-disk cache for `cuda.core.Program.compile` outputs. The high-level integration is one keyword on `Program.compile`:

```python
from cuda.core import Program, ProgramOptions
from cuda.core.utils import FileStreamProgramCache

source = 'extern "C" __global__ void k(int *a){ *a = 1; }'
options = ProgramOptions(arch="sm_80")

with FileStreamProgramCache() as cache:  # default: $XDG_CACHE_HOME/cuda-python/program-cache
    obj = Program(source, "c++", options=options).compile("cubin", cache=cache)
    obj.get_kernel("k")
```

A second invocation with the same inputs short-circuits the entire NVRTC compile — `cache.get(key)` (one stat + one read) and an `ObjectCode._init` from the bytes. No `Program_compile` is invoked. This is the fast path the cache exists to provide:

```python
# Fresh process / second run -- same source, same options.
with FileStreamProgramCache() as cache:
    obj = Program(source, "c++", options=options).compile("cubin", cache=cache)
    # ~10us round-trip on a warm page cache, vs hundreds of ms to seconds
    # for an actual NVRTC invocation.
```

## Public API

- `Program.compile(target_type, *, cache=...)` — convenience wrapper. Derives the key, returns a fresh `ObjectCode` on hit, stores the compile output on miss.
- `cuda.core.utils.ProgramCacheResource` — abstract bytes-in / bytes-out interface for custom backends. Provides `get`, `update` (Mapping or pairs), `clear`, and the mapping mutators (`__getitem__`/`__setitem__`/`__delitem__`/`__len__`). `__contains__` is intentionally omitted: `cache.get(key)` is the recommended idiom because the two-call `if key in cache: cache[key]` pattern is racy across processes.
- `cuda.core.utils.InMemoryProgramCache` — single-process LRU on `OrderedDict`, `threading.RLock`, size-only cap. For "compile once, look up many" workflows that don't need persistence.
- `cuda.core.utils.FileStreamProgramCache` — directory of atomic per-entry files. Safe across processes via `os.replace` + Windows sharing-violation retries on `os.replace` / read / `unlink`.
- `cuda.core.utils.make_program_cache_key` — escape hatch when the compile inputs require an `extra_digest` (`include_path`, `pre_include`, `pch`, `use_pch`, `pch_dir`, NVVM `use_libdevice=True`, NVRTC `options.name` with a directory component). `Program.compile(cache=...)` rejects those compiles with a `ValueError` pointing here.

## On-disk format

Each entry is the raw compiled binary verbatim — cubin / PTX / LTO-IR — with no pickle, JSON, length prefix, or framing of any kind. Cache files are directly consumable by external NVIDIA tools (`cuobjdump`, `nvdisasm`, `cuda-gdb`).

`ObjectCode.symbol_mapping` from `name_expressions` is not preserved across a cache round-trip; the wrapper rejects `Program.compile(name_expressions=..., cache=...)` outright so the first-call-works/second-call-breaks footgun can't surface. Callers that need `get_kernel(name_expression)` should compile without `cache=`.

## FileStreamProgramCache

- **Atomic writes**: stage to `tmp/`, fsync, `os.replace` into `entries/<2char>/<hash>`. Concurrent readers never observe partial writes. Windows `os.replace` retries on `ERROR_ACCESS_DENIED` / `ERROR_SHARING_VIOLATION` / `ERROR_LOCK_VIOLATION` (winerrors 5/32/33) within a bounded backoff (~185 ms); after the budget, the write is dropped and the next call recompiles. The same retry covers reads and `path.unlink` so eviction doesn't crash the writer that triggered it on win-64.
- **Sharing-violation predicate**: `_is_windows_sharing_violation(exc)` filters `EACCES` only when `winerror` is absent — non-sharing winerrors are real config errors and propagate. Off-Windows `PermissionError` always propagates.
- **Transparent input forms**: `cache[key] = value` (and `cache.update({key: value, ...})`) accept raw bytes, bytearray, memoryview, or any `ObjectCode` (path-backed too — the file is read at write time so the cached entry is the binary content, not a path that could move). Reads return the same bytes that went in.
- **Size-only bound**: `max_size_bytes` is the only knob — no element-count cap. `None` means unbounded.
- **True LRU via atime**: every successful read calls `os.utime` (fd-based on Linux/macOS via `os.supports_fd`, path-based on Windows) to bump `st_atime` regardless of mount options or `NtfsDisableLastAccessUpdate`. Eviction sorts by oldest `st_atime` first. The atime touch is stat-guarded so a racing rewriter's freshly-replaced file never has its mtime rolled back.
- **Stat-guarded prunes**: `clear()`, `_enforce_size_cap()`, and the atime touch all snapshot `(ino, size, mtime_ns)` per entry and refuse to unlink / overwrite stamps if a writer replaced the file mid-operation.
- **Cache key derivation** (`make_program_cache_key`): a backend-strategy pattern with one class per `code_type` (`_NvrtcBackend` / `_LinkerBackend` / `_NvvmBackend`). Each owns its own validate / encode_code / option_fingerprint / encode_name_expressions / hash_version_probe / hash_extra_payload. The orchestrator validates `code_type`/`target_type`, dispatches to the right backend, and assembles the digest in fixed order. Adding a new backend is one new class, not a five-place edit.
- **NVRTC `options.name` with a directory component**: rejected without `extra_digest` because NVRTC resolves quoted `#include` directives relative to that directory — neighbour-header changes wouldn't invalidate the cache otherwise.
- **PTX-loadability warning on cache hit**: when the active driver can't load freshly-generated PTX, the wrapper emits the same `RuntimeWarning` the uncached path emits — loadability depends on the driver, not on whether the bytes were freshly compiled.
- **Default cache directory**: when `path` is omitted, resolves via `platformdirs.user_cache_path("cuda-python", appauthor=False, opinion=False) / "program-cache"`:
  - Linux/BSD: `\$XDG_CACHE_HOME/cuda-python/program-cache` (default `~/.cache/cuda-python/program-cache`)
  - macOS: `~/Library/Caches/cuda-python/program-cache`
  - Windows: `%LOCALAPPDATA%\\cuda-python\\program-cache`
- **`tmp/` self-heal**: if something deletes `tmp/` after the cache is opened, the next write recreates it rather than crashing with `FileNotFoundError`.
- **Crashed-writer cleanup**: stale temp files older than 1 hour are swept on open and on size-cap enforcement.

## Test plan

- `tests/test_program_cache.py` — abstract-class contract, `update` accepts mapping or pairs, transparent input-form equivalence (bytes / bytearray / memoryview / bytes-backed `ObjectCode` / path-backed `ObjectCode` all round-trip to the same on-disk bytes), `make_program_cache_key` semantics (deterministic, supported-target matrix mirrors `Program.compile`, backend probe failures fail closed but stable, env-version changes don't perturb the key on the wrong backends, options-fingerprint canonicalization for the linker path, side-effect / external-content / NVRTC `options.name`-dir-component guards, schema version mixing), filestream CRUD, atomic-write race coverage, stat-guarded prune / atime-touch / clear / size-cap, atime LRU promotes recently-read, default-dir uses `platformdirs`, `_is_windows_sharing_violation` predicate's truth table including the regression case (non-sharing winerror plus EACCES propagates), `tmp/` recreation after external wipe.
- `tests/test_program_cache_multiprocess.py` — concurrent writers same key, distinct keys, reader-vs-writer torn-file safety, size-cap eviction race (rewriter vs. churner) under stat-guarded eviction.
- `tests/test_program_compile_cache.py` — `Program.compile(cache=...)` miss/hit/error paths against a recording stub, `name_expressions` rejection, `extra_digest`-required / side-effect / NVRTC `options.name`-dir-component rejection, PTX loadability warning on cache hit (positive + negative), real-NVRTC end-to-end roundtrip across reopen.


rwgk:
Generated with the help of Cursor GPT-5.4 Extra High Fast

___


## High: `make_program_cache_key()` misses implicit source-directory header dependencies

`make_program_cache_key()` only forces `extra_digest` for explicit include/PCH options in `cuda_core/cuda/core/utils/_program_cache.py:393` and `cuda_core/cuda/core/utils/_program_cache.py:592`, but NVRTC also implicitly searches the source file's directory unless `no_source_include` is set in `cuda_core/cuda/core/_program.pyx:1001`.

`Program` passes `options.name` straight to `nvrtcCreateProgram()` in `cuda_core/cuda/core/_program.pyx:635`, while the key builder only hashes that path string in `cuda_core/cuda/core/utils/_program_cache.py:778`. That means a workflow like `options.name="/path/to/kernel.cu"` plus `#include "local.h"` can reuse a stale cached `ObjectCode` after `local.h` changes.

The new tests cover explicit include/PCH knobs, but not this default source-directory include path (`cuda_core/tests/test_program_cache.py:765`).

## Medium: `FileStreamProgramCache._enforce_size_cap()` can over-evict under concurrent capped writers

After the re-stat at `cuda_core/cuda/core/utils/_program_cache.py:1515`, a concurrent deleter can remove the candidate before `path.unlink()`. That `FileNotFoundError` is suppressed at `cuda_core/cuda/core/utils/_program_cache.py:1530`, but `total` is not adjusted, so eviction continues and can delete newer entries unnecessarily.

For a backend explicitly documented for multi-process use, that turns ordinary contention into avoidable cache data loss. The current multiprocess coverage exercises concurrent writes and prune races, but not `max_size_bytes` under concurrency (`cuda_core/tests/test_program_cache_multiprocess.py`).

Reduced simulation I used locally:

```python
import time
from pathlib import Path

from cuda.core._module import ObjectCode
from cuda.core.utils import FileStreamProgramCache

cache = FileStreamProgramCache("/tmp/cuda_cache_review_race", max_size_bytes=1000)
cache[b"old"] = ObjectCode._init(b"a" * 600, "cubin", name="old")
time.sleep(0.01)

old_path = cache._path_for_key(b"old")
orig_unlink = Path.unlink
state = {"done": False}


def flaky_unlink(self, *args, **kwargs):
    if self == old_path and not state["done"]:
        state["done"] = True
        # Simulate another process deleting the file after stat() but before
        # _enforce_size_cap() updates its bookkeeping.
        orig_unlink(self, *args, **kwargs)
        raise FileNotFoundError(self)
    return orig_unlink(self, *args, **kwargs)


Path.unlink = flaky_unlink
try:
    cache[b"new"] = ObjectCode._init(b"b" * 600, "cubin", name="new")
finally:
    Path.unlink = orig_unlink

remaining = [key for key in (b"old", b"new") if cache.get(key) is not None]
print(remaining)  # []
```

That produced `[]` for me: once the first deletion race is swallowed without decrementing `total`, the loop keeps evicting and drops the fresh entry too.

## Low: `from cuda.core.utils import *` now eagerly imports the cache stack

The package conversion keeps explicit imports like `from cuda.core.utils import StridedMemoryView` lightweight, but `from cuda.core.utils import *` walks `__all__`, resolves the lazy cache symbols, and imports `_program_cache` (`cuda_core/cuda/core/utils/__init__.py:10`, `cuda_core/cuda/core/utils/__init__.py:32`).

I verified that star-import now loads `cuda.core.utils._program_cache`. That said, this only affects `import *`, which is already discouraged. I think a short comment explaining that the laziness guarantee is intended for explicit imports, not star-import, seems sufficient here.

cuda_core/cuda/core/utils/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+# StridedMemoryView`` stays lightweight -- the cache backends pull in driver,
+# NVRTC, and module-load machinery that memoryview-only consumers do not need.
+_LAZY_CACHE_ATTRS = frozenset(
+    {
+        "FileStreamProgramCache",
+        "ProgramCacheResource",
+        "SQLiteProgramCache",
+        "make_program_cache_key",
+    }
+)
```
rwgk, line 29: Small readability/maintenance cleanup suggestion:

 `__all__` and `_LAZY_CACHE_ATTRS` currently duplicate the same cache-export names, so defining the ordered lazy-export list once and reusing it in `__all__` seems a bit easier to scan and reduces the chance that the two drift apart later.

Something along these lines:

```python
_LAZY_CACHE_ATTRS = (
    "FileStreamProgramCache",
    "ProgramCacheResource",
    "SQLiteProgramCache",
    "make_program_cache_key",
)

__all__ = [
    "StridedMemoryView",
    "args_viewable_as_strided_memory",
    *_LAZY_CACHE_ATTRS,
]
```

Mostly just a readability nit, but I think this makes the relationship between "lazy exports" and "public exports" a little clearer.

leofang:
Thanks, Phillip! I have this PR in my review backlog 🙏

The most important question: Are these cache implementations multithreading/multiprocessing safe? This is the key challenge that real-world apps will stress test. In CuPy, our on-disk cache has been stress-tested in DOE supercomputers.

cuda_core/cuda/core/utils/__init__.py
cpcloud: Done in cad93d0efd. `_LAZY_CACHE_ATTRS` is now the single ordered source; `__all__` splices it in via `*_LAZY_CACHE_ATTRS` so the two lists can't drift. Same commit adds a note that the laziness guarantee is for explicit imports only -- star-import still walks `__all__` and therefore resolves every lazy attribute.

cpcloud:
Addressed in ff886d3585 (fixes) and cad93d0efd (refactor + star-import note).

**High -- source-directory include.** `make_program_cache_key()` now refuses to build an NVRTC key when `options.name` contains a directory separator and neither `extra_digest` nor `no_source_include=True` is set. Scoping the guard to names that actually introduce a new search directory (`/abs/kernel.cu`, `rel/kernel.cu`, `C:\src\kernel.cu`) keeps bare labels like `"default_program"` or `"kernel-a"` -- which fall back to CWD, the same search root every NVRTC compile sees -- accepted unchanged. Tests cover POSIX and Windows separators, the `extra_digest` and `no_source_include=True` accept paths, and confirm the guard is NVRTC-only (PTX and NVVM unaffected).

**Medium -- over-eviction race.** `FileStreamProgramCache._enforce_size_cap()` now decrements `total` whether it unlinks the candidate itself or a concurrent pruner already removed the file. The `FileNotFoundError` is still suppressed, but the accounting now matches reality, so the loop stops as soon as the cap is met. Added a test that monkeypatches `Path.unlink` to simulate a concurrent deleter winning exactly once, then verifies the freshly-committed entry survives.

**Low -- star-import.** Added a note in `cuda_core/cuda/core/utils/__init__.py` that the laziness guarantee is for explicit imports only -- `from cuda.core.utils import *` walks `__all__` and therefore resolves every lazy attribute. Star-imports are discouraged anyway, so treat that as expected.

cpcloud:
@leofang -- yes, all three backends are designed and tested for concurrent access, with different scopes:

**`InMemoryProgramCache` -- thread-safe, not process-safe.** Dict-backed (`OrderedDict`) cache that lives only in the owning process. `threading.RLock` serialises every method (`__getitem__`, `__setitem__`, `__contains__`, `__delitem__`, `__len__`, `clear`, and the internal eviction pass) so threads can share one cache object without external locking. It is **not** process-safe by design: each process has its own independent cache, there is no shared state or IPC; for multi-process sharing use `SQLiteProgramCache` or `FileStreamProgramCache`. Note that entries are stored by reference, not copied -- a thread that mutates a returned `ObjectCode` affects the cached entry, so callers must treat reads as read-only. Stressed by a threaded test with 4 writers + 4 readers x 200 ops against a cache with `max_entries` set, verifying the cache stays consistent and never exceeds the cap.

**`SQLiteProgramCache` -- thread-safe; multi-process best-effort.** `check_same_thread=False` on the connection plus a `threading.RLock` serialises every connection-touching method, so threads cannot interleave a read/update or a write/VACUUM pair. WAL + autocommit on open. Stressed by a 4 writers + 4 readers x 200 ops test in `test_program_cache.py`. Sharing the sqlite file across processes does work (sqlite3 WAL serialises writes at the file level and our size-cap/VACUUM passes run under the same WAL discipline), but the `threading.RLock` does not cross process boundaries and the `VACUUM` pass skips under active readers, so the on-disk file can temporarily grow above `max_size_bytes` until readers release. For workloads with many concurrent processes, `FileStreamProgramCache` is the better fit.

**`FileStreamProgramCache` -- thread-safe and process-safe.** Every write lands on a per-write temp file and is promoted via `os.replace`, so a reader/writer race either sees the old entry or the new one -- never a half-written file. Reader pruning, `clear()`, and `_enforce_size_cap` are all stat-guarded: before unlinking, the code re-`stat`s the candidate and refuses if `(ino, size, mtime_ns)` differs from the snapshot, so a concurrent writer's `os.replace` is preserved. Stale temp files are swept on open. On Windows, `os.replace` can surface `ERROR_SHARING_VIOLATION` (32) / `ERROR_LOCK_VIOLATION` (33) against a reader briefly holding the handle; the code retries with bounded backoff (~185ms total) before treating it as a non-fatal cache miss -- all other `PermissionError`s and POSIX failures propagate.

Cross-process coverage in `test_program_cache_multiprocess.py`:
- concurrent writers producing overlapping keys
- a writer/reader race exercising the stat-guarded prune path
- clear/eviction race injection via generator cleanup (the cleanup code after the last yield runs at `StopIteration`, which is exactly between `_enforce_size_cap`'s scan and its eviction loop)
- Windows `PermissionError` narrowing (winerror 32/33 swallow + retry, all others propagate)

One concurrency bug this review shook out (over-eviction after a suppressed `FileNotFoundError` in `_enforce_size_cap`) is fixed with its own test. If you see a DOE-style pattern from CuPy's cache that we don't cover yet, happy to add a test that reproduces it -- mapping that stress-testing onto this backend would be useful.

rwgk:
FWIW, I briefly explored "safe pickle" and "signed pickle blobs" in this chat:

* https://chatgpt.com/share/69efa9b2-83b8-83ea-8fa6-a158c1fb0088

The conclusion there is:
* Do not pickle compiled-kernel cache entries.
* Use JSON / simple binary files for metadata and artifacts.

cuda_core/pyproject.toml
```diff
@@ -50,6 +50,7 @@ classifiers = [
 dependencies = [
     "cuda-pathfinder >=1.4.2",
     "numpy",
+    "platformdirs >=3.0",
```
leofang: Please let us not introduce new dependencies.

cuda_core/pixi.toml
```diff
@@ -185,6 +185,7 @@ cuda-version = "*"
 numpy = "*"
 cuda-bindings = "*"
 cuda-pathfinder = "*"
+platformdirs = ">=3.0"
```
leofang: ditto

cuda_core/cuda/core/utils/_program_cache.py
```diff
@@ -0,0 +1,1702 @@
...
+      (Windows uses local AppData -- caches don't roam).
+
+    Delegates to :mod:`platformdirs`, which encodes the per-platform
+    rules canonically (and tracks corner cases like ``FOLDERID_*`` on
+    Windows that an ad-hoc reimplementation would miss).
+    ``opinion=False`` suppresses the extra ``Cache`` component
+    platformdirs would otherwise insert on Windows, keeping the layout
+    identical across platforms (``<root>/cuda-python/program-cache``).
+    """
+    return platformdirs.user_cache_path("cuda-python", appauthor=False, opinion=False) / "program-cache"
```
leofang: 1. Doesn't make sense to introduce a new dependency just for this niche use case. We can totally vibe this out ourselves.
2. CUDA does not support macOS for many years.

cuda_core/pixi.lock
leofang, line 1: If we drop platiformdir we won't need to update the lock here

cuda_core/cuda/core/_program.pyx
```diff
@@ -98,13 +103,99 @@ cdef class Program:
...
+        if name_expressions:
+            raise ValueError(
+                "Program.compile(cache=...) does not support name_expressions: "
+                "ObjectCode.symbol_mapping is populated by NVRTC at compile "
+                "time and is not preserved across a cache round-trip, so cache "
+                "hits would silently break get_kernel(name_expression) lookups "
+                "that the uncached path supports. Compile without cache= when "
+                "name_expressions are needed, or look up mangled symbols by "
+                "hand from the cached ObjectCode."
+            )
```
leofang, line 187: Note to self: I need to address this after 1.0 is out, xref: https://github.com/cupy/cupy/pull/9801

cuda_core/cuda/core/utils/_program_cache.py
```diff
@@ -0,0 +1,1702 @@
...
+        k = _as_key_bytes(key)
+        with self._lock:
+            try:
+                data, _size = self._entries[k]
+            except KeyError:
+                raise KeyError(key) from None
+            # Touch LRU: a real read promotes the entry to "most recent"
+            # so eviction prefers genuinely cold entries.
+            self._entries.move_to_end(k)
+            return data
```
leofang: Q: What would be our recommended way of using `InMemoryProgramCache` in a multi-GPU env? Wondering about this because we usually have each GPU driven by a thread, and if the intended use case is a global cache object (which makes sense on a homogeneous system like DGX) this would cause serialization.

In CuPy internally there is a per-device cache so this issue is avoided.

leofang (changes requested):
Thanks, Phillip! Sorry for the long wait. Sending out the first wave of my review. Will continue asap.

Note: It would be nice if we can break up the two largest files (`cuda_core/cuda/core/utils/_program_cache.py` and `cuda_core/tests/test_program_cache.py`, each are 1-2k lines) into smaller logical units.

cpcloud:
@leofang -- on the multi-GPU question. Two options worth weighing, both viable:

**Background.** `make_program_cache_key` already mixes `arch` into the digest, so on a homogeneous system (all sm_80, all sm_90) every device produces the same key for the same source -- sharing one cache amortises compiles across every GPU. The single-lock cost is real but bounded: the dict update + LRU bump runs in microseconds, and the compile itself runs outside the lock. So the contention only matters during compile-heavy startup with many threads, on heterogeneous-arch systems, or when profiling shows it.

**Option A -- document the dict-of-caches pattern, no API change.**
```python
caches = {d.device_id: InMemoryProgramCache() for d in devices}
# per thread:
program.compile(\"cubin\", cache=caches[Device().device_id])
```
- Pros: zero API surface added; \`Program.compile(cache=...)\` is duck-typed today, so callers can also wrap the dict in a thin router themselves.
- Cons: callers thread the dict through every compile site; easy to forget the device-id key.

**Option B -- ship a \`PerDeviceProgramCache\` routing wrapper.**
- Pros: one-liner for callers (\`cache = PerDeviceProgramCache()\`); routing on \`Device().device_id\` lives inside the class.
- Cons: couples the cache layer to \`Device.current()\` (every \`__getitem__\` reaches into the device API); more API surface to lock down before 1.0; harder to mock in tests.

I lean A for the first cut: the common case (homogeneous DGX, single SKU) is correctly served by one shared cache and benefits from cross-device amortisation, the heterogeneous-arch case is a 3-line dict away, and B can ship post-1.0 if real workloads make it pattern enough to deserve a class. Happy to pivot to B if you'd rather have it on the public surface from day one.

cpcloud:
Pushed d177450e76 addressing the first review wave:

**platformdirs dropped** (`cuda_core/pyproject.toml`, `cuda_core/pixi.toml`, `cuda_core/cuda/core/utils/_program_cache/_file_stream.py`). `_default_cache_dir` is now ~10 lines: \`$XDG_CACHE_HOME or ~/.cache\` on Linux, \`%LOCALAPPDATA% or ~/AppData/Local\` on Windows -- no macOS branch since CUDA doesn't support it. `pixi.lock` is reverted to upstream (the lock churn was solely from the now-removed dep). Tests in `test_default_cache_dir_lives_under_user_cache_root` parametrise both branches.

**`_program_cache.py` split into a package** (`cuda_core/cuda/core/utils/_program_cache/{__init__.py,_abc.py,_keys.py,_in_memory.py,_file_stream.py}`) -- 1700 lines → 5 focused submodules, each ≤ 700 lines. Public surface (\`FileStreamProgramCache\`, \`InMemoryProgramCache\`, \`ProgramCacheResource\`, \`make_program_cache_key\`) re-exported from the package, so external imports are unchanged. Tests that monkeypatch internals were updated to address the owning submodule directly (`_program_cache._keys._linker_backend_and_version`, `_program_cache._file_stream._IS_WINDOWS`, etc.) -- the package-level symbols are convenience aliases and don't intercept calls within submodules.

**test_program_cache.py file split deferred.** The 2179-line file is internally well-organised by section (key construction / InMemory / FileStream / multi-process), and after the source split the test reorganisation is mechanical churn rather than a clarity win. Happy to do it as a follow-up if you'd prefer.

Multi-GPU question replied to in [#issuecomment-4350987016](#issuecomment-4350987016).

cuda_core/cuda/core/utils/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+    if name in _LAZY_CACHE_ATTRS:
+        from cuda.core.utils import _program_cache
+
+        value = getattr(_program_cache, name)
+        globals()[name] = value  # cache for subsequent accesses
+        return value
+    raise AttributeError(f"module 'cuda.core.utils' has no attribute {name!r}")
+
+
+def __dir__():
```
rparolin: To be removed

cuda_core/docs/source/api.rst
```diff
@@ -247,7 +247,23 @@ Utility functions
...
 
    args_viewable_as_strided_memory
+   make_program_cache_key
 
    :template: autosummary/cyclass.rst
 
    StridedMemoryView
+
+Program caches
+--------------
```
leofang: Should we merge this section with `CUDA compilation toolchain` above, or at least move it and make them stay closer?

cuda_core/cuda/core/utils/_program_cache/_abc.py
```diff
@@ -0,0 +1,187 @@
...
+        and size-cap enforcement run on every entry. Not transactional --
+        a failure mid-iteration leaves earlier writes committed.
+        """
+        if isinstance(items, collections.abc.Mapping):
+            items = items.items()
+        for key, value in items:
+            self[key] = value
+
+    def close(self) -> None:  # noqa: B027
+        """Release backend resources. No-op by default."""
```
leofang: This is no-op only for trivial backends. If users don't use it as a context manager (well... it's a separate discussion why they should..), then they better call `.close()` for consistency and portability (to another backend).

cuda_core/docs/source/api.rst
leofang: Alternatively, just merging with the above section is fine too.

cuda_core/cuda/core/utils/_program_cache/_abc.py
```diff
@@ -0,0 +1,187 @@
...
+            options=options,
+            target_type="cubin",
+            extra_digest=header_fingerprint(),
+        )
+        data = cache.get(key)
+        if data is None:
+            obj = program.compile("cubin")
+            cache[key] = obj  # extracts bytes(obj.code)
+        else:
+            obj = ObjectCode._init(data, "cubin")
```
leofang: `ObjectCode` has many public constructors. We must not teach about any private methods.

cuda_core/tests/test_program_compile_cache.py
```diff
@@ -0,0 +1,357 @@
...
+
+class _RecordingCache:
+    """Minimal recording stub for the bytes-in / bytes-out cache protocol.
+
+    Mirrors :class:`FileStreamProgramCache`'s contract: ``__setitem__``
+    accepts bytes-like or :class:`ObjectCode` (extracts bytes), and
+    ``get`` returns the stored bytes (or ``None``).
+
+    Intentionally does NOT subclass ``ProgramCacheResource`` -- the wrapper
+    should be duck-typed, so we test the duck-typed surface directly.
```
leofang: Why don't we require each `cache=` instance to subclass from `ProgramCacheResource`?

cuda_core/cuda/core/_program.pyx
```diff
@@ -577,8 +681,16 @@ cdef inline void _process_define_macro(list options, object macro) except *:
     raise RuntimeError(f"Expected define_macro {union_type}, list[{union_type}], got {macro}")
 
 
-cpdef bint _can_load_generated_ptx() except? -1:
-    """Check if the driver can load PTX generated by the current NVRTC version."""
+def _can_load_generated_ptx():
```
leofang: `cpdef` functions are also accessible from Python, is this change needed?

cuda_core/cuda/core/_program.pyx
```diff
@@ -98,13 +103,99 @@ cdef class Program:
...
+            target_type=target_type,
+        )
+        hit_bytes = cache.get(key)
+        if hit_bytes is not None:
+            # The uncached NVRTC path warns when the active driver can't
+            # load freshly-generated PTX; that loadability is a property
+            # of the driver, not of how the bytes were produced, so the
+            # warning applies equally to cached PTX. Mirror it here so a
+            # cache hit doesn't silently hide an incompatibility that the
+            # uncached call would have surfaced.
```
leofang, line 210: **Critical**: Cache-write failure loses the successfully compiled `ObjectCode`.

If `cache[key] = compiled` raises (disk full, permissions, I/O error), the exception propagates and the caller never gets `compiled` back — even though the compile succeeded and the result is sitting right there. An _optional_ `cache=` parameter should not turn a successful compile into a failure.

Suggested fix — wrap the cache write so the compile result is always returned:

```python
compiled = _program_compile_uncached(self, target_type, name_expressions, logs)
try:
    cache[key] = compiled
except Exception:
    pass  # or warnings.warn(...)
return compiled
```

cuda_core/cuda/core/_program.pyx
```diff
@@ -98,13 +103,99 @@ cdef class Program:
...
+            code=code_for_key,
+            code_type=self._code_type,
+            options=self._options,
+            target_type=target_type,
+        )
+        hit_bytes = cache.get(key)
+        if hit_bytes is not None:
+            # The uncached NVRTC path warns when the active driver can't
+            # load freshly-generated PTX; that loadability is a property
+            # of the driver, not of how the bytes were produced, so the
```
leofang, line 207: **Consideration**: `logs` silently receives nothing on cache hit.

When the cache hits, no compilation occurs, so `logs.write()` is never called. A caller relying on `logs` to confirm compilation ran (or to capture PTX output) will get silence with no indication it was a cache hit vs. a compile that produced no output. Worth a note in the `cache=` or `logs` parameter docstring, e.g.:

> On a cache hit, no compilation is performed and `logs` receives no output.

cuda_core/cuda/core/utils/_program_cache/_keys.py
```diff
@@ -0,0 +1,782 @@
...
+        # just those fields so shared ProgramOptions carrying NVRTC-only
+        # flags (``include_path``, ``pch_*``, ``frandom_seed``, ...) don't
+        # force spurious cache misses on PTX.
+        return _linker_option_fingerprint(options, use_driver_linker=self._decide_driver())
+
+    def hash_version_probe(self, update):
+        # Only cuLink (driver-backed linker) goes through the CUDA driver
+        # for codegen. nvJitLink is a separate library, so a driver
+        # upgrade under it does not change the compiled bytes -- skip the
+        # driver version there. ``_linker_backend_and_version`` already
```
leofang, line 533: **Consideration**: `_decide_driver()` is called 2–3 times per key derivation for PTX inputs.

`validate()` calls `self._decide_driver()` at line 494, `option_fingerprint()` calls it again at line 515, and `hash_version_probe()` calls `_linker_backend_and_version()` which re-probes independently. Each call re-imports and re-invokes `_decide_nvjitlink_or_driver()`.

If the result were to differ between calls within the same invocation (transient import failure recovering), the key would be internally inconsistent — validation used one backend, fingerprint used another. Even though this is unlikely in practice, computing the result once and threading it through (or caching it on the instance for the duration of one `make_program_cache_key` call) would be cleaner and more robust.

cuda_core/cuda/core/utils/_program_cache/_keys.py
```diff
@@ -0,0 +1,782 @@
...
+    # Mirror Program.compile (_program.pyx Program_init lowercases code_type
+    # before dispatch); a caller that passes "PTX" or "C++" must get the
+    # same routing and the same cache key as the lowercase form.
+    code_type = code_type.lower() if isinstance(code_type, str) else code_type
+    if code_type not in _VALID_CODE_TYPES:
+        raise ValueError(f"code_type={code_type!r} is not supported (must be one of {sorted(_VALID_CODE_TYPES)})")
+    if target_type not in _VALID_TARGET_TYPES:
+        raise ValueError(f"target_type={target_type!r} is not supported (must be one of {sorted(_VALID_TARGET_TYPES)})")
+    supported_for_code = _SUPPORTED_TARGETS_BY_CODE_TYPE[code_type]
+    if target_type not in supported_for_code:
```
leofang, line 757: **Consideration**: `target_type` is not case-normalized, asymmetric with `code_type`.

`code_type` is lowered here to ensure `"PTX"` and `"ptx"` route the same way. But `target_type` is not normalized — `make_program_cache_key(target_type="PTX")` and `make_program_cache_key(target_type="ptx")` produce different keys for the same compilation. The comment on lines 728–730 says "a caller that passes 'PTX' or 'C++' must get the same routing" but only addresses `code_type`.

`Program.compile` may normalize `target_type` internally, but standalone callers of `make_program_cache_key` are exposed to this instability. Consider adding `target_type = target_type.lower()` here to match.

cuda_core/cuda/core/utils/_program_cache/_keys.py
```diff
@@ -0,0 +1,782 @@
...
+    "split_compile": _gate_identity,
+    "ptxas_options": _gate_ptxas_options,
+    "no_cache": _gate_is_true,
+}
+
+
+# LinkerOptions fields the ``cuLink`` driver backend silently ignores
+# (emits only a DeprecationWarning; no actual flag reaches the compiler).
+# When the driver backend is active, collapse them to a single sentinel in
+# the fingerprint so nvJitLink<->driver parity of ``ObjectCode`` doesn't
```
leofang, line 129: **Consideration**: `_LINKER_RELEVANT_FIELDS` and `_LINKER_FIELD_GATES` must be kept in sync manually.

A field added to `_LINKER_FIELD_GATES` but forgotten in `_LINKER_RELEVANT_FIELDS` would be silently ignored — the gate exists but is never iterated over in `_linker_option_fingerprint`. The reverse (missing from `_LINKER_FIELD_GATES`) would `KeyError` at runtime, which is fail-fast but late.

A test assertion like `assert set(_LINKER_RELEVANT_FIELDS) == set(_LINKER_FIELD_GATES.keys())` would close both directions cheaply. Same applies for sync with `_translate_program_options` in `_program.pyx` — the existing `test_make_program_cache_key_supported_targets_matches_program_compile` guards target drift but not field drift.

cuda_core/cuda/core/utils/_program_cache/_abc.py
```diff
@@ -0,0 +1,187 @@
...
+    if isinstance(value, ObjectCode):
+        code = value.code
+        if isinstance(code, str):
+            return Path(code).read_bytes()
+        return bytes(code)
+    raise TypeError(f"cache values must be bytes-like or ObjectCode, got {type(value).__name__}")
+
+
+def _as_key_bytes(key: object) -> bytes:
+    if isinstance(key, (bytes, bytearray)):
```
leofang, line 44: **Consideration**: `_extract_bytes` for path-backed `ObjectCode` raises a bare `FileNotFoundError` with no context.

If the underlying file was deleted after `ObjectCode` construction, the user sees `FileNotFoundError: '/some/path'` with no indication it came from the cache extracting bytes from a path-backed `ObjectCode`. This could be confusing — "I called `cache[key] = obj`, why is it looking for a file?"

Consider wrapping with a more descriptive error:

```python
if isinstance(code, str):
    try:
        return Path(code).read_bytes()
    except FileNotFoundError:
        raise FileNotFoundError(
            f"path-backed ObjectCode references {code!r}, "
            f"but the file no longer exists"
        ) from None
```

cuda_core/cuda/core/utils/_program_cache/_keys.py
```diff
@@ -0,0 +1,782 @@
...
+    "split_compile",
+    "ptxas_options",
+    "no_cache",
+)
+
+
+# Map each linker-relevant ProgramOptions field to the gate the Linker uses
+# to turn it into a flag (see ``_prepare_nvjitlink_options`` and
+# ``_prepare_driver_options`` in _linker.pyx). Collapsing inputs through
+# these gates means semantically-equivalent configurations
```
leofang: **Consideration**: `no_cache` naming collision with `Program.compile(cache=...)`.

`ProgramOptions.no_cache` controls the _linker's_ internal JIT cache. `Program.compile(cache=...)` controls the _program-level_ cache. A user who sets `no_cache=True` and passes `cache=` probably expects no caching at all, but gets program-level caching anyway. The two caching layers are independent, but the naming overlap is confusing.

At minimum, a note in the `cache=` parameter docstring clarifying this distinction would help. Alternatively, `Program.compile` could warn when `options.no_cache` is set alongside `cache=`.

cuda_core/tests/test_program_cache_multiprocess.py
leofang, line 1: **Consideration**: Multiprocess reader test (`test_concurrent_reader_never_sees_torn_file`) only checks `hits == 200` (every read returned non-`None`) but never verifies the returned bytes match the expected payload. A partially-written file that happens to be non-empty would pass. The title claims "never sees torn file" but the assertion only proves "never sees `None`."

Adding `assert got == expected_payload` inside the reader loop (or at least checking the length) would actually verify atomicity.

Similarly, `test_concurrent_eviction_does_not_delete_replaced_file`: the rewriter writes a final `cache[key] = payload + b"final"` _after_ `done_event` is set (churner already exited), so the assertion `got.endswith(b"final")` can pass even if the stat-guard was broken during the concurrent phase — the final write is uncontested.

cuda_core/cuda/core/utils/_program_cache/_in_memory.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def __setitem__(self, key: object, value: bytes | bytearray | memoryview | ObjectCode) -> None:
+        data = _extract_bytes(value)
+        size = len(data)
+        k = _as_key_bytes(key)
+        with self._lock:
+            existing = self._entries.pop(k, None)
+            if existing is not None:
+                self._total_bytes -= existing[1]
+            self._entries[k] = (data, size)
```
leofang, line 89: **Consideration**: No thread-safety tests for `InMemoryProgramCache`.

The docstring promises "the cache can be shared across threads without external locking" and the implementation wraps every method in `RLock`. But there are zero concurrent-thread tests — the multiprocess suite covers `FileStreamProgramCache`, but `InMemoryProgramCache` has no analogous stress test. A test with N threads doing concurrent reads + writes against a size-capped cache (verifying no crashes, no deadlocks from `RLock` reentrance in `__setitem__` → `_evict_to_caps` → `popitem`, and consistent final state) would validate the claim.

Also missing: a `FileStreamProgramCache` overwrite test analogous to `test_inmemory_cache_overwrite_replaces_value_and_updates_size` — verifying that overwriting a key via `os.replace` replaces the on-disk file and `__len__` stays at 1.

cuda_core/cuda/core/utils/_program_cache/_file_stream.py
```diff
@@ -0,0 +1,614 @@
...
+        digest = hashlib.blake2b(k, digest_size=32).hexdigest() if k else "empty"
+        if len(digest) < 3:
+            digest = digest.rjust(3, "0")
+        return self._entries / digest[:2] / digest[2:]
+
+    # -- mapping API ---------------------------------------------------------
+
+    def __getitem__(self, key: object) -> bytes:
+        path = self._path_for_key(key)
+        try:
```
leofang, line 436: **Nitpick**: `_path_for_key` special-cases empty key unnecessarily.

`blake2b(b"", digest_size=32)` is well-defined and produces a collision-resistant digest. The `if k else "empty"` branch adds a special case that's harder to reason about (the literal `"empty"` could collide with a key whose blake2b hex starts with `"empty"`, however unlikely). The `len(digest) < 3` guard on line 416 is also dead code — blake2b with `digest_size=32` always returns a 64-character hex string, and `"empty"` is 5 characters.

Suggest removing both branches and unconditionally hashing:
```python
digest = hashlib.blake2b(k, digest_size=32).hexdigest()
return self._entries / digest[:2] / digest[2:]
```

cuda_core/cuda/core/utils/_program_cache/_file_stream.py
```diff
@@ -0,0 +1,614 @@
...
+        count = 0
+        for path in self._iter_entry_paths():
+            if path.is_file():
+                count += 1
+        return count
+
+    def clear(self) -> None:
+        # Snapshot stat alongside path so we can refuse to unlink an entry
+        # that was concurrently replaced by another process between the
+        # snapshot scan and the unlink. Same stat-guard contract as
```
leofang, line 559: **Nitpick**: `__len__` double-stats every entry.

`_iter_entry_paths` already filters with `entry.is_file()` (line 531), so every yielded path is known to be a file. The `path.is_file()` check here in `__len__` is redundant — it issues a second `stat` syscall per entry. For a large cache this is 2N unnecessary stat calls. Safe to remove the inner check:

```python
def __len__(self) -> int:
    return sum(1 for _ in self._iter_entry_paths())
```

cuda_core/cuda/core/utils/_program_cache/_file_stream.py
```diff
@@ -0,0 +1,614 @@
...
+        max_size_bytes: int | None = None,
+    ) -> None:
+        if max_size_bytes is not None and max_size_bytes < 0:
+            raise ValueError("max_size_bytes must be non-negative or None")
+        self._root = Path(path) if path is not None else _default_cache_dir()
+        self._entries = self._root / _ENTRIES_SUBDIR
+        self._tmp = self._root / _TMP_SUBDIR
+        self._max_size_bytes = max_size_bytes
+        self._root.mkdir(parents=True, exist_ok=True)
+        self._entries.mkdir(exist_ok=True)
```
leofang, line 399: **Nitpick**: `max_size_bytes=0` is accepted but makes the cache a black hole.

Both `InMemoryProgramCache` and `FileStreamProgramCache` validate `max_size_bytes >= 0`, so `max_size_bytes=0` passes. But a zero-byte cap means every write is immediately evicted (any payload has `size > 0`). This is almost certainly a user error. Consider either rejecting it with a `ValueError`, or documenting it explicitly as "effectively disables caching."

cuda_core/cuda/core/utils/_program_cache/_keys.py
```diff
@@ -0,0 +1,782 @@
...
+        if len(v) == 0:
+            return None
+        return tuple(f"-Xptxas={s}" for s in v)
+    return v
+
+
+_LINKER_FIELD_GATES = {
+    "name": _gate_identity,
+    "arch": _gate_identity,
+    "max_register_count": _gate_identity,
```
leofang, line 111: **Nitpick**: `ptxas_options` fingerprint preserves list order.

`["-v", "-O2"]` and `["-O2", "-v"]` produce different cache keys even if ptxas treats them equivalently. This causes spurious cache misses (not collisions), so it's safe but suboptimal. The conservative choice (preserve order) is defensible if ptxas flag order _is_ significant for some flags. Worth a brief comment noting the choice is intentional.

cuda_core/cuda/core/utils/_program_cache/_keys.py
```diff
@@ -0,0 +1,782 @@
...
+import hashlib
+from typing import Sequence
+
+from cuda.core._program import ProgramOptions
+from cuda.core._utils.cuda_utils import (
+    driver as _driver,
+)
+from cuda.core._utils.cuda_utils import (
+    handle_return as _handle_return,
+)
```
leofang, line 32: **Nitpick**: Top-level import creates a fragile circular-import dependency.

`_keys.py` imports `ProgramOptions` from `cuda.core._program` at module level, while `_program.pyx` does a deferred import of `make_program_cache_key` inside `compile()`. The deferred import breaks the cycle today, but if someone later adds `from cuda.core.utils import ...` at the top of `_program.pyx`, it would break. Worth a `# NOTE: ...` comment here documenting the mutual dependency and why _keys.py gets the top-level import while _program.pyx defers.

cuda_core/cuda/core/utils/_program_cache/_keys.py
```diff
@@ -0,0 +1,782 @@
...
+    # must reflect that or a cache hit could hand back an entry with the
+    # wrong name. Universal across backends.
+    options_name = getattr(options, "name", None)
+    if options_name is not None:
+        _update("options_name", str(options_name).encode("utf-8"))
+
+    if extra_digest is not None:
+        _update("extra_digest", bytes(extra_digest))
+
+    return hasher.digest()
```
leofang, line 813: **Nitpick**: `options.name` is double-hashed for the linker (PTX) path.

For PTX inputs, `"name"` is in `_LINKER_RELEVANT_FIELDS` and goes through `_linker_option_fingerprint` via `_gate_identity`. Then this universal block hashes it again under the `"options_name"` label. Not a collision or instability bug (hashing the same value twice under different labels is safe), but it's redundant work for the linker path. Worth a comment noting the intentional redundancy, or dedup by skipping the linker path here.

leofang (commented):
Review of the persistent program cache implementation. Findings are categorized inline as:

- **Critical** (1): Must fix before merging — cache-write failure drops a successfully compiled `ObjectCode`.
- **Consideration** (8): Performance/functionality concerns worth discussing; can be deferred but should be tracked.
- **Nitpick** (6): Not blockers for merging.

This excludes items already captured in Leo's and rwgk's earlier review comments (platformdirs removal, over-eviction race, source-directory include guard, `ObjectCode._init` in docstrings, `cpdef`→`def` change, duck-typed test, `close()` vs context manager, multi-GPU usage, star-import laziness, doc section placement, SQLiteProgramCache removal).

Overall this is a well-engineered piece of work — the TOCTOU handling, stat-guards, and atomic-write design are thorough and well-documented. The main concern is the cache-write failure path losing the compile result.

cuda_core/cuda/core/_program.pyx
leofang: I would actually disagree with my bot here. Caches are already an opt-in construct, and if it fails and we swallow the error silently, it makes no use and is harder to debug why cache miss happens. Resolving.

cuda_core/cuda/core/utils/_program_cache/_file_stream.py
```diff
@@ -0,0 +1,614 @@
...
+            total += st.st_size
+        if self._tmp.exists():
+            for tmp in self._tmp.iterdir():
+                if not tmp.is_file():
+                    continue
+                try:
+                    total += tmp.stat().st_size
+                except FileNotFoundError:
+                    continue
+        if total <= self._max_size_bytes:
```
leofang: **Consideration**: Temp file accounting under burst writes can cause committed-entry thrashing.

`_enforce_size_cap` counts temp files toward `total` but can only evict committed entries (the `entries` list). During high-concurrency writes, many young temp files (each under the 1h stale-sweep threshold) could push `total` persistently above the cap with no way to recover — the loop would keep evicting committed entries while the in-flight temps hold the total above `max_size_bytes`. Once the burst subsides and the temps are committed/cleaned, the evicted entries are gone.

This is a soft-cap design so it's consistent with the documented contract, but worth noting that burst write concurrency can cause disproportionate eviction of committed entries.

cuda_core/cuda/core/utils/_program_cache/_file_stream.py
```diff
@@ -0,0 +1,614 @@
...
+            # propagate immediately (real config problem).
+            if not _replace_with_sharing_retry(tmp_path, target):
+                with contextlib.suppress(FileNotFoundError):
+                    tmp_path.unlink()
+                return
+        except BaseException:
+            with contextlib.suppress(FileNotFoundError):
+                tmp_path.unlink()
+            raise
+        self._enforce_size_cap()
```
leofang: **Consideration**: `_enforce_size_cap` is O(n) on every `__setitem__`.

Every write stats all files in `entries/` plus `tmp/` to compute the total. For a large cache (thousands of entries), this could be measurably costly on every compile. An incremental size tracker (add on write, subtract on eviction, periodic reconciliation to correct drift from external deletions) would make writes O(1) in the common case, falling back to a full scan only when reconciliation is needed.

cuda_core/cuda/core/_program.pyx
```diff
@@ -98,13 +103,99 @@ cdef class Program:
...
+        # intercept reliably.
+        from cuda.core.utils._program_cache import make_program_cache_key
+
+        # ``self._code`` is always stored as bytes (see ``Program_init``),
+        # but ``make_program_cache_key`` only accepts bytes when
+        # ``code_type == "nvvm"`` -- c++/ptx must be ``str``. Decode back
+        # to the original str for the NVRTC/linker paths so the generated
+        # key matches keys callers build by passing the str source
+        # directly.
+        code_for_key = self._code if self._code_type == "nvvm" else self._code.decode("utf-8")
```
leofang, line 195: **Consideration**: UTF-8 decode introduces a new failure mode in the cache path.

If the source code contains non-UTF-8 bytes (e.g. Latin-1 encoded comments), this `.decode("utf-8")` raises `UnicodeDecodeError` even though the uncached compile path would succeed (NVRTC accepts raw bytes). The cache path introduces a failure that doesn't exist without `cache=`.

This is likely rare in practice (CUDA source is almost always ASCII/UTF-8), but worth documenting in the `cache=` parameter docstring or catching with a clear error message pointing to the limitation.

cuda_core/cuda/core/utils/_program_cache/_file_stream.py
```diff
@@ -0,0 +1,614 @@
...
+            try:
+                st_now = os.fstat(fd)
+            except OSError:
+                return
+            if (st_now.st_ino, st_now.st_size, st_now.st_mtime_ns) != (
+                st_before.st_ino,
+                st_before.st_size,
+                st_before.st_mtime_ns,
+            ):
+                return
```
leofang: **Nitpick**: Stat-key triple `(st_ino, st_size, st_mtime_ns)` is repeated in 4 places.

This exact pattern appears in `_touch_atime` (twice — fd-based and path-based), `_prune_if_stat_unchanged`, and `_enforce_size_cap`. Extracting a small helper would reduce duplication and make the invariant easier to audit:

```python
def _stat_key(st: os.stat_result) -> tuple:
    return (st.st_ino, st.st_size, st.st_mtime_ns)
```

Then each site becomes `if _stat_key(st_now) != _stat_key(st_before): return`.

cuda_core/cuda/core/utils/_program_cache/_file_stream.py
```diff
@@ -0,0 +1,614 @@
...
+            os.replace(tmp_path, target)
+            return True
+        except PermissionError as exc:
+            if not _IS_WINDOWS or getattr(exc, "winerror", None) not in _SHARING_VIOLATION_WINERRORS:
+                raise
+            # Windows sharing violation; loop and try again unless this was the
+            # last attempt, in which case fall through and return False.
+            if i == len(_REPLACE_RETRY_DELAYS) - 1:
+                return False
+    return False
```
leofang: **Nitpick**: Windows sharing-violation retry pattern is repeated across 3 functions.

`_replace_with_sharing_retry`, `_stat_and_read_with_sharing_retry`, and `_unlink_with_sharing_retry` all iterate over `_REPLACE_RETRY_DELAYS`, sleep, try an operation, catch `PermissionError`, and check `_is_windows_sharing_violation`. A generic retry helper parameterized on the operation could deduplicate this:

```python
def _with_sharing_retry(op, *args, on_exhausted=None, **kwargs):
    last_exc = None
    for delay in _REPLACE_RETRY_DELAYS:
        if delay:
            time.sleep(delay)
        try:
            return op(*args, **kwargs)
        except PermissionError as exc:
            if not _is_windows_sharing_violation(exc):
                raise
            last_exc = exc
    if on_exhausted is not None:
        return on_exhausted(last_exc)
    raise last_exc
```

Not critical — the current code is clear and correct — but it's 3× the same loop structure.

cuda_core/cuda/core/utils/_program_cache/_keys.py
```diff
@@ -0,0 +1,782 @@
...
+    AND across processes that hit the same failure mode. The
+    ``_probe_failed`` label differs from every backend's success label, so a
+    broken environment never collides with a working one -- the cache
+    "fails closed" between broken and working environments while staying
+    persistent within either.
+    """
+    update(f"{label}_probe_failed", type(exc).__name__.encode())
+
+
+class _KeyBackend(abc.ABC):
```
leofang, line 311: **Nitpick**: `_KeyBackend` class hierarchy vs plain function dispatch.

Three stateless singleton classes (`_NvrtcBackend`, `_LinkerBackend`, `_NvvmBackend`) for what is essentially a 3-way dispatch on `code_type`. The ABC does document the contract clearly (6 abstract/overridable methods), but three plain functions (or a dict of named tuples of functions) dispatched from `make_program_cache_key` would be simpler and equally extensible — there will likely never be a 4th code_type. The class hierarchy adds indirection and cognitive overhead (readers must trace through the ABC to understand the dispatch) without a clear extensibility payoff.

Not a correctness issue — the current design works — just noting the complexity cost relative to a simpler alternative.

leofang (commented):
Follow-up batch — 6 additional inline comments from the first review round:

- **Consideration** (3): temp file burst-write thrashing, O(n) `_enforce_size_cap` on every write, UTF-8 decode introducing a new failure mode in the cache path.
- **Nitpick** (3): stat-key triple dedup, Windows sharing-retry dedup, `_KeyBackend` class hierarchy vs simpler function dispatch.

cuda_core/cuda/core/utils/_program_cache/_file_stream.py
cpcloud: Agreed — this is an optimization, not a correctness issue. The cache is bounded by `max_size_bytes`, so the typical case walks a few hundred cubins per write at microsecond `stat` cost. An incremental size tracker with periodic reconciliation is the right answer for very large caches; deferring to a follow-up.

cuda_core/cuda/core/_program.pyx
cpcloud: The decode is provably safe in the current shape: `Program_init` calls `assert_type(code, str)` for both `c++` and `ptx` paths, so `self._code = code.encode()` always lands UTF-8-valid bytes. NVVM is the only path that retains user-supplied bytes, and the cache uses `self._code` directly there (no decode). I'd prefer not to add a try/except for a code path that the assert_type guard already eliminates -- if we ever loosen `Program_init` to accept bytes for c++/ptx, the right place to handle non-UTF-8 input is the new entry, not as a fallback inside `compile`. Will update the inline comment to spell that out.

cuda_core/cuda/core/utils/_program_cache/_keys.py
cpcloud: The recent change to cache the cuLink-vs-nvJitLink probe per make_program_cache_key call (addressing your other comment about _decide_driver running 2-3 times) gives _LinkerBackend real per-instance state -- _cached_decision plus _cached_decision_exc that validate, option_fingerprint, and hash_version_probe all share. Flattening to free functions would force that state into a context object passed through every call site or back into module globals reset somehow per call. The class hierarchy now expresses exactly the per-call lifetime we want; keeping it.

cuda_core/docs/source/api.rst
cpcloud: Done in 205b15b — "Program caches" is now a subsection of "CUDA compilation toolchain" with a `currentmodule:: cuda.core.utils` switch around the autosummary. `make_program_cache_key` moved with it.

cuda_core/cuda/core/utils/_program_cache/_abc.py
cpcloud: Done in 45790ad — `close()` docstring now spells out the no-op default and tells callers to use the context manager or call `close()` explicitly so code stays portable across backends that do hold resources.

cuda_core/cuda/core/utils/_program_cache/_abc.py
cpcloud: Done in d77a5a9 — both docstring examples now use the public `from cuda.core import ObjectCode` + `ObjectCode.from_cubin(...)` pattern.

cuda_core/tests/test_program_compile_cache.py
cpcloud: Done in 701d00b — `Program.compile` now isinstance-checks `cache=` against `ProgramCacheResource` and raises `TypeError` up front. The recording test cache subclasses the ABC; added a regression test that a duck-typed cache is rejected.

cuda_core/cuda/core/_program.pyx
cpcloud: Done in 821131a — restored `cpdef bint _can_load_generated_ptx() except? -1`. The two cache-hit warning tests now mock `driver_version` (to `(0, 0, 0)` for the warning case, `(999, 0, 0)` for the no-warning case) so the cpdef body computes the desired result without needing a Python-level test seam.

cuda_core/cuda/core/_program.pyx
cpcloud: Done in 863a145 — added a note to the `logs=` docstring that cache hits don't run a compile and therefore don't write to `logs`.

cuda_core/cuda/core/utils/_program_cache/_keys.py
cpcloud: Done in 41806ed — `_LinkerBackend` caches the `_decide_nvjitlink_or_driver()` result (and any probe exception) on the instance; `_BACKENDS_BY_CODE_TYPE` now stores classes and `make_program_cache_key` instantiates fresh per call so the cache lives exactly one call. `_linker_backend_and_version()` takes `use_driver` as an arg so the third probe collapses too.

cuda_core/cuda/core/utils/_program_cache/_keys.py
cpcloud: Done in c877f1e — `Program.compile` lowercases `target_type` at the top, and `make_program_cache_key` mirrors the normalisation. Added a parametrised case-insensitivity test.

cuda_core/cuda/core/utils/_program_cache/_keys.py
cpcloud: Done in 8323be7 — dropped `_LINKER_RELEVANT_FIELDS` entirely; `_LINKER_FIELD_GATES` is now the single source of truth and `_linker_option_fingerprint` iterates the dict directly.

cuda_core/cuda/core/utils/_program_cache/_abc.py
cpcloud: Done in ec93588 — `_extract_bytes` now wraps `Path.read_bytes()`'s `FileNotFoundError` with a message naming the cache operation and the missing path. Added a regression test.

cuda_core/cuda/core/utils/_program_cache/_keys.py
cpcloud: Done in 2c09fe5 — added a docstring paragraph distinguishing program-level `cache=` from linker-level `ProgramOptions.no_cache`.

cuda_core/tests/test_program_cache_multiprocess.py
cpcloud: Done in 0a2916b — reader test now carries the seeded payload through the worker and counts byte mismatches (must be zero). Eviction-race test drops the post-race uncontested `payload + b"final"` write so survival depends on the in-race rewrites being preserved by the stat-guard; the assertion now requires the value to start with the rewriter's `payload` prefix.

cuda_core/cuda/core/utils/_program_cache/_in_memory.py
cpcloud: Done in 3ded273 — added a 4-writers + 4-readers × 200-ops `InMemoryProgramCache` stress test (no exceptions, no deadlocks, internal accounting stays consistent) and a `FileStreamProgramCache` overwrite test that asserts `__len__` stays at 1 and exactly one entry file lives on disk.

cuda_core/cuda/core/utils/_program_cache/_file_stream.py
cpcloud: Done in 1dd90b4 — dropped both the empty-key branch and the `len(digest) < 3` guard; `_path_for_key` now unconditionally `blake2b(k).hexdigest()`s.

cuda_core/cuda/core/utils/_program_cache/_file_stream.py
cpcloud: Done in e170ca9 — `__len__` returns `sum(1 for _ in self._iter_entry_paths())`; the inner `path.is_file()` is gone.

cuda_core/cuda/core/utils/_program_cache/_file_stream.py
cpcloud: Done in 5bb4016 — both backends reject `max_size_bytes=0` with `ValueError` ("must be positive or None"). Tests parametrise over `-1` and `0`.

cuda_core/cuda/core/utils/_program_cache/_keys.py
cpcloud: Done in 756a9b4 — added a comment noting order is intentionally significant (ptxas accepts ordering-sensitive flag pairs, e.g. `-O2` after `-O3` lowers the active level).

cuda_core/cuda/core/utils/_program_cache/_keys.py
cpcloud: Done in 8ef7507 — added a NOTE block on the `from cuda.core._program import ProgramOptions` import documenting the mutual-dependency contract with `_program.pyx`.

cuda_core/cuda/core/utils/_program_cache/_keys.py
cpcloud: Done in 5fa0de1 — extended the universal-block comment to acknowledge the double-hash on PTX is intentional (linker reads `name` too) and harmless because the two label-payload pairs can't collide.

cuda_core/cuda/core/utils/_program_cache/_file_stream.py
cpcloud: Done in 0beee3c — extended `_enforce_size_cap`'s comment to spell out the burst-write over-eviction trade-off (counting young temps toward `total` lets the cap drag committed entries out during high-concurrency bursts; the soft-cap contract permits it).

cuda_core/cuda/core/utils/_program_cache/_file_stream.py
cpcloud: Done in f1502af — extracted `_stat_key(st) -> (st_ino, st_size, st_mtime_ns)` and routed all four stat-guarded sites through it.

cuda_core/cuda/core/utils/_program_cache/_file_stream.py
cpcloud: Done in dbbb528 — collapsed the three retry loops into `_with_sharing_retry(op, on_exhausted=...)`. Net behaviour preserved per call site through the `on_exhausted` callback.

cuda_core/cuda/core/utils/_program_cache/_file_stream.py
leofang: @cpcloud How much work is it to address this? Can we just keep an internal size counter without accessing the file system? I think this is a severe issue from CuPy's experience. 

We can easily populate the cache with O(1-10k) CUBINs. Filesystem operations are also costly in the multi-process cases. Stat'ing over many small files can be quite slow (from which I already suffer every time I manage my local CuPy cache).

cpcloud: Implemented in f8c9264ae2.

`FileStreamProgramCache` now keeps an incremental `_tracked_size_bytes` (seeded by one full scan at open, mutated by the net delta on every `__setitem__`/`__delitem__`/`clear`, refreshed by every actual eviction pass). Writes that stay under `max_size_bytes` are now O(1) — `_enforce_size_cap` is only called when the running total crosses the cap. The eviction scan doubles as the reconciliation point, so cross-process drift (other writers/deleters) self-corrects the next time the cap fires.

Also switched `_iter_entry_paths` to `os.scandir` (with `follow_symlinks=False`) so `is_dir`/`is_file` consult the dirent type cache instead of issuing a per-entry `stat`.

Three new tests (`tests/test_program_cache.py`):
- `test_filestream_cache_writes_skip_scan_when_under_cap` — 100 writes under cap, `_enforce_size_cap` is never called.
- `test_filestream_cache_writes_trigger_scan_when_over_cap` — once a write crosses the cap, the scan fires.
- `test_filestream_cache_tracker_reconciles_after_external_drift` — external `unlink` behind the cache stale-highs the tracker; the next eviction observes reality and rebases.

leofang:
Thanks @cpcloud! Let's resolve the conflict and get this in.

cuda_core/cuda/core/utils/_program_cache/__init__.py
leofang, line 1: nit: I forgot that we already have a folder for collecting utils:
```
cuda_core/cuda/core/_utils
```
How about we move the program cache files to there? Say,
```
cuda_core/cuda/core/_utils/_program_cache/
```

cuda_core/cuda/core/utils/_program_cache/_file_stream.py
```diff
@@ -0,0 +1,739 @@
...
+                size = path.stat().st_size
+            except FileNotFoundError:
+                raise KeyError(key) from None
+        try:
+            _unlink_with_sharing_retry(path)
+        except FileNotFoundError:
+            raise KeyError(key) from None
+        if self._max_size_bytes is not None:
+            with self._size_lock:
+                self._tracked_size_bytes -= size
```
leofang: **Consideration** (correctness): `_tracked_size_bytes` can go permanently negative, disabling all future eviction.

Scenario: thread A calls `__delitem__` on key K and stats the file (gets `size`). Concurrently, thread B runs `_enforce_size_cap`, scans the directory (sees K, includes its size in `total`), finishes eviction, and re-seeds: `self._tracked_size_bytes = total` (which includes K's size). Then thread A's subtract runs here: `self._tracked_size_bytes -= size`. The tracker is now lower than reality by `size`. Repeat enough times across concurrent delete+eviction races, and the tracker goes negative.

Once negative, the check `self._tracked_size_bytes > self._max_size_bytes` never fires, so `_enforce_size_cap` is never called again. The cache grows without bound despite having a configured `max_size_bytes`. Unlike the "soft cap temporarily exceeded" behavior (which is fine), this is "eviction permanently disabled" — there is no self-healing path because the reconciliation point lives inside `_enforce_size_cap`, which is the function that never gets called.

Fix: clamp to zero so the tracker can't cross into the permanently-broken state:

```python
with self._size_lock:
    self._tracked_size_bytes = max(0, self._tracked_size_bytes - size)
```

This doesn't make the tracker perfectly accurate (it may still undercount), but the next `_enforce_size_cap` call will re-scan and correct the drift. The key property is that undercount eventually self-corrects (the cap fires late), while negative never self-corrects.

leofang (commented):
Re-review after the latest round of updates (33 commits). All 21 previous review comments have been addressed — nice work.

One new finding on the incremental size tracker (`_tracked_size_bytes`) that can affect correctness: a `__delitem__` racing with `_enforce_size_cap`'s re-seed can make the tracker go permanently negative, which disables all future eviction. Fix is a one-line `max(0, ...)` clamp. See inline comment for details.

Everything else looks solid. The per-instance `_decide_driver()` caching is correct, the key derivation is collision-free, the `os.scandir` migration is clean, and the test coverage has improved substantially.

cuda_core/cuda/core/utils/_program_cache/_file_stream.py
cpcloud: Good catch. Fixed in c347f3c83a with the clamp you suggested.

Race walkthrough confirmed: if `_enforce_size_cap`'s scan runs AFTER `__delitem__`'s unlink (so its reseed value excludes the deleted entry), the subtract here would undercount by `size`. The clamp leaves the tracker in the self-healing range — worst case is undercounting reality, which the next eviction's reseed corrects.

Added a regression test that simulates the post-reseed state directly and asserts the tracker stays at zero (`test_filestream_cache_tracker_clamps_at_zero_under_delete_race`).

leofang (approved):
Thanks a lot, Phillip. Merging.

PR #1897, opened 2026-04-13
[no-ci] cuda_core: fix SPDX identifiers and enforce Apache-2.0 headers

rwgk:
## Summary

- Correct the affected `cuda_core/` `SPDX-License-Identifier` headers so they consistently use `Apache-2.0`.
- Refactor `toolshed/check_spdx.py` so expected license identifiers can be enforced by path prefix, and require `Apache-2.0` for files under `cuda_core/`.
- Keep focused regression coverage in `toolshed/test_check_spdx.py` for the new `cuda_core/` rule, path normalization, and the existing staged-file copyright autofix flow.

## Validation

- `pre-commit run --all-files`
- `pixi run --manifest-path "cuda_pathfinder" pytest toolshed/test_check_spdx.py -q`
- Output after locally and temporarily reverting commit 7738635fac0cc2d3ec42f7e1f0e914ba6f5b9f1a:

```
$ pre-commit run --all-files
ruff check...............................................................Passed
ruff format..............................................................Passed
Check SPDX Headers.......................................................Failed
- hook id: check-spdx
- exit code: 1

INVALID SPDX license identifier 'LicenseRef-NVIDIA-SOFTWARE-LICENSE' (expected 'Apache-2.0') in 'cuda_core/tests/graph/test_graphdef.py'
INVALID SPDX license identifier 'LicenseRef-NVIDIA-SOFTWARE-LICENSE' (expected 'Apache-2.0') in 'cuda_core/cuda/core/_utils/driver_cu_result_explanations.py'
INVALID SPDX license identifier 'LicenseRef-NVIDIA-SOFTWARE-LICENSE' (expected 'Apache-2.0') in 'cuda_core/tests/graph/test_graphdef_lifetime.py'
INVALID SPDX license identifier 'LicenseRef-NVIDIA-SOFTWARE-LICENSE' (expected 'Apache-2.0') in 'cuda_core/cuda/core/_utils/clear_error_support.py'
INVALID SPDX license identifier 'LicenseRef-NVIDIA-SOFTWARE-LICENSE' (expected 'Apache-2.0') in 'cuda_core/tests/helpers/collection_interface_testers.py'
INVALID SPDX license identifier 'LicenseRef-NVIDIA-SOFTWARE-LICENSE' (expected 'Apache-2.0') in 'cuda_core/cuda/core/_utils/enum_explanations_helpers.py'
[...]

PR #1913, opened 2026-04-15
[no-ci] check_spdx.py: require explicit license decisions for top-level paths and related cleanup

rwgk:
# Accident note

The PR description below does **not** match what was actually merged into `main` in a8805e51.

GitHub merged the earlier branch state ending at c719fce4, which is what GitHub restored for `rwgk:expected_license_identifiers_cont`. The later follow-up changes described below were not pushed due to an unknown oversight.

Those later changes are now preserved for reference on `rwgk:expected_license_identifiers_cont_recovered` (see [comment below](https://github.com/NVIDIA/cuda-python/pull/1913#issuecomment-4271701561) for more details). Please read the remainder of this PR description as describing that recovered branch, not the merged result on `main`.

The missing commits were added to https://github.com/NVIDIA/cuda-python/pull/1948#issuecomment-4271805549 (needed there to guide fixing license identifiers).
___

Follow-on to PR #1897

## Summary

- Enforce an explicit SPDX license identifier for every top-level directory in the repository, so new paths can't slip in without a reviewed license decision.
- Keep top-level files on an explicit default of `Apache-2.0`.
- Continue to accommodate special cases, but treat them as exceptions we want to minimize rather than the default mechanism.
- Update `cuda_bindings/pixi.toml` to `LicenseRef-NVIDIA-SOFTWARE-LICENSE`.
- Fix the current non-benchmark special cases by updating
    - `.coveragerc`
    - `cuda_python_test_helpers/cuda_python_test_helpers/nvvm_bitcode.py`
    - `toolshed/build_static_bitcode_input.py`
    - `toolshed/dump_cutile_b64.py`

  to `Apache-2.0`.
- Keep `cuda_bindings/benchmarks/` as the only remaining special cases for now; those mixed-license cases will be handled in follow-on PRs (see [comment below](https://github.com/NVIDIA/cuda-python/pull/1913#issuecomment-4252968149)).

## Technical details

- Replace the old prefix-only SPDX policy handling in `toolshed/check_spdx.py` with explicit `TOP_LEVEL_DIRS_LICENSE_IDENTIFIERS` and `SPECIAL_CASE_LICENSE_IDENTIFIERS` mappings.
- Normalize repo-relative paths before matching so the checker behaves consistently across platforms.
- Make the checker fail when a file lives under a top-level directory that does not have an explicit expected license.
- Apply special cases by exact path or glob pattern, with longest-match wins so narrower overrides can coexist with broader rules.
- Improve license parsing by trimming common comment suffixes and decoding with replacement so malformed bytes still produce deterministic comparisons instead of being treated as missing headers.

### DIFF
--- a/.spdx-ignore
+++ b/.spdx-ignore
@@ -8,6 +8,9 @@ LICENSE
 requirements*.txt
 cuda_bindings/examples/*
 
+# Will be moved in (see https://github.com/NVIDIA/cuda-python/pull/1913#issuecomment-4252968149)
+cuda_bindings/benchmarks/*
+
 # Vendored
 cuda_core/cuda/core/_include/dlpack.h
 
--- a/cuda_bindings/pixi.toml
+++ b/cuda_bindings/pixi.toml
@@ -1,6 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
-# SPDX-License-Identifier: Apache-2.0
+# SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
 
 [workspace]
 channels = ["conda-forge"]
--- a/toolshed/check_spdx.py
+++ b/toolshed/check_spdx.py
@@ -17,7 +17,12 @@ SPDX_FILE_COPYRIGHT_TEXT_PREFIX = b"-".join((b"SPDX", b"FileCopyrightText: "))
 
 LICENSE_IDENTIFIER_REGEX = re.compile(re.escape(SPDX_LICENSE_IDENTIFIER_PREFIX) + rb"(?P<license_identifier>[^\r\n]+)")
 
-EXPECTED_LICENSE_IDENTIFIERS = (("cuda_core/", "Apache-2.0"),)
+EXPECTED_LICENSE_IDENTIFIERS = (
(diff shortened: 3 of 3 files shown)


danielfrg:
We agreed to move it the benchmarks to `{repo}/benchmarks/{pkg}` and that way we can keep the Apache license. I'll move them on my next PR for the benchmarks.

mdboom (approved):

rwgk:
**Forensics note on this PR**

I dug into why `toolshed/check_spdx.py` on `main` did not match the state I remembered locally.

The key point is that GitHub restored this PR branch exactly as GitHub last saw it at merge time. The restored branch `rwgk:expected_license_identifiers_cont` comes back at c719fce4, and the PR still shows only the early commits that were present when this PR was merged into a8805e51.

I also recovered a previously local-only branch tip with additional follow-up commits (c656fba8, 8b807157, 3f86b82d, c31b4fcf) that never made it to GitHub, so they were not part of the merged PR.

I pushed that recovered history to `rwgk:expected_license_identifiers_cont_recovered` for reference. Most likely I either forgot to run `git push`, or I overlooked an error message when trying to switch branches. **Sorry for the accident.**

The practical result is that the later work to make the top-level SPDX decisions explicit, handle the mixed-license `cuda_bindings/benchmarks` case, and fix the remaining license identifiers existed only locally and did not actually land in `main` through this PR. I will add a note to the PR description, and will follow up separately with the missing changes on top of current `main`.

PR #1917, opened 2026-04-15
cuda.core.system: Add ProcessInfo APIs

mdboom:
These APIs are required to support rapidsai/cudf.


cuda_core/cuda/core/system/_process.pxi
```diff
@@ -0,0 +1,43 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+
+class ProcessInfo:
+    """
+    Information about a process using the GPU.
```
leofang: Nit: "using"? "running on"?

cuda_core/cuda/core/system/_process.pxi
```diff
@@ -0,0 +1,43 @@
...
+    def used_gpu_memory(self) -> int:
+        """
+        The amount of GPU memory (in bytes) used by the process.
+        """
+        return self._process_info.used_gpu_memory
+
+    @property
+    def gpu_instance_id(self) -> int:
+        """
+        The GPU instance ID for MIG devices.
```
leofang, line 31: Oh interesting. This is MIG-only? Can we detect it and raise for non-MIG cases?

cuda_core/cuda/core/system/_process.pxi
```diff
@@ -0,0 +1,43 @@
...
+
+        Only valid for processes running on MIG devices.
+        """
+        return self._process_info.gpu_instance_id
+
+    @property
+    def compute_instance_id(self) -> int:
+        """
+        The Compute instance ID for MIG devices.
```
leofang, line 42: Ditto

cuda_core/cuda/core/system/_process.pxi
mdboom: These are the cuda.h docs:

```
    unsigned int        gpuInstanceId;      //!< If MIG is enabled, stores a valid GPU instance ID. gpuInstanceId is set to
                                            //  0xFFFFFFFF otherwise.
    unsigned int        computeInstanceId;  //!< If MIG is enabled, stores a valid compute instance ID. computeInstanceId is set to
                                            //  0xFFFFFFFF otherwise.
```

So, yes, it would be nice to raise on this rather than returning 0xFFFFFFFF

leofang (approved):

Issue #1851, opened 2026-04-02
[BUG]: Test: The generated ptx will not be loadable by the current driver.


mdboom:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Runtime Error

### Component

cuda.core

### Describe the bug

Occasionally in testing the following error is displayed:

```
RuntimeWarning: The CUDA driver version is older than the backend version. The generated ptx will not be loadable by the current driver.
```

This can come from the `cuda_core/examples/jit_lto_fractal.py` test in the "Test win-64 / py3.11, 13.2.0, wheels, rtx4090 (WDDM)" configuration.

[Full log](https://github.com/NVIDIA/cuda-python/actions/runs/23875576811/job/69618831328)

### How to Reproduce

Happens occasionally in CI.

I think the root cause is that the CI runner is in an incorrect state -- perhaps there is a race condition between installing the driver and the runtime that we could fix by determining the version once and installing the same version of both.  

I think it is better to fix that root cause than to make the test skippable.

### Expected behavior

All tests should pass.

leofang:
The warning is accurate. 
- The installed driver is 13.0 ([see here](https://github.com/NVIDIA/cuda-python/actions/runs/23875576811/job/69618831328#step:6:19))
- The installed CUDA compiler libraries are from 13.2.
- PTX generated by a newer (13.2) compiler cannot be loaded (JIT'd) by an older (13.0) driver, since MVC does not apply to PTX

My question is: Shouldn't this always fail in the CI? Is there a flakiness?

rwgk:
> My question is: Shouldn't this always fail in the CI? Is there a flakiness?

I'm guessing it's just that most runners have a newer driver. We should probably compare the driver version with the cuda-bindings version, and skip the test if the driver is older than the bindings.

rwgk:
Here is another flake of the same kind:

https://github.com/NVIDIA/cuda-python/actions/runs/24330602768/job/71036864789?pr=1898

**Test win-64 / py3.11, 13.0.2, wheels, rtx4090 (WDDM)**

```
================================== FAILURES ===================================
______________________ test_example[jit_lto_fractal.py] _______________________

example = 'jit_lto_fractal.py'

    @pytest.mark.parametrize("example", sample_files)
    def test_example(example):
[...]

PR #1921, opened 2026-04-15
bug fix: constrain cuda.core test installs to the requested CTK minor

rwgk:
xref: https://github.com/NVIDIA/cuda-python/issues/1851#issuecomment-4238025578

Fix latent bug discovered while reviewing #1851 with the team.

Without the explicit `cuda-toolkit==${CUDA_VER_MINOR}.*` pin, `pip` can satisfy the broad major-version requirements with a newer CUDA Toolkit minor than the one selected by the CI matrix. In practice, that can either pull a newer nvidia-cuda-runtime than the local CTK provides or make the wheel-backed job silently test the wrong CTK minor, reducing the value of the matrix and masking minor-version-specific regressions.

## Changes

- pin the `cuda-toolkit` dependency in `ci/tools/run-tests` to the matrix-selected CUDA toolkit minor for `cuda.core` test installs
- apply that pin in both the local-CTK and wheel-backed install paths so pip does not resolve a newer runtime than the job is meant to cover
- simplify the install flow by selecting the wheel target once via `WHL_EXTRA`, only adding the `cu${TEST_CUDA_MAJOR}` extra when the run needs published `cuda-bindings`

### DIFF
--- a/ci/tools/run-tests
+++ b/ci/tools/run-tests
@@ -74,14 +74,14 @@ elif [[ "${test_module}" == "core" ]]; then
 
   pushd ./cuda_core
   CUDA_VER_MINOR="$(cut -d '.' -f 1-2 <<< "${CUDA_VER}")"
+  # Constrain cuda-toolkit to the requested CTK version to avoid
+  # pip pulling in a newer nvidia-cuda-runtime that conflicts with it.
   if [[ "${LOCAL_CTK}" == 1 ]]; then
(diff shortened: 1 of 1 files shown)


rwgk:
@leofang @mdboom for visibility (we found this together in our review meeting on Monday)

rparolin (approved):

leofang:
LGTM, thx Ralf

PR #1927, opened 2026-04-16
cuda.bindings: Update generated code

mdboom:
This catches `cuda.bindings` up with a number of recent changes to the code generator.

These are

- whitespace
- documentation fixes, particularly matching documentation to the correct arguments and return values
- Fixing the `ind_ex` argument naming bug


leofang (approved):
LGTM! I consider the changes here to be bug-fix.

PR #1929, opened 2026-04-16
Add Cython ABI checking tool to toolshed

mdboom:
This is a rehash of the ABI-checking tool in #1067.  That PR was much more ambitious, and aimed to automatically detect changes to the ABI in CI, but required committing ABI descriptions to the repo, so was ultimately too much burden.

This just adds a commandline script that we can run when we want to compare the ABI of two versions of cuda-python.  Checking it in to `toolshed` makes it easier to find.

Though there is nothing cuda-python specific about it -- it could also be a standalone package.


toolshed/check_cython_abi.py
```diff
@@ -0,0 +1,167 @@
...
+     python check_cython_abi.py check <package_name> <dir>
+"""
+
+import importlib
+import json
+import re
+import sysconfig
+from pathlib import Path
+
+EXT_SUFFIX = sysconfig.get_config_var("EXT_SUFFIX")
```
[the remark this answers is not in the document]
mdboom: I think this is fine.  We don't need to run on platforms where this wouldn't work.

toolshed/check_cython_abi.py
```diff
@@ -0,0 +1,167 @@
...
+            if not abi_path.is_file():
+                print(f"New module added {so_path.relative_to(build_dir)}")
+                has_allowed_changes = True
+
+    if has_errors:
+        print("ERRORS FOUND")
+    elif has_allowed_changes:
+        print("Allowed changes found.")
+
+
```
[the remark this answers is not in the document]
mdboom: I think we should do the opposite -- just fix the type annotation.

toolshed/check_cython_abi.py
```diff
@@ -0,0 +1,178 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+
+"""
+Tool to check for Cython ABI changes in a given package.
```
leofang, line 7: Scope caveat worth adding to the docstring. Per [cython/cython#7603](https://github.com/cython/cython/issues/7603), Cython has **four** runtime ABI failure modes:

1. `cdef` function signatures (capsule strings) — *covered here*
2. `cdef class` struct size (`tp_basicsize`) — *not covered*
3. `cdef class` vtable layout / method reordering — *not covered*, and this one fails as **silent UB** rather than an import-time error
4. Fused specialization ordering — partially covered (reorders manifest as capsule-name deltas, but the mapping is non-obvious)

A one-liner pointing at #7603 would set expectations so users don't assume a clean run here means the full ABI is intact.

-- Leo's bot

toolshed/check_cython_abi.py
```diff
@@ -0,0 +1,178 @@
...
+
+def pyx_capi_to_json(d: dict[str, object]) -> dict[str, str]:
+    """
+    Converts the __pyx_capi__ dictionary to a JSON-serializable dictionary,
+    removing any memory addresses that are irrelevant for comparison.
+    """
+
+    def extract_name(v: object) -> str:
+        v = str(v)
+        match = re.match(r'<capsule object "([^\"]+)" at 0x[0-9a-fA-F]+>', v)
```
leofang, line 70: Optional robustness nit: `ctypes.pythonapi.PyCapsule_GetName` is the canonical way to read a capsule's name, and it's what the reference inspection snippet in [cython/cython#7603](https://github.com/cython/cython/issues/7603) uses. The regex works today because CPython's capsule `repr()` has been stable, but the direct C API removes that coupling:

```python
import ctypes
_get_name = ctypes.pythonapi.PyCapsule_GetName
_get_name.restype = ctypes.c_char_p
_get_name.argtypes = [ctypes.py_object]
name = _get_name(capsule).decode("utf-8")
```

Not a blocker — mentioning it since this tool is meant to outlive a handful of CPython/Cython revs.

-- Leo's bot

toolshed/check_cython_abi.py
```diff
@@ -0,0 +1,178 @@
...
+
+def check(package: str, abi_dir: Path) -> bool:
+    build_dir = get_package_path(package)
+
+    has_errors = False
+    has_allowed_changes = False
+    for abi_path in Path(abi_dir).glob(f"**/*{ABI_SUFFIX}"):
+        so_path = abi_path_to_so_path(abi_path, build_dir, abi_dir)
+        if so_path.is_file():
+            module = import_from_path(package, build_dir, so_path)
```
leofang, line 104: Non-blocking: if any one extension fails to import here (missing shared lib, init-time crash), the whole check run aborts. Same hazard at line 121 in this function and line 143 in `regenerate`. Copilot flagged this too. For a tool we may run over a whole package, a per-module `try/except` that records the failure and continues is more useful than a hard abort — especially in a "generate then diff" workflow where one flaky module shouldn't force the user to regenerate all baselines from scratch.

-- Leo's bot

toolshed/check_cython_abi.py
```diff
@@ -0,0 +1,178 @@
...
+
+    build_dir = get_package_path(package)
+    for so_path in Path(build_dir).glob(f"**/*{EXT_SUFFIX}"):
+        print(f"Generating ABI from {so_path.relative_to(build_dir)}")
+        module = import_from_path(package, build_dir, so_path)
+        if hasattr(module, "__pyx_capi__"):
+            abi_path = so_path_to_abi_path(so_path, build_dir, abi_dir)
+            abi_path.parent.mkdir(parents=True, exist_ok=True)
+            with open(abi_path, "w", encoding="utf-8") as f:
+                json.dump(pyx_capi_to_json(module.__pyx_capi__), f, indent=2)
```
leofang, line 148: Suggestion: consider stamping the Cython (and maybe Python) version into the generated JSON — either as a top-level `_meta` key alongside the symbols, or a sibling file. Capsule-name formatting has drifted across Cython versions in the past (typedef mangling in particular), so a user who regenerates under Cython X and checks under Y can see false-positive "ABI breakage" that is really just a format shift. At minimum, worth calling out in the docstring that `generate` and `check` should be run with the same Cython version.

-- Leo's bot

leofang (approved):
Thanks, Mike — I did an independent read and cross-referenced [cython/cython#7603](https://github.com/cython/cython/issues/7603) for the full ABI picture.

Approving. The core approach is right: `__pyx_capi__` capsule names are exactly what Cython's cross-module `cimport` string-compares at runtime, so snapshotting them is the correct signal for catching `cdef`-function ABI drift. I also agree with pulling back from #1067's ambition — no baselines checked in means no maintenance tax, and running this on demand against a tag is a sensible workflow.

Left four non-blocking suggestions inline. The one I'd most like to see before we lean on this heavily is a **scope caveat in the docstring** pointing at #7603, so future users don't read a clean run as "full ABI is intact" — it explicitly isn't covering `cdef class` layout or vtable reorders, and those are the silent-UB failure mode the Cython maintainers called out. The other three (direct `PyCapsule_GetName` vs regex on `repr`, per-module `try/except` on import, Cython version stamped in the JSON) are all "would be nice, can follow up."

Also +1 on declining Copilot's defensive `EXT_SUFFIX` fallback and the unchanged return signature on `check()` — right calls.

-- Leo's bot

toolshed/check_cython_abi.py
mdboom: That's a good callout.  3 (maybe 2 also) are an order of magnitude harder, because I think it requires parsing the `.pxd` files since the information is lost in the compiled artifact.

toolshed/check_cython_abi.py
mdboom: There is no direct way to get the Cython version from the compiled artifact.  You can look for the `string` like `_cython_3_4_2` in the module.  But if Cython decided to compile with compressed strings, that fails.

leofang: Oh I missed this, sorry @mdboom. This is doable similar to https://github.com/NVIDIA/cuda-python/pull/1817#discussion_r3047612886. We just stringify the Cython compiler version instead of the CUDA version.

PR #1937, opened 2026-04-17
nvbug6084457: Fix device architecture handling and NVLink link count query

mdboom:
Filing as a draft because this is still only a partial fix for the reported bug.  The final fix requires coordination with upstream NVML.


cuda_core/cuda/core/system/_device.pyx
```diff
@@ -165,7 +165,12 @@ cdef class Device:
         "VOLTA"``, and RTX A6000 will report ``DeviceArchitecture.name ==
         "AMPERE"``.
         """
-        return DeviceArch(nvml.device_get_architecture(self._handle))
+        arch = nvml.device_get_architecture(self._handle)
+        try:
+            arch = DeviceArch(arch)
+            return arch
```
cpcloud:
```suggestion
            return DeviceArch(arch)
```

Original seems overkill re assigning a variable and then immediately returning it.

mdboom:
Marking this as "ready to review".  As a follow-on when we do the 13.3 bring up, we will need to add logic to make `NVML_NVLINK_MAX_LINK` dynamic based on the CTK version, but no point in doing that now.

cuda_bindings/tests/nvml/test_nvlink.py
```diff
@@ -26,4 +26,4 @@ def test_nvlink_get_link_count(all_devices):
         # The feature_nvlink_supported detection is not robust, so we
         # can't be more specific about how many links we should find.
         if value.nvml_return == nvml.Return.SUCCESS:
-            assert value.value.ui_val <= nvml.NVLINK_MAX_LINKS, f"Unexpected link count {value.value.ui_val}"
+            assert value.value.ui_val[0] <= nvml.NVLINK_MAX_LINKS, f"Unexpected link count {value.value.ui_val[0]}"
```
rwgk, line 29: LGTM, based on my understanding that this fixes a subtle latent bug in the test: before, `value.value.ui_val` was a 1-element NumPy array, so this was doing an array-to-scalar comparison and relying on NumPy's size-1 truthiness. With `[0]`, the test now compares the actual scalar field value to `nvml.NVLINK_MAX_LINKS`, which I assume is the intended behavior.

cuda_core/cuda/core/system/_device.pyx
```diff
@@ -165,7 +165,11 @@ cdef class Device:
         "VOLTA"``, and RTX A6000 will report ``DeviceArchitecture.name ==
         "AMPERE"``.
         """
-        return DeviceArch(nvml.device_get_architecture(self._handle))
+        arch = nvml.device_get_architecture(self._handle)
+        try:
+            return DeviceArch(arch)
+        except ValueError:
+            return nvml.DeviceArch.UNKNOWN
```
rwgk, line 187: LGTM, based on my understanding that `nvml.device_get_architecture()` returns a raw integer architecture code, and `DeviceArch(arch)` is the enum conversion/validation step. This change seems to make `Device.arch` handle newer/unknown architecture IDs gracefully by returning `UNKNOWN` instead of raising `ValueError`. Please correct me if I'm missing any nuance.

How easy or difficult would it be to add a test that covers the `except` path?

cuda_bindings/tests/nvml/test_init.py
```diff
@@ -25,11 +25,18 @@ def test_devices_are_the_same_architecture(all_devices):
     # they won't be tested properly.  This tests for the (hopefully rare) case
     # where a system has devices of different architectures and produces a warning.
 
-    all_arches = {nvml.DeviceArch(nvml.device_get_architecture(device)) for device in all_devices}
+    def get_architecture_name(arch):
+        try:
+            arch = nvml.DeviceArch(arch)
+            return arch.name
+        except ValueError:
+            return f"UNKNOWN({arch})"
```
rwgk: Readability nit, to make the intent more obvious:

```
        try:
            arch = nvml.DeviceArch(arch)
        except ValueError:
            return f"UNKNOWN({arch})"
        return arch.name
```

Nit 2: `UNKNOWN_ARCH_ID`, so the resulting warning explains what the magic integer is (it can be guessed, this is just a little more helpful).

rwgk (approved):
LGTM, but since I'm not immersed in the context, it wasn't easy to be sure about the details. It'd be helpful to get an agent summary of the changes in the PR description.

PR #1942, opened 2026-04-17
Privatize system.Device helper classes

### LINKED ISSUE #1941: DOCS: properties do not link to their type
We have properties that return class instances where it would be very convenient to link directly to the type returned to see what members are available.  Unfortunately our autosummary template doesn't seem to do that, see, for example:

https://nvidia.github.io/cuda-python/cuda-core/latest/generated/cuda.core.system.Device.html#cuda.core.system.Device.memory_info

mdboom:
Also fixes #1941.

Remove helper/data-container classes from `cuda.core.system`'s public API (`__all__`) since they are not intended to be directly instantiated by users. These classes are returned by `Device` properties/methods and serve as nested data containers.

**Classes removed from `__all__`:** `BAR1MemoryInfo`, `ClockInfo`, `ClockOffsets`, `CoolerInfo`, `DeviceAttributes`, `DeviceEvents`, `EventData`, `FanInfo`, `FieldValue`, `FieldValues`, `GpuDynamicPstatesInfo`, `GpuDynamicPstatesUtilization`, `InforomInfo`, `PciInfo`, `RepairStatus`, `Temperature`, `ThermalSensor`, `ThermalSettings`.

Enums, exceptions, `Device` itself, and free functions remain public.

### Changes

- Remove helper classes from `__all__` in `_device.pyx`
- Update `api.rst` to document helper classes via their private module path (`system._device.ClassName`)
- Update `Device` property/method docstrings to link to helper classes using `:obj:` role
- Update tests to import helper classes from `_device` module

### DIFF
--- a/cuda_core/cuda/core/system/_clock.pxi
+++ b/cuda_core/cuda/core/system/_clock.pxi
@@ -129,7 +129,7 @@ cdef class ClockInfo:
 
         Returns
         -------
-        ClockOffsets
+        :obj:`~_device.ClockOffsets`
             An object with the min, max and current clock offset.
         """
         return ClockOffsets(nvml.device_get_clock_offsets(self._handle, self._clock_type, pstate))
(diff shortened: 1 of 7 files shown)


cpcloud (approved):

rwgk:
(I only glanced through. LGTM)

Issue #1945, opened 2026-04-17
cuda.core: method/member/property naming audit


mdboom:
## Summary

This is an audit of method/member/property naming in the `cuda_core` API.  (A subset of the larger audit work in #1919).

This is a bot-generated list, some of it good, some of it nonsense, and some of it requires some judgement calls.

Instructions for readers of this issue:

- Let's discuss in the comments below
- Once things have consensus, I will mark them with a :heavy_check_mark: or :x:.  I have taken the liberty of applying my own judgement as a first pass
- When the change has been merged, the checkbox will get checked

Based on this prompt:

---

I want to audit the naming of all members of every public class in cuda_core. You can ignore the private classes, I only care about properties and methods exposed through the public API.

- Check for PEP8 compliance
- Properties should be nouns
- Methods should be verbs
- Identify methods that could be properties (i.e. take no arguments and do minimal calculation)
- Check for any naming inconsistencies, such as when different english words or abbreviation forms are used for the same concept or datatype
- Make sure conceptual method pairs are matching, e.g. add/remove, get/set, create/delete

---

## Methods that could be properties

These take no arguments and return a value. Converting to properties would match existing patterns.

- [ ] `Buffer.get_ipc_descriptor()` → property `ipc_descriptor`
- [ ] `Event.get_ipc_descriptor()` → property `ipc_descriptor`
- [ ] `DeviceMemoryResource.get_allocation_handle()` → property `allocation_handle`
- [ ] `PinnedMemoryResource.get_allocation_handle()` → property `allocation_handle`
- [ ] `Linker.get_error_log()` → property `error_log`
- [ ] `Linker.get_info_log()` → property `info_log`
- [ ] `GraphDef.nodes()` → property `nodes`
- [ ] `GraphDef.edges()` → property `edges`

### `system` subpackage — no-arg methods that could be properties

- [ ] :heavy_check_mark: `system.Device.get_auto_boosted_clocks_enabled()` → property
- [ ] :heavy_check_mark: `system.Device.get_current_clock_event_reasons()` → property
- [ ] :heavy_check_mark: `system.Device.get_supported_clock_event_reasons()` → property
- [ ] ❌ `system.Device.get_supported_event_types()` → property
- [ ] :heavy_check_mark: `system.Device.get_supported_pstates()` → property
- [ ] :heavy_check_mark: `system.PciInfo.get_pcie_replay_counter()` → property
[...]

PR #1946, opened 2026-04-17
cuda.core.system: Naming improvements suggested in #1945

mdboom:
This applies the non-controversial renaming suggestions in #1945 within cuda.core.system.


cuda_core/cuda/core/system/_device.pyx
```diff
@@ -215,23 +215,23 @@ cdef class Device:
...
-    def is_c2c_mode_enabled(self) -> bool:
+    def is_c2c_enabled(self) -> bool:
         """
         Whether the C2C (Chip-to-Chip) mode is enabled for this device.
         """
         return bool(nvml.device_get_c2c_mode_info_v(self._handle).is_c2c_enabled)
 
     @property
-    def persistence_mode_enabled(self) -> bool:
+    def persistence_mode(self) -> bool:
```
rwgk: This is the only one that stands out to me as odd.

To be similar to `is_c2c_enabled` above, I'd expect `is_persistence_enabled` here. But really, `is_persistence_mode_enabled` would seem clearer. Then again, why are we dropping `mode` in `is_c2c_mode_enabled`? Overall my preference is:

```
is_c2c_mode_enabled
is_persistence_mode_enabled
```

leofang: Yes our convention for bool properties is `is_<something>`

cpcloud (approved):

cuda_core/cuda/core/system/_device.pyx
mdboom: Good callout.  I will change this to `is_persistence_mode_enabled` before merging.

PR #1952, opened 2026-04-17
[no-ci] Check struct and class changes in ABI checking tool

mdboom:
I had to push my agent really hard to use an actual parser and not build up crazy regexes, but I think the result is pretty ok.

This will now detect changes in the size or content of structs and also the sizes of `cdef classes`.

### DIFF
--- a/toolshed/check_cython_abi.py
+++ b/toolshed/check_cython_abi.py
@@ -6,12 +6,17 @@
 """
 Tool to check for Cython ABI changes in a given package.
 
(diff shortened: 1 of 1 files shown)


leofang:
Q: So this addition allows for checking a C struct's layout & size changes?

mdboom:
> Q: So this addition allows for checking a C struct's layout & size changes?

Yes.  Basically any change to a struct gets reported as an error.

leofang (approved):

PR #1936, opened 2026-04-17
WIP: pathfinder_with_compatibility_checks_v0

rwgk:
Continued under #1977, which started with systematically renaming `WithCompatibilityChecks` → `CompatibilityGuardRails`.

xref: #1038

WIP — Please don't look yet. I just want to see the CI results.

PR #1953, opened 2026-04-19
[FEA]: Add CUDA driver version query to `cuda.pathfinder` (private API)

rwgk:
xref: #1038, #1936

- add a small `cuda.pathfinder._utils.driver_info` helper backed by `cuDriverGetVersion()` so `cuda.pathfinder` can query the CUDA-facing driver version
- introduce `DriverCudaVersion`, `QueryDriverCudaVersionError`, and cached `query_driver_cuda_version()`, with naming and docstrings that make it explicit this matches `nvidia-smi`'s `CUDA Version`
- ~re-export the new API from `cuda.pathfinder`, document it in `cuda_pathfinder/docs/source/api.rst`, and add focused unit, real-driver, and public-import regression tests~

___

All new `INFO` lines are identical across all CI test logs:

https://github.com/NVIDIA/cuda-python/actions/runs/24639785901/attempts/1?pr=1953

```
INFO test_real_query_driver_cuda_version: driver_version=13.0 (encoded=13000)
```


cuda_pathfinder/cuda/pathfinder/__init__.py
leofang, line 1: I am confused. Why wouldn't people just use cuda.bindings or cuda.core for this purpose? Why are we duplicating the functionality in cuda-pathfinder?

leofang (changes requested):

cuda_pathfinder/cuda/pathfinder/__init__.py
rwgk: > I am confused. Why wouldn't people just use cuda.bindings or cuda.core for this purpose? Why are we duplicating the functionality in cuda-pathfinder?

cuda-bindings depends on cuda-pathfinder. We'd get a dependency cycle:

Note that I need this functionality for #1038 ("scoped" is not the main theme anymore, "compatibility checks" is, but I didn't want to change the title of the issue).

Another aspect: cuda-pathfinder is meant to require pure Python only (no compilation). This ties in directly with: cuda-pathfinder is meant to be usable as a build-time dependency (e.g. build hooks).

Luckily, the "duplication" is just a handful of ctypes-based lines of code.

leofang: Dropping a quick note based on concerns I raised during team sync. 

We definitely do not want to introduce any dependency cycle. However, we also don't want to expand the API surface unnecessarily. It is confusing to have every level of the CUDA Python stack (pathfinder, bindings, core, ...) expose the same API. 

As far as #1038 (and general pathfinder needs) is concerned, the driver version query is a mean to the end that can be hidden as implementation detail.

rwgk: Done: commit a450104ec6e0ec3f903678b3ae8c5c19291199c5

I removed the public re-exports from `cuda.pathfinder` and kept the driver-version query as an internal implementation detail.

To note the tradeoff I see: this does leave a potentially useful capability unavailable to users who may want to stay with `cuda.pathfinder` alone and avoid a `cuda-bindings` dependency, especially in lighter-weight or build-time contexts where `cuda.pathfinder` is already usable from `build_hooks.py`. A concern I still have is that, if we do not offer a small supported pathfinder-level entry point here, similar needs may become more likely to get handled with quick ad-hoc solutions in downstream code, which can unintentionally create duplication and tech debt. I’m fine with taking the narrower route here for now, but wanted to capture that consideration explicitly.

rwgk:
~~Oh, I forgot to remove the change in api.rst. Sec.~~

cuda_pathfinder/cuda/pathfinder/_utils/driver_info.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+@dataclass(frozen=True, slots=True)
+class DriverCudaVersion:
+    """
+    CUDA-facing driver version reported by ``cuDriverGetVersion()``.
+
+    The name ``DriverCudaVersion`` is intentionally specific: this dataclass
+    models the version shown as ``CUDA Version`` in ``nvidia-smi``, not the
+    graphics driver release shown as ``Driver Version``.
```
leofang: In other words, this class records the UMD version, not the KMD version

leofang (approved):
LGTM!

cuda_pathfinder/cuda/pathfinder/__init__.py
leofang: The counter argument is that build systems like CMake and setuptools (including our `build_hooks.py`) should never need to know the driver version. CUDA applications are meant to be buildable on CPU-only machines (which is what our CI does).

cuda_pathfinder/cuda/pathfinder/_utils/driver_info.py
rwgk: Thanks! I added commit 0d7c8f831bb730140cc861686bcee27d022903b8 to add that information on.

PR #1900, opened 2026-04-13
Replace nvrtc Tempita templates with pre-generated Cython files

leofang:
## Summary

Replace NVRTC's Tempita `.pyx.in` / `.pxd.in` templates with pre-generated plain Cython files, and update the build system to skip pyclibrary header parsing for NVRTC.

**Generated artifacts (all Tempita-free):**
- Remove 6 old `.pyx.in` / `.pxd.in` template files
- Add 7 pre-generated `.pyx` / `.pxd` files following cybind convention
- Internal layer moved from `_bindings/` to `_internal/` with platform split (`nvrtc_linux.pyx` / `nvrtc_windows.pyx`)

**Build system (`build_hooks.py`):**
- Remove `nvrtc` from `_REQUIRED_HEADERS` — no pyclibrary parsing needed
- Update `sources_list`: `cynvrtc.pyx` source path changed from `_bindings/` to `bindings/`
- `_rename_architecture_specific_files` already handles `_internal/nvrtc_linux.pyx` → `_internal/nvrtc.pyx`

This is a companion PR to cybind !369. When driver/runtime are also migrated, pyclibrary can be removed from `pyproject.toml` `build-system.requires` entirely.

## Test plan

- [ ] Review generated files match expected cybind convention
- [ ] Verify zero Tempita in all generated files
- [ ] Build with CUDA toolkit to verify Cython compilation succeeds
- [ ] Run nvrtc tests

-- Leo's bot

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1957, opened 2026-04-21
Backport of #1900 to 12.9.x

mdboom:
See #1900 for details.

PR #1958, opened 2026-04-21
Update .gitignore after #1900

mdboom:
I missed that we should probably have updated this in #1900, since these files are now "ok to be committed".  Not critical, but git tooling could get confused without this.


rwgk (approved):

rwgk:
Using `[no-ci]` in the PR title would seem appropriate (CI will still run on main).

leofang:
Ah... we need to merge #1957 first

PR #1962, opened 2026-04-22
Faster benchmarking smoke tests

mdboom:
The benchmarking smoke tests, while super valuable for making sure our benchmarks don't bitrot, are taking ~20min out of the ~27min total run time for CI.

This reduces the smoke test time to 13s.  Running each benchmark exactly once should be enough to detect any obvious breakage.

Hopefully this should still be good enough.  What do you think, @danielfrg?


danielfrg (approved):

danielfrg:
Yes, so much better like this!

PR #1693, opened 2026-02-26
Add support for `pathfinder.load_nvidia_dynamic_lib("cupti")`

rwgk:
Closes #1572

Complete support for (Linux, Windows) x (site-packages, Conda, CTK)

The CTK and site-packages support is fully tested in the cuda-python CI (`all_must_work`).

The Conda support is based on inspecting local Conda installations: (Linux, Windows) x (CTK 12, CTK 13)

In the cuda-python CI, Conda support is tested via mocks only. Additional manual testing:

```
smc120-0009.ipp2a2.colossus.nvidia.com:/wrk $ conda activate py314_cupti12

(py314_cupti12) smc120-0009.ipp2a2.colossus.nvidia.com:/wrk $ pip install cuda_pathfinder-1.4.1.dev31+g40f60225d-py3-none-any.whl
Processing ./cuda_pathfinder-1.4.1.dev31+g40f60225d-py3-none-any.whl
Installing collected packages: cuda-pathfinder
Successfully installed cuda-pathfinder-1.4.1.dev31+g40f60225d

(py314_cupti12) smc120-0009.ipp2a2.colossus.nvidia.com:/wrk $ python -c 'from cuda import pathfinder; print(pathfinder.__version__); print(pathfinder.load_nvidia_dynamic_lib("cupti"))'
1.4.1.dev31+g40f60225d
LoadedDL(abs_path='/wrk/miniforge3/envs/py314_cupti12/lib/libcupti.so.12', was_already_loaded_from_elsewhere=False, _handle_uint=94773688522640, found_via='conda')
```

```
smc120-0009.ipp2a2.colossus.nvidia.com:/wrk $ conda activate py314_cupti13

(py314_cupti13) smc120-0009.ipp2a2.colossus.nvidia.com:/wrk $ pip install cuda_pathfinder-1.4.1.dev31+g40f60225d-py3-none-any.whl
Processing ./cuda_pathfinder-1.4.1.dev31+g40f60225d-py3-none-any.whl
Installing collected packages: cuda-pathfinder
Successfully installed cuda-pathfinder-1.4.1.dev31+g40f60225d

(py314_cupti13) smc120-0009.ipp2a2.colossus.nvidia.com:/wrk $ python -c 'from cuda import pathfinder; print(pathfinder.__version__); print(pathfinder.load_nvidia_dynamic_lib("cupti"))'
1.4.1.dev31+g40f60225d
[...]

PR #1966, opened 2026-04-23
fix(pathfinder): Linux .so glob fallback prefers newest (#1732)

### LINKED ISSUE #1732: cuda_pathfinder: evaluate Linux glob fallback behavior when multiple .so versions match
Follow-up to PR review discussion: https://github.com/NVIDIA/cuda-python/pull/1693#discussion_r2896777451

## Context

In `cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_platform.py`, Linux fallback search does:

- try exact `lib<name>.so`
- otherwise glob `lib<name>.so*`
- sort matches and return the first one

This is deterministic, but when multiple candidates exist (for example both `libcupti.so.12` and `libcupti.so.13`), selection policy is not explicitly documented as a product decision.

## Decision to make

Does the current behavior make sense for pathfinder, or should fallback selection use a different policy?

Possible policies:

1. Keep current behavior (deterministic first match) and document assumptions.
2. Prefer newest candidate when multiple matches exist.
3. Treat multiple matches as ambiguous and raise an explicit error.

## Notes

- CTK installs typically provide unversioned `.so` symlink chains, so fallback may rarely be used there.
- Conda packaging (notably CUPTI) may expose only versioned files, so fallback can be relevant.

## Done criteria

- Team agrees on intended fallback behavior.
- Code comments/tests reflect that intent.
- If needed, implementation is updated in a follow-up PR.

cpcloud:
Resolves #1732.

## Summary

- `cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_platform.py`:
  `_find_so_in_rel_dirs` (the site-packages wheel lookup) now uses
  `sorted(..., reverse=True)` on its versioned-`.so` glob, matching the
  newest-first policy already used by `LinuxSearchPlatform.find_in_lib_dir`
  (the conda / `CUDA_PATH` lookup) and by the descriptor-driven
  already-loaded / system-search order in `load_dl_linux`/`load_dl_windows`.
- Added an inline comment at `_find_so_in_rel_dirs` documenting why the
  fallback exists, that a single match is expected in practice, the
  newest-first tie-breaker, and a pointer back to this issue.
- Appended the same issue pointer to the existing comment at
  `LinuxSearchPlatform.find_in_lib_dir`.
- `cuda_pathfinder/tests/test_search_steps.py`: added six tests (three per
  call site) locking in the newest-first policy: zero / one / multiple
  versioned matches at both the site-packages and conda/`CUDA_PATH` call
  sites, using `libcudart.so.12` / `libcudart.so.13` as the fixtures.

Behavioral changes to ambiguity handling (raising when multiple versioned
files coexist) remain deferred to #1038, consistent with the triage
decision recorded on #1732.

Original review thread:
https://github.com/NVIDIA/cuda-python/pull/1693#discussion_r2896777451

## Test plan

- [x] `pixi run pytest tests/` from `cuda_pathfinder/` — 966 passed, 4
      skipped (unchanged count except +6 new tests).
- [x] Ruff check + format pass on changed files.
- [x] Mypy on the changed source file — no new errors (same pre-existing
      set).
- [x] Sanity-checked that the new multi-match test fails on the
      pre-change sort by reverting the keyword locally.

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_platform.py
+++ b/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_platform.py
@@ -43,10 +43,16 @@ def _find_so_in_rel_dirs(
     for rel_dir in rel_dirs:
         sub_dir = tuple(rel_dir.split(os.path.sep))
         for abs_dir in find_sub_dirs_all_sitepackages(sub_dir):
+            # Exact unversioned match first; fall back to versioned names because some
+            # distros only ship lib<name>.so.<major> (e.g. conda libcupti). Only one match
+            # is expected in practice. Sort in reverse so the newest-sorting name wins if
+            # multiple coexist, matching the newest-first bias elsewhere in pathfinder
+            # (see LinuxSearchPlatform.find_in_lib_dir and load_dl_linux._candidate_sonames).
+            # Issue #1732 tracks the deferred question of raising on true ambiguity.
             so_name = os.path.join(abs_dir, so_basename)
             if os.path.isfile(so_name):
                 return so_name
-            for so_name in sorted(glob.glob(os.path.join(abs_dir, file_wild))):
+            for so_name in sorted(glob.glob(os.path.join(abs_dir, file_wild)), reverse=True):
                 if os.path.isfile(so_name):
                     return so_name
         sub_dirs_searched.append(sub_dir)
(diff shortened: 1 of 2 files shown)


rwgk (approved):
Thanks!

Issue #971, opened 2025-09-15
CI: Test `CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM=1`


leofang:
It appears that we haven't tested this at all. 

We should run this test in nightly or weekly tests, by turning on `CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM=1` and running all tests of cuda-bindings and cuda-core.

related: #133

leofang:
Mike added the PTDS tests in #1972, but unfortunately it actually has no effect because of a typo (https://github.com/NVIDIA/cuda-python/pull/1972/files#r3502934170). I'll fix it.

Issue #133, opened 2024-09-26
RFC: Enable per-thread default stream in free-threading builds


leofang:
tl;dr: For the Python 3.13 free-threading build (`cp313t`), the per-thread default stream is enabled and used by default. Users need to set `CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM=0` to explicitly opt out and restore the old behavior.

In CUDA, there are two kinds of default streams:
- Legacy default stream (synchronizing all blocking streams)
   - Unless some action is done as per the CUDA Programming Guide, this is the default. Most of the time the null/0 stream is a synonym of the legacy default stream
- Per-thread default stream (only synchronizing with the legacy default stream)

Today, CUDA Python offers a way to switch between the legacy and per-thread default streams (at the time of loading driver symbols) via the environment variable `CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM`, and the default has been as if it is set to 0 (so using the legacy default stream). 

However, it is a very common pitfall for performance-seeking applications and users who find themselves needing to create nonblocking streams explicitly to avoid implicit synchronization. This change would lift the need of creating nonblocking streams. This change would also allow GPU workloads launched from different host threads -- without an explicit stream in use -- to have an opportunity of overlapping and executing in parallel, instead of being serialized on the same (legacy default) stream.

The free threading build offers a natural opportunity and perfect timing for us to change the default to as if the env var is set to 1 and using the per-thread default stream. This also gives NVIDIA a path forward to assess the feasibility of deprecating (and eventually removing!) the legacy default stream, which has been a long-time quest we seek to conquer.

Users who use the regular build will not be affected, only those testing the experimental `cp313t` free-threading build will.

gmarkall:
From the perspective of numba-cuda I think this is a positive change and a good chance to make the change to the default. Numba does already support PTDS with the user explicitly setting it: https://numba.readthedocs.io/en/stable/reference/envvars.html#envvar-NUMBA_CUDA_PER_THREAD_DEFAULT_STREAM

The user can also explicitly force which kind of default stream they want: https://numba.readthedocs.io/en/stable/cuda-reference/host.html#numba.cuda.per_thread_default_stream / https://numba.readthedocs.io/en/stable/cuda-reference/host.html#numba.cuda.legacy_default_stream but I think this is orthogonal to what the default is.

bdice:
[...]

PR #1972, opened 2026-04-24
Move driver and nvrtc cython and internal layers to new generator

mdboom:
This is a continuation of the work in #1900.  Now adds `driver` to the mix and both `nvrtc` and `driver` are generated from the "real" new generator.


cuda_bindings/cuda/bindings/_internal/driver_windows.pyx
```diff
@@ -0,0 +1,8671 @@
...
+import threading
+from .utils import FunctionNotFoundError, NotSupportedError
+
+from cuda.pathfinder import load_nvidia_dynamic_lib
+
+from libc.stddef cimport wchar_t
+from libc.stdint cimport uintptr_t
+from cpython cimport PyUnicode_AsWideCharString, PyMem_Free
+
+# You must 'from .utils import NotSupportedError' before using this template
```
leofang, line 19: Note to self: check if this is from cybind template

cuda_bindings/cuda/bindings/_internal/driver_linux.pyx
```diff
@@ -0,0 +1,8653 @@
...
+        RTLD_LOCAL
+
+    const void* RTLD_DEFAULT 'RTLD_DEFAULT'
+
+cdef int get_cuda_version():
+    cdef void* handle = NULL
+    cdef int err, driver_ver = 0
+
+    # Load driver to check version
+    handle = dlopen('libcuda.so.1', RTLD_NOW | RTLD_GLOBAL)
```
leofang, line 41: Q: why don't we use pathfinder here?

cuda_bindings/cuda/bindings/_internal/driver_linux.pyx
```diff
@@ -0,0 +1,8653 @@
...
+            raise RuntimeError('Failed to open cuda')
+
+        # Get latest __cuGetProcAddress_v2
+        global __cuGetProcAddress_v2
+        __cuGetProcAddress_v2 = dlsym(handle, 'cuGetProcAddress_v2')
+        if __cuGetProcAddress_v2 == NULL:
+            raise RuntimeError("Failed to get __cuGetProcAddress_v2")
+        _F_cuGetProcAddress_v2 = <__cuGetProcAddress_v2_T>__cuGetProcAddress_v2
+
+        if os.getenv('CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM', default=0):
```
leofang: Wouldn't this evaluate to True for `export CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM=0`?
```python
>>> if '0': print(123)
...
123
```

cuda_bindings/cuda/bindings/_internal/driver_linux.pyx
```diff
@@ -0,0 +1,8653 @@
...
+        handle = load_library()
+        if handle == NULL:
+            raise RuntimeError('Failed to open cuda')
+
+        # Get latest __cuGetProcAddress_v2
+        global __cuGetProcAddress_v2
+        __cuGetProcAddress_v2 = dlsym(handle, 'cuGetProcAddress_v2')
+        if __cuGetProcAddress_v2 == NULL:
+            raise RuntimeError("Failed to get __cuGetProcAddress_v2")
+        _F_cuGetProcAddress_v2 = <__cuGetProcAddress_v2_T>__cuGetProcAddress_v2
```
leofang, line 609: IIRC the old code has a path where we do dlsym to get unversioned symbols, is it no longer relevant?

cuda_bindings/cuda/bindings/_internal/driver_windows.pyx
```diff
@@ -0,0 +1,8671 @@
...
+    return result
+
+cdef inline void *GetProcAddress(uintptr_t hModule, const char* lpProcName) nogil:
+    return _GetProcAddress(<HMODULE>hModule, lpProcName)
+
+cdef int get_cuda_version():
+    cdef int err, driver_ver = 0
+
+    # Load driver to check version
+    handle = LoadLibraryExW("nvcuda.dll", NULL, LOAD_LIBRARY_SEARCH_SYSTEM32)
```
leofang, line 60: ditto, re: pathfinder

leofang (commented):
First wave of questions

cuda_bindings/cuda/bindings/_internal/driver_windows.pyx
mdboom: It's not from the cybind template, but it's added as part of the `windows_externs.pxd` snippet.

cuda_bindings/cuda/bindings/_internal/driver_linux.pyx
mdboom: This comes from `linux_externs.pxd` snippet in cybind and is pasted into every `_internal/X_linux.pyx` file that the generator generates.  We should probably update this to use `pathfinder`, but it would have implications beyond `cuda_bindings`, so I didn't want to touch it.

I have an open PR in cybind to move away from snippets, because I think we are seeing one of the downsides of that approach here.

Importantly, this is **dead code** in this file.  `get_cuda_version` is never called elsewhere -- it's a utility for libraries that need to get the cuda version before loading their own library.  (Again, that should probably get moved to pathfinder as well, but is orthogonal to this change).

cuda_bindings/cuda/bindings/_internal/driver_linux.pyx
mdboom: Yes, the old code had a path to fall back to `dlsym` if we can't load `cuGetProcAddress_v2`.  This now assumes it's available and works.  IIUC, the old code path was for old CTK's where that might not be available and we don't need it anymore.  But I think the best way would be to have QA run this across a range of things.  It would be great to drop that complexity if we could.  The complexity is not just another branch of this that uses `dlsym` -- it's from re-creating the versioned symbol mapping that `cuGetProcAddress_v2` handles for us.  It's a big part of what the old generator did that doesn't yet exist in the new generator.

cuda_bindings/cuda/bindings/_internal/driver_windows.pyx
mdboom: Same reason not to update this now as for Linux.

cuda_bindings/cuda/bindings/_internal/driver_linux.pyx
mdboom: That's good catch, and we should probably fix it.  Note this was copied directly from the existing code which also has this bug:

https://github.com/NVIDIA/cuda-python/blob/main/cuda_bindings/cuda/bindings/_bindings/cydriver.pyx.in#L522

cuda_bindings/cuda/bindings/_internal/_fast_enum.py
```diff
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
 #
-# This code was automatically generated across versions from 12.9.1 to 13.3.0, generator version 0.3.1.dev1719+g565f73f4e. Do not modify it directly.
+# This code was automatically generated across versions from 12.9.0 to 13.2.0, generator version 0.3.1.dev1630+gadce055ea.d20260422. Do not modify it directly.
```
leofang: Not sure whether the 13.3 -> 13.2 change hints a potential issue?

cuda_bindings/cuda/bindings/_lib/utils.pxi
```diff
@@ -155,22 +155,22 @@ cdef void * _helper_input_void_ptr(ptr, _HelperInputVoidPtrStruct *helper):
             raise TypeError("Provided argument is of type {} but expected Type {}, {} or object with Buffer Protocol".format(type(ptr), type(None), type(int)))
 
 
-{{if 'CUmemPool_attribute_enum' in found_types}}
```
leofang, line 158: Q: Do we also start removing the build-time check logics?

leofang: Sorry, ignore me, the answer is yes (and I started it in #1900)...

.gitignore
```diff
@@ -17,31 +17,21 @@ __pycache__/
...
 
 # CUDA Python specific (auto-generated)
-cuda_bindings/cuda/bindings/_bindings/cydriver.pxd
-cuda_bindings/cuda/bindings/_bindings/cydriver.pyx
 cuda_bindings/cuda/bindings/_bindings/cyruntime.pxd
 cuda_bindings/cuda/bindings/_bindings/cyruntime.pyx
 cuda_bindings/cuda/bindings/_bindings/cyruntime_ptds.pxd
 cuda_bindings/cuda/bindings/_bindings/cyruntime_ptds.pyx
-cuda_bindings/cuda/bindings/_internal/_nvml.pyx
 cuda_bindings/cuda/bindings/_internal/cufile.pyx
```
leofang, line 29: nit: we should add them here now that both are auto-gen'd?
```suggestion
cuda_bindings/cuda/bindings/_internal/driver.pyx
cuda_bindings/cuda/bindings/_internal/nvrtc.pyx
cuda_bindings/cuda/bindings/_internal/cufile.pyx
```

leofang:
LGTM overall, @mdboom it seems the build fails at cythonization again 😛

cuda_bindings/cuda/bindings/_lib/utils.pxi
mdboom: We need to finish `runtime` first.  That will be follow-on work.  This only deals with `driver` (and also moving `nvrtc` further along than before).

mdboom:
> LGTM overall, @mdboom it seems the build fails at cythonization again 😛

Yeah, my agent found something easy, but it just fixed Python 3.15 by breaking everyone else ;)  Need to work deeper.

.gitignore
mdboom: We only want to ignore the `.pyx` files that come from `.pyx.in` files.  `.pyx` files that are checked into cuda-python (even if generated by cybind) should not be on this list.

cuda_bindings/cuda/bindings/_internal/_fast_enum.py
mdboom: Turns out not.  This file shouldn't have been updated at all.  That's a bug in cybind, now fixed.

mdboom:
> > LGTM overall, @mdboom it seems the build fails at cythonization again 😛
> 
> Yeah, my agent found something easy, but it just fixed Python 3.15 by breaking everyone else ;) Need to work deeper.

It turns out the exception signature needs to match exactly when declaring function pointer types (which cuda-core does in one spot).  cython-gen was generating `except ?CUDA_ERROR_NOT_FOUND`.  cybind generates `except ?_CURESULT_INTERNAL_LOADING_ERROR`.  The frustrating thing is that these are equivalent values.  (There is a line `_CURESULT_INTERNAL_LOADING_ERROR = CUDA_ERROR_NOT_FOUND`)...  But Cython can't see through that.  The solution I arrived at was to add another override so cybind will generate `CUDA_ERROR_NOT_FOUND` for driver only.  (Since we have to make cuda-core work for cuda-bindings from both before and after this PR).  Yet another special case for these "legacy" core libraries, but not too bad, I suppose.

mdboom:
@leofang: This is passing all the tests now.  I think this is good to go (and let's merge the generator-side thing at roughly the same time, if we can).

.gitignore
leofang: Ah @mdboom I see the confusion. Like `cufile.pyx` & co below, they are all auto-generated at build time too (from `cufile_<platform>.pyx`). It's different from the `.pyx.in` templates but still is generated code, hence it'd be nice to be consistent and ignore the generated files.

leofang:
> It turns out the exception signature needs to match exactly when declaring function pointer types (which cuda-core does in one spot).

I assume that this also helps us avoid any potential (Cython) ABI breakage?

leofang (approved):

leofang:
I wish we have the ABI test running in the CI 😛 Eyeball'd as much as I can. Thanks, Mike! 

The CI flakiness should be fixed in #2195. I retried a few times here to make it green.

cuda_bindings/cuda/bindings/_internal/driver_linux.pyx
```diff
@@ -0,0 +1,8653 @@
...
+        _F_cuGetProcAddress_v2('cuIpcCloseMemHandle', <void **>&__cuIpcCloseMemHandle, 4010, ptds_mode, NULL)
+
+        global __cuMemHostRegister_v2
+        _F_cuGetProcAddress_v2('cuMemHostRegister', <void **>&__cuMemHostRegister_v2, 6050, ptds_mode, NULL)
+
+        global __cuMemHostUnregister
+        _F_cuGetProcAddress_v2('cuMemHostUnregister', <void **>&__cuMemHostUnregister, 4000, ptds_mode, NULL)
+
+        global __cuMemcpy
+        _F_cuGetProcAddress_v2('cuMemcpy', <void **>&__cuMemcpy, 4000, ptds_mode, NULL)
```
leofang: For ~48 symbols (every `cuMemcpy*`, `cuMemset*`, and `cuStream*` core op), the old code used **different `cudaVersion` args for the PTDS vs DEFAULT load paths** — `cuMemcpy` for example was loaded with `cudaVersion=7000` in the PTDS branch and `cudaVersion=4000` in the default branch. The new unified path uses the default-mode value (`4000`) for both. The PTDS flag should still steer the driver to the `_ptds` variant, but this is a real behavior change that I don't think CI covers today.

Could we run a manual pass with `CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM=1` on the driver range we support, and confirm the resolved function pointers actually point at the `_ptds`/`_ptsz` variants (and not the plain ones)? Easiest check would be to snapshot the `__cuMemcpy` / `__cuMemcpyAsync` / `__cuMemsetD8` etc. addresses from a fixture under both modes against a known-good build.

-- Leo's bot

cuda_bindings/cuda/bindings/_internal/driver_linux.pyx
```diff
@@ -0,0 +1,8653 @@
...
+        _F_cuGetProcAddress_v2('cuStreamGetFlags', <void **>&__cuStreamGetFlags, 5050, ptds_mode, NULL)
+
+        global __cuStreamGetId
+        _F_cuGetProcAddress_v2('cuStreamGetId', <void **>&__cuStreamGetId, 12000, ptds_mode, NULL)
+
+        global __cuStreamGetCtx
+        _F_cuGetProcAddress_v2('cuStreamGetCtx', <void **>&__cuStreamGetCtx, 9020, ptds_mode, NULL)
+
+        global __cuStreamGetCtx_v2
+        _F_cuGetProcAddress_v2('cuStreamGetCtx_v2', <void **>&__cuStreamGetCtx_v2, 12050, ptds_mode, NULL)
```
leofang: The new generator passes the literal `_v2`-suffixed name to `cuGetProcAddress_v2`, while the old generator passed the **base** name and let `cudaVersion` disambiguate:

```diff
- _F_cuGetProcAddress_v2('cuStreamGetCtx',    &__cuStreamGetCtx_v2, 12050, PTDS,      NULL)  # main, _bindings/cydriver.pyx.in:799
+ _F_cuGetProcAddress_v2('cuStreamGetCtx_v2', &__cuStreamGetCtx_v2, 12050, ptds_mode, NULL)  # this PR
```

Same pattern applies to `cuLaunchHostFunc_v2` (line 2090), `cuCtxGetDevice_v2`, `cuCtxSynchronize_v2`, `cuMulticastBindAddr_v2`, `cuMulticastBindMem_v2` — six symbols in total.

The documented `cuGetProcAddress` contract is to pass the base name; that's the whole point of the API and what gives us versioned-symbol mapping for free. Passing the literal versioned name happens to work today (the driver likely also indexes by the literal name), but it's not the documented contract and bypasses the version arg. If a future driver tightens resolution to base-name-only, these six symbols silently become `NULL` and we get a hard-to-diagnose crash — exactly the breakage `cuGetProcAddress` was designed to prevent.

This is also the "re-creating the versioned symbol mapping" piece you flagged as [missing from the new generator](https://github.com/NVIDIA/cuda-python/pull/1972#discussion_r3234692933) — could cybind be updated to emit the base name + `cudaVersion` and reconstruct the mapping, matching the old behavior?

-- Leo's bot

leofang (changes requested):
Re-reviewing the symbol-loading rework end-to-end against `main`. Two file-anchored concerns below, plus one branch-state issue:

**Missing CUDA 13.3 symbols.** The following are declared and loaded in `_bindings/cydriver.pyx.in` on `main` (all at `cudaVersion=13030`) but absent from every new file in this PR (`driver_linux.pyx`, `driver_windows.pyx`, `_internal/driver.pxd`, `cydriver.pyx`):

- `cuLogicalEndpoint{Create,Destroy,AddDevice,BindAddr,BindMem,Unbind,Export,Import,Query,GetLimits,IdRelease,IdReserve}` — 12 functions
- `cuStreamBeginRecaptureToGraph` — 1 function

Looks like the branch was regenerated before these landed on `main`. Calling any of them post-merge would be a hard breakage, so this needs a rebase + regen before going in. Not a design fix.

Cancelling my prior approval until that and the inline comments are addressed.

-- Leo's bot

cuda_bindings/cuda/bindings/_internal/driver_linux.pyx
mdboom: It totally makes sense to drop the version suffix on the names passed to `cuGetProcAddress_v1`.

But I don't understand this bit:

> This is also the "re-creating the versioned symbol mapping" piece you flagged as https://github.com/NVIDIA/cuda-python/pull/1972#discussion_r3234692933 — could cybind be updated to emit the base name + cudaVersion and reconstruct the mapping, matching the old behavior?

cuda_bindings/cuda/bindings/_internal/driver_linux.pyx
mdboom: I think the testing suggestion here makes sense.  I don't think we necessarily need to make the suggested change unless the testing suggests otherwise.

mdboom:
> > It turns out the exception signature needs to match exactly when declaring function pointer types (which cuda-core does in one spot).
> 
> I assume that this also helps us avoid any potential (Cython) ABI breakage?

It does.  It's overly strict (it wouldn't have been a breakage in this specific case), but at least if we /really/ broke it we know Cython would tell us.

cuda_bindings/cuda/bindings/_internal/driver_linux.pyx
mdboom: So, it turns out PTDS is broken by this PR.  Back to the drawing board on some of these details.

Would changing at least one CI config to use PTDS be ok?  It would be great to be testing it on a regular basis.

leofang: Yes, see #971. We'll eventually flip to PTDS for FT builds anyway (#133), so definitely need to test it.

mdboom:
@leofang: I have updated this to include the ptds version numbers where appropriate and also added ptds testing to a couple of the configurations in CI.

cuda_bindings/cuda/bindings/_internal/driver_linux.pyx
leofang: Probably another late night fart of mine (and my bot's)... I don't understand it anymore either lol

leofang:
@mdboom this regression is not yet addressed?

> **Missing CUDA 13.3 symbols.** The following are declared and loaded in `_bindings/cydriver.pyx.in` on `main` (all at `cudaVersion=13030`) but absent from every new file in this PR (`driver_linux.pyx`, `driver_windows.pyx`, `_internal/driver.pxd`, `cydriver.pyx`):
> 
> * `cuLogicalEndpoint{Create,Destroy,AddDevice,BindAddr,BindMem,Unbind,Export,Import,Query,GetLimits,IdRelease,IdReserve}` — 12 functions
> * `cuStreamBeginRecaptureToGraph` — 1 function

mdboom:
> @mdboom this regression is not yet addressed?
> 
> > **Missing CUDA 13.3 symbols.** The following are declared and loaded in `_bindings/cydriver.pyx.in` on `main` (all at `cudaVersion=13030`) but absent from every new file in this PR (`driver_linux.pyx`, `driver_windows.pyx`, `_internal/driver.pxd`, `cydriver.pyx`):
> > 
> > * `cuLogicalEndpoint{Create,Destroy,AddDevice,BindAddr,BindMem,Unbind,Export,Import,Query,GetLimits,IdRelease,IdReserve}` — 12 functions
> > * `cuStreamBeginRecaptureToGraph` — 1 function

Sorry, I didn't see this comment yesterday.  These have been added now.

leofang (approved):

ci/test-matrix.yml
```diff
@@ -33,7 +39,7 @@ linux:
     - { ARCH: 'amd64', PY_VER: '3.11',  CUDA_VER: '12.9.1', LOCAL_CTK: '0', GPU: 'rtxpro6000', GPU_COUNT: '1', DRIVER: 'latest' }
     - { ARCH: 'amd64', PY_VER: '3.11',  CUDA_VER: '13.0.2', LOCAL_CTK: '1', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest' }
     - { ARCH: 'amd64', PY_VER: '3.11',  CUDA_VER: '13.3.0', LOCAL_CTK: '1', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest' }
-    - { ARCH: 'amd64', PY_VER: '3.12',  CUDA_VER: '12.9.1', LOCAL_CTK: '1', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest' }
+    - { ARCH: 'amd64', PY_VER: '3.12',  CUDA_VER: '12.9.1', LOCAL_CTK: '1', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest', ENV: { CUDA_PYTHON_PER_THREAD_DEFAULT_STREAM: '1' } }
```
leofang, line 42: Turns out that this is a typo... 😞 It should have been `CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM`. I'll fix it...

Issue #214, opened 2024-11-04
Segregate existing Cython tests to be self-sufficient and disconnected from other Python tests


vzhurba01:
Until recently both the Cython and Python tests were part of the `cuda_bindings/tests` directory. By having them together, Python tests are implicitly blocked on having a successfully build for the Cython test-suite. I want to relax this dependency so that it is more natural to test one or the other.

Furthermore, these tests are only built when the the full cuda-bindings subproject gets built. This introduces further dependencies on the test-suite and slowly down testing if any pre-requisite is missed.

I propose segregating the two types of tests (Cython vs. Python) into their own directories while providing the Cython test-suite a dedicated setup.py for Cythonizing and building their tests.

PR #1974, opened 2026-04-24
Build cython tests as part of the main test suite

### LINKED ISSUE #1973: Include cuda_bindings Cython tests in main cuda_bindings test suite
The Cython tests in the `cuda_bindings` test suite currently require a separate build step, otherwise they don't run as part of the main `cuda_bindings` test suite.

This makes them easy to forget to run, and, as in my case, I didn't even know they existed.

Instead, we can do the build step as part of a test fixture. This pattern is already used in cybind so it can test Cython-compiled things.

Opening an issue in case there is something I'm missing as to why this can't be done.

mdboom:
Fix #1973.  It's too easy to forget tests that aren't part of the regular test-running suite.

This automatically builds these tests if precompiled objects aren't found.  This allows us to keep compiling on the build machine and testing on a machine without a compiler, while also allowing for more seamless local development.


leofang (changes requested):
Apology, @mdboom, for my recollection coming in the late night. It did not click for me until now.

If you don't mind, I would like to propose to kill this effort, because we split it for a very good reason: Having the Cython tests run by default messes up QA tests every single time. We ended up being chased after due to QA's insufficient/imperfect environment issues, as repeatedly discussed in recent team syncs.

A similar discussion can be found in the Oct-Nov 2024 team meeting minutes. We also had a public issue tracking the splitting: https://github.com/NVIDIA/cuda-python/issues/214. Let us not revert the split and take on more unnecessary burdens.

rwgk:
@leofang I'm not sure if this is relevant here, but I believe QA has been running the cython tests in the 13.2 test cycle, probably also in the 13.1 cycle. For sure, I have them in our reference scripts. Please let me know if finding out for sure (by asking the QA team for confirmation) would change your opinion about this PR.

leofang:
> Please let me know if finding out for sure (by asking the QA team for confirmation) would change your opinion about this PR.

Yes, it'd be nice to know if/how they run the Cython tests in the past few releases of both cuda-bindings and cuda-core.

rwgk:
> > Please let me know if finding out for sure (by asking the QA team for confirmation) would change your opinion about this PR.
> 
> Yes, it'd be nice to know if/how they run the Cython tests in the past few releases of both cuda-bindings and cuda-core.

The QA team will only be back at work on 5/5, but I just remembered, I'm sure they did run the examples: see nvbugs 5861207 comment 14. Offline I helped them find the commands they had to remove from their automation. So clearly, they were using them before.

mdboom:
Closing due to @leofang's comment.  I might file an alternative that just improve the docs a bit around testing, but if having Cython available at testing (as this approach would require) messes up QA, I agree, let's not do it.

PR #1976, opened 2026-04-25
Add green context support

### LINKED ISSUE #112: Support green contexts

### LINKED ISSUE #1563: Support allocating SMs

leofang:
Close #1563. Close #112.

## Summary

Add green context support to cuda.core — the explicit-model API for querying device resources, splitting SMs, creating green contexts, and using them without touching the thread-local context stack.

## Design

See the companion [design doc](https://docs.google.com/document/d/1fNlXLJw_oy4GLyUhzxlDO5aC_q2GYbekJs-DN3y-hWg/edit) for full rationale. Key decisions:

- **Unified `Context` type** — no user-visible `GreenContext` subclass. A single `Context` wraps either a primary `CUcontext` or a `CUgreenCtx` + derived `CUcontext`. `ctx.is_green` distinguishes them. Inspired by the CUDA runtime's execution-context (EC) abstraction.
- **`dev.resources` namespace** — `DeviceResources` groups hardware resource queries (`dev.resources.sm`, `dev.resources.workqueue`). Follows the existing "plural = namespace" pattern (`dev.properties`, `kernel.attributes`).
- **`ctx.resources` / `stream.resources`** — same `DeviceResources` type, but queries the context's provisioned resources (`cuCtxGetDevResource` / `cuGreenCtxGetDevResource`) instead of the full device.
- **`SMResourceOptions` with SoA broadcasting** — single dataclass for `SMResource.split()`. Scalar fields broadcast; `count` drives the group count. `count=None` means discovery mode (translated to `smCount=0` internally).
- **Merged workqueue types** — `WorkqueueResource` merges `CU_DEV_RESOURCE_TYPE_WORKQUEUE_CONFIG` and `CU_DEV_RESOURCE_TYPE_WORKQUEUE` under one user-facing class. Strings for option values (e.g. `sharing_scope="green_ctx_balanced"`).
- **`ContextOptions(resources=[...])` → `dev.create_context()`** — resource descriptor generation and `cuGreenCtxCreate` are internal. The user passes pre-split resource objects.
- **Explicit model** — `ctx.create_stream()` creates streams bound to a green context without calling `dev.set_current()`. The C++ handle layer auto-dispatches between `cuGreenCtxStreamCreate` and `cuStreamCreateWithPriority` based on the context type. Green context streams must be non-blocking.
- **`ctx.close()` does not manage the context stack** — closing a current context raises `RuntimeError`. `dev.set_current(green_ctx)` still works for backward compatibility but is not the recommended path.

## New public API

- `Device.resources` → `DeviceResources` (namespace: `.sm`, `.workqueue`)
- `Context.resources` → `DeviceResources` (context-level query of provisioned resources)
- `Stream.resources` → `DeviceResources` (delegates to the stream's parent context)
- `Context.create_stream(options)` → `Stream` (green contexts only; raises on primary)
- `Context.is_green` → `bool`
- `SMResource` — properties: `sm_count`, `min_partition_size`, `coscheduled_alignment`, `flags`, `handle`; method: `split(options, *, dry_run=False)`
- `SMResourceOptions` — `count`, `coscheduled_sm_count`, `preferred_coscheduled_sm_count`
- `WorkqueueResource` — method: `configure(options)`
- `WorkqueueResourceOptions` — `sharing_scope`
- `ContextOptions.resources` — accepts `Sequence[SMResource | WorkqueueResource]`

## Implementation details

**C++ handle layer** (`resource_handles.hpp/cpp`):
- `GreenCtxHandle` (`shared_ptr<const CUgreenCtx>`) — owning handle; destructor calls `cuGreenCtxDestroy`.
- `ContextBox` gains a `GreenCtxHandle` field so the derived `CUcontext` keeps the green ctx alive. `get_context_green_ctx()` provides reverse lookup.
- `create_green_ctx_handle()` combines `cuDevResourceGenerateDesc` + `cuGreenCtxCreate` in one call — the descriptor is transient (no `DevResourceDescHandle` needed since CUDA has no explicit destroy for it).
- `create_stream_handle()` auto-dispatches: checks `get_context_green_ctx()` on the provided `ContextHandle` and calls `cuGreenCtxStreamCreate` for green contexts, `cuStreamCreateWithPriority` for primary. Returns `CUDA_ERROR_NOT_SUPPORTED` if the context is green but `cuGreenCtxStreamCreate` is unavailable (CUDA < 12.5).
- `context_registry` / `stream_registry` (`HandleRegistry`) deduplicate handles by raw CUDA pointer, enabling identity-preserving `set_current` swaps.

**Bug fix — stream context tracking**:
- `StreamBox` now carries a `ContextHandle` dependency, populated at creation time.
- `get_stream_context()` returns it without a driver call.
- `Stream._from_handle` and `Stream_ensure_ctx` prefer the registry-backed handle before falling back to `cuStreamGetCtx`. This fixes a latent issue where streams created in a green context would lose their context association after a `set_current` swap.

**Error handling**:
- `dev.create_context()` without resources raises `ValueError` with a clear message.
- Green context stream creation with `nonblocking=False` is caught by the driver (`CUDA_ERROR_INVALID_VALUE`) and re-raised as `ValueError` with a helpful message.
- `cuCtxGetStreamPriorityRange` failure (`CUDA_ERROR_INVALID_CONTEXT`) raises "Call dev.set_current() before creating streams."

**Version guards**:
- Compile-time: `IF CUDA_CORE_BUILD_MAJOR >= 13` gates `cuDevSmResourceSplit` (the general/structured form).
- Runtime: trinary cached checks (supported / unsupported / unchecked). `cy_driver_version() >= (12, 4, 0)` for all green ctx APIs; `>= (13, 1, 0)` for structured splits. Raises `ValueError` when unsupported.
- CUDA 12.x fallback: `cuDevSmResourceSplitByCount` for basic (homogeneous) splits. Per-group `coscheduled_sm_count` and heterogeneous counts require 13.1+ and raise `NotImplementedError` on 12.x.
- Green ctx function pointers loaded via `_get_optional_driver_fn` — graceful `NULL` when bindings lack the symbol.

## Test coverage

33 tests in `test_green_context.py`, organized with proper pytest fixtures and classes:

- **Fixtures**: `sm_resource`, `wq_resource`, `green_ctx` (with `CUDAError` → skip), `fill_kernel`
- **`_use_green_ctx`** context manager for safe push/pop in set_current regression tests
- `TestSMResourceQuery` — properties, arch constraints (pre-Hopper vs Hopper+)
- `TestWorkqueueResource` — query, configure valid/invalid
- `TestSMResourceSplitValidation` — scalar/Sequence mismatch, negative count, dry-run blocked
- `TestSMResourceSplit` — single/two-group splits with arch-aligned counts, discovery mode, alignment, dry-run parity
- `TestGreenContextLifecycle` — `is_green`, `create_stream` on primary raises, blocking stream raises, explicit stream creation, stream/event context tracking, close-while-current guard, set_current regression
- `TestContextResources` — green ctx SM resources are subset of device, two contexts have disjoint partitions, stream.resources matches ctx.resources (SM + workqueue)
- `TestGreenContextKernelLaunch` — compile + launch + host-verify via `ctx.create_stream()`, two independent green contexts with different fill values, SM + workqueue combined

## Validation

```
CUDA_HOME=... pip install -e . --no-build-isolation
python -m pytest tests/test_green_context.py -v                          # 32 passed, 1 skipped (arch)
python -m pytest tests/test_device.py tests/test_stream.py tests/test_event.py tests/test_context.py -v  # no regressions (257 total passed)
```

-- Leo's bot


leofang:
/ok to test [ac5c0fc](https://github.com/NVIDIA/cuda-python/pull/1976/commits/ac5c0fc4f6c5668bee0b618de2d174b4e65b53e5)

leofang:
/ok to test [2812c5b](https://github.com/NVIDIA/cuda-python/pull/1976/commits/2812c5bacfd322ec5de24714f6cbfd109985d54e)

leofang:
/ok to test [fa254a5](https://github.com/NVIDIA/cuda-python/pull/1976/commits/fa254a514dbd1ec051e0673a9b4271bd770b8aee)

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -223,12 +229,112 @@ void clear_last_error() noexcept {
...
+        return {};
+    }
+
+    auto box = std::shared_ptr<const ContextBox>(
+        new ContextBox{ctx, h_green_ctx},
+        [](const ContextBox* b) {
+            context_registry.unregister_handle(b->resource);
+            delete b;
+        }
+    );
```
Andy-Jost: This needs to check the registry after calling `p_cuCtxFromGreenCtx`. I'd suggest delegating to `create_context_handle_ref`

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -250,14 +356,16 @@ ContextHandle get_primary_context(int device_id) {
     }
 
     auto box = std::shared_ptr<const ContextBox>(
-        new ContextBox{ctx},
+        new ContextBox{ctx, {}},
         [device_id](const ContextBox* b) {
+            context_registry.unregister_handle(b->resource);
             GILReleaseGuard gil;
```
Andy-Jost, line 358: nit: put the GIL release first for consistency in this file and to make it easier to spot.

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -286,33 +394,79 @@ ContextHandle get_current_context() {
...
+        }
     }
 
     auto box = std::shared_ptr<const StreamBox>(
-        new StreamBox{stream},
-        [h_ctx](const StreamBox* b) {
+        new StreamBox{stream, h_ctx},
+        [](const StreamBox* b) {
+            stream_registry.unregister_handle(b->resource);
             GILReleaseGuard gil;
```
Andy-Jost, line 431: nit: same as above

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -324,16 +478,23 @@ StreamHandle create_stream_handle_with_owner(CUstream stream, PyObject* owner) {
     }
     Py_INCREF(owner);
     auto box = std::shared_ptr<const StreamBox>(
-        new StreamBox{stream},
+        new StreamBox{stream, {}},
         [owner](const StreamBox* b) {
+            stream_registry.unregister_handle(b->resource);
             GILAcquireGuard gil;
```
Andy-Jost: nit: as above

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -286,33 +394,79 @@ ContextHandle get_current_context() {
...
 
 StreamHandle create_stream_handle_with_owner(CUstream stream, PyObject* owner) {
+    if (auto h = stream_registry.lookup(stream)) {
+        // Reuse handles that already carry structural context metadata, e.g.
+        // cuda-core-owned streams. Owner-backed foreign streams still need a
+        // fresh handle so the supplied owner is retained.
+        if (get_box(h)->h_context) {
+            return h;
+        }
+    }
```
Andy-Jost, line 464: This stream registry only allows one entry per unique `CUstream`, but there can be multiple `PyObject` owners, which will compete for the one and only cache slot. It can easily lead to registry corruption.

At a minimum, I don't think the stream + Python owner path should use a registry at all, since it is really just tying a `Py_INCREF`/`Py_DECREF` pair to a `CUstream` lifetime, which can be stacked any number of times.

More broadly, I'm having trouble understanding why a stream registry is needed at all. The PR description says:

> `Stream._from_handle` and `Stream_ensure_ctx` prefer the registry-backed handle before falling back to `cuStreamGetCtx`. This fixes a latent issue where streams created in a green context would lose their context association after a `set_current` swap.

I don't see how this comes into play. A registry allows a raw handle such as `CUstream` to round-trip through the driver without losing metadata, but `set_current` is about `CUcontext`, not `CUstream`. Where is the path that needs to match up a raw `CUstream` with an existing `StreamHandle`?

cuda_core/cuda/core/_context.pyx
```diff
@@ -32,17 +48,88 @@ cdef class Context:
...
     @property
     def _handle(self):
         return self.handle
 
+    @property
+    def is_green(self) -> bool:
+        """True if this context was created from device resources."""
+        if not self._h_context:
+            return False
+        return get_context_green_ctx(self._h_context).get() != NULL
```
Andy-Jost, line 78: nit: consider putting this in the handle API as `is_green(self._h_context)`.

cuda_core/cuda/core/_context.pyx
```diff
@@ -32,17 +48,88 @@ cdef class Context:
...
+        """Release this context wrapper's underlying CUDA handles."""
+        cdef cydriver.CUcontext current_ctx
+        if self._h_context and as_cu(self._h_context) != NULL:
+            with nogil:
+                HANDLE_RETURN(cydriver.cuCtxGetCurrent(&current_ctx))
+            if current_ctx == as_cu(self._h_context):
+                raise RuntimeError(
+                    "Cannot close a CUDA context while it is current. "
+                    "Restore a previous context before closing this context."
+                )
```
Andy-Jost, line 130: Long term, we should create something like a global `std::stack<ContextHandle>` representing the current context stack. Then we would not need code like this.

cuda_core/cuda/core/_device.pyx
```diff
@@ -1266,7 +1292,63 @@ class Device:
...
+            )
+
+        resources = tuple(options.resources)
+        if len(resources) == 0:
+            raise ValueError("ContextOptions.resources must not be empty")
+
+        n_resources = len(resources)
+        c_resources = <cydriver.CUdevResource*>malloc(
+            n_resources * sizeof(cydriver.CUdevResource)
+        )
```
Andy-Jost: nit: I'm guilty of the same thing, but consider `std::vector`

cuda_core/cuda/core/_cpp/resource_handles.cpp
Andy-Jost:
```diff
diff --git a/cuda_core/cuda/core/_cpp/resource_handles.cpp b/cuda_core/cuda/core/_cpp/resource_handles.cpp
index 2413d9473c..4b7b3ecfa3 100644
--- a/cuda_core/cuda/core/_cpp/resource_handles.cpp
+++ b/cuda_core/cuda/core/_cpp/resource_handles.cpp
@@ -245,9 +245,9 @@ static const ContextBox* get_box(const ContextHandle& h) noexcept {

 // See REGISTRY_DESIGN.md (Level 1: Driver Handle -> Resource Handle)
 static HandleRegistry<CUcontext, ContextHandle> context_registry;
-}  // namespace

-ContextHandle create_context_handle_ref(CUcontext ctx) {
+// Create a context handle reference, with optional green context as source.
+ContextHandle create_context_handle_ref(CUcontext ctx, GreenCtxHandle h_green_ctx) {
     if (!ctx) {
         return {};
     }
@@ -255,7 +255,7 @@ ContextHandle create_context_handle_ref(CUcontext ctx) {
         return h;
     }
     auto box = std::shared_ptr<const ContextBox>(
-        new ContextBox{ctx, {}},
+        new ContextBox{ctx, std::move(h_green_ctx)},
         [](const ContextBox* b) {
             context_registry.unregister_handle(b->resource);
             delete b;
@@ -265,6 +265,11 @@ ContextHandle create_context_handle_ref(CUcontext ctx) {
     context_registry.register_handle(ctx, h);
     return h;
 }
+}  // namespace
+
+ContextHandle create_context_handle_ref(CUcontext ctx) {
+    return create_context_handle_ref(ctx, {});
+}

 ContextHandle create_context_handle_from_green_ctx(const GreenCtxHandle& h_green_ctx) {
     if (!h_green_ctx) {
@@ -281,16 +286,7 @@ ContextHandle create_context_handle_from_green_ctx(const GreenCtxHandle& h_green
         return {};
     }

-    auto box = std::shared_ptr<const ContextBox>(
-        new ContextBox{ctx, h_green_ctx},
-        [](const ContextBox* b) {
-            context_registry.unregister_handle(b->resource);
-            delete b;
-        }
-    );
-    ContextHandle h(box, &box->resource);
-    context_registry.register_handle(ctx, h);
-    return h;
+    return create_context_handle_ref(ctx, h_green_ctx);
 }

 GreenCtxHandle get_context_green_ctx(const ContextHandle& h) noexcept {
```

Andy-Jost (changes requested):
I see a few issues with the registries. It appears the context registry needs to be checked in one place.

I have a bigger concern with the stream registry. I don't see what problem it solves, and it appears to be corruptible through the user API.

cuda_core/cuda/core/_cpp/resource_handles.cpp
leofang: @Andy-Jost correct me if I am wrong, I would think every type we own should have its own registry, why would we differentiate `CUcontext` from others? It is not only confusing but also unsafe in general, considering that we do want to handle the general cases for each object via `from_handle()` (#1989). 

For example, our `Buffer` internally tracks the allocation `Stream`, so having a buffer registry would help express the intent (though for buffer we may have other considerations such as performance).

> Where is the path that needs to match up a raw CUstream with an existing StreamHandle?

On way to trigger this is:
```python
# cuda.core creates a green stream (cuGreenCtxStreamCreate under the hood)
ctx = dev.create_context(...)
s = ctx.create_stream()

# User extracts the raw handle and re-wraps it (round-trip through int)
raw = s.handle
s2 = Stream.from_handle(raw)

# Without registry: s2 is a fresh handle with no context dep
# With registry: s2 reuses the existing StreamHandle, preserving
#   the ContextHandle → GreenCtxHandle dependency chain
s2.context.is_green  # True with registry, might fail without
```
so in this case a (stream) registry lookup is indeed useful.

> It can easily lead to registry corruption.

I think we do want to address the corruption issue. Would it help if we keep the registry but exclude the owner path (i.e. foreign streams don't register)?

cuda_core/cuda/core/_context.pyx
leofang: This is an overkill because by cuda-core's design our stack depth is either 0 or 1. We don't do push/pop 🙂

cuda_core/cuda/core/_cpp/resource_handles.cpp
leofang: Thanks, Andy! Done in commit 37ebe2a64b0720316fde96c3d072c614625fcaba.

cuda_core/cuda/core/_device.pyx
leofang: Done in commit 37ebe2a64b0720316fde96c3d072c614625fcaba.

cuda_core/cuda/core/_cpp/resource_handles.cpp
leofang: For my own understanding: this seems to be consistent with the rest of the file? In the shared pointer deleter we:
1. unregister a handle from the registry 
2. get GIL guard
3. call C API

Are you saying we should swap 1 and 2 for the whole file, including code that is not touched by this PR?

Andy-Jost: Yeah, you're right. I was thinking the GIL releaser should appear first to make it more visible, but that's a weak argument. The other pattern (matching what you have) is already used and defers GIL acquisition longer. Plus, this isn't important enough to sweat over either way.

cuda_core/cuda/core/_cpp/resource_handles.cpp
Andy-Jost: Your example shows a stream round tripping through user code. Here, we have two options: (1) don't support this (users can just use `s`, which they already have) or (2) make raw streams promotable to handles. If we do want to support the latter, then adding a registry is the right approach.

Making this a rule is indeed better for consistency, and in that case we ought to follow up with an update to `resource_handles.cpp` to templatize the repetitive details.

> I think we do want to address the corruption issue.

The root of this conflict is that associating metadata with a handle and tying the handle's lifetime to that of a Python owner are orthogonal. The metadata comes into play strictly once, when an owning handle is created. A Python owner association can be added at any time, any number of times, and with any number of distinct Python objects. It suggests using separate mechanisms.

For `cuda.core` objects we own, the handle lifetime is tied to the Python owner through a handle member: e.g., `class Stream` contains a member of type `StreamHandle.` For arbitrary Python owners we need to embed the association in  the handle graph.

Thinking this through, I think a `*_with_owner` handle-creation function taking a raw CUDA resource (`raw`) and Python owner (`x`) should be implemented as follows:

1. Use the first level registry to promote `raw` to the cannonical `shared_ptr` handle (`h`), if it exists.
2. Use the second level registry to promote `h` to the cannonical owning `PyObject` (`y`), if it exists.
3. If `x` is `y`, return `h`. The owner is a `cuda.core` class that holds this handle.
4. Else, make a new aliasing `shared_ptr` tying the lifetime of 'h' to that of `x`. For `CUstream` it might look like this:

```c++
Py_INCREF(x);
auto h2 = std::shared_ptr<void>(
      nullptr
    , [x, h](void *) noexcept {
          GILAcquireGuard gil;
          if (gil.acquired()) {
              Py_DECREF(x);
          }
      }
);
return StreamHandle(h2, h.get());  // Tie h2's control block to the original StreamBox
```

A new `shared_ptr` is needed because it is managing a call to Py_DECREF rather than a CUDA resource. The original handle, `h`, is lambda-captured, keeping it alive. `h2` is an alias for the original stream and `StreamBox`; the metadata remains intact and internal getters for metadata (like `get_box`) work correctly because they operate relative to the resource address. The new handle should not be cached at all. Every call to `*_with_owner` creates a new association (unless a cache hit in step 2 produces a `cuda.core` object).

This seems like the principled long-term fix, which should also help address #1989. As this would be a fairly complex change, it's fine with me to merge this PR as-is and address the caching and owership issues in a follow-up.

Andy-Jost (approved):

cuda_core/cuda/core/_cpp/resource_handles.cpp
leofang: @Andy-Jost PTAL commit [9d21b6b](https://github.com/NVIDIA/cuda-python/pull/1976/commits/9d21b6b6fe288d20e149f6b845eacab665f36b7b)

Andy-Jost (approved):

leofang:
/ok to test [9d21b6b](https://github.com/NVIDIA/cuda-python/pull/1976/commits/9d21b6b6fe288d20e149f6b845eacab665f36b7b)

leofang:
/ok to test [4d82262](https://github.com/NVIDIA/cuda-python/pull/1976/commits/4d82262177fda7045907d7b5b281b1c05d08a853)

leofang:
/ok to test [d840bdc](https://github.com/NVIDIA/cuda-python/pull/1976/commits/d840bdc1dde5ea8b9e1fb72c4d1f23450c2b6e82)

leofang:
I am very impatient. My CI run was finally green after several retries, but someone cut my line and caused merge conflicts. I am going to admin-merge this PR, and leave any test failures to the next person to deal with.

PR #1979, opened 2026-04-26
[FEA]: Add per-library path override env var to cuda.pathfinder

### LINKED ISSUE #1054: [FEA]: Ability to provide an extra path for cuda-pathfinder to search with high priority
### Is this a duplicate?

### Area

cuda.pathfinder

The cuda-pathfinder's search path prioritizes libraries from wheels. For example, `libnvshmem_host.so` comes from `libnvidia-nvshmem-cuXX` wheel. If I'm working from a development branch of nvshmem, I may need to manually copy libraries over the `site-packages` directory in order to get my custom libraries to be opened. `LD_LIBRARY_PATH` is much lower in the priority order (on purpose), so that won't work either. 

A few different ideas come to mind:

- `<LIBNAME_UPPER_CASE>_PATH_OVERRIDE` so each library has an ability to be overridden independently
- `CUDA_PATHFINDER_TOP_PRIORITY_PATH` which applies to all pathfinder searches
- Adding a `<library>_HOME` environment variable, and prioritizing it above other search paths

 

I think I listed all 3 above. The current hacky WAR is to copy .so files around on the system.

### Additional context

_No response_

rparolin:
## Summary

Closes #1054.

Adds `CUDA_PATHFINDER_<LIBNAME_UPPER>_PATH_OVERRIDE` as a developer escape hatch so users working on a development branch of a library (e.g. nvshmem) can point `cuda.pathfinder` at a custom build without having to uninstall the wheel or copy `.so` files into `site-packages`.

This implements option 1 from the issue (per-library override), which matches @kkraus14's later suggestion in the thread for a per-library, pathfinder-specific knob — and avoids the cross-cutting precedence concerns @leofang raised against a global `TOP_PRIORITY_PATH`.

## Behavior

- **Env var name**: `CUDA_PATHFINDER_<LIBNAME_UPPER>_PATH_OVERRIDE`
  - `nvshmem_host` → `CUDA_PATHFINDER_NVSHMEM_HOST_PATH_OVERRIDE`
  - `cublasLt` → `CUDA_PATHFINDER_CUBLASLT_PATH_OVERRIDE`
- **Value**: either an absolute file path (used as-is) or a directory (searched with the same platform logic as conda / `CUDA_PATH`).
- **Priority**: highest. Runs before site-packages, conda, system search, `CUDA_PATH`, and the canary probe. Applies uniformly to CTK, third-party, and driver libraries.
- **Fail-loud**: if the env var is set but the library cannot be resolved from the path, the load raises `DynamicLibNotFoundError` immediately rather than silently falling through. Silent fallthrough would be the actual debugging footgun — a typo or stale path should be loud.
- **`found_via`**: tagged as `override(CUDA_PATHFINDER_<LIB>_PATH_OVERRIDE)` so the resolved source is distinguishable from `site-packages`, `conda`, etc.

## Files

- `cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_steps.py` — new `find_via_path_override` find step + `path_override_env_var` helper; prepended to `EARLY_FIND_STEPS`.
- `cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py` — driver fast-path also honors the override; updated docstring.
- `cuda_pathfinder/tests/test_search_steps.py` — 11 new unit tests.

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py
+++ b/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py
@@ -24,6 +24,7 @@ from cuda.pathfinder._dynamic_libs.search_steps import (
     SearchContext,
     derive_ctk_root,
     find_via_ctk_root,
+    find_via_path_override,
     run_find_steps,
 )
 from cuda.pathfinder._dynamic_libs.subprocess_protocol import (
@@ -60,8 +61,13 @@ def _load_driver_lib_no_cache(desc: LibDescriptor) -> LoadedDL:
     Driver libs (libcuda, libnvidia-ml) are part of the display driver, not
     the CUDA Toolkit. They are expected to be discoverable via the platform's
     native loader mechanisms, so the full CTK search cascade (site-packages,
-    conda, CUDA_PATH, canary) is unnecessary.
+    conda, CUDA_PATH, canary) is unnecessary. The per-library override env
+    var is still honored so developers can point at a custom build.
     """
+    override_ctx = SearchContext(desc)
+    override_find = find_via_path_override(override_ctx)
+    if override_find is not None:
+        return LOADER.load_with_abs_path(desc, override_find.abs_path, override_find.found_via)
     loaded = LOADER.check_if_already_loaded_from_elsewhere(desc, False)
     if loaded is not None:
         return loaded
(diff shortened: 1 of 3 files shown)


cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py
```diff
@@ -221,23 +227,37 @@ def load_nvidia_dynamic_lib(libname: str) -> LoadedDL:
         RuntimeError: If Python is not 64-bit.
 
     Search order:
-        0. **Already loaded in the current process**
```
rwgk, line 224: I'm almost certain we have to keep this first.

```
0. **Already loaded in the current process**
```

Otherwise we could have undefined behavior. It's a while ago that I looked into the details of this. If there are strong reasons, I could try to dig into this question again, but I hope it's acceptable to keep it first.

rparolin:
Closing b/c we are going to merge this functionality into another set of work.

PR #1983, opened 2026-04-28
Add CUDA process checkpointing helpers

### LINKED ISSUE #1343: Support CUDA Checkpointing
- https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__CHECKPOINT.html
- https://developer.nvidia.com/blog/checkpointing-cuda-applications-with-criu/

kkraus14:
## Summary
- add a dedicated `cuda.core.checkpoint` module for CUDA process checkpointing APIs while keeping `cuda.core.system` focused on CUDA system/NVML capabilities
- expose a narrow runtime API via `checkpoint.Process(pid)`: read-only `pid`, `state`, `restore_thread_id`, `lock`, `checkpoint`, `restore`, and `unlock`
- keep the checkpoint module public surface limited to `Process`; the state return type lives in `cuda.core.typing.ProcessStateT` and is rendered in the private API docs
- map `Process.state` from the CUDA driver `CUprocessState` enumerators rather than raw integer values
- support restore-time GPU UUID remapping using either driver `CUuuid` values or `Device.uuid` strings; migration docs and tests now describe the stricter kernel-mode-driver visibility requirement rather than user-space CUDA visibility
- document the coordinator/target-process model, Linux permission requirements such as `CAP_SYS_PTRACE`, the CRIU/CPU-process-image boundary, restore-thread requirement, and persistence mode/`cuInit` restore requirement
- validate checkpoint API availability lazily and cache the successful check, covering the `cuda-bindings` version, required binding symbols, and CUDA driver version
- consolidate checkpoint driver call handling in one boundary that translates missing checkpoint symbols and unsupported checkpoint CUDA results into a checkpoint-specific `RuntimeError`
- re-enable checkpoint driver coverage in CI by running driver-backed lifecycle tests in isolated coordinator/target subprocess scenarios with parent-side timeouts; migration tests skip when the CUDA device view is masked because the mapping cannot be proven KMD-complete
- run checkpoint scenario and target subprocesses through `python -I cuda_core/tests/test_checkpoint.py ...` as a file-backed child entrypoint, avoiding manual cwd/PYTHONPATH rewriting and stringified child Python while preserving installed-wheel imports in CI
- treat accepted-but-no-op GPU migration as a hardware/driver skip for both swap and rotation scenarios while still failing if the driver moves a context to an unexpected GPU

Closes #1343

## Testing
- `git commit -S` pre-commit hooks: ruff, formatting, SPDX, whitespace, RST, and related checks passed
- `git diff --check`
- `pixi run ruff check cuda_core/tests/test_checkpoint.py` (`All checks passed`)
- `pixi run --manifest-path cuda_core pytest cuda_core/tests/test_checkpoint.py cuda_core/tests/test_typing_imports.py` (`10 passed`, `6 skipped`) after the file-backed subprocess entrypoint simplification
- `pixi run ruff check cuda_core/cuda/core/checkpoint.py cuda_core/tests/test_checkpoint.py` (`All checks passed`)
- `pixi run --manifest-path cuda_core -e docs docs-build-latest` (Sphinx build succeeded)
- previous broader local run: `pixi run --manifest-path cuda_core pytest cuda_core/tests --ignore=cuda_core/tests/cython` (`2798 passed`, `352 skipped`, `2 failed`)

The two Python-suite failures in the broader run are existing local NVML/system environment failures and are not related to this checkpointing change:
- `cuda_core/tests/system/test_system_device.py::test_get_inforom_version` returns an empty InfoROM board part number locally.
- `cuda_core/tests/system/test_system_system.py::test_get_process_name` hits an NVML UTF-8 decode error locally.

Additional local build/test notes:
- `pixi run --manifest-path cuda_core test` stops before pytest in the existing `build-cython-tests` pre-step because `cuda_core/tests/cython/test_get_cuda_native_handle.pyx` cannot find the expected `cuda.bindings` `.pxd` files in this local pixi environment.
- `pixi build` from `cuda_core` reaches the existing native cuda-core extension build and then fails with CUDA 12.9 headers that do not declare `CU_MEM_ALLOCATION_TYPE_MANAGED`; this is in the existing graph/managed-memory extension build path and is not checkpoint-specific.

CI note:
- The previous CI attempt on `8192df67` exposed two unrelated/runner issues: one Windows py3.12 build failed inside the shared mini-CTK cache setup before any cuda.core build step, and CUDA 13.x GPU test jobs were canceled after the old in-process checkpoint test hung in `cuCheckpointProcessCheckpoint`.
- The current head removes the broad `CI` skip. Driver-backed checkpoint lifecycle tests now run through isolated subprocess coordinator/target scenarios, and the parent pytest process can kill and skip a scenario that times out instead of letting the CI job hang.
- The CI attempt on `7a2e6830` passed the checkpoint lifecycle tests and only failed the two multi-GPU jobs. Both failed in `test_rotation_migrates_context` because the driver accepted the rotation mapping but reported the target context still on the original UUID; the swap migration test already observed and skipped the same no-op behavior. The current head extends that no-op skip handling to rotation while preserving the assertion for any unexpected migrated UUID.

## Current Test Implementation
The checkpoint tests in `cuda_core/tests/test_checkpoint.py` are real driver/GPU tests, not broad mocks.

Input validation and public-symbol checks run everywhere. Driver-backed lifecycle tests create a target process that initializes a real CUDA context, then a coordinator scenario calls `checkpoint.Process(target.pid)` and exercises `state`, `restore_thread_id`, `lock`, `checkpoint`, `restore`, and `unlock` through the real driver. The parent pytest process enforces a timeout around each scenario so unsupported driver/hardware paths skip cleanly instead of hanging the test job.

The scenario subprocess is launched as `python -I cuda_core/tests/test_checkpoint.py scenario <name>`, and its target process is launched as `python -I cuda_core/tests/test_checkpoint.py target <device_index>`. Python isolated mode ignores `PYTHON*` environment variables and omits the script/current directory from import resolution, so wheel CI imports the installed package without manual `cwd` or `PYTHONPATH` surgery. The child processes still inherit non-Python CUDA environment such as driver/library paths and device visibility.

Migration tests require at least two same-chip GPUs and an unmasked CUDA device view. They build full UUID mappings using `Device.uuid` strings, then exercise rotation and pair-swap migration patterns through `Process.restore(gpu_mapping=...)` in the isolated target process. They skip gracefully when `CUDA_VISIBLE_DEVICES` is set, when the local hardware lacks a same-chip GPU pair, when the driver rejects checkpoint migration, or when the driver accepts the mapping but leaves the target context on the original GPU.

### DIFF
--- a/cuda_core/cuda/core/checkpoint.py
+++ b/cuda_core/cuda/core/checkpoint.py
@@ -207,8 +207,8 @@ def _make_restore_args(driver, gpu_mapping: _Mapping[_Any, _Any] | None):
     pairs = []
     for old_uuid, new_uuid in gpu_mapping.items():
         pair = driver.CUcheckpointGpuPair()
-        pair.oldUuid = old_uuid
-        pair.newUuid = new_uuid
+        pair.oldUuid = _as_cuuuid(driver, old_uuid)
+        pair.newUuid = _as_cuuuid(driver, new_uuid)
         pairs.append(pair)
 
     if not pairs:
@@ -218,6 +218,27 @@ def _make_restore_args(driver, gpu_mapping: _Mapping[_Any, _Any] | None):
     args.gpuPairs = pairs
     args.gpuPairsCount = len(pairs)
     return args
+
+
+def _as_cuuuid(driver, value):
+    """Convert *value* to a ``CUuuid``.
+
+    Accepts a ``CUuuid`` instance (returned as-is) or a UUID string in
+    the ``"xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"`` format returned by
+    :attr:`Device.uuid`.
+    """
+    if isinstance(value, str):
+        import ctypes
+
+        raw = bytes.fromhex(value.replace("-", ""))
+        if len(raw) != 16:
+            raise ValueError(
+                "GPU UUID string must be 32 hex characters "
+                f"(with optional hyphens), got {value!r}"
+            )
+        buf = ctypes.create_string_buffer(raw, 16)
+        return driver.CUuuid(ctypes.addressof(buf))
+    return value
 
 
 __all__ = [
--- a/cuda_core/cuda/core/checkpoint.py
+++ b/cuda_core/cuda/core/checkpoint.py
@@ -50,10 +50,17 @@ class Process:
         Process ID of the CUDA process.
     """
 
-    __slots__ = ("pid",)
+    __slots__ = ("_pid",)
 
     def __init__(self, pid: int):
-        self.pid = _check_pid(pid)
+        self._pid = _check_pid(pid)
+
+    @property
+    def pid(self) -> int:
+        """
+        Process ID of the CUDA process.
+        """
+        return self._pid
 
     @property
     def state(self) -> _ProcessStateT:
@@ -61,7 +68,7 @@ class Process:
         CUDA checkpoint state for this process.
         """
         driver = _get_driver()
-        state = _call_driver(driver, driver.cuCheckpointProcessGetState, self.pid)
+        state = _call_driver(driver, driver.cuCheckpointProcessGetState, self._pid)
         state_names = _get_process_state_names(driver)
         try:
             return state_names[state]
@@ -75,7 +82,7 @@ class Process:
         CUDA restore thread ID for this process.
         """
         driver = _get_driver()
-        return _call_driver(driver, driver.cuCheckpointProcessGetRestoreThreadId, self.pid)
+        return _call_driver(driver, driver.cuCheckpointProcessGetRestoreThreadId, self._pid)
 
     def lock(self, timeout_ms: int = 0) -> None:
         """
@@ -89,14 +96,14 @@ class Process:
         driver = _get_driver()
         args = driver.CUcheckpointLockArgs()
         args.timeoutMs = _check_timeout_ms(timeout_ms)
-        _call_driver(driver, driver.cuCheckpointProcessLock, self.pid, args)
+        _call_driver(driver, driver.cuCheckpointProcessLock, self._pid, args)
 
     def checkpoint(self) -> None:
         """
         Checkpoint the GPU memory contents of this locked process.
         """
         driver = _get_driver()
-        _call_driver(driver, driver.cuCheckpointProcessCheckpoint, self.pid, None)
+        _call_driver(driver, driver.cuCheckpointProcessCheckpoint, self._pid, None)
 
     def restore(self, gpu_mapping: _Mapping[_Any, _Any] | None = None) -> None:
         """
@@ -107,18 +114,20 @@ class Process:
         gpu_mapping : mapping, optional
             GPU UUID remapping from each checkpointed GPU UUID to the GPU UUID
             to restore onto. For migration workflows, provide mappings for
-            every CUDA-visible GPU.
+            every GPU visible to the kernel-mode driver. User-space masking
+            such as ``CUDA_VISIBLE_DEVICES`` does not reduce this mapping
+            requirement.
         """
         driver = _get_driver()
         args = _make_restore_args(driver, gpu_mapping)
-        _call_driver(driver, driver.cuCheckpointProcessRestore, self.pid, args)
+        _call_driver(driver, driver.cuCheckpointProcessRestore, self._pid, args)
 
     def unlock(self) -> None:
         """
         Unlock this locked process so it can resume CUDA API calls.
         """
         driver = _get_driver()
-        _call_driver(driver, driver.cuCheckpointProcessUnlock, self.pid, None)
+        _call_driver(driver, driver.cuCheckpointProcessUnlock, self._pid, None)
 
 
 def _get_driver():
@@ -169,10 +178,7 @@ def _call_driver(driver, func, *args):
                 "Upgrade to a driver version with CUDA checkpoint API support."
             ) from e
         raise
-    return _handle_return(driver, result)
 
-
-def _handle_return(driver, result):
     err = result[0]
     not_supported_errors = (
         getattr(driver.CUresult, "CUDA_ERROR_NOT_FOUND", None),
--- a/cuda_core/cuda/core/checkpoint.py
+++ b/cuda_core/cuda/core/checkpoint.py
@@ -2,13 +2,14 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+import ctypes as _ctypes
 from collections.abc import Mapping as _Mapping
 from typing import Any as _Any
-from typing import Literal as _Literal
 
 from cuda.core._utils.cuda_utils import handle_return as _handle_cuda_return
 from cuda.core._utils.version import binding_version as _binding_version
 from cuda.core._utils.version import driver_version as _driver_version
+from cuda.core.typing import ProcessStateT as _ProcessStateT
 
 try:
     from cuda.bindings import driver as _driver
@@ -16,14 +17,12 @@ except ImportError:
     from cuda import cuda as _driver
 
 
-ProcessStateT = _Literal["running", "locked", "checkpointed", "failed"]
-
-_PROCESS_STATE_NAMES: dict[int, ProcessStateT] = {
-    0: "running",
-    1: "locked",
-    2: "checkpointed",
-    3: "failed",
-}
+_PROCESS_STATE_NAME_ATTRS: tuple[tuple[str, _ProcessStateT], ...] = (
+    ("CU_PROCESS_STATE_RUNNING", "running"),
+    ("CU_PROCESS_STATE_LOCKED", "locked"),
+    ("CU_PROCESS_STATE_CHECKPOINTED", "checkpointed"),
+    ("CU_PROCESS_STATE_FAILED", "failed"),
+)
 
 _REQUIRED_BINDING_ATTRS = (
     "cuCheckpointProcessCheckpoint",
@@ -34,6 +33,7 @@ _REQUIRED_BINDING_ATTRS = (
     "cuCheckpointProcessUnlock",
     "CUcheckpointGpuPair",
     "CUcheckpointLockArgs",
+    "CUprocessState",
     "CUcheckpointRestoreArgs",
 )
 _REQUIRED_DRIVER_VERSION = (12, 8, 0)
@@ -56,16 +56,17 @@ class Process:
         self.pid = _check_pid(pid)
 
     @property
-    def state(self) -> ProcessStateT:
+    def state(self) -> _ProcessStateT:
         """
         CUDA checkpoint state for this process.
         """
         driver = _get_driver()
         state = _call_driver(driver, driver.cuCheckpointProcessGetState, self.pid)
-        state_value = int(state)
+        state_names = _get_process_state_names(driver)
         try:
-            return _PROCESS_STATE_NAMES[state_value]
+            return state_names[state]
         except KeyError as e:
+            state_value = int(state)
             raise RuntimeError(f"Unknown CUDA checkpoint process state: {state_value}") from e
 
     @property
@@ -105,8 +106,8 @@ class Process:
         ----------
         gpu_mapping : mapping, optional
             GPU UUID remapping from each checkpointed GPU UUID to the GPU UUID
-            to restore onto. If provided, the mapping must contain every
-            checkpointed GPU UUID.
+            to restore onto. For migration workflows, provide mappings for
+            every CUDA-visible GPU.
         """
         driver = _get_driver()
         args = _make_restore_args(driver, gpu_mapping)
@@ -152,6 +153,10 @@ def _get_driver():
 def _binding_version_supports_checkpoint(version) -> bool:
     major, minor, patch = version[:3]
     return (major == 12 and (minor, patch) >= (8, 0)) or (major == 13 and (minor, patch) >= (0, 2)) or major > 13
+
+
+def _get_process_state_names(driver) -> dict[_Any, _ProcessStateT]:
+    return {getattr(driver.CUprocessState, attr): state_name for attr, state_name in _PROCESS_STATE_NAME_ATTRS}
 
 
 def _call_driver(driver, func, *args):
@@ -207,8 +212,9 @@ def _make_restore_args(driver, gpu_mapping: _Mapping[_Any, _Any] | None):
     pairs = []
     for old_uuid, new_uuid in gpu_mapping.items():
         pair = driver.CUcheckpointGpuPair()
-        pair.oldUuid = _as_cuuuid(driver, old_uuid)
-        pair.newUuid = _as_cuuuid(driver, new_uuid)
+        buffers = []
+        pair.oldUuid = _as_cuuuid(driver, old_uuid, buffers)
+        pair.newUuid = _as_cuuuid(driver, new_uuid, buffers)
         pairs.append(pair)
 
     if not pairs:
@@ -220,7 +226,7 @@ def _make_restore_args(driver, gpu_mapping: _Mapping[_Any, _Any] | None):
     return args
 
 
-def _as_cuuuid(driver, value):
+def _as_cuuuid(driver, value, buffers):
     """Convert *value* to a ``CUuuid``.
 
     Accepts a ``CUuuid`` instance (returned as-is) or a UUID string in
@@ -228,13 +234,12 @@ def _as_cuuuid(driver, value):
     :attr:`Device.uuid`.
     """
     if isinstance(value, str):
-        import ctypes
-
         raw = bytes.fromhex(value.replace("-", ""))
         if len(raw) != 16:
             raise ValueError(f"GPU UUID string must be 32 hex characters (with optional hyphens), got {value!r}")
-        buf = ctypes.create_string_buffer(raw, 16)
-        return driver.CUuuid(ctypes.addressof(buf))
+        buf = _ctypes.create_string_buffer(raw, 16)
+        buffers.append(buf)
+        return driver.CUuuid(_ctypes.addressof(buf))
     return value
 
 
--- a/cuda_core/cuda/core/checkpoint.py
+++ b/cuda_core/cuda/core/checkpoint.py
@@ -1,0 +1,182 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from collections.abc import Mapping as _Mapping
+from dataclasses import dataclass as _dataclass
+from enum import IntEnum as _IntEnum
+from typing import Any as _Any
+
+from cuda.core._utils.cuda_utils import handle_return as _handle_cuda_return
+
+try:
+    from cuda.bindings import driver as _driver
+except ImportError:
+    from cuda import cuda as _driver
+
+
+class ProcessState(_IntEnum):
+    """
+    CUDA checkpoint state for a process.
+    """
+
+    RUNNING = 0
+    LOCKED = 1
+    CHECKPOINTED = 2
+    FAILED = 3
+
+
+@_dataclass(frozen=True)
+class Process:
+    """
+    CUDA process that can be locked, checkpointed, restored, and unlocked.
+
+    Parameters
+    ----------
+    pid : int
+        Process ID of the CUDA process.
+    """
+
+    pid: int
+
+    def __post_init__(self):
+        _check_pid(self.pid)
+
+    @property
+    def state(self) -> ProcessState:
+        """
+        CUDA checkpoint state for this process.
+        """
+        driver = _get_driver()
+        state = _handle_return(driver, driver.cuCheckpointProcessGetState(self.pid))
+        return ProcessState(int(state))
+
+    @property
+    def restore_thread_id(self) -> int:
+        """
+        CUDA restore thread ID for this process.
+        """
+        driver = _get_driver()
+        return _handle_return(driver, driver.cuCheckpointProcessGetRestoreThreadId(self.pid))
+
+    def lock(self, timeout_ms: int = 0) -> None:
+        """
+        Lock this process, blocking further CUDA API calls.
+
+        Parameters
+        ----------
+        timeout_ms : int, optional
+            Timeout in milliseconds. A value of 0 indicates no timeout.
+        """
+        driver = _get_driver()
+        args = driver.CUcheckpointLockArgs()
+        args.timeoutMs = _check_timeout_ms(timeout_ms)
+        _handle_return(driver, driver.cuCheckpointProcessLock(self.pid, args))
+
+    def checkpoint(self) -> None:
+        """
+        Checkpoint the GPU memory contents of this locked process.
+        """
+        driver = _get_driver()
+        _handle_return(driver, driver.cuCheckpointProcessCheckpoint(self.pid, None))
+
+    def restore(self, gpu_mapping: _Mapping[_Any, _Any] | None = None) -> None:
+        """
+        Restore this checkpointed process.
+
+        Parameters
+        ----------
+        gpu_mapping : mapping, optional
+            GPU UUID remapping from each checkpointed GPU UUID to the GPU UUID
+            to restore onto. If provided, the mapping must contain every
+            checkpointed GPU UUID.
+        """
+        driver = _get_driver()
+        args = _make_restore_args(driver, gpu_mapping)
+        _handle_return(driver, driver.cuCheckpointProcessRestore(self.pid, args))
+
+    def unlock(self) -> None:
+        """
+        Unlock this locked process so it can resume CUDA API calls.
+        """
+        driver = _get_driver()
+        _handle_return(driver, driver.cuCheckpointProcessUnlock(self.pid, None))
+
+
+def _get_driver():
+    required = (
+        "cuCheckpointProcessCheckpoint",
+        "cuCheckpointProcessGetRestoreThreadId",
+        "cuCheckpointProcessGetState",
+        "cuCheckpointProcessLock",
+        "cuCheckpointProcessRestore",
+        "cuCheckpointProcessUnlock",
+        "CUcheckpointGpuPair",
+        "CUcheckpointLockArgs",
+        "CUcheckpointRestoreArgs",
+    )
+    missing = [name for name in required if not hasattr(_driver, name)]
+    if missing:
+        raise RuntimeError(
+            f"CUDA checkpointing requires cuda.bindings with CUDA checkpoint API support. Missing: {', '.join(missing)}"
+        )
+    return _driver
+
+
+def _handle_return(driver, result):
+    err = result[0]
+    not_supported_errors = (
+        getattr(driver.CUresult, "CUDA_ERROR_NOT_FOUND", None),
+        getattr(driver.CUresult, "CUDA_ERROR_NOT_SUPPORTED", None),
+    )
+    if err in not_supported_errors:
+        raise RuntimeError(
+            "CUDA checkpointing is not supported by the installed NVIDIA driver. "
+            "Upgrade to a driver version with CUDA checkpoint API support."
+        )
+
+    return _handle_cuda_return(result)
+
+
+def _check_pid(pid: int) -> int:
+    if isinstance(pid, bool) or not isinstance(pid, int):
+        raise TypeError("pid must be an int")
+    if pid <= 0:
+        raise ValueError("pid must be a positive int")
+    return pid
+
+
+def _check_timeout_ms(timeout_ms: int) -> int:
+    if isinstance(timeout_ms, bool) or not isinstance(timeout_ms, int):
+        raise TypeError("timeout_ms must be an int")
+    if timeout_ms < 0:
+        raise ValueError("timeout_ms must be >= 0")
+    return timeout_ms
+
+
+def _make_restore_args(driver, gpu_mapping: _Mapping[_Any, _Any] | None):
+    if gpu_mapping is None:
+        return None
+    if not isinstance(gpu_mapping, _Mapping):
+        raise TypeError("gpu_mapping must be a mapping from checkpointed GPU UUID to restore GPU UUID")
+
+    pairs = []
+    for old_uuid, new_uuid in gpu_mapping.items():
+        pair = driver.CUcheckpointGpuPair()
+        pair.oldUuid = old_uuid
+        pair.newUuid = new_uuid
+        pairs.append(pair)
+
+    if not pairs:
+        return None
+
+    args = driver.CUcheckpointRestoreArgs()
+    args.gpuPairs = pairs
+    args.gpuPairsCount = len(pairs)
+    return args
+
+
+__all__ = [
+    "Process",
+    "ProcessState",
+]
--- a/cuda_core/cuda/core/checkpoint.py
+++ b/cuda_core/cuda/core/checkpoint.py
@@ -3,11 +3,12 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from collections.abc import Mapping as _Mapping
-from dataclasses import dataclass as _dataclass
-from enum import IntEnum as _IntEnum
 from typing import Any as _Any
+from typing import Literal as _Literal
 
 from cuda.core._utils.cuda_utils import handle_return as _handle_cuda_return
+from cuda.core._utils.version import binding_version as _binding_version
+from cuda.core._utils.version import driver_version as _driver_version
 
 try:
     from cuda.bindings import driver as _driver
@@ -15,18 +16,30 @@ except ImportError:
     from cuda import cuda as _driver
 
 
-class ProcessState(_IntEnum):
-    """
-    CUDA checkpoint state for a process.
-    """
+ProcessStateT = _Literal["running", "locked", "checkpointed", "failed"]
 
-    RUNNING = 0
-    LOCKED = 1
-    CHECKPOINTED = 2
-    FAILED = 3
+_PROCESS_STATE_NAMES: dict[int, ProcessStateT] = {
+    0: "running",
+    1: "locked",
+    2: "checkpointed",
+    3: "failed",
+}
+
+_REQUIRED_BINDING_ATTRS = (
+    "cuCheckpointProcessCheckpoint",
+    "cuCheckpointProcessGetRestoreThreadId",
+    "cuCheckpointProcessGetState",
+    "cuCheckpointProcessLock",
+    "cuCheckpointProcessRestore",
+    "cuCheckpointProcessUnlock",
+    "CUcheckpointGpuPair",
+    "CUcheckpointLockArgs",
+    "CUcheckpointRestoreArgs",
+)
+_REQUIRED_DRIVER_VERSION = (12, 8, 0)
+_driver_capability_checked = False
 
 
-@_dataclass(frozen=True)
 class Process:
     """
     CUDA process that can be locked, checkpointed, restored, and unlocked.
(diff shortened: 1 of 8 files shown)


kkraus14:
/ok to test https://github.com/NVIDIA/cuda-python/pull/1983/commits/7c66b2fbce82ed583a1d124049ef07a051a35ba2

cuda_core/cuda/core/system/__init__.py
```diff
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
kkraus14: Shouldn't update this since we're not updating the file

cuda_core/cuda/core/checkpoint.py
```diff
@@ -0,0 +1,192 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
```
kkraus14: Why is this needed?

cuda_core/cuda/core/checkpoint.py
```diff
@@ -0,0 +1,192 @@
...
+    Attributes
+    ----------
+    old_uuid
+        UUID of the GPU that was checkpointed.
+    new_uuid
+        UUID of the GPU to restore onto.
+    """
+
+    old_uuid: Any
+    new_uuid: Any
```
kkraus14: This class feels a bit awkward. Would it feel cleaner to just pass key-value pairs of GPUs as a dict or something similar?

cuda_core/cuda/core/checkpoint.py
```diff
@@ -0,0 +1,192 @@
...
+        "cuCheckpointProcessUnlock",
+        "CUcheckpointGpuPair",
+        "CUcheckpointLockArgs",
+        "CUcheckpointRestoreArgs",
+    )
+    missing = [name for name in required if not hasattr(driver, name)]
+    if missing:
+        raise RuntimeError(
+            f"CUDA checkpointing requires cuda.bindings with CUDA checkpoint API support. Missing: {', '.join(missing)}"
+        )
```
kkraus14: In addition to handling the case when the CUDA checkpoint APIs don't exist in the `cuda.bindings` package, I think we need to handle the case where someone is using a newer `cuda.bindings` package with an older driver version. I believe the Python APIs would exist but throw at runtime if tried to be used?

cuda_core/cuda/core/checkpoint.py
```diff
@@ -0,0 +1,192 @@
...
+    return driver
+
+
+def _handle_return(result):
+    from cuda.core._utils.cuda_utils import handle_return
```
kkraus14: Can we move this to a top level import instead of being function local?

cuda_core/cuda/core/checkpoint.py
kkraus14: Removed it; the annotations work without `from __future__ import annotations` on the supported Python versions.

cuda_core/cuda/core/system/__init__.py
kkraus14: Removed this from the final PR diff; `cuda_core/cuda/core/system/__init__.py` now matches the base tree.

cuda_core/cuda/core/checkpoint.py
kkraus14: Moved `handle_return` to a module-level import.

cuda_core/cuda/core/checkpoint.py
kkraus14: Agreed. Replaced the `GpuPair` dataclass with a `gpu_mapping` argument that accepts a mapping of checkpointed GPU UUID to restore GPU UUID, and removed `GpuPair` from the public docs/API.

cuda_core/cuda/core/checkpoint.py
kkraus14: Added a checkpoint-specific return handler for `CUDA_ERROR_NOT_FOUND` and `CUDA_ERROR_NOT_SUPPORTED`, so newer bindings on an older driver now raise a clearer `RuntimeError` explaining that the installed driver lacks checkpoint API support.

cuda_core/cuda/core/checkpoint.py
```diff
@@ -0,0 +1,182 @@
...
+        """
+        driver = _get_driver()
+        _handle_return(driver, driver.cuCheckpointProcessUnlock(self.pid, None))
+
+
+def _get_driver():
+    try:
+        from cuda.bindings import driver
+    except ImportError:
+        from cuda import cuda as driver
```
kkraus14: We should do this as a top level import if possible

kkraus14: Moved the driver import to module scope as a private `_driver` binding while keeping the checkpoint capability check in `_get_driver()`.

cuda_core/tests/test_checkpoint.py
```diff
@@ -0,0 +1,203 @@
...
+        self.timeoutMs = None
+
+
+class _CheckpointRestoreArgs:
+    def __init__(self):
+        self.gpuPairs = None
+        self.gpuPairsCount = None
+
+
+class _Driver:
```
rparolin: Nit: Consider renaming to indicate its a mock. For example, `_MockDriver`.

cuda_core/cuda/core/checkpoint.py
```diff
@@ -0,0 +1,182 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
```
rparolin, line 4: The module is a brand-new 1.0 public surface but has no narrative for users — there's no description of the required call sequence, the cross-thread restore requirement (which is why `restore_thread_id` is exposed), or platform support. Suggest adding a module docstring so the autosummary landing page isn't bare.

```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
#
# SPDX-License-Identifier: Apache-2.0

"""CUDA process checkpointing.

This module exposes a small, Pythonic surface over the CUDA driver's
process-checkpoint APIs. The typical lifecycle for checkpointing a CUDA
process is::

    proc = checkpoint.Process(pid)
    proc.lock()         # block further CUDA API calls in the target process
    proc.checkpoint()   # snapshot device state to host memory
    # ... persist / migrate / wait ...
    proc.restore()      # restore device state (optionally remapping GPUs)
    proc.unlock()       # resume CUDA API calls

Notes
-----
- Checkpointing is supported on Linux only and requires an NVIDIA driver
  that ships the ``cuCheckpointProcess*`` entry points. On unsupported
  drivers, constructing a :class:`Process` succeeds but operations raise
  :class:`RuntimeError`.
- ``restore`` must be invoked from the thread identified by
  :attr:`Process.restore_thread_id`; calling it from another thread is
  undefined behavior per the driver contract.
- :class:`Process` is a thin handle: instantiation does not validate that
  ``pid`` refers to a live CUDA process. Errors surface on the first
  driver call.
"""
```

cuda_core/cuda/core/checkpoint.py
```diff
@@ -0,0 +1,182 @@
...
+        "CUcheckpointGpuPair",
+        "CUcheckpointLockArgs",
+        "CUcheckpointRestoreArgs",
+    )
+    missing = [name for name in required if not hasattr(_driver, name)]
+    if missing:
+        raise RuntimeError(
+            f"CUDA checkpointing requires cuda.bindings with CUDA checkpoint API support. Missing: {', '.join(missing)}"
+        )
+    return _driver
```
leofang, line 159: I think this is not safe. Elsewhere in the codebase, we have helpers that do both checks (cuda-bindings version and driver version) to ensure API availability. We should follow the same pattern and lazy-init & cache the result.

rparolin, line 159: `_get_driver()` is called from every public method/property, so the `hasattr` loop over 9 names runs on each `state` read, each `lock`/`checkpoint`/`restore`/`unlock` call. The capability check only needs to succeed once per process. Cache it with a module-level flag so steady-state checkpoint loops don't pay the introspection cost:

```suggestion
_driver_capability_checked = False


def _get_driver():
    global _driver_capability_checked
    if _driver_capability_checked:
        return _driver
    required = (
        "cuCheckpointProcessCheckpoint",
        "cuCheckpointProcessGetRestoreThreadId",
        "cuCheckpointProcessGetState",
        "cuCheckpointProcessLock",
        "cuCheckpointProcessRestore",
        "cuCheckpointProcessUnlock",
        "CUcheckpointGpuPair",
        "CUcheckpointLockArgs",
        "CUcheckpointRestoreArgs",
    )
    missing = [name for name in required if not hasattr(_driver, name)]
    if missing:
        raise RuntimeError(
            f"CUDA checkpointing requires cuda.bindings with CUDA checkpoint API support. Missing: {', '.join(missing)}"
        )
    _driver_capability_checked = True
    return _driver
```

Note this only caches the binding-level capability check (the `hasattr` introspection). Per-call `CUDA_ERROR_NOT_FOUND` / `CUDA_ERROR_NOT_SUPPORTED` translation in `_handle_return` still has to run on each call — that catches the case where the binding wrapper exists but the loaded `libcuda.so` doesn't export the symbol, which `hasattr` can't detect.

cuda_core/cuda/core/checkpoint.py
```diff
@@ -0,0 +1,182 @@
...
+    CUDA checkpoint state for a process.
+    """
+
+    RUNNING = 0
+    LOCKED = 1
+    CHECKPOINTED = 2
+    FAILED = 3
+
+
+@_dataclass(frozen=True)
```
leofang: Why does this need to be a data class?

cuda_core/cuda/core/checkpoint.py
```diff
@@ -0,0 +1,182 @@
...
+
+class ProcessState(_IntEnum):
+    """
+    CUDA checkpoint state for a process.
+    """
+
+    RUNNING = 0
+    LOCKED = 1
+    CHECKPOINTED = 2
+    FAILED = 3
```
leofang: We don't use enums in cuda.core (but we do, currently, in cuda.core.systems, which is something that @mdboom and I were discussing yesterday). How about just return a string (and we can type the set of possible strings as `ProcessStateT = Literal[...]`)?

cuda_core/cuda/core/checkpoint.py
```diff
@@ -0,0 +1,182 @@
...
+
+    @property
+    def restore_thread_id(self) -> int:
+        """
+        CUDA restore thread ID for this process.
+        """
+        driver = _get_driver()
+        return _handle_return(driver, driver.cuCheckpointProcessGetRestoreThreadId(self.pid))
+
+    def lock(self, timeout_ms: int = 0) -> None:
```
leofang, line 87: Usually we use `<Something>Options` to encapsulate options. Though here we have 2 APIs that take two distinct sets of options. Maybe @Andy-Jost has some thoughts here?

leofang (changes requested):

cuda_core/cuda/core/checkpoint.py
leofang: Can module-level docstrings be rendered in Sphinx? If not, I'd rather us adding it to the Sphinx rst file directly.

cuda_core/tests/test_checkpoint.py
kkraus14: Addressed in `4992921dfe`: renamed the fake driver classes to `_MockDriver`, `_MockDriverProcessState`, and `_MockDriverResult`.

cuda_core/cuda/core/checkpoint.py
kkraus14: Addressed in `4992921dfe`: I moved the user-facing narrative into `cuda_core/docs/source/api.rst` so it renders in Sphinx. It now covers Linux support, the `lock -> checkpoint -> restore -> unlock` lifecycle, string process states, GPU remapping, the restore-thread requirement, and the restore-to-locked state transition.

cuda_core/cuda/core/checkpoint.py
kkraus14: Addressed in `4992921dfe`: `_get_driver()` now lazily checks the `cuda-bindings` version, required checkpoint binding symbols, and CUDA driver version before returning the driver, and caches a successful availability check. The driver-call path also translates checkpoint `Function ... not found` runtime errors into the checkpoint-specific unsupported-driver error.

kkraus14: Addressed in `4992921dfe`: the required-symbol scan is now on the lazy cached availability path, so public methods/properties no longer repeat the `hasattr` loop after the first successful check. Per-call driver result handling remains on each API call.

cuda_core/cuda/core/checkpoint.py
kkraus14: Addressed in `4992921dfe`: `Process` is now an explicit lightweight class with `__slots__` and `__init__` validation instead of a dataclass.

cuda_core/cuda/core/checkpoint.py
kkraus14: Addressed in `4992921dfe`: removed the public enum. `Process.state` now returns one of the strings typed by `ProcessStateT = Literal["running", "locked", "checkpointed", "failed"]`, and the tests cover the driver-state to string mapping. `__all__` is intentionally narrowed to `["Process"]`.

cuda_core/cuda/core/checkpoint.py
kkraus14: I kept these as direct method parameters for now rather than adding public `LockOptions` / `RestoreOptions` types. Each operation currently has one user-facing option (`timeout_ms` for lock, `gpu_mapping` for restore), and adding two public option classes would expand the 1.0 API surface without buying much structure yet. If we expect these argument sets to grow soon, I can switch to `ProcessLockOptions` and `ProcessRestoreOptions`; otherwise I think the direct parameters are the smaller stable API.

Andy-Jost: I'm +1 for starting with direct parameters.

Proactively bundling options makes a lot of sense in the driver, which needs to be extendable and ABI-stable at the same time. An example is `cuCheckpointProcessCheckpoint`, which accepts an empty `CUcheckpointCheckpointArgs` structure. By forcing callers to pass an empty struct now, options can be added later without changing the call signature.

We don't face the same pressure in `cuda.core`. Here, I would not expect either option list to grow significantly, the direct parameters are simpler, and we have a clear migration path to `ProcessLockOptions` / `ProcessRestoreOptions` if we ever change our minds

cuda_core/cuda/core/checkpoint.py
```diff
@@ -0,0 +1,225 @@
...
+from typing import Literal as _Literal
+
+from cuda.core._utils.cuda_utils import handle_return as _handle_cuda_return
+from cuda.core._utils.version import binding_version as _binding_version
+from cuda.core._utils.version import driver_version as _driver_version
+
+try:
+    from cuda.bindings import driver as _driver
+except ImportError:
+    from cuda import cuda as _driver
```
Andy-Jost, line 17: nit: do we still need this fallback? `cuda.bindings.driver` is imported unconditionally elsewhere.

Andy-Jost (approved):

cuda_core/cuda/core/checkpoint.py
leofang: > We don't use enums in cuda.core (but we do, currently, in cuda.core.systems, which is something that @mdboom and I were discussing yesterday).

xref: #1995

cuda_core/cuda/core/checkpoint.py
leofang: > and we have a clear migration path to `ProcessLockOptions` / `ProcessRestoreOptions` if we ever change our minds

How does the migration look like, for my own understanding?

cuda_core/cuda/core/checkpoint.py
leofang: It is used in `_get_driver` below.

cuda_core/cuda/core/checkpoint.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+ProcessStateT = _Literal["running", "locked", "checkpointed", "failed"]
```
leofang: Expose this to `cuda.core.typing` and then add it to `api_private.rst` to make it rendered by Sphinx

cuda_core/cuda/core/checkpoint.py
```diff
@@ -0,0 +1,225 @@
...
+
+_PROCESS_STATE_NAMES: dict[int, ProcessStateT] = {
+    0: "running",
+    1: "locked",
+    2: "checkpointed",
+    3: "failed",
```
leofang: nit: use the actual enumerators as key, instead of plain Python ints

cuda_core/docs/source/api.rst
```diff
@@ -174,6 +174,47 @@ CUDA compilation toolchain
...
+
+A checkpoint workflow operates on a CUDA process by process ID. The typical
+sequence is to lock the process, capture its GPU memory state, restore it
+when needed, and then unlock it so CUDA API calls can resume:
+
+.. code-block:: python
+
+   from cuda.core import checkpoint
+
+   process = checkpoint.Process(pid)
```
leofang: Q: Should we teach users how to get `pid` in Python?

cuda_core/cuda/core/checkpoint.py
```diff
@@ -0,0 +1,225 @@
...
+        pair.oldUuid = old_uuid
+        pair.newUuid = new_uuid
+        pairs.append(pair)
+
+    if not pairs:
+        return None
+
+    args = driver.CUcheckpointRestoreArgs()
+    args.gpuPairs = pairs
+    args.gpuPairsCount = len(pairs)
```
leofang, line 231: xref: https://github.com/NVIDIA/cuda-checkpoint/blob/1c71b048c779d30921f01172dda96e04033bb8ac/src/r580-migration-api.c#L77-L88

(The CUDA docs is pretty lacking unfortunately...)

cuda_core/docs/source/api.rst
leofang: xref: https://github.com/NVIDIA/cuda-checkpoint/blob/1c71b048c779d30921f01172dda96e04033bb8ac/src/r580-migration-api.c#L90

Maybe we should call `os.getpid()` here?

cuda_core/tests/test_checkpoint.py
leofang, line 1: Sorry, but I find this test suite very problematic. Why are we mocking the entire tests? This only works when we implement the checkpoint module in pure Python. Once we lower to Cython/C++, it won't work. Plus, across cuda-core we never, ever mock the tests. We always require GPU machines to test cuda-core functionalities. This seems like an agentic laziness to avoid writing/running GPU tests from within sandbox!

leofang (changes requested):

cuda_core/tests/test_checkpoint.py
kkraus14: I had my agent mock the test suite because CUDA check pointing requires more than just interacting with the CUDA driver / other libraries. It requires using CRIU which requires a whole bunch of kernel capabilities alongside building a harness for process management. It felt like it would be fragile to try to implement reliable actual tests because of this so I opted for mocking the driver given the API surface is quite small.

Happy to give it a shot of having actual tests that checkpoint and restore a process instead of mocking things if you think that would be fruitful.

cuda_core/docs/source/api.rst
kkraus14: This would allow check pointing the non current process so I don't think using `os.getpid()` is appropriate?

cuda_core/tests/test_checkpoint.py
leofang: If we write the tests using Andy’s [min-2-GPU ~~decorator~~ fixture](https://github.com/NVIDIA/cuda-python/blob/aac5bf50859420138563947c106283fe2d92c4bc/cuda_core/tests/conftest.py#L221), we can test the [GPU migration capability](https://github.com/NVIDIA/cuda-checkpoint/blob/main/src/r580-migration-api.c) without CRIU. The idea is that we shuffle each GPU’s state to the next one (and wrap around).

cuda_core/cuda/core/checkpoint.py
```diff
@@ -0,0 +1,225 @@
...
+    def restore(self, gpu_mapping: _Mapping[_Any, _Any] | None = None) -> None:
+        """
+        Restore this checkpointed process.
+
+        Parameters
+        ----------
+        gpu_mapping : mapping, optional
+            GPU UUID remapping from each checkpointed GPU UUID to the GPU UUID
+            to restore onto. If provided, the mapping must contain every
+            checkpointed GPU UUID.
```
leofang: I think this is why a real, instead of mocking, test is a MUST. Apparently, [the API doc](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__CHECKPOINT.html#group__CUDA__CHECKPOINT_1gf2066439091dfa0eae0cbca0144f5e91) and the [example code](https://github.com/NVIDIA/cuda-checkpoint/blob/1c71b048c779d30921f01172dda96e04033bb8ac/src/r580-migration-api.c#L80) diverge here. The latter requests "all devices visible to CUDA" must appear in the mapping, not just those participating in checkpointing (as indicated by the former). We should get clarification on this (even better, find a way to test this).

cuda_core/cuda/core/checkpoint.py
Andy-Jost: We can simulate this overload by discrimination over the arguments:

```python
def lock(self, timeout_ms: int = 0):
    pass

def lock(self, options: ProcessLockOptions):
    pass
```

This approach comes with a performance cost that would be negligible for the checkpoint API.

We have several implementation options:

1. Approximate the overload by adding a kw-only argument, `options`:

```python
def lock(self, timeout_ms: int = None, *, options : ProcessLockOptions = None):
    if options is not None:
        # verify timeout_ms is None
        # call with options
        . . .
    else:
        if timeout_ms is None:
            timeout_ms = 0
        # call with timeout_ms
        . . .
```

2. Hand-roll an exact implementation (allowing calls like `proc.lock(options)`) by inspecting `*args` and `**kwargs`.
3. Properly build the overload set with a helper:

```python
@functools.singledispatchmethod
def lock(self, timeout_ms: int = 0):
    # call with timeout_ms
    . . .

@lock.register
def _(self, options: Options):
    # call with options
    . . .
...
```

cuda_core/docs/source/api.rst
Andy-Jost: It might be worth pointing out in `api.rst` that this is typically used to checkpoint a _different_ process.

leofang: > This would allow check pointing the non current process

> this is typically used to checkpoint a _different_ process.

I think `os.getpid()` allows for checkpointing self, which is useful as demo'd in the linked code. 

I do not believe all PIDs are allowed. I assume only processes owned by the current user can be checkpointed (either limited by the Linux kernel or the CUDA driver).

In any case, the example snippet in `api.rst` isn't very clear with the current 4 lines of code (lock -> checkpoint -> restore -> unlock). It is not the full story. There are lots of things that need to happen behind the scene. It was the main reason I started digging all of these myself without relying on AI.

cuda_core/cuda/core/checkpoint.py
leofang: Yeah `singledispatchmethod` works here only by luck (we take exactly 1 argument here). But I see your point. Resolving.

cuda_core/docs/source/api.rst
Andy-Jost: FWIW, my understanding is that checkpointing is something you'd typically do **to** a process, analogous to sending a signal or attaching a debugger. Linux handles the permissions, and checkpointing requires `CAP_SYS_PTRACE`, the same permissions needed to attach a debugger or run `strace` against another user's process. One might expect a system admin to run it with sudo privileges.

The main purpose of CUDA checkpoint is to ensure everything managed by CUDA resides in CPU user space so that a tool such as CRIU can capture a complete process image. Without this, CRIU would miss the GPU state.

Use cases:

1. Migrate a GPU workload to a different system.
2. Periodically checkpoint a long-running job so it can be quickly resumed after a potential system failure.
3. Preempt GPU resources to favor a job with higher priority.

These fit naturally into a system-admin role. It looks like CUDA allows a process to checkpoint itself, but it seems to me the use cases would be niche.

cuda_core/cuda/core/checkpoint.py
kkraus14: Addressed in 8192df67. The restore docs now call out the stricter migration guidance from the CUDA example: for migration workflows, provide mappings for every CUDA-visible GPU. The implementation still passes the mapping through to the driver for final validation.

kkraus14 (commented):
Addressed the current CUDA checkpointing review comments in signed commit 8192df67.

cuda_core/tests/test_checkpoint.py
kkraus14: Addressed in 8192df67. The checkpoint tests are now real driver/GPU tests rather than broad mocks. They self-checkpoint the pytest process through lock/checkpoint/restore/unlock and include migration tests that rotate/swap full CUDA-visible `Device.uuid` string mappings on suitable >=2 same-chip GPU systems, with skips for unsupported hardware/driver cases.

kkraus14:
Review follow-up notes for signed commit `8192df67`:

GitHub accepted the inline reply for the restore-mapping xref thread, but the remaining inline reply attempts are currently returning GitHub server errors, so I am summarizing the responses here.

- `ProcessStateT`: moved to `cuda.core.typing`, included in `typing.__all__`, and listed in `api_private.rst`. `cuda.core.checkpoint` imports it privately and does not expose it in `checkpoint.__all__` or as `checkpoint.ProcessStateT`.
- Process state mapping: updated to use the actual `driver.CUprocessState.CU_PROCESS_STATE_*` enumerators as keys rather than raw Python ints.
- Checkpoint docs: rewrote the section to describe the coordinator/target PID model, show `os.getpid()` only as a self-checkpoint example, call out Linux permissions including `CAP_SYS_PTRACE`, explain the CRIU/CPU-process-image boundary, and document restore-thread plus persistence mode/`cuInit` requirements.
- Tests: replaced the broad mock suite with real driver/GPU tests. The lifecycle tests self-checkpoint the pytest process through `lock`, `checkpoint`, `restore`, and `unlock`; migration tests rotate/swap full CUDA-visible `Device.uuid` string mappings on suitable same-chip multi-GPU systems and skip unsupported hardware/driver cases.
- Mapping semantics: docs and tests now use full CUDA-visible mappings for migration. Tests also skip at call time when a driver exposes checkpoint symbols but returns `CUDA_ERROR_NOT_SUPPORTED` for actual checkpoint operations.

leofang:
Copy-pasta from my bot, with internal info redacted.

## What I did
- Discovered GPU migration doesn't work on this machine (heterogeneous GPUs + known driver bugs)
- Searched internal resources (Confluence, gdrive, NVBugs, P4) to understand the root cause
- Rewrote the full test suite with real GPU tests and no mocks
- Pushed 4 commits to Keith's PR branch

## Test suite structure (13 tests)
- 5 input validation (no GPU needed): invalid pid types/values, public symbols
- 6 lifecycle (single GPU, real driver): state transitions at every step, restore_thread_id, lock/unlock with timeouts, full checkpoint/restore cycle
- 2 migration (≥2 same-chip GPUs): rotation and swap — exercise real driver API, skip gracefully when unsupported

## Lessons learned
1. nvidia-smi order ≠ CUDA device order. We wasted time testing the wrong GPU pair because CUDA_VISIBLE_DEVICES=0,3 selected Ada + A6000 (different chips) instead of two A6000s. Always use Device.uuid to identify GPUs, never assume nvidia-smi indices match CUDA indices.

2. CUDA_VISIBLE_DEVICES breaks checkpoint migration. Tested locally: even a 2-pair identity mapping fails with CUDA_ERROR_INVALID_VALUE when only 2 of 4 GPUs are visible, while the same 4-pair identity mapping succeeds with all GPUs visible. The CUDA driver docs state the mapping must cover "every checkpointed GPU," and the driver records all physically attached GPUs at checkpoint time — not just the CUDA_VISIBLE_DEVICES subset.

3. Migration requires exact chip match. Tested locally: any mapping that remaps between different architectures (e.g. Ada :left_right_arrow: A6000) is rejected with CUDA_ERROR_INVALID_VALUE. Same-chip mappings (A6000 :left_right_arrow: A6000) are accepted. The public CUDA docs say "the GPU to restore onto needs to be of the same chip type as the old GPU."

4. Migration may be a no-op on some driver versions. Tested locally: same-chip A6000 swap is accepted (no error) but the context device UUID doesn't change — the driver silently no-ops. This is consistent with NVBug 5437334 (api_reverse_gpu_pairs fails on GA100x4) and NVBug 5544504 (customer report: "restore on a different GPU — the checkpointed process does not appear in the GPU process list").

5. Checkpoint state machine is strict. Can't unlock from "checkpointed" state — must restore first (CUDA_ERROR_ILLEGAL_STATE). Can't corrupt GPU memory between checkpoint and restore within the same process (CUDA APIs are frozen). The overwrite-then-restore scenario requires an external coordinator (CRIU).

6. Use cuda.core APIs in tests, not raw bindings. Device().uuid for current device, Device.uuid for mapping keys, Device.get_all_devices() for enumeration. The implementation (checkpoint.py) handles the string-to-CUuuid conversion internally.

leofang:
Pushed [e9c03de](https://github.com/NVIDIA/cuda-python/pull/1983/commits/e9c03de152d465b8bfcfb40c68886d5dc379670a) because the real tests hang in the CI...

leofang:
/ok to test [e9c03de](https://github.com/NVIDIA/cuda-python/pull/1983/commits/e9c03de152d465b8bfcfb40c68886d5dc379670a)

cuda_core/cuda/core/checkpoint.py
leofang: I see this has been modified. Just leave a quick note: as per https://github.com/NVIDIA/cuda-python/pull/1983#issuecomment-4372769138 the visibility here is w.r.t. the KMD, not the UMD. In particular, apps that use `CUDA_VISIBLE_DEVICES` to mask out GPUs in the user space (e.g. Dask & co) cannot be checkpointed. The mapping must be complete.

leofang:
@kkraus14 I noticed your force-push (to stringify the tests for subprocess) still includes my WAR (checking the env var `CI`), so the tests are still not run in the CI environment, is it intended?

cuda_core/cuda/core/checkpoint.py
```diff
@@ -0,0 +1,248 @@
...
+        if "cuCheckpointProcess" in str(e) and "not found" in str(e):
+            raise RuntimeError(
+                "CUDA checkpointing is not supported by the installed NVIDIA driver. "
+                "Upgrade to a driver version with CUDA checkpoint API support."
+            ) from e
+        raise
+    return _handle_return(driver, result)
+
+
+def _handle_return(driver, result):
```
leofang: Q: Can this check be consolidated with the above one (`_call_driver`)?

cuda_core/cuda/core/checkpoint.py
```diff
@@ -0,0 +1,248 @@
...
+    Parameters
+    ----------
+    pid : int
+        Process ID of the CUDA process.
+    """
+
+    __slots__ = ("pid",)
+
+    def __init__(self, pid: int):
+        self.pid = _check_pid(pid)
```
leofang: nit: move `.pid` to `._pid` and add
```python
@property 
def pid(self):
    return self._pid
```
to make it readonly.

leofang (commented):
Since I also made some changes, would be nice for @Andy-Jost to re-review 🙂

cuda_core/cuda/core/checkpoint.py
kkraus14: Addressed in signed commit `376acc7f18`: `Process` now stores `_pid`, exposes `pid` as a read-only property, and `test_pid_is_read_only` covers assignment rejection.

cuda_core/cuda/core/checkpoint.py
kkraus14: Agreed. Addressed in signed commit `376acc7f18`: `_call_driver` is now the single checkpoint driver-call boundary for both runtime missing-symbol errors and `CUDA_ERROR_NOT_FOUND` / `CUDA_ERROR_NOT_SUPPORTED` result translation.

cuda_core/cuda/core/checkpoint.py
kkraus14: Updated in signed commit `376acc7f18`: the docs now describe the KMD visibility requirement explicitly and call out that `CUDA_VISIBLE_DEVICES` does not reduce the required mapping. The migration tests skip when `CUDA_VISIBLE_DEVICES` is set because `Device.get_all_devices()` only sees the user-space CUDA view and cannot prove the KMD-complete mapping.

kkraus14:
Addressed Leo’s latest review comments in signed commit `376acc7f18`.

On CI: the remaining `CI` skip was a temporary workaround, not the intended final state. I removed the broad CI skip after moving all driver-backed lifecycle tests onto the isolated subprocess coordinator/target harness. The parent pytest process enforces timeouts and kills the scenario process group on timeout, so CI should get checkpoint coverage without repeating the previous job-level hang. Migration tests still skip when `CUDA_VISIBLE_DEVICES` is set or when the hardware/driver cannot provide valid same-chip migration, because the restore mapping must cover the KMD-visible GPU set.

Other updates:
- `Process.pid` is now read-only via private `_pid` storage plus a property, with test coverage.
- `_call_driver` now owns checkpoint missing-symbol and unsupported-result translation in one place.
- Docs and migration tests now describe/use the KMD visibility rule instead of the looser CUDA-visible wording.
- The PR description has been updated for the current implementation and validation state.

Local validation passed: `ruff`, `git diff --check`, focused pytest (`10 passed`, `6 skipped`), and `docs-build-latest`.

kkraus14:
Follow-up CI fix pushed in signed commit `7a2e683059`.

The first re-enabled CI run showed all eight checkpoint scenario tests failing in wheel jobs before reaching the driver: the subprocess imported `cuda.core` from the source checkout, where generated `cuda.core._version` is absent, instead of importing the installed wheel. The scenario harness now runs from a neutral temp cwd and removes the checkout `cuda_core` source root from `PYTHONPATH`, so wheel jobs should exercise the installed package. Local focused validation still passes: `10 passed`, `6 skipped`.

cuda_core/tests/test_checkpoint.py
```diff
@@ -206,6 +211,28 @@ def _run_checkpoint_scenario_or_skip(body: str, *, timeout: int = 90) -> None:
         )
 
 
+def _checkpoint_scenario_env() -> dict[str, str]:
+    """Avoid source-tree shadowing when wheel jobs spawn a child Python."""
```
kkraus14: This feels super hacky and feels like it shouldn't be needed

kkraus14:
Triage for the two failed multi-GPU jobs on `7a2e6830`:

- Both failures were isolated to `tests/test_checkpoint.py::TestCheckpointGpuMigration::test_rotation_migrates_context`.
- The lifecycle checkpoint tests passed on those runners.
- The driver accepted the rotation mapping, but the target process stayed on the original GPU UUID. The swap migration test already observed the same accepted-but-no-op behavior and skipped it as a hardware/driver limitation.
- I updated rotation to use the same no-op skip path while keeping the assertion for any unexpected migrated UUID.

Local validation on the new head:

- `git diff --check`
- `pixi run ruff check cuda_core/tests/test_checkpoint.py`
- `pixi run --manifest-path cuda_core pytest cuda_core/tests/test_checkpoint.py cuda_core/tests/test_typing_imports.py` (`10 passed`, `6 skipped`)

The new signed/verified head is `8aeb8e82e1d09661289e0a4c588af6fb5bf862fc`.

Andy-Jost (approved):
LGTM!

kkraus14:
@Andy-Jost @leofang PTAL, I reworked how the checkpoint test subprocess launching works to avoid having to stringify all of the code and hopefully make it easier to maintain

leofang (approved):

PR #1986, opened 2026-04-29
cuda.core: complete the naming audit from #1945

### LINKED ISSUE #657: [DOC]: Description of `EventOptions.busy_waited_sync` is contradictory
### Is this a duplicate?

### Is this for new documentation, or an update to existing docs?

Update

### Describe the incorrect/future/missing documentation

This documentation states the following:

```
busy_waited_sync

If True, event will use blocking synchronization. When a CPU thread calls synchronize, the
call will block until the event has actually been completed. Otherwise, the CPU thread will
busy-wait until the event has been completed. (Default to False)

Type:
[bool](https://docs.python.org/3/library/functions.html#bool), optional
```

This description is confusion because the text implies: `busy_waited_sync=True -> blocking`, but from the parameter name, you would think that `busy_waited_sync=True -> busy waiting`.

The upstream docs are no help because they use the opposite convention where the flag is named `cudaEventBlockingSync` https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__EVENT.html#group__CUDART__EVENT and busy waiting seems to be the default?

### If this is a correction, please provide a link to the incorrect documentation. If this is a new documentation request, please link to where you have looked.

https://nvidia.github.io/cuda-python/cuda-core/0.2.0/generated/cuda.core.experimental.EventOptions.html

Andy-Jost:
## Summary

Address the remaining items from the `cuda.core` naming audit (#1945)
in preparation for v1.0.0. Three logical chunks, three commits:

- **Mechanical batch**: convert no-arg deterministic getters to
  properties, rename boolean / non-noun properties (with one polarity
  fix), align graph allocation/event/embed verbs across the API.
- **Conditional API**: unify the conditional-graph factory on
  `GraphCondition`, rename `if_cond` -> `if_then`, and let a
  `GraphCondition` be passed straight to `launch()`.
- **`KernelAttributes`**: convert the 16 attribute methods to
  properties; expose per-device queries via `kernel.attributes[device]`.

## Changes

**Property conversions** (no-arg deterministic getters):

- `Buffer.get_ipc_descriptor()` -> `Buffer.ipc_descriptor`
- `Event.get_ipc_descriptor()` -> `Event.ipc_descriptor`
- `DeviceMemoryResource.get_allocation_handle()` -> `.allocation_handle`
- `PinnedMemoryResource.get_allocation_handle()` -> `.allocation_handle`

**Boolean / non-noun cleanup** (with polarity correction):

- `LaunchConfig.cooperative_launch` -> `.is_cooperative` (also the
  constructor keyword argument).
- `Event.is_timing_disabled` -> `.is_timing_enabled` (polarity flipped
  to avoid the double-negative form). The constructor keyword
  `EventOptions.enable_timing` is renamed to `.timing_enabled` (bare
  adjective).
- `Event.is_sync_busy_waited` -> `.is_blocking_sync`. The new name
  matches the underlying `CU_EVENT_BLOCKING_SYNC` flag: when `True`,
  the calling CPU thread blocks (yields) on `Event.sync` instead of
  busy-waiting. The previous name was inverted with respect to its
  semantics. The constructor keyword `EventOptions.busy_waited_sync`
  is correspondingly renamed to `.blocking_sync`. This also closes
  #657.
- This standardizes the `*Options` / property convention: bare
  adjective on the option (`timing_enabled`, `blocking_sync`) and
  `is_`-prefixed boolean on the property (`is_timing_enabled`,
  `is_blocking_sync`).
- The C++ `EventBox` field/accessor pair and the Cython mirrors are
  renamed in parallel to `is_blocking_sync` /
  `get_event_is_blocking_sync`; `IPCEventDescriptor._busy_waited`
  -> `._is_blocking_sync`.

**Graph allocation method consistency** (matches `MemoryResource`):

- `GraphDefinition.alloc` / `free` -> `.allocate` / `.deallocate`
- `GraphNode.alloc` / `free` -> `.allocate` / `.deallocate`

**Cross-API consistency for graph builders**:

- `GraphBuilder.add_child` -> `.embed` (matches `GraphDefinition.embed`
  and `GraphNode.embed`).
- `GraphDefinition.record_event` / `wait_event` -> `.record` / `.wait`
  and the same on `GraphNode`, matching `Stream.record` / `Stream.wait`.

**Conditional graph API unification**:

- `GraphBuilder.create_conditional_handle()` -> `.create_condition()`,
  now returning a `GraphCondition` (matches `GraphDefinition.create_condition`).
- `if_cond` -> `if_then` on `GraphBuilder`, `GraphDefinition`, and
  `GraphNode`. The new name parallels the existing `if_else`,
  `while_loop`, and `switch` methods (verb describing the construct,
  not an abbreviation of "condition") and matches Python's own
  `if/then/else` vocabulary.
- The four conditional builder methods on `GraphBuilder` (`if_then`,
  `if_else`, `while_loop`, `switch`) now accept a `GraphCondition`
  instead of a raw `CUgraphConditionalHandle`, with a TypeError check
  that mirrors the one already present in `_make_conditional_node`.
- A `GraphCondition` can be passed directly as a kernel argument;
  the kernel argument handler grew a `GraphCondition` fast path that
  reads the underlying `CUgraphConditionalHandle` directly. No
  `__int__` is exposed on the public class.

**`KernelAttributes` redesign**:

- The 16 attribute methods (`max_threads_per_block`, `num_regs`,
  `cluster_scheduling_policy_preference`, etc.) that previously took
  a `device_id` argument are now `@property`. The view returned by
  `Kernel.attributes` is bound to the current device by default
  (resolved at attribute-access time, so it follows context changes);
  `kernel.attributes[device]` returns a sibling view bound to a
  specific `Device` or device ordinal, sharing the underlying cache.
- Old: `kernel.attributes.num_regs()`, `kernel.attributes.num_regs(some_dev)`
- New: `kernel.attributes.num_regs`, `kernel.attributes[some_dev].num_regs`

**Docs**:

- New "Breaking changes" entries in `1.0.0-notes.rst` covering every
  rename above with issue refs.
- Three older release notes (`0.3.0`, `0.4.0`, `0.7.0`) had stray
  Sphinx cross-references to renamed/removed names; converted to inline
  literals so latest-only Sphinx builds keep working without rewriting
  history.

No backward-compatibility aliases (pre-1.0).

## Test Coverage

- Existing tests updated to the new names throughout
  (`test_event.py`, `test_memory.py`, `test_module.py`, `test_launcher.py`,
  `test_object_protocols.py`, the `tests/graph/` and
  `tests/memory_ipc/` suites).
- `pytest.raises` blocks adjusted (`_ = obj.property`) where property
  access replaced a method call, to satisfy `ruff B018`.
- `test_read_only_kernel_attributes` (parametrized over all 16
  attributes) was rewritten to exercise the new shape directly:
  property access on the default view and indexed access through
  `kernel.attributes[device]` and `kernel.attributes[device.device_id]`,
  with a type assertion at every access path (the original test only
  checked the loop's last value).
- New tests:
  - `test_kernel_attributes_per_device_view` exercises the
    `kernel.attributes[device]` shape and value parity with the
    default view.
  - `test_kernel_attributes_indexing_rejects_invalid_device`
    confirms `kernel.attributes["bad"]` errors via the `Device(...)`
    constructor.
- Verified locally on a GPU host: `tests/`, `tests/graph/`, and
  `tests/memory_ipc/` all green.
- Editable build clean, pre-commit clean (all-files).

## Related Work

- Closes the items assigned to me (and Leo's `get_allocation_handle`)
  in #1945, except the `DeviceProperties.cooperative_launch` rename
  (deferred pending team discussion) and the system/fan items (Mike).
- Builds on #1984 (`GraphDef` -> `GraphDefinition`,
  `Condition` -> `GraphCondition`), which is already merged.
- Closes #657 (the contradictory `EventOptions.busy_waited_sync`
  description is resolved by the rename to `EventOptions.blocking_sync`
  / `Event.is_blocking_sync`).


cuda_core/cuda/core/graph/_graph_definition.pyx
```diff
@@ -48,6 +61,9 @@ cdef class GraphCondition:
     def __hash__(self) -> int:
         return hash(<unsigned long long>self._c_handle)
 
+    def __int__(self) -> int:
+        return <unsigned long long>self._c_handle
```
mdboom: I'm not crazy about `int` implicitly returning the underlying handle (which should be an implementation detail).  The other classes (of ours) that need this, can just reach to `_c_handle`, which would avoid making it accessible from Python.

cuda_core/cuda/core/_event.pyx
```diff
@@ -100,40 +100,41 @@ cdef class Event:
         cdef Event self = cls.__new__(cls)
         cdef EventOptions opts = check_or_create_options(EventOptions, options, "Event options")
         cdef unsigned int flags = 0x0
-        cdef bint timing_disabled = False
-        cdef bint busy_waited = False
+        cdef bint timing_enabled = True
+        cdef bint uses_blocking_sync = False
```
mdboom: Minor nit bc it's not API, but there is `uses_blocking_sync` vs. `use_blocking_sync` here.

mdboom (changes requested):
I think we shouldn't add `__int__` to get underlying handles and pointers...  It doesn't look like we do that elsewhere in `cuda-core` (`cuda-bindings` does it in a bunch of places).

Other than that, LGTM.

cuda_core/cuda/core/_event.pyx
Andy-Jost: Good catch. This points to larger issue in the API about matching language between `*Options` classes and the properties that reflect those options. 

Here's what appears in events:

| `EventOptions` | `Event` |
| ------------- | ------------- |
| `enable_timing` | `is_timing_enabled` |
| `use_blocking_sync` | `uses_blocking_sync` |
| `ipc_enabled` | `is_ipc_enabled`

The code base consistently uses **adjective** -> **is_ + adjective**, as in `ipc_enabled` -> `is_ipc_enabled`, and `nonblocking` -> `is_nonblocking`. The imperative forms (e.g. `enable_timing`) are the odd ones.

Following that, I'll change events to say `timing_enabled` -> `is_timing_enabled` and `blocking_sync` ->`is_blocking_sync`.

Side note: `uses_blocking_sync` would be the only instance of `uses_`. While `is_blocking_sync` is slightly unnatural, this option is overall more consistent.

cuda_core/cuda/core/graph/_graph_definition.pyx
Andy-Jost: I'll remove this

mdboom (approved):
👍

PR #1987, opened 2026-04-29
Add nightly CI for optional-dependency testing (PyTorch, numba-cuda)

leofang:
Add a nightly CI pipeline that tests cuda-python wheels against optional dependencies (PyTorch and numba-cuda) without rebuilding wheels. Wheels are downloaded from the latest successful CI run on `main`.

## Design

- **`ci-nightly.yml`**: New orchestrator workflow (2 AM UTC daily + `workflow_dispatch` for manual testing). Finds the latest successful CI run on main via `ci/tools/lookup-run-id --branch` and passes its run-id + head SHA to the existing test workflows.
- **`test-wheel-linux/windows.yml`**: Extended with three new inputs:
  - `run-id`: enables `actions/download-artifact` to pull wheels from a different workflow run
  - `sha`: commit SHA for constructing artifact names (defaults to `github.sha`)
  - `test-mode`: `standard` (default, current behavior), `nightly-pytorch`, or `nightly-numba-cuda`
- **`test-matrix.yml`**: New `nightly:` entries with a `MODE` field. The orchestrator uses the existing `matrix_filter` input to select by mode.
- **`ci/tools/run-tests`**: New `nightly-pytorch` and `nightly-numba-cuda` modes that install all wheels (pathfinder + bindings + core) and the optional dep in a single pip call. Handles `BINDINGS_SOURCE` (main/backport/published) for correct bindings resolution.
- **`ci/tools/lookup-run-id`**: Extended with `--branch` mode (find latest successful run on a branch) alongside the existing `--tag` mode. All inline `gh run list` patterns across `ci-nightly.yml`, `test-wheel-linux.yml`, `test-wheel-windows.yml`, and `build-wheel.yml` are consolidated into this script. Windows backport download step converted from PowerShell to bash.

## Additional changes

- **Job names**: Use descriptive format: `Python 3.12, CUDA 12.6.3 (wheels), GPU l4, 2.11.0`
- **Display step**: Guarded with `if: BINDINGS_SOURCE != 'published'` (no local wheel for published mode)
- **VC++ Redistributable**: Installed on Windows before PyTorch (required per pytorch/pytorch#166628)

### PyTorch (12 jobs: 4 linux-64 + 4 linux-aarch64 + 4 win-64)
| TORCH_VER | TORCH_CUDA | CUDA_VER | Platforms |
|-----------|------------|----------|-----------|
| 2.11.0 | cu126 | 12.6.3 | linux-64, linux-aarch64, win-64 |
[...]

PR #1988, opened 2026-04-29
Fix tensor bridge DLL import failure on Windows

leofang:
Follow-up of #1894.

`aoti_torch_get_current_cuda_stream` lives in `torch_cuda.dll` / `libtorch_cuda.so`, not `torch_cpu.dll`. Linking against it at build time breaks CPU-only PyTorch installs and caused `ImportError: DLL load failed` on Windows.

## Root cause

The `.def` file declared all AOTI symbols under `LIBRARY torch_cpu.dll`, but `aoti_torch_get_current_cuda_stream` is CUDA-specific. On Linux this wasn't visible because `RTLD_GLOBAL` makes all symbols visible across shared libraries, but Windows DLL imports are strict.

## Fix

Resolve `aoti_torch_get_current_cuda_stream` lazily at runtime instead of link-time:
- Inline C helper using `LoadLibrary` + `GetProcAddress` (Windows) / `dlsym(RTLD_DEFAULT, ...)` (Linux)
- Cached function pointer — resolved once on first CUDA tensor stream sync, then pure C on subsequent calls
- CPU-only PyTorch works fine: the resolution only triggers when a CUDA tensor actually needs stream ordering

## Changes

- `_tensor_bridge.pyx`: Remove static extern declaration, add inline C resolver + cached function pointer in `sync_torch_stream()`
- `aoti_shim.h`: Remove `aoti_torch_get_current_cuda_stream` declaration, add explanatory comments
- `aoti_shim.def`: Remove the symbol (already done in previous commit)
- `build_hooks.py`: Revert to single-def-file (no second stub lib needed)

## Testing

Verified on Windows with both CUDA-enabled and CPU-only PyTorch:
- CUDA tensors: `StridedMemoryView.from_dlpack(cuda_tensor)` works correctly
- CPU tensors: module imports and works without triggering CUDA resolution
- CPU-only PyTorch: no import errors

-- Leo's bot

### DIFF
--- a/cuda_core/build_hooks.py
+++ b/cuda_core/build_hooks.py
@@ -183,19 +183,22 @@ def _build_cuda_core(debug=False):
         # related to free-threading builds.
         extra_compile_args += ["-DCYTHON_TRACE_NOGIL=1", "-DCYTHON_USE_SYS_MONITORING=0"]
 
-    # On Windows, _tensor_bridge.pyx needs a stub import library so the MSVC
-    # linker can resolve the AOTI symbols (they live in torch_cpu.dll at
-    # runtime).  We generate the .lib from a .def file at build time.
+    # On Windows, _tensor_bridge.pyx needs stub import libraries so the MSVC
+    # linker can resolve the AOTI symbols at link time.  At runtime the symbols
+    # resolve from the actual DLLs loaded by 'import torch'.
+    #   - aoti_shim.def       -> torch_cpu.dll   (dtype, device, tensor metadata)
+    #   - aoti_shim_cuda.def  -> torch_cuda.dll   (CUDA stream access)
     _aoti_extra_link_args = []
     if sys.platform == "win32":
-        _def_file = os.path.join("cuda", "core", "_include", "aoti_shim.def")
-        _lib_file = os.path.join("build", "aoti_shim.lib")
         os.makedirs("build", exist_ok=True)
-        subprocess.check_call(  # noqa: S603
-            ["lib", f"/DEF:{_def_file}", f"/OUT:{_lib_file}", "/MACHINE:X64"],  # noqa: S607
-            stdout=subprocess.DEVNULL,
-        )
-        _aoti_extra_link_args = [_lib_file]
+        for def_name in ("aoti_shim", "aoti_shim_cuda"):
+            def_file = os.path.join("cuda", "core", "_include", f"{def_name}.def")
+            lib_file = os.path.join("build", f"{def_name}.lib")
+            subprocess.check_call(  # noqa: S603
+                ["lib", f"/DEF:{def_file}", f"/OUT:{lib_file}", "/MACHINE:X64"],  # noqa: S607
+                stdout=subprocess.DEVNULL,
+            )
+            _aoti_extra_link_args.append(lib_file)
 
     def get_extra_link_args(mod_name):
         if mod_name == "_tensor_bridge" and _aoti_extra_link_args:
--- a/cuda_core/build_hooks.py
+++ b/cuda_core/build_hooks.py
@@ -183,22 +183,22 @@ def _build_cuda_core(debug=False):
         # related to free-threading builds.
         extra_compile_args += ["-DCYTHON_TRACE_NOGIL=1", "-DCYTHON_USE_SYS_MONITORING=0"]
 
-    # On Windows, _tensor_bridge.pyx needs stub import libraries so the MSVC
-    # linker can resolve the AOTI symbols at link time.  At runtime the symbols
-    # resolve from the actual DLLs loaded by 'import torch'.
-    #   - aoti_shim.def       -> torch_cpu.dll   (dtype, device, tensor metadata)
-    #   - aoti_shim_cuda.def  -> torch_cuda.dll   (CUDA stream access)
+    # On Windows, _tensor_bridge.pyx needs a stub import library so the MSVC
+    # linker can resolve the AOTI symbols (they live in torch_cpu.dll at
+    # runtime).  We generate the .lib from a .def file at build time.
+    # Note: aoti_torch_get_current_cuda_stream lives in torch_cuda.dll and
+    # is resolved lazily at runtime (not via the stub lib) — see
+    # _tensor_bridge.pyx.
     _aoti_extra_link_args = []
     if sys.platform == "win32":
+        _def_file = os.path.join("cuda", "core", "_include", "aoti_shim.def")
+        _lib_file = os.path.join("build", "aoti_shim.lib")
         os.makedirs("build", exist_ok=True)
-        for def_name in ("aoti_shim", "aoti_shim_cuda"):
-            def_file = os.path.join("cuda", "core", "_include", f"{def_name}.def")
-            lib_file = os.path.join("build", f"{def_name}.lib")
-            subprocess.check_call(  # noqa: S603
-                ["lib", f"/DEF:{def_file}", f"/OUT:{lib_file}", "/MACHINE:X64"],  # noqa: S607
-                stdout=subprocess.DEVNULL,
-            )
-            _aoti_extra_link_args.append(lib_file)
+        subprocess.check_call(  # noqa: S603
+            ["lib", f"/DEF:{_def_file}", f"/OUT:{_lib_file}", "/MACHINE:X64"],  # noqa: S607
+            stdout=subprocess.DEVNULL,
+        )
+        _aoti_extra_link_args = [_lib_file]
 
     def get_extra_link_args(mod_name):
         if mod_name == "_tensor_bridge" and _aoti_extra_link_args:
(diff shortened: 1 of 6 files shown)


leofang:
/ok to test [77c0b7b](https://github.com/NVIDIA/cuda-python/pull/1988/commits/77c0b7bfcdfa46f36febe44e1c4a8d5193f079ff)

leofang:
This bug is discovered by #1987:
https://github.com/NVIDIA/cuda-python/actions/runs/25112913255/job/73591754844?pr=1987#step:33:374

cuda_core/build_hooks.py
```diff
@@ -183,19 +183,22 @@ def get_sources(mod_name):
         # related to free-threading builds.
         extra_compile_args += ["-DCYTHON_TRACE_NOGIL=1", "-DCYTHON_USE_SYS_MONITORING=0"]
 
-    # On Windows, _tensor_bridge.pyx needs a stub import library so the MSVC
-    # linker can resolve the AOTI symbols (they live in torch_cpu.dll at
-    # runtime).  We generate the .lib from a .def file at build time.
+    # On Windows, _tensor_bridge.pyx needs stub import libraries so the MSVC
+    # linker can resolve the AOTI symbols at link time.  At runtime the symbols
+    # resolve from the actual DLLs loaded by 'import torch'.
+    #   - aoti_shim.def       -> torch_cpu.dll   (dtype, device, tensor metadata)
```
rwgk: Did you consider renaming the existing file to aoti_shim_cpu.def, for symmetry?

Related but different: this comment in `cuda_core/cuda/core/_include/aoti_shim.h` isn't accurate anymore:

```
/*
 * IMPORTANT: Keep the AOTI_SHIM_API declaration list below in sync with
 * aoti_shim.def. On Windows, build_hooks.py turns that .def file into the
 * stub import library that MSVC needs to link _tensor_bridge without making
 * PyTorch a build-time dependency. If you add, remove, or rename an imported
 * AOTI symbol here, update aoti_shim.def in the same change.
 */
```

I'd simply change it to `aoti_shim*.def` (two locations) and "... the .def files into stub import libraries ..."

rwgk (commented):
Generated with Cursor GPT-5.4 Extra High Fast with a few rounds of prompting:

___

Is the intent that this Windows fast path only supports CUDA-enabled PyTorch?

As written, `_tensor_bridge` now links against `torch_cuda.dll`, but the fast path is still taken for any supported `torch.Tensor`, including CPU tensors. On Windows, that looks like it could make the module fail to load under CPU-only PyTorch, or even for CPU-tensor-only use, before the code can reach the CPU branch. Could this be avoided by keeping the base bridge linked only to `torch_cpu.dll` and loading or resolving the CUDA-specific symbol lazily only when a CUDA tensor actually needs stream synchronization?

leofang:
> Is the intent that this Windows fast path only supports CUDA-enabled PyTorch?

Good idea. Fixed in commit 50294b8.

cuda_core/build_hooks.py
leofang: This is moot after commit 50294b8.

cuda_core/cuda/core/_tensor_bridge.pyx
```diff
@@ -274,10 +304,17 @@ cpdef int sync_torch_stream(int32_t device_index,
     the consumer stream wait on it.  This is a no-op if both streams are
     the same.
     """
+    global _cached_get_cuda_stream
     cdef void* producer_s
     cdef EventHandle h_event
 
-    check_aoti(aoti_torch_get_current_cuda_stream(device_index, &producer_s),
+    if _cached_get_cuda_stream == NULL:
+        _cached_get_cuda_stream = <_get_cuda_stream_fn_t>_resolve_cuda_stream_fn()
```
rwgk, line 312: We have `"freethreading_compatible": True` in `cuda_core/build_hooks.py`, therefore this can be a C-level data race.

Could we leave it as-is?

This is what Cursor GPT-5.4 Extra High Fast said when I asked about "worst-case consequences":

___

If both threads ultimately resolve the same correct function address, there are two answers.

The formal C answer is still: **undefined behavior**.

C does not say "same value is okay." Unsynchronized concurrent read/write or write/write to the same object is a data race regardless of whether the bits happen to match. Once you have that, the compiler is allowed to assume the race does not exist.

For this exact snippet in `cuda_core/cuda/core/_tensor_bridge.pyx`, the practical risk is probably much smaller than the formal one:

- most likely, both threads just call `_resolve_cuda_stream_fn()`, both get the same pointer, and both store it;
- on Windows that may mean an extra `LoadLibraryA("torch_cuda.dll")` refcount bump;
- after that, things keep working.

The worst credible consequences are:

- a thread observes a stale or torn value and calls through a bad pointer or `NULL`, leading to a crash;
- compiler optimization based on the "no races exist" assumption produces behavior that is not obviously tied to the source;
- a thread sees the uninitialized state longer than expected and does duplicate initialization or raises spuriously.

That said, in this particular pattern, on mainstream 64-bit Linux/Windows builds, I would rate the real-world risk as "probably benign most of the time," not "likely to explode." It is still a real bug for a module that claims free-threading compatibility, but it is more of a correctness-contract violation / sanitizer-finding than an obvious imminent crash bug.

___

I also asked Cursor for a fix, this is what it generated (untested):

```diff
diff --git a/cuda_core/cuda/core/_tensor_bridge.pyx b/cuda_core/cuda/core/_tensor_bridge.pyx
index 07eec56537b..THREADSAFE000 100644
--- a/cuda_core/cuda/core/_tensor_bridge.pyx
+++ b/cuda_core/cuda/core/_tensor_bridge.pyx
@@
 ctypedef AOTITorchError (*_get_cuda_stream_fn_t)(int32_t, void**) nogil
 
 cdef extern from *:
     """
     #ifdef _WIN32
     #include <windows.h>
-    static void* _resolve_cuda_stream_fn(void) {
-        HMODULE h = LoadLibraryA("torch_cuda.dll");
-        if (!h) return NULL;
-        return (void*)GetProcAddress(h, "aoti_torch_get_current_cuda_stream");
-    }
+    static INIT_ONCE _cuda_stream_init_once = INIT_ONCE_STATIC_INIT;
+    static void* _cached_cuda_stream_fn = NULL;
+
+    static BOOL CALLBACK _init_cuda_stream_fn(
+            PINIT_ONCE init_once, PVOID param, PVOID* context) {
+        HMODULE h = LoadLibraryA("torch_cuda.dll");
+        if (h) {
+            _cached_cuda_stream_fn =
+                (void*)GetProcAddress(h, "aoti_torch_get_current_cuda_stream");
+        }
+        return TRUE;
+    }
+
+    static void* _resolve_cuda_stream_fn(void) {
+        InitOnceExecuteOnce(&_cuda_stream_init_once, _init_cuda_stream_fn, NULL, NULL);
+        return _cached_cuda_stream_fn;
+    }
     #else
     #include <dlfcn.h>
+    #include <pthread.h>
     #ifndef RTLD_DEFAULT
     #define RTLD_DEFAULT ((void*)0)
     #endif
+    static pthread_once_t _cuda_stream_once = PTHREAD_ONCE_INIT;
+    static void* _cached_cuda_stream_fn = NULL;
+
+    static void _init_cuda_stream_fn(void) {
+        _cached_cuda_stream_fn =
+            dlsym(RTLD_DEFAULT, "aoti_torch_get_current_cuda_stream");
+    }
+
     static void* _resolve_cuda_stream_fn(void) {
-        return dlsym(RTLD_DEFAULT, "aoti_torch_get_current_cuda_stream");
+        pthread_once(&_cuda_stream_once, _init_cuda_stream_fn);
+        return _cached_cuda_stream_fn;
     }
     #endif
     """
     void* _resolve_cuda_stream_fn() nogil
-
-cdef _get_cuda_stream_fn_t _cached_get_cuda_stream = NULL
 
 @@
 cpdef int sync_torch_stream(int32_t device_index,
                             intptr_t consumer_s) except? -1:
 @@
-    global _cached_get_cuda_stream
     cdef void* producer_s
     cdef EventHandle h_event
+    cdef _get_cuda_stream_fn_t get_cuda_stream
 
-    if _cached_get_cuda_stream == NULL:
-        _cached_get_cuda_stream = <_get_cuda_stream_fn_t>_resolve_cuda_stream_fn()
-        if _cached_get_cuda_stream == NULL:
-            raise RuntimeError(
-                "Cannot resolve aoti_torch_get_current_cuda_stream from "
-                "torch_cuda — is CUDA-enabled PyTorch installed?")
-    check_aoti(_cached_get_cuda_stream(device_index, &producer_s),
+    get_cuda_stream = <_get_cuda_stream_fn_t>_resolve_cuda_stream_fn()
+    if get_cuda_stream == NULL:
+        raise RuntimeError(
+            "Cannot resolve aoti_torch_get_current_cuda_stream from "
+            "torch_cuda — is CUDA-enabled PyTorch installed?")
+    check_aoti(get_cuda_stream(device_index, &producer_s),
                b"aoti_torch_get_current_cuda_stream")
```

rwgk (approved):
Approving.

For peace of mind, I'd fix the race, but it'd seem defendable to assume that the C-level data race under free-threading is benign.

leofang:
Thanks, Ralf!

> For peace of mind, I'd fix the race, but it'd seem defendable to assume that the C-level data race under free-threading is benign.

Yeah I don't think it would cause issues in practice. Let's see who complains... This would have been a non-issue if we feed the shim header to cybind and generate cython/internal bindings, where we take good care of protecting the module-level variables.

PR #1992, opened 2026-04-30
BUG: NVML system events may only be registered once per process

mdboom:
This should fix the nightly tests that are currently failing.


leofang (approved):
Thanks, Mike!

Issue #1951, opened 2026-04-17
cuda.core: audit of C-style error-code return patterns in public API


mdboom:
## Motivation

cuda.core is intended to be a high-level Pythonic wrapper around lower-level bindings in cuda.bindings. In idiomatic Python, errors should be communicated via exceptions rather than requiring callers to inspect return values. This audit looked at all public functions and methods in cuda.core for places where the C convention of returning error/status codes leaks through — or more broadly, anywhere the caller must inspect the returned object for correctness rather than relying on exception flow.

## Summary

The codebase is largely well-designed. The `HANDLE_RETURN()` macro and `handle_return()` function consistently convert CUDA error codes into Python exceptions across the vast majority of the API. However, there are several notable deviations.

### 1. `Event.is_done` — boolean derived from CUDA error code

`_event.pyx`: Converts `CUDA_SUCCESS` → `True` and `CUDA_ERROR_NOT_READY` → `False`. The caller must inspect the return value rather than relying on exception flow. This is a common idiom in async GPU APIs and is arguably reasonable for polling, but it is worth noting as a deliberate deviation from pure exception-based error handling.

### 2. `Program.pch_status` — string status code the caller must interpret

`_program.pyx`: Returns `"created"`, `"not_attempted"`, `"failed"`, or `None`. The `"failed"` case is notable — PCH creation failure is reported as a string value rather than raised as an exception. The caller must know to check for `"failed"` and handle it. Internally, the helper `_read_pch_status()` also uses `None` as a sentinel for "heap exhausted, retry needed" (a classic C-style error pattern, though internal-only).

### 3. `Linker.get_error_log()` / `get_info_log()` — unchecked CUDA calls

`_linker.pyx`: These return diagnostic strings, but the underlying CUDA calls to `nvJitLinkGetErrorLogSize` / `nvJitLinkGetErrorLog` are not checked via `HANDLE_RETURN` — the results are used directly without error checking. If these calls fail, the failure is silently ignored.

### 4. `_MP_deallocate` silently swallows `CUDA_ERROR_INVALID_CONTEXT`

`_memory_pool.pyx`: The deallocation path explicitly suppresses `CUDA_ERROR_INVALID_CONTEXT`. The function is marked `noexcept` so it cannot raise, but this means a real error (e.g., deallocating after context destruction) is silently ignored. Callers have no way to know deallocation failed.

[...]

PR #1993, opened 2026-04-30
BUG: Handle error return in nvJitLinkGetErrorLogSize and nvJitLinkGetErrorLog

mdboom:
While not strictly an API issue, this was found as part of the API audit in #1951.


Andy-Jost (approved):
✔️

PR #1999, opened 2026-04-30
Fix torch-incompatible assertions in TestViewCudaArrayInterfaceGPU

leofang:
The `_check_view` method in `TestViewCudaArrayInterfaceGPU` was missed during the tensor bridge refactor (#1894) and still used raw numpy attributes that don't work with `torch.Tensor`:

- `in_arr.size` — torch's `.size` is a method (returns shape), not a property (returns element count)
- `in_arr.strides` / `in_arr.dtype.itemsize` — torch tensors don't have numpy-style `.strides`
- `in_arr.flags["C_CONTIGUOUS"]` — torch tensors don't have `.flags`
- `gpu_array_ptr(in_arr)` — didn't handle torch tensors
- `in_arr.dtype` — torch dtype != numpy dtype

## Changes

- Replace raw array accesses with `_arr_*` helpers (`_arr_ptr`, `_arr_size`, `_arr_strides_in_counts`, `_arr_is_c_contiguous`)
- Add `_arr_dtype` helper: uses `__cuda_array_interface__["typestr"]` for torch tensors to get a numpy-compatible dtype
- Flip strides assertion to branch on `view.strides is None` (per review) — clearer semantics, consistent across all three `_check_view` methods
- Restore dtype assertion using `_arr_dtype`

## Testing

Verified locally: 76/76 tests pass across `TestViewCPU`, `TestViewGPU`, and `TestViewCudaArrayInterfaceGPU` (including all torch and cupy variants).

Caught by the nightly optional-dependency CI (#1987).

-- Leo's bot


leofang:
/ok to test [dacfec9](https://github.com/NVIDIA/cuda-python/pull/1999/commits/dacfec9796274c71d5af75735704e0fa2c797316)

cuda_core/tests/test_utils.py
```diff
@@ -171,7 +171,10 @@ def _check_view(self, view, in_arr):
         assert view.shape == expected_shape
         assert view.size == _arr_size(in_arr)
         strides_in_counts = _arr_strides_in_counts(in_arr)
-        assert (_arr_is_c_contiguous(in_arr) and view.strides is None) or view.strides == strides_in_counts
+        if _arr_is_c_contiguous(in_arr):
+            assert view.strides in (None, strides_in_counts)
+        else:
+            assert view.strides == strides_in_counts
```
rwgk: Could we flip this to branch on `view.strides is None` instead?

```suggestion
        if view.strides is None:
            assert _arr_is_c_contiguous(in_arr)
        else:
            assert view.strides == strides_in_counts
```

This reads a bit more directly to me, and `==` expresses the explicit-strides case more clearly, while also avoiding the small semantic gap between equality and membership checks.

cuda_core/tests/test_utils.py
```diff
@@ -343,15 +346,15 @@ def test_cuda_array_interface_gpu(self, in_arr, use_stream):
...
-            assert view.strides is None
+        assert view.ptr == _arr_ptr(in_arr)
+        expected_shape = tuple(in_arr.shape)
+        assert view.shape == expected_shape
+        assert view.size == _arr_size(in_arr)
+        strides_in_counts = _arr_strides_in_counts(in_arr)
+        if _arr_is_c_contiguous(in_arr):
+            assert view.strides in (None, strides_in_counts)
         else:
             assert view.strides == strides_in_counts
```
rwgk: Same suggestion as above.

cuda_core/tests/test_utils.py
```diff
@@ -343,15 +346,15 @@ def test_cuda_array_interface_gpu(self, in_arr, use_stream):
...
-        assert view.dtype == in_arr.dtype
```
rwgk, line 354: Could this work, so we don't lose this `assert`?

```python
        expected_dtype = np.dtype(in_arr.__cuda_array_interface__["typestr"])
        assert view.dtype == expected_dtype
```

Or maybe add an `_arr_dtype` helper, similar to the other helpers?

cuda_core/tests/test_utils.py
```diff
@@ -343,15 +346,15 @@ def test_cuda_array_interface_gpu(self, in_arr, use_stream):
 
     def _check_view(self, view, in_arr, dev):
         assert isinstance(view, StridedMemoryView)
-        assert view.ptr == gpu_array_ptr(in_arr)
-        assert view.shape == in_arr.shape
-        assert view.size == in_arr.size
-        strides_in_counts = convert_strides_to_counts(in_arr.strides, in_arr.dtype.itemsize)
-        if in_arr.flags["C_CONTIGUOUS"]:
```
rwgk, line 350: Does the replacement still behaves correctly for Numba arrays? Cursor is telling me `numba.cuda` device arrays have a mapping-style `flags`, so key access like `flags["C_CONTIGUOUS"]` works, but attribute access like `flags.c_contiguous` does not. If that's right, would it be safer for `_arr_is_c_contiguous()` to use:

```python
return arr.flags.c_contiguous if hasattr(arr.flags, "c_contiguous") else arr.flags["C_CONTIGUOUS"]
```

That seems like it would preserve the NumPy/CuPy behavior while still handling a Numba-style `flags` object.

Andy-Jost (approved):

cuda_core/tests/test_utils.py
leofang: Done — flipped all three `_check_view` methods to branch on `view.strides is None`. Agreed it reads more directly.

cuda_core/tests/test_utils.py
leofang: Done — applied the same flip to all three `_check_view` methods.

cuda_core/tests/test_utils.py
leofang: Good idea — added an `_arr_dtype` helper that uses `__cuda_array_interface__["typestr"]` for torch tensors (gives a numpy-compatible dtype), and restored the assertion. 76/76 tests pass locally.

cuda_core/tests/test_utils.py
leofang: This was fixed upstream in #1998 (`hasattr(arr.flags, "c_contiguous")` instead of `hasattr(arr, "flags")`). Merged main into this branch to pick it up.

rwgk (approved):

Andy-Jost (approved):

PR #2008, opened 2026-05-01
cuda.core: Cythonize GraphBuilder and Graph with handle-layer cleanup

Andy-Jost:
## Summary

Convert `GraphBuilder` and `Graph` from Python classes (using `_MembersNeededForFinalize` + `weakref.finalize`) to Cython `cdef class` objects backed by typed C++ resource handles.

This does two things. First, it lays groundwork for step 3 of #1330 (graph updates) by giving graph objects the same handle-based ownership pattern as the rest of `cuda.core`. Second, it clarifies `GraphBuilder`'s state machine: what used to be a tangle of implicit flags and conditional cleanup paths is now two orthogonal enums — `_BuilderKind` (`PRIMARY`/`FORKED`/`CONDITIONAL_BODY`) describing how the builder was created, and `_CaptureState` (`CAPTURE_NOT_STARTED`/`CAPTURING`/`CAPTURE_ENDED`) tracking the capture lifecycle. Methods can now check exactly the state they care about, illegal transitions are detectable, and `__dealloc__` has a single, well-defined condition for ending capture.

Removing `weakref.finalize` and moving destroy into refcounted `GraphHandle` / `GraphExecHandle` deleters also closes Glasswing finding V18.1 (CWE-415 double-destroy on re-entrant `close()`).

## Changes

- Add `GraphExecHandle` to the resource-handle layer (`_cpp/resource_handles.{hpp,cpp}`, `_resource_handles.{pxd,pyx}`), wrapping `CUgraphExec` with a `cuGraphExecDestroy`-based deleter run under `GILReleaseGuard`.
- `GraphBuilder` becomes a `cdef class` with the explicit `_BuilderKind`/`_CaptureState` enums described above. Live-API methods (`begin_building`, `end_building`, `embed`, `split`, `join`, etc.) move to `nogil` `cydriver` paths where practical, and end-of-capture in `__dealloc__` runs against the cached `StreamHandle` rather than reaching into a possibly-cleared `Stream` attribute.
- `Graph` becomes a `cdef class` holding `GraphExecHandle _h_graph_exec` directly; `update`/`upload`/`launch` move to `nogil` `cydriver`. `weakref.finalize` is gone.
- `Device.create_graph_builder` and `Stream.create_graph_builder` cimport `GraphBuilder` and call its `_init` factory; quoted forward-reference annotations are removed (clears Cython "Strings should no longer be used for type declarations" warnings).
- Add `test_graph_close_is_idempotent` to verify repeated `Graph.close()` does not double-destroy the exec handle.

## Test Coverage

- `test_graph_close_is_idempotent` — repeated `Graph.close()` is safe and clears the exec handle (Glasswing V18.1 regression guard).

## Related work

- Groundwork for step 3 of #1330.
- Closes NVIDIA/cuda-python-private#370 (Glasswing V18.1 — Graph re-entrant double-destroy, NVBugs [6268912](https://nvbugspro.nvidia.com/bug/6268912)).


cuda_core/cuda/core/graph/_graph_builder.pyx
```diff
@@ -198,63 +236,48 @@ class GraphBuilder:
...
-        self = cls.__new__(cls)
-        self._mnff = GraphBuilder._MembersNeededForFinalize(
-            self, stream, is_stream_owner, conditional_graph, is_join_required
-        )
+    def __dealloc__(self):
+        # Note: _stream could be set to None by cyclic-GC tp_clear before
+        # __dealloc__, but _h_stream is guaranteed to be valid.
+        if self._h_stream and self._state == CAPTURING and self._kind != FORKED:
+            with nogil:
+                cydriver.cuStreamEndCapture(as_cu(self._h_stream), NULL)
```
leofang: 1. Shouldn't we set `_state` and `_kind` to a guarding value to prevent the destructor from being invoked again?
2. If a builder owns a stream, shouldn't we free the stream too (which is the old behavior)?

leofang: 3. Is this non-public behavior that `cuStreamEndCapture` accepts a nullptr for `CUgraph*`? I don't think it's documented.

cuda_core/cuda/core/graph/_graph_builder.pyx
```diff
@@ -198,63 +236,48 @@ class GraphBuilder:
...
 
+    def close(self):
+        """Destroy the graph builder."""
+        if self._h_stream and self._state == CAPTURING and self._kind != FORKED:
+            with nogil:
+                HANDLE_RETURN(cydriver.cuStreamEndCapture(as_cu(self._h_stream), NULL))
+        self._h_graph.reset()
+        self._h_stream.reset()
+        self._state = CAPTURE_ENDED
+        self._stream = None
```
leofang, line 273: Let's share the same destructor for both `close` and `__dealloc__` like what we do elsewhere in the codebase, instead of repeating.

cuda_core/cuda/core/graph/_graph_builder.pyx
```diff
@@ -162,7 +165,9 @@ def _instantiate_graph(h_graph, options: GraphCompleteOptions | None = None) ->
             flags |= driver.CUgraphInstantiate_flags.CUDA_GRAPH_INSTANTIATE_FLAG_USE_NODE_PRIORITY
         params.flags = flags
 
-    graph = Graph._init(handle_return(driver.cuGraphInstantiateWithParams(h_graph, params)))
+    py_exec = handle_return(driver.cuGraphInstantiateWithParams(h_graph, params))
+    c_exec = <cydriver.CUgraphExec><intptr_t>int(py_exec)
+    graph = Graph._init(c_exec)
```
leofang: Pre-existing pattern (the original wrapped before checking too), but worth fixing while we're here: if `params.result_out` indicates an instantiate error, `c_exec` may be NULL or otherwise invalid — yet we've already wrapped it in `Graph._init`, whose new RAII deleter will call `cuGraphExecDestroy` on it during the exception unwind that follows. Suggest checking `result_out` first and only constructing the `Graph` on `SUCCESS`.

cuda_core/cuda/core/graph/_graph_builder.pyx
```diff
@@ -198,63 +236,48 @@ class GraphBuilder:
...
-            if self.graph:
-                handle_return(driver.cuGraphDestroy(self.graph))
-            self.graph = None
-            self.conditional_graph = None
-
-    __slots__ = ("__weakref__", "_building_ended", "_mnff")
-
     def __init__(self):
         raise NotImplementedError(
             "directly creating a Graph object can be ambiguous. Please either "
```
leofang: Nit: this is `GraphBuilder.__init__`, so the message should say "directly creating a *GraphBuilder* object" — currently mirrors the wording from the old class, but it's confusing now that there's a distinct `Graph` class right below (line 830) with its own message saying "directly constructing a Graph instance".

cuda_core/cuda/core/graph/_graph_builder.pyx
```diff
@@ -198,63 +236,48 @@ class GraphBuilder:
...
         return self
 
+    def close(self):
+        """Destroy the graph builder."""
+        if self._h_stream and self._state == CAPTURING and self._kind != FORKED:
+            with nogil:
+                HANDLE_RETURN(cydriver.cuStreamEndCapture(as_cu(self._h_stream), NULL))
+        self._h_graph.reset()
+        self._h_stream.reset()
+        self._state = CAPTURE_ENDED
```
leofang: `close()` unconditionally lands every kind of builder in `_state = CAPTURE_ENDED` with `_h_graph` reset — including FORKED, which never legitimately reaches "ended" via the normal `end_building` path. After `forked.close()` (or `join()`, which calls `close()` on the non-root builders), a subsequent `some_graph.update(forked_builder)` then passes the `_state != CAPTURE_ENDED` check in `Graph.update` (line ~870) and dereferences a null `_h_graph` into `cuGraphExecUpdate`. Two options:

1. Introduce a distinct `CLOSED` (or `INVALIDATED`) state value so `Graph.update` / `complete` / `debug_dot_print` can reject closed builders explicitly.
2. Or guard in `Graph.update`'s `GraphBuilder` branch on `source._h_graph` being non-null with a clearer error than the raw CUDA failure.

Related to leofang's guarding-state question on L250.

cuda_core/cuda/core/graph/_graph_builder.pyx
```diff
@@ -272,64 +295,72 @@ class GraphBuilder:
...
+                    c_stream, c_graph, NULL, NULL, 0, c_mode))
         else:
-            handle_return(driver.cuStreamBeginCapture(self._mnff.stream.handle, capture_mode))
+            with nogil:
+                HANDLE_RETURN(cydriver.cuStreamBeginCapture(c_stream, c_mode))
+                # The driver rejects NULL captureStatus_out, so we pass a
+                # stack-local even though begin_capture just succeeded and we
+                # only care about the resulting graph handle.
+                _get_capture_info(c_stream, &c_status, &c_graph)
+            self._h_graph = create_graph_handle(c_graph)
```
leofang, line 337: Partial-failure window: if `cuStreamBeginCapture` succeeds but the immediately-following `_get_capture_info` raises (or `create_graph_handle` throws under memory pressure), the stream is now mid-capture from the driver's POV but `self._state` is still `CAPTURE_NOT_STARTED`. `__dealloc__`'s guard `_state == CAPTURING and _kind != FORKED` then won't end the orphaned capture, and the stream stays poisoned for the rest of its lifetime.

This is new in this PR (the original `begin_building` was a single driver call, so no intermediate failure window existed). Two cheap fixes:

- Set `self._state = CAPTURING` *before* the `_get_capture_info` call, so `__dealloc__` will clean up.
- Or wrap the whole block in a `try:` that calls `cuStreamEndCapture` on failure.

cuda_core/cuda/core/graph/_graph_builder.pyx
```diff
@@ -764,7 +747,76 @@ class GraphBuilder:
...
+    gb._h_stream = stream._h_stream
+    gb._kind = FORKED
+    gb._state = CAPTURING
+    gb._stream = stream
+    return gb
+
+
+cdef inline GraphBuilder _init_conditional(Stream stream, cydriver.CUgraph cond_graph, GraphBuilder parent):
+    cdef GraphBuilder gb = GraphBuilder.__new__(GraphBuilder)
+    gb._h_graph = create_graph_handle_ref(cond_graph, parent._h_graph)
```
leofang, line 825: This is correct for the common case where `parent` is a PRIMARY builder whose `_h_graph` actually owns the top-level CUgraph. But `_init_forked` (line 750) explicitly leaves `_h_graph` empty for FORKED builders ("captures to primary graph"), so if a user calls `forked_builder.if_then(...)` (or `switch`/`while_loop`), `_cond_with_params` passes the FORKED's `_h_graph` (null) here as the parent reference.

The result is a CONDITIONAL_BODY builder whose `_h_graph` is a `create_graph_handle_ref` to the body sub-graph with no parent pinning — the actual owning handle (the PRIMARY's `_h_graph`) is never grabbed. So while the *physical* conditional node is wired into the primary's CUgraph, the *handle dependency chain* is broken.

Original code didn't track parents either, so this is a pre-existing semantic gap rather than a regression. But the new structure makes it *look* like the dependency is being tracked, which is more misleading than no tracking. Either (a) walk up to the real owning handle (somehow — there's currently no link from FORKED back to its PRIMARY), or (b) add a comment here documenting that this ref can be null when the parent is FORKED.

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -952,6 +953,28 @@ GraphHandle create_graph_handle_ref(CUgraph graph, const GraphHandle& h_parent)
...
 
+// ============================================================================
+// Graph Exec Handles
+// ============================================================================
+
+namespace {
+struct GraphExecBox {
+    CUgraphExec resource;
+};
+}  // namespace
```
leofang, line 1148: Design question, not a blocker: this is the only one of the graph-family boxes that holds no extra state — `GraphBox` tracks `h_parent`, `GraphNodeBox` tracks `h_graph`, but `GraphExecBox` has just the raw `CUgraphExec`. That's *semantically* correct today: after `cuGraphInstantiateWithParams`, the exec is independent of its source `CUgraph`, and `cuGraphExecUpdate` takes the new source at call time, so nothing needs to be held.

But the PR description calls this "groundwork for step 3 of #1330 (graph updates)". If that work ends up wanting to remember which `GraphBuilder` / `CUgraph` an exec was last updated from (for diagnostics, error messages, or to extend the source's lifetime through the exec), this Box is the natural place to grow a `GraphHandle h_source_graph;`. Worth either a one-line `// TODO:` noting that, or a confirmation that no such reference is planned and removing the asymmetry from any reviewer's mental model.

leofang (commented):
A few more observations beyond the inline thread already in flight — see the new inline comments. The two I'd actually want answers to before merge are (a) the FORKED-parent / conditional-body handle dependency on L762, and (b) the partial-failure window in `begin_building` on L322–328. The rest are minor.

-- Leo's bot

leofang:
During the offline review with Leo we found a gap in the test coverage, see below. No need to address them in this PR, but we should track it after the PR is merged.

The existing suite at `cuda_core/tests/graph/` covers two of the illegal state transitions (`test_graph_repeat_capture`, `test_graph_capture_errors`). The following scenarios — most introduced or behaviorally changed by this PR — are not exercised today:

1. **`begin_building()` called twice in a row.** Pre-PR this fell through to a cryptic CUDA error from `cuStreamBeginCapture`; post-PR the new `_state == CAPTURING` guard converts it to a clear `RuntimeError("Graph builder is already building.")`. This is the headline case for the "illegal transitions are detectable" claim in the PR description and is currently unverified.

2. **`complete()` on a FORKED builder after it has been `close()`d or `join()`ed.** `close()` unconditionally sets `_state = CAPTURE_ENDED` and resets `_h_graph`, so `complete()` passes the state guard and dereferences a null graph handle (see inline comment on L269).

3. **`if_then` / `switch` / `while_loop` invoked on a FORKED builder.** `_init_conditional` stores `create_graph_handle_ref(cond_graph, parent._h_graph)`, but `parent._h_graph` is null for FORKED, so the resulting CONDITIONAL_BODY's keep-alive chain to the actual owning primary handle is broken (see inline comment on L762).

4. **GC of a builder mid-capture (no `close()` / `end_building()` called).** This is the path that exercises `cuStreamEndCapture(stream, NULL)` (the L250 concern about the undocumented NULL `phGraph` behavior) and the partial-failure window in `begin_building` between `cuStreamBeginCapture` and `_get_capture_info` (inline comment on L328). `__dealloc__` swallows all driver return codes, so a regression here would be silent.

5. **`Graph.update(source)` after `source.close()`.** Same root cause as #2 — `_state` is `CAPTURE_ENDED` post-close, the type-check passes, and `cuGraphExecUpdate` is called with a null `_h_graph`.

6. **Nested conditionals (`if_then` inside a CONDITIONAL_BODY body).** This is the only path that exercises the multi-level `create_graph_handle_ref` chain introduced by this PR (nested-body → outer-body → primary). Expected to work, but untested.

7. **`embed(non_GraphBuilder)`.** The new `def embed(self, GraphBuilder child)` Cython signature raises `TypeError` immediately on a non-GraphBuilder argument; pre-PR this `AttributeError`ed on the first attribute access. Minor error-class change worth pinning down.

-- Leo's bot

cuda_core/cuda/core/graph/_graph_builder.pyx
Andy-Jost: 1. Added a `CLOSED` state and guards to check for it.
2. The stream is now held by a `StreamHandle` (`shared_ptr`) and the reference is dropped on close.
3. Yes. I confirmed that the driver performs the necessary `NULL` checks. This is undocumented, though.

cuda_core/cuda/core/graph/_graph_builder.pyx
Andy-Jost: Fixed

cuda_core/cuda/core/graph/_graph_builder.pyx
Andy-Jost: fixed

cuda_core/cuda/core/graph/_graph_builder.pyx
Andy-Jost: fixed

cuda_core/cuda/core/graph/_graph_builder.pyx
Andy-Jost: Good catch; added the CLOSED state and related guards.

cuda_core/cuda/core/graph/_graph_builder.pyx
Andy-Jost: Fixed — set `_state = CAPTURING` immediately after begin-capture succeeds, before `_get_capture_info`/`create_graph_handle`, so `dealloc`/`close` can unwind the capture on failure.

cuda_core/cuda/core/graph/_graph_builder.pyx
Andy-Jost: fixed: forked builders now carry the primary's `GraphHandle` via `GB_init_forked(stream, self._h_graph)`, so conditional bodies get a valid parent ref in `create_graph_handle_ref`. Test added

cuda_core/cuda/core/_cpp/resource_handles.cpp
Andy-Jost: Agreed: no source-graph ref on `GraphExecBox` is planned. The exec is independent post-instantiate and `Graph.update()` supplies the source at call time. Clarification: step 3 lays the foundation for source graph updates; exec graph updates are step 5.

Andy-Jost:
> During the offline review with Leo...

All seven gaps called out here are now covered in this PR (see `cuda_core/tests/graph/test_graph_builder.py` and `cuda_core/tests/graph/test_graph_builder_conditional.py`).

For #3, `test_graph_conditional_on_forked_builder` exercises `if_then` on a forked builder. `switch`/`while_loop` share the same `_cond_with_params` path, so I didn't duplicate them.

| Identified gap | Test added |
| --- | --- |
| 1. `begin_building()` twice | `test_graph_begin_building_twice` |
| 2. `complete()` after forked `close()`/`join()` | `test_graph_complete_after_close_forked` |
| 3. Conditionals on FORKED builder | `test_graph_conditional_on_forked_builder` (`if_then` only) |
| 4. GC mid-capture | `test_graph_gc_mid_capture` |
| 5. `Graph.update()` after `source.close()` | `test_graph_update_after_source_close` |
| 6. Nested conditionals | `test_graph_conditional_nested` |
| 7. `embed(non_GraphBuilder)` | `test_graph_embed_non_builder` |

cuda_core/cuda/core/graph/_graph_builder.pyi
```diff
@@ -335,15 +330,7 @@ class GraphBuilder:
 
         """
 
-    def close(self) -> None:
```
mdboom, line 338: Drive-by comment to test the theory that the `.pyi` files help us see API changes.

I think dropping `close` here is API breakage -- if a user is calling it directly, they will no longer be able to.  If that isn't resolvable to make everything in this PR work, then we need a "breaking changes" changelog entry and a deprecation decorator etc.

mdboom: Actually I stand corrected, I see this just got moved a few lines above.

Andy-Jost: Ah, good. I was worried for a moment!

rwgk:
The test failures here are almost certainly due to the new cupti dll name. Working on it.

rwgk:
> The test failures here are almost certainly due to the new cupti dll name. Working on it.

Interestingly, this was updated already two hours ago:

* https://pypi.org/project/nvidia-cuda-cupti/13.3.75/

I was monitoring https://developer.download.nvidia.com/compute/cuda/redist/redistrib_13.3.1.json, it still doesn't exist.

Next time I'll monitor nvidia-cuda-cupti instead.

rwgk:
@Andy-Jost this should fix the CI failures here:

https://github.com/NVIDIA/cuda-python/pull/2276

rwgk (commented):
@mdboom see the second (P3) finding.

@Andy-Jost is the first (P2) finding actionable?

___

gpt-5.5

# PR 2008 Refactor-Focused Review Findings

- P2: `Device.create_graph_builder()` still creates a private stream, but `GraphBuilder.close()` now only drops its own shared handle/reference. If user code keeps `s = gb.stream`, `gb.close()` no longer closes that owned stream, unlike the old `is_stream_owner` path. Either preserve the ownership close behavior or document/test the new shared-lifetime semantics. See `cuda_core/cuda/core/_device.pyx:1457` and `cuda_core/cuda/core/graph/_graph_builder.pyx:267`.

- P3: Public stub annotations regressed for `GraphBuilder.close()` and `GraphBuilder.embed()`; both now infer `Any` return in the `.pyi` despite being public APIs that return `None`. See `cuda_core/cuda/core/graph/_graph_builder.pyi:121` and `cuda_core/cuda/core/graph/_graph_builder.pyi:333`.

Andy-Jost:
I approved #2276.

Regarding P2, I think it's fine as-is. In the new ownership model, `close` means to drop a shared reference and that's not a problem.

rwgk (approved):

PR #1977, opened 2026-04-25
WIP: pathfinder_compatibility_guard_rails

rwgk:
Resolves #1038

Continuation of #1936

WIP — CI testing

PR #2012, opened 2026-05-04
pathfinder: add toolkit-info parser and display-driver release helpers

rwgk:
## Summary

Extracted from #1977 (xref #1038) so the underlying `cuda.pathfinder._utils` plumbing can be reviewed independently of the larger compatibility guard-rails work that consumes it. Five commits, scoped strictly to `cuda_pathfinder/cuda/pathfinder/_utils/{driver_info,toolkit_info}.py` and their unit tests; nothing else is touched.

## Changes

- **toolkit-info parser.** Add `cuda.pathfinder._utils.toolkit_info`, a small utility that reads the `CUDA_VERSION` macro from `cuda.h` so callers can infer CTK major.minor from toolkit headers without depending on `version.json`.

- **Shared encoded-CUDA-version decoding.** Centralize encoded CUDA version parsing and validation so toolkit and driver version helpers stay aligned and `cuda.h` parsing gets consistent string conversion and error reporting.

- **Display-driver release metadata.** Add `DriverReleaseVersion` and `query_driver_release_version()` in `cuda.pathfinder._utils.driver_info`. The helper queries `nvmlSystemGetDriverVersion()` via `ctypes` against the NVML library loaded through pathfinder's existing dynamic-lib loader, and parses the result into a frozen dataclass exposing both the raw text (e.g. `"595.58.03"`) and the branch number (e.g. `595`) used by NVIDIA's published minor-version compatibility tables.

- **`try`/`finally` around NVML init/shutdown.** The matched `nvmlInit_v2()` / `nvmlShutdown()` pair runs under a single `try`/`finally`, with `sys.exc_info()[1]` preserving the asymmetric error precedence: when both the body and shutdown fail, the body's error wins (Python keeps the shutdown error on `__context__` for debugging); when only shutdown fails, the shutdown error surfaces. Comments document that NVML's init/shutdown is reference-counted, so this balanced pair is safe even when the caller has already initialized NVML elsewhere in the process.

- **Tests.** Direct coverage of the dataclass parsers, the cache-clear lifecycle, the `QueryDriverReleaseVersionError` wrapping, the fake-NVML end-to-end path, and the full success/failure cleanup matrix (body-ok+shutdown-fail, body-fail+shutdown-ok, body-fail+shutdown-fail).

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_utils/driver_info.py
+++ b/cuda_pathfinder/cuda/pathfinder/_utils/driver_info.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import ctypes
 import functools
+import re
 from collections.abc import Callable
 from dataclasses import dataclass
 from typing import cast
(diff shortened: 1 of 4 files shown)


rwgk:
Closing this PR (and PR #1977) so I can rebuild the work as a native stacked PR series using GitHub's newly announced stacked pull requests feature, which is currently rolling out in public preview:

* https://github.blog/changelog/2026-07-30-stacked-pull-requests-are-now-in-public-preview/

This indicates that the new feature is already available in this repo (the list is empty because there are no stacks yet):

```
$ gh api repos/NVIDIA/cuda-python/stacks
[]
```

I’ll link the replacement PRs here once they're available.

rwgk:
The corresponding new PR is #2505

PR #2013, opened 2026-05-04
[no-ci] toolshed: modernize `conda_create_for_pathfinder_testing` scripts

rwgk:
Modernizations as used for manually testing changes under PR #1977 (xref issue #1038) with Conda Linux and Windows:

Test results: https://github.com/NVIDIA/cuda-python/pull/1977#issuecomment-4332733658

This is a separate PR to remove a distraction from the bigger set of changes.

## What this PR modernizes

The two scripts had drifted from their de-facto source of truth — the Linux- and Windows-installable subsets of `cuda_pathfinder/pyproject.toml`. This PR brings them back in sync and tightens a few rough edges:

- **Track the current pyproject.toml package lists.** The previous scripts predate several additions, removals, and renamings in the pathfinder dependency groups; the package arrays in both shells now mirror the platform-scoped subsets that `pyproject.toml` declares.
- **The Windows list is properly platform-scoped instead of a copied-from-Linux superset.** The Windows script previously ran with the Linux list and silently swallowed `conda install` errors for packages that don't have win-64 builds. The new list contains only the win-64-installable subset, so a non-zero exit now means something is actually wrong.
- **Caller-chosen Python version.** Both scripts now take `python_major_minor` as an explicit argument instead of hard-coding it. The trigger was that `python=3.14` with `cuda-toolkit=12.9.x` had stopped resolving on Windows, so the fixed mapping was no longer adequate.
- **Harden the script invariants.** `set -euo pipefail` (bash) and `Set-StrictMode -Version Latest` plus `$ErrorActionPreference = "Stop"` (PowerShell) so unset variables and non-zero exits surface immediately instead of silently producing partial environments.
- **Single-solve `conda install`** (per separate review feedback on this PR). Both scripts now hand the full package list to one `conda install` invocation, replacing the prior per-package loop. Faster, and more importantly: each independent solve in the old loop could downgrade or upgrade shared deps placed by a previous solve, leaving the final environment in an order-dependent state.

## On `toolshed/` in general

`toolshed/` is intentionally a shared scratchpad for scripts that are useful intermittently — environment bootstrap helpers, ad-hoc reproducers, glue around manual workflows — but don't warrant CI coverage, unit tests, or the rest of the production-code apparatus. It is accepted that entries bitrot between uses; having them in a shared, versioned location just means the next person who needs the same workflow has a better starting point than starting from scratch.

That framing is also why these helpers stay as two thin shell scripts rather than getting consolidated into a Python module: for the current cadence of use, parallel shell idioms in bash and PowerShell are a smaller maintenance surface than a third script (Python) that would in turn need its own subprocess handling, error model, and entry-point story. Worth revisiting if/when the scripts grow further.


toolshed/conda_create_for_pathfinder_testing.ps1
```diff
@@ -7,22 +7,30 @@ param(
 )
 
 $ErrorActionPreference = "Stop"
+Set-StrictMode -Version Latest
```
rparolin, line 12: It would be good to spell out more what "modernization's" you are doing.  I'm not sure what you are modernizing based on the code.

toolshed/conda_create_for_pathfinder_testing.sh
```diff
@@ -1,28 +1,65 @@
...
 # SPDX-License-Identifier: Apache-2.0
 
+set -euo pipefail
+
 if [[ $# -ne 1 ]]; then
     echo "Usage: $(basename "$0") ctk-major-minor-patch" 1>&2
     exit 1
 fi
 
+cuda_version="$1"
```
rparolin: Consider: Refactoring these two scripts into a single python script to reduce maintenance headaches.

toolshed/conda_create_for_pathfinder_testing.ps1
rparolin: Consider: Using batch install rather then running the resolve N times.

"This runs the conda solver N times instead of once, which is both much slower and can produce a different (worse) environment than solving all packages together — each install can independently downgrade/upgrade shared deps. A single conda install -y -c conda-forge "${cpkgs[@]}" is faster and more consistent."

rparolin (approved):

toolshed/conda_create_for_pathfinder_testing.ps1
rwgk: Done: See the massively expanded PR description.

toolshed/conda_create_for_pathfinder_testing.sh
rwgk: See last paragraph in the new expanded PR description.

I more-or-less went into the opposite direction: this feedback prompted me to simplify the scripts, by removing the "derive Python version from CUDA version" logic and simply requiring the Python version to be passed in.

toolshed/conda_create_for_pathfinder_testing.ps1
rwgk: Done in commit 76d2524

rwgk:
Re-tested locally after the consolidating changes on this PR.

Linux:

```
toolshed/conda_create_for_pathfinder_testing.sh 3.12 12.9.2
toolshed/conda_create_for_pathfinder_testing.sh 3.13 13.2.1
```

Windows (PowerShell):

```
toolshed\conda_create_for_pathfinder_testing.ps1 3.12 12.9.2
toolshed\conda_create_for_pathfinder_testing.ps1 3.14 13.2.1
```

All four runs completed cleanly: the explicit `python_major_minor` argument was honored, the platform-scoped package list installed in a single `conda install` solve, and CUDA-specific resolutions picked the expected cu12/cu13 builds.

The asymmetric Python pick — 3.13 on Linux vs. 3.14 on Windows for the CUDA 13 run — is deliberate: 3.14 paired with `cuda-toolkit=12.9.x` is the combination that originally stopped resolving on Windows and motivated taking the Python version as an explicit argument, so this exercises that the newer Python is now reachable when paired with a matching CUDA toolkit.

Issue #1995, opened 2026-04-30
cuda.core: Replace string literals with enums in public API


mdboom:
> This is just a result of a sweep for using strings where enums could be used. This is just to aid the discussion about whether they should be addressed.

Several places in the `cuda.core` public API accept or return string values chosen from a small, fixed vocabulary. These are candidates for replacement with proper Python `enum` types, which would improve discoverability, IDE autocompletion, `isinstance` checking, and catch typos at the call site rather than at runtime.

---

### Compilation / Linking pipeline

| Location | Kind | Values |
|---|---|---|
| `Program.__init__(code_type)` | arg | `"c++"`, `"ptx"`, `"nvvm"` |
| `Program.compile(target_type)` | arg | `"ptx"`, `"cubin"`, `"ltoir"` |
| `Program.backend` | return | `"NVRTC"`, `"NVVM"`, `"nvJitLink"`, `"driver"` |
| `Program.pch_status` | return | `"created"`, `"not_attempted"`, `"failed"` |
| `Linker.link(target_type)` | arg | `"cubin"`, `"ptx"` |
| `Linker.backend` | return | `"nvJitLink"`, `"driver"` |
| `ObjectCode.code_type` | return | `"cubin"`, `"ptx"`, `"ltoir"`, `"fatbin"`, `"object"`, `"library"` |

### Memory

| Location | Kind | Values |
|---|---|---|
| `ManagedMemoryResourceOptions.preferred_location_type` | arg | `"device"`, `"host"`, `"host_numa"` |
| `ManagedMemoryResource.preferred_location` | return (tuple[0]) | `"device"`, `"host"`, `"host_numa"` |
| `VirtualMemoryResourceOptions.allocation_type` | arg | `"pinned"`, `"managed"` |
| `VirtualMemoryResourceOptions.location_type` | arg | `"device"`, `"host"`, `"host_numa"`, `"host_numa_current"` |
| `VirtualMemoryResourceOptions.handle_type` | arg | `"posix_fd"`, `"generic"`, `"win32_kmt"`, `"fabric"` |
| `VirtualMemoryResourceOptions.granularity` | arg | `"recommended"`, `"minimum"` |
| `VirtualMemoryResourceOptions.self_access` / `peer_access` | arg | `"rw"`, `"r"` |

[...]

Issue #1919, opened 2026-04-15
Audit cuda.core API for 1.0 release


rparolin:
We want to audit the cuda.core API before we release 1.0 to highlight anything that we want to change to prior to release. Once cuda.core-1.0.0 release, we will be locked into a few years support cycle so if there is any APIs that we want to break we must be do it now.

@mdboom plans to do this as a series of prompts to find issues and tackle from there.

## Unintended exposure

In the `cuda_core` package, please analyse all of the `__init__.py` files and find any symbols that are exposed unintendedly.  Especially symbols defined outside of the `cuda_core` package, private members or implementation details.

- #1931

leofang:
@mdboom I think this can be closed now?

Issue #1715, opened 2026-03-04
[BUG]: cuda.core.system.Device unavailable with cuda-bindings 13.1.1 — requires unreleased 13.1.2


Pedrexus:
Hello all, thanks for the nice library!

After installing `cuda-core==0.6.0` + `cuda-bindings==13.1.1` (latest on PyPI, conda-forge, and the NVIDIA conda channel), `CUDA_BINDINGS_NVML_IS_COMPATIBLE` is always `False` and `Device` is never exported from `cuda.core.system`.

Tracing through the source: `_system.pyx` in the 0.6.0 tag checks `_BINDINGS_VERSION >= (13, 1, 2)`, but `cuda-bindings 13.1.2` was never published — PR #1404 bumped the source version to unblock `cuda-core` development, and issue #1521 (the actual release) was closed as `not_planned`. So every user of `cuda-core==0.6.0` gets `Device` silently unavailable.

Additionally, five symbols are missing from `_nvml` 13.1.1 that `_device.pyx` / `_fan.pxi` / `_event.pxi` / `_system_events.pyx` alias at import time: `DeviceArch`, `FieldId`, `ClocksEventReasons`, `EventType`, `FanControlPolicy`, `SystemEventType`.

**Workaround** (confirmed working): create `cuda/bindings/nvml.py` re-exporting `_nvml` with the six missing stubs, and patch `_version.py` to report `13.1.2`:

```python
# cuda/bindings/_version.py
__version__ = "13.1.2"
__version_tuple__ = (13, 1, 2)

# cuda/bindings/nvml.py
import enum
from cuda.bindings._nvml import *

class DeviceArch(enum.IntEnum):
    DEVICE_ARCH_AMPERE = 7; DEVICE_ARCH_ADA = 8; DEVICE_ARCH_HOPPER = 9; ...

class FieldId(enum.IntEnum): pass
class ClocksEventReasons(enum.IntFlag): GpuIdle = 0x01; ...
class EventType(enum.IntFlag): SingleBitEccError = 0x1; ...
class SystemEventType(enum.IntFlag): GpuDriverUnbind = 0x1; ...
class FanControlPolicy(enum.IntEnum): TEMPERATURE_CONTINOUS_SW = 0; MANUAL = 1
```

After that, `Device(index=0).name`, `.memory_info`, `.temperature.sensor()` etc. all work correctly against 13.1.1.

[...]

PR #2014, opened 2026-05-04
Re-wrap most of the enums in cuda.bindings.nvml for cuda.core.system.

mdboom:
This rewraps most of the enums (except the extremely large and unorganized `FieldId`) for `cuda.core.system` rather than passing them directly through.  This also creates an optional string interface for all of these enum values.

### DIFF
--- a/cuda_core/cuda/core/system/_clock.pxi
+++ b/cuda_core/cuda/core/system/_clock.pxi
@@ -16,7 +16,7 @@ Current actual clock value.
 ClockId.CUSTOMER_BOOST_MAX.__doc__ = """
 OEM-defined maximum clock rate
 """
-cdef dict _CLOCK_ID_MAPPING = {
+_CLOCK_ID_MAPPING = {
     ClockId.CURRENT: nvml.ClockId.CURRENT,
     ClockId.CUSTOMER_BOOST_MAX: nvml.ClockId.CUSTOMER_BOOST_MAX,
 }
@@ -36,7 +36,7 @@ class ClocksEventReasons(StrEnum):
     HW_THERMAL_SLOWDOWN = "hw_thermal_slowdown"
     HW_POWER_BRAKE_SLOWDOWN = "hw_power_brake_slowdown"
     DISPLAY_CLOCK_SETTING = "display_clock_setting"
-cdef dict _CLOCKS_EVENT_REASONS_MAPPING = {
+_CLOCKS_EVENT_REASONS_MAPPING = {
     nvml.ClocksEventReasons.EVENT_REASON_NONE: ClocksEventReasons.NONE,
     nvml.ClocksEventReasons.EVENT_REASON_GPU_IDLE: ClocksEventReasons.GPU_IDLE,
     nvml.ClocksEventReasons.EVENT_REASON_APPLICATIONS_CLOCKS_SETTING: ClocksEventReasons.APPLICATIONS_CLOCKS_SETTING,
@@ -58,7 +58,7 @@ class ClockType(StrEnum):
     SM = "sm"
     MEMORY = "memory"
     VIDEO = "video"
-cdef dict _CLOCK_TYPE_MAPPING = {
+_CLOCK_TYPE_MAPPING = {
     ClockType.GRAPHICS: nvml.ClockType.CLOCK_GRAPHICS,
     ClockType.SM: nvml.ClockType.CLOCK_SM,
     ClockType.MEMORY: nvml.ClockType.CLOCK_MEM,
--- a/cuda_core/cuda/core/system/_clock.pxi
+++ b/cuda_core/cuda/core/system/_clock.pxi
@@ -3,9 +3,67 @@
 # SPDX-License-Identifier: Apache-2.0
 
 
-ClockId = nvml.ClockId
-ClocksEventReasons = nvml.ClocksEventReasons
-ClockType = nvml.ClockType
+class ClockId(StrEnum):
+    """
+    Clock Ids. These are used in combination with :class:`ClockType` to specify a single clock value.
+    """
+    CURRENT = "current"
+    CUSTOMER_BOOST_MAX = "customer_boost_max"
+    # APP_CLOCK_TARGET and APP_CLOCK_DEFAULT are deprecated so not included here
+ClockId.CURRENT.__doc__ = """
+Current actual clock value.
+"""
+ClockId.CUSTOMER_BOOST_MAX.__doc__ = """
+OEM-defined maximum clock rate
+"""
+cdef dict _CLOCK_ID_MAPPING = {
+    ClockId.CURRENT: nvml.ClockId.CURRENT,
+    ClockId.CUSTOMER_BOOST_MAX: nvml.ClockId.CUSTOMER_BOOST_MAX,
+}
+
+
+class ClocksEventReasons(StrEnum):
+    """
+    Reasons for a clocks event.  These are used in combination with :class:`ClockType` to specify the reason for a clocks event.
+    """
+    NONE = "none"
+    GPU_IDLE = "gpu_idle"
+    APPLICATIONS_CLOCKS_SETTING = "applications_clocks_setting"
+    SW_POWER_CAP = "sw_power_cap"
+    HW_SLOWDOWN = "hw_slowdown"
+    SYNC_BOOST = "sync_boost"
+    SW_THERMAL_SLOWDOWN = "sw_thermal_slowdown"
+    HW_THERMAL_SLOWDOWN = "hw_thermal_slowdown"
+    HW_POWER_BRAKE_SLOWDOWN = "hw_power_brake_slowdown"
+    DISPLAY_CLOCK_SETTING = "display_clock_setting"
+cdef dict _CLOCKS_EVENT_REASONS_MAPPING = {
+    nvml.ClocksEventReasons.EVENT_REASON_NONE: ClocksEventReasons.NONE,
+    nvml.ClocksEventReasons.EVENT_REASON_GPU_IDLE: ClocksEventReasons.GPU_IDLE,
+    nvml.ClocksEventReasons.EVENT_REASON_APPLICATIONS_CLOCKS_SETTING: ClocksEventReasons.APPLICATIONS_CLOCKS_SETTING,
+    nvml.ClocksEventReasons.EVENT_REASON_SW_POWER_CAP: ClocksEventReasons.SW_POWER_CAP,
+    nvml.ClocksEventReasons.THROTTLE_REASON_HW_SLOWDOWN: ClocksEventReasons.HW_SLOWDOWN,
+    nvml.ClocksEventReasons.EVENT_REASON_SYNC_BOOST: ClocksEventReasons.SYNC_BOOST,
+    nvml.ClocksEventReasons.EVENT_REASON_SW_THERMAL_SLOWDOWN: ClocksEventReasons.SW_THERMAL_SLOWDOWN,
+    nvml.ClocksEventReasons.THROTTLE_REASON_HW_THERMAL_SLOWDOWN: ClocksEventReasons.HW_THERMAL_SLOWDOWN,
+    nvml.ClocksEventReasons.THROTTLE_REASON_HW_POWER_BRAKE_SLOWDOWN: ClocksEventReasons.HW_POWER_BRAKE_SLOWDOWN,
+    nvml.ClocksEventReasons.EVENT_REASON_DISPLAY_CLOCK_SETTING: ClocksEventReasons.DISPLAY_CLOCK_SETTING,
+}
+
+
+class ClockType(StrEnum):
+    """
+    Clock types. All speeds are in Mhz.
+    """
+    GRAPHICS = "graphics"
+    SM = "sm"
+    MEMORY = "memory"
+    VIDEO = "video"
+cdef dict _CLOCK_TYPE_MAPPING = {
+    ClockType.GRAPHICS: nvml.ClockType.CLOCK_GRAPHICS,
+    ClockType.SM: nvml.ClockType.CLOCK_SM,
+    ClockType.MEMORY: nvml.ClockType.CLOCK_MEM,
+    ClockType.VIDEO: nvml.ClockType.CLOCK_VIDEO,
+}
 
 
 cdef class ClockOffsets:
@@ -48,11 +106,18 @@ cdef class ClockInfo:
     cdef intptr_t _handle
     cdef int _clock_type
 
-    def __init__(self, handle, clock_type: ClockType):
+    def __init__(self, handle, clock_type: ClockType | str):
         self._handle = handle
+        try:
+            clock_type = _CLOCK_TYPE_MAPPING[clock_type]
+        except KeyError:
+            raise ValueError(
+                f"Invalid clock type: {clock_type}. "
+                f"Must be one of {list(ClockType.__members__.values())}"
+            ) from None
         self._clock_type = int(clock_type)
 
-    def get_current_mhz(self, clock_id: ClockId = ClockId.CURRENT) -> int:
+    def get_current_mhz(self, clock_id: ClockId | str = ClockId.CURRENT) -> int:
         """
         Get the current clock speed of a specific clock domain, in MHz.
 
@@ -60,14 +125,21 @@ cdef class ClockInfo:
 
         Parameters
         ----------
-        clock_id: :class:`ClockId`
-            The clock ID to query.
+        clock_id: :class:`ClockId` | str
+            The clock ID to query.  Defaults to the current clock value.
 
         Returns
         -------
         int
             The clock speed in MHz.
         """
+        try:
+            clock_id = _CLOCK_ID_MAPPING[clock_id]
+        except KeyError:
+            raise ValueError(
+                f"Invalid clock ID: {clock_id}. "
+                f"Must be one of {list(ClockId.__members__.values())}"
+            ) from None
         return nvml.device_get_clock(self._handle, self._clock_type, clock_id)
 
     def get_max_mhz(self) -> int:
@@ -99,24 +171,26 @@ cdef class ClockInfo:
         """
         return nvml.device_get_max_customer_boost_clock(self._handle, self._clock_type)
 
-    def get_min_max_clock_of_pstate_mhz(self, pstate: Pstates) -> tuple[int, int]:
+    def get_min_max_clock_of_pstate_mhz(self, pstate: int) -> tuple[int, int]:
         """
         Get the minimum and maximum clock speeds for this clock domain
         at a given performance state (Pstate), in MHz.
 
         Parameters
         ----------
-        pstate: :class:`Pstates`
-            The performance state to query.
+        pstate: int
+            The performance state to query.  Must be an int between 0 and 15,
+            where 0 is the highest performance state (P0) and 15 is the lowest
+            (P15).
 
         Returns
         -------
         tuple[int, int]
             A tuple containing the minimum and maximum clock speeds in MHz.
         """
-        return nvml.device_get_min_max_clock_of_p_state(self._handle, self._clock_type, pstate)
+        return nvml.device_get_min_max_clock_of_p_state(self._handle, self._clock_type, _pstate_to_enum(pstate))
 
-    def get_offsets(self, pstate: Pstates) -> ClockOffsets:
+    def get_offsets(self, pstate: int) -> ClockOffsets:
         """
         Retrieve min, max and current clock offset of some clock domain for a given Pstate.
 
@@ -124,12 +198,14 @@ cdef class ClockInfo:
 
         Parameters
         ----------
-        pstate: :class:`Pstates`
-            The performance state to query.
+        pstate: int
+            The performance state to query.  Must be an int between 0 and 15,
+            where 0 is the highest performance state (P0) and 15 is the lowest
+            (P15).
 
         Returns
         -------
         :obj:`~_device.ClockOffsets`
             An object with the min, max and current clock offset.
         """
-        return ClockOffsets(nvml.device_get_clock_offsets(self._handle, self._clock_type, pstate))
+        return ClockOffsets(nvml.device_get_clock_offsets(self._handle, self._clock_type, _pstate_to_enum(pstate)))
--- a/cuda_core/cuda/core/system/_clock.pxi
+++ b/cuda_core/cuda/core/system/_clock.pxi
@@ -10,12 +10,12 @@ class ClockId(StrEnum):
     CURRENT = "current"
     CUSTOMER_BOOST_MAX = "customer_boost_max"
     # APP_CLOCK_TARGET and APP_CLOCK_DEFAULT are deprecated so not included here
-ClockId.CURRENT.__doc__ = """
-Current actual clock value.
-"""
-ClockId.CUSTOMER_BOOST_MAX.__doc__ = """
-OEM-defined maximum clock rate
-"""
+
+
+ClockId.CURRENT.__doc__ = "Current actual clock value."
+ClockId.CUSTOMER_BOOST_MAX.__doc__ = "OEM-defined maximum clock rate"
+
+
 _CLOCK_ID_MAPPING = {
     ClockId.CURRENT: nvml.ClockId.CURRENT,
     ClockId.CUSTOMER_BOOST_MAX: nvml.ClockId.CUSTOMER_BOOST_MAX,
@@ -36,6 +36,8 @@ class ClocksEventReasons(StrEnum):
     HW_THERMAL_SLOWDOWN = "hw_thermal_slowdown"
     HW_POWER_BRAKE_SLOWDOWN = "hw_power_brake_slowdown"
     DISPLAY_CLOCK_SETTING = "display_clock_setting"
+
+
 _CLOCKS_EVENT_REASONS_MAPPING = {
     nvml.ClocksEventReasons.EVENT_REASON_NONE: ClocksEventReasons.NONE,
     nvml.ClocksEventReasons.EVENT_REASON_GPU_IDLE: ClocksEventReasons.GPU_IDLE,
@@ -58,6 +60,8 @@ class ClockType(StrEnum):
     SM = "sm"
     MEMORY = "memory"
     VIDEO = "video"
+
+
 _CLOCK_TYPE_MAPPING = {
     ClockType.GRAPHICS: nvml.ClockType.CLOCK_GRAPHICS,
     ClockType.SM: nvml.ClockType.CLOCK_SM,
--- a/cuda_core/cuda/core/system/_device.pyx
+++ b/cuda_core/cuda/core/system/_device.pyx
@@ -5,22 +5,27 @@
 from libc.stdint cimport intptr_t, uint64_t
 from libc.math cimport ceil
 
+from enum import StrEnum
 from multiprocessing import cpu_count
 from typing import Iterable
+import warnings
 
 from cuda.bindings import nvml
+from cuda.bindings._internal._fast_enum import FastEnum
 
 from ._nvml_context cimport initialize
 
 
-AddressingMode = nvml.DeviceAddressingModeType
-AffinityScope = nvml.AffinityScope
-BrandType = nvml.BrandType
-DeviceArch = nvml.DeviceArch
-GpuP2PCapsIndex = nvml.GpuP2PCapsIndex
-GpuP2PStatus = nvml.GpuP2PStatus
-GpuTopologyLevel = nvml.GpuTopologyLevel
-Pstates = nvml.Pstates
+cdef object _pstate_to_int(object pstate):
+    if pstate == nvml.Pstates.PSTATE_UNKNOWN:
+        return None
+    return int(pstate) - int(nvml.Pstates.PSTATE_0)
+
+
+cdef int _pstate_to_enum(int pstate):
+    if pstate < 0 or pstate > 15:
+        raise ValueError(f"Invalid P-state: {pstate}. Must be between 0 and 15 inclusive.")
+    return int(pstate) + int(nvml.Pstates.PSTATE_0)
 
 
 include "_clock.pxi"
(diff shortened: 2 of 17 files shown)


rwgk:
# PR 2014 Agent Review

- Size: 16 files, +800 / -191
- Reviewer: Claude Opus 4.7 (1M context, Max thinking) — initial review

With small manual edits: I reduced the **Compatibility** subsection. I deleted the Next steps section entirely. The rest looks useful to me.

The "Sync guard (structural gap)" finding wasn't in the initial automatic review; I asked about it in a follow-on prompt. It seems pretty easy to at least add basic protections.

___


## High-level summary (asked for by reviewer)

This PR is the implementation of issue
[#1995](https://github.com/NVIDIA/cuda-python/issues/1995) ("Replace string
literals with enums in public API"), which is itself a sub-item of the
[#1919](https://github.com/NVIDIA/cuda-python/issues/1919) "Audit `cuda.core`
API for 1.0 release" epic. The motivation, captured in the issue thread
(especially the 2026-05-01 comment), is the `cuda.core` 1.0 milestone (#43, due
2026-05-07): once 1.0 ships, the team is locked into a multi-year support
window, so any awkward-to-fix public API surface must be cleaned up now. The
decision was to:

- For "logically enum-like" things, accept either an enum or a string
  everywhere, and return a string-flavored enum (`StrEnum`).
- For NVML-derived enums (currently `_FastEnum` ints), wrap them in
  `cuda.core.system` with hand-curated `StrEnum`s so users see a uniform
  `Enum | str` interface and don't get C-style names like
  `EVENT_TYPE_XID_CRITICAL_ERROR` or `TEMPERATURE_THRESHOLD_SHUTDOWN` leaking
  into the public API.

The PR rewraps most NVML enums (except `FieldId`, intentionally) into
`StrEnum` types, adds `_<NAME>_MAPPING` / `_<NAME>_INV_MAPPING` dicts for
round-tripping, accepts `Enum | str` everywhere, and cleans up a few adjacent
things while the file is open:

- `Pstates` → plain `int` (0..15) with `None` for unknown.
- `PciInfo.get_throughput(counter)` → `rx_throughput` / `tx_throughput`
  properties.
- `device.brand` → free-form `str` (with `"Unknown"` fallback).
- NvLink version → `(major, minor)` tuple.

It also depends on `backports.strenum` for Python 3.10 (`cuda_core` supports
3.10+).

The direction is right and consistent with the design agreed in #1995.
However, the PR has several concrete bugs, a doc-build risk, a structural gap
(no mechanism to keep the wrappers in sync with the bindings — see "Sync
guard" below), and a few breaking-change footnotes that are not currently
called out in the PR body.

## Findings (most severe first)

### Bugs

**Bug — `test_nvlink` references a removed symbol.**
`cuda_core/tests/system/test_system_device.py:759` still does
`assert isinstance(version, system.NvlinkVersion)`. `NvlinkVersion` is no
longer in the `system` namespace (the `__all__` change in `_device.pyx`
dropped it; `_nvlink.pxi` now returns a plain `tuple[int, int]`). This test
will fail with `AttributeError: module 'cuda.core.system' has no attribute
'NvlinkVersion'` on any host with NVLink. Replace with
`assert isinstance(version, tuple)` plus length / element checks.

**Bug — `supported_pstates` can iterate beyond the valid range.**
`cuda_core/cuda/core/system/_device.pyx:1001` walks
`nvml.device_get_supported_performance_states(...)` and only filters out
`PSTATE_UNKNOWN` (= 32). The NVML header says unused trailing slots are
`PSTATE_UNKNOWN`, but if the driver ever returned a value outside 0..15
(e.g., a future `PSTATE_*`), `_pstate_to_int` would silently return
`int(x) - 0`, producing an out-of-contract integer. Either drop the value or
raise. The same risk applies to `device.performance_state`
(`cuda_core/cuda/core/system/_device.pyx:971`).

**Bug — `_pstate_to_enum` name is wrong; takes an `int` and returns an `int`.**
`cuda_core/cuda/core/system/_device.pyx:29` is named `_pstate_to_enum` but
the body is
`return int(pstate) + int(nvml.Pstates.PSTATE_0)`. It just shifts an int, and
since `PSTATE_0 = 0` it's literally the identity for valid input. The
misnomer is confusing; rename to `_int_to_pstate` (or inline). The cast back
through `int(...)` is also redundant.

**Bug — error messages report bit index, not the failing bit value.**
In `current_clock_event_reasons` / `supported_clock_event_reasons`
(`cuda_core/cuda/core/system/_device.pyx:670` and
`cuda_core/cuda/core/system/_device.pyx:691`), in `CoolerInfo.target`
(`cuda_core/cuda/core/system/_cooler.pxi:76`), and in
`get_supported_event_types`
(`cuda_core/cuda/core/system/_device.pyx:811`), the error path is:

```python
for reason in _unpack_bitmask(reasons):       # reason is a bit *index* (0,1,2,...)
    try:
        output_reason = _CLOCKS_EVENT_REASONS_MAPPING[1 << reason]  # bit *value*
    except KeyError:
        raise ValueError(f"Unknown clock event reason bit: {reason}")  # reports index
```

The lookup uses `1 << reason` but the error message reports `reason`. If a
future driver introduces a new bit, the user will see a confusing message
("bit: 9" when the real unmapped value is `0x200`). Use `1 << reason` in the
message and also include the bit index for context.

**Bug — `device.brand` silently downgrades several brands to `"Unknown"`.**
`_BRAND_TYPE_MAPPING` (`cuda_core/cuda/core/system/_device.pyx:95`) is
missing `BRAND_QUADRO_RTX`, `BRAND_NVIDIA_RTX`, and `BRAND_NVIDIA` (all
defined in `nvml.BrandType`). The old code returned a typed `BrandType` enum
that could be matched against any of those; the new code uses
`.get(..., "Unknown")` so a Quadro RTX or "NVIDIA"-branded card will now
report as `"Unknown"` instead of its real brand. Either add the missing
entries or fall through to `nvml.BrandType(brand).name`.

**Bug — pre-existing `Device.__new__()` call in `EventData.device`.**
`cuda_core/cuda/core/system/_event.pxi:64` reads
`device = Device.__new__()` (no `cls` argument). This is a pre-existing bug,
not introduced by this PR, but the PR is the right time to fix it because the
file is being touched. It will raise
`TypeError: object.__new__(): not enough arguments` the first time
`event.device` is accessed.

### Sync guard (structural gap)

There is no mechanism today, and none added by this PR, to ensure the
`cuda_core` enum wrappers stay in sync with the underlying `cuda_bindings`
NVML enums. I checked for tests over `__members__`, code-gen hooks, CI
diff-checks, and runtime self-checks; none exist. The only thing close is the
runtime fallback pattern this PR introduces — `_<NAME>_MAPPING.get(value,
default)` for the inbound direction and `try: ... except KeyError: raise
ValueError(...)` for the outbound direction. That catches *missing* mappings
at the point of first encounter (a property access on a real device), but
only when the NVML driver actually returns the new value on the test host —
which CI labs don't control. Several of the bugs above (the missing
`BRAND_*` entries, the dropped `THERMAL_GPU_RELATED`) survived precisely
because nothing else flags them.

This is a real risk for `cuda.core` 1.0:

- `cuda_bindings` is auto-generated from the NVML header. The header at the
  top of `cuda_bindings/cuda/bindings/_internal/_fast_enum.py` confirms it is
  generated, and the `nvml.pyx` comment says
  *"automatically generated across versions from 12.9.1 to 13.2.0"*. NVML
  enums grow on every CUDA toolkit refresh.
- This PR turns the wrappers into hand-curated `StrEnum`s. So every NVML
  release will silently widen the binding's `_FastEnum`, while the
  `cuda.core` wrapper stays stuck at whatever was current when the wrapper
  was written. The fallback `.get()` pattern then either reports `None` /
  `"Unknown"`, or — for outbound (user → driver) — raises `ValueError` for a
  value that NVML actually supports.
- The repo already has a closed bug
  ([#1712](https://github.com/NVIDIA/cuda-python/issues/1712)) for a similar
  problem (the explanation dicts going out of sync between `cuda_core` and
  `cuda_bindings`), and a known related bug
  ([#1663](https://github.com/NVIDIA/cuda-python/issues/1663)) about Cython
  type-redefinition. So this category of drift is a known-recurring footgun
  in the codebase.

What a guard could look like, in rough order of cost vs. coverage:

1. **Cheapest — a single parametrized test** in
   `cuda_core/tests/system/` that imports both `cuda.bindings.nvml` and
   `cuda.core.system`, and for each wrapper asserts that every NVML member
   has a corresponding entry in `_<NAME>_MAPPING` (or is on a documented
   allow-list). Sketch:

   ```python
   import pytest
   from cuda.core.system import _device

   WRAPPER_TO_BINDING = [
       (_device._ADDRESSING_MODE_MAPPING, nvml.DeviceAddressingModeType,
        {"DEVICE_ADDRESSING_MODE_NONE"}),
       (_device._AFFINITY_SCOPE_MAPPING, nvml.AffinityScope, set()),
       (_device._GPU_TOPOLOGY_LEVEL_MAPPING, nvml.GpuTopologyLevel, set()),
       (_device._EVENT_TYPE_MAPPING, nvml.EventType, set()),
       (_device._BRAND_TYPE_MAPPING, nvml.BrandType, {"BRAND_COUNT"}),
       # ... one entry per wrapper
   ]

   @pytest.mark.parametrize("mapping, binding, intentionally_unmapped",
                            WRAPPER_TO_BINDING)
   def test_wrapper_covers_all_binding_members(mapping, binding,
                                               intentionally_unmapped):
       binding_keys = set(binding.__members__) - intentionally_unmapped
       mapped_keys = (
           {m.name for m in mapping.keys() if isinstance(m, binding)}
           | {m.name for m in mapping.values() if isinstance(m, binding)}
       )
       missing = binding_keys - mapped_keys
       assert not missing, (
           f"{binding.__name__} is missing wrapper entries for: {missing}"
       )
   ```

   The `intentionally_unmapped` set is the explicit allow-list (e.g. `*_COUNT`
   sentinels, deprecated `APP_CLOCK_*`, the typo'd
   `P2P_STATUS_CHIPSET_NOT_SUPPORED`, etc.). When NVML adds a new member,
   this test fails on every CI host (no GPU required), and a maintainer
   either adds a wrapper entry or extends the allow-list with a comment
   explaining why.

2. **Medium — a small import-time `_validate_mappings()`** behind
   `if __debug__:` or behind a `CUDA_PYTHON_VALIDATE_ENUMS=1` env var. Same
   idea as #1, but lives next to the mappings so the failure mode is
   `ImportError` on `cuda.core.system`. I'd lean against this for a 1.0
   library — too noisy at import — but it's an option.

3. **Heavier — code-gen the mappings.** Since `cuda_bindings` is already
   generator-driven, the `_<NAME>_MAPPING` dicts could be too: feed the same
   NVML header → emit a `_generated_mappings.pxi` with the forward dict, plus
   a side file listing "human-curated" `StrEnum` names that humans then
   maintain. The generator emits a placeholder `# TODO map <new_member>`
   comment when a new NVML member appears, which fails CI via a regex check.
   This is the strongest guarantee but is a much bigger change.

4. **Adjacent — pin a binding floor.** `cuda_core` already has
   `cuda-bindings[all]==12.*` / `13.*` in extras; if the wrappers target a
   specific NVML enum surface, pin `cuda-bindings>=X,<Y` so a *user* who
   mixes `cuda-core` 1.0.0 with a newer `cuda-bindings` gets a clean failure
   rather than silent `"Unknown"` / `ValueError`. (Issue
   [#1715](https://github.com/NVIDIA/cuda-python/issues/1715) was exactly the
   inverse problem — `cuda.core` demanding an unreleased bindings version —
   so the team already has scar tissue here.)

Given the milestone is May 7 and this review already turned up missing brand
entries and a missing cooler target, I'd push for option **#1** as a blocker
for 1.0: it's ~80 lines, table-driven, no GPU required, and the allow-list
doubles as documentation for *why* certain NVML members are deliberately not
surfaced (deprecated / sentinel / typo / composite bitmask). I'm happy to
draft that test against the current branch so it's ready to drop in on top of
this PR.

### Behavior / compatibility

**Behavior — `get_supported_event_types` includes `EventType.NONE` mapping,
but the bitmask path can never produce it.**
`EventType.NONE = "none"` is added to the wrapper but the bitmask path can
never produce it (no bit is set). `_EVENT_TYPE_MAPPING` includes
`nvml.EventType.NONE → EventType.NONE` only because it is needed for the
`EventData.event_type` property. That's fine, but consider documenting that
`EventType.NONE` is reserved for "no event" and isn't a registrable type.
Currently `device.register_events([EventType.NONE])` is a silent no-op
(bitmask stays 0), which is surprising; consider raising.

**Compatibility — Call out that this PR is a breaking change in the PR description /
   release notes. The `breaking` label is missing.**

**Behavior — `device.brand: BrandType → str`** (also see the bug above).
Switching to a free-form `str` (with `"Unknown"` fallback) means callers
can't reliably enumerate brands or do `==` against a name they didn't see in
CI. Was a `Brand` `StrEnum` considered and rejected? If yes, mention it in
the PR body so reviewers don't ask again.

**Behavior — `NvLink.version: NvlinkVersion → tuple[int, int]`.** This is a
clean improvement, but it's not in the issue #1995 scope. It would be good to
call out in the PR description that it's an additional change so it doesn't
surprise users.

**Behavior — `PciInfo.get_throughput(counter)` →
`rx_throughput` / `tx_throughput` properties.** Same — this is good
cleanup, but it changes the public API shape and isn't in the PR title or
description. Worth noting.

### Docs

**Docs — three private-doc enums still listed as cyclasses.**
`cuda_core/docs/source/api_private.rst:74` keeps
`system._device.GpuP2PCapsIndex`, `system._device.GpuP2PStatus`, and
`system._device.GpuTopologyLevel` under
`:template: autosummary/cyclass.rst`. After this PR they are pure-Python
`StrEnum`s, not Cython classes. They should be moved to the lower section
(`cuda_core/docs/source/api_private.rst:92`+) alongside `AddressingMode`,
`AffinityScope`, etc., or rendered with the default template; otherwise the
docs build is likely to warn or render incorrectly.

**Docs — `Device.performance_state` documentation feels split.**
`cuda_core/cuda/core/system/_device.pyx:957` returns `int | None`. The
current doc and runtime contract say "0 is highest, 15 is lowest, `None` if
unknown". That's reasonable, but `dynamic_pstates_info` and
`register_events([EventType.PSTATE])` still use Pstate concepts; users who
read those docstrings have to mentally context-switch. Consider documenting
once on `system.Device` and cross-referencing.

**Docs — stale references to old enum names.**

- `cuda_core/cuda/core/system/_event.pxi:79`,
  `cuda_core/cuda/core/system/_event.pxi:92`, and
  `cuda_core/cuda/core/system/_event.pxi:105` still reference
  `EventType.EVENT_TYPE_XID_CRITICAL_ERROR` in docstrings. The new value is
  `EventType.XID_CRITICAL_ERROR`. The Sphinx links will fail (or worse,
  render as broken refs).
- `cuda_core/cuda/core/system/_system_events.pyx:168` example uses
  `SystemEventType.SYSTEM_EVENT_TYPE_GPU_DRIVER_UNBIND`. The new name is
  `SystemEventType.UNBIND`.
- `cuda_core/cuda/core/system/_event.pxi:95` (`gpu_instance_id`) and
  `cuda_core/cuda/core/system/_event.pxi:108` (`compute_instance_id`)
  docstrings still use the old `EVENT_TYPE_XID_CRITICAL_ERROR` name.

### Tests

**Test — temperature-thresholds slice loses one threshold.**
`cuda_core/tests/system/test_system_device.py:680` is
`for threshold in list(system.TemperatureThresholds)[:-1]:`. Pre-PR, `[:-1]`
stripped `TEMPERATURE_THRESHOLD_COUNT`. Post-PR, the new `StrEnum` has 8 real
values and no sentinel; `[:-1]` now strips `GPS_CURR` instead. Drop the
`[:-1]`. Same audit for any other `[:-1]` over an enum in tests.

**Test — `register_events` test misses positive case for `str` input.**
`cuda_core/tests/system/test_system_device.py:265-286` was simplified, but it
only checks `register_events([])`, `register_events(0)` (now correctly
rejected), and a pre-existing typed-list path on the systems test. It would
be good to also assert that `device.register_events("xid_critical_error")`
and `device.register_events([EventType.XID_CRITICAL_ERROR])` both work, since
the whole point of this PR is dual support.

### Style / nits

**Style — module-level docstring assignment is repeated and noisy.**
The pattern

```python
class ClockId(StrEnum):
    """..."""
    CURRENT = "current"
ClockId.CURRENT.__doc__ = "Current actual clock value."
```

is duplicated across 8+ enums. Per #1995 mdboom called this "ugly but
works." Consider a small helper, e.g.
`_set_member_docs(ClockId, {"CURRENT": "...", "CUSTOMER_BOOST_MAX": "..."})`,
both for readability and to make it easy to lift these into autodoc later.

**Style — `import sys` + version branch duplicated in two `.pyx` files.**
`cuda_core/cuda/core/system/_device.pyx:8` and
`cuda_core/cuda/core/system/_system_events.pyx:8` have identical
`if sys.version_info >= (3, 11):` guards. With
`cuda_core/pyproject.toml:22` requiring `>=3.10`, this is necessary. Consider
centralizing in a shared `_compat.pxi` (or in `_device_utils.pxi`) and
`include`-ing it from both, so future bumps to 3.11 are a single edit.

**Style — inverse mappings constructed inconsistently.**
`_GPU_TOPOLOGY_LEVEL_INV_MAPPING`, `_EVENT_TYPE_INV_MAPPING`,
`_THERMAL_TARGET_INV_MAPPING`, `_SYSTEM_EVENT_TYPE_INV_MAPPING` are all built
as `{v: k for k, v in <forward>.items()}`. Several other dicts (e.g.,
`_CLOCK_ID_MAPPING`, `_GPU_P2P_CAPS_INDEX_MAPPING`) are forward-only and used
only one way. Be explicit and consistent: either always have both, or comment
why the asymmetry exists.

**Style — `get_thermal_settings` has a stranded `TODO` comment.**
`cuda_core/cuda/core/system/_temperature.pxi:257` has
`# TODO: The above docstring is from the NVML header, but it doesn't seem to
make sense.` *after* the docstring. It would be clearer in the docstring
itself, or fixed properly while you're here.

**Misc — `nvml.GpuP2PStatus` typo handled but with redundant key.**
`_GPU_P2P_STATUS_MAPPING` (`cuda_core/cuda/core/system/_device.pyx:164`) maps
both `P2P_STATUS_CHIPSET_NOT_SUPPORED` (typo) and
`P2P_STATUS_CHIPSET_NOT_SUPPORTED`. Per the binding, these are aliases (same
int value), so one of these dict entries silently overwrites the other at
construction time. That's harmless, but worth a comment so a future cleanup
doesn't accidentally drop the typo'd alias and break callers using older
NVML.

**Misc — `CoolerTarget.GPU_RELATED` not exposed.**
`nvml.CoolerTarget.THERMAL_GPU_RELATED` is a composite bitmask
(`GPU | MEMORY | POWER_SUPPLY`); the new `CoolerTarget` `StrEnum` drops it
entirely. If the underlying NVML field is set to `THERMAL_GPU_RELATED` (and
the device reports it as a single composite rather than three individual
bits), the new code may iterate three bits and produce a longer list than
before. That's probably what's wanted, but worth a sentence in the property
docstring — and is exactly the sort of thing the sync-guard test would have
flagged.

## Open questions

1. Was a `Brand` `StrEnum` considered for `device.brand` and rejected? The
   current free-form `str` makes typed comparisons in user code impossible.
2. Were the `NvlinkVersion → tuple[int, int]` and
   `PciInfo.get_throughput → rx_throughput / tx_throughput` changes intended
   to land here, or as separate PRs? They expand the scope beyond #1995.
3. Is the team OK with the silent breaking change for callers who were
   passing `nvml.<EnumType>.X` directly? If yes, please call it out in the PR
   body and add the `breaking` label; if no, consider a one-release deprecation
   path that accepts the old `_FastEnum` value with a `DeprecationWarning`.

cuda_core/cuda/core/system/_device.pyx
```diff
@@ -42,6 +51,151 @@ include "_temperature.pxi"
...
+    GPU peer-to-peer status.
+    """
+    OK = "ok"
+    CHIPSET_NOT_SUPPORTED = "chipset not supported"
+    GPU_NOT_SUPPORTED = "GPU not supported"
+    IOH_TOPOLOGY_NOT_SUPPORTED = "IOH topology not supported"
+    DISABLED_BY_REGKEY = "disabled by regkey"
+    NOT_SUPPORTED = "not supported"
+    UNKNOWN = "unknown"
+cdef dict _GPU_P2P_STATUS_MAPPING = {
```
leofang: nit: two empty lines

mdboom:
`<rant>`I think these epic auto-reviews are really hard to address, because they include a lot of reasonable comments with a lot of noise.  Can whatever robo-tool being used actually comment on lines in the PR?  I already pre-reviewed this with the same model, so a lot of these ideas I already rejected, but then of course there are mistakes that this found that my run didn't so it's not entirely without value.  Just seems like a lot of extra time vs. how GitHub was designed to be used  :shrug:`</rant>`

I am going to respond to everything I think is wrong.  If not mentioned, assume I have addressed and fixed it.

> Bug — test_nvlink references a removed symbol. cuda_core/tests/system/test_system_device.py:759 still does
assert isinstance(version, system.NvlinkVersion)

Nope.

```
 rg system\.NvLinkVersion
```

> Bug — supported_pstates can iterate beyond the valid range.
cuda_core/cuda/core/system/_device.pyx:1001 walks
nvml.device_get_supported_performance_states(...) and only filters out
PSTATE_UNKNOWN (= 32). The NVML header says unused trailing slots are
PSTATE_UNKNOWN, but if the driver ever returned a value outside 0..15
(e.g., a future PSTATE_*), _pstate_to_int would silently return
int(x) - 0, producing an out-of-contract integer.

Disagree.  If NVML doesn't follow the enum contract, all bets are off.  I don't think we do this kind of defensive programming elsewhere.  And it's Python -- it will raise, not segfault.

> Bug — _pstate_to_enum name is wrong; takes an int and returns an int.
cuda_core/cuda/core/system/_device.pyx:29 is named _pstate_to_enum but
the body is
return int(pstate) + int(nvml.Pstates.PSTATE_0)

Since the enums used are Python enums, not Cython enums, we need to type it this way.  Confusing, sure, but it's an internal convenience function.  The fact that it's a no-op is fine -- it's to be resilient to changes in the underlying enum.

> Bug — device.brand silently downgrades several brands to "Unknown".

Yep.  Looks like your Opus 4.7 caught the error from my Opus 4.7 🙃 

> Behavior — get_supported_event_types includes EventType.NONE mapping,
but the bitmask path can never produce it. ... Currently device.register_events([EventType.NONE]) is a silent no-op
(bitmask stays 0), which is surprising; consider raising.

I think this is fine as-is.

> Behavior — device.brand: BrandType → str

Yes, brands aren't really designed to be acted on programmatically -- they are primarily just a name to display to the user.

> Docs — Device.performance_state documentation feels split.

Yes, this is the downside of moving away from numerical enums to an int.  I'm not sure I agree with the solution.

> Style — module-level docstring assignment is repeated and noisy.

Yes, but it's very explicit.  I think the model's suggestion here is too magical.

> Style — import sys + version branch duplicated in two .pyx files.

Again, this is fine, and pretty standard practice.  If we do want to unify Python compat code, we should do it as a separate sweep.

> Style — inverse mappings constructed inconsistently.

Python doesn't have dead code elimination, so we shouldn't create private objects that we would never use.

> Style — get_thermal_settings has a stranded TODO comment.

No.  This is where it should go.  It can't go above the docstring.

> Misc — nvml.GpuP2PStatus typo handled but with redundant key.
_GPU_P2P_STATUS_MAPPING (cuda_core/cuda/core/system/_device.pyx:164) maps
both P2P_STATUS_CHIPSET_NOT_SUPPORED (typo) and
P2P_STATUS_CHIPSET_NOT_SUPPORTED. Per the binding, these are aliases (same
int value), so one of these dict entries silently overwrites the other at
construction time.

I don't think we can guarantee that will always be the case, so it's safer as-is.

> Was a Brand StrEnum considered for device.brand and rejected? The
current free-form str makes typed comparisons in user code impossible.

Yes, I think that's the right choice.

> Were the NvlinkVersion → tuple[int, int] and
PciInfo.get_throughput → rx_throughput / tx_throughput changes intended
to land here, or as separate PRs? They expand the scope beyond https://github.com/NVIDIA/cuda-python/issues/1995.

Yes.  The scope was really to "be Pythonic", not to strictly adhere to wrapping enums as-is.  If that were the goal, why bother with this at all? 

> Is the team OK with the silent breaking change for callers who were
passing nvml.<EnumType>.X directly? If yes, please call it out in the PR
body and add the breaking label; if no, consider a one-release deprecation
path that accepts the old _FastEnum value with a DeprecationWarning.

Yes, prior to 1.0 we are fine with any breaking change without notice.


## Sync guards

The agent's analysis of this problem really gets to the heart of why I wasn't sure any of this was a good idea.  Though a lot of its analysis is based on an incomplete understanding about version compatibility and guarantees between `cuda_core` and `cuda_bindings`.  However, I plan to experiment with its first suggestion.  If that doesn't bear immediate fruit, we can separate it out into a separate issue, and consider sync guards across all of `cuda_core`.  That is not blocking for 1.0

rwgk:
> Can whatever robo-tool being used actually comment on lines in the PR?

It can, but I intentionally didn't make use of that feature. I believe it'll make it even harder to know what came from where and when. — I figure an agent can easily go from the details in the comments to code edits, so figured it's better to have one review in a one-piece comment.

> I already pre-reviewed this with the same model, so a lot of these ideas I already rejected,

We could post such things as comments on the PRs before sending them out for review. — But again, I wouldn't want to have the tool auto-post comments. I agree there is a lot of noise, I want to weed out what I can, and actually read at least once what it wrote. So I always post my comments manually. (Maybe one day that'll be futile, as the tools get better, but I don't think we're there yet.)

cuda_core/cuda/core/system/_device.pyx
```diff
@@ -5,22 +5,31 @@
...
-BrandType = nvml.BrandType
-DeviceArch = nvml.DeviceArch
-GpuP2PCapsIndex = nvml.GpuP2PCapsIndex
-GpuP2PStatus = nvml.GpuP2PStatus
-GpuTopologyLevel = nvml.GpuTopologyLevel
-Pstates = nvml.Pstates
+cdef object _pstate_to_int(object pstate):
+    if pstate == nvml.Pstates.PSTATE_UNKNOWN:
+        return None
+    return int(pstate) - int(nvml.Pstates.PSTATE_0)
```
rwgk, line 32: I wouldn't let this go unchecked (bugs happen), but keep it simple:

```
    assert int(pstate) >= int(nvml.Pstates.PSTATE_0)
```

This is the only item my agent still pulled out as noteworthy.

rwgk (approved):

cuda_core/cuda/core/system/_clock.pxi
```diff
@@ -3,9 +3,67 @@
...
-ClocksEventReasons = nvml.ClocksEventReasons
-ClockType = nvml.ClockType
+class ClockId(StrEnum):
+    """
+    Clock Ids. These are used in combination with :class:`ClockType` to specify a single clock value.
+    """
+    CURRENT = "current"
+    CUSTOMER_BOOST_MAX = "customer_boost_max"
+    # APP_CLOCK_TARGET and APP_CLOCK_DEFAULT are deprecated so not included here
+ClockId.CURRENT.__doc__ = """
```
leofang: Format: Two empty lines

leofang: (this and everywhere else, can we make it easier for eyeballing? I trust my eyes more than agents if this is made easier.)

cuda_core/cuda/core/system/_clock.pxi
```diff
@@ -48,26 +106,40 @@ cdef class ClockInfo:
...
+        clock_id: :class:`ClockId` | str
+            The clock ID to query.  Defaults to the current clock value.
 
         Returns
         -------
         int
             The clock speed in MHz.
         """
+        try:
+            clock_id = _CLOCK_ID_MAPPING[clock_id]
```
leofang, line 141: Wow, this is dark magic... I wouldn't expect `str` to be convertible to an enumerator... OK...

cuda_core/cuda/core/system/_event.pxi
```diff
@@ -3,7 +3,50 @@
...
+    FATAL_POISON_ERROR = "fatal_poison_error"
+    GPU_UNAVAILABLE_ERROR = "gpu_unavailable_error"
+    GPU_RECOVERY_ACTION = "gpu_recovery_action"
+EventType.PSTATE.__doc__ = """
+Event about PState changes
+
+On Fermi™ architecture, PState changes are also an indicator that GPU is throttling down due to
+no work being executed on the GPU, power capping or thermal capping. In a typical situation,
+Fermi-based GPU should stay in P0 for the duration of the execution of the compute process.
+"""
```
leofang, line 33: Q: What is our guidance for repeating `__doc__`? I see we don't always do this

cuda_core/cuda/core/system/_device.pyx
```diff
@@ -5,22 +5,31 @@
...
+if sys.version_info >= (3, 11):
+    from enum import StrEnum
+else:
+    from backports.strenum import StrEnum
 from multiprocessing import cpu_count
 from typing import Iterable
+import warnings
 
 from cuda.bindings import nvml
+from cuda.bindings._internal._fast_enum import FastEnum
```
leofang: Q: Do we need a try-except to check if `FastEnum` exists (and fall back to `IntEnum` if not)?

leofang (commented):
Thanks, Mike, no major issues found.

cuda_core/cuda/core/system/_event.pxi
mdboom: If `nvml.h` has a meaningful docstring for an enum value, I've included it here.  Otherwise I don't.  (Some are completely undocumented, some just repeat the name, which seems useless.  So I don't include those here).  But we can always add more docs in the future...

cuda_core/cuda/core/system/_device.pyx
mdboom: Oh, yeah, I suppose this limits us to a recent-ish cuda_bindings.  But all of `cuda.core.system` is already limited in the same way...  Anyway, it can't hurt to be careful.

cuda_core/cuda/core/system/_clock.pxi
mdboom: I can make these changes.  This is actually something where I overrode my agent.  I found it easier to see the enum+mapping is a logical unit by not putting the space, but I know that's not `ruff-compliant`.

cuda_core/cuda/core/system/_clock.pxi
mdboom: Do you mean the dark magic will be surprising to the user as an API?

Internally, this is just a dictionary mapping strings to an integer enum value, so not terribly magical.

leofang: Yeah this is not user-visible, it's only surprising to me that this works 😄

cuda_core/cuda/core/system/_clock.pxi
leofang: Interesting that our cython-lint does not complain 😛

cuda_core/cuda/core/system/_device.pyx
leofang: That is actually a good point -- IIRC `FastEnum` was introduced at the same time when the `nvml` bindings were released?

mdboom: It was a little bit after.  So I think doing a `try/except ImportError` thing is not a bad idea, at least for a little while.

cuda_core/cuda/core/system/_clock.pxi
mdboom: Yeah, it's just a linter, not a formatter.  We need to nerd-snipe Astral into adding Cython support.

cuda_core/cuda/core/system/_clock.pxi
```diff
@@ -99,37 +175,41 @@ cdef class ClockInfo:
         """
         return nvml.device_get_max_customer_boost_clock(self._handle, self._clock_type)
 
-    def get_min_max_clock_of_pstate_mhz(self, pstate: Pstates) -> tuple[int, int]:
+    def get_min_max_clock_of_pstate_mhz(self, pstate: int) -> tuple[int, int]:
```
rparolin, line 178: Are we intending to move to a integral type here: `Pstates -> int`?

cuda_core/cuda/core/system/_device.pyx
```diff
@@ -5,22 +5,37 @@
 from libc.stdint cimport intptr_t, uint64_t
 from libc.math cimport ceil
 
+import sys
+if sys.version_info >= (3, 11):
+    from enum import StrEnum
```
rparolin, line 10: Nit: Would it better to have a common .py file that holds this implementation and all other cuda-python code can just import it?  Rather than copy this version check to all the sites that need it?

cuda_core/cuda/core/system/_device.pyx
```diff
@@ -5,22 +5,37 @@
...
+else:
+    from backports.strenum import StrEnum
 from multiprocessing import cpu_count
 from typing import Iterable
+import warnings
 
 from cuda.bindings import nvml
+try:
+    from cuda.bindings._internal._fast_enum import FastEnum
```
rparolin, line 19: Nit: Ditto above about hoisting this backwards compat code into a common location that can be referenced.

cuda_core/cuda/core/system/_event.pxi
```diff
@@ -27,17 +76,17 @@ cdef class EventData:
         """
         The type of event that was triggered.
         """
-        return EventType(self._event_data.event_type)
+        return _EVENT_TYPE_MAPPING[self._event_data.event_type]
```
rparolin, line 79: Nit: Bare dict lookup will raise `KeyError` if the driver returns an event type not in `_EVENT_TYPE_MAPPING`. Consider `.get(...)` with a sentinel/`None` (or wrap with a clearer error) so a property accessor doesn't blow up on values introduced by newer drivers.

cuda_core/cuda/core/system/_system_events.pyx
```diff
@@ -28,7 +48,7 @@ cdef class SystemEvent:
         """
         The :obj:`~SystemEventType` that was triggered.
         """
-        return SystemEventType(self._event_data.event_type)
+        return _SYSTEM_EVENT_TYPE_MAPPING[self._event_data.event_type]
```
rparolin, line 51: Nit: Same as in `_event.pxi` — bare lookup raises `KeyError` on unmapped values. Consider `.get(...)` with a fallback for forward-compat with newer drivers.

cuda_core/cuda/core/system/_fan.pxi
```diff
@@ -94,7 +105,7 @@ cdef class FanInfo:
 
         For all CUDA-capable discrete products with fans.
         """
-        return FanControlPolicy(nvml.device_get_fan_control_policy_v2(self._handle, self._fan))
+        return _FAN_CONTROL_POLICY_MAPPING[nvml.device_get_fan_control_policy_v2(self._handle, self._fan)]
```
rparolin, line 108: Nit: Bare `_FAN_CONTROL_POLICY_MAPPING[...]` lookup will raise `KeyError` for any policy value the wrapper doesn't know. Other wrappers in this PR use `.get(..., fallback)` — worth being consistent.

cuda_core/cuda/core/system/_device.pyx
```diff
@@ -5,22 +5,37 @@
...
-GpuP2PCapsIndex = nvml.GpuP2PCapsIndex
-GpuP2PStatus = nvml.GpuP2PStatus
-GpuTopologyLevel = nvml.GpuTopologyLevel
-Pstates = nvml.Pstates
+cdef object _pstate_to_int(object pstate):
+    if pstate == nvml.Pstates.PSTATE_UNKNOWN:
+        return None
+    assert (
+        int(pstate) >= 0 and int(pstate) <= 15
+    ), f"Invalid P-state: {pstate}. Must be between 0 and 15 inclusive, or PSTATE_UNKNOWN."
```
rparolin, line 31: `assert` is stripped under `python -O`, so this bounds check silently disappears in optimized runs. Prefer raising `ValueError` for runtime input validation.

cuda_core/cuda/core/system/_nvlink.pxi
```diff
@@ -18,18 +26,24 @@ cdef class NvlinkInfo:
...
         -------
-        NvlinkVersion
-            The Nvlink version.
+        tuple[int, int]
+            The Nvlink version as a tuple of (major, minor).
         """
-        return NvlinkVersion(nvml.device_get_nvlink_version(self._device._handle, self._link))
+        version = nvml.device_get_nvlink_version(self._device._handle, self._link)
+        if version == nvml.NvlinkVersion.VERSION_INVALID:
+            raise RuntimeError(f"Invalid NvLink version returned for device")
```
rparolin, line 42: f-string with no `{}` interpolation — either drop the `f` prefix or include the offending version value in the message (more useful for debugging). Ruff would flag this as F541; cython-lint won't.

cuda_core/cuda/core/system/_device.pyx
```diff
@@ -42,6 +57,174 @@ include "_temperature.pxi"
...
+    DISABLED_BY_REGKEY = "disabled by regkey"
+    NOT_SUPPORTED = "not supported"
+    UNKNOWN = "unknown"
+
+
+_GPU_P2P_STATUS_MAPPING = {
+    nvml.GpuP2PStatus.P2P_STATUS_OK: GpuP2PStatus.OK,
+    # Typo in upstream library
+    nvml.GpuP2PStatus.P2P_STATUS_CHIPSET_NOT_SUPPORED: GpuP2PStatus.CHIPSET_NOT_SUPPORTED,
+    nvml.GpuP2PStatus.P2P_STATUS_CHIPSET_NOT_SUPPORTED: GpuP2PStatus.CHIPSET_NOT_SUPPORTED,
```
rparolin, line 193: The typo'd `P2P_STATUS_CHIPSET_NOT_SUPPORED` and the corrected `P2P_STATUS_CHIPSET_NOT_SUPPORTED` alias to the same integer value, so the second entry overwrites the first in the dict — the first line is dead. Either drop one or add a comment noting it's intentional dedup coverage.

cuda_core/tests/test_enum_coverage.py
```diff
@@ -0,0 +1,275 @@
...
+
+import pytest
+
+import cuda.core
+from cuda.core import system
+
+if sys.version_info >= (3, 11):
+    from enum import StrEnum
+else:
+    from backports.strenum import StrEnum
```
rparolin, line 23: Nit: This `StrEnum` / `backports.strenum` compat block is duplicated yet again here. Folds into the hoist suggested in the existing threads on `_device.pyx` (r3185426568, r3185428204).

cuda_core/cuda/core/system/_system_events.pyx
mdboom: This is fine and not worth the performance impact.  The value comes from C++ code.  It would be a runtime, not a user error, if that were ever to happen.

cuda_core/cuda/core/system/_fan.pxi
mdboom: The other wrappers use `.get` because the value is either explicitly unbounded in the NVML docs or comes from the user.  That is not the case here.

cuda_core/cuda/core/system/_device.pyx
mdboom: Again, we don't need to validate values coming from NVML, IMHO.

cuda_core/tests/test_enum_coverage.py
mdboom: See #2019.

PR #2015, opened 2026-05-04
[FIX]: pathfinder: Add missing static-lib CUDA 12 path for Windows conda  layouts.

rwgk:
Latent bug discovered while testing PR #1977 (xref issue #1038) with Windows Conda.

**Source change — `find_static_lib.py`**

- The `_StaticLibInfo` schema had a single conda lookup directory per library (`conda_rel_path: str`). On Windows that was hard-coded to `Library/lib/x64`, but only CUDA 13 conda environments lay `cudadevrt.lib` out that way; CUDA 12 conda environments put it in `Library/lib`. So on a Windows + CUDA 12 conda setup, pathfinder skipped the conda anchor entirely and fell through to `CUDA_PATH`, often producing a misleading "not found" error pointing at the wrong location.
- The fix changes the schema to a *tuple* of lookup paths (`conda_rel_paths: tuple[str, ...]`) so each lib can declare every layout it knows about. For `cudadevrt`:
  - Windows: `("lib/x64", "lib")` — try the CUDA 13 layout first, fall back to the CUDA 12 layout.
  - Linux: `("lib",)` — unchanged, single entry.
- `_FindStaticLib.try_with_conda_prefix` now loops through `self.conda_rel_paths` and returns the first hit instead of probing one fixed subdirectory.

**Test change — `test_find_static_lib.py`**

- Existing search-order test (`test_locate_static_lib_search_order`) is updated mechanically: `CUDADEVRT_INFO["conda_rel_path"]` → `CUDADEVRT_INFO["conda_rel_paths"][0]`.
- New test `test_locate_static_lib_conda_rel_path_fallback`:
  - Asserts the fallback semantic by injecting a deliberately-missing first entry (`"missing-first"`) when the platform only ships a single path. On Windows, where the tuple already has two entries, it uses them as-is.
  - Drops the file at the *second* listed path under the conda anchor and verifies `locate_static_lib("cudadevrt")` finds it with `found_via="conda"`.
  - Real coverage: confirms the loop tries entries in order and a missing first entry doesn't abort the search.

Real (not mock-based) testing was reported here:

* https://github.com/NVIDIA/cuda-python/pull/1977#issuecomment-4332733658 — **PR 1977 conda testing results analysis**

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_static_libs/find_static_lib.py
+++ b/cuda_pathfinder/cuda/pathfinder/_static_libs/find_static_lib.py
@@ -28,7 +28,7 @@ class LocatedStaticLib:
 class _StaticLibInfo(TypedDict):
     filename: str
     ctk_rel_paths: tuple[str, ...]
-    conda_rel_path: str
+    conda_rel_paths: tuple[str, ...]
     site_packages_dirs: tuple[str, ...]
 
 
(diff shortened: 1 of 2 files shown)


rparolin (approved):

rwgk:
Thanks a lot @rparolin!

PR #1775, opened 2026-03-17
Add managed-memory advise, prefetch, and discard-prefetch free functions

rparolin:
## Summary

Adds managed-memory range operations to `cuda.core`:

- **Free functions** in `cuda.core.utils`: `advise`, `prefetch`, `discard`, `discard_prefetch`. Each accepts either a single `Buffer` or a sequence; N==1 dispatches to the per-range driver entry point and N>1 dispatches to the corresponding `cuMem*BatchAsync` (CUDA 13+).
- **`Host`** — new top-level singleton class symmetric to `Device`. `Host()` (any host), `Host(numa_id=N)`, `Host.numa_current()`. Same-argument constructions are interned (`Host() is Host()`). Used together with `Device` to express managed-memory locations.
- **`ManagedBuffer`** — `Buffer` subclass returned by `ManagedMemoryResource.allocate`. Exposes a Pythonic property-style advice API on top of the same free functions. Wrap an external managed pointer with `Buffer.from_handle(...)` (now a `@classmethod`, so `ManagedBuffer.from_handle(...)` returns a `ManagedBuffer`).

Closes #1332. Addresses the managed-memory portion of #1333 (P1: `cuMemPrefetchBatchAsync`, `cuMemDiscardBatchAsync`, `cuMemDiscardAndPrefetchBatchAsync`). The P0 `cuMemcpyBatchAsync` from #1333 is intentionally out of scope and tracked separately; the holistic batched-API contract this PR commits to is documented in [#issuecomment-4355502334](https://github.com/NVIDIA/cuda-python/pull/1775#issuecomment-4355502334) so the upcoming `cuMemcpyBatchAsync` work can mirror it.

### `ManagedBuffer` — property-style advice on managed allocations

`ManagedMemoryResource.allocate` returns a `ManagedBuffer` (a `Buffer` subclass). All ManagedBuffer-specific behavior is layered on top of the free functions, so the two surfaces stay consistent.

```python
from cuda.core import Device, Host, ManagedMemoryResource

mr = ManagedMemoryResource()
buf = mr.allocate(size)                # ManagedBuffer

# Driver-backed properties — getter queries the driver, setter calls cuMemAdvise.
buf.read_mostly = True
buf.preferred_location = Device(0)     # or Host(), or Host(numa_id=N)
buf.preferred_location = None          # unset

# Live set-like view of `set_accessed_by` advice.
buf.accessed_by.add(Device(1))
buf.accessed_by.discard(Device(1))
[...]

PR #2016, opened 2026-05-04
Fix #1995: Use StrEnum for enum-like strings

### LINKED ISSUE #1995: cuda.core: Replace string literals with enums in public API

mdboom:
See discussion in #1995.


rwgk:
# PR 2016 Agent Review

**Reviewer:** Opus 4.7 1M (Max), 2026-05-04

With a few small manual edits. I read through the entire review but only checked a few of the many details.

---

## Context

This PR is the second half of the cuda.core v1.0.0 enum hardening that PR #2014 started. Where 2014 wrapped NVML enums for `cuda.core.system`, this PR introduces eleven `enum.StrEnum` types for the compilation/linking pipeline and the memory APIs:

- `_program.pyx`: `SourceType`, `CodeType`, `CompilerBackend`, `PchStatus`
- `_memory/_virtual_memory_resource.py`: `VirtualMemoryAllocationType`, `VirtualMemoryLocationType`, `VirtualMemoryHandleType`, `VirtualMemoryGranularityType`, `VirtualMemoryAccessType`
- `_memory/_managed_memory_resource.pyx`: `ManagedMemoryLocationType`
- `graph/_graph_definition.pyx`: `GraphMemoryType`

It uses `from enum import StrEnum` with a `backports.strenum` fallback for Python < 3.11 (added to both `pyproject.toml` and `pixi.toml`). Argument-side methods accept either the `StrEnum` member or a plain string; return-side values are still mostly plain strings, matching the migration approach proposed in #1995.

---

## Blocking Issues

These should be addressed in this PR.

### 1. `MemoryLocationType` is undefined (typo)

`cuda_core/cuda/core/_memory/_managed_memory_resource.pyx:110`

```python
def preferred_location(self) -> tuple[MemoryLocationType, int | None] | None:
```

There is no symbol named `MemoryLocationType` anywhere in the workspace. The intended name is `ManagedMemoryLocationType` (introduced in this same PR). Because the file uses `from __future__ import annotations`, the typo is silent at import time, but `typing.get_type_hints(...)`, IDE inference, and Sphinx autodoc all raise `NameError` when resolving the annotation.

**Fix:** rename to `ManagedMemoryLocationType` (already in scope at line 26).

### 2. `Program.backend` / `Program.pch_status` claim enum return types but return `str`

`cuda_core/cuda/core/_program.pyx:178-180` and `:151-175`

The properties advertise `-> CompilerBackend` and `-> PchStatus | None`, but:

- `_program.pxd` declares `cdef str _backend` and `cdef str _pch_status`.
- `_backend` is set with `self._backend = str(CompilerBackend.NVRTC)` at lines 684, 694, 728.
- `_pch_status` is set with `str(status)` at line 862, but with the bare enum (`PchStatus.FAILED`) at line 884.

Net result: `program.backend` is always a plain `str`; `program.pch_status` is sometimes a plain `str` and sometimes a `PchStatus`, depending on the code path. Compare with `Linker.backend` (`_linker.pyx:171-174`), which now returns a real `CompilerBackend` member — so two related public properties advertising the same return type behave differently.

**Fix:** pick one direction and apply it consistently across `Program.backend`, `Program.pch_status`, and `Linker.backend`. Recommended:

- Change `cdef str _backend` / `cdef str _pch_status` in `_program.pxd` to `cdef object`.
- Drop the `str(...)` wrappers in `_program.pyx` (lines 684, 694, 728, 862).
- Keep the doc-stated enum return types.

This makes `isinstance(p.backend, CompilerBackend)` and `isinstance(p.pch_status, PchStatus)` actually true. All existing test assertions (`program.backend == "NVRTC"`, etc.) continue to pass via `StrEnum` equality.

### 3. New public enums are documented under `api_private.rst`

`cuda_core/docs/source/api_private.rst:24-34`

All eleven new enums (`CodeType`, `CompilerBackend`, `GraphMemoryType`, `ManagedMemoryLocationType`, `PchStatus`, `SourceType`, `VirtualMemory*Type`) are listed in `api_private.rst`, which carries `:orphan:` and is described in its own header comment as "for private classes exposed to users". These enums are reachable as `cuda.core.CodeType` etc., live in `cuda.core.__all__`, and are named in user-facing docstrings — they are public API.

**Fix:** move them into `api.rst`, ideally as a new "Enums" subsection within each owning section:

- `CodeType`, `CompilerBackend`, `PchStatus`, `SourceType` under "CUDA compilation toolchain".
- `GraphMemoryType` under "CUDA graphs".
- `VirtualMemory*Type` and `ManagedMemoryLocationType` under "Memory management".

Mirror the layout the system module already uses in `api.rst:204-229` ("Enums" subsection under "CUDA system information…").

---

## Recommended Polish (this PR or a quick follow-up)

These are smaller items the author may want to roll into the same PR while touching the files.

- **Stale comment**, `_program.pyx:943`: `# Keys here should match _utils.enum.CompilerBackend` references a module path that does not exist (only `_utils/enum_explanations_helpers.py` is in the tree). Either move `CompilerBackend` (and likely `CodeType`/`SourceType`) into a shared module to also remove the lazy import in `Linker.backend`, or drop the comment.
- **Duplicated import shim**: the `try: from enum import StrEnum except ImportError: from backports.strenum import StrEnum` block is copy-pasted in five files. Hoist to a single helper (e.g. `cuda.core._utils._strenum`) so the shim can be removed in one place when 3.10 support ends.
- **Dead `__doc__` mutation**, `_program.pyx:69`: `SourceType.CXX.__doc__ = "C++ source code."` is fragile (some `Enum` implementations make members read-only) and inconsistent — only one member of one enum gets a doc string. Drop it and document the `"c++"` value in the class docstring instead.
- **Tuple → list message change**, `_program.pyx:732`: `supported_code_types = [x.value for x in SourceType]` changes the user-visible error message from `(...)` to `[...]`. The test was patched to a regex that accepts both (`test_program.py:330`), but it's cleaner to keep the tuple form: `tuple(x.value for x in SourceType)`.
- **Stale assert**, `_program.pyx:733`: `assert code_type not in supported_code_types` is trivially true at this point and reads as a leftover from earlier code. Remove.
- **Misleading `tuple[ManagedMemoryLocationType, ...]` annotation** (post-fix to issue 1): once the typo is corrected, the property still returns plain strings (because `_resolve_preferred_location` stores plain `"device"` etc. into `_pref_loc_type`). Either weaken the annotation back to `tuple[str, int | None] | None` or promote `_pref_loc_type` to actually hold the enum.

---

## Test Gaps

Worth adding while in this PR; cheap and they catch the regression that the blocking-issue fixes are guarding against.

- `assert isinstance(program.backend, CompilerBackend)` somewhere in `test_program.py`. Currently every test only checks string equality, so a future regression that silently re-introduces `str(...)` would not be caught.
- Same for `program.pch_status` returning `PchStatus`.
- Parametrized coverage of `SourceType.PTX` and `SourceType.NVVM` in `Program.__init__` (only `SourceType.CXX` is exercised in `test_launch_invalid_values`).
- `ProgramOptions.as_bytes(CompilerBackend.NVRTC)` (the function does `backend.lower()`; works, but not currently tested with enum input).
- For `GraphAllocOptions.memory_type`: parametrize `_build_alloc_managed_node` to also exercise enum input for `DEVICE` and `HOST`, plus string input for all three, so both halves of the "either is OK" promise are covered.

---

## Suggested Follow-on PRs

The two items below are direct extensions of this work. They are scoped cleanly and would be easier to review separately than to fold in here.

### Follow-on 1: `ConditionalNode.cond_type` enum

**Background.** Issue #1995 explicitly lists `ConditionalNode.cond_type` (returning `"if"`, `"while"`, `"switch"`) as a string-vocabulary candidate for an enum, alongside everything PR 2016 already addresses. The current implementation lives at `cuda_core/cuda/core/graph/_subclasses.pyx:701` and still returns plain `str`. It is not touched by PR 2016. Since #1995 is P0 on the v1.0.0 milestone (due 2026-05-07), closing this gap before 1.0.0 keeps the issue closable.

**Suggested follow-on PR.**

- Add a `ConditionalType(StrEnum)` next to the other graph enums in `cuda.core.graph._graph_definition` (or a new `_conditional.pyx` if you prefer to keep `_subclasses.pyx` lean).
- Members: `IF = "if"`, `WHILE = "while"`, `SWITCH = "switch"`. (None of these are valid Python identifiers, so the `NAME = "value"` pattern matches what `SourceType.CXX = "c++"` already does in this PR.)
- Update `_subclasses.pyx` to return `ConditionalType` instead of `str` from `ConditionalNode.cond_type`.
- Add it to `cuda.core.graph.__all__` and re-export from `cuda.core`.
- Add an isinstance test mirroring whatever pattern the team picks for `Program.backend` in this PR.

This follow-on is small (~30 lines + tests) and directly closes the remaining #1995 surface.

### Follow-on 2: bindings-vs-core enum sync test

**Background.** PR 2014's review surfaced a structural risk that wrappers in `cuda_core` can drift from auto-generated `cuda_bindings` enums when NVML adds new values. The flip side applies to PR 2016: most of the new `StrEnum`s here wrap **driver-level** C enums (`CUmemAllocationType`, `CUmemAllocationHandleType`, `CUmemLocationType`, `CUmemAllocationGranularity_flags`, `CUmemAccess_flags`, `nvrtcResult`, …) selectively, not exhaustively. When the CUDA driver introduces a new value (e.g. a future `CU_MEM_HANDLE_TYPE_*`), the wrapper `StrEnum` will silently fall behind and a user calling, say, `VirtualMemoryResourceOptions(handle_type=...)` with the new value would have to bypass the enum to use it — defeating the type-safety goal.

This is not a blocker for v1.0.0 (today's enums match today's bindings), but it is the kind of debt that compounds across CTK releases.

**Suggested follow-on PR.** Add a parametrized sync test (e.g. `cuda_core/tests/test_enum_sync.py`) that imports the relevant `cuda.bindings.driver` / `cuda.bindings.nvrtc` enums and asserts the wrapper `StrEnum` value-set is a known subset, with a small allow-list for intentional omissions:

```python
@pytest.mark.parametrize("wrapper, source, omitted", [
    (VirtualMemoryAllocationType, cydriver.CUmemAllocationType,
     {"CU_MEM_ALLOCATION_TYPE_INVALID", "CU_MEM_ALLOCATION_TYPE_MAX"}),
    (VirtualMemoryHandleType, cydriver.CUmemAllocationHandleType,
     {"CU_MEM_HANDLE_TYPE_NONE", "CU_MEM_HANDLE_TYPE_MAX"}),
    (VirtualMemoryLocationType, cydriver.CUmemLocationType,
     {"CU_MEM_LOCATION_TYPE_INVALID", "CU_MEM_LOCATION_TYPE_NONE",
      "CU_MEM_LOCATION_TYPE_MAX"}),
    # ...etc.
])
def test_strenum_covers_bindings(wrapper, source, omitted):
    bindings_values = {m.name for m in source} - omitted
    wrapper_values = {m.name.upper() for m in wrapper}  # or a normalization fn
    missing = bindings_values - wrapper_values
    assert not missing, f"{wrapper.__name__} is missing: {missing}"
```

The test fails on the cuda-bindings update PR that adds a new driver value, prompting an explicit decision (extend the wrapper, or extend the `omitted` set with a comment). This is exactly the kind of thing that would have caught the gap quietly during the 12.x → 13.x bindings refresh.

A version of this test should also be applied to the NVML wrappers introduced in PR 2014, so the sync guard covers both halves of the v1.0.0 enum work.

---

## Notes on Items NOT Addressed in This Review

- **`MemoryType` unification from #1995.** Issue 1995 proposed a single shared `MemoryType` enum for `device`/`host`/`managed`. PR 2016 splits it into three (`GraphMemoryType`, `ManagedMemoryLocationType`, `VirtualMemoryLocationType`) — reasonable for documentation provenance, but worth a one-line note in the PR description so reviewers don't expect the unified form. (Could also be revisited in a future PR by deriving them from a shared base or making them aliases.)

mdboom:
> 3. New public enums are documented under api_private.rst
cuda_core/docs/source/api_private.rst:24-34

> All eleven new enums (CodeType, CompilerBackend, GraphMemoryType, ManagedMemoryLocationType, PchStatus, SourceType, VirtualMemory*Type) are listed in api_private.rst, which carries :orphan: and is described in its own header comment as "for private classes exposed to users". These enums are reachable as cuda.core.CodeType etc., live in cuda.core.__all__, and are named in user-facing docstrings — they are public API.

This is to address @leofang's concern offline about "class explosion" from using enums.  By documenting them as linked from functions that use the enums, and not putting them at the top-level table of contents, I think this helps quite a bit with that concern.  If anything the `api_private.rst` (which is just a hack to document things not in the table of contents) is now mis-named and maybe we address that separately.

> Duplicated import shim: the try: from enum import StrEnum except ImportError: from backports.strenum import StrEnum block is copy-pasted in five files. Hoist to a single helper (e.g. cuda.core._utils._strenum) so the shim can be removed in one place when 3.10 support ends.

Let's separate that out.  Not required for 1.0

https://github.com/NVIDIA/cuda-python/issues/2019

> Dead __doc__ mutation, _program.pyx:69: SourceType.CXX.__doc__ = "C++ source code." is fragile (some Enum implementations make members read-only) and inconsistent — only one member of one enum gets a doc string. Drop it and document the "c++" value in the class docstring instead.

That suggestion does not work with our current Sphinx setup, and it is also not as ergonomic at the terminal or Jupyter.

## Sync tests

As with #2014, I'll experiment with what sync testing would look like, but it is not critical for 1.0

cuda_core/cuda/core/_program.pyx
```diff
@@ -130,12 +172,12 @@ cdef class Program:
            use the NVRTC backend. For PTX and NVVM programs this property
            always returns ``None``.
         """
-        return self._pch_status
+        return PchStatus(self._pch_status)
```
rwgk:
```suggestion
        if self._pch_status is None:
            return None
        return PchStatus(self._pch_status)
```

cuda_core/cuda/core/_memory/_managed_memory_resource.pyx
```diff
@@ -109,7 +119,7 @@ cdef class ManagedMemoryResource(_MemPool):
             return None
         if self._pref_loc_type == "host":
             return ("host", None)
```
rwgk:
```suggestion
            return (ManagedMemoryLocationType.HOST, None)
```

rwgk (commented):
Only two tiny fixes. The rest looks good to me.

cuda_core/cuda/core/graph/_conditional_type.py
```diff
@@ -0,0 +1,18 @@
...
+    from enum import StrEnum
+except ImportError:
+    from backports.strenum import StrEnum
+
+__all__ = ["ConditionalType"]
+
+
+# Since this enum uses `IF`, which is a reserved keyword in Cython, we have to
+# define it in a separate Python file just for this purpose.
+class ConditionalType(StrEnum):
```
leofang: Should we name this `GraphConditionalType`?

cuda_core/cuda/core/_module.pyx
```diff
@@ -569,7 +570,7 @@ cdef class Kernel:
 
 CodeTypeT = bytes | bytearray | str
 
-cdef tuple _supported_code_type = ("cubin", "ptx", "ltoir", "fatbin", "object", "library")
+cdef tuple _supported_code_type = tuple(CodeType.__members__.values())
```
leofang: Ah, when reading this side-by-side it is easy to catch: We have both `CodeTypeT` and `CodeType` meaning two very different things, we probably should fix this

cuda_core/cuda/core/_program.pyx
```diff
@@ -49,6 +60,38 @@ The ``int`` type covers NVVM handles, which don't have a wrapper class.
...
+
+
+class CompilerBackend(StrEnum):
+    NVRTC = "NVRTC"
+    NVVM = "NVVM"
+    NVJITLINK = "nvJitLink"
+    DRIVER = "driver"
+
+
+class PchStatus(StrEnum):
```
leofang: Pch -> PCH

cuda_core/cuda/core/__init__.py
```diff
@@ -41,11 +41,17 @@ def _import_versioned_module():
...
     ManagedMemoryResource,
     ManagedMemoryResourceOptions,
     MemoryResource,
     PinnedMemoryResource,
     PinnedMemoryResourceOptions,
+    VirtualMemoryAccessType,
+    VirtualMemoryAllocationType,
+    VirtualMemoryGranularityType,
+    VirtualMemoryHandleType,
+    VirtualMemoryLocationType,
```
leofang: These `*Type` are where I backed out last time. I think they should really go to `cuda.core.typings`. 

FWIW, at least for location, @Andy-Jost and I are pushing for the addition of a `Host` object (#1775). So this might change if we land that PR.

cuda_core/cuda/core/__init__.py
```diff
@@ -63,11 +76,13 @@ def _import_versioned_module():
 )
 from cuda.core._tensor_map import TensorMapDescriptor, TensorMapDescriptorOptions
 from cuda.core.graph import (
+    ConditionalType,
```
leofang: ditto, GraphConditionalType?

leofang: (also moves to `cuda.core.typings`)

leofang (changes requested):

cuda_core/cuda/core/_module.pyx
mdboom: Would `CodeFormat` (for the enum with "cubin", "ptx" etc.) make sense?

cuda_core/cuda/core/__init__.py
mdboom: Do you think that's where we should put all of the enums?  I sort of put them "near use" here, but even that got a bit messy because some of them are used in multiple modules etc.

cuda_core/cuda/core/_module.pyx
leofang: Maybe `ObjectCodeFormatT`? The `ObjectCode` class can be constructed from any supported format.

Another naming convention discussion: Our type annotations have been made a bit C++-ish (but consistently) by appending a suffix T: `IsStreamT`, `CodeTypeT`, .... Maybe a good time to decide a convention and stick to it. My only concern (as raised elsewhere) is we might have both actual classes and associated type annotations refer to overlapping or orthogonal concepts. Would be nice to avoid ambiguity.

cuda_core/cuda/core/graph/_conditional_type.py
leofang: Or `GraphConditionlT`, see https://github.com/NVIDIA/cuda-python/pull/2016#discussion_r3189031176.

cuda_core/cuda/core/_module.pyx
mdboom: In think in this case, enums are not type annotations (they have members and runtime semantics and such), so shouldn't have the `T`.  I know we are putting them in `typing`.

cuda_core/cuda/core/__init__.py
mdboom: (Offline discussion): We are going to put all the enums in `cuda.core.typing`.

mdboom:
This got a pretty significant rework based on @leofang's feedback.  All the new enums are defined in `cuda.core.typing` now.  The suggested renames of the enums was applied.  And this ties into the testing added in #2014 that ensures that new enum values added upstream will be detected.  There were a few bugfixes required to that new testing infrastructure to make work, which then revealed a couple small bugs in #2014, fixed here.

cuda_core/tests/graph/test_graph_definition.py
leofang, line 1: Q: As part of the move to use StrEnum, I'd assume that all existing tests should continue to work without any change?

cuda_core/tests/test_launcher.py
leofang, line 1: ditto

cuda_core/cuda/core/typing.py
```diff
@@ -2,12 +2,103 @@
...
 from cuda.core._stream import IsStreamT
 
 __all__ = [
+    "CompilerBackend",
     "DevicePointerT",
+    "GraphConditionalType",
+    "GraphMemoryType",
     "IsStreamT",
+    "ManagedMemoryLocationType",
+    "ObjectCodeFormat",
```
leofang: ObjectCodeFormat**Type**?

cuda_core/cuda/core/typing.py
```diff
@@ -2,12 +2,103 @@
...
+    "PCHStatus",
```
leofang: PCHStatus**Type**?

cuda_core/cuda/core/typing.py
```diff
@@ -2,12 +2,103 @@
...
+    "SourceType",
```
leofang: CodeSourceType or SourceCodeType?

cuda_core/cuda/core/typing.py
```diff
@@ -2,12 +2,103 @@
...
+    from enum import StrEnum
+except ImportError:
+    from backports.strenum import StrEnum
 
 from cuda.core._memory._buffer import DevicePointerT
 from cuda.core._stream import IsStreamT
 
 __all__ = [
+    "CompilerBackend",
     "DevicePointerT",
```
leofang: DevicePointerType?

cuda_core/cuda/core/typing.py
```diff
@@ -2,12 +2,103 @@
...
+    "GraphConditionalType",
+    "GraphMemoryType",
     "IsStreamT",
```
leofang: IsStreamType or StreamType?

leofang: (ps: it's actually a protocol, not a type annotation)

cuda_core/cuda/core/typing.py
```diff
@@ -2,12 +2,103 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-"""Public type aliases and protocols used in cuda.core API signatures."""
+"""Public type aliases, protocols, and enumerations used in cuda.core API signatures."""
+
+try:
+    from enum import StrEnum
+except ImportError:
+    from backports.strenum import StrEnum
```
leofang, line 10: del StrEnum at the end of the file?

leofang (commented):
LGTM overall! Left a bunch of minor comments mainly for consistency.

cuda_core/tests/graph/test_graph_definition.py
mdboom: Yes, I think so.  The test changes here are only to test the /new/ code path of passing in enums, but the old path of passing in strings still works.

cuda_core/cuda/core/typing.py
leofang: Sorry, this can be ignored, I was being dumb and forgot that we already had `__all__`.

mdboom:
> LGTM overall! Left a bunch of minor comments mainly for consistency.

Ok.  I renamed everything so it has a consistent `Type` suffix for all of these type-annotation-like things.

leofang (approved):
We will need a release note entry, feel free to add it in this PR or the next one. Approving.

Issue #1950, opened 2026-04-17
cuda.core: public API class naming audit (PEP 8 and consistency)


mdboom:
@mdboom's commentary:  This all seems pretty low stakes, but now is the time to address any inconsistencies before 1.0.0.

## Summary

An audit of all public class names in `cuda.core` found three categories of naming inconsistency: a terminology mismatch ("Config" vs "Options"), one unusual abbreviation, and several PEP 8 acronym-capitalization violations in the `cuda.core.system` submodule.

### Original prompt

> Can you please look at all classes exposed in the public API of cuda_core. I want you to look just at the names of the classes, not any members or methods. Please make sure they are all PEP8 compliant and that the same words are used for the same concepts.

---

### 1. "Config" vs "Options" inconsistency

`LaunchConfig` is the only class that uses "Config" for a configuration/options object. Every other such class uses "Options":

| Pattern-breaker | Consistent peers |
|---|---|
| `LaunchConfig` | `EventOptions`, `LinkerOptions`, `StreamOptions`, `ProgramOptions`, `DeviceMemoryResourceOptions`, `ManagedMemoryResourceOptions`, `PinnedMemoryResourceOptions`, `VirtualMemoryResourceOptions`, `TensorMapDescriptorOptions`, `GraphAllocOptions`, `GraphCompleteOptions`, `GraphDebugPrintOptions` |

**Suggestion:** `LaunchConfig` → `LaunchOptions`

### 2. `GraphDef` abbreviation

`GraphDef` abbreviates "Definition" while the rest of the API spells words out (e.g. `TensorMapDescriptor`, not `TensorMapDesc`; `ConditionalNode`, not `CondNode`). The `AllocNode` / `GraphAllocOptions` abbreviations are defensible since "alloc/free" are standard CUDA/C terms, but `GraphDef` is less conventional.

**Suggestion:** `GraphDef` → `GraphDefinition`

### 3. Acronym capitalization (PEP 8 violation in `cuda.core.system`)

[PEP 8](https://peps.python.org/pep-0008/#descriptive-naming-styles) says: *"When using abbreviations in CapWords, capitalize all the letters of the abbreviation. Thus `HTTPServerError` is better than `HttpServerError`."*

The system module is internally inconsistent — `P2P` and `BAR1` are correctly all-caps, while other acronyms are treated as regular CamelCase words:

| Current name | PEP 8 name | Acronym |
|---|---|---|
[...]

PR #2018, opened 2026-05-04
cuda.core: convert peer_accessible_by to a live MutableSet view

Andy-Jost:
## Summary

`DeviceMemoryResource.peer_accessible_by` previously returned a sorted `tuple[int, ...]` backed by a Python-level cache. This is fragile in two ways: the cache can diverge from driver state across multiple wrappers around the same memory pool (#1720), and tuple semantics force callers into bespoke splice/replace patterns instead of standard set operations. This PR replaces the property with a live driver-backed `collections.abc.MutableSet` view, matching the proxy patterns established by `AdjacencySetProxy` (graphs) and the in-flight `AccessedBySet` (managed-memory advice, #1775).

This is a breaking change, captured in the v1.0.0 release notes.

### New: `PeerAccessibleBySetProxy`

A `MutableSet` subclass living in `cuda_core/cuda/core/_memory/_peer_access_utils.pyx`.

- Reads (`__contains__`, `__iter__`, `__len__`) call `cuMemPoolGetAccess`.
- Writes (`add`, `discard`, and bulk ops) call `cuMemPoolSetAccess`.
- Iteration yields `Device` objects.
- `add`, `discard`, `__contains__` accept `Device | int`.
- The owner device is silently filtered (matching the existing planner contract).
- All bulk operations (`update`, `|=`, `&=`, `-=`, `^=`, `clear`) issue **exactly one** `cuMemPoolSetAccess` call. This matters: peer-access transitions can take seconds per pool because every existing memory mapping is updated, so coalescing into a single driver call lets the toolkit handle the mappings in parallel.

The proxy is constructed fresh on every property access. There is nothing to cache or pickle.

### Cache removal

The `_peer_accessible_by` field on `DeviceMemoryResource` is dropped, along with its initializations in `__cinit__`, `_DMR_init`, and `from_allocation_handle`. The owned/non-owned read split is gone — every read path now queries the driver directly. This eliminates the bug class fixed in #1720 and simplifies the implementation.

### Module consolidation

All peer-access internals now live in `_peer_access_utils.pyx` (promoted from a `.py`). The `cdef inline` driver helpers (`_query_peer_access_ids`, `_peer_access_includes`, `_set_pool_access`) and the `cpdef replace_peer_accessible_by` setter helper moved out of `_device_memory_resource.pyx`. `DeviceMemoryResource` now exposes only the public `peer_accessible_by` property: the getter returns `PeerAccessibleBySetProxy(self)` and the setter delegates to `replace_peer_accessible_by(self, devices)`. This keeps the DMR surface focused on its memory-management role.

### Property setter

`mr.peer_accessible_by = [...]` is preserved and unchanged in behavior. It still does a single batched `cuMemPoolSetAccess` via the same shared `plan_peer_access_update` path the proxy uses for bulk ops.

### `_query_peer_access_ids` GIL optimization

The driver loop that enumerates peer device IDs now runs inside a single `nogil` block instead of acquiring/releasing the GIL once per device. Per-call data is collected into a `libcpp.vector[int]`, and because `range(total)` ascends the result is already sorted (so the trailing `sorted()` is gone). This is a local performance tweak with no behavior change.

## Test coverage

The existing integration tests (`test_memory_peer_access.py`, `memory_ipc/test_peer_access.py`) migrated from tuple equality to set-of-`Device` equality. Eight new tests pin the proxy's contract end-to-end, all using the existing `mempool_device_x2` fixture so they run on CI:

- **MutableSet conformance.** A new `assert_single_member_mutable_set_interface(subject, member, non_member)` helper exercises every `MutableSet` method against subjects whose backing store admits at most one insertable element (the peer-access proxy can hold at most `{dev1}` on a 2-GPU box). The original `assert_mutable_set_interface(subject, items)` is unchanged for normal multi-element subjects (`AdjacencySetProxy`).
- **`Device`/`int` interchangeability** on `add`/`discard`/`__contains__`.
- **Owner-filtering contract** on every write path (silent no-op).
- **Error paths**: `add(out_of_range)` and `add(non_coercible)` raise; `discard`/`__contains__` swallow the same inputs; `remove(non_member)` raises `KeyError`.
- **Live driver view**: a proxy obtained before another wrapper modifies the pool reflects the change with no refresh step.
- **Iteration order** is ascending by `device_id`; elements are `Device` instances; `__repr__` shape; getter return type.
- **Single-call batching** via a `monkeypatch` spy on `_set_pool_access` (the thin Python-visible helper extracted from `_apply_peer_access_diff` purely so tests can spy on the actual driver call). Every bulk op (`|=`/`&=`/`-=`/`^=`/`update`/`difference_update`/`clear`) and the property setter is asserted to issue exactly one `cuMemPoolSetAccess`, zero on no-op. This includes the `dmr.peer_accessible_by |= {...}` augmented-assignment-on-property pattern, where the empty-diff short-circuit prevents a second driver call from the trailing setter write.

## Breaking change

`DeviceMemoryResource.peer_accessible_by` no longer returns a `tuple[int, ...]`. Callers must update:

- Tuple comparisons (`mr.peer_accessible_by == (0,)`) -> set comparisons (`mr.peer_accessible_by == {Device(0)}`).
- Iteration that expected ints -> iteration over `Device` objects (use `[d.device_id for d in mr.peer_accessible_by]` if int IDs are needed).

Setter usage (`mr.peer_accessible_by = [...]`) is unchanged.

## Test plan

- [x] Existing `test_memory_peer_access.py` tests pass after migration to set semantics.
- [x] Existing `test_memory_peer_access_utils.py` tests still pass (`plan_peer_access_update` and `normalize_peer_access_targets` are unchanged).
- [x] Existing `memory_ipc/test_peer_access.py` tests pass (IPC-enabled pools, parent-child IPC scenarios).
- [x] Full `cuda_core` test suite passes on a single-GPU box (2931 passed, 212 skipped — peer-access tests gated on `mempool_device_x2`/`x3`).
- [x] All new peer-access tests pass on a 2-GPU box.
- [x] `pre-commit run --all-files` passes (ruff, format, SPDX, cython-lint, RST checks).

## Related work

- #1720 — root cause of the cache-divergence bug; this PR removes the cache entirely.
- #1775 — Rob Parolin's managed-memory `AccessedBySet`, the second driver-backed set proxy in `cuda.core`.
- `cuda_core/cuda/core/graph/_graph_node.pyx` (`AdjacencySetProxy`) — reference pattern for driver-backed set proxies. Same `MutableSet` shape, same `_from_iterable` plain-set fallback, same single-driver-call bulk-op contract.


leofang:
FYI: A breaking change must be labeled with P0.

cuda_core/tests/helpers/collection_interface_testers.py
```diff
@@ -136,7 +162,157 @@ def assert_mutable_set_interface(subject, items):
...
+def assert_single_member_mutable_set_interface(subject, member, non_member):
+    """Exercise every MutableSet method on a subject with capacity one.
+
+    Use this for proxies whose backing store admits at most one insertable
+    element at a time (typically because the underlying resource is bounded
+    by hardware, e.g. a peer-access view on a system with a single valid
+    peer device). The subject only ever holds ``set()`` or ``{member}``;
+    *non_member* supplies the right-hand side of comparisons, ``isdisjoint``,
+    subset/superset, and binary/in-place operators so every ``MutableSet``
+    method is exercised at least once.
```
Andy-Jost, line 177: @rparolin you might be able to use this to test `AccessedBySetProxy` in #1775.

leofang:
Q: It is a big diff with a very niche use case. Are we sure this is P0?

Andy-Jost:
> Q: It is a big diff with a very niche use case. Are we sure this is P0?

The peer access list is currently a tuple. This change aligns with the set proxies in the graph module (`AdjacencySetProxy`) and the managed memory hint functions (`AccessedBySetProxy`). Without this, users would have to remember which interfaces uses a proxy and which do not. I think as a matter of course we should favor container proxies; they are more Pythonic.

This change is not functionally as big as it might appear. (1) Most of the bulk is just code movement: I moved the peer access implementation from `_device_memory_resource.pyx` to `_peer_access_utils.pyx` just to keep the DMR implementation clean. (2) I had to change `_peer_access_utils` from a `.py` to a `.pyx` file to accommodate the moved code, and git didn't identify it as a move/rename.

cuda_core/cuda/core/_memory/_device_memory_resource.pyx
```diff
@@ -217,30 +213,23 @@ cdef class DeviceMemoryResource(_MemPool):
...
         Examples
         --------
         >>> dmr = DeviceMemoryResource(0)
-        >>> dmr.peer_accessible_by = [1]  # Grant access to device 1
-        >>> assert dmr.peer_accessible_by == (1,)
-        >>> dmr.peer_accessible_by = []  # Revoke access
+        >>> dmr.peer_accessible_by = {1}   # grant access to device 1
+        >>> assert 1 in dmr.peer_accessible_by
+        >>> dmr.peer_accessible_by.add(2)  # update access to include device 2
+        >>> dmr.peer_accessible_by = []    # revoke peer access
```
mdboom: Not strictly necessary, but for the sake of consistency in the example?

```suggestion
        >>> dmr.peer_accessible_by = {}    # revoke peer access
```

leofang (commented):
(Disregard — test review from debugging API.)

cuda_core/cuda/core/_memory/_peer_access_utils.pyx
```diff
@@ -0,0 +1,392 @@
...
+        # _from_iterable. Returning a plain set lets the user reason about
+        # the result independently of any pool's driver state.
+        return set(it)
+
+    # --- abstract MutableSet methods ---
+
+    def __contains__(self, value) -> bool:
+        try:
+            dev_id = _resolve_peer_device_id(value)
+        except (TypeError, ValueError):
```
leofang, line 238: **Critical** — `__contains__` can raise on out-of-range int, violating Python convention.

`_resolve_peer_device_id` calls `Device(value)`. For out-of-range ints, `Device()` may raise `CUDAError` (not `ValueError` or `TypeError`), which isn't caught here. So `999 in proxy` would propagate a `CUDAError` instead of returning `False`.

Python's `__contains__` contract is that it never raises — callers rely on `in` being a safe, boolean predicate. The same gap exists in `discard` (line 261) and the bulk-op resolution loops (`difference_update`, `intersection_update`, `symmetric_difference_update`).

Suggest either:
- catching `CUDAError` here too (and in the other resolution try/except blocks), or
- having `Device()` normalize out-of-range ordinals to `ValueError` upstream.

The existing test (`test_peer_accessible_by_rejects_invalid_inputs`) covers `"not-a-device" not in proxy` (TypeError, caught) but not `bad_id in proxy` (CUDAError, uncaught).

cuda_core/cuda/core/_memory/_peer_access_utils.pyx
```diff
@@ -0,0 +1,392 @@
...
+        """Grant peer access from ``value`` to allocations in this pool."""
+        self._apply([value], ())
+
+    def discard(self, value) -> None:
+        """Revoke peer access from ``value`` to allocations in this pool."""
+        try:
+            dev_id = _resolve_peer_device_id(value)
+        except (TypeError, ValueError):
+            return
+        self._apply((), [dev_id])
```
leofang: **Consideration** — double resolution in `discard`.

`discard` resolves `value` to `dev_id` (int) here, then passes `[dev_id]` into `_apply`, which iterates `removals` and calls `_resolve_peer_device_id(value)` again on each element (line 381). That second call does `Device(int).device_id` — an extra `Device()` construction per element.

For a single-element `discard` this is one redundant `Device()` allocation. Not wrong, but since cuda.core cares about O(100 ns)-level overhead, you could either:
- have `_apply` accept already-resolved ints for removals (a separate parameter or a sentinel), or
- pass the raw `value` through and let `_apply` do the only resolution.

cuda_core/cuda/core/_memory/_peer_access_utils.pyx
```diff
@@ -0,0 +1,392 @@
...
+        only the owner device is filtered out. Adds are validated through
+        :meth:`Device.can_access_peer` via :func:`plan_peer_access_update`;
+        removals bypass that check (revoking is always permitted).
+        """
+        from cuda.core._device import Device
+
+        cdef DeviceMemoryResource mr = <DeviceMemoryResource>self._mr
+        owner_id = mr._dev_id
+        owner = Device(owner_id)
+        current = _query_peer_access_ids(mr)
```
leofang, line 376: **Consideration** — every `add`/`discard` does a full device scan.

`_apply` unconditionally calls `_query_peer_access_ids(mr)` here, which loops over all devices with one `cuMemPoolGetAccess` per device. For `add(dev1)`, a cheaper path would be:
1. `_peer_access_includes(mr, dev_id)` — already granted? return.
2. validate `can_access_peer` for the single device.
3. `_set_pool_access(mr, (dev_id,), ())`.

That's 1 driver call for the common "already granted" case vs. N+1. For `discard`, similarly check `_peer_access_includes` first.

The current approach is correct and consistent (reuses the planner, validates `can_access_peer`), so this is about whether the per-device scan matters in practice. On a system with many GPUs (DGX-H100: 8 GPUs = 7 `cuMemPoolGetAccess` calls per `add`), it could add up.

cuda_core/tests/test_memory_peer_access.py
```diff
@@ -122,18 +125,287 @@ def test_peer_access_shared_pool_queries_driver(mempool_device_x2):
...
+    assert dmr3.peer_accessible_by == set()
+
+
+# ---------------------------------------------------------------------------
+# Set-proxy interface coverage
+#
+# These tests exercise the ``PeerAccessibleBySetProxy`` surface added in
+# v1.0.0. They run against ``mempool_device_x2`` because every CI machine has
+# at most 2 GPUs, which means at most one valid peer device. The relaxed
+# ``support_multi_insert=False`` path on ``assert_mutable_set_interface``
```
leofang: **Nitpick** — stale comment. This mentions `support_multi_insert=False` path on `assert_mutable_set_interface`, but the actual code below calls `assert_single_member_mutable_set_interface` — a separate function, not a flag. Looks like a leftover from an earlier iteration.

cuda_core/docs/source/release/1.0.0-notes.rst
```diff
@@ -120,6 +120,10 @@ Breaking changes
     ``CUgraphConditionalHandle`` value. Previously, ``.handle`` had to be
     extracted explicitly.
 
+- :attr:`DeviceMemoryResource.peer_accessible_by` now returns a
```
leofang, line 123: **Nitpick** — other breaking-change entries link to their PR/issue (e.g. `#1950`, `#2016`). This one could link to #2018 for consistency.

leofang (commented):
Overall well-structured PR. Cache elimination is the right call, proxy pattern is consistent with `AdjacencySetProxy` / `AccessedBySetProxy`, and test coverage is thorough. Comments tagged Critical / Consideration / Nitpick.

leofang:
@Andy-Jost there are merge conflicts that need to be resolved

cuda_core/tests/helpers/collection_interface_testers.py
leofang: Let's get this PR merged and apply it in #1775 (assuming we still have time).

cuda_core/cuda/core/_memory/_peer_access_utils.pyx
Andy-Jost: My 2c is this level of optimization is not needed these APIs. Peer access changes iterate through all memory mappings and can take on the order of seconds to complete. I'd favor simpler code here.

cuda_core/cuda/core/_memory/_peer_access_utils.pyx
Andy-Jost: `Device(bad_id)` raises `ValueError` at `cuda/core/_device.pyx:984`. I will add a test to confirm in the next commit.

Andy-Jost:
commit `3233538afb` addresses:                                                                                                                                                                                                                                                                                                            
- Release notes: added `#2018` link to the `peer_accessible_by` entry for consistency with other breaking-change items                                                                                                                                                                                                                     
- `__contains__` with out-of-range int: investigated and confirmed not a bug — `Device(bad_id)` raises `ValueError` (not `CUDAError`) at `_device.pyx:984`, which the existing `except (TypeError, ValueError)` catches correctly. Added a test for `bad_id not in proxy` to lock this in.
- Stale comment in `test_memory_peer_access.py`: updated to reference `assert_single_member_mutable_set_interface` instead of the removed `support_multi_insert=False` flag                                                                                                                                                                
- `add`/`discard` full device scan: optimized both to use a single `_peer_access_includes` call (1 `cuMemPoolGetAccess`) instead of scanning all devices via `_query_peer_access_ids` (N calls). Bulk ops are unchanged — they already need the full set for diff computation.

cuda_core/cuda/core/_memory/_peer_access_utils.pyx
leofang: I think it's the problem of going through the driver API... We need to discuss some form of reverting post-1.0. CUDA driver APIs aren't friendly for peer access queries, and this query has a real cost in the user apps.

leofang (approved):
Approving with a caveat (https://github.com/NVIDIA/cuda-python/pull/2018#discussion_r3197700023).

cuda_core/cuda/core/_memory/_peer_access_utils.pyx
Andy-Jost: It looks to me like `Device(int)` does not call any driver APIs though?

leofang: You're right that `Device(int)` itself doesn't call driver APIs — that part is just Python object allocation.

My concern was about the broader pattern, not the `Device()` construction specifically. With the cache removed, every read and write path on the proxy now goes through `_query_peer_access_ids`, which calls `cuMemPoolGetAccess` once per device. So even a simple `dev1 in proxy` triggers `_peer_access_includes` (1 driver call), and `len(proxy)` or `list(proxy)` triggers N calls. An equality check like `proxy == {dev1}` calls both `__len__` and `__iter__` under the hood, so that's 2N driver calls for a single `==`.

That's the "going through the driver API" cost I was referring to — the no-cache design means we pay driver roundtrips on every access, not just on writes.

Andy-Jost: Maybe we're on the same page then. I updated add/remove to only call the driver APIs for the device in question, rather than all devices. I 100% agree with that direction.

I'm of two minds about using attribute caches in general due to the added complexity and risk of becoming desync'd. I'm coming around the approach that we should use proxy containers without caches at first, and then optimize if and when we have performance data. It's definitely a grey area. For peer access specifically, I'd be surprised if the difference matters.

Issue #1989, opened 2026-04-29
Inconsistent lifetime management across `from_handle` / `from_*` APIs


leofang:
## Summary

Audit of all public `from_handle` / `from_*` methods in `cuda.core` reveals several inconsistencies in how lifetime management is exposed to users. These should be resolved before the 1.0 stable release.

## Inventory

| Class | Method | Ownership Model |
|---|---|---|
| `Stream` | `from_handle(handle)` | Always non-owning (borrowed) |
| `Kernel` | `from_handle(handle, mod=)` | Non-owning ref; optional `_keepalive` holds `ObjectCode` |
| `Buffer` | `from_handle(ptr, size, mr=, owner=)` | 3 modes: non-owning / mr-deallocates / owner-ref |
| `Event` | `from_ipc_descriptor(...)` | Owns imported event |
| `Buffer` | `from_ipc_descriptor(...)` | Owns imported buffer |
| `GraphicsResource` | `from_gl_buffer(...)` / `from_gl_image(...)` | Owns GL registration |
| `DeviceMemoryResource` | `from_allocation_handle(...)` / `from_registry(uuid)` | Owns imported pool / retrieves existing |
| `PinnedMemoryResource` | `from_allocation_handle(...)` / `from_registry(uuid)` | Owns imported pool / retrieves existing |
| `ObjectCode` | `from_cubin/ptx/ltoir/fatbin/object/library()` | Owns module data |
| `StridedMemoryView` | `from_dlpack/cai/array_interface/buffer/any_interface()` | Always non-owning (view) |

### 1. `Stream.from_handle` cannot opt into keeping an owner alive

- `Buffer.from_handle` accepts `owner=` to hold a Python reference and prevent GC of the foreign allocator.
- `Kernel.from_handle` accepts `mod=` to hold a reference and prevent GC of the parent `ObjectCode`/library.
- `Stream.from_handle` has **no equivalent parameter**. It creates a borrowed stream with no way for the user to tie lifetime to an owner object.

The internal machinery already supports this — `create_stream_handle_with_owner()` in C++ accepts a Python owner and does `Py_INCREF/DECREF`. But `from_handle()` only passes the ephemeral `_stream_holder` wrapper as the owner, not the user's actual owner object.

**Suggestion:** Add an `owner` parameter to `Stream.from_handle`, consistent with `Buffer.from_handle`.

### 2. `Event` has no public `from_handle()`

[...]

PR #2020, opened 2026-05-04
cuda.core: require explicit stream for stream-scheduling APIs

### LINKED ISSUE #2001: cuda.core: reject stream=None — require explicit stream everywhere
## Summary

Several `cuda.core` APIs accept `stream=None` and silently fall back to `default_stream()` (or NULL). This makes the stream choice implicit and environment-dependent (`CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM`), which is error-prone. Users should always pass a stream explicitly — including `device.default_stream` when they want the default.

## Design rule

`stream` should be a **keyword-only argument with no default value** (`*, stream: Stream`), so callers must always pass a stream explicitly.

APIs that already follow this convention:
- `Buffer.copy_to(dst=None, *, stream: Stream | GraphBuilder)`
- `Buffer.copy_from(src, *, stream: Stream | GraphBuilder)`
- `Buffer.fill(value, *, stream: Stream | GraphBuilder)`

### `stream=None` → `default_stream()` (implicit fallback, remove it)

- `MemoryPool.allocate()` / `deallocate()`
- `GraphMemoryResource.allocate()` / `deallocate()`
- `GraphicsResource.map()`
- `LegacyPinnedMemoryResource.allocate()`
- `_SynchronousMemoryResource.allocate()`
- `Device.allocate()` (delegates to the above)

### `stream=None` → NULL / legacy default (same issue)

- `Kernel.max_potential_cluster_size(config, stream=None)`
- `Kernel.max_active_clusters(config, stream=None)`

### `stream` should become keyword-only

- `GraphBuilder.launch(self, stream: Stream)` — currently positional, should be `(self, *, stream: Stream)`

## APIs that are fine (no change needed)

- `launch(stream, config, kernel, *args)` — stream is the 1st positional arg by design
- `Buffer.close(stream=None)` — None means "reuse the original stream", by design
- `GraphicsResource.unmap(stream=None)` / `close(stream=None)` — same, reuses the mapping stream

## Proposed implementation

1. Have `Stream_accept()` raise `TypeError` when `stream is None`, so the check is centralized and cannot be forgotten.
2. Remove the `= None` default from every affected API signature.
3. Make `stream` keyword-only where it isn't already (except `launch()` where it's intentionally positional).

## Prior art: CCCL's cccl-runtime

CCCL's modern CUDA runtime (libcudacxx, `cuda::` namespace) follows the same explicit-stream principle:

> The CUDA default (NULL) stream is not exposed as a first-class runtime object because it is tied to implicit per-device state and encourages hidden dependencies.
>
> — `docs/libcudacxx/runtime/cudart_interactions.rst`

Concretely:
- `cuda::launch(stream, config, kernel, args...)` requires an explicit `stream_ref` as the first parameter, with no default.
- The `cuda::mr::resource` concept requires `allocate(cuda::stream_ref, ...)` / `deallocate(cuda::stream_ref, ...)` — no fallback to a default stream.
- `cuda::stream_ref` deletes constructors from `int` and `nullptr`, and deprecates the parameterless (default stream) constructor.

Andy-Jost:
Closes #2001.

## Summary

Removes the implicit fallback to `default_stream()` (or NULL) on every `cuda.core` API that schedules work on a stream. `stream` is now a required keyword-only argument; `Stream_accept(None)` raises `TypeError`. Callers that want the previous behavior must pass `device.default_stream` (or any explicit `Stream`) at the call site.

## Changes

**API surface**

- `MemoryResource.allocate` / `MemoryResource.deallocate` and overrides on `DeviceMemoryResource`, `PinnedMemoryResource`, `ManagedMemoryResource`, `LegacyPinnedMemoryResource`, and `GraphMemoryResource`.
- `Device.allocate`.
- `GraphicsResource.map`.
- `KernelOccupancy.max_potential_cluster_size` / `max_active_clusters`.
- `Buffer.from_ipc_descriptor`.

`deallocate` is keyword-only on every memory-resource subclass. The synchronous resources (`LegacyPinnedMemoryResource`, `_SynchronousMemoryResource`, `VirtualMemoryResource`) keep `stream=None` as a default, since those resources do not actually use the stream.

`Stream_accept` is promoted to `cpdef` so the pure-Python legacy/sync resources can call the centralized validation helper.

**Exemptions (no change)**

- `Graph.launch(stream)` — same shape as `Graph.upload(stream)` and the kernel `launch(stream, config, kernel, *args)` API; stream stays the first positional argument.
- `VirtualMemoryResource.allocate` / `deallocate` — `cuMemCreate` / `cuMemMap` are synchronous and not stream-ordered. Stream is keyword-only but optional and validated when provided.
- `Buffer.close(stream=None)`, `GraphicsResource.unmap(stream=None)`, `GraphicsResource.close(stream=None)` — `None` here means "reuse the existing stream stored in the handle", not "fall back to default" — already on the exemption list in the issue.

## Test Coverage

- `tests/test_module.py`: added `pytest.raises(TypeError, ...)` for `KernelOccupancy.max_potential_cluster_size` / `max_active_clusters` when `stream` is omitted.
- `tests/test_graphics.py`: added `test_map_requires_explicit_stream` (replaces `test_map_with_default_stream`).
- All existing test sites and examples updated to pass `stream=` explicitly.
- Full `cuda_core/tests` suite passes locally on Linux/A10/CUDA 13.1.

## Related Work

- Issue #2001


cuda_core/cuda/core/_memory/_buffer.pyx
```diff
@@ -124,7 +124,11 @@ cdef class Buffer:
 
     @staticmethod
     def _reduce_helper(mr, ipc_descriptor):
-        return Buffer.from_ipc_descriptor(mr, ipc_descriptor)
+        # The parent process's stream is not portable across processes, so the
+        # pickle path cannot thread an explicit stream through. Seed the
+        # imported buffer's deallocation with the current context's default
+        # stream; the receiver can override via buffer.close(stream).
+        return Buffer.from_ipc_descriptor(mr, ipc_descriptor, stream=default_stream())
```
leofang, line 138: This makes perfect sense. I thought we already did this.

cuda_core/cuda/core/_module.pyx
```diff
@@ -367,7 +367,7 @@ cdef class KernelOccupancy:
             ))
         return dynamic_smem_size
 
-    def max_potential_cluster_size(self, config: LaunchConfig, stream: Stream | None = None) -> int:
+    def max_potential_cluster_size(self, config: LaunchConfig, *, stream: Stream) -> int:
```
leofang, line 371: To add to the context: here (and all methods touched in this file) we need to pass a stream because the available hw resources, from which the sizes are computed, could be inferred in the cases of "green" streams (#1976).

cuda_core/examples/graph_update.py
```diff
@@ -63,7 +63,7 @@ def main():
         module = program.compile("cubin")
         kernel = module.get_kernel("add_one")
 
-        buffer = pinned_mr.allocate(2 * np.dtype(np.int32).itemsize)
+        buffer = pinned_mr.allocate(2 * np.dtype(np.int32).itemsize, stream=device.default_stream)
```
leofang: Isn't a stream created at line 53?

cuda_core/tests/helpers/buffers.py
```diff
@@ -140,7 +140,7 @@ def _get_pattern_buffer(self, seed, value):
 
 def make_scratch_buffer(device, value, nbytes):
     """Create a unified memory buffer with the specified value."""
-    buffer = DummyUnifiedMemoryResource(device).allocate(nbytes)
+    buffer = DummyUnifiedMemoryResource(device).allocate(nbytes, stream=device.default_stream)
     set_buffer(buffer, value)
```
leofang: Q: Is this stream-ordered?

leofang (changes requested):
Thanks, Andy! LGTM overall. Caught one minor bug.

cuda_core/cuda/core/_memory/_buffer.pyx
```diff
@@ -54,7 +54,7 @@ cdef void _mr_dealloc_callback(
         stream = None
         if h_stream:
             stream = Stream._from_handle(Stream, h_stream)
-        mr.deallocate(int(ptr), size, stream)
+        mr.deallocate(int(ptr), size, stream=stream)
```
leofang, line 69: **Consideration:** When `h_stream` is falsy, `stream` is set to `None` (line 54, unchanged), then passed to `mr.deallocate(int(ptr), size, stream=stream)` here. After this PR, pool-based MRs (`DeviceMemoryResource`, `PinnedMemoryResource`, `ManagedMemoryResource`) require a non-`None` stream in `deallocate` — `Stream_accept(None)` will raise `TypeError`.

The `except Exception` handler catches it and prints a warning (no crash), but the memory leaks silently. Is it guaranteed that pool-backed buffers always carry a deallocation stream handle? If not, this is a latent bug.

cuda_core/cuda/core/_memory/_legacy.py
```diff
@@ -27,33 +28,33 @@ class LegacyPinnedMemoryResource(MemoryResource):
...
+        Stream_accept(stream)
         if size:
             err, ptr = driver.cuMemAllocHost(size)
             raise_if_driver_error(err)
         else:
             ptr = 0
         return Buffer._init(ptr, size, self)
 
-    def deallocate(self, ptr: DevicePointerT, size, stream):
+    def deallocate(self, ptr: DevicePointerT, size, *, stream: Stream | None = None):
```
leofang: **Consideration:** When `stream` IS provided here, the method calls `stream.sync()` directly without first running it through `Stream_accept` for type validation. A non-`Stream` object would get `AttributeError` on `.sync()` instead of the clean `TypeError` that `allocate` gives (which does go through `Stream_accept`). Same applies to `_SynchronousMemoryResource.deallocate` below.

cuda_core/tests/helpers/buffers.py
```diff
@@ -22,11 +22,11 @@ class DummyUnifiedMemoryResource(MemoryResource):
     def __init__(self, device):
         self.device = device
 
-    def allocate(self, size, stream=None) -> Buffer:
+    def allocate(self, size, *, stream) -> Buffer:
```
leofang: Nit: `cuMemAllocManaged` is synchronous — the `stream` param here satisfies the `MemoryResource` interface but is never passed to any CUDA call. Same for `deallocate` (`cuMemFree`) and `TrackingMR` below. A brief comment (e.g. `# stream accepted for interface conformance; cuMemAllocManaged is synchronous`) would clarify intent.

leofang (commented):
Overall the core API changes look correct and complete. All APIs from #2001 are properly converted to keyword-only with no default, `Stream_accept` correctly rejects `None`, and the exemptions are well-justified.

Inline comments flag one potential runtime concern (`_mr_dealloc_callback` passing `None` to pool-based MRs) and a couple of minor consistency items. See individual threads for details.

cuda_core/tests/helpers/buffers.py
leofang: No — `cuMemAllocManaged` and `cuMemFree` are both synchronous, non-stream-ordered calls. The `stream` parameter is accepted here purely to satisfy the `MemoryResource` interface contract but is never passed to any CUDA API. The managed memory pointer is immediately valid on all streams.

Same applies to `TrackingMR` right below (`cuMemAlloc`/`cuMemFree` are also synchronous).

cuda_core/examples/graph_update.py
leofang: +1 — a dedicated `stream` is created at line 53 and used for all graph operations. Should pass `stream=stream` here instead of `stream=device.default_stream`. Harmless since `LegacyPinnedMemoryResource` ignores the stream (`cuMemAllocHost` is synchronous), but bad as a model for users who might copy this pattern into pool-backed code where stream ordering matters.

cuda_core/cuda/core/_memory/_buffer.pyx
Andy-Jost: I think we relied on the default value `stream=None` to pick up the default stream on the receiving end.

Andy-Jost:
- `ca2f5f70` adds missing `Stream_accept` calls for validating optional streams. Also fixes a case where the wrong stream was supplied in an ignored context (though this is removed in the subsequent commit).
- `91807eea` removes ignored stream arguments supplied to synchronous resources, including `Legacy*` and `Dummy*` MRs, and `TrackingMR`, among others. Adds comments indicating which APIs accept but ignore streams for compatibility.
- `5e9a5ea2` addresses [this comment](https://github.com/NVIDIA/cuda-python/pull/2020#discussion_r3185711454), by passing the default stream from `Buffer` to `mr.deallocate` when no deallocation stream was supplied. Although this introduces an implicit stream, which is discouraged, the alternative (leaking memory) is worse.

cuda_core/cuda/core/_memory/_buffer.pyx
```diff
@@ -49,12 +49,24 @@ cdef void _mr_dealloc_callback(
...
+    policy because the teardown has no other source of truth.
+    """
+    cdef Stream stream
     try:
-        stream = None
-        if h_stream:
-            stream = Stream._from_handle(Stream, h_stream)
-        mr.deallocate(int(ptr), size, stream)
+        stream = Stream._from_handle(Stream, h_stream) if h_stream else default_stream()
+        mr.deallocate(int(ptr), size, stream=stream)
```
leofang, line 69: My apology. It seems my bot was hallucinating? I thought we guaranteed already that we always have a deallocation stream. There should be no way that we would need the default stream here. @Andy-Jost could you confirm? 

Below is bot-generated:

---

For pool allocations via _MemPool.allocate(): the invariant holds — h_stream is always non-zero. In deviceptr_alloc_from_pool() (resource_handles.cpp:719), the stream is stored in the DevicePtrBox at creation and captured by the shared_ptr deleter. Same for Buffer.from_ipc_descriptor(). So for the normal async MR allocation path, the default_stream() fallback is unreachable.

The one path that breaks the invariant: Buffer.from_handle(ptr, size, mr=some_mr). This calls deviceptr_create_with_mr() (resource_handles.cpp:847) which creates the box with StreamHandle{} (empty). No stream parameter is accepted by from_handle(). So if someone constructs a buffer via from_handle(mr=pool_mr), the callback sees h_stream=0.

So the question is: is Buffer.from_handle(mr=pool_mr) with an async MR a legitimate use case? If not, the default_stream() fallback papers over something that should be an error. If yes, then from_handle() arguably should accept an optional stream when mr is provided.

For the test Andy added (test_mr_dealloc_callback_falls_back_to_default_stream), it constructs a buffer via Buffer.from_handle(1, 1024, mr=mr) — which is exactly the from_handle path, not the pool allocation path. This confirms the fallback only matters for from_handle, not for normal pool allocations.

Should this be flagged in the review? The options seem to be:
- a) Keep the fallback but document it's specifically for the from_handle(mr=...) edge case, not for pool allocations
- b) Make from_handle() accept an optional stream param when mr is provided, and treat missing-stream-on-async-MR as an error in the callback instead

cuda_core/tests/test_memory.py
```diff
@@ -513,16 +519,66 @@ def test_mr_deallocate_receives_stream():
...
-            super().deallocate(ptr, size, stream)
+            super().deallocate(ptr, size, stream=stream)
 
     mr = StreamCaptureMR()
     buf = mr.allocate(1024)
     buf.close(stream)
     assert received["stream"].handle == stream.handle
 
 
+def test_mr_dealloc_callback_falls_back_to_default_stream():
```
leofang, line 532: ditto

cuda_core/cuda/core/_memory/_buffer.pyx
Andy-Jost: `from_handle` was the case I had in mind: when someone constructs a Buffer from a raw handle without a deallocation stream should we raise an exception or try to deallocate on the default stream? Prior to this change, the default stream was used, so I kept it with this commit. I didn't consider reworking from-handle construction to avoid missing streams in the context of this PR.

My take is this is a corner that we can punt on for now. The ordinary paths going through `cuda.core` APIs objects provide a deallocation stream. We have an issue (#1989) to do a pass over the `from_handle` methods, cleaning them up. Can we revisit this then? In the meantime, I don't think there's much risk leaving this as-is.

leofang (approved):

PR #2029, opened 2026-05-05
cuda.core: fix invalid python_version target in pixi.toml

rparolin:
## Summary

- `cuda_core/pixi.toml` had `[package.target.'python_version < "3.11"'.run-dependencies]`, but pixi only accepts platform identifiers in `[package.target.<X>]`. Every `pixi run` against this manifest failed at parse time:
  ```
  × 'python_version < "3.11"' is not a known platform.
  ```
- Move `backports.strenum` into the unconditional `[package.run-dependencies]`. Conda-forge ships per-Python variants (real impl gated to `python <3.11`, plus a 1.3.1 stub gated to `python >=3.12`), so the solver picks the right build per `workspace.build-variants` Python (3.10–3.14).
- Lockfile regenerated; `backports.strenum-1.3.1` now appears in the relevant environments.

## Why this slipped in

No CI workflow currently runs `pixi`, so the broken stanza was never validated. Worth a follow-up to add a `pixi task list` smoke check, but out of scope here.

## Test plan

- [x] `pixi task list --manifest-path cuda_core/pixi.toml` — manifest parses
- [x] `pixi run --manifest-path cuda_core -e default test` — `3039 passed, 178 skipped, 2 xfailed`
- [x] `pixi run test` from repo root — `3026 passed, 187 skipped, 2 xfailed`

### DIFF
--- a/cuda_core/pixi.lock
+++ b/cuda_core/pixi.lock
@@ -11,6 +11,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/alsa-lib-1.2.15.3-hb03c661_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/aom-3.9.1-hac33072_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/attr-2.5.2-h39aace5_0.conda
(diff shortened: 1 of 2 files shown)


mdboom (commented):
Seems fine for `pixi`.  Wouldn't want to ship this metadata in any package we ship, but fine for local use.

kkraus14 (approved):

PR #2037, opened 2026-05-06
Move enums from typing -> enums

mdboom:
As discussed offline, we could instead put all our enums into their own module, rather than in typing.


leofang:
I don't like it 🥲

leofang:
I prefer cuda.core.typing for all-in-one dumpster, including enums. 

By the original design this is a non-issue. Instead of allowing passing `cuda.core.enum.AddressingMode.HMM` we would have limited to only passing `"HMM"` (a str), and then we would have added `cuda.core.typing.AddressingModeType = Literal["HMM", "ATS"]`. So by this argument they belong to `typing` too, not `enum`. Enum is IMHO an artificial construct and implementation detail.

PR #2040, opened 2026-05-06
cuda.core: consolidate driver version query functions

leofang:
## Summary

Consolidate `get_driver_version()` and `get_driver_version_full()` into a single `get_driver_version()` that returns `(umd_version, kmd_version)` — a pair of version tuples (UMD is a 2-tuple `(MAJOR, MINOR)`, KMD is a 3-tuple `(MAJOR, MINOR, PATCH)`).

The function now requires NVML and raises `RuntimeError` if it is not available.

- Removes the `kernel_mode` parameter and the separate `get_driver_version_full` function
- Updates tests, docs, and 1.0.0 release notes


leofang:
/ok to test [bfd9ab9](https://github.com/NVIDIA/cuda-python/pull/2040/commits/bfd9ab9fd9fbeb7d7f52ea1cb8a899f266b0b5e7)

leofang:
/ok to test [d2d387e](https://github.com/NVIDIA/cuda-python/pull/2040/commits/d2d387ea1431a06c20d2988e891dbad127d5b521)

cuda_core/tests/system/test_system_system.py
```diff
@@ -19,11 +19,25 @@
 from .conftest import skip_if_nvml_unsupported
 
 
+@skip_if_nvml_unsupported
```
rparolin: Why does nvml have to be supported to check the driver version?

cuda_core/tests/system/test_system_system.py
```diff
@@ -41,28 +55,11 @@ def test_devices():
...
-    assert len(driver_version) in (2, 3)
-
-    (ver_maj, ver_min, *ver_patch) = driver_version
-    assert 400 <= ver_maj < 1000
-    assert ver_min >= 0
-    if ver_patch:
-        assert 0 <= ver_patch[0] <= 99
+def test_driver_version_requires_nvml():
+    if system.CUDA_BINDINGS_NVML_IS_COMPATIBLE:
+        pytest.skip("NVML is available, cannot test the error path")
```
rparolin, line 60: More comments explaining why NVML being available prevents testing would be a good trace for the reader.

rparolin:
Nit: the file-level comment at [`_system.pyx#L6-L8`](https://github.com/NVIDIA/cuda-python/blob/d2d387ea1431a06c20d2988e891dbad127d5b521/cuda_core/cuda/core/system/_system.pyx#L6-L8) still states that functions must "fall back to non-NVML-based methods for backward compatibility" when `cuda.bindings.nvml` isn't available. After this PR, `get_driver_version()` no longer follows that pattern (it raises `RuntimeError` instead), joining `get_driver_branch()` / `get_nvml_version()` which already required NVML. Worth tweaking the header to reflect that the file is now a mix — some functions fall back (e.g. `get_num_devices`), others require NVML.

cuda_core/tests/system/test_system_system.py
leofang: Because this PR now uses NVML API calls exclusively.

rparolin (approved):
The changes are correct but some added context would make it obvious to reader why its correct.  I've labeled them as nit to not block this PR.

cuda_core/tests/system/test_system_system.py
leofang: Because this test needs a negation for `@skip_if_nvml_unsupported`.

cuda_core/tests/system/test_system_system.py
rparolin: Yeah, I'd argue there should be more traces to the reader to make that obvious. That coupled with the incorrect header at the top of the file make it difficult to parse.

leofang:
Let me wait for @mdboom blessing before merging

mdboom (changes requested):
I don't think we can make NVML a hard requirement, because it means we now require cuda_bindings >=12.9.6 or cuda_bindings >= 13.2.0.  Don't we make broader guarantees in general for that?

To resolve that, maybe we just have two separate functions for kernel and cuda version, but remove the "_full" distinction (which was just there for backward compat).

mdboom (approved):
LGTM

cuda_core/cuda/core/system/__init__.py
```diff
@@ -11,10 +11,10 @@
 
 __all__ = [
     "CUDA_BINDINGS_NVML_IS_COMPATIBLE",
-    "get_driver_version",
-    "get_driver_version_full",
+    "get_kernel_mode_driver_version",
     "get_num_devices",
     "get_process_name",
+    "get_user_mode_driver_version",
```
rwgk, line 17: For the cuda-pathfinder guard rails that I'm adding in pending PR #1977, I need both functions, too:

https://github.com/NVIDIA/cuda-python/pull/1977/changes#diff-e4eb8e0d6231c5ab21d8993a187523d4ef49b2b6eba39ee84b606968ea726722

Currently I have different naming. It might be good to keep the names aligned.

I was going with what `nvidia-smi` is showing, but that's changing. In the future it'll look like this:

```
| NVIDIA-SMI NNN.NN                 KMD Version: NNN.NN        CUDA UMD Version: NN.N     |
```

So

`get_driver_kmd_version`
`get_driver_cuda_umd_version`

would be most aligned.

rwgk (commented):
Ah, looks like I was too slow.

PR #2041, opened 2026-05-06
cuda.core: keep kernel-argument objects alive in graph kernel nodes

### LINKED ISSUE #2039: Graph kernel nodes don't keep kernel argument objects alive
## Summary

`GraphDefinition.launch()` does not extend the lifetime of Python kernel-argument objects to the lifetime of the graph. Inside `GN_launch`, a `ParamHolder` is built from the call's `args` tuple and holds the only strong references to those objects; when `GN_launch` returns, the `ParamHolder` is destroyed and its references drop.

The CUDA driver records the raw pointer values into the kernel node, but it does not (and cannot) retain the Python objects those pointers refer to. So if a kernel argument is reachable only through the `.launch()` call - for example, a `Buffer` passed inline - it can be garbage-collected before the graph runs. The graph then holds a stale device pointer, and instantiating/launching it produces undefined behavior: typically memory corruption or a crash.

`KernelHandle` and `EventHandle` are already retained via `_attach_user_object` for exactly this reason. Kernel arguments need the same treatment.

## Reproducer

```python
def test_kernel_args_buffer_not_kept_alive():
    """Buffer passed as a kernel arg is not kept alive by the graph.

    The ParamHolder built in GN_launch holds the only references to the
    kernel args and is destroyed when GN_launch returns. With no other
    reference to the Buffer, its device memory is freed and the graph
    is left with a stale pointer.
    """
    import ctypes
    import gc
    import weakref

    from cuda.bindings import driver as drv
    from cuda.core import Device, DeviceMemoryResource, LaunchConfig, Program, ProgramOptions
    from cuda.core.graph import GraphDefinition

    dev = Device()
    dev.set_current()
    stream = dev.create_stream()
    mr = DeviceMemoryResource(dev)

    arch = "".join(f"{i}" for i in dev.compute_capability)
    prog = Program(
        "__global__ void add_one(int *a) { *a += 1; }",
        code_type="c++",
        options=ProgramOptions(std="c++17", arch=f"sm_{arch}"),
    )
    add_one = prog.compile("cubin", name_expressions=("add_one",)).get_kernel("add_one")
    config = LaunchConfig(grid=1, block=1)

    buf = mr.allocate(ctypes.sizeof(ctypes.c_int), stream=stream)
    stream.sync()
    buf_weak = weakref.ref(buf)
    drv.cuMemcpyHtoD(int(buf.handle), (ctypes.c_int * 1)(0), ctypes.sizeof(ctypes.c_int))

    g = GraphDefinition()
    g.launch(config, add_one, buf)

    del buf
    gc.collect()
    assert buf_weak() is None  # nothing in the graph kept the Buffer alive

    g.instantiate().launch(stream)
    stream.sync()  # crash or memory corruption
```

## Proposed fix

In `GN_launch`, attach the `ParamHolder` (or just its `kernel_args` tuple) to the graph as a CUDA user object, mirroring the existing handling of `KernelHandle`. That ties the Python-side argument lifetimes to the graph itself.

Andy-Jost:
## Summary

Closes #2039.

`GraphDefinition.launch()` did not extend the lifetime of Python kernel-argument objects (e.g. `Buffer`) to the lifetime of the graph. The ownership represented by a `ParamHolder` constructed in `GN_launch` needs to be attached to the graph to avoid the possibility of stale arguments producing memory corruption or a crash on launch.

## Changes

- `cuda_core/cuda/core/graph/_graph_node.pyx`: in `GN_launch`, attach the `kernel_args` tuple to the graph as a CUDA user object, mirroring the existing handling of `KernelHandle` and `EventHandle`. Reuses the `_py_host_destructor` path already used by the host-callback machinery.
- `cuda_core/cuda/core/graph/_utils.pxd`: expose `_py_host_destructor` so the new caller can use it.

The new attachment runs only on the graph-construction path and is paid once per kernel node at build time, not at execution time. It does not affect the regular (non-graph) launch path in `_launcher.pyx`.

## Test Coverage

Two tests added in `cuda_core/tests/graph/test_graph_definition_lifetime.py`:

- `test_kernel_args_buffer_kept_alive_through_execution`: a `Buffer` passed as a kernel arg survives `del buf` + `gc.collect()` (weakref check) and the graph executes correctly against its memory after instantiation (value check).
- `test_kernel_args_survive_graph_clone`: same scenario but via `cuGraphClone`, which doesn't carry Python-level references — only CUDA user objects can keep the args alive across the clone.

## Related Work

- Closes #2039.
- A follow-up issue will track hardening `_py_host_destructor` against being invoked after `Py_Finalize`. That is a pre-existing risk (also present on the host-callback path) that this PR inherits but does not introduce or widen.

### DIFF
--- a/cuda_core/tests/graph/test_graph_definition_lifetime.py
+++ b/cuda_core/tests/graph/test_graph_definition_lifetime.py
@@ -3,13 +3,15 @@
 
 """Tests for GraphDefinition resource lifetime management and RAII correctness."""
 
+import ctypes
 import gc
+import weakref
 
 import pytest
 from helpers.graph_kernels import compile_common_kernels
 from helpers.misc import try_create_condition
 
-from cuda.core import Device, EventOptions, Kernel, LaunchConfig
+from cuda.core import Device, DeviceMemoryResource, EventOptions, Kernel, LaunchConfig
 from cuda.core.graph import (
     ChildGraphNode,
     ConditionalNode,
@@ -485,3 +487,78 @@ def test_kernel_node_reconstruction_preserves_validity(init_cuda):
     stream = Device().create_stream()
     graph.launch(stream)
     stream.sync()
+
+
+# =============================================================================
+# Kernel argument lifetime — kernel nodes should keep argument objects alive
+# =============================================================================
+
+
+def test_kernel_args_buffer_kept_alive_through_execution(init_cuda):
+    """Buffer passed as a kernel arg is kept alive by the graph, and the kernel
+    actually executes against its memory after the original Python ref drops.
+
+    Without the user-object attachment, the ParamHolder is destroyed when the
+    kernel node is added, the Buffer is GC'd, and the graph is left with a
+    stale device pointer.
+    """
+    from cuda.core._utils.cuda_utils import driver, handle_return
+
+    _skip_if_no_mempool()
+    dev = Device()
+    mr = DeviceMemoryResource(dev)
+    add_one = compile_common_kernels().get_kernel("add_one")
+    buf = mr.allocate(ctypes.sizeof(ctypes.c_int), stream=dev.default_stream)
+    buf.fill(0, stream=dev.default_stream)
+    dev.default_stream.sync()
+    buf_weak = weakref.ref(buf)
+    dptr = int(buf.handle)
+
+    g = GraphDefinition()
+    g.launch(LaunchConfig(grid=1, block=1), add_one, buf)
+
+    del buf
+    gc.collect()
+    assert buf_weak() is not None  # graph kept the Buffer alive
+
+    stream = dev.create_stream()
+    g.instantiate().launch(stream)
+    stream.sync()
+
+    out = (ctypes.c_int * 1)(0)
+    handle_return(driver.cuMemcpyDtoH(out, dptr, ctypes.sizeof(ctypes.c_int)))
+    assert out[0] == 1
+
+
+def test_kernel_args_survive_graph_clone(init_cuda):
+    """Cloned graph keeps Buffer alive via CUDA user objects.
+
+    A graph clone does not inherit Python-level references, so only user
+    objects (which propagate through cuGraphClone) can keep the args alive.
+    """
+    from cuda.core._utils.cuda_utils import driver, handle_return
+
+    _skip_if_no_mempool()
+    dev = Device()
+    mr = DeviceMemoryResource(dev)
+    add_one = compile_common_kernels().get_kernel("add_one")
+    buf = mr.allocate(ctypes.sizeof(ctypes.c_int), stream=dev.default_stream)
+    buf.fill(0, stream=dev.default_stream)
+    dev.default_stream.sync()
+    dptr = int(buf.handle)
+
+    g = GraphDefinition()
+    g.launch(LaunchConfig(grid=1, block=1), add_one, buf)
+    cloned_cu_graph = handle_return(driver.cuGraphClone(driver.CUgraph(g.handle)))
+
+    del buf, g
+    gc.collect()
+
+    graph_exec = handle_return(driver.cuGraphInstantiate(cloned_cu_graph, 0))
+    stream = dev.create_stream()
+    handle_return(driver.cuGraphLaunch(graph_exec, driver.CUstream(int(stream.handle))))
+    stream.sync()
+
+    out = (ctypes.c_int * 1)(0)
+    handle_return(driver.cuMemcpyDtoH(out, dptr, ctypes.sizeof(ctypes.c_int)))
+    assert out[0] == 1
--- a/cuda_core/tests/graph/test_graph_definition_lifetime.py
+++ b/cuda_core/tests/graph/test_graph_definition_lifetime.py
@@ -494,9 +494,10 @@ def test_kernel_node_reconstruction_preserves_validity(init_cuda):
 # =============================================================================
 
 
-def test_kernel_args_buffer_kept_alive_through_execution(init_cuda):
-    """Buffer passed as a kernel arg is kept alive by the graph, and the kernel
-    actually executes against its memory after the original Python ref drops.
+def test_kernel_args_buffer_lifetime(init_cuda):
+    """Buffer passed as a kernel arg is kept alive by the graph, the kernel
+    executes against its memory after the original Python ref drops, and the
+    Buffer is released once the graph is destroyed.
 
     Without the user-object attachment, the ParamHolder is destroyed when the
     kernel node is added, the Buffer is GC'd, and the graph is left with a
@@ -528,6 +529,10 @@ def test_kernel_args_buffer_kept_alive_through_execution(init_cuda):
     out = (ctypes.c_int * 1)(0)
     handle_return(driver.cuMemcpyDtoH(out, dptr, ctypes.sizeof(ctypes.c_int)))
     assert out[0] == 1
+
+    del g
+    gc.collect()
+    assert buf_weak() is None  # graph released, Buffer freed
 
 
 def test_kernel_args_survive_graph_clone(init_cuda):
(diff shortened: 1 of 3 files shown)


cuda_core/tests/graph/test_graph_definition_lifetime.py
```diff
@@ -485,3 +487,78 @@ def test_kernel_node_reconstruction_preserves_validity(init_cuda):
...
+    dev.default_stream.sync()
+    buf_weak = weakref.ref(buf)
+    dptr = int(buf.handle)
+
+    g = GraphDefinition()
+    g.launch(LaunchConfig(grid=1, block=1), add_one, buf)
+
+    del buf
+    gc.collect()
+    assert buf_weak() is not None  # graph kept the Buffer alive
```
rparolin, line 547: Test prove the buffer is kept alive, but it doesn't validate that its cleaned up after the graph is released.

rparolin (changes requested):
The tests should validate that buffer is eventually freed once the graph is refcount is decremented.

cuda_core/tests/graph/test_graph_definition_lifetime.py
Andy-Jost: I added a test for this. If it is flakey, we might need to adjust the `CU_USER_OBJECT_NO_DESTRUCTOR_SYNC` flag so that graph destructors cannot be invoked asynchronously.

Update: I confirmed this is not a concern for source graphs. Asynchronous destruction only comes into play for exec graphs.

rwgk:
Below are the Cursor GPT-5.4 Extra High Fast findings. It was thinking far longer than I'd have expected for a PR this size.

I'm not sure which of these are actually actionable:

Re 1. Do we care about stream-captured graphs?
Re 2. Could we simply document that we don't protect against explicit release?
Re 3. This seems OK to me? (I.e. I mean we can ignore this finding, or document the behavior?)

@Andy-Jost 
___


1. High: Stream-captured graphs still use the unfixed launcher path

   `launch()` still accepts a `GraphBuilder` and creates a stack-local `ParamHolder` in `cuda_core/cuda/core/_launcher.pyx:23` and `cuda_core/cuda/core/_launcher.pyx:47`, but the new ownership transfer exists only in `GN_launch()` at `cuda_core/cuda/core/graph/_graph_node.pyx:623`. That means `launch(gb, ..., buf)` can still capture a kernel node whose `Buffer` or other Python owner is freed after capture, leaving the graph with a stale raw pointer. This is a real public path, not just a hypothetical one; existing tests exercise captured kernel launches with `Buffer` args in `cuda_core/tests/graph/test_graph_memory_resource.py:171`. The new regression tests only cover the explicit `GraphDefinition` path in `cuda_core/tests/graph/test_graph_definition_lifetime.py:497` and `cuda_core/tests/graph/test_graph_definition_lifetime.py:533`.

2. High: Retaining the `Buffer` wrapper does not protect against explicit release

   The PR now keeps `ker_args.kernel_args` alive at `cuda_core/cuda/core/graph/_graph_node.pyx:623`, but `ParamHolder` snapshots a `Buffer` argument as a raw device pointer value in `cuda_core/cuda/core/_kernel_arg_handler.pyx:283` and `cuda_core/cuda/core/_kernel_arg_handler.pyx:287`. If user code explicitly calls `buf.close()` or exits a context manager, `Buffer_close()` resets `_h_ptr` and can free the allocation immediately at `cuda_core/cuda/core/_memory/_buffer.pyx:596`. At that point the graph still owns only the stale integer pointer copied during node creation; keeping the Python `Buffer` object alive does not keep the underlying allocation handle alive.

3. Medium: The new graph-scoped attachment can pin kernel-argument owners after node deletion

   `_attach_user_object()` creates a CUDA user object and moves it into the graph at `cuda_core/cuda/core/graph/_utils.pyx:41`, but it does not return any handle that could later be released per node. `GraphNode.destroy()` only calls `cuGraphDestroyNode()` at `cuda_core/cuda/core/graph/_graph_node.pyx:159`. Because CUDA user objects are retained and released at graph scope (`cuGraphRetainUserObject` / `cuGraphReleaseUserObject`), deleting or rewiring a kernel node can now leave large kernel-argument owners, including `Buffer`s, pinned until the entire graph is destroyed. Node mutation is already a supported workflow in `cuda_core/tests/graph/test_graph_definition_mutation.py:159`.

rparolin (approved):

Andy-Jost:
Thanks @rwgk 

> 1. High: Stream-captured graphs still use the unfixed launcher path

I'll look into this. It might need to be deferred because AFAIK the stream capture path does not create any user objects.

> 2. High: Retaining the Buffer wrapper does not protect against explicit release

We have a huge class of possible errors of this type, unfortunately. A better approach than storing the `Buffer` would be to store the `DevicePtrHandle` (a `std::shared_ptr` owning the buffer). Definitely out of scope for this change.

> 3. Medium: The new graph-scoped attachment can pin kernel-argument owners after node deletion

I have to rework the whole user object design for #1330 (step 4) and I plan to address this.

leofang:
> High: Stream-captured graphs still use the unfixed launcher path

This is a good catch, and it is not fixed with this PR. (Which is why kernel arg update is so messy, as noted during the team sync today). I am fine with this PR only fixing the explicit graph construction path.

rwgk (approved):

cuda_core/tests/graph/test_graph_definition_lifetime.py
Andy-Jost: This test creates an exec graph, so there is a race. CI for free-threaded Python seems more likely to trigger it. 9f2c8f2aba6825b8b3648e03c820631d55c879d7 adds polling, but removing the test would also be defensible.

cuda_core/tests/graph/test_graph_definition_lifetime.py
```diff
@@ -3,13 +3,34 @@
...
+    Python, by deferred reference-count processing. A bounded poll keeps the
+    test correct without depending on undocumented driver timing guarantees.
+    """
+    deadline = time.monotonic() + timeout
+    while time.monotonic() < deadline:
+        gc.collect()
+        if predicate():
+            return
+        time.sleep(interval)
+    raise AssertionError(f"condition not satisfied within {timeout}s")
```
leofang, line 30: Wouldn't this still welcome flakiness? I am concerned about this being tested in SWQA hands

cuda_core/tests/graph/test_graph_definition_lifetime.py
```diff
@@ -485,3 +506,87 @@ def test_kernel_node_reconstruction_preserves_validity(init_cuda):
...
+    Without the user-object attachment, the ParamHolder is destroyed when the
+    kernel node is added, the Buffer is GC'd, and the graph is left with a
+    stale device pointer.
+
+    The final freeing assertion uses a bounded poll because CUgraphExec
+    releases its user-object references via an asynchronous DPC, and on
+    free-threaded Python the resulting Py_DECREF chain may need an extra
+    GC pass to settle.
+    """
+    from cuda.core._utils.cuda_utils import driver, handle_return
```
leofang, line 530: nit: move imports to the top, no need to defer import to here

cuda_core/tests/graph/test_graph_definition_lifetime.py
```diff
@@ -485,3 +506,87 @@ def test_kernel_node_reconstruction_preserves_validity(init_cuda):
...
+
+    _skip_if_no_mempool()
+    dev = Device()
```
leofang, line 533:
```suggestion
    dev = init_cuda
```

Andy-Jost:
I merged due to the release deadline, but I will follow-up on the open comments.

cuda_core/tests/graph/test_graph_definition_lifetime.py
Andy-Jost: Agreed, it's not perfect, though this is much better than before. Realistically, I think it's either this or we don't test the release condition Rob pointed out. There will be much more work on the graph ownership model, so I expect to revisit all these tests.

Andy-Jost: https://github.com/NVIDIA/cuda-python/issues/2044

PR #2046, opened 2026-05-07
Fix sccache summary to count CUDA sub-tool cache hits

leofang:
## Summary

The sccache step summary under-reports the cache hit rate because CUDA sub-tool compilations are excluded from the count.

The rapidsai/sccache fork tracks CUDA compilation sub-phases under separate language keys in its JSON stats:
- `"CUDA"` — nvcc driver pass
- `"CUDA (Device code)"` — cudafe++ pass
- `"PTX"` — cicc pass
- `"CUBIN"` — ptxas pass

The summary script only counted `{"C/C++", "CUDA"}`, which captures the nvcc driver pass only. That pass typically has 0 cache hits — the actual hits all land in the sub-tool categories.

This PR adds the missing keys so the reported rate matches sccache's own output.

Related:
- NVIDIA/numba-cuda#878 (same fix for numba-cuda)
- NVIDIA/cccl#8858 (same fix for CCCL)

## Test plan

- Verified locally with mock JSON matching real CI data
- The fix will be validated on the next CI run that produces sccache stats

### DIFF
--- a/.github/actions/sccache-summary/action.yml
+++ b/.github/actions/sccache-summary/action.yml
@@ -6,8 +6,6 @@ name: sccache summary
 description: Parse sccache stats JSON and write a summary table to GITHUB_STEP_SUMMARY
 
 # Inspired by NVIDIA/cccl's prepare-execution-summary.py (PR #3621).
-# Only counts C/C++ and CUDA language hits (excludes PTX/CUBIN which are
-# not included in sccache's compile_requests counter).
 
 inputs:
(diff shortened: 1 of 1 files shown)


leofang:
/ok to test [4be4714](https://github.com/NVIDIA/cuda-python/pull/2046/commits/4be47143fddb000392a0183d8e619a38f0bf6dbe)

kkraus14 (approved):

PR #2055, opened 2026-05-08
Fix tab completion

### LINKED ISSUE #2053: [BUG]: Tab completion with `DeviceMemoryResource` is broken
### Is this a duplicate?

### Type of Bug

Runtime Error

### Component

cuda.core

### Describe the bug

Tab completion is not prompting anything for the MR's method:
```python
mr.<tab>
```

git bisect tells me the offending commit is https://github.com/NVIDIA/cuda-python/commit/8b55a6cef6544fbe93b00d4d9bc8a1f3a2aaaa0f

### How to Reproduce

- build cuda-core locally
- run this script via `python -I test_mr.py` to enter the Python REPL:
```python
from cuda.core import Device
from cuda.core import DeviceMemoryResource
dev = Device(0)
dev.set_current()
mr = DeviceMemoryResource(dev)
```
- hit tab completion for `mr.`

### Expected behavior

it should just work

### Operating System

_No response_

### nvidia-smi output

_No response_

mdboom:
This is a possible fix for #2053.

I almost never reach for monkey-patching as a solution, however in this case:

- The alternatives are way worse and intrusive and force us into API design corners for the niche use case of tab completion (see my [comments about the agent's other suggestions](https://github.com/NVIDIA/cuda-python/issues/2053#issuecomment-4407017833)).

- The monkeypatch is quite narrow -- it doesn't patch builtins that might have broad implications.  It should only affect autocompletion with `readline`.  A project like that might exist, but it's probably pretty niche.  IPython and Jupyter are not affected by the bug to begin with, and the patch has no effect on them.  (The patch is /installed/ with IPython, but then the code never gets called.  I experimented with detecting IPython and not installing the patch, but it was complex, specific to IPython, and ultimately doesn't matter).

- This monkeypatch is idempotent. If it gets cargo-culted to multiple projects, and those projects get imported in the same process, it will still work and be effective.

- (Pending the CI run passing) It should be fine across all versions of Python we support, and there is a test here to confirm that going forward.  If Python "fixes" [the issue](https://github.com/python/cpython/pull/27401) such that this patch is no longer needed, the test here should catch that and we can gate it at that point (as part of the usual "bring up on a new Python version" process).

### DIFF
--- a/cuda_core/cuda/core/__init__.py
+++ b/cuda_core/cuda/core/__init__.py
@@ -41,7 +41,7 @@ def _patch_rlcompleter_for_cython_properties():
     # rlcompleter side effect.
     import os
 
-    if os.environ.get("CUDA_CORE_DONT_FIX_TAB_COMPLETION"):
+    if int(os.environ.get("CUDA_CORE_DONT_FIX_TAB_COMPLETION", "0")):
         # Explicit opt-out for users who don't want the global rlcompleter
         # side effect, even in an interactive session.
         return
--- a/cuda_core/cuda/core/__init__.py
+++ b/cuda_core/cuda/core/__init__.py
@@ -28,6 +28,47 @@ _import_versioned_module()
 del _import_versioned_module
 
 
+def _patch_rlcompleter_for_cython_properties():
+    # Cython @property on cdef class compiles to a C-level getset_descriptor,
+    # which rlcompleter's narrow isinstance(..., property) check misses; the
+    # fallback getattr() then invokes the descriptor and any non-AttributeError
+    # it raises kills tab completion. Extend that isinstance check to also
+    # match getset_descriptor / member_descriptor. Only installed in
+    # interactive mode so library users running scripts see no global
+    # rlcompleter side effect.
+    import os
+    import sys
+
+    if os.environ.get("CUDA_CORE_DONT_FIX_TAB_COMPLETION"):
+        # Explicit opt-out for users who don't want the global rlcompleter
+        # side effect, even in an interactive session.
+        return
+    if not (hasattr(sys, "ps1") or sys.flags.inspect):
+        # Plain `python script.py`, `python -c ...`, pytest, etc.
+        return
+
+    import rlcompleter
+    from types import GetSetDescriptorType, MemberDescriptorType
+
+    # This works by overriding the `property` built-in with a custom subclass of
+    # property, but only in the rlcompleter module.  This subclass overrides the
+    # `__instancecheck__` method to also return True for getset_descriptor and
+    # member_descriptor types, which are what Cython uses for properties on cdef
+    # classes.
+    class _PatchedPropMeta(type):
+        def __instancecheck__(cls, inst):
+            return isinstance(inst, (property, GetSetDescriptorType, MemberDescriptorType))
+
+    class _PatchedProperty(metaclass=_PatchedPropMeta):
+        pass
+
+    rlcompleter.property = _PatchedProperty
+
+
+_patch_rlcompleter_for_cython_properties()
+del _patch_rlcompleter_for_cython_properties
+
+
 from cuda.core import checkpoint, system, utils
 from cuda.core._context import Context, ContextOptions
 from cuda.core._device import Device
--- a/cuda_core/cuda/core/__init__.py
+++ b/cuda_core/cuda/core/__init__.py
@@ -29,6 +29,9 @@ del _import_versioned_module
 
 
 def _patch_rlcompleter_for_cython_properties():
+    # TODO: This can be removed when Python 3.13 is our minimum-supported version:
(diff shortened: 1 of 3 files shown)


cuda_core/tests/test_rlcompleter_patch.py
```diff
@@ -0,0 +1,110 @@
...
+# Probe script: reproduces the bug-report repro literally, then runs
+# rlcompleter against `mr` and reports the outcome.
+_PROBE_SCRIPT = textwrap.dedent("""
+    import rlcompleter
+    from cuda.core import Device, DeviceMemoryResource
+
+    dev = Device(0)
+    dev.set_current()
+    mr = DeviceMemoryResource(dev)
+    assert not mr.is_ipc_enabled, "test setup: mr should not be IPC-enabled"
```
mdboom, line 48: This is only testing one specific class that we know broke things.  If we accept this PR, we should probably add testing that /all/ of our Cython classes autocomplete correctly.

mdboom:
Test failures indicate something wrong in the design of the test -- not a problem with the monkey-patch itself.

cuda_core/cuda/core/__init__.py
```diff
@@ -28,6 +28,47 @@ def _import_versioned_module():
...
+    import os
+    import sys
+
+    if os.environ.get("CUDA_CORE_DONT_FIX_TAB_COMPLETION"):
+        # Explicit opt-out for users who don't want the global rlcompleter
+        # side effect, even in an interactive session.
+        return
+    if not (hasattr(sys, "ps1") or sys.flags.inspect):
+        # Plain `python script.py`, `python -c ...`, pytest, etc.
+        return
```
kkraus14: This doesn't seem quite correct, where if I just run `python` from a shell it drops me to a REPL that I should be able to get autocompletion with but `sys.flags.inspect` is `0`.

cuda_core/cuda/core/__init__.py
```diff
@@ -28,6 +28,47 @@ def _import_versioned_module():
...
+
+    import rlcompleter
+    from types import GetSetDescriptorType, MemberDescriptorType
+
+    # This works by overriding the `property` built-in with a custom subclass of
+    # property, but only in the rlcompleter module.  This subclass overrides the
+    # `__instancecheck__` method to also return True for getset_descriptor and
+    # member_descriptor types, which are what Cython uses for properties on cdef
+    # classes.
```
kkraus14, line 56: Are we concerned about any implications of other modules Cython classes `getset_descriptor` objects?

cuda_core/cuda/core/__init__.py
mdboom: Yes, but `ps1` is there, right.  It WFM with `python`.

cuda_core/cuda/core/__init__.py
mdboom: No. If they wanted them to be private, raising an exception unconditionally would not be the way.

cuda_core/cuda/core/__init__.py
```diff
@@ -28,6 +28,46 @@ def _import_versioned_module():
...
+    # Cython @property on cdef class compiles to a C-level getset_descriptor,
+    # which rlcompleter's narrow isinstance(..., property) check misses; the
+    # fallback getattr() then invokes the descriptor and any non-AttributeError
+    # it raises kills tab completion. Extend that isinstance check to also
+    # match getset_descriptor / member_descriptor. Only installed in
+    # interactive mode so library users running scripts see no global
+    # rlcompleter side effect.
+    import os
+
+    if os.environ.get("CUDA_CORE_DONT_FIX_TAB_COMPLETION"):
```
leofang: nit:
```suggestion
    if int(os.environ.get("CUDA_CORE_DONT_FIX_TAB_COMPLETION", "0")):
```

cuda_core/cuda/core/__init__.py
```diff
@@ -28,6 +28,46 @@ def _import_versioned_module():
...
+    # member_descriptor types, which are what Cython uses for properties on cdef
+    # classes.
+    class _PatchedPropMeta(type):
+        def __instancecheck__(cls, inst):
+            return isinstance(inst, (property, GetSetDescriptorType, MemberDescriptorType))
+
+    class _PatchedProperty(metaclass=_PatchedPropMeta):
+        pass
+
+    rlcompleter.property = _PatchedProperty
```
leofang, line 64: Note for posterity: This patches the built-in `property` but only within this module.

leofang (approved):

mdboom:
This still had a test to confirm things aren't fixed when the patch isn't installed.  Since we always install the patch now (except when the envvar is set), that test was failing.  Removed the test -- hopefully things pass now.

leofang:
Admin-merging this to unblock the release... There seems to exist a job scheduling problem that the L4 job refuses to run.

PR #2056, opened 2026-05-08
Use Python properties for Cython descriptors

kkraus14:
## Summary
- Add a shared `python_property` decorator that creates real Python `property` descriptors inside Cython `cdef class` bodies.
- Convert the known public Cython descriptors that may raise or perform CUDA/NVML work during completion to use `@python_property` instead of generated `getset_descriptor`s.
- Add regression coverage that checks those public descriptors are `property` instances and not `getset_descriptor` objects.

## Testing
- pixi run ruff check cuda_core/cuda/core/_utils/properties.py cuda_core/tests/test_cython_property_descriptors.py
- pixi run --manifest-path cuda_core -e cu13 pytest cuda_core/tests/test_cython_property_descriptors.py *(blocked locally because `cuda_core/pixi.toml` on this branch still has the known `python_version < "3.11"` package target parse error; the pixi fix is intentionally left to the separate PR)*
- commit hooks: ruff, SPDX, toml, cython-lint, and other pre-commit checks passed

### DIFF
--- a/cuda_core/cuda/core/_memoryview.pyx
+++ b/cuda_core/cuda/core/_memoryview.pyx
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from __future__ import annotations
+from cuda.core._utils.properties import python_property
 
 from ._dlpack cimport *
 from ._dlpack import classify_dl_device
@@ -484,7 +485,7 @@ cdef class StridedMemoryView:
         _smv_get_dl_device(self, &device_type, &device_id)
         return (<int>device_type, int(device_id))
 
-    @property
+    @python_property
     def _layout(self) -> _StridedLayout:
         """
         The layout of the tensor. For StridedMemoryView created from DLPack or CAI,
@@ -492,25 +493,25 @@ cdef class StridedMemoryView:
         """
         return self.get_layout()
 
-    @property
+    @python_property
     def size(self) -> int:
         return self.get_layout().get_volume()
 
-    @property
+    @python_property
     def shape(self) -> tuple[int, ...]:
         """
         Shape of the tensor.
         """
         return self.get_layout().get_shape_tuple()
 
-    @property
+    @python_property
     def strides(self) -> tuple[int, ...] | None:
         """
         Strides of the tensor (in **counts**, not bytes).
         """
         return self.get_layout().get_strides_tuple()
 
-    @property
+    @python_property
     def dtype(self) -> numpy.dtype | None:
         """
         Data type of the tensor.
--- a/cuda_core/cuda/core/_memoryview.pyx
+++ b/cuda_core/cuda/core/_memoryview.pyx
@@ -485,8 +485,7 @@ cdef class StridedMemoryView:
         _smv_get_dl_device(self, &device_type, &device_id)
         return (<int>device_type, int(device_id))
 
-    @python_property
-    def _layout(self) -> _StridedLayout:
+    def _get_layout(self) -> _StridedLayout:
         """
         The layout of the tensor. For StridedMemoryView created from DLPack or CAI,
         the layout is inferred from the tensor object's metadata.
--- a/cuda_core/cuda/core/_utils/properties.py
+++ b/cuda_core/cuda/core/_utils/properties.py
@@ -3,6 +3,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
 
-def python_property(func):
+def python_property(fget=None, fset=None, fdel=None, doc=None):
     """Create a Python property without Cython's cdef-class @property lowering."""
-    return property(func)
+    return property(fget, fset, fdel, doc)
--- a/cuda_core/cuda/core/_utils/properties.py
+++ b/cuda_core/cuda/core/_utils/properties.py
@@ -1,0 +1,8 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+
+def python_property(func):
+    """Create a Python property without Cython's cdef-class @property lowering."""
+    return property(func)
--- a/cuda_core/cuda/core/_utils/properties.py
+++ b/cuda_core/cuda/core/_utils/properties.py
@@ -2,7 +2,4 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-
-def python_property(fget=None, fset=None, fdel=None, doc=None):
-    """Create a Python property without Cython's cdef-class @property lowering."""
-    return property(fget, fset, fdel, doc)
+python_property = property
--- a/cuda_core/tests/test_cython_property_descriptors.py
+++ b/cuda_core/tests/test_cython_property_descriptors.py
@@ -18,7 +18,11 @@ import cuda.core.system
 pytestmark = pytest.mark.no_cuda
 
 
-_NOT_ALLOWED_TO_IMPORT = {"cuda.core._tensor_bridge"}
+# Suffixes of module names that must not be imported (e.g. _tensor_bridge
+# depends on PyTorch symbols that are unavailable in most test environments).
+# Matched against the end of each discovered module name so that versioned
+# subpackages like cuda.core.cu12._tensor_bridge are also excluded.
+_NOT_ALLOWED_TO_IMPORT_SUFFIXES = {"._tensor_bridge"}
 
 _GETSET_FIELD_ALLOWLIST = {
     ("cuda.core._kernel_arg_handler", "ParamHolder", "ptr"),
@@ -41,7 +45,10 @@ def _iter_cuda_core_modules():
         for info in pkgutil.walk_packages(root.__path__, root.__name__ + "."):
             module_names.add(info.name)
 
-    module_names -= _NOT_ALLOWED_TO_IMPORT
+    module_names = {
+        n for n in module_names
+        if not any(n.endswith(s) for s in _NOT_ALLOWED_TO_IMPORT_SUFFIXES)
+    }
     for module_name in sorted(module_names):
         yield importlib.import_module(module_name)
 
--- a/cuda_core/tests/test_cython_property_descriptors.py
+++ b/cuda_core/tests/test_cython_property_descriptors.py
@@ -45,10 +45,7 @@ def _iter_cuda_core_modules():
         for info in pkgutil.walk_packages(root.__path__, root.__name__ + "."):
             module_names.add(info.name)
 
-    module_names = {
-        n for n in module_names
-        if not any(n.endswith(s) for s in _NOT_ALLOWED_TO_IMPORT_SUFFIXES)
-    }
+    module_names = {n for n in module_names if not any(n.endswith(s) for s in _NOT_ALLOWED_TO_IMPORT_SUFFIXES)}
     for module_name in sorted(module_names):
         yield importlib.import_module(module_name)
 
--- a/cuda_core/tests/test_cython_property_descriptors.py
+++ b/cuda_core/tests/test_cython_property_descriptors.py
@@ -2,56 +2,136 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+from __future__ import annotations
+
+import importlib
 import inspect
+import re
 import types
+from dataclasses import dataclass
+from pathlib import Path
 
 import pytest
 
-from cuda.core import (
-    Buffer,
-    Context,
-    DeviceMemoryResource,
-    DeviceResources,
-    Event,
-    PinnedMemoryResource,
-)
 from cuda.core.system import CUDA_BINDINGS_NVML_IS_COMPATIBLE
 
-_PYTHON_PROPERTIES = [
-    pytest.param(DeviceMemoryResource, "allocation_handle", id="DeviceMemoryResource.allocation_handle"),
-    pytest.param(PinnedMemoryResource, "allocation_handle", id="PinnedMemoryResource.allocation_handle"),
-    pytest.param(Event, "ipc_descriptor", id="Event.ipc_descriptor"),
-    pytest.param(Buffer, "ipc_descriptor", id="Buffer.ipc_descriptor"),
-    pytest.param(Context, "resources", id="Context.resources"),
-    pytest.param(DeviceResources, "workqueue", id="DeviceResources.workqueue"),
-]
+pytestmark = pytest.mark.no_cuda
 
-if CUDA_BINDINGS_NVML_IS_COMPATIBLE:
-    from cuda.core.system import Device as SystemDevice
-
-    _PYTHON_PROPERTIES.extend(
-        [
-            pytest.param(SystemDevice, "attributes", id="system.Device.attributes"),
-            pytest.param(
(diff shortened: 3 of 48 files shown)


cuda_core/cuda/core/_utils/properties.py
```diff
@@ -0,0 +1,8 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+
+def python_property(fget=None, fset=None, fdel=None, doc=None):
+    """Create a Python property without Cython's cdef-class @property lowering."""
+    return property(fget, fset, fdel, doc)
```
mdboom: Does this work?

```suggestion
python_property = property
```

cuda_core/tests/test_cython_property_descriptors.py
```diff
@@ -0,0 +1,137 @@
...
+def test_cython_properties_are_python_properties(cython_property: CythonProperty):
+    if cython_property.module.startswith("cuda.core.system.") and not CUDA_BINDINGS_NVML_IS_COMPATIBLE:
+        pytest.skip("cuda.core.system extension modules require NVML-compatible bindings")
+
+    module = importlib.import_module(cython_property.module)
+    cls = getattr(module, cython_property.class_name)
+    descriptor = inspect.getattr_static(cls, cython_property.name)
+
+    assert isinstance(descriptor, property)
+    assert not isinstance(descriptor, types.GetSetDescriptorType)
```
mdboom, line 138: It seems kind of brittle to do a source-level analysis here.

Could we just iterate over all classes and assert that they don't have any `GetSetDescriptorType`'s?  It's possible I'm missing something (maybe there are implicit descriptors always that we can't remove), but that seems less complicated.

mdboom (changes requested):
One suggestion is just a micro-op for startup time that ultimately doesn't matter.

It would be nice to simplify the testing (if possible) though.  My concern is corner case syntactic constructions that might not get caught.

leofang:
All Linux-64 CI segfaulted. (Ignored linux-arm64 CIs, there seems to be a widespread outage.)

cuda_core/cuda/core/_utils/properties.py
kkraus14: This did work, good catch!

mdboom: I confirmed that this works.

It turns out it's more than just an optimization, but I think is a requirement.  I merged this branch in with my WIP branch to add auto-generated `.pyi` files and then type-check the results.  If `python_property is property` holds, everything works find and mypy and pyright can understand that things are properties and type check them accordingly.  If you hide `python` inside of another function like this, it doesn't have access to the body of the function and it can't reason about it.  So I think this change is a must.

Additionally, if we want to avoid having a module with a single assignment in it, we can just put:

```
python_property = property
```

instead of 

```
from cuda.core._utils.properties import python_property
```

but that's just a matter of style, really.

cuda_core/tests/test_cython_property_descriptors.py
kkraus14: Good call. I replaced the source-level parser with runtime introspection over the cuda-core modules/classes.

A literal “no `GetSetDescriptorType` anywhere” check still needs filtering because Cython also emits getset descriptors for non-property things like `__dict__`, `__weakref__`, and typed cdef fields such as option dataclass fields. The updated test now allows those known field descriptors, but fails on any unexpected getset descriptor exposed by cuda-core classes.

cuda_core/tests/test_cython_property_descriptors.py
```diff
@@ -0,0 +1,82 @@
...
+    ("cuda.core._layout", "_StridedLayout", "itemsize"),
+    ("cuda.core._layout", "_StridedLayout", "slice_offset"),
+    ("cuda.core._memoryview", "StridedMemoryView", "device_id"),
+    ("cuda.core._memoryview", "StridedMemoryView", "exporting_obj"),
+    ("cuda.core._memoryview", "StridedMemoryView", "is_device_accessible"),
+    ("cuda.core._memoryview", "StridedMemoryView", "ptr"),
+    ("cuda.core._memoryview", "StridedMemoryView", "readonly"),
+    ("cuda.core._memoryview", "_StridedMemoryViewProxy", "has_dlpack"),
+    ("cuda.core._memoryview", "_StridedMemoryViewProxy", "obj"),
+}
```
kkraus14: This still feels very brittle where I'm not too happy about this and don't want to have to maintain a list like this.

cuda_core/tests/test_cython_property_descriptors.py
```diff
@@ -0,0 +1,82 @@
...
+
+def _is_allowed_getset_descriptor(cls, name, descriptor) -> bool:
+    if name in {"__dict__", "__weakref__"}:
+        return True
+    # Typed cdef fields generate getset descriptors too, but their docs follow
+    # this compact "field: type" form rather than a property docstring.
+    doc = descriptor.__doc__
+    if doc is not None and doc.startswith(f"{name}:"):
+        return True
+    return (cls.__module__, cls.__qualname__, name) in _GETSET_FIELD_ALLOWLIST
```
kkraus14: This also feels quite brittle

cuda_core/tests/test_cython_property_descriptors.py
mdboom: Yeah, thanks for humoring me.  The `__dict__` and `__weakref__` thing doesn't bother me (that's a pretty static set), but I forgot that `cdef readonly`/`cdef public` things would be indistinguishable from Cython properties from the outside.  Probably best to revert to the first revision that parsed the code.  /Maybe/ the agent could do better than regexes by using the Cython parser (I've had reasonable success with that elsewhere), but we could also cross that bridge if we find the regex approach breaks on some future code.

cuda_core/cuda/core/_memoryview.pyx
```diff
@@ -484,33 +485,33 @@ cdef class StridedMemoryView:
         _smv_get_dl_device(self, &device_type, &device_id)
         return (<int>device_type, int(device_id))
 
-    @property
+    @python_property
```
mdboom: Looking at the segfault in CI (which I think is really just infinite recursion) -->

I think what's happening is that this is now masking the `cdef _layout` on the class for Cython code, whereas before it was somehow only exposing it to Python code.  Renaming one or the other should hopefully do the trick.  I assume the one expected internally from this file is the `cdef` and this property was only expected to be used from outside.

leofang:
> (Ignored linux-arm64 CIs, there seems to be a widespread outage.)

FYI, this should be restored now.

cuda_core/tests/test_cython_property_descriptors.py
```diff
@@ -0,0 +1,82 @@
...
+import pkgutil
+import types
+
+import pytest
+
+import cuda.core
+import cuda.core.graph
+import cuda.core.system
+
+pytestmark = pytest.mark.no_cuda
```
leofang: Is this really needed...?

cuda_core/cuda/core/_memoryview.pyx
leofang: How about making `_layout` a `cdef readonly` and just remove this one? We really don't want general users to access it.

mdboom: It's used from testing in the test suite, but we can just update the test suite to call `get_layout()` since that's all this does.  The `cdef readonly _layout` isn't exposed to Python so the tests otherwise fail.

mdboom: @kkraus14 I pushed a commit here that will hopefully fix the tests.

cuda_core/tests/test_cython_property_descriptors.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+_OPTIONAL_IMPORT_FAILURES = {"cuda.core._tensor_bridge"}
```
leofang:
```suggestion
_NOT_ALLOWED_TO_IMPORT = {"cuda.core._tensor_bridge"}
```

cuda_core/tests/test_cython_property_descriptors.py
```diff
@@ -0,0 +1,82 @@
...
+        for info in pkgutil.walk_packages(root.__path__, root.__name__ + "."):
+            module_names.add(info.name)
+
+    for module_name in sorted(module_names):
+        try:
+            yield importlib.import_module(module_name)
+        except ImportError:
+            if module_name in _OPTIONAL_IMPORT_FAILURES:
+                continue
+            raise
```
leofang:
```suggestion
    module_names -= _NOT_ALLOWED_TO_IMPORT
    for module_name in sorted(module_names):
        yield importlib.import_module(module_name)
```

leofang (changes requested):
The tensor bridge module is not designed to be importable casually, so it causes a test error.

cuda_core/cuda/core/_memoryview.pyx
leofang: It seems to work indeed!

leofang (commented):
## Review

### Approach

`python_property = property` is elegant — zero runtime overhead, works with mypy/pyright (as confirmed [in this thread](https://github.com/NVIDIA/cuda-python/pull/2056#discussion_r3210969873)), and makes Cython `cdef class` properties produce real `property` descriptors that `rlcompleter's` `isinstance(..., property)` guard recognizes. This is the right fix.

### The `_layout` segfault (resolved)

The CI segfaults were caused by `StridedMemoryView._layout`: there's both a `cdef _layout` field and a `@property _layout` on the class. With Cython's native `getset_descriptor`, these coexist (the cdef field is C-only, the property is Python-only). But when `@python_property` is used, the Python `property` descriptor masks the cdef field even from Cython code, causing infinite recursion (`_layout` getter → `get_layout()` → `self._layout` → `_layout` getter → …). The rename to `_get_layout()` method is the correct fix.

### Issues

1. **CI failure: `_tensor_bridge` exclusion mismatch.** The guardrail test fails in CI because the installed package uses a versioned subpackage (`cuda.core.cu12._tensor_bridge`), but the exclusion set only matched the unversioned name (`cuda.core._tensor_bridge`). I pushed a fix ([`6e182ef`](https://github.com/kkraus14/cuda-python/commit/6e182ef9ad)) that matches by suffix instead of exact name, so `cuda.core.cu12._tensor_bridge`, `cuda.core.cu13._tensor_bridge`, etc. are all excluded.

2. **No functional `rlcompleter` test.** The guardrail test checks descriptor types structurally, which is good for preventing regressions. But a test that actually calls `rlcompleter.Completer.complete()` on affected objects and asserts no exception would be stronger end-to-end proof. For example:

   ```python
   import rlcompleter
   completer = rlcompleter.Completer({"mr": mr})
   assert completer.complete("mr.", 0) is not None  # should not raise
   ```

3. **Minor: setter pattern could use a comment.** The pattern used for read-write properties:

   ```python
   def peer_accessible_by(self):
       ...
   def _set_peer_accessible_by(self, devices):
       ...
   peer_accessible_by = python_property(peer_accessible_by, _set_peer_accessible_by)
   ```

   …works correctly, but it's non-obvious why the more Pythonic `@peer_accessible_by.setter` pattern isn't used. A brief comment noting that Cython `cdef class` bodies don't support `.setter` on a `python_property` would help future maintainers.

### Not an issue

- **Over-broad scope** (converting all ~439 properties, not just the ~13 affected ones) — defensively correct, prevents future surprises. Worth the churn.
- **Error message changes** (`"attribute 'X' ... is not writable"` → `"can't set attribute 'X'"`) — correct, Python `property` uses different wording than `getset_descriptor`.
- **`graph/_graph_builder.pyx` is not modified** — this is fine. `GraphBuilder` is a regular Python `class` (not `cdef class`), so its `@property` decorators already produce Python `property` descriptors that `rlcompleter` handles correctly.

-- Leo's bot

leofang:
(fixing it locally)

leofang (commented):
*(Corrections merged into the review above.)*

leofang:
/ok to test [6e182ef](https://github.com/NVIDIA/cuda-python/pull/2056/commits/6e182ef9adf9f372708deabab6bc06254573bc02)

leofang:
/ok to test [7c52713](https://github.com/NVIDIA/cuda-python/pull/2056/commits/7c5271306760a1657d3a986425992e9cbebbe76c)

leofang (approved):

mdboom (changes requested):
1. (As noted elsewhere) We should revert the test to the first revision -- I think it's the better of two imperfect options.  Though it's brittle due to regex parsing, it doesn't have the downside of requiring a specific list of special cases as the current revision does.

2. This suggestion from Leo's agent is a good one:

```
No functional rlcompleter test. The guardrail test checks descriptor types structurally, which is good for preventing regressions. But a test that actually calls rlcompleter.Completer.complete() on affected objects and asserts no exception would be stronger end-to-end proof. For example:

import rlcompleter
completer = rlcompleter.Completer({"mr": mr})
assert completer.complete("mr.", 0) is not None  # should not raise
```

Since we are already walking the tree to find all classes, we should test all of them this way (not just one as suggested here).  It's completely possible there is some other Cython usage we move to in the future that will break autocompletion, so asserting the exact outcome we want would be good.

3. The mixture of `cdef` members and these Python properties changes member resolution order, and is not tested in the Cython test suite.  The exact change is that Python properties take precedence over `cdef` members, whereas Cython properties do not.  This is something that is changing the /compile/ time member lookup and changes a compile time binding to a late binding, which is really surprising and even when it doesn't create infinite recursion as we saw here, it could be an unintended performance regression.  I would feel a lot better if we could write a test here that asserts that there are no `cdef` members and Python properties with the same name on the same class.

leofang:
@mdboom @kkraus14 let's try to get the ball rolling across the finish line

mdboom:
> @mdboom @kkraus14 let's try to get the ball rolling across the finish line

Ok -- I'll take on addressing the 3 comments in my review.

mdboom (approved):
I have addressed the first two comments in my last review:

1) Moving back to the original version of the test here that "parses" the .pyx source files with regexes looking for Cython properties.  This approach feels marginally better because it doesn't require listing exceptions to the rule.

2) Added a test to make sure `rlcompleter` works on some class instances that were known to be broken.  Writing a fully-automatic test that would check all classes in the code base doesn't seem feasible -- creating a class instance is different, may require some pre-requisites etc.

3) @leofang did an experiment elsewhere showing that the only real risk of shadowing problems is between private `cdef` members and Python properties.  `cdef readonly` and `cdef public` will cause create compiler errors.  That makes that problem more contained than I realized, so I'm more comfortable with it.  We can always follow-up with a test for that case if we run into it again.  The tl;dr for contributors to this repo is don't reuse the same member names for properties and other members.

leofang:
Update: We decided to abandon this approach since it is a bit more brittle that it should have been for working around the upstream (CPython) bug.
- The bug is now fixed and being backported: https://github.com/python/cpython/pull/149577
- The tests to capture the issue is hard to get robust (either regex or rlcompleter)

We'd prefer #2055 instead. Closing.

Issue #2050, opened 2026-05-07
__eq__ operator overloading should be prepared to accept any type


mdboom:
Some of the `__eq__` operators in cuda_core assuming that the rhs side will be the same type as the lhs.  This will result in raising an `AttributeError` (or worse) rather than returning `False`.

We need to do something like this (across the whole codebase):

```diff
-    def __eq__(self : _StridedLayout, other : _StridedLayout) -> bool:
-        return self.itemsize == other.itemsize and self.slice_offset == other.slice_offset and _base_layout_equal(self.base, other.base)
+    def __eq__(self, other) -> bool:
+        if not isinstance(other, _StridedLayout):
+            return NotImplemented
+        cdef _StridedLayout _other = <_StridedLayout>other
+        return self.itemsize == _other.itemsize and self.slice_offset == _other.slice_offset and _base_layout_equal(self.base, _other.base)
```

(PS: Found as part of adding type checking)

Issue #2049, opened 2026-05-07
Bug in `_grow_allocation_fast_path`


mdboom:
I'm ramping up `.pyi` generation, which lead to type checking our code for the first time, and it's discovered a potential bug at this line:

https://github.com/NVIDIA/cuda-python/blob/main/cuda_core/cuda/core/_memory/_virtual_memory_resource.py#L338

```
cuda_core/cuda/core/_memory/_virtual_memory_resource.py:338: error: "Buffer" has no attribute "_size"; maybe "size"?
```

Indeed, `Buffer` does not expose `_size` to Python (and `_virtual_memory_resource` is Python, not Cython).

```
(Pdb) p buffer._size
*** AttributeError: 'cuda.core._memory._buffer.Buffer' object has no attribute '_size'. Did you mean: 'size'?
(Pdb) buffer._size = 42
*** AttributeError: 'cuda.core._memory._buffer.Buffer' object has no attribute '_size' and no __dict__ for setting new attributes. Did you mean: 'size'?
```

It doesn't look like we have any test coverage for this file.

We probably need to make `Buffer.size` a read/write property.

leofang:
Interesting, this path (`_grow_allocation_fast_path`) should be covered by this test
https://github.com/NVIDIA/cuda-python/blob/2ddbe6bcf40d394981300a47b85be7666109fe0a/cuda_core/tests/test_memory.py#L950
not sure what's missing.

Issue #468, opened 2025-02-24
Avoid relying on `typing.TYPE_CHECKING` to the maximum possible extent


leofang:
During investigation of #454 and related doc rendering issues, it has come to my attention that the type checking detection logic does not work well with Sphinx (ex: https://github.com/sphinx-doc/sphinx/issues/13137, https://github.com/sphinx-doc/sphinx/issues/11225). 

Independently, Ralf has pointed out that it makes code debugging harder (https://github.com/NVIDIA/cuda-python/pull/463#discussion_r1967029167).

The idea of having a `if TYPE_CHECKING:` check is to avoid unnecessary imports or circular imports. If such issues do not exist, we should avoid using it.

mdboom:
In the process of completing #2061 and getting type checking working as part of pre-commit, it's clear that `TYPE_CHECKING` is still necessary in some cases to avoid circular imports.  I agree we should only use it when absolutely necessary, but I don't think there is any way to avoid it fully.

The hypothesis that it prevents Sphinx from linking things seems to be untrue.

PR #2061, opened 2026-05-11
Add automatically-generated .pyi files to cuda_core

mdboom:
This adds automatically-generated `.pyi` files to cuda_core.  This allows IDE auto-completion to work (which is also used by IDE-integrated coding agents).  This has also found 2 real bugs in our code already (#2049 and #2050).  The ability to catch a certain class of bugs with this will be really helpful going forward, especially since our linting abilities with `cython-lint` are a bit behind what they are in pure Python.

(Marked as a draft because it requires a small patch to stubgen-pyx and is currently using my personal fork.  We can update this after the patch, or something like it, is merged -- https://github.com/jon-edward/stubgen-pyx/pull/19)

## Reviewing

Three separate commits to make this easier to review:

- Infrastructure / configuration updates to integrate the tools
- Updates to the type annotations to make everything consistent and pass type-checking
- Adding all of the generated .pyi files

## Workflow

The workflow I chose here is that `.pyi` files are committed explicitly into the repo.  A pre-commit hook regenerates them and then runs a type checker over the entire result.

**Alternative:** An alternative approach would be to not store them in the repo: generate them at the time of wheel building, and then test them as part of the test suite.

I prefer the approach taken here, because it will make it obvious in a PR when our public API has changed, and we can validate that it is making an allowed transformation (adding, but not removing etc.)  We could (as a future step), use something like [griffe](https://mkdocstrings.github.io/griffe/).

## Type-checking

This adds mypy-based type checking as part of the pre-commit hook.  I ran this code over all of the major type checkers (mypy, pyright, pyrefly and ty).  mypy and pyright were very similar in behavior, we could use either, but given that pathfinder is already using mypy and I have more confidence in its long-term viability, I went with that.  pyrefly seems worth following up on in a year, and ty was pretty limited at this point.  It takes 0.25s to run mypy over cuda-core on my machine, so I don't think moving to a Rust-based checker matters much at this point.

## Assessment

Overall, I was very impressed by how far `stubgen-pyx` has come since last I looked, and I think it's good enough to be relied on at this point.

One minor rough edge is with code at the top-level.  Code that calls `cdef` functions at the top-level will not type check/resolve, because (correctly) those `cdef` functions are not exposed to the `.pyi` file.  The solution I used here was just to put these blocks of code at the top level into functions and call them, which is probably not a bad practice in terms of global namespace hygiene anyway.

The biggest shortcoming I see with this approach is that the line numbers in the `.pyi` files don't map back to the original line numbers in the `.pyx` or `.pxi` files.  So when mypy displays a type error, you have to weave through and find the original source to edit it.  (This is sort of what the `#line` directive in the C preprocessor is for -- I don't think `.pyi` files have an equivalent).  I think we can live with that, but it's a bit of a pain point.  All that said, my agent had no trouble realizing it had to go back and make changes to the `.pyx` file, not directly in the `.pyi` file.

Cc: @jon-edward (for viz)


leofang:
cc @ZzEeKkAa for vis

jon-edward:
Hello! The patch that @mdboom mentioned is merged, and is included in stubgen-pyx v0.2.6 which is now available via pip

mdboom:
pre-commit.ci autofix

cuda_core/cuda/core/_memory/_buffer.pyx
```diff
@@ -28,16 +28,22 @@ from cuda.core._stream cimport Stream, Stream_accept, default_stream
...
 if sys.version_info >= (3, 12):
     from collections.abc import Buffer as BufferProtocol
-else:
-    BufferProtocol = object
 
 from cuda.core._dlpack import classify_dl_device, make_py_capsule
 from cuda.core._device import Device
 
+if TYPE_CHECKING:
+    from cuda.core.graph import GraphBuilder
```
Andy-Jost, line 38: Previously @leofang mentioned he wanted to remove all `if TYPE_CHECKING` blocks (but I don't recall the reason). Is that no longer the guidance? It appears real imports are needed by the type checking tools.

cuda_core/cuda/core/_memory/_peer_access_utils.pyx
```diff
@@ -4,9 +4,9 @@
 
 from __future__ import annotations
 
-from collections.abc import Callable, Iterable, MutableSet
+from collections.abc import Callable, Iterable, MutableSet, Set as AbstractSet
```
Andy-Jost: Does the distinct name `AbstractSet` buy anything?

cuda_core/cuda/core/_memory/_virtual_memory_resource.py
```diff
@@ -335,7 +343,9 @@ def _grow_allocation_fast_path(
             trans.commit()
 
         # Update the buffer size (pointer stays the same)
-        buf._size = new_size
+        # TODO: #2049 This is a real bug, accessing _size which doesn't exist.
+        # Fix bug and remove the "type: ignore[attr-defined]" comment.
+        buf._size = new_size  # type: ignore[attr-defined]
```
Andy-Jost, line 348: This seems like a good enough reason to Cythonize even when we don't expect performance benefits.

cuda_core/cuda/core/_memory/_virtual_memory_resource.py
```diff
@@ -559,20 +569,25 @@ def allocate(self, size: int, *, stream: Stream | None = None) -> Buffer:
...
         size : int
             The size in bytes of the memory to deallocate.
         stream : Stream, optional
             Keyword-only. Unused because virtual memory operations are
             synchronous.
         """
+        if ptr is None:
+            ptr = 0
+        else:
+            ptr = int(ptr)
```
Andy-Jost: nit: `ptr = 0 if ptr is None else int(ptr)` or even `ptr = int(ptr or 0)` since we can rely on `ptr` being in `driver.CUdeviceptr | int | None`

Andy-Jost (approved):
No issues here. I checked the `.pyx` files carefully but didn't read the generated `.pyi` files.

cuda_core/cuda/core/_memory/_buffer.pyx
mdboom: Yes.  Even if you want to use the quoted syntax, i.e. `def my_function() -> "GraphBuilder":`, you need to import those things so the type checker knows what you are referring to.  The `if TYPE_CHECKING:` block is just a way to prevent cyclical imports that this otherwise might cause.

cuda_core/cuda/core/_memory/_peer_access_utils.pyx
mdboom: Probably not.  The agent did this.  On reflection it's probably clearer to not rename.  I can update the PR.

cuda_core/cuda/core/_memory/_virtual_memory_resource.py
mdboom: Or, conversely, it's a reason not to Cythonize because the type checker would not have been able to catch this bug if it were Cython 😄 

But it's all tradeoffs.  We tend to take runtime performance over developer productivity in this project, which is reasonable.

cuda_core/cuda/core/_memory/_peer_access_utils.pyx
mdboom: I think the agent did this because it's the convention in typeshed:

https://github.com/search?q=repo%3Apython%2Ftypeshed%20abstractset&type=code

Not an argument that we need go either way.  It's functionality the same thing, and personally I don't like to rename things unless it's to workaround a clash or something.

cuda_core/cuda/core/_utils/cuda_utils.pyx
```diff
@@ -12,12 +12,21 @@ from collections.abc import Sequence
...
-    from cuda import cuda as driver
-    from cuda import cudart as runtime
-    from cuda import nvrtc
+# TODO: Are we sure we don't need this fallback anymore?
+
+# (Previously wrapped in try/except ImportError for the legacy
+# `from cuda import cuda as driver` etc. import path.)
+# `as X` form is the PEP 484 explicit re-export marker, which type checkers
+# need to treat these names as part of the public API of this module.
+from cuda.bindings import driver as driver, nvrtc as nvrtc, runtime as runtime
```
mdboom, line 15: Question for @leofang: Do we still need the fallback to the pre-cuda-bindings library?

```
from cuda import nvrtc
```

It makes the type-checking a lot more complicated.  There is probably a solution, but I didn't want to add it if we can just drop this because `cuda_bindings` has been out long enough by now.

cuda_core/cuda/core/_memory/_buffer.pyx
mdboom: There was mention elsewhere of [this bug](https://github.com/NVIDIA/cuda-python/issues/468) and concern that the introduction of `if TYPE_CHECKING:` blocks will have negative consequences for the docs.

This PR only adds new imports inside of `if TYPE_CHECKING:` blocks. It does move anything from the top-level to these blocks, so nothing that was previously visible is now hidden.

That said, it is clear that this PR actually makes things better, so I think Sphinx is actually picking up references that it wasn't before.  For example, looking at the [upstream documentation for GraphDefinition](https://nvidia.github.io/cuda-python/cuda-core/latest/generated/cuda.core.graph.GraphDefinition.html#cuda.core.graph.GraphDefinition), the return types are not linked:

image

With [this PR](https://nvidia.github.io/cuda-python/pr-preview/pr-2061/cuda-core/latest/generated/cuda.core.graph.GraphDefinition.html#cuda.core.graph.GraphDefinition), they are:

image

I think if anything, this helps with Sphinx and mypy here.

cuda_core/cuda/core/_utils/cuda_utils.pyx
leofang: We can purge them now

PR #2067, opened 2026-05-12
Fix #2050: Fix __eq__ implementations

### LINKED ISSUE #2050: __eq__ operator overloading should be prepared to accept any type

mdboom:
This bug was found with the static type-checking in #2061.

`__eq__` overrides have to handle /any/ type being passed into them, and should not blindly access members of what is passed in.

### DIFF
--- a/cuda_core/cuda/core/_event.pyx
+++ b/cuda_core/cuda/core/_event.pyx
@@ -330,9 +330,11 @@ cdef class IPCEventDescriptor:
         self._is_blocking_sync = is_blocking_sync
         return self
 
-    def __eq__(self, IPCEventDescriptor rhs):
+    def __eq__(self, other) -> bool:
+        if not isinstance(other, IPCEventDescriptor):
+            return NotImplemented
         # No need to check self._is_blocking_sync.
-        return self._reserved == rhs._reserved
+        return self._reserved == (<IPCEventDescriptor>other)._reserved
 
     def __reduce__(self):
         return IPCEventDescriptor._init, (self._reserved, self._is_blocking_sync)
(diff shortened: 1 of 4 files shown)


rwgk (approved):

___

# Findings

- No issues found. For the scope of PR 2067, the tests look reasonably complete.
- `cuda_core/tests/test_event.py` covers both sides of the `IPCEventDescriptor.__eq__` fix: unrelated-type comparisons now follow Python's rich-comparison protocol, and same-type comparisons still preserve the intended value semantics.
- `cuda_core/tests/test_strided_layout.py` does the same for `_StridedLayout`: it exercises `==`, `!=`, reverse comparison, and the direct `__eq__` `NotImplemented` contract, plus a same-type equality/inequality smoke test.

# Assumptions

- `cuda_core/tests/test_object_protocols.py` does not include `IPCEventDescriptor` or `_StridedLayout` in the shared equality matrix, so these focused tests are the primary regression coverage for those two types.
- That looks acceptable for this PR because the production change is narrowly scoped to those two `__eq__` implementations.

# Verification

- Ran `TestVenv/bin/python -m pytest cuda_core/tests/test_event.py -q` and got `28 passed, 2 skipped`.
- Ran `TestVenv/bin/python -m pytest cuda_core/tests/test_strided_layout.py -q` and got `689 passed`.
- The added coverage matches the regression described in issue `#2050` and also checks that same-type equality behavior remains intact.

kkraus14 (approved):

PR #2068, opened 2026-05-12
Fix #2049: Expose Buffer._size to Python

### LINKED ISSUE #2049: Bug in `_grow_allocation_fast_path`

mdboom:
`_grow_allocation_fast_path` was not being tested in CI, and static type-checking discovered an attribute error lurking within it.

This new test is entirely agent-written and I don't personally understand this part of the system, so please give it extra scrutiny.

The bugfix itself is simple, but also could be removed when/if `_virtual_memory_resource.py` is Cythonized, since it could `cimport` Buffer and get access to `_size` that way.

### DIFF
--- a/cuda_core/cuda/core/_memory/_buffer.pxd
+++ b/cuda_core/cuda/core/_memory/_buffer.pxd
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # SPDX-License-Identifier: Apache-2.0
 
@@ -18,13 +18,16 @@ cdef struct _MemAttrs:
 cdef class Buffer:
     cdef:
         DevicePtrHandle _h_ptr
-        size_t          _size
         MemoryResource  _memory_resource
         object          _ipc_data
         object          _owner
         _MemAttrs       _mem_attrs
         bint            _mem_attrs_inited
         object          __weakref__
+    cdef public:
+        # Python code in _memory/_virtual_memory_resource.py needs to update
+        # this value, though it is technically private.
+        size_t          _size
 
 
 cdef class MemoryResource:
--- a/cuda_core/tests/test_memory.py
+++ b/cuda_core/tests/test_memory.py
@@ -1,6 +1,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+import contextlib
 import ctypes
 import sys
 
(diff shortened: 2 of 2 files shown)


cuda_core/cuda/core/_memory/_buffer.pxd
```diff
@@ -18,13 +18,16 @@ cdef struct _MemAttrs:
...
         MemoryResource  _memory_resource
         object          _ipc_data
         object          _owner
         _MemAttrs       _mem_attrs
         bint            _mem_attrs_inited
         object          __weakref__
+    cdef public:
+        # Python code in _memory/_virtual_memory_resource.py needs to update
+        # this value, though it is technically private.
+        size_t          _size
```
leofang, line 30: This seems odd. Didn't we already expose `Buffer.size`?
https://nvidia.github.io/cuda-python/cuda-core/latest/generated/cuda.core.Buffer.html#cuda.core.Buffer.size

leofang:
Do we know why this did not trip in the CI? https://github.com/NVIDIA/cuda-python/issues/2049#issuecomment-4400974363

mdboom:
> Do we know why this did not trip in the CI? [#2049 (comment)](https://github.com/NVIDIA/cuda-python/issues/2049#issuecomment-4400974363)

My agent's assessment:

vmm_mr.allocate(2 * 1024 * 1024) reserves only that 2 MB range; nothing guarantees the next VA page is unreserved. In practice the CUDA driver's cuMemAddressReserve with a fixedAddr hint typically returns CUDA_SUCCESS but with a different address (the hint is advisory). That trips the new_ptr != int(buf.handle) + aligned_prev_size branch, the reservation is freed, and execution falls through to _grow_allocation_slow_path.

The test author already noticed this — the comment at [test_memory.py:955-956](vscode-webview://1lq1g48ng0sclg835m7ufv5pq987ujci9cjh6dj4ho9llgobp0ub/cuda_core/tests/test_memory.py#L955-L956) literally says "Because of the slow path, the pointer may change … we can assert that a new pointer was assigned". So the test was written knowing only the slow path runs.

cuda_core/cuda/core/_memory/_buffer.pxd
mdboom: We could make that property settable (it's currently readonly), but I assumed we didn't want to make settabilility part of the public API -- seems kinda dangerous. So I exposed `_size` as a `_`-prefixed member instead.

Andy-Jost (approved):
LGTM!

I don't particularly like the VVM interface forcing `Buffer` to be mutable but that's not in scope for this PR.

mdboom:
> I don't particularly like the VVM interface forcing Buffer to be mutable but that's not in scope for this PR.

Should we create a new issue to follow-up for that?  I'm not as familiar with this code so I don't know what that would look like...

PR #2072, opened 2026-05-12
Follow-up to #2065: A couple of additional test fixes

mdboom:


cuda_bindings/tests/test_cuda.py
```diff
@@ -560,29 +559,6 @@ def test_get_error_name_and_string():
     assert s == b"CUDA_ERROR_INVALID_DEVICE"
 
 
-@pytest.mark.skipif(not callableBinary("nvidia-smi"), reason="Binary existence needed")
```
mdboom, line 563: This test is redundant to this one: https://github.com/NVIDIA/cuda-python/blob/main/cuda_bindings/tests/nvml/test_cuda.py#L57

(Hinted at already in the comment here).

Just removing this seems like the better choice.

rwgk (approved):
I verified that all Tegra Thor NVML-related errors are resolved with this PR.

PR #2074, opened 2026-05-12
cuda.core: harden graph user-object destructor during Python shutdown

### LINKED ISSUE #2042: Harden _py_host_destructor against invocation after Py_Finalize
## Summary

`_py_host_destructor` (`cuda_core/cuda/core/graph/_utils.pyx`) is the destroy callback we attach to CUDA user objects that hold Python references for graph-resource lifetime. It is declared `noexcept with gil` and unconditionally calls `Py_DECREF`:

```python
cdef void _py_host_destructor(void* data) noexcept with gil:
    _py_decref(data)
```

It is attached via `_attach_user_object` with the `CU_USER_OBJECT_NO_DESTRUCTOR_SYNC` flag, which explicitly allows CUDA to invoke the destructor asynchronously on a worker thread, decoupled from `cuGraphDestroy`.

This creates a window during interpreter shutdown:

1. Python starts shutdown; `GraphDefinition.__dealloc__` runs and calls `cuGraphDestroy`.
2. CUDA queues the destructor for the user object on a worker thread.
3. `Py_Finalize` completes.
4. The CUDA worker later runs `_py_host_destructor` -> `with gil` calls `PyGILState_Ensure` after the runtime is gone -> undefined behavior, typically a crash.

In practice this is usually masked because CUDA tends to run the destructor synchronously inside `cuGraphDestroy`, but the contract permits the bad ordering and the codebase should not depend on the lucky timing.

## Affected callers

All current users of `_py_host_destructor`:

- Host callbacks attached via `_attach_host_callback_to_graph` (Python callable and ctypes `CFuncPtr` paths, plus bytes-backed `user_data`) — pre-existing.
- Kernel-argument lifetime attachment added in #2039 / PR #2041.

## Proposed fix

Guard the decref with `Py_IsInitialized()`:

```python
cdef extern from \"Python.h\":
    int Py_IsInitialized()

cdef void _py_host_destructor(void* data) noexcept with gil:
    if Py_IsInitialized():
        _py_decref(data)
    # else: process is exiting; the OS will reclaim everything.
```

An alternative is to drop `CU_USER_OBJECT_NO_DESTRUCTOR_SYNC` for Python-typed user objects so destructors always run synchronously inside `cuGraphDestroy` (where we are guaranteed to hold the GIL). That is safer but may have performance implications and changes existing semantics for the host-callback path; the `Py_IsInitialized()` guard is the smaller, safer change.

## Context

This should be fixed in the context of #1330. The broader graph-lifetime/update work tracked there is the natural place to review the user-object lifetime model end to end.

aryanputta:
## Summary

Closes #2042.

`_py_host_destructor` unconditionally entered `with gil` before calling
`Py_DECREF`. That is fine during normal runtime, but it is unsafe in the
graph user-object path because CUDA may invoke the destructor
asynchronously after interpreter finalization has begun.

This change makes `_py_host_destructor` `nogil`, checks
`Py_IsInitialized()`, and only enters a GIL section when Python is still
initialized.

## Changes

- `cuda_core/cuda/core/graph/_utils.pyx`: declare `Py_IsInitialized()` and
  harden `_py_host_destructor` so it only acquires the GIL when Python is
  still initialized.
- `cuda_core/cuda/core/graph/_utils.pxd`: update the destructor signature
  to match.

The normal runtime path is unchanged: if Python is still alive, the
destructor still decrefs the attached object. The change only affects the
shutdown edge where acquiring the GIL is no longer safe.

## Validation

Locally verified:

- `git diff --check` passes.
- The change is limited to the existing graph user-object destructor path in
  `_utils.pyx` / `_utils.pxd`.
- The commit is DCO-signed and SSH-signed.

Not fully verified on this machine:

- full `cuda.core` test suite
- editable build of `cuda_core`
## Related Work

- Closes #2042.
- This is the hardening follow-up called out in #2041 after graph
  user-object lifetime handling was extended to kernel arguments.


cuda_core/cuda/core/graph/_utils.pyx
```diff
@@ -14,15 +14,18 @@ from cuda.core._utils.cuda_utils cimport HANDLE_RETURN
...
 
 
 cdef void _py_host_trampoline(void* data) noexcept with gil:
     (<object>data)()
 
 
-cdef void _py_host_destructor(void* data) noexcept with gil:
-    _py_decref(data)
+cdef void _py_host_destructor(void* data) noexcept nogil:
+    if Py_IsInitialized():
```
rwgk: Mostly generated with Cursor GPT-5.4 Extra High Fast, after a few rounds of prompting and having it dig into the cpython sources (git clone).

__

I think this should also gate on **finalization**, not just **initialization**.

Something along these lines seems like the best fit here:

```cython
cdef extern from *:
    """
    #include <Python.h>

    static inline int _py_runtime_is_finalizing(void) {
    #if PY_VERSION_HEX >= 0x030D0000
        return Py_IsFinalizing();
    #else
        return _Py_IsFinalizing();  /* CPython-private, available in 3.10-3.12 */
    #endif
    }
    """
    int _py_runtime_is_finalizing() nogil

cdef void _py_host_destructor(void* data) noexcept nogil:
    if data == NULL:
        return

    # Leak once shutdown has started or completed.
    if not Py_IsInitialized():
        return
    if _py_runtime_is_finalizing():
        return

    with gil:
        _py_decref(data)
```

I inspected the CPython sources before suggesting this. The key detail is that during shutdown CPython marks the runtime as finalizing **before** it clears the `initialized` flag:

`Python/pylifecycle.c` lines 2077-2081:

```
    /* Remaining daemon threads will be trapped in PyThread_hang_thread
       when they attempt to take the GIL (ex: PyEval_RestoreThread()). */
    _PyInterpreterState_SetFinalizing(tstate->interp, tstate);
    _PyRuntimeState_SetFinalizing(runtime, tstate);
    runtime->initialized = 0;
```

In other words, there is a real source-level window where `Py_IsInitialized()` is still true even though attempting to enter Python is already unsafe. In the cpython source tree (git tag: `v3.14.4`) this ordering is not only visible in `Python/pylifecycle.c`, but the docs in `Doc/c-api/interp-lifecycle.rst` / `Doc/c-api/threads.rst` also explicitly describe GIL/thread-state attachment during finalization as unsafe.

That means the current proposed change:

```cython
cdef void _py_host_destructor(void* data) noexcept nogil:
    if Py_IsInitialized():
        with gil:
            _py_decref(data)
```

still leaves a shutdown race: the destructor can observe `Py_IsInitialized() == 1`, then finalization can begin, and then `with gil` can run into CPython's shutdown path.

Why I think the version above is the best we can do:

* It preserves the desired behavior during normal operation: we still decref instead of leaking.
* It matches the stated shutdown policy: once finalization has started (or Python is already gone), we intentionally leak rather than trying to force a decref through a fragile runtime state.
* It closes the specific `initialized == true but runtime is already finalizing` hole that the current patch still has.
* For `3.10+`, it is implementable today: `Py_IsFinalizing()` is public in `3.13+`, and CPython exposes `_Py_IsFinalizing()` in `3.10`-`3.12`.

Remaining risk: this is still only **best effort**, not a proof of safety. There is still an unavoidable race between the finalization check and `with gil`. If finalization starts in that tiny window, CPython may still take the shutdown path when the thread tries to re-enter Python. From the CPython sources, that outcome is version-dependent: older versions can terminate the thread, while newer versions can hang it. I do not think we can eliminate that residual risk here without either (1) deferring cleanup onto a known Python-owned execution context, or (2) leaking more aggressively before finalization. Considering both of those as out of scope, this looks like the narrowest and most defensible compromise.

aryanputta: Got it, that makes sense. I was only gating on Py_IsInitialized(), but I see the issue now: there is a shutdown window where Python is still initialized while the runtime is already finalizing, so entering the GIL can still be unsafe. I’ll update the destructor to also check finalization and intentionally leak once shutdown has started. Thanks for the quick responce!

cuda_core/cuda/core/graph/_utils.pyx
```diff
@@ -14,15 +14,42 @@ from cuda.core._utils.cuda_utils cimport HANDLE_RETURN
...
+    static inline int _py_runtime_is_finalizing(void) {
+    #if PY_VERSION_HEX >= 0x030D0000
+        return Py_IsFinalizing();
+    #else
+        return _Py_IsFinalizing();
+    #endif
+    }
+    """
+    int _py_runtime_is_finalizing() nogil
+
```
Andy-Jost: This replicates machinery that appears in `_cpp/resource_handles.cpp`. It would be better to add something like `py_object_user_object_destroy` to that layer and use it here. That way, the Python finalization checks would appear in only one place.

rwgk (approved):
This looks good to me.

It’d be nice to add a comment like this in `cuda_core/cuda/core/_cpp/resource_handles.hpp` next to the existing `py_is_finalizing()` helper. It documents an important subtlety from CPython shutdown semantics: the check is only best-effort, and there is still a narrow race where non-finalizer threads can be killed on older CPython or hang on newer CPython. That means teardown in this layer can still be cut short in rare cases, so the guard reduces the risk but does not fully eliminate it.

```
  // Best-effort probe for interpreter shutdown.
  //
  // In CPython this is not a hard guarantee: finalization can begin after this
  // returns false but before a later PyGILState_Ensure() / other Python C-API
  // call.
  //
  // If that race is lost on a non-finalizer thread, CPython's behavior is
  // version-dependent: on older supported versions (3.10-3.13) it may abruptly the
  // current thread (historically via PyThread_exit_thread(), i.e. without normal
  // C++ unwinding), while on newer versions (3.14+) it may hang the thread until
  // process exit.
  //
  // We still use this check because the policy in this layer is to avoid Python
  // work once shutdown is underway and accept an intentional leak / skipped
  // Python conversion in that edge case rather than add more complex deferral
  // machinery.
  inline bool py_is_finalizing() noexcept {
  #if PY_VERSION_HEX >= 0x030D0000
      return Py_IsFinalizing();
  #else
      return _Py_IsFinalizing() != 0;
  #endif
  }
```

rwgk (approved):
Thanks!

aryanputta:
@rwgk, I’ve updated the resource handles to include the suggested comments on shutdown semantics. I believe all feedback has been addressed; please let me know if any additional modifications are needed before merging.

rwgk:
This looks ready to merge to me, but @Andy-Jost is the original author of the changed code, I'm waiting to give him a chance to approve, too.

rwgk:
@aryanputta at this stage it's best if you don't merge main anymore, unless the UI here reports conflicts.

Each time you merge main, we have to rerun the CI again.

aryanputta:
> @aryanputta at this stage it's best if you don't merge main anymore, unless the UI here reports conflicts.
> 
> 
> 
> Each time you merge main, we have to rerun the CI again.

Okay, sorry!

Andy-Jost (approved):
LGTM! Thanks for the contribution!

leofang:
I think the WSL failures are fixed in https://github.com/NVIDIA/cuda-python/pull/2118 later the same day (after the CI was triggered). Let's retry.

leofang:
/ok to test [78e86b1](https://github.com/NVIDIA/cuda-python/pull/2074/commits/78e86b154dc254b5404fc783533d1f244d980f6a)

aryanputta:
On the latest run (78e86b1) the only failure is `Test win-64 / Python 3.12, CUDA 12.9.1 (wheels), GPU l4 (MCDM)` (plus the `Check job status` rollup that reflects it). It failed in the **Set up Python 3.12** step with `Version 3.12 was not found in the local cache`, before any test ran (every test step is `skipped`). The other 93 jobs in the matrix are green, so this looks like a transient self-hosted-runner cache miss rather than anything from this PR.

I don't have permission to re-run individual jobs here. @leofang could you re-run just that job (or re-trigger with `/ok to test`) when convenient? Happy to push an empty commit instead if that's easier. Thanks!

leofang:
As Ralf said (and I noted elsewhere: https://github.com/NVIDIA/cuda-python/pull/2086#issuecomment-4664776748),
> Each time you merge main, we have to rerun the CI again.

leofang:
/ok to test [2e91fc8](https://github.com/NVIDIA/cuda-python/pull/2074/commits/2e91fc83cae4a1e6babb66921511c615a2624f19)

PR #2076, opened 2026-05-13
BUG: Handle CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM correctly

mdboom:
As @leofang pointed out [here](https://github.com/NVIDIA/cuda-python/pull/1972/files/9bb9f5a62a625d0fe117349b67de10d9c62d43c3#diff-05f9f01058c2371b760076824cb01359afd9373333cbd2c480815ce2bf385974)...

Per-thread default stream would be enabled if `CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM` was set to any value, including '0'.

### DIFF
--- a/cuda_bindings/cuda/bindings/_bindings/cydriver.pyx.in
+++ b/cuda_bindings/cuda/bindings/_bindings/cydriver.pyx.in
@@ -519,7 +519,7 @@ cdef int _cuPythonInit() except -1 nogil:
     cdef char libPath[260]
 
     with gil, __symbol_lock:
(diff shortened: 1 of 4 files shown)


kkraus14 (approved):

kkraus14:
@mdboom defer to you to merge since I imagine this requires syncing a change in the backend codegen

mdboom:
> @mdboom defer to you to merge since I imagine this requires syncing a change in the backend codegen

Yep.  If somebody could approve the backend codegen (which in this case is very simple), I'll merge both at the same time.

PR #2078, opened 2026-05-13
Fix #2019: Move Python-compatibility shims to their own file

### LINKED ISSUE #2019: Handle Python backward compatibility shims in a single file
As suggested in #2016 and #2014, we have some repeated backward compatibility shims that import either from the stdlib or backward compat libraries and we maybe should move this into a centralized module to make it easier to remove in the future.

mdboom:
A bunch of robo-reviews kept suggesting this.  It's probably a good practice / convention.


cuda_core/cuda/core/_utils/pycompat.py
leofang, line 1: nit: Move this to `cuda/core/_utils`?

kkraus14 (approved):

PR #2080, opened 2026-05-13
Fix #2079: Add PyTorch testing in some CI configs

### LINKED ISSUE #2079: Add pytorch to one CI configuration to increase test coverage
None of our tests that depend on pytorch are currently covered in our CI.  It's a heavy dependency to add /everywhere/ but it would be useful to add it to at least one configuration.

mdboom:
Added the ability to specify extra packages to install before running the tests, and then added `torch` to a couple of configurations.

This is part of reducing the number of tests we have that never run anywhere in our CI.

Will need to compare the job runtimes before and after this change to determine cost/benefit.

### DIFF
--- a/.github/workflows/test-wheel-linux.yml
+++ b/.github/workflows/test-wheel-linux.yml
@@ -77,7 +77,7 @@ jobs:
           echo "OLD_BRANCH=${OLD_BRANCH}" >> "$GITHUB_OUTPUT"
 
   test:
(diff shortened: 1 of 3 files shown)


leofang:
Would you mind letting me finish this in #1987 ?

leofang:
(fixed typo)

mdboom:
> Would you mind letting me finish this in #1987 ?

Sure.  Sorry, I just missed that #1987 existed.  I'll close this.

leofang:
No worries! I do agree we should test against PyTorch and #1987 actually helped catch bugs!

PR #2086, opened 2026-05-14
Validate checkpoint GPU UUID inputs early

aryanputta:
## Summary
- reject non-UUID gpu_mapping values in cuda.core.checkpoint before they reach driver-facing restore args
- normalize malformed UUID string failures to the documented ValueError path
- add source-level checkpoint helper tests that run without importing the built cuda.core package

## Testing
- python3 -m pytest cuda_core/tests/test_checkpoint_helpers.py --noconftest


cuda_core/tests/test_checkpoint_helpers.py
leofang, line 1: Hi @aryanputta as you might have seen in your other PRs, we generally do not accept mock tests like this. Please make sure new/revised tests fit in the existing test style, and GPU tests actually pass locally before submitting.

leofang (changes requested):

cuda_core/cuda/core/checkpoint.py
```diff
@@ -239,14 +239,21 @@ def _as_cuuuid(driver, value, buffers):
     the ``"xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"`` format returned by
     :attr:`Device.uuid`.
     """
+    if isinstance(value, driver.CUuuid):
+        return value
```
leofang, line 238: Actually, reading this more carefully: What is the motivation for accepting `CUuuid` here? Do you have a use case?

Note that `Device.uuid` returns a string:
https://nvidia.github.io/cuda-python/cuda-core/latest/generated/cuda.core.Device.html#cuda.core.Device.uuid

aryanputta:
@leofang You were right on both points here. 
What I changed in commit 083e41112a:
- removed the mock-only cuda_core/tests/test_checkpoint_helpers.py
- narrowed _as_cuuuid() so restore mappings now accept UUID strings only
- updated the docs to say gpu_mapping should use the strings returned by Device.uuid

On your question specifically... I dont' have a strong public facing use case for accepting CUuuid objects here. The intended caller-facing path is the string UUID returned by Device.uuid, and the existing real checkpoint tests already follow that contract. For example, cuda_core/tests/test_checkpoint.py builds migration mappings from devices[i].uuid in _build_rotation_mapping() and runs the driver-backed checkpoint/restore scenarios without mocks.

cuda_core/cuda/core/checkpoint.py
```diff
@@ -235,18 +235,22 @@ def _make_restore_args(driver, gpu_mapping: _Mapping[_Any, _Any] | None):
...
+            raise ValueError(
+                f"GPU UUID string must be 32 hex characters (with optional hyphens), got {value!r}"
+            ) from None
         if len(raw) != 16:
             raise ValueError(f"GPU UUID string must be 32 hex characters (with optional hyphens), got {value!r}")
         buf = _ctypes.create_string_buffer(raw, 16)
         buffers.append(buf)
         return driver.CUuuid(_ctypes.addressof(buf))
-    return value
+    raise TypeError("GPU UUID values must be UUID strings")
```
leofang: Apology @aryanputta, I missed that we already had accepted `driver.CUuuid`. Now that cuda-core is stable, we cannot break the backward compatibility anymore despite this is the right thing to do. I think your previous commit (having two separate `isinstance` checks) makes sense, would you mind restoring it?

aryanputta:
@leofang I pushed the requested CUuuid compatibility restore in 3261c10. The metadata check is still failing because the PR has no milestone, and GitHub does not let me set one on the upstream PR. Could you set the milestone to `cuda.core v1.1.0` when you get a chance?

leofang (approved):

leofang:
/ok to test [3261c10](https://github.com/NVIDIA/cuda-python/pull/2086/commits/3261c10a17e1a0b003f7974dc423e14d25af4b73)

aryanputta:
@leofang I repushed this as 1520ee3 to retrigger CI because I do not have permission to rerun failed jobs on the upstream repo. I inspected the previous run before repushing: the failures looked unrelated to the checkpoint change (aarch64 sccache/GitHub cache 503, WSL NVML process-name UnicodeDecodeError, Windows jit_lto_fractal driver/backend mismatch, and one IPC flake). Could you approve the new SHA with /ok to test when convenient?

aryanputta:
@leofang quick update: I moved this branch to current main as 41507c5 after confirming the previous WSL get_process_name failures were covered by #2118. The effective diff against current main is still only cuda_core/cuda/core/checkpoint.py. Could you approve the new SHA with /ok to test when convenient?

leofang:
@aryanputta plz be patient and wait for one of us to check the CI. We can trigger re-runs of the failing jobs only, but if you merge with main we have to start from beginning and run all jobs, not just the failing one.

leofang:
Also our CI is known flakey for a few things and rerun usually solves the problem.

aryanputta:
> @aryanputta plz be patient and wait for one of us to check the CI. We can trigger re-runs of the failing jobs only, but if you merge with main we have to start from beginning and run all jobs, not just the failing one.

Sorry about that, @leofang! I was told in the past not to force-push/rebase on shared branches to avoid disrupting reviewers, so I thought moving it to main would be the cleaner approach. I will definitely keep a note for myself to just leave it be and wait for a maintainer to trigger the specific job reruns next time.

Also, just wanted to let you know that I fixed the other PR #2074 we were working on whenever you have some time to check it out.

Thank you!

leofang:
> I was told in the past not to force-push/rebase on shared branches to avoid disrupting reviewers

That is correct, for the very same reasons: When the CI pipelines are triggered, it's based on a particular snapshot (commit). Any action that results in a new commit (ex: merge with main, rebase, force-push, ...) would forfeit the opportunity of re-running.

PR #2092, opened 2026-05-15
test: skip IPC peer access tests when IPC is unavailable

rluo8:
nvbug 6176793

Problem:
FAILED memory_ipc\test_peer_access.py::TestPeerAccessNotPreservedOnImport::test_main - RuntimeError: IPC is not available on Windows
FAILED memory_ipc\test_peer_access.py::TestBufferPeerAccessAfterImport::test_main[True] - RuntimeError: IPC is not available on Windows
FAILED memory_ipc\test_peer_access.py::TestBufferPeerAccessAfterImport::test_main[False] - RuntimeError: IPC is not available on Windows

The three tests in `cuda_core/tests/memory_ipc/test_peer_access.py` depend on the `mempool_device_x2` fixture, which only checks for two GPUs with mutual peer access — it does not check for IPC support.

The fix is to add an `ipc_mempool_device_x2` fixture in `cuda_core/tests/conftest.py` that composes `mempool_device_x2` with the same IPC-capability checks already used by `ipc_device` . 

After the fix, the tests could be successfully skipped.
tests/memory_ipc/test_peer_access.py::TestPeerAccessNotPreservedOnImport::test_main SKIPPED (Device does not...)
tests/memory_ipc/test_peer_access.py::TestBufferPeerAccessAfterImport::test_main[True] SKIPPED (Device does ...)
tests/memory_ipc/test_peer_access.py::TestBufferPeerAccessAfterImport::test_main[False] SKIPPED (Device does...)

### DIFF
--- a/cuda_core/tests/conftest.py
+++ b/cuda_core/tests/conftest.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
 import multiprocessing
@@ -284,6 +284,20 @@ def mempool_device_x2():
 def mempool_device_x3():
     """Fixture that provides three devices if available, otherwise skips test."""
     return _mempool_device_impl(3)
+
+
+@pytest.fixture
+def ipc_mempool_device_x2(mempool_device_x2):
+    """Fixture that provides two IPC-capable mempool devices, or skips."""
+    from helpers import IS_WSL, supports_ipc_mempool
+
+    if not all(device.properties.handle_type_posix_file_descriptor_supported for device in mempool_device_x2):
+        pytest.skip("Device does not support IPC")
+
+    if IS_WSL or not all(supports_ipc_mempool(device) for device in mempool_device_x2):
+        pytest.skip("Driver rejects IPC-enabled mempool creation on this platform")
+
+    return mempool_device_x2
 
 
 @pytest.fixture(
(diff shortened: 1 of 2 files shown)


rluo8:
Hi @mdboom @rwgk ,
Could you please help review it?
Thanks.

leofang (changes requested):
I left a comment in the bug, I don't think this is correct.

rwgk:
@leofang , thanks for calling out the distinction between peer access and IPC. I agree that they are separate requirements, and that was the gap in the original fixture setup.

The key point is that `cuda_core/tests/memory_ipc/test_peer_access.py` is still an IPC test even though it also exercises peer-access behavior. It creates `DeviceMemoryResource(..., ipc_enabled=True)` and transfers the resource across processes, so it needs both:

- peer-access-capable devices, and
- IPC-capable mempools on those devices.

Before the Cursor-generated follow-up commit 73d6797e718f26fe34808081d54ea41ea7d129e7 that I just pushed, `ipc_mempool_device_x2` and `ipc_device` each carried similar IPC skip logic separately. The new commit factors those checks into one shared helper in `cuda_core/tests/conftest.py` so that we consistently gate IPC tests on the same preconditions.

The intent is:

- `mempool_device_x2`: peer access + mempool support
- `ipc_device`: mempool support + IPC-capable mempool support
- `ipc_mempool_device_x2`: peer access + mempool support + IPC-capable mempool support

That keeps the nvbug 6176793 fix targeted at the real issue: the peer-access IPC tests were only gated on peer-access readiness, not on IPC-capable mempool readiness.

@rluo8 , your earlier fixture split was the right direction; this commit just removes the duplicated IPC checks and makes the relationship between the fixtures clearer.

I think this leaves the product behavior unchanged and confines the fix to test preconditions, which seems like the safest resolution for this bug.

___

Cursor model used: GPT-5.4 272K Medium

rwgk (approved):
Looks good to me.

@leofang, could you please re-review (you're currently holding a lock on this PR).

Andy-Jost (approved):

leofang:
> The key point is that `cuda_core/tests/memory_ipc/test_peer_access.py` is still an IPC test even though it also exercises peer-access behavior.

Thanks, this is what I missed.

leofang (approved):

Issue #467, opened 2025-02-24
Add `TextureObject` and `SurfaceObject`?


leofang:
Prior art: https://docs.cupy.dev/en/stable/reference/cuda.html#texture-and-surface-memory

PR #2095, opened 2026-05-15
cuda.core: add CUDAArray, TextureObject, SurfaceObject, MipmappedArray (texture/surface API, #467)

### LINKED ISSUE #2188: Design: cuda.core Texture/Surface API surface
## Purpose

Design discussion for the **texture / surface API** in `cuda.core` — to settle the API shape and
naming *before* code review of the implementation. Reviewers asked for design sign-off in an issue
before we commit to a ~9k-line feature.

- Implementation PR: #2095
- Feature request: #467

cc @leofang @mdboom @Andy-Jost @kkraus14 — you asked for a design pass; this is the home for it.

## Proposed public surface (from #2095)

- `Array` + `ArrayFormat` — opaque, hardware-laid-out GPU allocations backing textures/surfaces.
- `MipmappedArray` — wraps `CUmipmappedArray`; `get_level` returns a non-owning `Array` kept alive
  by a strong ref to the parent.
- `TextureObject` + `TextureDescriptor` — bindless texture handle + sampling state. (read-only)
- `SurfaceObject` — bindless surface handle; requires `Array(surface_load_store=True)`. (read/write)
- `ResourceDescriptor` — factories `from_array`, `from_mipmapped_array`, `from_linear`, `from_pitch2d`.

## Decisions to make

1. **Name of `Array`. ✅ Decided — rename `Array` → `CUDAArray`.**

   This type is an opaque `cudaArray_t` — the GPU stores it in a scrambled, hardware-defined layout
   with no linear pointer, so it **cannot** expose `__cuda_array_interface__` / DLPack and cannot
   share memory zero-copy with cupy / numba-cuda / torch. The name `Array` implies an n-dimensional
   array that participates in that ecosystem — it can't. CuPy names the identical type `CUDAarray`,
   and its whole `cupy.cuda.texture` module already matches this PR's surface 1:1.

   **Resolution: use `CUDAArray`** — the PEP 8 CapWords form (deliberately differing from CuPy's exact
   `CUDAarray` casing to follow Python's class-naming standard). The name signals "CUDA texture/surface
   backing store," not "n-dimensional array."

   **Open detail resolved: keep `ArrayFormat` (do *not* rename to `CUDAArrayFormat`).** The sibling
   enums in these modules — `AddressMode`, `FilterMode`, `ReadMode` — are all unprefixed, so
   `ArrayFormat` matches the established enum-naming pattern; and the "`Array` implies an
   ndarray/DLPack participant" concern that motivated `CUDAArray` does not apply to a format enum
   (nobody mistakes `ArrayFormat` for an n-dimensional array).

2. **Interop path. ✅ Decided — ship only `copy_from` / `copy_to`; no allocation helper.**

   Zero-copy is impossible (opaque layout, no linear pointer), so copying is the only option —
   this was purely about how polished the path is. The copy path to/from linear `cuda.core`
   `Buffer`s already exists: `copy_from` / `copy_to` accept a device `Buffer` or a host
   buffer-protocol object, in both directions. The only thing an extra helper would add is
   allocating the linear `Buffer` for the caller — folding `mr.allocate(arr.size_bytes, stream=s)`
   + `arr.copy_to(buf, stream=s)` into a one-liner, i.e. ~2 lines of convenience.

    The "extra helper" would collapse this:
    ```
    mr = DeviceMemoryResource(device)
    buf = mr.allocate(arr.size_bytes, stream=s)
    arr.copy_to(buf, stream=s)

    into something like 
   buf = arr.to_buffer(mr, stream=s).
    ```

   **Resolution: ship `copy_from` / `copy_to` only, and document the copy-only contract.** We will
   not add an allocating convenience helper now. It is purely additive and non-breaking, so we can
   add one later if users request it.

3. **Factory set. ✅ Not a real decision — driver-mandated, all four required.**

   A texture can be backed by four kinds of memory — the PR exposes one factory per kind:
   - `from_array` — texture over a `CUDAArray` *(the headline feature)*
   - `from_mipmapped_array` — texture over a `MipmappedArray` *(the headline feature)*
   - `from_linear` — texture over a plain 1D device buffer *(ordinary linear memory, no `CUDAArray`)*
   - `from_pitch2d` — texture over a plain 2D pitched buffer *(ordinary linear memory, no `CUDAArray`)*

   `ResourceDescriptor` binds `CUDA_RESOURCE_DESC`, a driver union whose `resType` is exactly one of
   `ARRAY` / `MIPMAPPED_ARRAY` / `LINEAR` / `PITCH2D` — one factory per union arm. A faithful binding
   of that type *must* cover all four; shipping only two would be an incomplete binding of a mandatory
   driver struct, not a smaller-but-valid surface. So there was no real optionality here — the CTK
   driver API dictates the set. (Listed only because it sat next to the genuine decisions.)

   **Resolution: ship all four factories — required by the driver API, not a tradeoff.**

4. **Channel format. ✅ Decided — keep the folded `format` + `num_channels` parameters.**

   Each array element has a component type (e.g. 8-bit uint, 32-bit float) and a channel count
   (1 = grayscale … 4 = RGBA). Two ways to surface that:
   - **Folded (this PR):** `CUDAArray.from_descriptor(shape=..., format=ArrayFormat.FLOAT32, num_channels=4)`
   - **Separate (CuPy):** one `ChannelFormatDescriptor(...)` object passed as a unit

   The driver descriptor `cuda.core` actually fills in (`CUDA_ARRAY3D_DESCRIPTOR`) already stores
   these as two separate fields — `Format` (a `CUarray_format`, mirrored 1:1 by `ArrayFormat`) and
   `NumChannels`. So the folded form maps straight onto the driver struct with no translation, and
   read-back is already exposed as two properties (`.format`, `.num_channels`). The bundled
   `ChannelFormatDescriptor` is the *runtime* API's (`cudaChannelFormatDesc`) modeling — the form
   CuPy wraps because its texture module sits on the runtime API. Adopting it in a driver-based
   library would mean a translation wrapper the underlying API doesn't use (and the shapes don't even
   map cleanly: the driver uses one uniform component format × channel count, while
   `cudaChannelFormatDesc` allows per-channel bit widths).

   **Resolution: keep folded `format` + `num_channels`.** It's the driver-faithful surface
   (consistent with #1 favoring correctness over CuPy parity and #3 following the driver API); the
   bundled form's only wins are CuPy look-alike and a single read-back object, neither worth a
   runtime-style wrapper here.

5. **Descriptor type consistency. ✅ Not a real decision — divergence is intentional and harmless.**

   *Note: the original framing here was factually wrong.* `ResourceDescriptor` is **not** a `cdef
   class` and holds **no** native C struct — it is a plain Python class with `__slots__`, storing a
   reference to the backing resource plus a few Python fields. The `CUDA_RESOURCE_DESC` struct is
   assembled later, in `TextureObject.from_descriptor`. So this was never a `@dataclass`-vs-`cdef
   class` / performance question. Both descriptors are pure Python.

   The genuine difference is only *how you construct each*, and it reflects what each type is:
   - `TextureDescriptor` — a flat bag of independent sampling settings, built directly with keyword
     args (`@dataclass` fits perfectly).
   - `ResourceDescriptor` — a "pick exactly one of four backings" union (array / mipmap / linear /
     pitch2d), built via `from_*` factories because each kind carries different fields. A single
     `__init__` would be a pile of mutually-exclusive optional args plus a kind tag.

   Consistency is not a goal in itself — it only matters when inconsistency makes the API harder to
   learn or use, and here it doesn't: a user learns each type once and never has to reconcile them.
   The only behavioral gap is equality (`TextureDescriptor` compares by value; `ResourceDescriptor`
   by identity), which is essentially never exercised on these objects and is arguably correct since
   `ResourceDescriptor` wraps a live device resource. Forcing both to the same *kind* of type would
   be uniformity for its own sake and would make `ResourceDescriptor`'s constructor worse.

   **Resolution: keep the split — the divergence is intentional and does not hurt usability.** Like
   #3, this resolves to a non-issue once examined (and on a mistaken premise to begin with).

6. **Bool naming. ✅ Decided — adopt the `is_<something>` convention.**

   `surface_load_store` is a boolean on `Array`: it records whether the array was created with the
   surface load/store capability (CUDA's `CUDA_ARRAY3D_SURFACE_LDST`), which a `SurfaceObject`
   requires. Exposed both as a constructor keyword (`surface_load_store=True`) and a read-only
   property (`arr.surface_load_store`).

   The repo convention for boolean properties is `is_<something>`, so a property named
   `surface_load_store` doesn't read as a boolean the way `arr.is_managed` does. **Resolution: rename
   the property to follow the `is_<x>` convention (e.g. `is_surface_load_store`) for consistency with
   the cuda-python codebase.**

   **Open detail resolved:** the property name is `is_surface_load_store` (already implemented), and
   the constructor keyword is renamed to match — `from_descriptor(..., is_surface_load_store=False)` —
   so one symmetric name serves both set and read-back. This follows the existing
   `StridedMemoryView(is_readonly=...)` precedent in cuda.core, where an `is_<x>` boolean is used as
   both the constructor argument and the attribute. (The keyword rename is a small implementation
   follow-up in the PR; the property is already done.)

7. **Scope. ✅ Decided — split the examples into a follow-up PR.**

   The nine `gl_interop_*.py` examples (~5k lines, not CI-wired, need a GL context CI lacks) are
   orthogonal to the core API. **Resolution: drop them from this PR and land them in a separate
   follow-up PR once this core texture/surface PR merges**, since the examples depend on the new API
   it introduces.

rparolin:
## Summary

Implements [#467](https://github.com/NVIDIA/cuda-python/issues/467) (`TextureObject` / `SurfaceObject`) plus the supporting texture/surface stack on `cuda.core`. The API shape and naming were settled in the design issue [#2188](https://github.com/NVIDIA/cuda-python/issues/2188).

- **`OpaqueArray`** + **`ArrayFormat`** — opaque, hardware-laid-out GPU allocations backing textures/surfaces (`Array` → `CUDAArray` → `OpaqueArray`; see Naming updates below).
- **`MipmappedArray`** — wraps `CUmipmappedArray`; `get_level` returns a non-owning `OpaqueArray` level view whose parent is kept alive structurally.
- **`TextureObject`** + **`TextureDescriptor`** — bindless texture handle with the full sampling-state surface (filter/read/address modes, border color, mipmap clamps, anisotropy, sRGB, seamless cubemap).
- **`SurfaceObject`** — bindless surface handle for kernel-side typed load/store; requires `OpaqueArray(is_surface_load_store=True)` (renamed per #2188 decision 6).
- **`ResourceDescriptor`** — four factories (`from_array`, `from_mipmapped_array`, `from_linear`, `from_pitch2d`) covering all texture-eligible arms of `CUDA_RESOURCE_DESC`.

All of these are exposed under the **`cuda.core.texture`** namespace (not the flat `cuda.core` namespace) — see Review feedback below.

## Naming updates (post-review)

- **`CUDAArray` → `OpaqueArray`** — the backing-storage class is now `cuda.core.texture.OpaqueArray` (the internal handle type was renamed `CUDAArrayHandle` → `OpaqueArrayHandle` to match). This supersedes #2188 decision 1, which had settled on `CUDAArray`.
- **`cuda.core.textures` → `cuda.core.texture`** (singular) — the grouping namespace was renamed for parity with CuPy's `cupy.cuda.texture`, and the impl modules now live in a `cuda/core/texture/` subpackage mirroring `cuda/core/graph/`.

## Design decisions reflected (#2188)

| # | Decision | Status |
|---|----------|--------|
| 1 | `Array` → `OpaqueArray` (supersedes the earlier `CUDAArray`); keep `ArrayFormat` | ✅ |
| 2 | Copy-only interop (`copy_from`/`copy_to`), documented; no allocation helper | ✅ |
| 3 | All four `ResourceDescriptor` factories | ✅ |
| 4 | Folded `format` + `num_channels` | ✅ |
| 5 | Keep the descriptor type split (`@dataclass` vs `from_*` factories) | ✅ |
| 6 | `is_surface_load_store` property **and** constructor keyword | ✅ |
| 7 | Split the `gl_interop_*` examples into a follow-up PR | ⏳ see Remaining |

## Review feedback addressed (#2188)

- **Copy API parity with `Buffer`** (@Andy-Jost): `OpaqueArray.copy_from` / `copy_to` now accept `Stream | GraphBuilder` (via `Stream_accept`), so array copies can be captured into a CUDA graph; and `copy_to` returns `dst`.
- **Module grouping** (@Andy-Jost): the texture/surface types are grouped under `cuda.core.texture` (grouped-only — no flat `cuda.core` exports; pre-release, so no deprecation needed).

## Resource lifetime

All four types route CUDA resource lifetime through `cuda.core`'s C++ `std::shared_ptr` resource-handle layer (`_cpp/resource_handles.{hpp,cpp}`): each owning handle's deleter calls the matching `cu*Destroy` with the GIL released, and inter-resource dependencies (mipmap level → parent, texture/surface → backing) are embedded structurally in the box. No raw handles, `_owning` flags, Python-reference keepalives, or `__dealloc__` — consistent with the rest of `cuda.core` (`_cpp/DESIGN.md`).

## What's included

- Public API under `cuda.core.texture`; docs in `docs/source/api.rst` (Textures and surfaces), including the copy-only interop contract.
- Examples kept here as representative end-to-end coverage:
  - `examples/texture_sample.py` — allocate a 2D `OpaqueArray`, bind a `LINEAR/CLAMP` `TextureObject`, verify POINT-exact + bilinear sampling from a kernel.
  - `examples/gl_interop_mipmap_lod.py` — `MipmappedArray` pyramid built per-level via `SurfaceObject`, sampled through a `TextureObject` with trilinear (LINEAR mipmap) filtering and LOD bias (covers the `MipmappedArray` / `get_level` headline path).
  - `examples/gl_interop_fluid.py` (`cuda.core` `OpaqueArray`/`TextureObject`/`SurfaceObject`) and `examples/gl_interop_fluid_numba_cuda_mlir.py` (numba-cuda-mlir port, in its own pixi env) — Stable-Fluids.

  The remaining `gl_interop_*` showcase is split out (see Remaining).
- 65 texture/surface unit tests: happy paths, negative-path validation for every `raise` site, boundary cases, structural-lifetime (level outlives dropped parent), idempotent `close()`, `copy_to`-returns-`dst`, and graph-capture of array copies.

## Test plan

- [x] texture/surface suite — 65 passed.
- [x] `pixi run` full `cuda_core` suite — 3451 passed, 214 skipped, 3 xfailed.
- [x] Cython ABI test (`tests/cython/`) — passed.
- [x] All `examples/` byte-compile; the `numba-cuda-mlir` port runs end-to-end.
- [x] CI green on NVIDIA runners.

## Remaining (follow-up PRs)

- The other 14 standalone `gl_interop_*.py` demos have been moved to the `feature/cuda-core-texture-examples-467` branch for continued iteration and will land in a separate examples PR (#2188 decision 7).
- Layered / cubemap / sparse `OpaqueArray` variants (currently `NotImplementedError`-deferred).
- Descriptor round-trip introspection (`cuTexObjectGetResourceDesc`, etc.).


leofang (commented):
We should have a design review as both @mdboom and @Andy-Jost suggested in the past meetings. At least, the API surface should be sketched in the issue before a PR is fired up. Having major feature PRs vibed without design discussion makes it very easy to merge regrettable changes. 

`Array` is in particular a terrible name.

rparolin:
Opened #2188 to host the design discussion (API surface + `Array` naming + interop contract) ahead of merging this, per review feedback. Let's settle the design there; I'll split the `gl_interop_*` examples into a separate PR. xref #467.

rparolin:
### `cuda.core` texture / surface examples

**Stable Fluids** (`gl_interop_fluid.py`) — a Jos Stam fluid solver whose velocity and dye fields are `CUDAArray`s bound both as a `TextureObject` (hardware-filtered reads) and a `SurfaceObject` (writes); semi-Lagrangian advection is a single `tex2D` linear fetch.

cuda-fluid-sim

**Mipmap / LOD** (`gl_interop_mipmap_lod.py`) — a `MipmappedArray` pyramid built per level via `SurfaceObject` writes, then sampled through a `TextureObject` with trilinear (LINEAR-mipmap) filtering and an adjustable LOD bias.

cuda-mipmap-lod

---

The two shots below were part of a larger set of texture/surface showcase examples that may be integrated in a later PR:

**Caustics** (`gl_interop_caustics.py`) — animated water caustics; `TextureObject` sampling with `MIRROR` addressing, `NORMALIZED_FLOAT` reads, sRGB, and anisotropic filtering.

cuda-caustics

**Clouds** (`gl_interop_clouds.py`) — volumetric clouds raymarched against a 3D `CUDAArray` sampled as a volume `TextureObject` with `WRAP` addressing.

cuda-clouds

rparolin:
@leofang @mdboom @Andy-Jost @kkraus14 — this is ready for review. 🎉

- **Design signed off in #2188**, and all the review feedback there is addressed: `Array` → `CUDAArray` and `is_surface_load_store` (decisions 1 & 6), copy-API parity with `Buffer` (`copy_from`/`copy_to` accept `Stream | GraphBuilder`; `copy_to` returns `dst`), and the texture/surface types are grouped under the new **`cuda.core.textures`** namespace (per @Andy-Jost's suggestion).
- **Resource lifetime** now routes through cuda.core's C++ `std::shared_ptr` handle layer (structural dependencies in the box, GIL-guarded deleters, no `__dealloc__`/`_parent_ref`) — consistent with the rest of cuda.core.
- **Scope** is the core API plus representative examples (`texture_sample`, `gl_interop_mipmap_lod`, and both Stable-Fluids ports); the remaining `gl_interop_*` showcase is split into a follow-up PR (decision 7).
- **CI is green** — full GPU test matrix (linux-64 / aarch64 / win, py3.10–3.14t, CUDA 12.9 / 13.0 / 13.3), docs build, CodeQL/analyze, and pre-commit.ci all passing.

Example screenshots are in the comment above. Thanks!

cuda_core/cuda/core/_cpp/resource_handles.hpp
```diff
@@ -171,6 +190,10 @@ using NvvmProgramHandle = std::shared_ptr<const NvvmProgramValue>;
 using NvJitLinkHandle = std::shared_ptr<const NvJitLinkValue>;
 using CuLinkHandle = std::shared_ptr<const CUlinkState>;
 using FileDescriptorHandle = std::shared_ptr<const int>;
+using ArrayHandle = std::shared_ptr<const CUarray>;
```
Andy-Jost: nit: `CUDAArrayHandle`?

cuda_core/cuda/core/_surface.pyx
```diff
@@ -0,0 +1,132 @@
...
+        )
+
+    @classmethod
+    def from_array(cls, array):
+        """Create a surface object directly from an :class:`CUDAArray`.
+
+        The array must have been created with ``is_surface_load_store=True``.
+        """
+        if not isinstance(array, CUDAArray):
+            raise TypeError(f"array must be an CUDAArray, got {type(array).__name__}")
```
Andy-Jost: Cosmetic: several error messages read "must be an CUDAArray" (should be "a CUDAArray")

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -1337,6 +1347,168 @@ FileDescriptorHandle create_fd_handle_ref(int fd) {
...
+
+ArrayHandle create_array_handle_ref(CUarray arr) {
+    if (!arr) {
+        return {};
+    }
+    auto box = std::make_shared<const ArrayBox>(ArrayBox{arr, {}});
+    return ArrayHandle(box, &box->resource);
+}
+
+ArrayHandle create_array_handle_owning(CUarray arr) {
```
Andy-Jost: `create_array_handle` and `create_array_handle_owning` only differ in who calls `cuArray3DCreate`. Could we pick one or the other?

cuda_core/cuda/core/texture/__init__.py
Andy-Jost, line 1: The `cuda.core.graph` submodule lives in a `graph/` subdirectory containing `.pyx`/`.pxd` files. This PR builds a virtual module from files at the `cuda/core` level. Is the divergence intentional?

Andy-Jost (approved):
No blockers from me

cuda_core/cuda/core/_cpp/resource_handles.hpp
rparolin: Done in 40037e21ca — renamed the alias to `CUDAArrayHandle` across the .hpp/.cpp and Cython .pyx/.pxd/.pyi (lowercase `create_array_handle*` functions and `MipmappedArrayHandle` unchanged).

cuda_core/cuda/core/_surface.pyx
rparolin: Done in 40037e21ca — fixed to "must be a CUDAArray" in both `_surface.pyx` and `_texture.pyx`.

cuda_core/cuda/core/_cpp/resource_handles.cpp
rparolin: Done in ed97f7c5fa — `create_array_handle(desc)` now delegates to `create_array_handle_owning(arr)` after `cuArray3DCreate`, so the owning box + destroy deleter live in exactly one place. Kept both entry points since they do different jobs (allocate-from-descriptor vs adopt-an-existing `CUarray` for graphics interop), and `cuArray3DCreate` stays in the C++ lifetime layer per `_cpp/DESIGN.md`.

cuda_core/cuda/core/texture/__init__.py
rparolin: Done in 6df91a9987 — moved the four impl modules (`_array`/`_texture`/`_surface`/`_mipmapped_array`) into a `cuda/core/texture/` subpackage mirroring `graph/`, with `textures.py` becoming `texture/__init__.py`. Note the public namespace is now `cuda.core.texture` (singular), matching `cupy.cuda.texture`; I'll update the PR description and #2188 which still say `textures`.

rparolin:
Heads-up on a naming change in the latest push (1b76277655): the backing-storage class `CUDAArray` has been renamed to **`OpaqueArray`** (and the internal handle type `CUDAArrayHandle` → `OpaqueArrayHandle` to match). The public type is now `cuda.core.texture.OpaqueArray`.

PR #2096, opened 2026-05-16
Harden pinned NUMA mempool tests against constructor OOM flakes

rwgk:
This is a small follow-up to PR #2084.

The `cuda_core_5.15.log` attached to `nvbug 5815123` shows two remaining intermittent `cuda_core` failures after the PR #2084 patch:

- `tests/test_memory.py::test_pinned_mr_numa_id_default_no_ipc`
- `tests/test_memory.py::test_pinned_mr_numa_id_explicit`

Both tests are meant to validate `PinnedMemoryResource.numa_id` behavior, but they instantiate `PinnedMemoryResource(PinnedMemoryResourceOptions(...))` directly. Those constructor paths create a real pinned memory pool immediately, so they can still hit the same constructor-time `CUDA_ERROR_OUT_OF_MEMORY` failure mode that is already handled elsewhere in the Windows mempool workaround coverage.

This change routes those constructor sites through the existing `create_pinned_memory_resource_or_xfail(...)` helper. It also updates the nearby `test_pinned_mr_numa_id_default_with_ipc` case for consistency, since it exercises the same pool-creation path.

The goal is to keep these tests focused on NUMA-ID semantics instead of failing on the known Windows MCDM mempool-constructor flake.

___

xref: `nvbug 5815123`

### DIFF
--- a/cuda_core/tests/test_memory.py
+++ b/cuda_core/tests/test_memory.py
@@ -1474,11 +1474,11 @@ def test_pinned_mr_numa_id_default_no_ipc(init_cuda):
     device = Device()
     skip_if_pinned_memory_unsupported(device)
 
-    mr = PinnedMemoryResource(PinnedMemoryResourceOptions())
+    mr = create_pinned_memory_resource_or_xfail(PinnedMemoryResourceOptions(), xfail_device=device)
     assert mr.numa_id == -1
     mr.close()
 
-    mr = PinnedMemoryResource(PinnedMemoryResourceOptions(ipc_enabled=False))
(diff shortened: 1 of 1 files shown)


leofang:
btw nightly also failed for the OOM issue (which I don't quite understand why would happen): https://github.com/NVIDIA/cuda-python/actions/runs/25947528269/job/76279804206#step:26:17561

Merging not because I agree we need to address cuda-core test issues during CTK bring-up, but because this actually blocks our CI and thus development, as seen in #2087.

leofang (approved):

PR #2098, opened 2026-05-16
Adapt green-context SM split tests to topology

rwgk:
## Summary
- Stop assuming Hopper+ devices always expose `min_partition_size` and `coscheduled_alignment` values of `8`.
- Probe for supported explicit SM split sizes in the tests instead of assuming `min_partition_size` is always a valid request for multi-group splits.
- Keep backfill coverage, but skip the backfill-specific case when the device does not expose a backfill-only two-group split.
- Interactively confirmed to resolve nvbug 6097301 on a Tegra Thor system.
- Agent used: Cursor GPT-5.4 Extra High Fast

### DIFF
--- a/cuda_core/tests/test_green_context.py
+++ b/cuda_core/tests/test_green_context.py
@@ -82,16 +82,58 @@ def fill_kernel(init_cuda):
     return mod.get_kernel("fill")
 
 
-def _safe_two_group_count(sm):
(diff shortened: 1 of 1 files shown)


leofang:
Let's re-run the CI with the latest driver!

leofang (approved):

rwgk:
Thanks @leofang!

PR #2099, opened 2026-05-16
cuda.core: bump tensor bridge PyTorch upper-bound to 2.12

### LINKED ISSUE #2089: Bump tensor bridge version cap for PyTorch 2.12
## Summary

The tensor bridge's `_torch_version_check()` in `_memoryview.pyx` caps the supported PyTorch version range at `(2, 3) <= (major, minor) <= (2, 11)`. PyTorch 2.12 was released on ~May 14, 2026, causing the version check to fail. When this happens, `_is_torch_tensor()` returns `False` and torch tensors fall through to the generic CAI/DLPack paths instead of the fast AOTI tensor bridge.

For `from_cuda_array_interface`, this causes a hard failure because PyTorch reports `__cuda_array_interface__` version 2, while cuda-python requires version 3+:

```
BufferError: only CUDA Array Interface v3 or above is supported
```

## Root cause

```python
# _memoryview.pyx
cdef inline bint _torch_version_check():
    ...
    _torch_version_ok = (2, 3) <= (major, minor) <= (2, 11)  # ← needs bumping
```

The upper bound exists because the tensor bridge relies on undocumented PyTorch internals (`THPVariable` struct layout, `AtenTensorHandle == at::Tensor*`). The docstring says: "Bump the upper bound after verifying a new PyTorch release."

## Action needed

1. Verify PyTorch 2.12's `THPVariable` struct layout is unchanged (check `torch/csrc/autograd/python_variable.h`)
2. Bump the upper bound to `(2, 12)`
3. Re-add `latest` to the nightly CI pytorch test matrix (currently pinned to `2.11.0` as a workaround in #1987)

## Discovered by

Nightly optional-dependency CI (#1987) — the `latest` torch entry started failing when PyTorch 2.12 was released.

-- Leo's bot

aryanputta:
Closes #2089.

## Problem

PyTorch 2.12 was released on ~May 14 2026. The version guard in `_torch_version_check()` (`cuda_core/cuda/core/_memoryview.pyx`) capped the AOTI tensor bridge at `(2, 11)`, so `_is_torch_tensor()` returned `False` for any torch tensor under 2.12.

As a result, torch tensors fell through to the generic CAI/DLPack paths and raised:

```
BufferError: only CUDA Array Interface v3 or above is supported
```

## Fix

Extend the upper bound from `(2, 11)` to `(2, 12)`. The `THPVariable` struct layout (`PyObject_HEAD` followed by `at::Tensor cdata`) and the `AtenTensorHandle == at::Tensor*` identity are stable across PyTorch 2.x minor releases, as they have been across 2.3–2.11.

```diff
-    _torch_version_ok = (2, 3) <= (major, minor) <= (2, 11)
+    _torch_version_ok = (2, 3) <= (major, minor) <= (2, 12)
```

## Testing

This change cannot be exercised without a PyTorch 2.12 build in CI. The existing tensor bridge tests (`cuda_core/tests/`) cover the code path; they will validate correctness once a 2.12 runner is available. The nightly CI matrix already references a `latest` PyTorch entry (see `nightly.yml`) which should pick up 2.12.

### DIFF
--- a/cuda_core/cuda/core/_memoryview.pyx
+++ b/cuda_core/cuda/core/_memoryview.pyx
@@ -43,7 +43,7 @@ cdef dict _torch_type_cache = {}
 cdef object _torch_version_ok = None
 
 cdef inline bint _torch_version_check():
-    """Return True if 2.3 <= torch <= 2.11 (known AOTI ABI range). Memoized.
+    """Return True if 2.3 <= torch <= 2.12 (known AOTI ABI range). Memoized.
 
     Lower bound: AOTI functions we use were introduced in PyTorch 2.3.
     Upper bound: the ``pyobj_to_aten_handle`` trick relies on the
@@ -64,7 +64,7 @@ cdef inline bint _torch_version_check():
     try:
         major, minor = int(torch.__version__.split(".")[0]), \
                        int(torch.__version__.split(".")[1])
-        _torch_version_ok = (2, 3) <= (major, minor) <= (2, 11)
+        _torch_version_ok = (2, 3) <= (major, minor) <= (2, 12)
     except (ValueError, IndexError):
         _torch_version_ok = False
     return <bint>_torch_version_ok


rwgk (approved):
I think this immediate bump makes sense as a short-term unblock for PyTorch 2.12, but the hard cap may be causing more user pain than the underlying ABI risk warrants.

Right now, every new PyTorch minor release effectively requires a new `cuda-core` release, even if the AOTI ABI and `THPVariable` layout have not changed. When the cap trips, users also do not get a very direct explanation of what happened; they fall back to the generic path and can end up seeing a confusing downstream error instead.

Would it make sense to keep the conservative default behavior, but make it explicit and overridable? For example:

```bash
export CUDA_PYTHON_TORCH_TENSOR_BRIDGE_COMPATIBILITY_CHECK=ERROR  # or WARN / OFF
```

With `ERROR` as the default, we could raise a clear message explaining that the detected PyTorch version is outside the validated range and point users to the `WARN` and `OFF` options as escape hatches when they are willing to assume ABI compatibility.

That would preserve the safety-first default while avoiding the need to cut a new `cuda-core` release for every PyTorch minor version unless there is evidence that compatibility actually broke.

rwgk:
I assume the right validation for this PR is a quick source-level check against PyTorch 2.12. Has that already been done?

In particular, I think the key files to inspect would be:

- `torch/csrc/autograd/python_variable.h`, to confirm that `THPVariable` still has the expected layout (`PyObject_HEAD` followed by `at::Tensor cdata`)
- `torch/csrc/inductor/aoti_torch/c/macros.h`, to confirm the `AtenTensorHandle` contract is unchanged
- `torch/csrc/inductor/aoti_torch/c/shim.h`, to confirm the AOTI entry points we use are still present and compatible

If someone already compared the relevant 2.11 vs 2.12 sources, it would be great to call that out in the PR description or comments.

rwgk:
While I was at it, I asked Cursor (GPT-5.4 272K Medium) to go ahead and do the source-based validation based on a local git clone of pytorch. Below is what it found. I'll go ahead and merge this PR.

___

I did a source-level comparison locally against the PyTorch `v2.11.0` and `v2.12.0` tags to validate the assumptions behind this PR.

What I checked:

- `torch/csrc/autograd/python_variable.h` between `v2.11.0` and `v2.12.0`: no diff. `THPVariable` still has `PyObject_HEAD` followed by `at::Tensor cdata`, which is the core assumption behind the `pyobj_to_aten_handle` trick.
- `torch/csrc/inductor/aoti_torch/c/macros.h`: no diff. PyTorch still documents `AtenTensorHandle` as being represented under the hood as `at::Tensor*`.
- `torch/csrc/inductor/aoti_torch/utils.h`: no diff. The helper still does `reinterpret_cast<at::Tensor*>(handle)`.
- `torch/csrc/inductor/aoti_torch/c/shim.h`: there is a diff, but it appears additive and unrelated to the APIs used here. The specific metadata getters used by `cuda-core` are still present with the same signatures.
- `torch/csrc/inductor/aoti_torch/shim_common.cpp`: the diff also appears additive; I only saw new dtype helper implementations. The getter implementations used by this bridge are unchanged and still go through the same `tensor_handle_to_tensor_pointer(...)` path before reading tensor metadata.

My conclusion is that, for the specific `2.11` -> `2.12` bump in this PR, the source-based validation looks good. I do not see evidence of a break in either of the two assumptions the PR description relies on:

- the `THPVariable` layout stayed the same
- `AtenTensorHandle` still maps to `at::Tensor*`, and the exact AOTI getter path used by `cuda-core` is unchanged

The caveat is that this is still not a formal compatibility guarantee from PyTorch for the `THPVariable` trick itself, so the cap remains a conservative policy choice rather than pure paranoia. But for this specific release bump, the source comparison does support the PR.

PR #2034, opened 2026-05-06
Add cuDLA bindings

nikshayshrivastava:
## Description

Generated from cudla.h using cybind.

Files added:
- cycudla.pxd/pyx: Cython layer exposing C header types and functions
- cudla.pxd/pyx: lowpp Python layer with POD classes, enums, and wrappers
- _internal/cudla.pxd, cudla_linux.pyx, cudla_windows.pyx: dynamic library loading
- docs/source/module/cudla.rst: API documentation
- tests/cudla/: pytest unit tests for enums, POD types, error handling, API surface, and hardware-gated function tests (verified on L4T/Orin)

Build/CI updates:
- pyproject.toml: added cudla to cuda-toolkit optional dependencies
- .github/actions/fetch_ctk/action.yml: added libcudla to CTK components
- docs/source/api.rst: added cudla to toctree

## Checklist

- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #2101, opened 2026-05-18
Add numba-cuda-mlir docs, cuda-core dependency, and toolkit extras

leofang:
## Summary

- Mention **numba-cuda-mlir** in the CUDA Python README and metapackage docs.
- Add `cuda-core~=1.0.0` as a required dependency of the `cuda-python` metapackage.
- Add CUDA Toolkit metapackage dependencies to `cuda-bindings[all]` and `cuda-core`'s `cu12` / `cu13` extras for improved dependency resolution.
- Keep cuDLA available through `cuda-bindings[all]` on Linux aarch64 via a guarded `nvidia-cudla==13.*` dependency.
- Update installation docs to include `nvFatBin` and `cuDLA` in the optional dependency list.

## Notes

- `cuda-cccl` is intentionally left out of the `cuda-python` metapackage for now because it does not yet have Python 3.14 free-threaded builds. It can be added in a follow-up once that support is available.


cuda_bindings/pyproject.toml
```diff
@@ -36,8 +36,9 @@ dependencies = ["cuda-pathfinder >=1.4.2"]
 
 [project.optional-dependencies]
 all = [
-    "cuda-toolkit[nvrtc,nvjitlink,nvvm,nvfatbin,cudla]==13.*",
+    "cuda-toolkit[nvrtc,nvjitlink,nvvm,nvfatbin]==13.*",
```
leofang, line 39: I noticed this change sneaked in from PR #2034. This is a bit problematic because cuDLA is not a universally available component in the CTK, and the majority of users cannot use it but they'd see this at install time:
```
WARNING: cuda-toolkit 13.1.2.0 does not provide the extra 'cudla'
```
We should just document that DLA users need to install cuDLA if they haven't already (which I suspect is not the case, because there should be a system CTK already installed on embedded devices).

rwgk: I approved this knowingly, after discovering that the meta package has the filters we need.

I missed that warning message though. If we want to get rid of it, we have do add the conditions similar to what we're doing under `cuda_pathfinder/pyproject.toml`:

```
    "nvidia-cudla; platform_system == 'Linux' and platform_machine == 'aarch64'",
```

cuda_python/setup.py
```diff
@@ -32,6 +32,8 @@
     version=version,
     install_requires=[
         f"cuda-bindings{matcher}{version}",
+        "cuda-core~=1.0.0",
+        "cuda-cccl~=1.0.0",
```
leofang: To discuss: If we need numba-cuda-mlir here

rparolin: What's the argument for adding it?

leofang: It is a CUDA Python compiler and part of the CUDA Python program.

cuda_bindings/pyproject.toml
kkraus14: That would still fire on aarch64 server machines where I don't think we have a way to differentiate Jetson / Tegra with current pip capabilities today

cuda_python/setup.py
leofang: Based on the internal discussion (in the RFE gdoc), we would not add numba-cuda-mlir for now.

gmarkall (approved):

leofang:
/ok to test [0587448](https://github.com/NVIDIA/cuda-python/pull/2101/commits/0587448b646ba19a41a0e63096d8e624164ac814)

rparolin (approved):

cuda_python/setup.py
leofang: Life has been not very kind to us. Because `cuda-cccl` does not yet have a 3.14t build, and because there is no environment marker that would allow us to append a condition here (PEP 780 is still in draft status and the DPO discussion is brutal), all Python 3.14t pipelines failed. We have no other way but drop `cuda-cccl` here:
```suggestion
```

cuda_bindings/docs/source/install.rst
```diff
@@ -12,7 +12,7 @@ Runtime Requirements
 * Linux (x86-64, arm64) and Windows (x86-64)
 * Python 3.10 - 3.14
 * Driver: Linux (580.65.06 or later) Windows (580.88 or later)
-* Optionally, NVRTC, nvJitLink, NVVM, and cuFile from CUDA Toolkit 13.x
+* Optionally, NVRTC, nvJitLink, nvFatBin, NVVM, and cuFile from CUDA Toolkit 13.x
```
kkraus14: Should we include cuDLA here now as well?

cuda_bindings/docs/source/install.rst
```diff
@@ -52,11 +52,16 @@ Where the optional dependencies include:
 
 * ``nvidia-cuda-nvrtc`` (NVRTC runtime compilation library)
 * ``nvidia-nvjitlink`` (nvJitLink library)
+* ``nvidia-nvfatbin`` (nvFatBin library)
```
kkraus14, line 55: +1 cuDLA here?

cuda_bindings/pyproject.toml
kkraus14: My 2c: I think we should include cuDLA in `all` still as it would only error at runtime if someone tried to use a cuDLA binding from a non Tegra system. The cuDLA binary is tiny (< 1MB), so binary size impact isn't a big concern.

cuda_python/setup.py
kkraus14: Let's add cuda-cccl in a follow up because of this.

cuda_bindings/pyproject.toml
leofang: @kkraus14 My concerns:
1. the warning (see above)
2. would `cuda-toolkit==13.0` still work for constraining the installed packages? cuDLA wheel was not shipped until 13.2

kkraus14: If we guard it based on @rwgk's comment in using `platform_system` and `platform_machine` does it still warn?

I assume if we added it to `all` and someone did `pip install cuda-bindings[all] cuda-toolkit==13.0` that it would error because there isn't a `cuda-toolkit[cudla]` package for 13.0?

cuda_python/setup.py
rwgk: Done: commit f4c693247d9acfe9af1a51b33c0977e867c5faf0

@shwina for visibility

cuda_bindings/docs/source/install.rst
rwgk: Done: commit f4c693247d9acfe9af1a51b33c0977e867c5faf0

@nikshayshrivastava for visibility

cuda_bindings/docs/source/install.rst
rwgk: Done: commit f4c693247d9acfe9af1a51b33c0977e867c5faf0

@nikshayshrivastava for visibility

cuda_bindings/pyproject.toml
rwgk: Done: commit f4c693247d9acfe9af1a51b33c0977e867c5faf0

rwgk (approved):
Looks good to me.

I looked through manually, and I used GPT-5.5 for a full agent review.

PR #2102, opened 2026-05-18
[doc-only] Add a short toolshed/README.md

rwgk:
Clarify expectations for the `toolshed/` directory as a shared home for useful, best-effort scripts that do not require production-level maintenance.

### DIFF
--- a/toolshed/README.md
+++ b/toolshed/README.md
@@ -1,0 +1,10 @@
+# Shared Toolshed
+
+This directory is a shared scratchpad for scripts that are useful
+intermittently, such as environment bootstrap helpers, ad-hoc reproducers, and
+glue for manual workflows. These tools do not warrant CI coverage, unit tests,
+or the rest of the production-code apparatus.
+
+Entries may bitrot between uses. Keeping them in a shared, versioned location
+gives the next person who needs the same workflow a better starting point than
+starting from scratch.


kiwigitops (commented):
Nice — having an in-tree note about `toolshed/`'s scope is a real improvement over the implicit convention.

One question on the "do not warrant CI coverage, unit tests, or the rest of the production-code apparatus" line: is that expectation enforced anywhere today, or only intended? Looking at the existing contents (`check_cython_abi.py`, `check_spdx.py`, the `build_pathfinder_*` family, etc.), some of these look CI-shaped enough that a future contributor might wire one into a workflow by reflex.

Concretely:
- Is `toolshed/` excluded from the test discovery patterns (e.g. `pyproject.toml` `pytest` config, ruff/mypy include rules) so a stray `test_`-prefixed scratch file doesn't accidentally become CI surface?
- Would it be worth a short bullet in this README spelling out the boundary — e.g. "Files here are excluded from `pytest` collection and `ruff`/`mypy` checks; if a tool graduates past scratchpad status, move it to `<canonical-location>` first"?

Happy either way — just thinking about how a future reader who only sees this README knows where the guardrails actually live. Not blocking for a doc-only PR.

kkraus14 (approved):

Issue #2104, opened 2026-05-18
Review of Restricted Paths Trust Signal Explorations


rwgk:
## Scope

I fetched the full `gh pr view` history for `.github/workflows/restricted-paths-guard.yml` and the relevant linked discussions I could access:

- #1878
- #1930
- #2010
- #1871
- #1896
- #1901
- #1821
- #2011
- #1139
- #1886
- private `#294`

I could not access the internal `github-apps` PRs `#78` or `#79` with the available token.

## What I Learned

The path was twisty because each candidate signal solved one failure mode and exposed another.

#1139 started with the practical policy: external contributions are generally OK, but changes under `cuda_bindings/` need early/manual review.

#1871 first tried folding this into `pr-metadata-check`, then #1878 split it into a dedicated `restricted-paths-guard.yml` workflow.

`author_association` looked attractive because it is directly available in the pull request payload. But testing showed webhook payloads can say `CONTRIBUTOR` for fork PRs even when the author is an NVIDIA org member or explicit repo collaborator. In later live API reads, those same users could appear as `MEMBER`. That made the event payload too lossy for enforcement.

Direct organization membership APIs would be conclusive in principle. In practice, the normal workflow `GITHUB_TOKEN` could not reliably query private NVIDIA org membership. That meant the workflow could not distinguish "not a member" from "not allowed to know" well enough.

Public org membership/profile checks were also explored. They behave cleanly as true-positive signals for public org members, but many legitimate internal users do not expose public membership. So that would still leave too many internal contributors in the manual-review bucket.

A checked-in allowlist was considered as a rollout fallback, then removed. It was simple and explicit, but too manual and not really a scalable trust source.

#1930 switched to `GET /repos/{owner}/{repo}/collaborators/{username}/permission`, trusting `admin|maintain|write`. This sidestepped the stale/inconsistent webhook `author_association` value and worked for known collaborators with elevated repo access.

#2010 broadened the trusted permission set to `admin|maintain|write|triage|read` because #1821 showed an internal/collaborator author returning `read`, which caused a false positive review label.

The new discovery from PR #2103 is that on a public repo, `read` is not a collaborator signal. It is just effective public repository access. For example, both `LeSingh1` and `octocat` return `permission: read` from:

```bash
gh api repos/NVIDIA/cuda-python/collaborators/<user>/permission
```

[...]

PR #2105, opened 2026-05-18
[no-ci] CI: stop trusting public read permission in restricted paths guard

rwgk:
Reference: #2104

## Summary

This PR fixes the restricted-paths guard trust check so public repository `read` permission is no longer treated as a collaborator/trust signal.

The guard now:

- queries the live pull request API for `author_association` on restricted-path PRs
- trusts live `MEMBER` and `OWNER` author associations
- falls back to the collaborator permission API only when the live author association is not trusted
- trusts collaborator permissions `triage`, `write`, `maintain`, and `admin`
- treats collaborator permission `read` as untrusted, because public repositories can report `read` for any GitHub user

## Background

Earlier versions of `restricted-paths-guard.yml` used `github.event.pull_request.author_association` from the workflow event payload. That value turned out to be stale or unreliable for some fork PRs, so PR #1930 switched the workflow to the collaborator permission API.

PR #2010 later expanded the trusted collaborator permissions to include `read`, after seeing a false positive for an internal/collaborator author whose permission was reported as `read`.

Issue #2104 captures the follow-up discovery: on a public repository, the collaborator permission API can return `read` for users who are not repository collaborators. For example, a first-time external contributor can get effective `read` access simply because the repository is public.

This PR keeps the useful part of the previous workaround, but tightens the trust model:

- live `author_association` is queried from `GET /repos/{owner}/{repo}/pulls/{pull_number}` instead of using the event payload
- `read` from `GET /repos/{owner}/{repo}/collaborators/{username}/permission` is no longer trusted

## Testing via reverted temporary commits

- 5ede77269f — TEMPORARY: Switch to pull_request trigger for testing
- 5226846b47 — TEMPORARY: Exclude write permission from trusted collaborators

### DIFF
--- a/.github/workflows/restricted-paths-guard.yml
+++ b/.github/workflows/restricted-paths-guard.yml
@@ -6,8 +6,7 @@ name: "CI: Restricted Paths Guard"
 on:
   # Run on drafts too so maintainers get early awareness on WIP PRs.
   # Label updates on fork PRs require pull_request_target permissions.
-  # TEMPORARY: Using pull_request for testing; revert to pull_request_target before merge.
-  pull_request:
+  pull_request_target:
     types:
       - opened
       - synchronize
--- a/.github/workflows/restricted-paths-guard.yml
+++ b/.github/workflows/restricted-paths-guard.yml
@@ -6,7 +6,8 @@ name: "CI: Restricted Paths Guard"
 on:
   # Run on drafts too so maintainers get early awareness on WIP PRs.
   # Label updates on fork PRs require pull_request_target permissions.
-  pull_request_target:
+  # TEMPORARY: Using pull_request for testing; revert to pull_request_target before merge.
+  pull_request:
     types:
       - opened
       - synchronize
--- a/.github/workflows/restricted-paths-guard.yml
+++ b/.github/workflows/restricted-paths-guard.yml
@@ -217,7 +217,7 @@ jobs:
               fi
 
               case "$COLLABORATOR_PERMISSION" in
-                admin|maintain|write|triage)
+                admin|maintain|triage)
                   HAS_TRUSTED_SIGNAL=true
                   LABEL_ACTION="not needed (collaborator permission is a trusted signal)"
                   TRUSTED_SIGNALS="collaborator_permission:$COLLABORATOR_PERMISSION"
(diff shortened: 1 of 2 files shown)


.github/workflows/restricted-paths-guard.yml
```diff
@@ -135,46 +146,87 @@ jobs:
...
           if [ "$TOUCHES_RESTRICTED_PATHS" = "true" ]; then
-            # Distinguish a legitimate 404 "not a collaborator" response from
-            # actual API failures. The former is an expected untrusted case;
-            # the latter fails the workflow so it can be rerun later.
-            if COLLABORATOR_PERMISSION_RESPONSE=$(
-              gh api "repos/$REPO/collaborators/$PR_AUTHOR/permission" \
-                --jq '.permission' 2>&1
+            if AUTHOR_ASSOCIATION_RESPONSE=$(
+              gh api "repos/$REPO/pulls/$PR_NUMBER" \
+                --jq '.author_association // "NONE"' 2>&1
```
leofang, line 151: Nit: this fetches the full PR JSON just to extract `.author_association`. Not a real concern for a CI guard (one extra REST call, and `--jq` filters the output), just noting it in case a future reader wonders why the full PR endpoint is used rather than a lighter query.

leofang (approved):
Trust model tightening looks correct and well-reasoned. Walked through the key edge cases (org MEMBER with read-only perm, outside COLLABORATOR with write, FIRST_TIME_CONTRIBUTOR) — all handled properly by the two-tier check. One minor inline note, otherwise LGTM.

.github/workflows/restricted-paths-guard.yml
leofang: Please ignore my bot. I don't think this makes sense. `gh api ... --jq ...` is perfectly fine.

rwgk:
Thanks @leofang!

rwgk:
Tested via this dummy PR:

* https://github.com/NVIDIA/cuda-python/pull/1895#issuecomment-4492448614

The workflow worked as intended:

* https://github.com/NVIDIA/cuda-python/actions/runs/26128203440

Copy-pasting the Summary view:

___

Restricted Paths Guard Completed

Author: rwgkunplugged

Author association: NONE

Collaborator permission: read

Touches restricted paths: true

Restricted paths: cuda_bindings/, cuda_python/

Trusted signals: (none)

Label action: added

Comment action: posted

Matched restricted paths:

cuda_bindings/pyproject.toml

Manual follow-up: No trusted signal was found, so Needs-Restricted-Paths-Review is required.

PR #2107, opened 2026-05-19
Fix #1960: Codify our cuda-core API-auditing process in AGENTS.md

### LINKED ISSUE #1960: Add API auditing to AGENTS.md
Once #1919 is completed and we've learned what was useful or not useful, we should convert the queries that audited the API into AGENTS.md instructions so that it can help review any new APIs going forward.

(This would be useful for cuda.core.  Not really useful for cuda.bindings which follows upstream CUDA APIs more closely and we don't really have the ability to set our own guidelines).

mdboom:
This codifies the work from #1919.  This tries to encode what we /arrived/ at after the whole audit process, not just what we started with in #1919.

### DIFF
--- a/cuda_core/AGENTS.md
+++ b/cuda_core/AGENTS.md
@@ -63,3 +63,82 @@ This file describes `cuda_core`, the high-level Pythonic CUDA subpackage in the
   call-site consistency.
 - Prefer explicit error propagation over silent fallback paths.
 - If you change public behavior, update tests and docs under `docs/source/`.
+
+## API design guidelines
+
+These are some API design guidelines we try to follow when adding new APIs to
+`cuda.core`.  These rules only apply to public APIs.  Private implementation
+details can violate these rules at any time.
+
+Public APIs are defined as symbols defined in `__all__` within modules or
+subpackages that are not prefixed with `_`.
+
+In code reviews, any violations of this section should be considered
+suggestions, not hard rules.  Consistency with existing API design in this code
+base is also important.
+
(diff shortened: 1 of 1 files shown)


kkraus14 (approved):

cuda_core/AGENTS.md
```diff
@@ -63,3 +63,82 @@ This file describes `cuda_core`, the high-level Pythonic CUDA subpackage in the
...
+Python or Cython type annotations should be included for all public APIs.  Avoid
+the use of `Any` unless absolutely necessary.  The argument and return types as
+defined in the docstrings should match the type annotations.
+
+### Semantics
+
+Designs involving manual resource management should be avoided.  Where
+appropriate, provide context managers (implemented with `__enter__` and
+`__exit__`, not `contextlib.contextmanager`) or RAII using a `__del__` or
+`__dealloc__` method.
```
leofang, line 133: this does not match our current design guidelines

mdboom: The context manager part or the destructor part, or both?  We have a ton of the latter in the codebase already.

leofang: Right, I meant the former. But also the whole bullet point in general. For example, all of our public objects have a `.close()` method, and Phillip ensured that all code samples use them, so this does not hold:
> Designs involving manual resource management should be avoided.

mdboom: Ok.  I guess it's more both things, then: an API for manual resource management and helper context managers where appropriate.

PR #2112, opened 2026-05-19
Fix a double-free bug

mdboom:
This is part of my experimentation with using an LLM agent for static analysis.

Fixes a double-free in the generated setters for list-valued struct members
(e.g. `CUlaunchConfig.attrs`, `CUDA_MEM_ALLOC_NODE_PARAMS.accessDescs`,
external-semaphore and batch-mem-op node parameter arrays, and their runtime
counterparts). Assigning an empty list freed the internal buffer but left
the cached pointer non-NULL, so a subsequent assignment or `__dealloc__`
would call `free()` again on the dangling pointer.

### DIFF
--- a/cuda_bindings/tests/test_cuda.py
+++ b/cuda_bindings/tests/test_cuda.py
@@ -3,6 +3,8 @@
 
 import ctypes
 import shutil
+import subprocess
+import sys
 import textwrap
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


cuda_bindings/tests/test_cuda.py
leofang, line 1: Q: So the new tests would break without this fix?

mdboom: Yes.  The first test is really easy to do from Python.  The second test is a bit harder, but still possible.  The whole weirdness in the middle is just because it's hard to write a test for it.  Reproducing the bug just takes one object referencing another, setting an array on it and then using the referencing object after the first is freed.

leofang (approved):

PR #2113, opened 2026-05-19
[no-ci] Follow-on to cuda.core AGENTS.md

mdboom:
This is a follow up to #2107 based on [this comment](https://github.com/NVIDIA/cuda-python/pull/2107#pullrequestreview-4323433672) and an offline discussion about `typing.TYPE_CHECK`.

### DIFF
--- a/cuda_core/AGENTS.md
+++ b/cuda_core/AGENTS.md
@@ -125,12 +125,19 @@ Python or Cython type annotations should be included for all public APIs.  Avoid
 the use of `Any` unless absolutely necessary.  The argument and return types as
 defined in the docstrings should match the type annotations.
 
+The use of `if typing.TYPE_CHECK:` blocks is often necessary to avoid import
+cycles.  However, this can occasionally mask symbols used by Sphinx to make
+cross-references in the docs.  If importing an object doesn't create a cycle, it
+should be imported outside of an `if typing.TYPE_CHECK:` block, even if it is
+only used in type annotations.
+
 ### Semantics
 
-Designs involving manual resource management should be avoided.  Where
-appropriate, provide context managers (implemented with `__enter__` and
-`__exit__`, not `contextlib.contextmanager`) or RAII using a `__del__` or
-`__dealloc__` method.
+APIs should exist for both manual resource management (such as `close()`) and
+automatic resource management, using context managers or destructors where
+appropriate.  Context managers should be implemented with `__enter__` and
+`__exit__`, not `contextlib.contextmanager`.  For destructors use `__dealloc__`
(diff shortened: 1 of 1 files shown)


cuda_core/AGENTS.md
```diff
@@ -125,12 +125,19 @@ Python or Cython type annotations should be included for all public APIs.  Avoid
 the use of `Any` unless absolutely necessary.  The argument and return types as
 defined in the docstrings should match the type annotations.
 
+The use of `if typing.TYPE_CHECK:` blocks is often necessary to avoid import
+cycles.  However, this can occasionally mask symbols used by Sphinx to make
+cross-references in the docs.  If importing an object doesn't create a cycle, it
+should be imported outside of an `if typing.TYPE_CHECK:` block, even if it is
+only used in type annotations.
```
rwgk: I'd turn this around: common case first, exception second, the "why" last (because it doesn't really matter for the desired agent behavior). Roughly:

Python imports should generally be outside of an `if typing.TYPE_CHECK:` block, even if the imported object is only used in type annotations. Use `if typing.TYPE_CHECK:` only to avoid creating import cycles. (This guidance maximizes compatibility with the cross-reference mechanisms in Sphinx.)

rwgk (approved):
I'd reword, in style of an actual guidance, but LGTM otherwise.

cuda_core/AGENTS.md
mdboom:
```suggestion
Python imports should generally be outside of an if typing.TYPE_CHECK: block, even if the imported object is only used in type annotations. Use if typing.TYPE_CHECK: only to avoid creating import cycles. (This guidance maximizes compatibility with the cross-reference mechanisms in Sphinx.)
```

leofang:
LGTM but I'd let @Andy-Jost take a look on `TYPE_CHECKING` as he seems to have reservation based on the offline chat.

Andy-Jost:
> LGTM but I'd let @Andy-Jost take a look on `TYPE_CHECKING` as he seems to have reservation based on the offline chat.

The wording here looks fine to me. Regarding `TYPE_CHECKING`, I'm mostly confused. I'm looking for rules (a), (b), (c), ... such that following them avoids certain identifiable problems, e.g., with circular imports, tools relying on type annotations, and so on. The rules would probably need to cover package structure, `__init__.py` file contents, `TYPE_CHECKING` blocks, and other `[c]import` guidelines.

AFAICT, we have not identified either the rules or even the complete list of problems we need to avoid. None of that should gate this PR, though.

Andy-Jost (approved):

Issue #2004, opened 2026-05-01
memory_ipc/test_send_buffers.py::TestIpcReexport::test_main[DeviceMR] hanging in CI


rwgk:
I manually canceled this job when it reached 58m 42s:

[Test linux-64 / py3.12, 12.9.1, local, l4](https://github.com/NVIDIA/cuda-python/actions/runs/25202833731/job/73897828000?pr=2003#logs)

It was hanging here, for almost 54 minutes:

```
2026-05-01T05:08:24.6867513Z tests/memory_ipc/test_send_buffers.py::TestIpcSendBuffers::test_main[3] PASSED [ 13%]
2026-05-01T06:02:21.2264769Z tests/memory_ipc/test_send_buffers.py::TestIpcReexport::test_main[DeviceMR]
2026-05-01T06:02:21.2758059Z ##[error]The operation was canceled.
```

A rerun worked, i.e. the job appears to be flaky.

I'm attaching the log for easy future reference:

[job-logs.txt](https://github.com/user-attachments/files/27271452/job-logs.txt)

rwgk:
Here is another case (this one hung for almost 4 hours):

[Test linux-64 / py3.12, 12.9.1, local, l4](https://github.com/NVIDIA/cuda-python/actions/runs/25409424774/job/74528666791#logs)

```
2026-05-06T00:45:35.6404725Z tests/memory_ipc/test_send_buffers.py::TestIpcReexport::test_main[DeviceMR-3-5] PASSED [ 12%]
2026-05-06T03:39:05.1997826Z tests/memory_ipc/test_send_buffers.py::TestIpcReexport::test_main[DeviceMR-4-5] 
2026-05-06T03:39:05.2534642Z ##[error]The operation was canceled.
```

[job-logs.txt](https://github.com/user-attachments/files/27423695/job-logs.txt)

rwgk:
Here is another case (this one timed out after blocking the runner for 6 hours):

[Test linux-aarch64 / Python 3.12, CUDA 12.9.1 (local), GPU a100](https://github.com/NVIDIA/cuda-python/actions/runs/25955161493/job/76301019309?pr=2097#logs)

```
2026-05-16T06:54:28.5249320Z tests/memory_ipc/test_send_buffers.py::TestIpcSendBuffers::test_main[3] PASSED [ 11%]
[...]

PR #2118, opened 2026-05-20
nvbug-6193808: Work around mojibake in nvml.system_get_process_name on WSL

mdboom:
## Summary

`cuda.core.system.get_process_name(pid)` raises `UnicodeDecodeError` under
WSL whenever the calling process has a non-`C` locale (which is the default
state for any CPython process, since the interpreter calls
`setlocale(LC_ALL, "")` at startup). This is reproducible by running the
`cuda_core` test suite with any seed that schedules
`tests/system/test_system_device.py::test_compute_running_processes` before
`tests/system/test_system_system.py::test_get_process_name`.

The underlying defect is in NVML's WSL implementation (see *Root cause*
below). This PR adds a scoped, defensive workaround in `cuda_core` so the
public API returns a correct value on WSL. It also fixes a latent issue
where `get_process_name` was effectively unusable from a fresh process
because it never primed NVML's per-PID name cache.

## Root cause: the WSL mojibake

NVML's `nvmlSystemGetProcessName` on WSL takes a different code path
depending on the process's current locale. With the default `"C"` locale,
the function returns the basename portion of `/proc/<pid>/exe` correctly.
With any other locale (including the typical `en_US.UTF-8`), it instead
walks an internal UTF-16LE buffer holding the executable path but uses a
**4-byte stride** (as if the buffer were UTF-32LE). Each "code point" it
pulls is therefore two adjacent ASCII bytes packed into the low and
next-higher bytes of a single 24-bit value. That value is then emitted as
an extended 5-byte UTF-8 sequence (the `0xF8`-prefixed encoding used to
represent code points beyond U+10FFFF).

The net result for, say, a Python process whose `/proc/<pid>/exe` resolves
to:

```
/home/mdboom/.local/share/uv/python/cpython-3.14.0-linux-x86_64-gnu/bin/python3.14
```

is that the returned buffer looks like ~180 bytes of `f8 …` chunks
followed by the correctly-encoded trailing basename, e.g.:

```
f8 9a 80 80 af  f8 9b 90 81 af  f8 8b b0 81 a5  f8 99 80 81 ad  …  /python3.14\0
```

Decoding the first chunk illustrates the pattern:

- bytes `f8 9a 80 80 af` decode as the extended-UTF-8 code point `0x68002F`
- that 24-bit value packs `'h'` (`0x68`) in the high byte and `'/'` (`0x2F`)
  in the low byte — i.e. the source ASCII bytes `/h` read as a
  little-endian 32-bit value padded with zeros

Every chunk has this structure; together they spell out the prefix
`/home/mdboom/.local/share/uv/python/cpython-3.14.0-linux-x86_64-gnu/bin/`
two characters at a time. The trailing `/python3.14` is unaffected because
of where the buggy stride leaves the cursor.

### Why the workaround needs to "re-prime"

`nvmlSystemGetProcessName` is cache-driven: the per-PID name is populated
the first time NVML enumerates compute processes that include the PID
(typically via `nvmlDeviceGetComputeRunningProcesses_v3`). Critically:

- The mojibake is produced *during* the prime call, not during the read.
- Once a buggy entry is in NVML's cache, switching to the `"C"` locale and
  re-reading does **not** unscramble it — the cache survives the locale
  flip.
- Re-running the prime call under the `"C"` locale overwrites the cached
  entry with the correct UTF-8 string. Subsequent reads (in any locale)
  then return correctly.

So the workaround must do prime + read together under `"C"`.

## Behaviour after this PR

- **Native Linux / Windows**: behaviour is identical to before, except that
  `get_process_name` now primes the NVML cache automatically. This makes
  it usable from a fresh process (previously a caller had to have
  manually queried `device.compute_running_processes` first or accept
  `NotFoundError`).
- **WSL**: the locale flip is applied around the prime + read sequence,
  so the returned name is the correct UTF-8 string regardless of the
  caller's locale.

## Discussion

- Should we instead try to fix this at the ``cuda_bindings`` layer and fix it
  for all ``cuda_bindings`` users? In that case I guess ``cuda_core`` should raise
  an exception from ``get_process_name`` if on WSL and the ``cuda_bindings``
  installed is too old?
- I do plan to file the underlying bug with NVML. Locale-sensitive APIs should
  generally be avoided.

### DIFF
--- a/cuda_bindings/docs/source/release/13.2.0-notes.rst
+++ b/cuda_bindings/docs/source/release/13.2.0-notes.rst
@@ -80,3 +80,7 @@ Known issues
 ------------
 
 * Updating from older versions (v12.6.2.post1 and below) via ``pip install -U cuda-python`` might not work. Please do a clean re-installation by uninstalling ``pip uninstall -y cuda-python`` followed by installing ``pip install cuda-python``.
+* ``nvml.system_get_process_name`` on WSL can return incorrect values.  To work
+  around this, set the locale to "C" before calling
+  ``nvml.system_get_process_name``. ``cuda_core`` does this automatically, but
+  users of the raw NVML API will need to do this manually.
--- a/cuda_bindings/docs/source/release/13.2.0-notes.rst
+++ b/cuda_bindings/docs/source/release/13.2.0-notes.rst
@@ -80,7 +80,4 @@ Known issues
 ------------
 
 * Updating from older versions (v12.6.2.post1 and below) via ``pip install -U cuda-python`` might not work. Please do a clean re-installation by uninstalling ``pip uninstall -y cuda-python`` followed by installing ``pip install cuda-python``.
-* ``nvml.system_get_process_name`` on WSL can return incorrect values.  To work
-  around this, set the locale to "C" before calling
-  ``nvml.system_get_process_name``. ``cuda_core`` does this automatically, but
-  users of the raw NVML API will need to do this manually.
+* ``nvml.system_get_process_name`` on WSL can return incorrect values.  To work around this, set the locale to "C" before calling ``nvml.device_get_compute_running_processes_v3`` (which sets the process names) and before calling ``nvml.system_get_process_name``. ``cuda_core`` does this automatically, but users of the raw NVML API will need to do this manually.
--- a/cuda_core/cuda/core/system/_system.pyx
+++ b/cuda_core/cuda/core/system/_system.pyx
@@ -10,6 +10,30 @@
 
 CUDA_BINDINGS_NVML_IS_COMPATIBLE: bool
 
+
+# POSIX per-thread locale APIs. We use these (rather than setlocale(3))
+# so the WSL workaround in get_process_name() doesn't perturb the locale
+# observed by other threads. locale_t is an opaque pointer in glibc.
+cdef extern from "locale.h" nogil:
+    ctypedef void *locale_t
+    int LC_ALL_MASK
+    locale_t LC_GLOBAL_LOCALE
+    locale_t newlocale(int category_mask, const char *locale, locale_t base)
+    locale_t uselocale(locale_t newloc)
+    void freelocale(locale_t locobj)
+
+
+cdef bint _detect_wsl():
+    try:
+        with open("/proc/sys/kernel/osrelease") as f:
+            data = f.read().lower()
+    except OSError:
+        return False
+    return "microsoft" in data or "wsl" in data
+
+
+cdef bint _IS_WSL = _detect_wsl()
+
 try:
     from cuda.bindings._version import __version_tuple__ as _BINDINGS_VERSION
 except ImportError:
@@ -127,8 +151,43 @@ def get_process_name(pid: int) -> str:
     name: str
         The process name.
     """
+    def _get_process_name(pid) -> str:
+        # NVML caches process names on a per-PID basis when queried via
+        # nvmlSystemGetProcessName, and the cache is populated when enumerating
+        # running processes on devices. To ensure the name is cached for the
+        # requested PID, we walk all devices and query their running processes.
+        for i in range(nvml.device_get_count_v2()):
+            dev_h = nvml.device_get_handle_by_index_v2(i)
+            nvml.device_get_compute_running_processes_v3(dev_h)
+        return nvml.system_get_process_name(pid)
+
+    cdef locale_t c_locale
+    cdef locale_t prev_locale
+
     initialize()
-    return nvml.system_get_process_name(pid)
+    if not _IS_WSL:
+        return _get_process_name(pid)
+
+    # WSL workaround: nvmlSystemGetProcessName on WSL takes a wide-char
+    # conversion path when the process locale is non-"C". That path walks
+    # a UTF-16LE source buffer with a 4-byte stride (as if it were UTF-32LE)
+    # and emits 5-byte UTF-8 sequences that look like garbage preceding the
+    # trailing basename of /proc/<pid>/exe. CPython's startup unconditionally
+    # calls setlocale(LC_ALL, ""), so essentially every cuda.core caller hits
+    # this. The cached entry for the PID is set the first time NVML resolves
+    # it (typically inside nvmlDeviceGetComputeRunningProcesses_v3), so to
+    # recover a correct value we re-prime the cache under the "C" locale
+    # before reading the name. We use the POSIX per-thread locale APIs so
+    # other threads' view of the locale is unaffected.
+    c_locale = newlocale(LC_ALL_MASK, b"C", <locale_t>0)
+    if c_locale == <locale_t>0:
+        raise RuntimeError("Failed to create C locale")
+    prev_locale = uselocale(c_locale)
+    try:
+        return _get_process_name(pid)
+    finally:
+        uselocale(prev_locale)
+        freelocale(c_locale)
 
 
 __all__ = [
--- a/cuda_core/cuda/core/system/_system.pyx
+++ b/cuda_core/cuda/core/system/_system.pyx
@@ -157,8 +157,11 @@ def get_process_name(pid: int) -> str:
         # running processes on devices. To ensure the name is cached for the
         # requested PID, we walk all devices and query their running processes.
         for i in range(nvml.device_get_count_v2()):
-            dev_h = nvml.device_get_handle_by_index_v2(i)
-            nvml.device_get_compute_running_processes_v3(dev_h)
+            try:
+                dev_h = nvml.device_get_handle_by_index_v2(i)
+                nvml.device_get_compute_running_processes_v3(dev_h)
+            except nvml.NvmlError:
+                continue
         return nvml.system_get_process_name(pid)
 
     initialize()
--- a/cuda_core/cuda/core/system/_system.pyx
+++ b/cuda_core/cuda/core/system/_system.pyx
@@ -11,18 +11,6 @@
 CUDA_BINDINGS_NVML_IS_COMPATIBLE: bool
 
 
-# POSIX per-thread locale APIs. We use these (rather than setlocale(3))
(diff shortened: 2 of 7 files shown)


rwgk (commented):
I used Cursor GPT-5.5 1M Extra High

___

## GPT Findings

1. Medium: `cuda_bindings/docs/source/release/13.2.0-notes.rst` and
   `cuda_bindings/docs/source/release/13.3.0-notes.rst` give an incomplete
   raw-NVML workaround. The PR and nvbug context indicate the corruption happens
   when `nvmlDeviceGetComputeRunningProcesses_v3` first primes the PID cache, so
   setting the locale to `"C"` only before `nvml.system_get_process_name` can be
   too late if the cache was already populated. The notes should say to prime and
   read under `"C"`, matching the `cuda.core` workaround.

2. Medium: `cuda_core/cuda/core/system/_system.pyx` now makes
   `get_process_name()` depend on successfully enumerating compute processes on
   every device. Any unrelated device-level NVML failure now breaks a per-PID
   lookup, including on non-WSL where this is a behavior change. Consider trying
   the direct read first on non-WSL, or making priming failures narrower.

3. Low: `cuda_core/docs/source/release/1.1.0-notes.rst` says the WSL workaround
   may hold a global lock, but the implementation uses POSIX per-thread locale
   APIs and no global lock is present. That note looks stale or misleading.

## Lightweight Thread-Safety Recommendation

The new locale switching implementation is reasonably thread-safe because it
uses POSIX `newlocale`/`uselocale`/`freelocale`, which scope the `"C"` locale to
the calling OS thread rather than mutating process-global locale state.

The remaining race is around NVML's process-name cache, which appears to be
process/global driver state. On WSL, another thread could call a cache-priming
path such as `nvmlDeviceGetComputeRunningProcesses_v3` under a non-`"C"` locale
between `get_process_name()`'s prime and read steps, reintroducing corrupted
cached data.

A lightweight improvement would be to add a module-private Python
`threading.RLock` shared by the `cuda.core` paths most likely to touch this
cache. Hold it around:

- `system.get_process_name()`'s WSL `c_locale_guard()` + prime + read sequence.
- `Device.compute_running_processes` on WSL, since that is the main in-package
  path that can prime the process-name cache.

This would not protect raw `cuda.bindings.nvml.*` calls or external users, but
it would cover the most likely `cuda.core` race without globally serializing all
NVML access. A Python `RLock` is sufficient here: even if Cython releases the GIL
inside NVML wrappers, the lock remains held until the surrounding Python `with`
block exits.

rwgk (approved):

cuda_bindings/docs/source/release/13.2.0-notes.rst
```diff
@@ -80,3 +80,4 @@ Known issues
 ------------
 
 * Updating from older versions (v12.6.2.post1 and below) via ``pip install -U cuda-python`` might not work. Please do a clean re-installation by uninstalling ``pip uninstall -y cuda-python`` followed by installing ``pip install cuda-python``.
+* ``nvml.system_get_process_name`` on WSL can return incorrect values.  To work around this, set the locale to "C" before calling ``nvml.device_get_compute_running_processes_v3`` (which sets the process names) and before calling ``nvml.system_get_process_name``. ``cuda_core`` does this automatically, but users of the raw NVML API will need to do this manually.
```
kkraus14, line 83: Why is this in both 13.2 and 13.3 release notes?

cuda_core/cuda/core/system/_system.pyx
```diff
@@ -10,6 +10,30 @@
 
 CUDA_BINDINGS_NVML_IS_COMPATIBLE: bool
 
+
+cdef bint _detect_wsl():
```
kkraus14, line 14: There is already a WSL test helper with broader detection logic: [`cuda_python_test_helpers._detect_wsl`](https://github.com/NVIDIA/cuda-python/blob/main/cuda_python_test_helpers/cuda_python_test_helpers/__init__.py#L18-L24) checks `/proc/sys/kernel/osrelease` and also falls back to `WSL_DISTRO_NAME` / `WSL_INTEROP`. Because the test helpers are used by `cuda.bindings` tests and should not depend on `cuda.core`, we unfortunately probably need to keep this small functionality duplicated. Could we pull the env-var fallback into this runtime helper too, so the two helpers have the same detection coverage?

cuda_bindings/docs/source/release/13.2.0-notes.rst
rwgk: It's a compatibility note that's in line with the existing `v12.6.2.post1 and below` note, which we've been carrying forward since `12.8.0`, I believe in every single release since then:

```
$ git grep 'v12.6.2.post1 and below' | cutniq | cut -d/ -f-4 | uniq -c
     16 cuda_bindings/docs/source/release
     15 cuda_python/docs/source/release
```

cuda_core/cuda/core/system/_system.pyx
rwgk: I looked at the git history, `cuda_python_test_helpers._detect_wsl` was added in #1045. That PR introduced the `WSL_DISTRO_NAME` / `WSL_INTEROP`. To me it looks like purely defensive AI-generated code that I often saw in the older agents; even in newer ones I still sometimes need to clean out defensive code. I quizzed GPT-5.5 about it; it told me:

> So yes: those env vars look mostly defensive. ... env vars can be inherited into containers/subprocesses, so the fallback is slightly more false-positive-prone than the /proc check.

I'd argue for removing the fallback, although definitely not in this PR.

kkraus14: Fair enough. We should just keep the logic consistent if possible. Ideally we'd only have one implementation, but because we want the test helpers to not depend on cuda-core and we don't want to ship this as part of cuda-bindings / cuda-pathfinder I don't think it's worth a new utilities package or anything of the like.

cuda_bindings/docs/source/release/13.2.0-notes.rst
mdboom: I was working on the assumption that we put "known issues" on previous releases.  (Since the issue applies to all of them).

mdboom:
@kkraus14, @rwgk: How do you feel about my question? 

> Should we instead try to fix this at the cuda_bindings layer and fix it for all cuda_bindings users? In that case I guess cuda_core should raise an exception from get_process_name if on WSL and the cuda_bindings installed is too old?

kkraus14:
> @kkraus14, @rwgk: How do you feel about my question?
> 
> > Should we instead try to fix this at the cuda_bindings layer and fix it for all cuda_bindings users? In that case I guess cuda_core should raise an exception from get_process_name if on WSL and the cuda_bindings installed is too old?

My 2c: We very intentionally don't do any magic in `cuda_bindings` where if the underlying C library has this behavior / issue then I think it's fine for `cuda_bindings` to have it as well. Having a workaround in `cuda_core` feels correct to me.

kkraus14 (approved):

rwgk:
> > @kkraus14, @rwgk: How do you feel about my question?
> > > Should we instead try to fix this at the cuda_bindings layer and fix it for all cuda_bindings users? In that case I guess cuda_core should raise an exception from get_process_name if on WSL and the cuda_bindings installed is too old?
> 
> My 2c: We very intentionally don't do any magic in `cuda_bindings` where if the underlying C library has this behavior / issue then I think it's fine for `cuda_bindings` to have it as well. Having a workaround in `cuda_core` feels correct to me.

That's the conclusion I'm coming to, too. I discussed this rationale with GPT-5.5:

* evidence is strong that there is a bug in NVML
* it's not the purpose of the cuda_bindings layer to add sophisticated workarounds for bugs in the wrapped libraries
* cuda_core is by design adding a higher-level layer, sophisticated workarounds are more suitable and feasible in that layer (and this PR looks great to me)
* the compatibility note in this PR is helpful, but it's not even really a compatibility note, but more a heads-up: we know there is a bug in NVML, we know how to work around it, and we added the workaround in cuda_core, but if you use cuda_bindings directly you need to be aware of it and work around it yourself
* ultimately we expect that the NVML bug will be fixed, and no workaround is needed in the future

Additional idea, to be helpful to our users:

Separately from the cuda_core workaround, I still think cuda_bindings would benefit from routing C-string decoding through a tiny helper, to make failures actionable. The current UnicodeDecodeError tells us only the first bad byte and position; in this case the useful information was the raw buffer contents and which NVML call produced them. A helper could keep the success path identical, but on decode failure raise an exception/message that includes the source API name plus a bounded repr/hex dump of the bytes. That would have made the nvbug much easier to diagnose and should be generally useful for future driver/library string issues. — I would avoid special system_get_process_name behavior; it's too niche.

rwgk:
I cancelled the CI manually after seeing that Python 3.12 with 12.9 is hanging again (two jobs). I'll report the details on the tracking bug we have already. I triggered a rerun.

rwgk:
The hanging jobs from the previous attempt are now tracked here:

https://github.com/NVIDIA/cuda-python/issues/2004#issuecomment-4509896700

rwgk:
It looks like the jobs tracked under https://github.com/NVIDIA/cuda-python/issues/2004#issuecomment-4509896700 are hanging again. Maybe the driver update made the issue a lot worse than before? Each time this happens, two runners are blocked for 4 hours, and of course merging the PR is blocked, too.

Interestingly, I just see this:

```
tests/memory_ipc/test_send_buffers.py::TestIpcReexport::test_main[DeviceMR] RERUN [ 12%]
```

But after that it's still hanging.

rwgk:
Motivated by the experience here, I created #2122 — [ENH]: Make cuda_bindings UnicodeDecodeError more actionable

rwgk:
The two jobs were hanging again in the 4th attempt. I cancelled them to unblock the runners. In the meantime @aryanputta sent PR #2121, I just triggered the CI there. I'll come back here to try again after PR #2121 is merged.

PR #2120, opened 2026-05-21
ci: trim PR matrix to PY3.12 + CUDA 12.9.1 only (debug only)

Andy-Jost:
## Summary

Temporary debugging PR. Cuts the `pull-request` test matrix in
`ci/test-matrix.yml` down to a single Python 3.12 + CUDA 12.9.1 row per
platform/arch (linux-64 L4, linux-aarch64 A100, win-64 L4) so the
`TestIpcReexport::test_main[DeviceMR]` hang seen on the r595-driver
runners (e.g. https://github.com/NVIDIA/cuda-python/actions/runs/26162055334)
reproduces fast without burning compute on configurations we already
know are green. The `nightly` matrices are intentionally untouched.

This is **not** intended to merge; it exists only to drive CI iterations
while we investigate the hang. The patch is fully self-contained and
reverts cleanly.

## Test plan

- [ ] Linux-64 L4 PY3.12 / CUDA 12.9.1 job reproduces the
      `TestIpcReexport::test_main[DeviceMR]` hang (or surfaces a real
      failure once a fix lands).
- [ ] Linux-aarch64 A100 PY3.12 / CUDA 12.9.1 job behaves the same.
- [ ] Windows L4 PY3.12 / CUDA 12.9.1 job stays green (sanity that
      trimming the matrix didn't break the workflow plumbing).
- [ ] No nightly job kicked off for this PR.

### DIFF
--- a/ci/test-matrix.yml
+++ b/ci/test-matrix.yml
@@ -17,51 +17,12 @@
 
 linux:
   pull-request:
+    # TEMPORARY: cut down to just Python 3.12 + CUDA 12.9.1 to debug the
+    # IPC re-export hang on r595 driver runners. Revert before merging.
     # linux-64
-    - { ARCH: 'amd64', PY_VER: '3.10',  CUDA_VER: '12.9.1', LOCAL_CTK: '1', GPU: 'v100',       GPU_COUNT: '1', DRIVER: 'latest' }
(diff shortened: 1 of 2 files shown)


leofang:
The fix was merged (https://github.com/NVIDIA/cuda-python/pull/2124), let's close this!

Issue #663, opened 2025-05-29
[FEA]: Tests for `ltoir`, `fatbin`, `object`, and `library` constructors


brandon-b-miller:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

cuda.bindings

### Is your feature request related to a problem? Please describe.

https://github.com/NVIDIA/cuda-python/pull/652 adds these constructors for these various code types to the `ObjectCode` class, but can't leverage the convenient pattern for testing that the existing ones do and need their own set of tests.

### Describe the solution you'd like

 Some ideas that have been brought up:

- Take a similar approach to Numba's [test binary makefile](https://github.com/NVIDIA/numba-cuda/blob/0a31c267b9dabfe912e6ab35c45f5efa73466b55/numba_cuda/numba/cuda/tests/test_binary_generation/Makefile#L4) and vendor some/all of it here with appropriate changes
- Other ideas raised in the comments https://github.com/NVIDIA/cuda-python/pull/652#issuecomment-2902069400

benhg:
Adding a note here about the `library` constructor. When I pass in a string with a path to the location of the library rather than the bytes of the library, I get TypeError: Expected type bytes, but got str. The exception at  https://github.com/NVIDIA/cuda-python/blob/8f1dd40ada363a8e4a65454818bcb2dbab1eb2f9/cuda_core/cuda/core/experimental/_linker.py#L398 likely needs to be adjusted.

leofang:
@copilot let's do the following steps to resolve this issue:
1. Fix the bug that Ben mentioned above. It should be as simple as removing the type assertion and adding another path that calls either `nvJitLinkAddFile` or `cuLinkAddFile` (plus test coverage). 
2. Add simple tests for all `ObjectCode.from_<something>()` constructors. Make sure we cover both in-memory (bytes) and on-disk (str for file path) inputs. This should be enough to ensure the fix for 1 is correct.

leofang:
`ltoir` tests are added in #890, but not `fatbin`, `object`, or `library`

leofang:
@isVoid added nvFatbin bindings in #1467. We can test the fatbin coverage now!

leofang:
> We can test the fatbin coverage now!

This is being done in #1875. Pushing the rest of this task to the next milestone.

[...]

PR #2123, opened 2026-05-21
ENH: Add support instantiating ObjectCode with path-like instances

### LINKED ISSUE #755: Support `pathlib` wherever we accept a string for the file path
Example: `ObjectCode.from_<something>()`.

acosmicflamingo:
Resolves #755

Before looking into addressing https://github.com/NVIDIA/cuda-python/issues/663, I noticed @kkraus14 had pointed out missing pathlib-like support and figured I could give it a try it a go.

One improvement I could make to `_lazy_load_module` is the ask-for-forgiveness approach to avoid checking module type twice. It would only require declaring `path_bytes` before the try block:
```diff
diff --git a/cuda_core/cuda/core/_module.pyx b/cuda_core/cuda/core/_module.pyx
index 96ac65effc..189971aed7 100644
--- a/cuda_core/cuda/core/_module.pyx
+++ b/cuda_core/cuda/core/_module.pyx
@@ -733,8 +733,16 @@ cdef class ObjectCode:
         if self._h_library:
             return 0
         module = self._module
-        assert_type_str_or_bytes_like(module)
         cdef bytes path_bytes
+        try:
+            assert_type(module, os.PathLike)
+            path_bytes = os.fsencode(module)
+            self._h_library = create_library_handle_from_file(<const char*>path_bytes)
+            if not self._h_library:
+                HANDLE_RETURN(get_last_error())
+            return 0
+        except TypeError:
+            assert_type_str_or_bytes_like(module)
         if isinstance(module, str):
             path_bytes = module.encode()
             self._h_library = create_library_handle_from_file(<const char*>path_bytes)
```

### DIFF
--- a/cuda_core/cuda/core/_module.pyx
+++ b/cuda_core/cuda/core/_module.pyx
@@ -627,9 +627,10 @@ cdef class ObjectCode:
 
         Parameters
         ----------
-        module : Union[bytes, str]
+        module : Union[bytes, str, os.PathLike]
             Either a bytes object containing the in-memory cubin to load, or
-            a file path string pointing to the on-disk cubin to load.
+            a file path object (or its string representation) pointing to the
+            on-disk cubin to load.
         name : Optional[str]
             A human-readable identifier representing this code object.
         symbol_mapping : Optional[dict]
@@ -645,9 +646,10 @@ cdef class ObjectCode:
 
         Parameters
         ----------
-        module : Union[bytes, str]
+        module : Union[bytes, str, os.PathLike]
             Either a bytes object containing the in-memory ptx code to load, or
-            a file path string pointing to the on-disk ptx file to load.
+            a file path object (or its string representation) pointing to the
+            on-disk ptx file to load.
         name : Optional[str]
             A human-readable identifier representing this code object.
         symbol_mapping : Optional[dict]
@@ -663,9 +665,10 @@ cdef class ObjectCode:
 
         Parameters
         ----------
-        module : Union[bytes, str]
-            Either a bytes object containing the in-memory ltoir code to load, or
-            a file path string pointing to the on-disk ltoir file to load.
+        module : Union[bytes, str, os.PathLike]
+            Either a bytes object containing the in-memory ltoir code to load,
+            or a file path object (or its string representation) pointing to the
+            on-disk ltoir file to load.
         name : Optional[str]
             A human-readable identifier representing this code object.
         symbol_mapping : Optional[dict]
@@ -681,9 +684,10 @@ cdef class ObjectCode:
 
         Parameters
         ----------
-        module : Union[bytes, str]
+        module : Union[bytes, str, os.PathLike]
             Either a bytes object containing the in-memory fatbin to load, or
-            a file path string pointing to the on-disk fatbin to load.
+            or a file path object (or its string representation) pointing to the
+            on-disk fatbin to load.
         name : Optional[str]
             A human-readable identifier representing this code object.
         symbol_mapping : Optional[dict]
--- a/cuda_core/cuda/core/_module.pyx
+++ b/cuda_core/cuda/core/_module.pyx
@@ -5,7 +5,7 @@
 from __future__ import annotations
 
 import os
-from cpython.object cimport PyObject_HasAttrString
+
 from libc.stddef cimport size_t
 
 from collections import namedtuple
@@ -745,10 +745,8 @@ cdef class ObjectCode:
             self._h_library = create_library_handle_from_file(<const char*>path_bytes)
         elif isinstance(module, (bytes, bytearray)):
             self._h_library = create_library_handle_from_data(<const void*><char*>module)
-        elif PyObject_HasAttrString(module, "__fspath__"):
-            path_str = module.__fspath__()
-            assert_type_str_or_bytes_like(path_str)
-            path_bytes = path_str.encode() if isinstance(path_str, str) else path_str
+        elif isinstance(module, os.PathLike):
+            path_bytes = os.fsencode(module)
             self._h_library = create_library_handle_from_file(<const char*>path_bytes)
         else:
             assert_type_str_or_bytes_like(module)
--- a/cuda_core/cuda/core/_module.pyx
+++ b/cuda_core/cuda/core/_module.pyx
@@ -5,7 +5,7 @@
 from __future__ import annotations
 
 import os
-
+from cpython.object cimport PyObject_HasAttrString
 from libc.stddef cimport size_t
 
 from collections import namedtuple
@@ -30,7 +30,6 @@ from cuda.core._resource_handles cimport (
 )
 from cuda.core._stream import Stream
 from cuda.core._utils.clear_error_support import (
-    assert_type,
     assert_type_str_or_bytes_like,
     raise_code_path_meant_to_be_unreachable,
 )
@@ -736,31 +735,23 @@ cdef class ObjectCode:
         if self._h_library:
             return 0
         module = self._module
-        try:
-            assert_type(module, os.PathLike)
-        except TypeError:
-            assert_type_str_or_bytes_like(module)
         cdef bytes path_bytes
-        # TODO: should code below move to try block?
-        if isinstance(module, os.PathLike):
-            path_bytes = os.fsencode(module)
-            self._h_library = create_library_handle_from_file(<const char*>path_bytes)
-            if not self._h_library:
-                HANDLE_RETURN(get_last_error())
-            return 0
         if isinstance(module, str):
             path_bytes = module.encode()
             self._h_library = create_library_handle_from_file(<const char*>path_bytes)
-            if not self._h_library:
-                HANDLE_RETURN(get_last_error())
-            return 0
-        if isinstance(module, (bytes, bytearray)):
+        elif isinstance(module, (bytes, bytearray)):
             self._h_library = create_library_handle_from_data(<const void*><char*>module)
-            if not self._h_library:
-                HANDLE_RETURN(get_last_error())
-            return 0
-        raise_code_path_meant_to_be_unreachable()
-        return -1
+        elif PyObject_HasAttrString(module, "__fspath__"):
+            path_str = module.__fspath__()
+            path_bytes = path_str.encode('utf-8') if isinstance(path_str, str) else path_str
+            self._h_library = create_library_handle_from_file(<const char*>path_bytes)
+        else:
+            assert_type_str_or_bytes_like(module)
+            raise_code_path_meant_to_be_unreachable()
+            return -1
+        if not self._h_library:
+            HANDLE_RETURN(get_last_error())
+        return 0
 
     def get_kernel(self, name) -> Kernel:
         """Return the :obj:`~_module.Kernel` of a specified name from this object code.
--- a/cuda_core/cuda/core/_module.pyx
+++ b/cuda_core/cuda/core/_module.pyx
@@ -7,7 +7,7 @@ from __future__ import annotations
 from libc.stddef cimport size_t
 
 from collections import namedtuple
-from os import PathLike
+from os import fsencode, fspath, PathLike
 
 from cuda.core._device import Device
 from cuda.core._launch_config cimport LaunchConfig
@@ -608,10 +608,10 @@ cdef class ObjectCode:
 
         self._code_type = str(code_type)
 
-        if isinstance(module, CodeTypeT):
+        if isinstance(module, (str, bytes, bytearray)):
             self._module = module
         elif isinstance(module, PathLike):
-            self._module = os.fspath(module)
+            self._module = fspath(module)
         else:
             self._module = module
         self._sym_map = {} if symbol_mapping is None else symbol_mapping
@@ -751,7 +751,7 @@ cdef class ObjectCode:
         elif isinstance(module, (bytes, bytearray)):
             self._h_library = create_library_handle_from_data(<const void*><char*>module)
         elif isinstance(module, PathLike):
-            path_bytes = os.fsencode(module)
+            path_bytes = fsencode(module)
             self._h_library = create_library_handle_from_file(<const char*>path_bytes)
         else:
             assert_type_str_or_bytes_like(module)
--- a/cuda_core/cuda/core/_module.pyx
+++ b/cuda_core/cuda/core/_module.pyx
@@ -4,6 +4,8 @@
 
 from __future__ import annotations
 
+import os
+
 from libc.stddef cimport size_t
 
 from collections import namedtuple
@@ -28,6 +30,7 @@ from cuda.core._resource_handles cimport (
 )
 from cuda.core._stream import Stream
 from cuda.core._utils.clear_error_support import (
+    assert_type,
     assert_type_str_or_bytes_like,
     raise_code_path_meant_to_be_unreachable,
 )
@@ -620,7 +623,7 @@ cdef class ObjectCode:
         return ObjectCode._reduce_helper, (self._module, self._code_type, self._name, self._sym_map)
 
     @staticmethod
-    def from_cubin(module: bytes | str, *, name: str = "", symbol_mapping: dict | None = None) -> ObjectCode:
+    def from_cubin(module: bytes | str | os.PathLike, *, name: str = "", symbol_mapping: dict | None = None) -> ObjectCode:
         """Create an :class:`ObjectCode` instance from an existing cubin.
 
         Parameters
@@ -638,7 +641,7 @@ cdef class ObjectCode:
         return ObjectCode._init(module, ObjectCodeFormatType.CUBIN, name=name, symbol_mapping=symbol_mapping)
 
     @staticmethod
-    def from_ptx(module: bytes | str, *, name: str = "", symbol_mapping: dict | None = None) -> ObjectCode:
+    def from_ptx(module: bytes | str | os.PathLike, *, name: str = "", symbol_mapping: dict | None = None) -> ObjectCode:
         """Create an :class:`ObjectCode` instance from an existing PTX.
 
         Parameters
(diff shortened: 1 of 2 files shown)


acosmicflamingo:
This could also be the updated documentation for the change:
```diff
diff --git a/cuda_core/cuda/core/_module.pyx b/cuda_core/cuda/core/_module.pyx
index ea334aae07..b5b0b441aa 100644
--- a/cuda_core/cuda/core/_module.pyx
+++ b/cuda_core/cuda/core/_module.pyx
@@ -628,9 +628,10 @@ cdef class ObjectCode:
 
         Parameters
         ----------
-        module : Union[bytes, str]
-            Either a bytes object containing the in-memory cubin to load, or
-            a file path string pointing to the on-disk cubin to load.
+        module : Union[bytes, str, os.PathLike]
+            Either a bytes object containing the in-memory cubin to load,
+            the file path pointing to the on-disk cubin to load,
+            or the string representation of the file path
         name : Optional[str]
             A human-readable identifier representing this code object.
         symbol_mapping : Optional[dict]
```

cuda_core/cuda/core/_module.pyx
```diff
@@ -733,8 +736,18 @@ cdef class ObjectCode:
         if self._h_library:
             return 0
         module = self._module
-        assert_type_str_or_bytes_like(module)
+        try:
```
kkraus14: Can we switch this to branch-based dispatch and put the built-in cases first for performance? In Cython, `isinstance(module, str)` / `isinstance(module, (bytes, bytearray))` compile to cheap C API checks (`PyUnicode_Check`, `PyBytes_Check`, `PyByteArray_Check`), while `os.PathLike` remains a dynamic `PyObject_IsInstance` path. The current exception-driven form also makes every existing `str`/`bytes`/`bytearray` caller raise and catch `TypeError` before loading the module.

A shape like this keeps the common cases cheap and avoids using exceptions for normal control flow:

```python
if isinstance(module, str):
    path_bytes = module.encode()
    ...
elif isinstance(module, (bytes, bytearray)):
    ...
elif isinstance(module, os.PathLike):
    path_bytes = os.fsencode(module)
    ...
else:
    assert_type_str_or_bytes_like(module)
```

I would also avoid moving `os.fsencode(module)` into the `try`/`except TypeError`; a `TypeError` from a broken `__fspath__()` should surface as the real conversion failure rather than being reclassified as a non-PathLike input.

cuda_core/cuda/core/_module.pyx
```diff
@@ -620,7 +623,7 @@ cdef class ObjectCode:
         return ObjectCode._reduce_helper, (self._module, self._code_type, self._name, self._sym_map)
 
     @staticmethod
-    def from_cubin(module: bytes | str, *, name: str = "", symbol_mapping: dict | None = None) -> ObjectCode:
+    def from_cubin(module: bytes | str | os.PathLike, *, name: str = "", symbol_mapping: dict | None = None) -> ObjectCode:
```
kkraus14: Since these signatures now accept `os.PathLike`, can you update the matching public docs/type surfaces too?

A few places look stale after this change:

- The `module` docstrings below still say `Union[bytes, str]` and describe only file path strings. They should include `os.PathLike` / path-like file paths for each updated constructor.
- `CodeTypeT` still excludes `os.PathLike`, but `ObjectCode.code` returns the original object passed to `_init()`, so it can now return a path-like object.
- `from_object()` and `from_library()` still advertise `bytes | str`, but this shared `_lazy_load_module()` implementation would accept `os.PathLike` for them too. Either update those constructors/docstrings as well, or explicitly reject `PathLike` there if the support is intended only for cubin/ptx/ltoir/fatbin.

cuda_core/cuda/core/_module.pyx
acosmicflamingo: Oof, performance of that commit was certainly nowhere close to speed-of-light ;) my apologies! Thank you for the very great suggestions! I rewrote the logic so it's as performant as it can be.

cuda_core/cuda/core/_module.pyx
acosmicflamingo: Absolutely I can!

acosmicflamingo: Now that you bring up the `from_object` and `from_library` API, do you think this PR should be reworked so it makes resolving [this](https://github.com/NVIDIA/cuda-python/issues/663) the main priority? I was going to append tests to make sure the `os.PathLike` behavior works for those functions, but there aren't any tests to begin with (as that ticket states).

Or, should I make sure this PR is only applicable for integrating path-like support to cubin/ptx/ltoir/fatbin and then I open a separate one to add the library/object tests (which would also include the path-like support)? I'm happy to go with either of the two options.

kkraus14: Let's focus this PR on integrating the path-like support and then a follow up PR for the test gaps.

Additionally, there was an offline discussion related to how we want to handle the return type of `ObjectCode.code` and the conclusion we came to is that we don't want to expand the possible return types, so we want to convert pathlike objects to strings before storing them in the internal attribute that's used for the `ObjectCode.code` property.

cuda_core/cuda/core/_module.pyx
```diff
@@ -733,21 +735,24 @@ cdef class ObjectCode:
...
-            return 0
-        if isinstance(module, (bytes, bytearray)):
+        elif isinstance(module, (bytes, bytearray)):
             self._h_library = create_library_handle_from_data(<const void*><char*>module)
-            if not self._h_library:
-                HANDLE_RETURN(get_last_error())
-            return 0
-        raise_code_path_meant_to_be_unreachable()
-        return -1
+        elif PyObject_HasAttrString(module, "__fspath__"):
```
kkraus14: What is the benefit of doing this versus using `isinstance(module, os.PathLike)`?

acosmicflamingo: Yes, that conditional sure looks like an odd duck out of all the others. When you said that cython is able to compile code like `isinstance(module, str)` to cheap C API checks, I took it to mean that in this context, it's more performant to use it for the simpler data types like `str`, `bytes` and that it could be more expensive for types that conform to `os.PathLike`. Since any class implements the `__fspath__` dunder function can be seen as path-like, I thought the cleanest way was to just use the `PyObject_HasAttrString` function.

Looking at what's actually being compiled, here's the check for `PyObject_HasAttrString`:
```C
  /* "cuda/core/_module.pyx":748
 *         elif isinstance(module, (bytes, bytearray)):
 *             self._h_library = create_library_handle_from_data(<const void*><char*>module)
 *         elif PyObject_HasAttrString(module, "__fspath__"):             # <<<<<<<<<<<<<<
 *             path_str = module.__fspath__()
 *             assert_type_str_or_bytes_like(path_str)
*/
  __pyx_t_3 = PyObject_HasAttrString(__pyx_v_module, __pyx_k_fspath);
  if (__pyx_t_3) {
```

If I were to use `isinstance(module, os.PathLike)`, there'll be a tiny bit more dereferencing going on:
```C
 /* "cuda/core/_module.pyx":748
 *         elif isinstance(module, (bytes, bytearray)):
 *             self._h_library = create_library_handle_from_data(<const void*><char*>module)
 *         elif isinstance(module, os.PathLike):             # <<<<<<<<<<<<<<
 *             path_str = module.__fspath__()
 *             assert_type_str_or_bytes_like(path_str)
*/
  __Pyx_GetModuleGlobalName(__pyx_t_2, __pyx_mstate_global->__pyx_n_u_os); if (unlikely(!__pyx_t_2)) __PYX_ERR(0, 748, __pyx_L1_error)
  __Pyx_GOTREF(__pyx_t_2);
  __pyx_t_4 = __Pyx_PyObject_GetAttrStr(__pyx_t_2, __pyx_mstate_global->__pyx_n_u_PathLike); if (unlikely(!__pyx_t_4)) __PYX_ERR(0, 748, __pyx_L1_error)
  __Pyx_GOTREF(__pyx_t_4);
  __Pyx_DECREF(__pyx_t_2); __pyx_t_2 = 0;
  __pyx_t_3 = PyObject_IsInstance(__pyx_v_module, __pyx_t_4); if (unlikely(__pyx_t_3 == ((int)-1))) __PYX_ERR(0, 748, __pyx_L1_error)
  __Pyx_DECREF(__pyx_t_4); __pyx_t_4 = 0;
  if (__pyx_t_3) {
```

Meanwhile using `isinstance(module, str)` gives you a one-liner:
```C
  /* "cuda/core/_module.pyx":743
 *         module = self._module
 *         cdef bytes path_bytes
 *         if isinstance(module, str):             # <<<<<<<<<<<<<<
 *             path_bytes = module.encode()
 *             self._h_library = create_library_handle_from_file(<const char*>path_bytes)
*/
  __pyx_t_3 = PyUnicode_Check(__pyx_v_module); 
  if (__pyx_t_3) {
```

With all that being said, I can see now that what you were trying to say was more that `os.PathLike` being a more complicated data type is why I should make sure that both `isinstance(module, str)` and `isinstance(module, bytes)` are called first without any of that exception-catching logic, not that I should outright avoid using `os.PathLike`.

I can revert that change so it looks like this if you'd prefer :)
```python
diff --git a/cuda_core/cuda/core/_module.pyx b/cuda_core/cuda/core/_module.pyx
index 71980a1f98..af693c25af 100644
--- a/cuda_core/cuda/core/_module.pyx
+++ b/cuda_core/cuda/core/_module.pyx
@@ -745,10 +745,8 @@ cdef class ObjectCode:
             self._h_library = create_library_handle_from_file(<const char*>path_bytes)
         elif isinstance(module, (bytes, bytearray)):
             self._h_library = create_library_handle_from_data(<const void*><char*>module)
-        elif PyObject_HasAttrString(module, "__fspath__"):
-            path_str = module.__fspath__()
-            assert_type_str_or_bytes_like(path_str)
-            path_bytes = path_str.encode() if isinstance(path_str, str) else path_str
+        elif isinstance(module, os.PathLike):
+            path_bytes = os.fsencode(module)
             self._h_library = create_library_handle_from_file(<const char*>path_bytes)
         else:
             assert_type_str_or_bytes_like(module)
```

cuda_core/cuda/core/_module.pyx
acosmicflamingo: Great! That makes things pretty clean for me ;) 

In that case, I can push this:
```diff
diff --git a/cuda_core/cuda/core/_module.pyx b/cuda_core/cuda/core/_module.pyx
index 71980a1f98..8781db4e11 100644
--- a/cuda_core/cuda/core/_module.pyx
+++ b/cuda_core/cuda/core/_module.pyx
@@ -608,7 +608,11 @@ cdef class ObjectCode:
         self._h_library = LibraryHandle()  # Empty handle
 
         self._code_type = str(code_type)
-        self._module = module
+
+        if isinstance(module, os.PathLike):
+            self._module = os.fspath(module)
+        else:
+            self._module = module
         self._sym_map = {} if symbol_mapping is None else symbol_mapping
         self._name = name if name else ""
 
# Will apply these to ptx/fatbin/etc
diff --git a/cuda_core/tests/test_module.py b/cuda_core/tests/test_module.py
index 2523d40adb..3a438f825a 100644
--- a/cuda_core/tests/test_module.py
+++ b/cuda_core/tests/test_module.py
@@ -244,7 +244,7 @@ def test_object_code_load_ptx_from_file(get_saxpy_kernel_ptx, tmp_path, convert_
     ptx_file.write_bytes(ptx)
     arg = convert_path(ptx_file)
     mod_obj = ObjectCode.from_ptx(arg, symbol_mapping=sym_map)
-    assert mod_obj.code == arg
+    assert mod_obj.code == str(arg)
     assert mod_obj.code_type == "ptx"
     if not _can_load_generated_ptx():
         pytest.skip("PTX version too new for current driver")
```

cuda_core/cuda/core/_module.pyx
kkraus14: Have you happened to do any microbenchmarking of this? I suspect the performance difference here is quite minimal and the `isinstance(module, os.PathLike)` is much clearer code intention wise. We could also store a local to `os.PathLike` to avoid having to look it up in every use.

cuda_core/cuda/core/_module.pyx
kkraus14: Yes, other than instead of checking for `isinstance(module, os.PathLike)` we should fastpath strings and bytes first.

cuda_core/cuda/core/_module.pyx
acosmicflamingo: Ah, I had not done any microbenchmarking when I wrote that 😅 

However, I was curious myself and found some interesting results.

The cython file:
```cython
import os

from cpython.object cimport PyObject_HasAttrString

def test_isinstance(x, int n):
    cdef int i
    for i in range(n):
        isinstance(x, os.PathLike)

def test_hasattrstring(x, int n):
    cdef int i
    for i in range(n):
        PyObject_HasAttrString(x, "__fspath__")
```

Here's the benchmarking script I wrote:
```python
import pathlib
import timeit

from functools import partial

import numpy as np
import pyximport
pyximport.install()

import bench

x = pathlib.Path('.').expanduser()
n = 1_000_000 #_000_000

t1_times = []
t2_times = []
for i in range(0, 100):
    t1 = timeit.timeit(partial(bench.test_isinstance, x, n), number=1)
    t2 = timeit.timeit(partial(bench.test_hasattrstring, x, n), number=1)
    t1_times.append(t1)
    t2_times.append(t2)

for idx, times in enumerate([t1_times, t2_times]):
    print(
f"""t{ idx }:
- max: { max(times) } s
- min: { min(times) } s
- avg: { np.average(times) } s
- std: { np.std(times) } s
"""
)
```

From that, I got these results
```
t0:
- max: 0.1483691829998861 s
- min: 0.06976235900037864 s
- avg: 0.11347771822997857 s
- std: 0.02798347527330753 s

t1:
- max: 0.10476064799968299 s
- min: 0.045422338000207674 s
- avg: 0.07302942716995858 s
- std: 0.019299640757812583 s
```

Of course, each collected time represents the time it took to run the check a million times. If I were to remove the looping logic and then increase the number of times I run the timeit commands to a million, I get something like this:
```
t0:
- max: 7141.460000639199 μs
- min: 1.720000000204891 μs
- avg: 4.733994145299221 μs
- std: 12.200230006400012 μs

t1:
- max: 3081.579998251982 μs
- min: 1.2599957699421793 μs
- avg: 3.3473681767736707 μs
- std: 7.80333719000798 μs
```

In summary, there is (expectedly) so much variance between each run, and the time "saved" greatly pales in comparison to maintaining simple, elegant code. I'm in favor of using `isinstance(module, os.PathLike)`.

cuda_core/cuda/core/_module.pyx
acosmicflamingo: In addition to providing a fastpath for strings/bytes/bytearray, should I throw in a `assert_type_str_or_bytes_like` in the `else` clause?
```diff
         self._h_library = LibraryHandle()  # Empty handle
 
         self._code_type = str(code_type)
-        self._module = module
+
+        if isinstance(module, CodeTypeT):
+            self._module = module
+        elif isinstance(module, os.PathLike):
+            self._module = os.fspath(module)
+        else:
+            self._module = module
```

cuda_core/cuda/core/_module.pyx
kkraus14: Based on this let's move forward with the `isinstance(module, os.PathLike)` approach (ideally creating a local to `os.PathLike` so we don't have to do the lookup from `os` each time.

acosmicflamingo: Done!

cuda_core/cuda/core/_module.pyx
```diff
@@ -606,7 +607,13 @@ cdef class ObjectCode:
         self._h_library = LibraryHandle()  # Empty handle
 
         self._code_type = str(code_type)
-        self._module = module
+
+        if isinstance(module, CodeTypeT):
```
kkraus14: Can we confirm this still retains the fast cpp generated code from Cython?

kkraus14 (approved):

kkraus14:
/ok to test https://github.com/NVIDIA/cuda-python/pull/2123/commits/837e70c64cadc0980f55085ea964d1abdabbd47a

acosmicflamingo:
Whoops, I expect tests to fail because `os.fsencode` and `os.fspath` is still there after I removed `import os`. I'll push fix after verifying that cython injects clean C-API in that `isinstance(modules, CodeTypeT)` call.

cuda_core/cuda/core/_module.pyx
acosmicflamingo: Ah, it does not! It generates this instead:
```C
/* "cuda/core/_module.pyx":611
 *         self._code_type = str(code_type)
 * 
 *         if isinstance(module, CodeTypeT):             # <<<<<<<<<<<<<<
 *             self._module = module
 *         elif isinstance(module, PathLike):
*/
  __Pyx_GetModuleGlobalName(__pyx_t_4, __pyx_mstate_global->__pyx_n_u_CodeTypeT); if (unlikely(!__pyx_t_4)) __PYX_ERR(0, 611, __pyx_L1_error)
  __Pyx_GOTREF(__pyx_t_4);
  __pyx_t_1 = PyObject_IsInstance(__pyx_v_module, __pyx_t_4); if (unlikely(__pyx_t_1 == ((int)-1))) __PYX_ERR(0, 611, __pyx_L1_error)
  __Pyx_DECREF(__pyx_t_4); __pyx_t_4 = 0;
  if (__pyx_t_1) {
```
I'll add the original suggestion back.

kkraus14:
/ok to test https://github.com/NVIDIA/cuda-python/pull/2123/commits/6ed4a7fda1226472483ed657322462ac64d2fb79

kkraus14:
We require maintainers to trigger CI 😄

kkraus14 (approved):

kkraus14:
Thanks for the contribution @acosmicflamingo!

acosmicflamingo:
Woohoo! Glad I could help out 😄

PR #2121, opened 2026-05-21
tests: kill zombie IPC child processes after join timeout

aryanputta:
## What

`TestIpcSendBuffers` and `TestIpcReexport` call `process.join(timeout=30)` to wait for child processes but never kill them if the join times out. This causes multi-hour CI hangs.

## Root cause

`ci/tools/env-vars` enables compute-sanitizer (`--target-processes=all`) on Python 3.12 + local CTK + Linux — exactly the combination in every hanging run from issue #2004. With `--target-processes=all`, compute-sanitizer attaches to every `mp.Process` child. On CUDA 12.9.1 its IPC teardown analysis gets stuck, so the child never exits.

The parent's `join(timeout=30)` returns after 30 seconds, but the child is still alive. That zombie holds an open IPC memory handle. When pytest runs fixture teardown (`ipc_memory_resource`'s `mr.close()`), it blocks on the held handle indefinitely. The runner gets tied up for 4–6 hours until GitHub Actions force-cancels the job.

## Fix

After `join(timeout=...)`, kill any process that is still alive. This forces IPC handle release before fixture teardown runs. The test still fails (exit code is non-zero, or `completed` is False), just in seconds instead of hours.

No behavior change when children exit cleanly.

## Related

Closes #2004

PR #2124, opened 2026-05-21
tests: layered defense against IPC child-process hangs (#2004)

### LINKED ISSUE #2004: memory_ipc/test_send_buffers.py::TestIpcReexport::test_main[DeviceMR] hanging in CI

Andy-Jost:
## Summary

Implements layered fixes that defend the IPC test suite against failures that wedge CI. With these layers in place, no IPC test can wedge the GHA runner for hours, even if a new failure mode emerges.

Deadlock was observed due to the following sequence:

1. The parent process spawns a child and waits with a 30 second timeout.
2. The child process fails to complete in 30 seconds and eventually blocks while attempting communication with the parent.
3. The parent moves on, eventually blocking on `mr.close()` because the subprocess holds an open IPC handle.

This is a bug in the test source: the result of the parent's wait needs to be checked. #2121 fixes some of these tests; this PR generalizes the fix to every IPC test.

This was observed in Python 3.12 test runs, which run under compute-sanitizer, presumably because compute-sanitizer slows down process spawning.

## Changes

**CI layer.** `ci/tools/setup-sanitizer` now passes `--target-processes=application-only` so spawned `multiprocessing.Process` children run without compute-sanitizer attached. This should improve the start-up time for spawned processes.

**In-test layer.** New helper `helpers.child_processes.kill_subprocesses(*processes)` kills any Process objects still alive at call time and returns them. Every `process.join(timeout=...)` site in `tests/memory_ipc/` now follows the canonical pattern below, so a timeout produces a clean failure message ("timed out waiting on: ['Process-3']") instead of `assert None == 0`, and held IPC handles are released before any further test code runs:

```python
process.join(timeout=CHILD_TIMEOUT_SEC)
survivors = kill_subprocesses(process)
assert not survivors, "child did not exit within timeout"
assert process.exitcode == 0
```

For `test_send_buffers.py:TestIpcReexport` (the test from #2121), the `event_c.wait` return value is also captured and asserted on, so a timeout there is reported explicitly instead of being dropped on the floor.

**Fixture layer.** `ipc_device` and `ipc_mempool_device_x2` now wrap their `yield` in `track_child_processes()`, a context manager in `helpers.child_processes` that patches `multiprocessing.process.BaseProcess.__init__` to record every `Process` instance constructed during the test and kills any survivor at teardown. This is defense in depth for tests that may still slip through the in-test layer — it protects `ipc_memory_resource.mr.close()` from blocking on IPC handles held by a stuck child, regardless of the original failure mode.

**Outer-guard layer.** `pytest-timeout` is added to the test dep group, and `tests/memory_ipc/conftest.py` applies `pytest.mark.timeout(child_timeout_sec() + 30)` to every test in the directory (60 s without compute-sanitizer, 150 s under it). IPC test children run concurrently, so a well-behaved test finishes in ~`CHILD_TIMEOUT_SEC` regardless of how many joins it chains; exceeding that already means something is genuinely stuck, at which point the outer guard firing is the right outcome (the autouse `track_child_processes()` context manager still cleans up survivors). A meta-test `test_outer_timeout_marker_is_applied` in `tests/memory_ipc/test_errors.py` verifies via `request.node.get_closest_marker("timeout")` that the per-directory conftest is being loaded and applies the marker with the expected value; if a future change breaks conftest discovery or the formula drifts, this test fails immediately with a clear message.

(`tests/memory_ipc/` also gains an empty `__init__.py` so its conftest is imported as `memory_ipc.conftest` rather than colliding with the top-level `conftest` module that other tests reference via `from conftest import ...`.)

**Timeout helper.** The hardcoded `CHILD_TIMEOUT_SEC = 30` in every IPC test module is replaced with a call to `helpers.child_processes.child_timeout_sec()`, which returns 30 by default and 120 under compute-sanitizer. The unused `CHILD_TIMEOUT_SEC` declaration in `test_workerpool.py` is removed.

## Test Coverage

- The helpers are exercised by the existing IPC suite: every IPC test runs inside the `track_child_processes()` context manager via `ipc_device` (and `ipc_mempool_device_x2`), and every join/exitcode pair now goes through `kill_subprocesses`. Happy-path tests verify no regression; any test that hits the original hang will fail fast with a clear timeout message instead of wedging.
- The meta-test `test_outer_timeout_marker_is_applied` will fail if `tests/memory_ipc/conftest.py` is not loaded or does not apply the timeout marker.
- Local smoke tests confirm: `kill_subprocesses` returns a list of still-alive processes (after killing them with SIGKILL, exitcode `-9`), an empty list for cleanly-exited or never-started Process objects, and never raises on either kind. `track_child_processes()` correctly kills a child still alive at scope exit and leaves cleanly-exiting children untouched.
- `child_timeout_sec()` returns 30 by default and 120 when `CUDA_PYTHON_UNDER_SANITIZER=1` (or any of the other signals checked by `under_compute_sanitizer()`).

## Related Work

- Closes #2004.
- Supersedes #2121 (this PR generalizes the per-test fix to every IPC test via `kill_subprocesses` and adds an autouse cleanup tracker in `ipc_device`).

### DIFF
--- a/cuda_core/tests/conftest.py
+++ b/cuda_core/tests/conftest.py
@@ -213,14 +213,24 @@ def deinit_all_contexts_function():
 
 @pytest.fixture
 def ipc_device():
-    """Obtains a device suitable for IPC-enabled mempool tests, or skips."""
+    """Obtains a device suitable for IPC-enabled mempool tests, or skips.
+
+    The fixture also tracks every ``multiprocessing.Process`` spawned during
+    the test and kills any survivors at teardown. This prevents a stuck child
+    (e.g., compute-sanitizer wedged during IPC teardown -- see issue #2004)
+    from blocking ``ipc_memory_resource``'s ``mr.close()`` for hours.
+    """
+    from helpers.child_processes import track_child_processes
+
     device = Device(0)
     device.set_current()
 
     if not device.properties.memory_pools_supported:
         pytest.skip("Device does not support mempool operations")
 
-    return _require_ipc_mempool_devices((device,))[0]
+    device = _require_ipc_mempool_devices((device,))[0]
+    with track_child_processes():
+        yield device
 
 
 @pytest.fixture(
@@ -291,8 +301,16 @@ def mempool_device_x3():
 
 @pytest.fixture
 def ipc_mempool_device_x2(mempool_device_x2):
-    """Fixture that provides two IPC-capable mempool devices, or skips."""
-    return _require_ipc_mempool_devices(mempool_device_x2)
+    """Fixture that provides two IPC-capable mempool devices, or skips.
+
+    Also tracks/kills any leftover ``multiprocessing.Process`` children at
+    teardown for the same reasons documented on :func:`ipc_device`.
+    """
+    from helpers.child_processes import track_child_processes
+
+    devices = _require_ipc_mempool_devices(mempool_device_x2)
+    with track_child_processes():
+        yield devices
 
 
 @pytest.fixture(
--- a/ci/tools/setup-sanitizer
+++ b/ci/tools/setup-sanitizer
@@ -14,11 +14,12 @@ if [[ "${SETUP_SANITIZER}" == 1 ]]; then
   COMPUTE_SANITIZER_VERSION=$(${COMPUTE_SANITIZER} --version | grep -Eo "[0-9]{4}\.[0-9]\.[0-9]" | sed -e 's/\.//g')
   # --target-processes=application-only: attach the sanitizer to the parent
   # pytest process only. Spawned multiprocessing.Process children run without
-  # the sanitizer. This avoids a class of CI hangs where compute-sanitizer's
-  # IPC teardown analysis wedges a child on certain CUDA driver / toolkit
-  # combinations (see issue #2004). The parent process is still fully
-  # sanitized, which is where most of the interesting host-side IPC plumbing
-  # runs anyway.
+  # the sanitizer. This aims to mitigate a class of CI hangs where child
+  # processes take an extreme amount of time to spawn (>30 seconds). Test bugs
+  # triggered by that specific condition are typically uncovered only in CI,
+  # where they become emergencies and are difficult to debug. The parent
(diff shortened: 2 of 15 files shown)


cuda_core/tests/conftest.py
```diff
@@ -213,14 +213,24 @@ def pop_all_contexts():
...
+
     device = Device(0)
     device.set_current()
 
     if not device.properties.memory_pools_supported:
         pytest.skip("Device does not support mempool operations")
 
-    return _require_ipc_mempool_devices((device,))[0]
+    device = _require_ipc_mempool_devices((device,))[0]
+    with track_child_processes():
```
rwgk, line 232: Non-blocking: since `ipc_memory_resource` depends on `ipc_device`, its teardown still runs before this `track_child_processes()` context exits. So this fixture-level guard may not protect `mr.close()` if a test fails before reaching its in-test `kill_subprocesses()` cleanup. The per-test cleanup and `pytest-timeout` still cover the known hang, so I don’t think this should block.

rwgk (approved):
One caveat identified by GPT-5.5. But the CI succeeds now, so it can wait until later, or if we actually see problems.

rwgk:
I merged: I'll update #2118 next.

PR #2128, opened 2026-05-22
cuda.bindings: add decode_c_str helper for actionable UnicodeDecodeError (#2122)

### LINKED ISSUE #2122: [ENH]: Make cuda_bindings UnicodeDecodeError more actionable
## Motivation

When CUDA Python receives a string from an underlying CUDA library, we generally expect it to decode cleanly into a Python `str`. When that assumption is wrong, the resulting `UnicodeDecodeError` is often technically accurate but not very actionable.

For users, the important question is not just "which byte failed to decode?" It is also:

- Which CUDA API returned the non-decodable string?
- What bytes did the CUDA library actually return?
- Is there any readable fragment in the returned data that helps identify the underlying issue?

PR #2118 exposed this gap while investigating a WSL/NVML process-name issue. The root cause there is handled separately, but the debugging experience showed that our current error message leaves out the information most useful for diagnosing driver/library string issues.

Today, if a CUDA library returns bytes that cannot be decoded as UTF-8, Python's default `UnicodeDecodeError` reports the codec, the first bad byte, and the byte offset. That is useful, but incomplete.

For example, an error like this:

```text
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf8 in position 0: invalid start byte
```

does not say which CUDA API produced the bytes, nor does it show the returned buffer. Without that context, users and maintainers have to add ad hoc instrumentation or repro scripts before they can tell whether the CUDA library returned garbage, a string in an unexpected encoding, an unterminated buffer, or some other invalid data.

This makes rare string-decoding failures much harder to report, triage, and escalate to the owning CUDA library team.

## Suggested enhancement

Add a small internal helper for decoding C strings returned by CUDA libraries. The helper should keep the successful path unchanged: if the bytes decode normally, return the same Python `str` as today.

On decode failure, the helper should raise an error with additional context, including at least:

- The CUDA API or binding function that produced the string.
- The original decode failure.
- A bounded representation of the returned bytes, such as a capped `repr` or hex dump up to the first NUL byte or a fixed maximum length.

The goal is better diagnostics, not silent recovery. The helper should not guess encodings, run expensive detection, or return replacement-decoded strings. It should make failures easier to understand while preserving existing behavior for valid strings.

This can start with the string-returning APIs where we already convert C buffers to Python strings, and can be expanded incrementally as needed.

aryanputta:
## Summary

Adds `cuda.bindings._internal.strdecode.decode_c_str(data, api_name)`. Success path = `bytes.decode()` as today. On `UnicodeDecodeError`, re-raises the same exception type with `reason` extended to include the originating CUDA API and a bounded hex preview of the buffer.

Implements the helper proposed inline on #2118 ("a tiny helper... source API name plus a bounded repr/hex dump"). #2118 fixed the WSL symptom in `cuda_core`; this closes the diagnostics gap in `cuda_bindings`.

## Design
- Pure-Python `_internal/strdecode.py` next to `_fast_enum.py` — no Cython rebuild
- Same exception type; original chained via `from e`
- Hex preview, stops at first NUL with `stopped at NUL@<offset>` marker,
  caps at 64 bytes with `+N more`

## Example
```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf8 in position 0:
invalid start byte (returned by nvmlSystemGetProcessName;
bytes=<5 bytes; hex='f8 9a 80 80 af'>)
```

## Scope of this PR

Two files, both hand-written:
- `cuda_bindings/cuda/bindings/_internal/strdecode.py` — the helper
- `cuda_bindings/tests/test_strdecode.py` — 11 unit tests

The existing `_output_.decode()` sites in `nvml.error_string`,
`nvvm.get_error_string`, `nvfatbin.get_error_string`, and
`cufile.cufileop_status_error` are good candidates to adopt this helper.
Those modules are maintained on the NVIDIA side, so wiring is left to the
maintainers rather than included here.

Refs #2118. Closes #2122.

### DIFF
--- a/cuda_bindings/cuda/bindings/_internal/strdecode.py
+++ b/cuda_bindings/cuda/bindings/_internal/strdecode.py
@@ -1,0 +1,42 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
+"""Decode C strings returned by CUDA libraries with actionable failure context."""
+
+from __future__ import annotations
+
+# Cap mirrors the #2118 mojibake size; keeps exception text readable in logs.
+_PREVIEW_MAX_BYTES = 64
+
+
+def _bounded_hex_preview(data: bytes, max_bytes: int = _PREVIEW_MAX_BYTES) -> str:
+    # NUL terminates C strings; bytes past it are not the returned value.
+    # Marker is explicit so a reader does not misread truncation as empty.
+    nul = data.find(b"\x00")
+    nul_stopped = nul != -1
+    visible_end = len(data) if not nul_stopped else nul
+    snippet_end = min(visible_end, max_bytes)
+    snippet = data[:snippet_end]
+    body = snippet.hex(" ") if snippet else ""
+    parts = []
+    if snippet_end < visible_end:
+        parts.append(f"+{visible_end - snippet_end} more")
+    if nul_stopped:
+        parts.append(f"stopped at NUL@{nul}")
+    suffix = f" ...({'; '.join(parts)})" if parts else ""
+    return f"<{visible_end} bytes; hex='{body}'{suffix}>"
+
+
+def decode_c_str(data: bytes, api_name: str) -> str:
+    """Decode ``data`` as UTF-8, or raise ``UnicodeDecodeError`` with ``api_name`` and a bounded hex preview in ``reason``.
+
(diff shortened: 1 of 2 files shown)


aryanputta:
Requesting review when you have a chance. I kept this PR scoped to the helper itself. If this direction looks good, I can send a follow-up wiring it into the current decode sites in nvml, nvvm, nvfatbin, and cufile.

rwgk:
Hi @aryanputta, unfortunately we cannot accept external contributions to cuda-bindings :-(

It's not compatible with the cuda-bindings LICENSE terms.

I should probably add that explanation to the [automatic bot comment above](https://github.com/NVIDIA/cuda-python/pull/2128#issuecomment-4514730739).

PR #2134, opened 2026-05-22
[doc-only] cuda.core: add docstrings to Host properties and public StrEnum types

aryanputta:
## Summary

Contributes to the doc-finalization step of #2131.

Two doc-only changes to `cuda.core`, touching no logic:

### `cuda/core/_host.py` — Host property docstrings

`Host` is a new 1.1.0 API (release notes: #2131 checklist item "Finalize the doc update"). The class-level docstring is already complete, but `numa_id` and `is_numa_current` had no docstrings, leaving them blank in generated API docs and `help()`:

```python
@property
def numa_id(self) -> int | None:
    """NUMA node ID, or ``None`` if not pinned to a specific NUMA node."""

@property
def is_numa_current(self) -> bool:
    """``True`` if this ``Host`` represents the calling thread's NUMA node (constructed via :meth:`numa_current`)."""
```

### `cuda/core/typing.py` — StrEnum class docstrings

All 12 public `StrEnum` subclasses (`SourceCodeType`, `ObjectCodeFormatType`, `CompilerBackendType`, `PCHStatusType`, `GraphConditionalType`, `GraphMemoryType`, `ManagedMemoryLocationType`, `VirtualMemoryHandleType`, `VirtualMemoryLocationType`, `VirtualMemoryGranularityType`, `VirtualMemoryAccessType`, `VirtualMemoryAllocationType`) previously had zero class-level docstrings. Each now has a short description and a one-liner per member so `help()` and Sphinx show meaningful output.

## Checklist

- [x] `cuda.core` only — no `cuda.bindings` files touched
- [x] No logic changes — docstrings only
- [x] Consistent with existing single-line property docstring style (`Host.numa_current`, `ManagedBuffer` methods)

### DIFF
--- a/cuda_core/cuda/core/_host.py
+++ b/cuda_core/cuda/core/_host.py
@@ -62,10 +62,12 @@ class Host:
 
     @property
     def numa_id(self) -> int | None:
+        """NUMA node ID, or ``None`` if not pinned to a specific NUMA node."""
         return self._numa_id
 
     @property
     def is_numa_current(self) -> bool:
+        """Whether this ``Host`` represents the calling thread's NUMA node (constructed via :meth:`numa_current`)."""
         return self._is_numa_current
 
     @classmethod
--- a/cuda_core/cuda/core/_host.py
+++ b/cuda_core/cuda/core/_host.py
@@ -67,7 +67,7 @@ class Host:
 
     @property
     def is_numa_current(self) -> bool:
-        """Whether this ``Host`` represents the calling thread's NUMA node (constructed via :meth:`numa_current`)."""
+        """``True`` if this ``Host`` represents the calling thread's NUMA node (constructed via :meth:`numa_current`)."""
         return self._is_numa_current
 
     @classmethod
--- a/cuda_core/cuda/core/typing.py
+++ b/cuda_core/cuda/core/typing.py
@@ -43,12 +43,28 @@ ProcessStateType = _Literal["running", "locked", "checkpointed", "failed"]
 
 
 class SourceCodeType(StrEnum):
+    """Source language passed to :class:`~cuda.core.Program`.
+
+    ``CXX`` selects CUDA C++, ``PTX`` selects PTX assembly text, and
+    ``NVVM`` selects NVVM IR (LLVM bitcode).
+    """
+
     CXX = "c++"
     PTX = "ptx"
     NVVM = "nvvm"
 
 
 class ObjectCodeFormatType(StrEnum):
+    """Output format produced by :meth:`~cuda.core.Program.compile`.
+
+    ``PTX`` — PTX assembly text.
+    ``CUBIN`` — device-native CUDA binary.
+    ``LTOIR`` — LTO (link-time optimization) IR for later linking.
+    ``FATBIN`` — fat binary bundling multiple device images.
+    ``OBJECT`` — relocatable device object.
+    ``LIBRARY`` — device code library.
+    """
+
     PTX = "ptx"
     CUBIN = "cubin"
     LTOIR = "ltoir"
@@ -58,6 +74,14 @@ class ObjectCodeFormatType(StrEnum):
 
 
 class CompilerBackendType(StrEnum):
+    """Compiler backend selected via :class:`~cuda.core.ProgramOptions`.
+
+    ``NVRTC`` — NVIDIA Runtime Compilation.
+    ``NVVM`` — NVVM LLVM backend.
+    ``NVJITLINK`` — nvJitLink device-side linker.
+    ``DRIVER`` — CUDA driver PTX JIT compiler.
+    """
+
     NVRTC = "NVRTC"
     NVVM = "NVVM"
     NVJITLINK = "nvJitLink"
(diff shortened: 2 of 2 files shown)


aryanputta:
@leofang @rwgk — flagging this for your awareness in the context of #2131.

This is a doc-only contribution to `cuda.core` (no `cuda.bindings` files touched — I know external contributions there aren't permitted). Both changes are pure docstring additions, zero logic:

- **`_host.py`**: `Host.numa_id` and `Host.is_numa_current` were undocumented. Since `Host` is new in 1.1.0 and explicitly in the release notes, having blank property docs felt like the most directly useful thing I could help with for the doc-finalization step.
- **`typing.py`**: added class-level docstrings to all 11 public `StrEnum` types so `help()` and Sphinx show something meaningful.

Happy to split into two PRs if that's easier to review, or adjust wording on any of the docstrings.

aryanputta:
@Andy-Jost — one style question before this is review-ready.

For the `StrEnum` classes, I went with per-member descriptions inside the class docstring body, e.g.:

```python
class ObjectCodeFormatType(StrEnum):
    """Output format produced by :meth:`~cuda.core.Program.compile`.

    ``PTX`` — PTX assembly text.
    ``CUBIN`` — device-native CUDA binary.
    ``LTOIR`` — LTO (link-time optimization) IR for later linking.
    ``FATBIN`` — fat binary bundling multiple device images.
    ``OBJECT`` — relocatable device object.
    ``LIBRARY`` — device code library.
    """
```

The question is whether that level of detail is right for members that are fairly self-explanatory from their name and string value (`PTX = "ptx"` probably doesn't need "PTX assembly text"). Two options I see:

**A) Keep per-member descriptions** i tried to keep it consistent across all 11 types, useful in `help()` output, and makes Sphinx more readable without needing to cross-reference the CUDA driver docs.

**B) Class-level summary only, drop the member list** for this i led to a  shorter approach which  avoids redundancy for obvious members. Something like:

```python
class ObjectCodeFormatType(StrEnum):
    """Output format produced by :meth:`~cuda.core.Program.compile`.

    Accepted values: ``PTX``, ``CUBIN``, ``LTOIR``, ``FATBIN``, ``OBJECT``, ``LIBRARY``.
    """
```

Happy to go either way — just want to match the preferred convention here before this gets reviewed.

Andy-Jost:
Thanks for the doc-only contribution — this is useful for the 1.1.0 doc-finalization work.

**Style (A vs B):** Option **A** (per-member descriptions) matches what we already do elsewhere in `cuda.core` — see the `Host` class docstring in `_host.py`, which uses `*` bullets with em-dashes. Please keep the per-member detail, but format the enum member lines as an RST bullet list (blank line after the summary paragraph, then `* ``MEMBER`` — description` per line). Without bullets, Sphinx tends to run the member lines together into one paragraph even though `help()` looks fine.

**A few fixes before merge:**

1. **`VirtualMemoryAllocationType.PINNED`** — the description should not say "page-locked host memory." `PINNED` maps to `CU_MEM_ALLOCATION_TYPE_PINNED` (VMM physical allocation type); backing location is controlled by `location_type` (default `DEVICE`). Something like "pinned/non-migratable physical allocation (placement via `VirtualMemoryLocationType`)" would be more accurate.

2. **`CompilerBackendType`** — backend isn't selected via `ProgramOptions`; it's determined from program `code_type` and exposed on `Program.backend`. Please reword the class summary accordingly.

3. **Version notes** — `ManagedMemoryLocationType.HOST_NUMA` is marked "CUDA 13+ only," but `ManagedMemoryResource` as a whole already requires CUDA 13, so that note is a bit misleading. `VirtualMemoryAllocationType.MANAGED` is the member that's actually 13+-gated — worth a note there if we keep version callouts.

**Nits (optional):** unify `SourceCodeType` with the same bullet style as the others; `ObjectCodeFormatType` is also used by `Linker.link` / `Program.as_bytes`, not only `Program.compile`.

Otherwise looks good — no broken xrefs spotted, and the `Host` property docstrings are fine.

Andy-Jost (changes requested):
See comments. Bullets are needed to render correctly in Sphinx. There are some minor accuracy issues.

aryanputta:
Thanks for the thorough review @Andy-Jost. All points addressed in this push:

- **RST bullets**: converted all enum member lines to `* ``MEMBER`` -- description` format (blank line after summary, one bullet per member) across all StrEnum classes.
- **PINNED description**: removed "page-locked host memory" -- now reads "pinned/non-migratable physical allocation (placement via `VirtualMemoryLocationType`)" to reflect that backing location is device-controlled.
- **CompilerBackendType summary**: rewrote to "inferred from the program's code type and exposed on `Program.backend`" -- no longer implies ProgramOptions selects it.
- **Version note moved**: dropped CUDA 13+ from `ManagedMemoryLocationType.HOST_NUMA` (redundant since ManagedMemoryResource already requires 13); added it to `VirtualMemoryAllocationType.MANAGED` where it's actually meaningful.
- **Nits applied**: unified `SourceCodeType` to bullet style; expanded `ObjectCodeFormatType` summary to mention `Linker.link` and `Program.as_bytes` alongside `Program.compile`.

Andy-Jost (approved):
Looks good -- thanks for the contribution!

PR #2136, opened 2026-05-25
TST: Add remaining tests for ObjectCode.from_<library/object>

### LINKED ISSUE #663: [FEA]: Tests for `ltoir`, `fatbin`, `object`, and `library` constructors

acosmicflamingo:
Resolves #663 

Unlike adding tests for `ObjectCode.from<ptx/ltoir/cubin/fatbin>`, adding tests for `ObjectCode.from_<object/library>` was not as straightforward since `cuda.core.Program` module's `SUPPORTED_TARGETS` list do not include `ObjectCodeFormatType.OBJECT` or `ObjectCodeFormatType.LIBRARY`, so doing something like this throws an exception:
```
@pytest.fixture
def get_saxpy_kernel_object(init_cuda):
    # prepare program
    prog = Program(SAXPY_KERNEL, code_type="c++")
    mod = prog.compile(
        "object",
        name_expressions=("saxpy<float>", "saxpy<double>"),
    )
    return object
```

Going through different PRs and conversations related to testing for `ObjectCode.from_<object/library>` led me to the very illuminating discussion [here](https://github.com/NVIDIA/cuda-python/pull/652).

Based on the very helpful back-and-forth, I took the approach of generating the library/object files within the test itself (assuming an NVCC is present in the environment, otherwise test will be skipped). Not adding a Makefile would mean we have one less file to worry about maintaining, although it does mean that the kernel is being recompiled over-and-over again when it could just be compiled once, saved, and the test relies on the persisted binaries.

I'm open to refactoring this code, whether it means creating a Makefile, generating the binaries, persisting them in the codebase, and having the tests point to them, or another approach I have not considered.


rparolin:
Hi @acosmicflamingo — first, thank you for putting in the work here, and I'm genuinely sorry for the situation. Unfortunately, issue #663 was recently assigned to @lijinf2 as an introductory task for him to ramp up on the codebase, and he has work in flight against it. I really regret that you spent time on this without us catching it sooner.

To avoid this happening again in the future (to you or anyone else), the best practice is to comment on an issue and wait to be assigned before starting work. That way maintainers can flag if the issue is already spoken for. I checked our CONTRIBUTING.md and the cuda.core contributor guide and noticed we don't currently document this expectation explicitly. I'll work internally to ensure that text is added in an effort to avoid this situation in the future.

Again, apologies for the wasted effort, and thank you for the contribution. Please don't let this discourage you from picking up another issue.

acosmicflamingo:
@rparolin no worries at all! I know how busy y'all are and didn't want to spam your inboxes with yet another "Can I work on this" message. I also wouldn't say it's a wasted effort at all; I have a much better understanding of the inner-workings of the codebase by doing that. But I do apologize because my attempt at being proactive is stepping on toes of folks who have been assigned a particular task. I'll keep this in mind going forward. Thanks!

PR #2142, opened 2026-05-27
Fix cuda.core example links in docs

uday1o1:
## Summary

- Replace the `|cuda_core_github_ref|` substitution inside cuda.core GitHub example URLs with Sphinx `extlinks` roles.
- Update `examples.rst`, `getting-started.rst`, and `interoperability.rst` to use the new roles.
- Keep existing version-aware behavior so latest/preview docs link to `main`, while release docs link to `cuda-core-v{release}`.

## Root Cause

The `|cuda_core_github_ref|` replacement was defined through `rst_epilog`, but the substitution was used inside embedded hyperlink URL targets. Sphinx/Docutils leaves that placeholder literal in generated HTML, which browsers encode as `%7Ccuda_core_github_ref%7C`; GitHub then treats it as a branch/tag name and returns 404.

## Validation

- Built a Sphinx smoke test for latest docs and verified generated links use `main`.
- Built a Sphinx smoke test for release docs with `SPHINX_CUDA_CORE_VER=1.0.1` and verified generated links use `cuda-core-v1.0.1`.
- Checked all generated CUDA example links in the latest and 1.0.1 smoke builds; all returned HTTP 200.
- Confirmed generated HTML no longer contains `cuda_core_github_ref` or `%7Ccuda_core`.
- Ran `git diff --check`.

### DIFF
--- a/cuda_core/docs/source/conf.py
+++ b/cuda_core/docs/source/conf.py
@@ -46,6 +46,7 @@ extensions = [
     "sphinx.ext.autosummary",
     "sphinx.ext.napoleon",
     "sphinx.ext.intersphinx",
+    "sphinx.ext.extlinks",
     "myst_nb",
     "sphinx_copybutton",
     "sphinx_toolbox.more_autodoc.autoprotocol",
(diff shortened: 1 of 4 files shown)


uday1o1:
This is ready for review. The PR is docs-only. It looks like the metadata check still needs maintainer-owned fields: assignee, type label (`documentation`), and milestone.

PR #2147, opened 2026-05-27
Fix release run lookup for annotated tags

### LINKED ISSUE #2146: [BUG]: release run lookup fails for signed/annotated tags
### Type of Bug
Something else

### Component
Infrastructure

### Describe the bug
The release workflow's auto-detection of the CI run ID breaks for signed release tags.

`.github/RELEASE-core.md` currently tells release managers to create signed tags with `git tag -s ...`. Those are annotated tags. In `ci/tools/lookup-run-id`, the tag is resolved with:

```bash
COMMIT_SHA=$(git rev-parse "${REF}")
```

For an annotated tag, that returns the tag object SHA rather than the underlying commit SHA. We then pass that value to `gh run list --commit`, which is keyed on the commit SHA, so the lookup can fail to find the successful tag-triggered CI run.

This means the documented release path can fail even when the tag-triggered CI run completed successfully.

### How to Reproduce
1. Create a signed or annotated tag for a commit on `main`.
2. Push the tag and wait for the tag-triggered `CI` workflow to finish successfully.
3. Run `ci/tools/lookup-run-id --tag <tag> NVIDIA/cuda-python`, or dispatch `CI: Release` and let it auto-detect the run ID.
4. The lookup may fail to find the successful run for that tag.

### Expected behavior
Signed release tags should work on the documented happy path, and the release workflow should be able to resolve the successful tag-triggered CI run automatically.

### Notes
I think the fix is just to peel the tag before passing it to `gh run list`, for example:

```bash
COMMIT_SHA=$(git rev-list -n 1 "${REF}")
```

or:

```bash
COMMIT_SHA=$(git rev-parse "${REF}^{commit}")
```

This looks release-blocking for `cuda.core`, since the release guide explicitly recommends `git tag -s`.

fallintoplace:
Fixes #2146.

`lookup-run-id` was resolving `--tag` inputs with `git rev-parse <tag>`, which returns the tag object SHA for annotated tags. That works for lightweight tags, but it breaks the documented release flow because `cuda.core` release tags are created with `git tag -s`, so the workflow can end up querying GitHub Actions with the wrong SHA.

This change peels tags to their underlying commit before calling `gh run list`.

Checks:
- `bash -n ci/tools/lookup-run-id`
- `git diff --check upstream/main...HEAD`


ci/tools/tests/test_lookup_run_id.py
leofang, line 1: Please kindly drop this test:
- Files in this folder are not tested anywhere in the CI, files here will bit-rot
- This is disproportionate since `lookup-run-id` is already regularly used (on a per-PR basis)

leofang (changes requested):
Thanks, @fallintoplace, let's keep the one-liner fix.

ci/tools/tests/test_lookup_run_id.py
fallintoplace: Thanks Leo, I will have a look soon.

leofang:
/ok to test [d32f306](https://github.com/NVIDIA/cuda-python/pull/2147/commits/d32f30605aabee8a38df6fa559c63b23c3b0c95f)

leofang (approved):

PR #2150, opened 2026-05-28
Make checkpoint tests fail on missing required binding symbols

### LINKED ISSUE #2149: Checkpoint tests should fail, not skip, when required cuda.bindings symbols are missing
The CUDA checkpoint test availability guard currently masks a class of cuda-bindings regressions as unsupported-environment skips.

This mattered for the CUDA 13.3.0 `CUcheckpointRestoreArgs` regression fixed by PR #2144. In that release, `cuda.bindings.driver.CUcheckpointRestoreArgs` was missing because a CUDA 13.3 header layout change caused generation to silently omit the restore-argument binding. `cuda.core.checkpoint._get_driver()` correctly detects this as a missing required binding attribute and raises:

```text
CUDA checkpointing requires cuda.bindings with CUDA checkpoint API support. Missing: CUcheckpointRestoreArgs
```

However, the checkpoint tests use `_checkpoint_available()` as a broad skip gate. That helper catches any `RuntimeError` from `checkpoint._get_driver()` and returns `False`, so a missing required binding symbol is treated the same as a truly unsupported driver/platform:

- Skip gate: https://github.com/NVIDIA/cuda-python/blob/cc50515648190d866cd4cd0ad269f33b7273513a/cuda_core/tests/test_checkpoint.py#L31-L43
- Checkpoint tests guarded by that skip marker: https://github.com/NVIDIA/cuda-python/blob/cc50515648190d866cd4cd0ad269f33b7273513a/cuda_core/tests/test_checkpoint.py#L397-L423
- Missing required binding detection in `checkpoint._get_driver()`: https://github.com/NVIDIA/cuda-python/blob/cc50515648190d866cd4cd0ad269f33b7273513a/cuda_core/cuda/core/checkpoint.py#L133-L149

As a result, the tests that would have exercised checkpoint restore were skipped instead of failing, even though the installed cuda-bindings version was expected to provide checkpoint API support.

Suggested fix:

- Keep skipping true unsupported environments, such as non-Linux platforms, drivers older than checkpoint API support, or hardware/driver combinations where runtime checkpoint scenarios cannot complete.
- Fail when `checkpoint._get_driver()` reports missing required cuda-bindings attributes for a cuda-bindings version that is expected to support checkpointing.
- Consider adding a lower-level cuda-bindings API completeness test that asserts the required checkpoint symbols are present when the binding version advertises checkpoint support. This would catch omissions before cuda-core runtime tests depend on them.

Related:

- PR #2144 fixes the `CUcheckpointRestoreArgs` generation/layout issue.
- cybind MR 396 is the related generator-side follow-up.
- See also: nvbug 6097301 comment 34.

rwgk:
Closes #2149

## Summary

This PR closes the checkpoint test escape that allowed the CUDA 13.3.0 `CUcheckpointRestoreArgs` binding regression to pass as a skip instead of a failure.

It has three layers, in priority order:

- **Fix the test escape:** `cuda.core` checkpoint tests still skip genuinely unsupported environments, but missing required `cuda.bindings` checkpoint symbols now fail loudly.
- **Add focused `cuda.bindings` coverage:** a narrow checkpoint binding-surface test checks the symbols that `cuda.core.checkpoint` depends on, including `CUcheckpointRestoreArgs`.
- **Keep compact `cuda_core` policy coverage:** the added tests cover the skip/fail boundary directly, while preserving skips for old bindings, unsupported drivers, and CUDA 12.9 environments that lack CUDA 13.x GPU-remapping support.

## Context

This is a follow-up to #2144, which fixed the generated `CUcheckpointRestoreArgs` binding. The separate issue addressed here is that the existing `cuda.core` checkpoint test guard treated every `RuntimeError` from `checkpoint._get_driver()` as an unsupported environment. That included the actual regression signal:

```text
CUDA checkpointing requires cuda.bindings with CUDA checkpoint API support. Missing: CUcheckpointRestoreArgs
```

So the checkpoint tests that knew `CUcheckpointRestoreArgs` was required were skipped instead of failing. That is the test-suite escape tracked in #2149 and described in nvbug 6097301 comment 34.

This PR keeps the intended skips for truly unsupported configurations, but distinguishes them from missing required binding attributes.

## Why The Tests Stay Focused

The `cuda_bindings` test is not intended to become a broad generated-API inventory. It only checks the checkpoint symbols required by `cuda.core.checkpoint`, because this area has already proven more brittle than most of the bindings and the failure mode is otherwise easy to miss.

The `cuda_core` tests were trimmed to a compact availability-policy check. They remain valuable because they protect the exact skip/fail boundary that let this bug escape, without adding broader checkpoint scenario coverage.

## Validation

Before #2144, this PR exposed the original breakage in both relevant places:

```bash
pytest cuda_core/tests/test_checkpoint.py
```

failed during collection with:

```text
RuntimeError: CUDA checkpointing requires cuda.bindings with CUDA checkpoint API support. Missing: CUcheckpointRestoreArgs
```

and:

```bash
pytest cuda_bindings/tests/test_cuda.py::test_cuCheckpoint_required_bindings_present
```

failed with:

```text
missing == ['CUcheckpointRestoreArgs']
```

After rebasing onto #2144 and the trimming commit, local retesting passed:

```text
cuda_pathfinder: 975 passed, 4 skipped
cuda_bindings: 418 passed, 23 skipped
cuda_bindings: 418 passed, 23 skipped
cuda_bindings cython tests: 9 passed
cuda_core: 3379 passed, 205 skipped, 2 xfailed
cuda_core cython tests: 1 passed
```

## Related

- Follow-up to #2144.
- Related backport: #2145.
- Related generator-side fix: cybind MR 396.
- Related generator-side cleanup/alignment: cybind MR 397.
- See also: nvbug 6097301 comment 34.


rwgk:
# PR 2150 first CI failure analysis

Workflow: https://github.com/NVIDIA/cuda-python/actions/runs/26591678170

Commit: 293258d9d60620c091ece04a7a89c43ba52d24e6

Workflow result: failed.

## High-level result

The build and non-test infrastructure mostly passed:

- Build jobs passed.
- Docs passed.
- `pre-commit.ci` passed.
- The final `Check job status` job failed because matrix test jobs failed.

The failures are concentrated in test matrix jobs. There were 37 failed test jobs plus the final status aggregation job.

Failure counts by CUDA version:

- CUDA 13.3.0: 24 failed test jobs.
- CUDA 12.9.1: 13 failed test jobs.
- CUDA 13.0.2: no failures observed in the failed-job list.

Failure counts by platform:

- `linux-64`: 17 failed test jobs.
- `linux-aarch64`: 12 failed test jobs.
- `win-64`: 8 failed test jobs.

## Failure mode 1: CUDA 13.3 missing `CUcheckpointRestoreArgs`

This is the expected proof-of-coverage failure for the issue fixed by PR #2144.

Representative failed jobs:

- https://github.com/NVIDIA/cuda-python/actions/runs/26591678170/job/78353772395
- https://github.com/NVIDIA/cuda-python/actions/runs/26591678170/job/78354692898

Observed failure:

```text
tests/test_cuda.py::test_cuCheckpoint_required_bindings_present FAILED
E       AssertionError: assert ['CUcheckpointRestoreArgs'] == []
E         Left contains one more item: 'CUcheckpointRestoreArgs'
```

This shows the new `cuda_bindings/tests/test_cuda.py::test_cuCheckpoint_required_bindings_present` test catches the missing generated binding directly. This is exactly the failure PR #2150 was intended to expose before PR #2144 is merged/rebased in.

## Failure mode 2: Linux CUDA 12.9 missing `CUcheckpointGpuPair`

This is a separate checkpoint binding gap surfaced by tightening the `cuda.core` checkpoint availability guard.

Representative failed jobs:

- https://github.com/NVIDIA/cuda-python/actions/runs/26591678170/job/78353772318
- https://github.com/NVIDIA/cuda-python/actions/runs/26591678170/job/78354682523

Observed failure:

```text
ERROR collecting tests/test_checkpoint.py
RuntimeError: CUDA checkpointing requires cuda.bindings with CUDA checkpoint API support. Missing: CUcheckpointGpuPair
binding_ver = (12, 9, 7)
missing    = ['CUcheckpointGpuPair']
```

This happens during `cuda_core/tests/test_checkpoint.py` collection on Linux. Windows does not hit this path because the checkpoint tests are platform-skipped before `_checkpoint_available()` is evaluated.

This is not the CUDA 13.3 `CUcheckpointRestoreArgs` issue. It means the tightened guard also exposes that CUDA 12.9 Linux bindings report enough checkpoint API surface to reach `cuda.core.checkpoint._get_driver()`, but still lack `CUcheckpointGpuPair`, which `cuda.core` considers required.

## Interpretation

CI behaved as intended for the main goal: PR #2150 converts the previous skip-masked CUDA 13.3 binding regression into clear test failures.

The run also identifies a follow-up decision for CUDA 12.9:

- If CUDA 12.9 should support the `cuda.core.checkpoint` mapping helpers, then `CUcheckpointGpuPair` needs to be present in the CUDA 12.9 bindings.
- If CUDA 12.9 should not support that surface, the `cuda.core` availability check needs to classify this specific older-binding condition as skippable rather than as a binding regression.

After PR #2144 is merged and PR #2150 is rebased onto it, the CUDA 13.3 `CUcheckpointRestoreArgs` failures should disappear. The CUDA 12.9 `CUcheckpointGpuPair` failures may remain unless they are handled separately.

rwgk:
I looked into the CUDA 12.9 failures from the first PR #2150 CI run.

The short version: these failures look separate from the CUDA 13.3 `CUcheckpointRestoreArgs` regression that PR #2144 fixes.

In `/usr/local/cuda-12.9`, I do not see `CUcheckpointGpuPair` at all. Even

```bash
grep -r -i GpuPair /usr/local/cuda-12.9
```

returns no matches.

The CUDA 12.9 `cuda.h` checkpoint restore args are still the older reserved-only layout:

```c
typedef struct CUcheckpointRestoreArgs_st {
    cuuint64_t reserved[8]; /**< Reserved for future use, must be zeroed */
} CUcheckpointRestoreArgs;
```

That matches the CUDA 12.9 CI failure mode from https://github.com/NVIDIA/cuda-python/actions/runs/26591678170: Linux CUDA 12.9 jobs now fail during `cuda_core/tests/test_checkpoint.py` collection with:

```text
RuntimeError: CUDA checkpointing requires cuda.bindings with CUDA checkpoint API support. Missing: CUcheckpointGpuPair
binding_ver = (12, 9, 7)
missing    = ['CUcheckpointGpuPair']
```

So my current interpretation is:

- The CUDA 13.3 failures are expected and useful: PR #2150 proves that missing `CUcheckpointRestoreArgs` would now fail loudly instead of being skip-masked.
- The CUDA 12.9 failures are a separate compatibility issue surfaced by the tighter guard.
- Since `CUcheckpointGpuPair` does not appear to exist in the CUDA 12.9 headers, this is not a missing 12.9 Python binding. It is more likely that `cuda.core.checkpoint` is treating `CUcheckpointGpuPair` as required too broadly for CUDA 12.9.

Possible follow-up direction: keep missing required symbols as failures for APIs that should exist in the active CUDA version, but treat the CUDA 12.9/no-`CUcheckpointGpuPair` path as an older checkpoint API shape that should remain skippable or should avoid enabling the GPU remapping surface.

rwgk:
# PR 2150 second CI failure analysis

Workflow: https://github.com/NVIDIA/cuda-python/actions/runs/26596635176

Commit: cd730c11b9162423df5e2d8748242fd6075ef60d

Current workflow state at inspection time:

- 65 jobs passed.
- 23 jobs failed.
- 1 job was still queued: `Test win-64 / Python 3.14, CUDA 13.3.0 (wheels), GPU h100 (x2) (MCDM)`.

## High-level result

The second CI run matches expectations after splitting baseline checkpoint support from GPU-remapping support.

All completed failures are CUDA 13.3.0 test jobs. CUDA 12.9.1 and CUDA 13.0.2 jobs that completed are passing.

Failure counts by CUDA version:

- CUDA 13.3.0: 23 failed test jobs.
- CUDA 12.9.1: 0 failed test jobs observed.
- CUDA 13.0.2: 0 failed test jobs observed.

Failure counts by platform:

- `linux-64`: 10 failed test jobs, all CUDA 13.3.0.
- `linux-aarch64`: 6 failed test jobs, all CUDA 13.3.0.
- `win-64`: 7 failed test jobs, all CUDA 13.3.0.

## Remaining failure mode: CUDA 13.3 missing `CUcheckpointRestoreArgs`

Representative failed jobs:

- https://github.com/NVIDIA/cuda-python/actions/runs/26596635176/job/78371061954
- https://github.com/NVIDIA/cuda-python/actions/runs/26596635176/job/78372022514

Both sampled logs show the same intended failure:

```text
tests/test_cuda.py::test_cuCheckpoint_required_bindings_present FAILED
E       AssertionError: assert ['CUcheckpointRestoreArgs'] == []
E         Left contains one more item: 'CUcheckpointRestoreArgs'
missing    = ['CUcheckpointRestoreArgs']
```

This is the expected pre-PR-2144 signal. PR #2150 is still proving that the CUDA 13.3 `CUcheckpointRestoreArgs` binding regression would fail loudly instead of being skip-masked.

## Fixed compared with the first CI run

The CUDA 12.9 `CUcheckpointGpuPair` failure mode is gone.

In the first run, Linux CUDA 12.9 jobs failed during `cuda_core/tests/test_checkpoint.py` collection with:

```text
RuntimeError: CUDA checkpointing requires cuda.bindings with CUDA checkpoint API support. Missing: CUcheckpointGpuPair
binding_ver = (12, 9, 7)
missing    = ['CUcheckpointGpuPair']
```

After commit cd730c11b9162423df5e2d8748242fd6075ef60d, completed CUDA 12.9.1 jobs are passing. This confirms that baseline checkpoint tests no longer require the CUDA 13.x GPU-remapping type, while the CUDA 13.3 missing-`CUcheckpointRestoreArgs` regression remains a hard failure.

## Interpretation

The results are the desired intermediate state:

- PR 2150 no longer over-requires `CUcheckpointGpuPair` for CUDA 12.9.
- PR 2150 still catches the original CUDA 13.3 `CUcheckpointRestoreArgs` generation issue.
- After PR 2144 lands and PR 2150 is rebased onto it, the CUDA 13.3 failures should disappear.

leofang (commented):
Do we really need this PR as-is? It seems pretty AI-slop to me... If binding is broken, we just fix it and move on. Same if it's the codegen that's broken. I don't think adding tests to either cuda-bindings/cuda-core like this is maintainable.

rwgk:
Thanks Leo, fair concern on maintainability.

This was AI-assisted, but I manually guided and reviewed it. The intent is not to build a broad ad hoc API coverage framework in this PR. The new cuda_bindings test is a deliberately narrow regression guard for the exact checkpoint binding surface that silently disappeared in 13.3.0. Keith also raised the broader version of this in Slack: we should have something that enumerates all public APIs and types per CUDA version so parser failures cannot silently drop symbols. I agree with that, but it is a much larger ask.

For cuda_core, the goal is also narrower than adding new checkpoint scenario coverage. The existing checkpoint tests already covered the restore path, but the availability helper treated missing required bindings as an unsupported-environment skip. This PR fixes that boundary: unsupported drivers/old bindings still skip, but missing required symbols now fail. The added cuda_core tests are focused on that skip/fail behavior and on separating baseline checkpoint support from GPU remapping support.

leofang:
Copying my reply here for posterity.

In general, I find agents very easily introduce code bloating that would otherwise not happen in the old days.
- for cuda-bindings:  
> we should have something that enumerates all public APIs and types per CUDA version so parser failures cannot silently drop symbols.

A big part of work that Mike and I are working on will make this moot. This issue only happens with cython-gen (aka driver/runtime/~~nvrtc~~) due to the way tempita+pyclibrary is used. We’re in the middle of removing it: https://github.com/NVIDIA/cuda-python/pull/1972.

- for cuda-core: can’t we just detect if cuda-bindings 13.3.0 is in use, and raise an exception if checkpointing is in use? Our typical check pattern is: 
   - is driver version sufficient?
   - does the needed binding exist?

this is really an edge case that should just be fixed in a one-off fashion. The codegen migration would be the eventual fix. The agents do not know the bigger picture that the team is seeing, so we need to help guide them and define a proper scope.

rwgk:
> Copying my reply here for posterity.

My concern is not the 13.3.0 binding bug itself; #2144 fixes that. My concern is the separate test-suite escape: an existing checkpoint availability guard converted a missing required binding into an unsupported-environment skip.

That was not hypothetical. The missing `CUcheckpointRestoreArgs` binding was already detected by `cuda.core.checkpoint._get_driver()`, but the test guard swallowed that `RuntimeError` and skipped the checkpoint tests. Without that error-masking skip, this particular regression would have failed very early on in local testing, immediately after running cybind.

I agree that removing cython-gen should reduce the risk of parser-driven symbol drops in the future. But it does not address this test behavior: when an API is expected to exist, missing required symbols should fail tests, not be classified as an unsupported environment.

That is the narrow scope I was trying to cover:

- In `cuda_bindings`, add a focused completeness check for the checkpoint binding surface that `cuda.core` depends on.
- In `cuda_core`, keep the intended skips for unsupported drivers / old bindings, but let missing required symbols fail.
- Split baseline checkpoint support from GPU remapping support so CUDA 12.9 remains properly exercised where `CUcheckpointGpuPair` does not exist.

If there is a smaller maintainable version of this, I am happy to reshape the PR. What I do not think is sufficient is only adding a one-off runtime check for `cuda-bindings==13.3.0`: that helps after a broken release exists, but it does not close the test escape that let this regression get released.

rwgk:
I made a small trimming pass in commit 41187f8ecf.

The change consolidates the three `cuda_core` checkpoint availability-policy tests into one parametrized test. This keeps the same behavioral coverage, but reduces the added test surface and makes the intended policy easier to see:

- unsupported installed driver: still skip
- old `cuda.bindings` without checkpoint API support: still skip
- supported `cuda.bindings` version with a missing required checkpoint symbol: fail loudly

I did not trim further because those distinctions are the core reason for this PR. The 13.3.0 `CUcheckpointRestoreArgs` issue did not escape because the broad test suite was absent; it escaped because the checkpoint test guard treated a missing required binding exactly like an unsupported environment and skipped. A one-off check for the already-broken `cuda-bindings==13.3.0` release would not close that test escape for future required-symbol drops.

I also kept the `cuda_bindings` presence test. It is intentionally narrow: it checks the checkpoint binding surface that `cuda.core.checkpoint` depends on, not a broad generated-API inventory. That still seems worth keeping because this particular area has already proven brittle, and the test is cheap, direct, and gives a clearer failure than waiting for `cuda_core` to trip over a missing symbol.

The remaining `cuda_core` GPU-remapping availability test is separate on purpose. CUDA 12.9 has baseline checkpoint support but does not define `CUcheckpointGpuPair`, while CUDA 13.x adds the GPU-remapping restore layout. Keeping that split prevents us from over-requiring 13.x-only remapping support when the baseline checkpoint lifecycle should still be tested.

__
Cursor GPT-5.5 1M Extra High on behalf of Ralf

PR #2156, opened 2026-05-29
Fix cuda.core and cuda.bindings example links in docs

uday1o1:
## Summary

- Replace raw `cuda.core` and `cuda.bindings` GitHub example URLs that used `|..._github_ref|` substitutions with Sphinx `extlinks` roles.
- Update affected links in `cuda_core/docs/source` and `cuda_bindings/docs/source`.
- Preserve version-aware behavior so latest/preview docs link to `main`, while release docs link to the appropriate release tag/ref.

## Root Cause

The GitHub ref replacements were defined through `rst_epilog`, but they were used inside embedded hyperlink URI targets. Sphinx/Docutils left those substitutions literal in generated HTML, so browsers/GitHub saw URLs containing values like `|cuda_core_github_ref|` or `|cuda_bindings_github_ref|`. Those are treated as branch/tag names and the links return 404.

## Validation

- Crawled 250 live `latest` docs pages and found the affected pages were limited to `cuda.core` and `cuda.bindings` docs pages touched by this PR.
- Built Sphinx smoke docs for:
  - `cuda.core` latest
  - `cuda.core` release `1.0.1`
  - `cuda.bindings` latest
  - `cuda.bindings` release `13.3.0`
- Checked generated GitHub example links from all four smoke builds; all returned HTTP 200.
- Confirmed generated HTML no longer contains `cuda_core_github_ref`, `cuda_bindings_github_ref`, or encoded `%7Ccuda_...` refs.
- Ran `git diff --check`.

### DIFF
--- a/cuda_bindings/docs/source/conf.py
+++ b/cuda_bindings/docs/source/conf.py
@@ -44,6 +44,7 @@ GITHUB_EXAMPLES_REF = _github_examples_ref()
 extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",
+    "sphinx.ext.extlinks",
     "sphinx.ext.napoleon",
     "sphinx.ext.intersphinx",
     "myst_nb",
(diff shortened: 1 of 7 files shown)


kkraus14 (approved):

kkraus14:
/ok to test https://github.com/NVIDIA/cuda-python/pull/2156/commits/9a93adf273850ab849cfec6af695b1921ccff4dd

Issue #2114, opened 2026-05-19
Use `pytest-run-parallel` to test thread safety


leofang:
It should work for both FT & non-FT builds.
- https://github.com/Quansight-Labs/pytest-run-parallel
- https://py-free-threading.github.io/testing/

kkraus14:
I suspect this would need a non-trivial amount of test refactoring to avoid teardowns that interact with the CUDA driver in a stateful way? There's so much state in the CUDA driver that it's hard to imagine doing this in an automated way safely.

leofang:
Maybe, maybe not. @seberg has invested a lot of efforts in CuPy and we're generally happy with us gaining more thread safety, following community standard, and having a CUDA project reaching FT-readiness. I am hopeful.

seberg:
Yeah, I would propose to try it, after the CuPy release maybe I can have a look as well.  CuPy had fun caches, the memory manager, and an awful lot of legacy unittest-style tests that were painful...  I would think this is all much easier for cuda-python.

I am not sure what might be problematic, e.g. graph capture related tests, seems like one that one may need to just mark as thread-unsafe in a first iteration.  (In CuPy they are largely fine, but not fully e.g. because module loading/unloading in multiple threads can be problematic.)

EDIT: I started to have a bit of a look.

> I suspect this would need a non-trivial amount of test refactoring to avoid teardowns that interact with the CUDA driver in a stateful way?

Yeah, I now see what you mean.  `cuda-pathfinder` tests seem fine with a few small thread-unsafe markers.  But `cuda-bindings`/`cuda-core` would skip almost all tests without larger re-factor since it's fixtures need to run per-thread (e.g. due to `cuInit()`).
It may work to convert these to normal decorators, but without that the `pytest-run-parallel` run wouldn't add much.

seberg:
This is implemented now running on the FT, CI jobs. A follow-up here may be to see if we can run this with TSAN to find more things and also to make it easier for downstream to use TSAN.

If we worry about missing `with nogil:` leading to possible deadlocks, then it could also make sense to expand it to a non-FT job.

PR #2162, opened 2026-06-02
Fix threading issues in graphnode cleanup and SMResource splitting

seberg:
This fixes a few threading issues, but we may want to discuss some details still.
* The GraphNode cleanup order is an important fix: Another thread may end up with the same pointer (but new object) as soon as we clean it up.  So we have to remove it from the cache before cleaning it up.
* The split mutex: This is thread-unsafe and failed threaded testing.  But I am honestly not sure if that isn't just expected, or whether the mutex is good but it should also be safe from within CUDA?

The following was split out:

<details>

* Use of atomics: I think this is needed, but for this one place an atomic seemed more reasonable.  (However, hard to test and if it can fail IIUC only on ARM.)
  (A critical section would work as well, since it is a lock which enforces order.) 
* The critical sections should be pretty safe.  I am not sure they will all ensure that the object is always the _identity_ but I am pretty sure it protects from worse races.
  `pytest-run-parallel` did find this for `MemPool.attributes` and I could not reproduce after adding it.  The other critical sections are added with a code-search, though.

* Use of `setdefault` cached pattern is largely just normalizing.  Without the `return dict.setdefault` a different instance may be returned on different threads (or a cache entry replaced). For the `cyGraphMemoryResource` that triggered a test with pytest-run-parallel although that doesn't mean it is problematic as such. `cuda-pathfinder` uses functools.cache, but usually for strings; the one we may want to look at is `load_nvidia_dynamic_lib`.

</details>

---

## Checklist

Implementing gh-2114 would add implicit coverage for most but the atomics and some of the critical sections (if it adds it might be so rare to effectively never trigger).  So this is split out.
(If preferred I can add some explicitly threaded tests.)

I have no full runs with pytest-run-parallel, but I want to clean up the test harness a bit more.  I also still should test without free-threading (useful to flush out forgotten `with nogil:` blocks that can cause deadlocks).

### DIFF
--- a/cuda_core/cuda/core/_device_resources.pyx
+++ b/cuda_core/cuda/core/_device_resources.pyx
@@ -498,10 +498,11 @@ cdef class SMResource:
         )
         _resolve_group_count(opts)
         _check_green_ctx_support()
-        if _can_use_structured_sm_split():
-            return _split_with_general_api(self, opts, dry_run)
-        # SplitByCount requires the same 12.4+ as green ctx support (already checked above)
-        return _split_with_count_api(self, opts, dry_run)
+        with self._split_mutex:
+            if _can_use_structured_sm_split():
+                return _split_with_general_api(self, opts, dry_run)
+            # SplitByCount requires the same 12.4+ as green ctx support (already checked above)
+            return _split_with_count_api(self, opts, dry_run)
 
 
 cdef class WorkqueueResource:
--- a/cuda_core/cuda/core/_device_resources.pxd
+++ b/cuda_core/cuda/core/_device_resources.pxd
@@ -2,6 +2,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+cimport cython
+
 from cuda.bindings cimport cydriver
 from cuda.core._resource_handles cimport ContextHandle, GreenCtxHandle
 
@@ -15,6 +17,7 @@ cdef class SMResource:
         unsigned int _flags
         bint _is_usable
         object __weakref__
+        cython.pymutex _split_mutex
 
     @staticmethod
     cdef SMResource _from_dev_resource(cydriver.CUdevResource res, int device_id)
--- a/cuda_core/cuda/core/graph/_graph_node.pyx
+++ b/cuda_core/cuda/core/graph/_graph_node.pyx
@@ -161,10 +161,11 @@ cdef class GraphNode:
         cdef cydriver.CUgraphNode node = as_cu(self._h_node)
         if node == NULL:
             return
-        with nogil:
-            HANDLE_RETURN(cydriver.cuGraphDestroyNode(node))
+
         _node_registry.pop(<uintptr_t>self._h_node.get(), None)
         invalidate_graph_node(self._h_node)
+        with nogil:
+            HANDLE_RETURN(cydriver.cuGraphDestroyNode(node))
 
     @property
     def pred(self) -> AdjacencySetProxy:


cuda_core/cuda/core/_device_resources.pyx
```diff
@@ -493,10 +493,11 @@ cdef class SMResource:
...
         _check_green_ctx_support()
-        if _can_use_structured_sm_split():
-            return _split_with_general_api(self, opts, dry_run)
-        # SplitByCount requires the same 12.4+ as green ctx support (already checked above)
-        return _split_with_count_api(self, opts, dry_run)
+        with self._split_mutex:
+            if _can_use_structured_sm_split():
+                return _split_with_general_api(self, opts, dry_run)
+            # SplitByCount requires the same 12.4+ as green ctx support (already checked above)
+            return _split_with_count_api(self, opts, dry_run)
```
seberg, line 505: Pointing this out inline, because this is probably the thing I am least sure about.  Should `SMResource.split()` be thread-safe if called on the same resource from multiple threads?

mdboom:
This should probably be a separate PR, but are there best practices for CI to help ensure these sorts of issues don't come back?  Should we be running `pytest-run-parallel` on our `3.14t` configurations, for example?

seberg:
> Should we be running pytest-run-parallel on our 3.14t configurations, for example?

Yeah, I just thought I'd split out this PR first and get it in and then open a PR for that.  Running `pytest-run-parallel` is +maybe not a huge PR but also a non-trivial PR because we probably want a "mini pytest plugin" for `init_cuda` style logic.

(*ideally* we use `pytest-run-parallel` together with a TSAN environment, I haven't actually done TSAN yet locally.  Since we have ASAN jobs, I think that may well be very doable.)

rparolin:
It would be great if at least the critical section and atomics changes were broken out into a separate PR.

seberg:
OK, I split out most things and will update after that (or split out the final chunks), until then marking as draft.

What is remaining is the GraphNode cleanup order and the mutex for `SMResource.split()`.

Andy-Jost (commented):
Just commenting (not approving) because this PR is marked draft (I think it will be split).

I'm fine with these fixes as they are, though I'd also like to explore more structural fixes to decrease the likelihood problems are reintroduced. None of that blocks this change for me. Thematically, I think it amounts to moving more things into the C++ layer. Some ideas below from my bot (where Level 1 and Level 2 refer to the registry levels documented in `REGISTRY_DESIGN.md`):

There are more structural options, and they're largely aligned with where the handle layer was already heading. This PR is tactical patching; the strategic fix is centralize lifetime + identity + lazy state in C++ boxes/factories, and make Cython thin.

1. Push lazy-init into Boxes (replaces atomics/critical sections for CUDA metadata)
  A lot of what got `@cython.critical`_section / atomics is really "query driver once per handle": Buffer mem attrs, pool attrs, kernel attrs. That belongs in the box next to the resource, with `std::once_flag` or a box mutex:

```
  struct DevicePtrBox {
      CUdeviceptr resource;
      mutable MemAttrs attrs;
      mutable std::once_flag attrs_init;
  };
```

  Cython properties become `return box->get_attrs()` — no Python-side check-then-act. This matches the existing handle design: state lives with ownership, not with the Python wrapper.

2. Unified C++ factory + destroy protocol (fixes L1 ordering in one place)  `create_graph_node_handle` already owns Level 1. Extend it to own the full lifecycle:
```
  create_graph_node(CUgraphNode, GraphHandle) → GraphNodeHandle + Python wrapper
  destroy_graph_node(GraphNodeHandle) → { pop L2, unregister L1, null resource, cuGraphDestroyNode }
```

  All under one C++ mutex. Then Cython `GraphNode.destroy()` is a one-liner. The invalidate-then-destroy rule becomes impossible to get wrong at call sites.

3. Templatized Level-2 / as_py identity cache in C++
  REGISTRY_DESIGN.md already cites pybind11's registered_instances. A generic:

```
  template<typename Handle>
  class PyIdentityRegistry {
      // key: owner identity (owner_before / box pointer)
      // value: PyObject* via weakref
  };
```

  could serve GraphNode, Kernel, Event, etc. Key on handle ownership, not raw CU*. Registration happens in the factory; unregistration in the destroy protocol. Cython stops maintaining `_node_registry` by hand.

as_py(h) in DESIGN.md was meant to return Python wrappers — today GraphNode's as_py still goes to cuda.bindings. Evolving it to return the cached cuda.core object would collapse L2 into the handle layer.

4. What you still can't fully move out of Cython
  • Polymorphic construction: GN_create_impl dispatches to EmptyNode, KernelNode, … C++ either needs a typed factory table or a Cython callback registered at init — still one entry point, but not zero Cython.
  • Pure-Python lazy caches: IPCDataForBuffer, DLPack/StridedMemoryView inference, ComputeCapability tuples — these aren't handle metadata; they'll need Python locks or redesign.
  • APIs that aren't handle-backed: SMResource.split() — mutex or a documented "caller serializes" contract unless the driver guarantees more.

seberg:
Not sure I am following all of it, but C++ threading does seem much more mature than Python/Cython so pushing more to C++ is probably a good idea.
Some notes/thoughts:
* In most cases here the `critical_section` is on functions that return Python objects (one where it wasn't uses the `atomic`, I forgot to delete it), so there are two considertions:
  1. Creating the object is quick but not free performance wise.
  2. It may be nice if things are idempotent which a C++ side re-wrapping can't solve?
     (But: I don't think the `@critical_section` doesn't guarantee idempotence fully.)
  I am slightly wondering if we couldn't create a tiny pattern ourselves for this but not sure how nice it'll be my best brainstorming is to create a C++ helper (mirroring dicts):
     ```
     cdef atomic_object:
         cdef get(..., order=something):
         cdef set(..., order=something):
         cdef object setdefault(..., order=something):
     ```
     but if we go down that route it may make more sense to talk to Cython folks.  I think Cython support will improve.
* Moving the `as_py` adds the need to deal with cleaning up weakrefs, but I suspect it is a good idea even just to not split the logic.

seberg:
/ok to test [ec725b4](https://github.com/NVIDIA/cuda-python/commit/ec725b4f274c6b955cbdabe8ab673683c68bc0df)

seberg:
@Andy-Jost since you had looked at this before.  This is now hopefully very simple with just two changes:
* Re-order graph cleanup for thread-safety.
* Guard `SMResource.split()` (this one makes sense to me, but I am not sure it should be fixed here strictly or is maybe expected to be thread unsafe)

Andy-Jost (approved):
LGTM

PR #2176, opened 2026-06-07
CI: allow specifying custom driver versions in test matrix

### LINKED ISSUE #293: CI: Expand test matrix to cover more driver versions on Linux
CUDA Python heavily relies on the UMD to provide functionalities. We need to include driver version as a new dimension of our CI test matrix. Need to discuss with @ajschmidt8 @m3vaz on how to approach this (as I don't think this is doable today).

### LINKED ISSUE #1265: CI: Test different driver versions on Windows
After #1242 is merged, we have a nice test matrix for different Windows configurations. However, currently they lock at the same driver version `latest`:
https://github.com/NVIDIA/cuda-python/blob/c4079ddfb6b6dcdd7c9773b727530f122021fa9e/ci/test-matrix.json#L37-L48
But looking at this more closely, one would find that this `DRIVER` label is not used on Windows at all. Instead, the version is hard-wired in the installer script:
https://github.com/NVIDIA/cuda-python/blob/c4079ddfb6b6dcdd7c9773b727530f122021fa9e/ci/tools/install_gpu_driver.ps1#L8-L10

The reason we need the `DRIVER` label on Linux is because the driver is pre-installed in the runner VMs (and maintained by the runner team), and we need to use the label to compute the runner name, whereas on Windows due to technical challenges we need to install the driver ourselves as part of the CI jobs. 

But, it gives us a unique opportunity to do something that we cannot do on Linux runners today, which is to select the driver versions that we intend to cover.

I think the `DRIVER` label on Windows could be repurposed to specify the UMD version, with the test matrix expanded:
| |CTK version | UMD version | purpose
|---|---|---|---|
| prev major | 12.x | 12.0 | test CUDA minor version compatibility |
| prev major | 12.x | 13.x | test CUDA backward compatibility |
| curr major | 13.0 | 13.0 | |
| curr major | 13.0 | 13.x | test CUDA backward compatibility |
| curr major | 13.x | 13.0 | test CUDA minor version compatibility |
| curr major | 13.x | 13.x | |

and we find a way to map the UMD version (ex: 13.0) to KMD version (ex: 581.15). Currently there is no public way to do this mapping and we need to hard-code a small table.

Perhaps this can be added to the nightly runs? #294

leofang:
## Description

closes #293
closes #1265

Extends the `DRIVER` field in `ci/test-matrix.yml` to accept an explicit driver version string (e.g. `580.65.06`) in addition to the existing `latest` / `earliest`.

- **Linux** — new `ci/tools/install_gpu_driver.sh` swaps the driver in-job. It is adapted nearly verbatim from the runner team's `nvgha-driver` CLI (copied so we don't depend on its rollout schedule). The script `nsenter`s onto the host for the install and refreshes the toolkit bind mounts back inside the test container. Because our script lives in the GH workspace mount (container-only), the host-side re-exec reads the script from stdin via `bash -s < "$0"` rather than running `"$0"` directly (the relative path doesn't resolve after the mount-namespace flip).
- **Windows** — `ci/tools/install_gpu_driver.ps1` is split into two scripts: `install_gpu_driver.ps1` (now install-only, reads `DRIVER` from env, errors on `latest`/`earliest`) and a new `ci/tools/configure_driver_mode.ps1` (driver-mode + `pnputil` device cycle, runs on every job). This also fixes a long-standing wart: the previous script unconditionally installed a hardcoded `581.15` even when the matrix row used a `latest`/`earliest` runner that already carried the right driver.

Matrix wiring (in both `test-wheel-linux.yml` and `test-wheel-windows.yml`):

- `compute-matrix` adds a new `RUNNER_DRIVER` field per row — equal to `DRIVER` for `latest` / `earliest`, otherwise `latest`. `runs-on:` is keyed on `RUNNER_DRIVER` so custom-DRIVER rows land on the most recent pre-installed runner image (the install scripts perform the actual swap).
- On Linux, `container.options` only adds `--privileged --pid=host` for custom-DRIVER rows (required by the nsenter dance).
- On Linux, custom `DRIVER` combined with `FLAVOR=wsl` is rejected eagerly in `compute-matrix` — the in-container swap doesn't work under WSL.
- The "Ensure GPU is working" step (`nvidia-smi`) now runs *after* the install / configure step in every workflow, so it validates the post-install driver state on custom-DRIVER rows.
- `coverage.yml` (Windows path, hardcoded `DRIVER: latest`) was updated alongside since it was the other caller of the old combined script.

Matrix rows flipped to exercise the new code path:

- **PR matrix (Linux, every PR):**
  - `amd64 / 3.13 / 13.3.0 / local-CTK / rtxpro6000` → `DRIVER: '610.43.02'`
  - `amd64 / 3.14 / 13.3.0 / local-CTK / l4` → `DRIVER: '610.43.02'`
- **Nightly numba-cuda:**
  - Linux `amd64 / 3.12 / 13.3.0 / l4` → `DRIVER: '580.65.06'`
  - Windows `amd64 / 3.12 / 13.3.0 / l4` → `DRIVER: '596.36'`

Also enables `workflow_dispatch:` on `ci.yml` so the main CI pipeline can be re-run manually from the Actions UI (no inputs — the workflow already builds every wheel it tests, and the existing `should-skip` / `detect-changes` gates handle non-PR events correctly).

All other matrix rows continue to use `DRIVER: latest` and are unaffected (same runners, no install step, no privileged container).

## Checklist


leofang:
/ok to test [b1b6070](https://github.com/NVIDIA/cuda-python/pull/2176/commits/b1b6070a823898a01e87c49cb1682f6d65a96eb9)

leofang:
/ok to test [3e016b5](https://github.com/NVIDIA/cuda-python/pull/2176/commits/3e016b572dcded7701f9b2f12c25cec3cb7e5b1d)

ci/test-matrix.yml
```diff
@@ -113,4 +122,4 @@ windows:
     - { MODE: 'nightly-pytorch',    ARCH: 'amd64', PY_VER: '3.12', CUDA_VER: '13.0.2', LOCAL_CTK: '0', GPU: 'l4', GPU_COUNT: '1', DRIVER: 'latest', DRIVER_MODE: 'TCC', TORCH_VER: '2.9.1',  TORCH_CUDA: 'cu130' }
     # nightly-numba-cuda
     - { MODE: 'nightly-numba-cuda', ARCH: 'amd64', PY_VER: '3.12', CUDA_VER: '12.9.1', LOCAL_CTK: '0', GPU: 'l4', GPU_COUNT: '1', DRIVER: 'latest', DRIVER_MODE: 'TCC' }
-    - { MODE: 'nightly-numba-cuda', ARCH: 'amd64', PY_VER: '3.12', CUDA_VER: '13.3.0', LOCAL_CTK: '0', GPU: 'l4', GPU_COUNT: '1', DRIVER: 'latest', DRIVER_MODE: 'TCC' }
+    - { MODE: 'nightly-numba-cuda', ARCH: 'amd64', PY_VER: '3.12', CUDA_VER: '13.3.0', LOCAL_CTK: '0', GPU: 'l4', GPU_COUNT: '1', DRIVER: '596.36',  DRIVER_MODE: 'TCC' }
```
leofang, line 128: xref: https://github.com/NVIDIA/cuda-python/actions/runs/27081260952/job/79927472454

ci/test-matrix.yml
```diff
@@ -74,7 +83,7 @@ linux:
     - { MODE: 'nightly-pytorch',    ARCH: 'arm64', PY_VER: '3.12', CUDA_VER: '13.0.2', LOCAL_CTK: '0', GPU: 'l4', GPU_COUNT: '1', DRIVER: 'latest', TORCH_VER: '2.9.1',  TORCH_CUDA: 'cu130' }
     # nightly-numba-cuda
     - { MODE: 'nightly-numba-cuda', ARCH: 'amd64', PY_VER: '3.12', CUDA_VER: '12.9.1', LOCAL_CTK: '0', GPU: 'l4', GPU_COUNT: '1', DRIVER: 'latest' }
-    - { MODE: 'nightly-numba-cuda', ARCH: 'amd64', PY_VER: '3.12', CUDA_VER: '13.3.0', LOCAL_CTK: '0', GPU: 'l4', GPU_COUNT: '1', DRIVER: 'latest' }
+    - { MODE: 'nightly-numba-cuda', ARCH: 'amd64', PY_VER: '3.12', CUDA_VER: '13.3.0', LOCAL_CTK: '0', GPU: 'l4', GPU_COUNT: '1', DRIVER: '580.65.06' }
```
leofang, line 89: xref https://github.com/NVIDIA/cuda-python/actions/runs/27081260952/job/79927472925

leofang:
/ok to test [4a23b23](https://github.com/NVIDIA/cuda-python/pull/2176/commits/4a23b23a26111a8ead0cc40519a2b558ea9bfe66)

ci/tools/configure_driver_mode.ps1
```diff
@@ -0,0 +1,45 @@
...
+        exit 1
+    }
+
+    # Only restart NVIDIA display adapters, not other display devices (e.g. QEMU VGA)
+    $nvidia_devices = Get-PnpDevice -Class Display -FriendlyName "NVIDIA*"
+    foreach ($device in $nvidia_devices) {
+        Write-Output "Restarting device: $($device.FriendlyName) ($($device.InstanceId))"
+        pnputil /disable-device "$($device.InstanceId)"
+        pnputil /enable-device "$($device.InstanceId)"
+    }
```
rwgk, line 39: It seems this function now cycles display devices on every Windows job, even when no driver was installed and `nvidia-smi -fdm` reports the mode was already correct (e.g. `Driver model is already set to MCDM for GPU 00000000:0A:00.0.` in the log). Is that intentional? In the [h100/MCDM CI failure](https://github.com/NVIDIA/cuda-python/actions/runs/27081591837/job/79929244255), NVML becomes unavailable right after the `pnputil` cycle here, so I’m wondering if the device restart should only happen after an install or an actual mode change.

leofang: This is now fixed in commit d33a928eb464987b6de6eeceaf7ab9454ffe1e91 by relaxing the timeout for the mode change to settle.

Even before this PR, the mode change already must happen unconditionally, because the default driver mode may not be aligned with what we want to test against.

ci/test-matrix.yml
```diff
@@ -29,10 +38,10 @@ linux:
     - { ARCH: 'amd64', PY_VER: '3.12',  CUDA_VER: '13.3.0', LOCAL_CTK: '0', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest' }
     - { ARCH: 'amd64', PY_VER: '3.13',  CUDA_VER: '12.9.1', LOCAL_CTK: '0', GPU: 'v100',       GPU_COUNT: '1', DRIVER: 'latest' }
     - { ARCH: 'amd64', PY_VER: '3.13',  CUDA_VER: '13.0.2', LOCAL_CTK: '1', GPU: 'rtxpro6000', GPU_COUNT: '1', DRIVER: 'latest' }
-    - { ARCH: 'amd64', PY_VER: '3.13',  CUDA_VER: '13.3.0', LOCAL_CTK: '1', GPU: 'rtxpro6000', GPU_COUNT: '1', DRIVER: 'latest' }
+    - { ARCH: 'amd64', PY_VER: '3.13',  CUDA_VER: '13.3.0', LOCAL_CTK: '1', GPU: 'rtxpro6000', GPU_COUNT: '1', DRIVER: '610.43.02' }
```
leofang, line 41: xref https://github.com/NVIDIA/cuda-python/actions/runs/27105815636/job/79997044862?pr=2176

ci/test-matrix.yml
```diff
@@ -29,10 +38,10 @@ linux:
...
     - { ARCH: 'amd64', PY_VER: '3.14',  CUDA_VER: '12.9.1', LOCAL_CTK: '0', GPU: 't4',         GPU_COUNT: '1', DRIVER: 'latest' }
     - { ARCH: 'amd64', PY_VER: '3.14',  CUDA_VER: '13.0.2', LOCAL_CTK: '1', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest' }
-    - { ARCH: 'amd64', PY_VER: '3.14',  CUDA_VER: '13.3.0', LOCAL_CTK: '1', GPU: 'l4',         GPU_COUNT: '1', DRIVER: 'latest' }
+    - { ARCH: 'amd64', PY_VER: '3.14',  CUDA_VER: '13.3.0', LOCAL_CTK: '1', GPU: 'l4',         GPU_COUNT: '1', DRIVER: '610.43.02' }
```
leofang, line 44: xref https://github.com/NVIDIA/cuda-python/actions/runs/27105815636/job/79997044897?pr=2176

.github/workflows/ci.yml
```diff
@@ -24,6 +24,7 @@ on:
   schedule:
     # every 24 hours at midnight UTC
     - cron: "0 0 * * *"
+  workflow_dispatch:
```
mdboom: Might just be me, but this always looks like a bug in YAML when I see it.

```suggestion
  workflow_dispatch: {}
```

.github/workflows/test-wheel-linux.yml
```diff
@@ -101,21 +106,21 @@ jobs:
...
-    runs-on: "${{ matrix.FLAVOR || 'linux' }}-${{ matrix.ARCH }}-gpu-${{ matrix.GPU }}-${{ matrix.DRIVER }}-${{ matrix.GPU_COUNT }}"
+    runs-on: "${{ matrix.FLAVOR || 'linux' }}-${{ matrix.ARCH }}-gpu-${{ matrix.GPU }}-${{ matrix.RUNNER_DRIVER }}-${{ matrix.GPU_COUNT }}"
     # The build stage could fail but we want the CI to keep moving.
     if: ${{ github.repository_owner == 'nvidia' && !cancelled() }}
     # Our self-hosted runners require a container
     # TODO: use a different (nvidia?) container
     container:
-      options: -u root --security-opt seccomp=unconfined --shm-size 16g
+      # Custom-DRIVER rows need --privileged --pid=host so install_gpu_driver.sh
+      # can nsenter to the host for the install + refresh the toolkit bind mounts
```
mdboom, line 118: nsenter?

.github/workflows/test-wheel-linux.yml
```diff
@@ -131,6 +136,21 @@ jobs:
...
+      - name: Install GPU driver
+        if: ${{ matrix.DRIVER != 'latest' && matrix.DRIVER != 'earliest' }}
+        env:
+          DRIVER: ${{ matrix.DRIVER }}
+          GPU_TYPE: ${{ matrix.GPU }}
+        run: |
+          # util-linux for nsenter; install_gpu_driver.sh re-execs onto the
+          # host (requires --privileged --pid=host on the container, set
+          # conditionally above) and refreshes the toolkit bind mounts here.
+          apt-get -y install --no-install-recommends util-linux
```
mdboom: Should we add this to "Install dependencies" above instead, to avoid two invocations of `apt-get`?

ci/tools/install_gpu_driver.sh
```diff
@@ -0,0 +1,220 @@
...
+KEEP_RE='^(nvidia-container-toolkit(-base)?|libnvidia-container1|libnvidia-container-tools)$'
+
+in_container() {
+  [ -f /.dockerenv ] || grep -qE '/(docker|kubepods|containerd)' /proc/1/cgroup 2>/dev/null
+}
+
+host_install() {
+  apt-get -y install build-essential dkms "linux-headers-$(uname -r)" psmisc kmod
+
+  systemctl stop nvidia-persistenced dcgm-exporter 2>/dev/null || true
```
mdboom: `2>/dev/null` seems like it could hide useful error information.  Is this particularly chatty?  It shouldn't be, I don't think.

mdboom (changes requested):

.github/workflows/test-wheel-linux.yml
leofang: https://man7.org/linux/man-pages/man1/nsenter.1.html

It's used in `install_gpu_driver.sh` so that we can execute commands on the host from within the container, because driver installation/update must happen there.

.github/workflows/ci.yml
leofang: Good call — applied in fa7940aecd. Not a YAML parser bug FWIW (the spec says `key:` with nothing after means the value is `null`), but I agree `key: {}` reads as "deliberately empty" while a bare `key:` always looks like the value got cut off.

.github/workflows/test-wheel-linux.yml
leofang: Good catch — `util-linux` is actually already in the `ubuntu:22.04` base image (the CI logs from earlier runs showed `util-linux is already the newest version (2.37.2-4ubuntu3.5). 0 upgraded, 0 newly installed`), so my inline `apt-get install util-linux` was a no-op all along. Moved it into the `Install dependencies` step's declarative list anyway so the deps are declared in one place, and dropped the inline apt-get. Applied in fa7940aecd.

ci/tools/install_gpu_driver.sh
leofang: Agreed — the `2>/dev/null` was inherited verbatim from nv-gha-runners/vm-images#256's `nvgha-driver` with no specific need on our side, and `|| true` already keeps the script non-fatal on the common "unit not loaded" path. Dropped the redirect so real errors surface. Applied in fa7940aecd.

leofang:
/ok to test [fa7940a](https://github.com/NVIDIA/cuda-python/pull/2176/commits/fa7940aecd2be6f8104d41156de12ebfa00a43f6)

.github/workflows/test-wheel-linux.yml
```diff
@@ -85,8 +85,13 @@ jobs:
...
           TEST_MATRIX=$(yq -o json ".linux[\"${MATRIX_TYPE}\"] | map(select(.ARCH == \"${ARCH}\"))" ci/test-matrix.yml)
 
-          # Apply matrix filter and wrap in include structure
-          MATRIX=$(echo "$TEST_MATRIX" | jq -c '${{ inputs.matrix_filter }} | if (. | length) > 0 then {include: .} else "Error: Empty matrix\n" | halt_error(1) end')
+          # Apply matrix filter; reject custom DRIVER + FLAVOR=wsl (the
+          # in-container driver swap doesn't work under WSL); add a
+          # RUNNER_DRIVER field that maps any custom version back to
+          # 'latest' (the install script swaps the driver itself, so we
+          # need to land on the runner that ships with the most recent
+          # pre-installed driver); wrap in include structure.
```
kkraus14, line 93: Does this mean we can't test various driver versions under WSL?

.github/workflows/test-wheel-linux.yml
```diff
@@ -127,10 +132,22 @@ jobs:
...
+          # example; util-linux for `nsenter` (custom-DRIVER rows re-exec
+          # install_gpu_driver.sh onto the host through nsenter)
+          dependencies: "jq wget libgl1 libegl1 g++ util-linux"
           dependent_exes: "jq wget"
 
+      - name: Install GPU driver
+        if: ${{ matrix.DRIVER != 'latest' && matrix.DRIVER != 'earliest' }}
+        env:
+          DRIVER: ${{ matrix.DRIVER }}
+          GPU_TYPE: ${{ matrix.GPU }}
```
kkraus14, line 147: Do we need to handle the proprietary vs open kernel modules?

ci/tools/install_gpu_driver.sh
```diff
@@ -0,0 +1,220 @@
...
+    ARCH_DIR=Linux-aarch64
+    ARCH_SUFFIX=aarch64
+    ;;
+  *)
+    echo "::error::unsupported arch: $(uname -m)" >&2
+    exit 1
+    ;;
+esac
+
+URL="https://us.download.nvidia.com/XFree86/${ARCH_DIR}/${VERSION}/NVIDIA-Linux-${ARCH_SUFFIX}-${VERSION}.run"
```
kkraus14, line 69: Just to confirm is this just a driver runfile or does this have the toolkit bundled in as well?

We should also make sure this feeds through some kind of cache similar to the apt cache that exists.

ci/tools/install_gpu_driver.sh
```diff
@@ -0,0 +1,220 @@
...
+  # captured by KEEP_RE. Tolerate apt failures: postrm scripts can trip
+  # and the .run installer is about to replace everything anyway.
+  dpkg-query -W -f='${Package}\n' 'nvidia-*' 'libnvidia-*' 2>/dev/null \
+    | awk -v re="$KEEP_RE" '$0 !~ re' \
+    | xargs -r apt-get -y remove --purge || true
+
+  local d
+  d=$(mktemp -d)
+  ( cd "$d" \
+    && wget -q -O installer.run "$URL" \
```
kkraus14, line 113: We should probably validate some kind of SHA here before installing if possible

.github/workflows/test-wheel-linux.yml
leofang: M.V.P. use proprietary (hard-coded in `install_gpu_driver.sh`), the rest use open.

.github/workflows/test-wheel-linux.yml
leofang: Correct. When WSL runners start we're already in the WSL terminal, but the rule is any WSL driver update must be done on the native Windows side, not inside WSL.

ci/tools/install_gpu_driver.sh
leofang: Wouldn't it require us to hard-code the SHA in `test-matrix.yaml`?

ci/tools/install_gpu_driver.sh
leofang: Yes. https://download.nvidia.com/XFree86/Linux-x86_64/ is a known url linked from the driver download page (from [here](https://www.nvidia.com/en-in/drivers/), go to Linux Drivers -> [Unix Driver Archive](https://www.nvidia.com/en-in/drivers/unix/) and then select the CPU arch).

AFAIK, the runner team does not (yet) offer HTTPS cache. But most of CI pipelines aren't flipped in this PR (still on pre-installed earliest/latest drivers), and those that do get flipped have quick installation (<2 mins), so I am not concerned with perf.

leofang:
/ok to test [8dfd06e](https://github.com/NVIDIA/cuda-python/pull/2176/commits/8dfd06e5b90c7ad80c657e5963deecda2175729c)

ci/tools/install_gpu_driver.sh
kkraus14: It would involve building a lookup table somewhere, yes, mapping driver versions to SHAs. I don't think it's high priority, but something we should do in the fullness of time to guard against supply chain attacks (though I think this one is very unlikely and would have much wider spread impact than just us)

kkraus14 (approved):

mdboom (approved):

Issue #2174, opened 2026-06-06
Remove helper scripts and tests not exercised by CI


leofang:
While reviewing #2147 we noticed a few helper scripts and tests in the repo that are documented for developers (and AI agents) but aren't invoked by any CI workflow. They risk silently rotting because nothing validates them on a regular basis.

**Stance: remove them, rather than wiring them into CI.**

Reasoning: our CI/CD already runs the relevant tooling on every PR. Writing additional tests to validate that tooling is recursive — tests for tests, tests for tests for tests, and so on. There is no principled place to stop. If the CI/CD breaks, the team will see it immediately on the next PR and fix it. We should draw the line at the per-PR validation layer rather than maintain an orphaned meta-test suite.

## Concrete examples

**`ci/tools/tests/`**
- Contains `test_check_release_notes.py` and `test_fetch_ctk_redistrib.py` today (PR #2147 would add a third).
- Listed in `pytest.ini` `testpaths`, so a developer who runs plain `pytest` from the repo root would collect them — but no CI job does that.
- `.github/workflows/release.yml` has a `Self-test release-notes checker` step that runs `pytest ci/tools/tests`, but it is commented out.

**`scripts/run_tests.sh`**
- Recommended in `cuda_core/README.md` and the per-package `AGENTS.md` files as the orchestrated way to run tests locally.
- No `.github/workflows/` job invokes it; CI uses inline `pytest` calls against specific dirs in `test-wheel-linux.yml`, `test-wheel-windows.yml`, and `coverage.yml`.
- Its targets are hardcoded to `pathfinder | bindings | core | smoke | all` — there is no `tools` target, so it would not cover `ci/tools/tests` either.

## Proposed action

- Delete `ci/tools/tests/` and the corresponding `pytest.ini` entry.
- Delete `scripts/run_tests.sh` and drop / rewrite the README and `AGENTS.md` guidance that points at it.

-- Leo's bot

PR #2178, opened 2026-06-07
Accept post-release wheel versions in release validation

fallintoplace:
## Summary
- reuse the release-note tag parser in `validate-release-wheels`
- preserve `.postN` versions from release tags instead of truncating to `X.Y.Z`
- keep release wheel tag parsing aligned with the existing release-note tag rules

## Why
Post-release tags such as `v12.6.2.post1` were parsed as `12.6.2`, so valid wheels like `cuda_bindings-12.6.2.post1-...whl` failed release validation before upload or publishing.

## Testing
- `uvx ruff check ci/tools/validate-release-wheels`
- `uvx ruff format --check ci/tools/validate-release-wheels`
- direct smoke test: `./ci/tools/validate-release-wheels v12.6.2.post1 cuda-bindings <tmp-wheel-dir>`
- direct negative smoke test: `./ci/tools/validate-release-wheels v12.6.2 cuda-bindings <tmp-wheel-dir>` rejects a `12.6.2.post1` wheel


ci/tools/tests/test_validate_release_wheels.py
leofang, line 1: As per #2174 I have a strong preference to not add more files in this folder. Let us remove this and just focus on the `ci/tools/validate-release-wheels` change.

leofang (changes requested):

ci/tools/validate-release-wheels
```diff
@@ -62,9 +68,14 @@ def parse_wheel_dist_and_version(path: Path) -> tuple[str, str]:
     return parts[0], parts[1]
 
 
-def main() -> int:
-    args = parse_args()
-    expected_version = version_from_tag(args.git_tag)
+def main(argv: list[str] | None = None) -> int:
```
leofang: The `argv` plumbing here and on `parse_args` was added for the test file that the prior review asked to drop. Without tests it's dead API surface — recommend reverting to keep the diff minimal.

ci/tools/validate-release-wheels
```diff
@@ -39,18 +42,21 @@ def parse_args() -> argparse.Namespace:
...
+        if (version := parse_version_from_tag(tag, tag_component)) is not None
+    }
+    if len(versions) == 1:
+        return versions.pop()
     raise ValueError(
         "Unsupported git tag format "
-        f"{tag!r}; expected tags beginning with vX.Y.Z, cuda-core-vX.Y.Z, "
-        "or cuda-pathfinder-vX.Y.Z."
+        f"{tag!r} for component {component!r}; expected vX.Y.Z, cuda-core-vX.Y.Z, "
+        "or cuda-pathfinder-vX.Y.Z with a valid release version."
```
leofang: This message still says `vX.Y.Z` only and doesn't reflect the new `.postN` support, so the error contradicts the behavior. Suggest `vX.Y.Z[.postN]` (and the same for the prefixed variants), or drop the format examples and reference the canonical tag rules.

leofang (commented):
Two nits before merge: the `argv` plumbing on `main`/`parse_args` is unused without tests and can be reverted, and the `ValueError` message should mention the new `.postN` support so it stops contradicting the behavior. Otherwise the rewrite looks good — also closes a latent unanchored-regex bug where `v12.6.2.dev1` would silently validate a `12.6.2` wheel.

leofang (approved):

lijinf2 (approved):
Thank you for the PR. This fixes git tag matching for both regular releases and .postN post-releases. I see all comments have been addressed. Looks good to me.

PR #1978, opened 2026-04-25
fix(test): unblock root pixi run test workflow

rparolin:
## Summary

`pixi run test` was failing end-to-end. Two stacked root causes:

1. **List-form pixi `cmd` arrays didn't expand `$PIXI_ENVIRONMENT_NAME`.** The root `test-*` tasks dropped into each sub-package without forwarding the active environment, so the inner `pixi run` picked the sub-package's `default` environment. In `cuda_bindings/pixi.toml` and `cuda_core/pixi.toml`, `default` has no `cuda-version` pin — the conda solver resolved a CUDA-12.x toolkit, and the build failed compiling Cython output that references `CUatomicOperation_enum` (a CUDA-13.x-only symbol). Fix: wrap each `test-*` task in `bash -c '…'` so the shell expands `$PIXI_ENVIRONMENT_NAME` and forwards it explicitly via `-e` to the inner `pixi run`.

2. **Cython couldn't find cross-package `.pxd` files under pixi-build's editable install.** pixi-build installs `cuda-bindings` via a PEP 660 finder hook; runtime imports honor it but Cython's filesystem `.pxd` resolver does not. Fix: replace the `cythonize` CLI invocation in both `cuda_core/tests/cython/build_tests.sh` and `cuda_bindings/tests/cython/build_tests.sh` with a small Python wrapper (`build_tests.py`) that calls `Cython.Build.cythonize()` with an explicit `include_path` resolved at runtime from `cuda.bindings.__file__`. This avoids platform-specific `PYTHONPATH` separator handling and surfaces missing-import failures as Python exceptions instead of silent fallbacks.

## Files changed

- `pixi.toml` — three `test-*` tasks switched to `bash -c '…'` form forwarding `-e "$PIXI_ENVIRONMENT_NAME"`; comment block expanded with rationale and Linux-only scope note.
- `cuda_core/tests/cython/build_tests.sh` — adds `set -euo pipefail`; invokes new `build_tests.py` instead of the `cythonize` CLI; still owns `CPLUS_INCLUDE_PATH` / `CL` setup.
- `cuda_core/tests/cython/build_tests.py` — **new**; resolves the cuda namespace package's parent dir at runtime and runs `cythonize() + setup(... build_ext --inplace)` with an explicit `include_path`.
- `cuda_bindings/tests/cython/build_tests.sh` — adds `set -euo pipefail`; switches to the new wrapper; preserves `CPLUS_INCLUDE_PATH` / `CL` setup.
[...]

PR #2180, opened 2026-06-08
Fix pixi run test: CUDA 13.3 pin lag + cython-test .so placement

### LINKED ISSUE #2182: pixi run test fails on main: CUDA 13.2 pin lags 13.3 bindings + cython-test .so misplacement
## Summary

`pixi run test` (env `cu13`) does not complete from a clean checkout on `main`. Two independent root causes:

1. **CUDA version pin lag.** The bindings were regenerated against CUDA 13.3.0 (#2139), adding NVRTC symbols (`NVRTC_ERROR_BUSY`, `nvrtcBundledHeadersInfo`, `nvrtcGetBundledHeadersInfo`), but the pixi `cuda-version` pins stayed at `13.2.*` in `cuda_bindings/pixi.toml` and `cuda_core/pixi.toml`. The local build then compiles 13.3-referencing Cython code against a 13.2 `nvrtc.h`:
   ```
   'nvrtcBundledHeadersInfo' was not declared in this scope
   ```
   CI was unaffected because it builds wheels from `ci/versions.yml` (13.3.0), not via `pixi run test`.

2. **Cython-test `.so` placement.** `tests/cython/build_tests.py` runs `build_ext --inplace`, which writes the `.so` relative to cwd; pixi runs the task from the project root, so the `.so` lands in the package root instead of `tests/cython/`, where pytest imports it by bare module name. Latent since #1978 — a stale, correctly-placed `.so` masked it. A clean checkout fails with `ModuleNotFoundError`.

## Fix

PR #2180 — bumps the `cuda-version` pins to `13.3.*` and `chdir`s before `build_ext --inplace` in both `build_tests.py`.

## Notes

There were also two local-only stale-artifact symptoms while bisecting (stale pyclibrary parser caches `cache_driver`/`cache_runtime`, and a stale compiled `.so`) that don't exist on a clean checkout — captured in the follow-up issue on drift prevention.

🤖 Filed with [Claude Code](https://claude.com/claude-code)

rparolin:
## Summary

`pixi run test` did not complete from a clean checkout. Two independent root causes, both downstream of CUDA toolkit drift:

1. **CUDA version pin lag.** The bindings were regenerated against CUDA **13.3.0** (cc50515648, "Support CUDA 13.3"), adding NVRTC symbols (`NVRTC_ERROR_BUSY`, `nvrtcBundledHeadersInfo`, `nvrtcGetBundledHeadersInfo`), but the pixi `cuda-version` pins stayed at **13.2** in `cuda_bindings/pixi.toml` and `cuda_core/pixi.toml`. The local build then compiled 13.3-referencing Cython code against a 13.2 `nvrtc.h`:
   ```
   'nvrtcBundledHeadersInfo' was not declared in this scope
   ```
   CI was unaffected because it builds wheels from `ci/versions.yml` (13.3.0), not via `pixi run test`.

2. **Cython-test `.so` placement bug (latent since #1978).** `tests/cython/build_tests.py` runs `build_ext --inplace`, which writes the `.so` relative to cwd. pixi runs the task from the project root, so the `.so` landed in the package root instead of `tests/cython/`, where pytest imports it by bare module name. #1978's verification silently reused a correctly-placed `.so` from an earlier build; a clean checkout fails with `ModuleNotFoundError`.

## Changes

- **Pin bump** `13.2.* → 13.3.*` (build-variants + `feature.cu13`) in both `pixi.toml`, matching the regenerated sources and `ci/versions.yml`; re-solved `pixi.lock` ×2.
- **`os.chdir(script_dir)`** before `build_ext --inplace` in both `build_tests.py` (kept aligned per #1978) so the `.so` lands next to its `.pyx` regardless of cwd.

## Verification

`pixi run test` (env `cu13`) now completes, exit 0:

| Subpackage | Passed | Skipped | xfailed |
|------------|--------|---------|---------|
| pathfinder | 975    | 4       | —       |
| bindings   | 429    | 20      | —       |
| core       | 3373   | 210     | 3       |

Confirmed the rebuilt `.so` lands in `tests/cython/` (package root clean), proving the placement fix rather than a leftover artifact.

---
Closes #2182. Follow-up: #2183.

### DIFF
--- a/cuda_bindings/pixi.toml
+++ b/cuda_bindings/pixi.toml
@@ -10,7 +10,7 @@ preview = ["pixi-build"]
 [workspace.build-variants]
 python = ["3.10.*", "3.11.*", "3.12.*", "3.13.*", "3.14.*"]
 # Keep source-package metadata aligned with the consuming environment's CUDA major.
-cuda-version = ["12.*", "13.2.*"]
+cuda-version = ["12.*", "13.3.*"]
 
 [feature.test.dependencies]
 cuda-bindings = { path = "." }
@@ -74,7 +74,7 @@ CUDA_HOME = "$CONDA_PREFIX/Library"
 cuda-version = "12.*"
 
 [feature.cu13.dependencies]
-cuda-version = "13.2.*"
+cuda-version = "13.3.*"
 
 [environments]
 default = { features = ["test", "cython-tests"], solve-group = "default" }
--- a/cuda_bindings/pixi.lock
+++ b/cuda_bindings/pixi.lock
@@ -559,21 +559,21 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/cairo-1.18.4-he90730b_1.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/colorama-0.4.6-pyhd8ed1ab_1.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/conda-gcc-specs-15.2.0-h53410ce_16.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-cccl_linux-64-13.2.27-ha770c72_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/cuda-crt-dev_linux-64-13.2.51-ha770c72_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-cudart-13.2.51-hecca717_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/cuda-cudart-dev-13.2.51-hecca717_0.conda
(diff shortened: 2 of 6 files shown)


cuda_bindings/pixi.toml
```diff
@@ -74,7 +74,7 @@ CUDA_HOME = "$CONDA_PREFIX/Library"
 cuda-version = "12.*"
 
 [feature.cu13.dependencies]
-cuda-version = "13.2.*"
+cuda-version = "13.3.*"
```
leofang, line 77: Did we always manually update the version here for every CTK bring-up? lol

leofang (approved):

cuda_bindings/pixi.toml
rparolin: Looks like it. That's why I'm adding another CI validation pass to catch these.

PR #2168, opened 2026-06-05
coverage: add cuda.core tests for host, launcher cluster, strided layout, program-cache, and DLPack utils

rluo8:
## Description

Tests covered:

| File | New tests | Target source |
|------|-----------|----------------|
| `test_host.py` | `repr` for all three `Host` forms; pickle/copy round-trip preserves the singleton via `__reduce__` | `_host.py` |
| `test_launcher.py` | thread-block-cluster cc gate (accept ≥9.0 / reject <9.0, `Device` mocked); cluster branch of `_to_native_launch_config` | `_launch_config.pyx` |
| `test_program_cache.py` | `FileStream` size-cap `__delitem__`/`clear`, entry-path iteration with missing/stray dirs, tmp-size accounting, `touch_atime` fallback + swallowed open/fstat failures | `utils/_program_cache/_file_stream.py` |
| `test_strided_layout.py` | validation/error paths: `dense_like`, repr-with-offset, min/max offset, `broadcast_to`, `unsqueezed`, `repacked` pack/unpack, `max_compatible_itemsize` | `_layout.pyx` |
| `test_utils.py` | `as_tensor_map` kwargs assembly; DLPack export round-trip over all NumPy-native dtypes + scalar/zero-volume shapes; versioned/unversioned capsule deleters; `from_dlpack` TypeError→unversioned fallback; CAI device-pointer/stream-ordering; DLPack C-exchange-API helpers | `_memoryview.pyx` |

 Coverage impact

Measured with paired full-coverage runs (no-patch baseline vs this patch):

| Scope | Before | After | Δ |
|-------|-------:|------:|---|
| `cuda/core/` overall | 79.81% | 81.84% | +2.03% (−213 missed) |
| `_memoryview.pyx` | 65.82% | 85.73% | +19.92% |
| `_host.py` | 83.64% | 100.00% | +16.36% |
| `_launch_config.pyx` | 77.46% | 91.55% | +14.08% |
| `_layout.pyx` | 89.59% | 96.23% | +6.64% |
| `_file_stream.py` | 84.51% | 90.49% | +5.99% |

Testing

- Full local run on py314: all added tests pass.
- Cross-platform on local GPU runners: **Linux x64 + Windows x64**, green.
- Additionally verified on a B100 windows + Tesla T4 host.

PR #2181, opened 2026-06-09
coverage: add cuda.core test_utils.py tests for DLPack/StridedMemoryView

rluo8:
## Description

Split out from #2168 per review. Adds tests targeting
cuda/core/experimental/_memoryview.pyx:

- DLPack export round-trip over all NumPy-native dtypes and
  scalar/zero-volume shapes
- versioned/unversioned capsule + deleter paths
- from_dlpack: ambiguous-stream and unsupported-device errors, TypeError -> unversioned-import fallback
- CAI path: device-pointer + stream-ordering view, deprecated init path, has_dlpack=False proxy, device-accessible view, None-dtype repr
- DLPack C exchange-API helpers 

Coverage change:
_memoryview.pyx coverage 65.82% -> 85.73%.


cuda_core/tests/test_utils.py
```diff
@@ -1094,3 +1094,425 @@ def test_strided_memory_view_dtype_roundtrip_all(dtype):
...
     assert view.dtype == np.dtype(dtype)  # .dtype triggers dtype_dlpack_to_numpy
+
+
+def test_as_tensor_map_assembles_kwargs(monkeypatch):
+    """``as_tensor_map`` forwards the view + box_dim and only the non-None
+    tiled options to ``TensorMapDescriptor._from_tiled``.
+
+    The real ``_from_tiled`` requires a device-accessible, 16-byte-aligned view
+    on TMA-capable hardware (sm90+), so we replace the (module-level) class the
+    method imports with a recorder and assert the assembled call instead.
```
seberg: Not sure, but can't we just skip the test if the hardware isn't available?  And if it is, I expect we are already testing this?

I feel this test belongs into `test_tensormap` as a single test that checks that tensormap instantiation fails gracefully when TMA isn't available.

cuda_core/tests/test_utils.py
```diff
@@ -1094,3 +1094,425 @@ def test_strided_memory_view_dtype_roundtrip_all(dtype):
...
+    )
+    obj.__cuda_array_interface__["stream"] = int(producer.handle)
+
+    view = StridedMemoryView.from_cuda_array_interface(obj, stream_ptr=consumer.handle)
+
+    assert view.is_device_accessible is True
+    assert view.ptr == int(buffer.handle)
+    assert view.device_id == dev.device_id
+    assert view.shape == (8,)
+    dev.default_stream.sync()
```
seberg, line 1102: I find this test a bit deceptive since it adds coverage without actually checking stream order correctness.  (It does check for device correctness of course).
Now, testing stream-order correctness is annoying (you need either large buffers, or launch many kernels on the first stream and then e.g. a copy-to-host on after export and see that all of them finished)
So maybe it's OK, but the bot oversells the test by a lot.

cuda_core/tests/test_utils.py
```diff
@@ -1094,3 +1094,425 @@ def test_strided_memory_view_dtype_roundtrip_all(dtype):
...
+
+
+def test_strided_memory_view_init_cai_path_deprecated(init_cuda):
+    """The deprecated ``StridedMemoryView(obj)`` constructor routes a CAI-only
+    object through the CAI branch (warn + ``view_as_cai``), not the DLPack one."""
+    obj = _make_cuda_array_interface_obj(shape=(4,), strides=None, typestr="<f4", data=(0, False))
+    with pytest.deprecated_call(match="deprecated"):
```
seberg:
```suggestion
    with pytest.deprecated_call(match="CUDA-array-interface-supporting"):
```
if we bother to match a string, try to make it unique.

cuda_core/tests/test_utils.py
```diff
@@ -1094,3 +1094,425 @@ def test_strided_memory_view_dtype_roundtrip_all(dtype):
...
+    assert r.startswith("StridedMemoryView(ptr=")
+    assert "dtype=NoneType" in r
+
+
+# ---------------------------------------------------------------------------
+# DLPack C exchange API (`__dlpack_c_exchange_api__`)
+#
+# Drive the C function pointers exposed by the capsule the way a native
+# consumer would, exercising the StridedMemoryView exchange-API implementation.
+# ---------------------------------------------------------------------------
```
seberg: We can add this, but I am starting to think this should be in it's own `test_dlpack` file, there is a lot of complexity here.

cuda_core/tests/test_utils.py
```diff
@@ -1094,3 +1094,425 @@ def test_strided_memory_view_dtype_roundtrip_all(dtype):
...
+    assert out.ndim == 2
+    assert out.device.device_type == int(DLDeviceType.kDLCPU)
+    assert out.data == src.ctypes.data
+    assert [out.shape[i] for i in range(out.ndim)] == [3, 4]
+
+
+def _exchange_api_cause(exc):
+    """Underlying exception raised by the noexcept C fn (surfaced by ctypes as
+    SystemError, with the real error chained as __cause__ or __context__)."""
+    return exc.value.__cause__ or exc.value.__context__
```
seberg: Unpacking the `SystemError` is a bad workaround for not supplying an `errcheck` function to the ctypes definition.
We shouldn't test for `SystemError` (unless we set them ourselves, which we don't).

cuda_core/tests/test_utils.py
```diff
@@ -1094,3 +1094,425 @@ def test_strided_memory_view_dtype_roundtrip_all(dtype):
...
+    assert isinstance(_exchange_api_cause(exc), RuntimeError)
+    assert not out_obj.value  # set to NULL before the error
+
+
+def test_dlpack_c_exchange_api_managed_tensor_allocator_not_supported():
+    """``managed_tensor_allocator`` is unsupported (NotImplementedError)."""
+    api = _get_exchange_api()
+    out = ctypes.c_void_p(123)
+    with pytest.raises(SystemError) as exc:
+        api.managed_tensor_allocator(None, ctypes.byref(out), None, None)
```
seberg:
```suggestion
        # Currently sets a Python error when `SetErr` isn't passed
        api.managed_tensor_allocator(None, ctypes.byref(out), None, None)
```
This is OK, honestly, I am not sure if there is a point in even forcing this function to be non-NULL on the dlpack side if we just always make an error, but whatever.

I just prefer mentioning in the test that we know it's a weird test.

cuda_core/tests/test_utils.py
```diff
@@ -1094,3 +1094,425 @@ def test_strided_memory_view_dtype_roundtrip_all(dtype):
         pytest.skip(f"NumPy does not export {np.dtype(dtype)} via DLPack: {e}")
     view = StridedMemoryView.from_dlpack(src, stream_ptr=-1)
     assert view.dtype == np.dtype(dtype)  # .dtype triggers dtype_dlpack_to_numpy
```
seberg: Might as well delete this test and move the comment about bfloat16 to `_NUMPY_NATIVE_DLPACK_DTYPES` your new test seems basically identical but expands it a bit.

cuda_core/tests/test_utils.py
```diff
@@ -1060,25 +1060,25 @@ def test_dlpack_export_non_native_endian_rejected():
...
+    np.int16,
+    np.int32,
+    np.int64,
+    np.float16,
+    np.float32,
+    np.float64,
+    np.complex64,
+    np.complex128,
+    np.bool_,
 )
```
seberg:
```suggestion
)
if ml_dtypes is not None:
    # Supported on NumPy 2.5 and ml_dtypes (probably) 0.5.5+
    _NUMPY_NATIVE_DLPACK_DTYPES += (ml_dtypes.bfloat16,)
```
Although then the "native" is slightly off :).  But we have the `try/except` that should make this skipped and NumPy 2.5 + next `ml_dtypes` version this should actually work.
(In fact, we can probably add more dtypes if we want to.)

cuda_core/tests/test_utils.py
```diff
@@ -1094,3 +1094,425 @@ def test_strided_memory_view_dtype_roundtrip_all(dtype):
...
+    with pytest.raises(BufferError, match="device not supported"):
+        StridedMemoryView.from_dlpack(_FakeUnsupportedDevice(), stream_ptr=0)
+
+
+class _DLPackNoMaxVersion:
+    """Wraps a NumPy array but rejects the ``max_version`` kwarg, forcing the
+    TypeError fallback in ``view_as_dlpack`` and an *unversioned* capsule import."""
+
+    def __init__(self, arr):
+        self._arr = arr
```
seberg:
```suggestion
        self._arr = StridedMemoryView.from_any_interface(arr)
```
I would suggest to use a `StridedMemoryView` here rather than NumPy.  Should do the same thing, but means we don't care if NumPy changes the behavior (which eventually it may) and stops exporting version 0.x (unversioned).

seberg (commented):
I'd wait for @mdboom to have a brief look. As we discussed yesterday, I am not sure about all tests but I guess the improved coverage is really nice.
FWIW, in these changes I don't really see trivial gains for parametrization.

That said, I left a few comments and especially the TMD tests seem like a lot of effort to add coverage that should already exists (but you need the support to see that in the coverage).
I think it may make sense to add *one* test (which could use funky arguments) to check that an error is given when it isn't available.  But I wouldn't add monkeypatch tests just to get coverage up here, I think.
(We should cover this with *real* tests not with dummy tests.)

The DLPack tests are pretty complex unittests with ctypes, I guess they are fine (this is unused, if there are bugs mirrored in implementation and tests we won't even notice, though -- which is a bit deceptive; it seemed fine but I'd have to look closer to be confident).
Either way, I feel this file is big enough and those tests complicated enough, that it may be good to split DLPack tests into a new one?  (`test_utils_dlpack.py` or a new `utils` folder?)

cuda_core/tests/test_utils.py
rluo8: Thanks for the thorough review, @seberg !  I'll update based on these comments.

rluo8: Thanks @seberg . I have updated the test cases. Dropped both monkeypatch tests, and added one real test test_as_tensor_map_host_view_rejected_without_tma to test_tensor_map.py.

cuda_core/tests/test_utils.py
rluo8: Thanks for pointing this out. I've updated the docstring to be honest about what the test does and doesn't check.

cuda_core/tests/test_utils.py
rluo8: Updated.

cuda_core/tests/test_utils.py
rluo8: I have moved all the DLPack/exchange-API tests into a new file test_utils_dlpack.py.

cuda_core/tests/test_utils.py
rluo8: Thanks. I dropped the SystemError unpacking. The exchange-API function pointers are now declared with ctypes.PYFUNCTYPE instead of CFUNCTYPE, so ctypes surfaces the Python exception set by the Cython function directly. The tests now assert the intended TypeError/RuntimeError/NotImplementedError instead of inspecting SystemError.__cause__/__context__.

cuda_core/tests/test_utils.py
rluo8: I have updated the test with suggested comments.

cuda_core/tests/test_utils.py
rluo8: Done. Deleted the test, and moved the bfloat16 comment to _NUMPY_NATIVE_DLPACK_DTYPES.

cuda_core/tests/test_utils.py
rluo8: Done. I added ml_dtypes.bfloat16 to _NUMPY_NATIVE_DLPACK_DTYPES when ml_dtypes is available. And unsupported NumPy/ml_dtypes combinations are still skipped by the per-test DLPack probe.

cuda_core/tests/test_utils.py
rluo8: Updated. Used a StridedMemoryView instead of NumPy, and passed stream_ptr=-1.

rwgk:
Drive-by suggestion: @rluo8 could you please try adopting #2232 here, then ask your agent to add the tags? It'll be interesting to see if it works as intended.

rluo8:
> Drive-by suggestion: @rluo8 could you please try adopting #2232 here, then ask your agent to add the tags? It'll be interesting to see if it works as intended.

Hi @rwgk, I asked the claude code to apply #2232 to add the tags. It has added the HUMAN-REVIEWED tag in the tests.

seberg (approved):
Thanks for the updates! I am happy enough, but will keep it for someone to press the green-button.

About the "human reviewed".  I am not sure the placement is as intended by @rwgk as it is file scoped now and above tests rather than inside a test (or at least class?)
I might stick with "AI generated" for the "exchange API" tests.  I had a look for correctness but if one of those fails, I think it's safe to assume they test current behavior more than correctness.

(One other thing is the `test_utils_dlpack.py` naming, but it is also easy to change at any point.)

rwgk:
Hi @rluo8: thanks for trying out PR #2232. Based on feedback I got from @mdboom in the meantime, I created an alternative PR #2245, which I prefer myself. For this PR, I think it'll be best to take out the tags and merge. Maybe we can add back annotations later, after the AGENTS.md change has landed.

rwgk:
@rluo8 QQ: did your agent automatically decide to use `HUMAN-REVIEWED` for all added tests? Or did you direct it to?

rluo8:
Thanks @seberg @rwgk . I have removed the tags for the tests. Happy to rename test_utils_dlpack.py whenever - just let me know your preference.

rluo8:
> @rluo8 QQ: did your agent automatically decide to use `HUMAN-REVIEWED` for all added tests? Or did you direct it to?

Hi @rwgk , the agent's default was actually AGENT-AUTHORED <model> — it wanted to leave the HUMAN-REVIEWED upgrade to a reviewer. Then I told it about the review comments , and it decided to change it to HUMAN-REVIEWED.

Issue #2183, opened 2026-06-09
Make pixi cuda-version pins track ci/versions.yml to prevent toolkit-version drift


rparolin:
## Problem

The canonical CUDA build version lives in `ci/versions.yml` (`cuda.build.version`), and the generated Cython bindings carry a `# generated with version X` stamp. But the local `pixi run test` path pins CUDA independently via `cuda-version` in `cuda_bindings/pixi.toml` and `cuda_core/pixi.toml` (`[workspace.build-variants]` + `[feature.cu13.dependencies]`).

These drifted: #2139 regenerated the bindings to 13.3 but left the pixi pins at 13.2, which silently broke `pixi run test` while CI stayed green (see #2182, fixed by #2180). This will recur on the next CUDA bump.

## Proposal

Have the pixi `cuda-version` pins derive from / stay in lockstep with `ci/versions.yml` (e.g. a check or a generation step), so a toolkit bump can't leave the local test env behind.

## Related latent weakness

While diagnosing #2182, two local caches did **not** invalidate when the CUDA header version changed underneath them, producing confusing errors:
- the pyclibrary parser cache (`CUDA_PYTHON_PARSER_CACHING=1` → `cache_driver`/`cache_runtime`) regenerated `driver.pyx` referencing a struct member the new 13.3 header had dropped;
- setuptools `build_ext` left a stale compiled cython-test `.so` (a `size changed` binary-incompat error).

Both are gitignored local artifacts (absent on a clean checkout), but worth considering whether a CUDA-version change should invalidate them.

🤖 Filed with [Claude Code](https://claude.com/claude-code)

Issue #2091, opened 2026-05-15
CI: Optimize `fetch-depth: 0` usage across workflows


leofang:
All 14 `fetch-depth: 0` checkouts across our workflows do a full clone with every blob in history. Most don't need that. Optimizing these would speed up CI, especially on Windows where git operations are slower.

### Can use `fetch-depth: 1` (no history needed — 6 occurrences)

| File | Job / line | Why history is unnecessary |
|---|---|---|
| `ci.yml` | `ci-vars` (L38) | Just reads `ci/versions.yml` with yq |
| `build-docs.yml` | docs build (L50) | Checks out a specific tag/ref, reads config |
| `release-upload.yml` | upload (L46) | `git archive` on a tag already checked out |
| `release.yml` | `check-tag` (L65) | Only uses GitHub CLI, no git ops |
| `release-cuda-pathfinder.yml` | `upload-assets` (L164) | `git archive` on a tag already checked out |
| `coverage.yml` | `coverage-windows` (L272) | Downloads pre-built wheels, no build step |

### Can use `fetch-depth: 1` + `fetch-tags: true` (only need tag resolution — 2 occurrences)

| File | Job / line | Why |
|---|---|---|
| `release.yml` | `determine-run-id` (L47) | `lookup-run-id` calls `git rev-parse` on a tag |
| `release-cuda-pathfinder.yml` | `prepare` (L64) | Same — `lookup-run-id` resolves tag to SHA |

### Should use `fetch-depth: 0` + `filter: blob:none` (need commit graph, not blobs — 7 occurrences)

These need the full commit graph (for `setuptools-scm` / `git describe`, `git merge-base`, or `git worktree`), but don't need historical blobs. A [treeless clone](https://github.blog/open-source/git/get-up-to-speed-with-partial-clone-and-shallow-clone/) (`filter: blob:none`) fetches the commit graph without blob data, which is significantly faster.

Verified locally that `git describe` produces identical output with a treeless clone.

| File | Job / line | Why it needs the commit graph |
|---|---|---|
| `build-wheel.yml` | build (L45) | setuptools-scm (`git describe`) |
| `ci.yml` | `detect-changes` (L109) | `git merge-base` + `git diff --name-only` |
| `cleanup-pr-previews.yml` | cleanup (L34) | `git worktree` on gh-pages |
| `coverage.yml` | `coverage-linux` (L69) | setuptools-scm |
[...]

PR #2185, opened 2026-06-09
ci: validate `pixi run test` source-build (PR smoke + nightly GPU)

rparolin:
## Summary

Main CI tests prebuilt **wheels** (`ci.yml` / `test-wheel-*.yml`) and never exercises the pixi **source build**, so the `pixi run test` developer path rots silently whenever the CUDA pin, generated bindings, conda-forge packages, or cython-test build mechanics drift. This adds the missing guard.

Two tiers, to spend GPU minutes deliberately:
- **`build-smoke`** (PRs touching the at-risk files): **CPU-only** on `ubuntu-latest`. Source-builds bindings + core, imports them, builds the cython test extensions and checks the `.so` lands in `tests/cython/`. Catches the compile / ABI / `.so`-placement regressions **without a GPU**.
- **`full-test`** (nightly cron + manual `workflow_dispatch`): GPU runner, full `pixi run test`.

Shared pixi install is factored into a composite action (`.github/actions/setup-pixi`) with an **explicit, asserted** version pin.

## Why separate from #2180

This is the *prevention* mechanism; #2180 is the *fix* for the current breakage. Kept separate per review.

## Sequencing / expected CI

This branch is based on `main`, which still has the breakage until #2180 merges. So **`build-smoke` is expected to FAIL on this PR** — that failure is the guard correctly catching #2182. Merge #2180 first (or rebase this on it) and it goes green.

## For reviewers to confirm

- Runner label `linux-amd64-gpu-l4-latest-1` (from `ci/test-matrix.yml`) — swap if a nightly-reserved label is preferred.
- pixi installed via the official installer pinned to `PIXI_VERSION`; `prefix-dev/setup-pixi` pinned to a SHA is a reasonable alternative.
- Suggest landing the new check **non-blocking** until proven stable.

Relates to #2183 (validate the source-build path over time). Guards against #2182 (fixed by #2180).

---
**Update:** rebased onto the #2180 fix branch so the source build is green (this PR was previously based on clean `main`, where `build-smoke` correctly caught #2182). It now stacks on #2180 — the pin/placement commits drop out once #2180 merges and this is rebased onto `main`.


.github/actions/setup-pixi/action.yml
```diff
@@ -0,0 +1,36 @@
...
+  using: "composite"
+  steps:
+    - name: Install pixi
+      shell: bash
+      env:
+        # The official installer honors PIXI_VERSION; pass it explicitly so the
+        # pin is visible at the call site, not implicit via a workflow-level env.
+        PIXI_VERSION: ${{ inputs.pixi-version }}
+      run: |
+        curl -fsSL https://pixi.sh/install.sh | bash
```
mdboom: We should download a pinned version of this and check the hash.

The setup-pixi GHA is probably more secure than this -- it doesn't seem to validate the hash of the binary, but the installation commands are in `setup-pixi` itself and can be pinned.  We could probably just use that instead of this entire file.

(I have the scars for the [codecov.io](https://about.codecov.io/security-update/) incident that stole keys from my employer using this method).

mdboom (changes requested):
LGTM, other that the strong security recommendation.

.github/actions/setup-pixi/action.yml
rparolin: Good call — thanks, and yikes re: codecov. Switched to `prefix-dev/setup-pixi` pinned to a commit SHA (`5185adf`, v0.9.6) and deleted the composite `curl|bash` action entirely (2195344). The install logic now lives in the pinned action rather than an unverified installer. `run-install: false` so it just puts the pinned pixi binary on PATH; the jobs drive `pixi run` themselves.

mdboom (approved):

.github/workflows/ci-pixi-source-test.yml
```diff
@@ -0,0 +1,149 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    steps:
+      - name: Checkout ${{ github.event.repository.name }}
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          # Full history + tags so setuptools-scm derives the real (13.x)
+          # package version; a shallow checkout yields 0.1.dev1, which trips
+          # cuda.core's "cuda.bindings 12.x or 13.x must be installed" guard.
+          fetch-depth: 0
```
leofang, line 71: Would be nice to add `filter: blob:none` here, see #2091.

.github/workflows/ci-pixi-source-test.yml
```diff
@@ -0,0 +1,149 @@
...
+          apt-get install -y --no-install-recommends \
+            ca-certificates git libgl1 libegl1
+
+      - name: Checkout ${{ github.event.repository.name }}
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          # Full history + tags so setuptools-scm derives the real (13.x)
+          # package version; a shallow checkout yields 0.1.dev1, which trips
+          # cuda.core's "cuda.bindings 12.x or 13.x must be installed" guard.
+          fetch-depth: 0
```
leofang, line 134: ditto

PR #2175, opened 2026-06-07
[doc-only] cuda.core: document how to install nightly (top-of-tree) builds

matysanchez:
Closes #2166

## What
Adds an **Installing the latest nightly (top-of-tree builds)** section to `cuda_core/docs/source/install.rst`. The section explains how to use `gh run download` to grab the wheel artifact from the most recent successful CI run on `main`, with notes for picking your Python version, target platform, and CUDA major version.

## Why
Per the issue, the equivalent section already exists in [numba-cuda-mlir](https://github.com/NVIDIA/numba-cuda-mlir/blob/main/INSTALL.md#top-of-tree-builds). The two repos share CI infrastructure, so the recipe transfers directly — only repo name, workflow file (`ci.yml`), and artifact pattern names needed adapting.

## Tested
- Verified `gh run list -R NVIDIA/cuda-python -w ci.yml -b main -s success -L1` resolves cleanly against the live repo (returns the latest successful CI run).
- Pure additive change — no surrounding content modified. Indentation matches the surrounding RST blocks.

## Verification
```
branch: hunter/issue-2166
files-changed: 1
commit: 6479b7f
```

Issue #2166, opened 2026-06-03
Document how to get the latest nightly in the developer guide


leofang:
Copy/paste this section from numba-cuda-milir:
https://github.com/NVIDIA/numba-cuda-mlir/blob/main/INSTALL.md#top-of-tree-builds
(the CI infra is the same.)

PR #2187, opened 2026-06-09
[doc-only] cuda.{core,bindings,pathfinder}: nightly install docs and unify "Development environment" structure

leofang:
## What

Extends #2175 to the remaining components named in #2166 (which covers all three of cuda.pathfinder, cuda.bindings, and cuda.core — the original PR only added the nightly recipe for cuda.core), and unifies the install-doc structure across all three components.

**Commit 1 — nightly recipes for the other two components**
- `cuda_bindings/docs/source/install.rst`: add a new **Development environment** section with an **Installing the latest nightly (top-of-tree builds)** subsection, between *Installing from Conda* and *Installing from Source*. Mirrors the cuda.core layout. The artifact pattern includes the CUDA major (`cuda13*`) because cuda-bindings wheels are CUDA-versioned.
- `cuda_pathfinder/docs/source/install.rst` (NEW): cuda.pathfinder previously had no install page. Adds a minimal install.rst (PyPI / Conda / Development environment → nightly / Source). Nightly uses the single pure-Python `cuda-pathfinder-wheel` artifact — no python/platform/CUDA selection needed.
- `cuda_pathfinder/docs/source/index.rst`: wire `install` into the toctree.

**Commit 2 — consistent nesting of "Installing from Source"**
- Across all three components, drop *Installing from Source* from `---` to `~~~` so it lives under *Development environment* as a sibling of the nightly / uv / pixi subsections (instead of as a top-level sibling of *Development environment*).
- For cuda.bindings, also drop its *Requirements* and *Editable Install* subsections from `~~~` to `^^^` so the depth still works.
- Bump SPDX copyright on the three touched existing files to `2025-2026`; the new cuda_pathfinder `install.rst` carries `2026`.

## Why

#2166 asks for the top-of-tree install recipe to be documented in the developer guide for all components. #2175 only covered cuda.core; this PR closes the gap for the other two and additionally unifies the section nesting so the developer-workflow group reads consistently across all three install pages.

## Tested

- `gh run list -R NVIDIA/cuda-python -w ci.yml -b main -s success -L1` resolves cleanly against the live repo.
- Local `sphinx-build` of all three docs sets (`cuda_core/docs`, `cuda_bindings/docs`, `cuda_pathfinder/docs`) succeeds with no new warnings on the changed/added install.rst files. Pre-existing autosummary warnings in cuda_pathfinder (the package isn't installed in the build env) are unchanged.
- Verified the cuda.bindings artifact name pattern against `ci/tools/env-vars` (`cuda-bindings-python{PY}-cuda{CUDA_VER}-{PLATFORM}-{SHA}`) and the cuda.pathfinder upload step (`cuda-pathfinder-wheel`, single artifact per run).

## Notes

- Content is intentionally additive in commit 1; commit 2 only changes underline characters and copyright years.
- Not auto-closing #2166 since you already closed it on #2175; happy to reopen-and-close on merge if you prefer.

### DIFF
--- a/cuda_bindings/docs/source/install.rst
+++ b/cuda_bindings/docs/source/install.rst
@@ -113,10 +113,10 @@ major version is built on ``main``; wheels for the prior CUDA major are
 published from the corresponding backport branch.
 
 Installing from Source
-----------------------
+~~~~~~~~~~~~~~~~~~~~~~
 
 Requirements
-~~~~~~~~~~~~
+^^^^^^^^^^^^
 
 * CUDA Toolkit headers[^1]
 * CUDA Runtime static library[^2]
@@ -138,7 +138,7 @@ See :doc:`Environment Variables <environment_variables>` for a description of ot
    Only ``cydriver``, ``cyruntime`` and ``cynvrtc`` are impacted by the header requirement.
 
 Editable Install
-~~~~~~~~~~~~~~~~
+^^^^^^^^^^^^^^^^
 
 You can use:
 
--- a/cuda_bindings/docs/source/install.rst
+++ b/cuda_bindings/docs/source/install.rst
@@ -80,6 +80,38 @@ For example:
 
    Tegra users can install the cuDLA conda package from conda-forge through ``conda install -c conda-forge libcudla cuda-version=13``, if it does not already exist on the system.
 
+Development environment
+-----------------------
+
+The sections above cover end-user installation. The section below focuses on
+a repeatable *development* workflow (editable installs and running tests).
+
+Installing the latest nightly (top-of-tree builds)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+These are useful for users looking to test new features or bug fixes prior to
+their inclusion in a release.
+
+CI publishes wheels as GitHub Actions artifacts on every push to ``main``. To
+obtain the most recent build, use the following commands:
+
+.. code-block:: console
+
+   $ # Find the latest successful CI run on main:
+   $ RUN_ID=$(gh run list -R NVIDIA/cuda-python -w ci.yml -b main -s success -L1 --json databaseId -q '.[0].databaseId')
+
+   $ # Download the wheel (pick your Python version and platform):
+   $ gh run download "$RUN_ID" -R NVIDIA/cuda-python -p "cuda-bindings-python312-cuda13*-linux-64-*"
+
+   $ # Install the downloaded wheel:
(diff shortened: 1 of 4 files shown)


leofang:
@rparolin this is ready

cuda_bindings/docs/source/install.rst
```diff
@@ -80,11 +80,43 @@ For example:
 
    Tegra users can install the cuDLA conda package from conda-forge through ``conda install -c conda-forge libcudla cuda-version=13``, if it does not already exist on the system.
 
+Development environment
+-----------------------
+
+The sections above cover end-user installation. The section below focuses on
+a repeatable *development* workflow (editable installs and running tests).
```
mdboom, line 87: What are the advantages of an editable install when it's almost 100% compiled code?  Personally, I find them more confusing/trouble than a regular install, since you have to recompile for virtually every change anyway.

cuda_bindings/docs/source/install.rst
```diff
@@ -80,11 +80,43 @@ For example:
...
+.. code-block:: console
+
+   $ # Find the latest successful CI run on main:
+   $ RUN_ID=$(gh run list -R NVIDIA/cuda-python -w ci.yml -b main -s success -L1 --json databaseId -q '.[0].databaseId')
+
+   $ # Download the wheel (pick your Python version and platform):
+   $ gh run download "$RUN_ID" -R NVIDIA/cuda-python -p "cuda-bindings-python312-cuda13*-linux-64-*"
+
+   $ # Install the downloaded wheel:
+   $ pip install cuda-bindings-python312-cuda13*-linux-64-*/cuda_bindings*.whl[all]
```
mdboom, line 107: It would be great to put these three lines in a script in `toolshed`.  Something like `toolshed/install-latest-wheel cuda_bindings`.

mdboom (approved):
My comments are just nice-to-haves / follow-ups.

cuda_bindings/docs/source/install.rst
leofang: For many Cython-based projects like CuPy, editable install works the same as pure Python projects, in that if a Cython module is not touched, it would not get re-compiled and local development is fast. However, my impression has been that it doesn't quite work for cuda-bindings for whatever reason. I am hoping that dropping Tempita might be the solution, but we'll see.

leofang:
Thanks, Mike!

cuda_bindings/docs/source/install.rst
leofang: I thought we do not want to document anything from `toolshed`?

PR #2192, opened 2026-06-10
Fix nvbug6084457: Make NVLINK_MAX_LINKS version-dependent

mdboom:
## Summary

This PR fixes `NvlinkInfo.max_links` being a static compile-time constant (`NVML_NVLINK_MAX_LINKS`) that overestimates the actual number of NVLink ports on a given device. Because not all devices support all links up to the macro ceiling, iterating over that range would attempt to query links the device doesn't have, producing incorrect behavior or errors.

### Changes

- **New `Device` APIs** (`get_nvlink_count`, `get_nvlinks`) — query the device-specific NVLink count at runtime via `nvmlDeviceGetFieldValues(NVML_FI_DEV_NVLINK_LINK_COUNT)`, and iterate over only the links actually present on that device.

- **`get_nvlink` validation fix** — the link-index range check now uses `get_nvlink_count()` instead of the static `NvlinkInfo.max_links`, so `ValueError` is raised precisely for links not supported by the specific device.

- **`NvlinkInfo.max_links` deprecated** — the class attribute now emits a `DeprecationWarning` (via a metaclass property) pointing users to `Device.get_nvlink_count()`. The static value is still returned for compatibility but callers should migrate.

- **Vendored `Deprecated` library** — a trimmed copy of the [`Deprecated`](https://pypi.org/project/Deprecated/) package (v1.3.1, MIT) is vendored under `cuda_core/cuda/core/_vendored/deprecated/` with the `wrapt` dependency removed. Used to emit `DeprecationWarning` with Sphinx-compatible docstring decorators (`@deprecated`, `@versionadded`, `@versionchanged`).

- **Tests** — `test_nvlink` updated to iterate via `get_nvlink_count()` and exercise `get_nvlinks()`; new `test_nvlink_max_links_deprecated` asserts the deprecation warning fires.

- **Release notes** — `1.1.0-notes.rst` updated with new APIs, the behaviour change to `get_nvlink`, and the deprecation of `NvlinkInfo.max_links`.

### DIFF
--- a/cuda_core/AGENTS.md
+++ b/cuda_core/AGENTS.md
@@ -157,11 +157,12 @@ Reviews should point out where existing public APIs are broken.
 - Changes should be notated in the code and also in the release notes in the
   "Deprecated APIs" section.
 
-**Annotating a new API** — Use the `deprecated.sphinx.versionadded` decorator:
+**Annotating a new API** — Use the `versionadded` decorator from the vendored
+`cuda.core._vendored.deprecated.sphinx` module:
 
 ```python
 
-from deprecated.sphinx import versionadded
+from cuda.core._vendored.deprecated.sphinx import versionadded
 
 @versionadded("1.2.0")
 def new_feature(...):
@@ -180,11 +181,12 @@ def new_feature(...):
     """
 ```
 
-**Annotating a changed API** — Use the `deprecated.sphinx.versionchanged` decorator:
+**Annotating a changed API** — Use the `versionchanged` decorator from the
+vendored `cuda.core._vendored.deprecated.sphinx` module:
 
 ```python
 
-from deprecated.sphinx import versionchanged
+from cuda.core._vendored.deprecated.sphinx import versionchanged
 
 @versionchanged("1.2.0", reason="The old version was broken because...")
 def new_feature(...):
@@ -205,13 +207,13 @@ def new_feature(...):
 ```
 
 **Deprecating an existing API** — use the `@deprecated` decorator from the
-`Deprecated` PyPI package (`deprecated.sphinx` import name) and add a
+vendored `cuda.core._vendored.deprecated.sphinx` module and add a
 `.. deprecated::` directive in the docstring.  The decorator emits a
 `DeprecationWarning` at call time; the docstring directive surfaces it in the
 generated docs.
 
 ```python
-from deprecated.sphinx import deprecated
+from cuda.core._vendored.deprecated.sphinx import deprecated
 
 @deprecated(version="1.2.0", reason="Use `new_feature` instead.")
 def old_feature(...):
--- a/cuda_core/AGENTS.md
+++ b/cuda_core/AGENTS.md
@@ -164,7 +164,7 @@ Reviews should point out where existing public APIs are broken.
 
 from cuda.core._vendored.deprecated.sphinx import versionadded
 
-@versionadded("1.2.0")
+@versionadded(version="1.2.0")
 def new_feature(...):
     """Short description.
     """
@@ -188,7 +188,7 @@ vendored `cuda.core._vendored.deprecated.sphinx` module:
 
 from cuda.core._vendored.deprecated.sphinx import versionchanged
 
-@versionchanged("1.2.0", reason="The old version was broken because...")
+@versionchanged(version="1.2.0", reason="The old version was broken because...")
 def new_feature(...):
     """Short description.
     """
--- a/cuda_core/AGENTS.md
+++ b/cuda_core/AGENTS.md
@@ -146,14 +146,17 @@ so that they are documented but don't appear in the main index.
 
 Reviews should point out where existing public APIs are broken.
 
-### Deprecation and API lifecycle
+### API lifecycle and deprecations
 
 `cuda.core` follows SemVer (see `docs/source/support.rst`):
 
-- **New APIs** may be added at any time (`x.Y.0`).
+- **New APIs** may be added at any time (`x.Y.0`).  They MUST have a
+  `@versionadded` decorator, unless the docstring formatting requires it to be
+  manually-specified.
 - **Breaking removals** only happen in **major releases** (`X.0.0`).
 - Per the support policy, a deprecation notice must be present for **at least
-  one minor release** before the API is actually removed.
+  one minor release** before the API is actually removed.  The deprecation notice
+  should use the `@deprecated` decorator, unless
 - Changes should be notated in the code and also in the release notes in the
   "Deprecated APIs" section.
 
@@ -215,7 +218,7 @@ generated docs.
 ```python
 from cuda.core._vendored.deprecated.sphinx import deprecated
 
-@deprecated(version="1.2.0", reason="Use `new_feature` instead.")
+@deprecated(version="1.2.0", reason="Use `new_feature` instead.")`
 def old_feature(...):
     """Short description.
     """
@@ -237,5 +240,9 @@ removing, verify that the deprecation has been present since at least the
 previous minor release.  Remove the decorator, the implementation, and any
 `__all__` entry; update `api.rst` and the release notes accordingly.
 
-At some point in the future, we will provide automation for removal of
-deprecated APIs.
+## Vendored code
+
+The `cuda/core/_vendored` directory contains vendored code from third-party
+sources.  It should not be modified, except to update the dependency or under
+exceptional circumstances, in which case any modifications should be clearly
+marked.
--- a/cuda_core/AGENTS.md
+++ b/cuda_core/AGENTS.md
@@ -148,7 +148,7 @@ Reviews should point out where existing public APIs are broken.
 
 ### Deprecation and API lifecycle
 
-`cuda.core` follows SemVer (see `docs/source/:
+`cuda.core` follows SemVer (see `docs/source/support.rst`):
 
 - **New APIs** may be added at any time (`x.Y.0`).
 - **Breaking removals** only happen in **major releases** (`X.0.0`).
--- a/cuda_core/AGENTS.md
+++ b/cuda_core/AGENTS.md
@@ -145,3 +145,95 @@ so that they are documented but don't appear in the main index.
 ### API stability
 
 Reviews should point out where existing public APIs are broken.
+
+### Deprecation and API lifecycle
+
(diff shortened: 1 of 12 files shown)


cuda_bindings/cuda/bindings/nvml.pyx
```diff
@@ -1593,7 +1593,11 @@ class FieldId(_FastEnum):
 
     MAX = 289
 
-NVLINK_MAX_LINKS = 18
+
+if tuple(int(x) for x in system_get_nvml_version().split(".")) < (3, 13):
```
rwgk: Could we obtain this value from the source of truth?

E.g. I see:

```
$ grep NVLINK_MAX_LINKS /usr/local/cuda_*/include/nvml.h
/usr/local/cuda_13.3.0_610.43.02_linux_kitpick035/include/nvml.h:#define NVML_NVLINK_MAX_LINKS 36 //!< Maximum number of NVLink links supported.
```

I have every single `/usr/local/cuda-*.*` from `12.0` to `13.3`, but I only see `NVML_NVLINK_MAX_LINKS` in the `13.3` include file.

Maybe something like this could work?

```
int internal_only_get_NVLINK_MAX_LINKS() {
#ifdef NVML_NVLINK_MAX_LINKS
    return NVML_NVLINK_MAX_LINKS;
#else
    return 18; // Prior to CUDA 13.3 this value was hard-wired. 
#endif
}
```

If there isn't a practical solution, could you please add a comment to explain?

mdboom: > I have every single /usr/local/cuda-*.* from 12.0 to 13.3, but I only see NVML_NVLINK_MAX_LINKS in the 13.3 include file.

Are you sure?  I see it in 12.9 through 13.3 (though the value changed in 13.3).

We have to build a single binary that works for every version 12.9 - 13.3, so the only way to do this is with a runtime computation.  Additionally, we build without the `nvml.h` header present.  I'll add a comment.

rwgk: Oh, sorry, I messed up like this (retrieved from my bash history; note the `cuda_`):

```
grep NVLINK_MAX_LINKS /usr/local/cuda_*/include/nvml.h
```

(I have a custom softlink for 13.3, which is why that matched.)

It looks much different like this:

```
$ grep NVLINK_MAX_LINKS /usr/local/cuda-*/include/nvml.h
/usr/local/cuda-12.0/include/nvml.h:#define NVML_NVLINK_MAX_LINKS 18
/usr/local/cuda-12.1/include/nvml.h:#define NVML_NVLINK_MAX_LINKS 18
/usr/local/cuda-12.2/include/nvml.h:#define NVML_NVLINK_MAX_LINKS 18
/usr/local/cuda-12.3/include/nvml.h:#define NVML_NVLINK_MAX_LINKS 18
/usr/local/cuda-12.4/include/nvml.h:#define NVML_NVLINK_MAX_LINKS 18
/usr/local/cuda-12.5/include/nvml.h:#define NVML_NVLINK_MAX_LINKS 18
/usr/local/cuda-12.6/include/nvml.h:#define NVML_NVLINK_MAX_LINKS 18
/usr/local/cuda-12.8/include/nvml.h:#define NVML_NVLINK_MAX_LINKS 18
/usr/local/cuda-12.9/include/nvml.h:#define NVML_NVLINK_MAX_LINKS 18
/usr/local/cuda-13.0/include/nvml.h:#define NVML_NVLINK_MAX_LINKS 18
/usr/local/cuda-13.1/include/nvml.h:#define NVML_NVLINK_MAX_LINKS 18 //!< Maximum number of NVLink links supported.
/usr/local/cuda-13.2/include/nvml.h:#define NVML_NVLINK_MAX_LINKS 18 //!< Maximum number of NVLink links supported.
/usr/local/cuda-13.3/include/nvml.h:#define NVML_NVLINK_MAX_LINKS 36 //!< Maximum number of NVLink links supported.
```

rwgk: I didn't think it through before, but at second look, of course the helper function I was envisioning would need to live in `nvml.h` itself.

I had my agent do a quick check in CUDA 13.3's nvml.h and the NVML docs. It didn't surface any runtime C API that reports NVML_NVLINK_MAX_LINKS. Your solution does indeed appear to be our only option. Ideally we'd request a runtime API, but that's for another day.

rwgk (approved):

cuda_bindings/cuda/bindings/nvml.pyx
leofang: This is a module-level constant so there's not much we can do with a helper function. @mdboom that said, this would force nvml to be loaded at import time, which breaks CPU-only envs. Can we hide this behind module `__getattr__`

leofang: (For example this would break [the import test](https://github.com/conda-forge/cuda-python-feedstock/blob/5c53a98d7d3deefb1c127ad4389b75ddfcb34c50/recipe/recipe.yaml#L69).)

mdboom: Experimentally, it doesn't break CPU-only builds, but it does break on systems without NVML installed, so I agree the `__getattr__` trick is probably justified, but it's narrower than you think.  `nvmlSystemGetVersion` doesn't require `nvmlInit` and doesn't require a GPU.

leofang: CPU-only envs have a broad definition, including but not limited to: GPU driver is not installed 😉

mdboom:
@leofang: There seem to be failures on specific hardware (a100, t4) on Windows only that we can't use all of the stated nvlinks.  I have asked internally why this might be the case.  It /shouldn't/ be a driver vs. CTK difference since NVML ships with the driver and we are asking NVML (not the CTK or cuda-bindings itself or something) what version we have to determine how many links we have.  But check my work, it could be that I'm not checking the appropriate thing.

mdboom:
As discussed in the meeting this has been changed so that:

- The `deprecated` library is vendored.  Unfortunately, dependabot doesn't support vendored libraries for Python, so it is on us to keep it updated, but this is not a big or complex dependency.
- Getting `NVML_NVLINK_MAX_LINKS` (the header macro) to work correctly has been given up on and its usage in cuda.core deprecated.  The recommended alternative is the device-specific field lookup implemented here.

rwgk:
codex gpt-5.5 had two medium findings and one low finding. I asked it to create commits with suggested fixes, which is easier these days than explaining:

* https://github.com/rwgk/cuda-python/commit/44e0197cab3349ac9509edcb94b83a869bbcedb4

* https://github.com/rwgk/cuda-python/commit/d0606ff1c851dc5e608113efcd396673f045b6fb

* https://github.com/rwgk/cuda-python/commit/b8eded275b451344bc7664149cbfbc2504bd4b38

rwgk:
It'd be nice to update the PR description.

mdboom:
> https://github.com/rwgk/cuda-python/commit/44e0197cab3349ac9509edcb94b83a869bbcedb4

This one seems out of date

> https://github.com/rwgk/cuda-python/commit/d0606ff1c851dc5e608113efcd396673f045b6fb

This has been cherry-picked

> https://github.com/rwgk/cuda-python/commit/b8eded275b451344bc7664149cbfbc2504bd4b38

This one is wrong.  It shouldn't be updating the vendored dependency.

> It'd be nice to update the PR description.

Sure.

cuda_core/cuda/core/system/_nvlink.pxi
```diff
@@ -18,7 +18,24 @@ if _NVLINK_VERSION_6_0 is not None:
...
+    )
+    def max_links(cls):
+        """
+        The statically-defined maximum number of Nvlinks available.  Defined in
+        upstream NVML as ``NVML_NVLINK_MAX_LINKS``.
+
+        To find the actual number of Nvlinks available on a device, use
+        :py:attr:`Device.get_nvlink_count`.
+        """
+        return nvml.NVLINK_MAX_LINKS
```
mdboom, line 36:
```suggestion
        # This will always return 18, even on CTK 13.3 where it should be 36.
        return nvml.NVLINK_MAX_LINKS
```

kkraus14:
@mdboom is there a brief explanation as far as why we chose to vendor the `Deprecated` library and remove the `wrapt` dependency from it?

cuda_core/AGENTS.md
```diff
@@ -145,3 +145,97 @@ so that they are documented but don't appear in the main index.
...
+- The deprecated API must remain in the codebase for **at least one full minor
+  release cycle** before it can be removed in a subsequent major release.
+
+**Removing a deprecated API** — removals land in a **major release**.  Before
+removing, verify that the deprecation has been present since at least the
+previous minor release.  Remove the decorator, the implementation, and any
+`__all__` entry; update `api.rst` and the release notes accordingly.
+
+At some point in the future, we will provide automation for removal of
+deprecated APIs.
```
rwgk: This seems purely informational and speculative, not actionable for an agent. Do we need it here?

cuda_core/AGENTS.md
```diff
@@ -145,3 +145,97 @@ so that they are documented but don't appear in the main index.
 ### API stability
 
 Reviews should point out where existing public APIs are broken.
+
+### Deprecation and API lifecycle
```
rwgk: This is primarily about the API lifecycle, deprecation is just one aspect:

```
### API lifecycle and deprecation
```

would seem more fitting.

I assume the decorators are additive? We should spell that out explicitly, so agents know what we expect them to do.

Should we also add explicitly that agents remind us to systematically add the `versionadded` and `versionchanged` decorators in the future? I imagine this will easily slip if there is no automated mechanism. Less desirable (IMO) alternative: spell out that the `versionadded` and `versionchanged` decorators are optional, worded in a way that ensures we achieve uniform behavior over different agents or agent versions.

mdboom:
> @mdboom is there a brief explanation as far as why we chose to vendor the Deprecated library and remove the wrapt dependency from it?

We chose to vendor because consensus in the meeting was that any dependency has a cost, especially when our users may be installing cuda_core alongside other libraries that may have a different version range for `Deprecated`, causing version conflicts out of our control.  The wrapt dependency was a single function that wasn't providing much value, so it was simpler to remove it rather than vendor it as well.

PR #2194, opened 2026-06-10
Test using pytest-run-parallel and related fixups in the tests

seberg:
## Description

Enable testing with `pytest-run-parallel` on CI for 3.14t and 3.15t as well some residual related fixes.

## Checklist

### DIFF
--- a/ci/tools/run-tests
+++ b/ci/tools/run-tests
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # SPDX-License-Identifier: Apache-2.0
 
@@ -20,6 +20,14 @@ fi
 
 test_module=${1}
 
+FREE_THREADING=""
+PYTEST_PARALLEL_ARGS=()
+if python -c 'import sys; assert not sys._is_gil_enabled()' 2> /dev/null; then
+  FREE_THREADING="-ft"
+  PYTEST_PARALLEL_ARGS=(--parallel-threads=4)
+  pip install pytest-run-parallel
+fi
+
 # For standard modes, install pathfinder up front (it is a direct dependency
 # of bindings, and a transitive dependency of core).  Nightly modes install
 # all wheels together in a single pip call further below.
@@ -36,7 +44,7 @@ if [[ "${test_module}" == "pathfinder" ]]; then
     "LD:${CUDA_PATHFINDER_TEST_LOAD_NVIDIA_DYNAMIC_LIB_STRICTNESS} " \
     "FH:${CUDA_PATHFINDER_TEST_FIND_NVIDIA_HEADERS_STRICTNESS} " \
     "BC:${CUDA_PATHFINDER_TEST_FIND_NVIDIA_BITCODE_LIB_STRICTNESS}"
-  pytest -ra -s -v --durations=0 tests/ |& tee /tmp/pathfinder_test_log.txt
+  pytest -ra -s -v --durations=0 "${PYTEST_PARALLEL_ARGS[@]}" tests/ |& tee /tmp/pathfinder_test_log.txt
   # Report the number of "INFO test_" lines (including zero)
   # to support quick validations based on GHA log archives.
   line_count=$(awk '/^INFO test_/ {count++} END {print count+0}' /tmp/pathfinder_test_log.txt)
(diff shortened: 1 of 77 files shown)


seberg:
/ok to test [ed40f60](https://github.com/NVIDIA/cuda-python/pull/2194/commits/ed40f6038cc5e6e774f9587bf559e5c9146b071f)

seberg:
Fun, the refactor made the cufile xfail-strict tests pass on CI, but I didn't set up the parallel run correctly... one more try:

/ok to test [eb6a2ff](https://github.com/NVIDIA/cuda-python/pull/2194/commits/eb6a2ff4313b3483f16c40366098708ebb1ac125)

seberg:
/ok to test [eb6a2ff](https://github.com/NVIDIA/cuda-python/pull/2194/commits/eb6a2ff4313b3483f16c40366098708ebb1ac125)

seberg:
/ok to test [7b59bff](https://github.com/NVIDIA/cuda-python/commit/7b59bffb66889a9d0aae7806a8eb2ae174d3a644)

seberg:
Still quite a few changes in this now (and let's see if CI passes) including gh-2321.  But maybe it's still fine to review.

The most annoying one is probably the `test_object_protocols.py` changes.  IIRC the problem was that the design with the `.getfixturevalue()` deferred things into threads, which might even have been fine with initialized.
But none of the tests actually test things deep enough to require cuda set up in the test itself.

ci/tools/run-tests
```diff
@@ -20,6 +20,14 @@ fi
 
 test_module=${1}
 
+FREE_THREADING=""
+PYTEST_PARALLEL_ARGS=()
+if python -c 'import sys; assert not sys._is_gil_enabled()' 2> /dev/null; then
+  FREE_THREADING="-ft"
+  PYTEST_PARALLEL_ARGS=(--parallel-threads=4)
+  pip install pytest-run-parallel
```
Andy-Jost: Can you add this to the test groups in the normal places under `cuda_{bindings,core,pathfinder}/pyproject.toml`?

Andy-Jost (changes requested):
This needs another review round after #2454 merges and it is rebased/merged with main

cuda_core/tests/conftest.py
```diff
@@ -72,44 +71,16 @@ def pytest_terminal_summary(terminalreporter):
...
-        # before the next test runs. Without this, a memory pool's VA
-        # reservation is not returned until both have happened, and per-test
-        # leftovers accumulate across the run -- which is how full-suite runs
-        # can exhaust address space and hit CUDA_ERROR_OUT_OF_MEMORY on a
-        # device with plenty of free physical memory (issue #2381). gc.collect()
-        # must run first: cuCtxSynchronize alone cannot drain frees that were
-        # never enqueued because their owning object had not been collected yet.
-        gc.collect()
-        driver.cuCtxSynchronize()
-        _ = _device_unset_current()
```
seberg, line 102: This was causing issues for some graph tests (because a device sync disrupts graph capture).

But, actually, it also seems to fix the `mr` cleanup issue...  I.e. the dead-lock I was seeing happens only if there is a device synchronization on each thread.

That should allow a bunch of cleanups, but I would prefer to follow-up for it.

cuda_core/tests/test_object_protocols.py
```diff
@@ -751,30 +751,31 @@ def sample_object_b(request):
 
 
 @pytest.mark.agent_authored(model="gpt-5.6")
-@pytest.mark.parametrize("fixture_name", CLOSEABLE_TYPES)
-def test_closeable_object_state_and_safe_inspection(fixture_name, request):
+@pytest.mark.thread_unsafe(reason="closes a fixture object shared between threads")
```
seberg, line 754: Just marked as unsafe here, but also changed to follow the same indirect fixture pattern as the rest of the file.

cuda_core/cuda/core/_stream.pyx
```diff
@@ -596,7 +596,6 @@ cdef cydriver.CUstream _handle_from_stream_protocol(obj) except*:
         info = cuda_stream_attr()
     else:
         info = cuda_stream_attr
-        warnings.simplefilter("once", DeprecationWarning)
```
seberg, line 600: This triggered a failure.  Could also just mark the test as unsafe, but mutating global state here seems wrong (e.g. what if the user wants to raise `DeprecationWarning`s?).

ci/tools/run-tests
seberg: Makes sense, I added it as a `test-ft` but it currently only contains `pytest-run-parallel`.  Happy to do it differently.

(Semi-related, removed the special logic for cupy not support free-threaded as it does now.)

seberg (commented):
(Few comments for review 🤞 that it will run through successfully in CI with all the drift going on.)

seberg:
Ping @Andy-Jost for awareness.  Not burning, but would love to move this in so that the test fixups don't keep diverging (hopefully the amount of new tests needing skipping isn't too bad; on the plus side, a lot of the fixups here would have been cargo-culted by agents already if it had been merged at the time).

Andy-Jost (changes requested):
Fable 5.1 points out one correctness issue: `thread_unsafe_on_windows` in `cuda_core/tests/helpers/buffers.py` returns the mark without applying it to `func`. On Windows/WSL the decorated name becomes a `MarkDecorator`, so pytest silently skips collection. The Windows CI logs for this PR show 21 warnings of the form:

```
PytestCollectionWarning: cannot collect 'test_options_none_copy_to_data_correct' because it is not a function.
```

Decorated classes (`TestCopyBatchCore`, `TestCopyBatchEquivalence`, `TestCopyBatchOptions`) disappear with no warning at all. Windows collects 4053 items vs 4202 on Linux, on both the 3.14t and 3.13 jobs.

The bug is pre-existing from #2454, but this PR adds ~33 more affected tests. One-line fix:

```python
        return pytest.mark.thread_unsafe(reason="...")(func)
```

Since these tests haven't actually run on Windows since #2454, please confirm the Windows jobs still pass once they do.

seberg:
Ouch, good spot :/, I'll check the CI log carefully when it's done!

Andy-Jost (approved):
LGTM!

PR #2195, opened 2026-06-10
CI: require N consecutive nvidia-smi successes after Windows device cycle

leofang:
## Description

Fix flakiness on the 2× H100 MCDM Windows runner, observed since #2176 landed (e.g. <https://github.com/NVIDIA/cuda-python/actions/runs/27293276910/job/80661578895>).

*What broke*: pre-#2176, every Windows row ran `install_gpu_driver.ps1` unconditionally, and that script ended with a fixed `Start-Sleep -Seconds 5` after the `pnputil` cycle. #2176 replaced that fixed 5-sec settle with a poll-until-success loop in `configure_driver_mode.ps1` — and that loop exits on the *first* `nvidia-smi` exit-0, ~2 sec into the cycle. On the H100 pair, NVML briefly reports success mid-init and then flaps back to "Not Found" a few seconds later, so the workflow's next "Ensure GPU is working" step lands in the flap window:

```
Waiting for nvidia-smi/NVML to come back up after device cycle...
+ nvidia-smi                    # next workflow step, ~4 sec later
Failed to initialize NVML: Not Found
Write-Error: Switching to driver mode MCDM failed!
```

*Fix*: two layers.

1. *Restore the pre-#2176 5-sec floor* — unconditional `Start-Sleep -Seconds 5` *before* the poll. That's the known-good baseline that worked across all Windows rows for a long time.
2. *Then require N consecutive successes in the poll*, where `N == Get-PnpDevice -Class Display -FriendlyName "NVIDIA*"`. On single-GPU rows that's still effectively one short success after the 5-sec floor. On the H100 pair it's two consecutive 3-sec-apart successes, so a mid-flap "ok" can't fool the loop. The 60-sec deadline cap is unchanged.

Net effect:
- *Single-GPU rows*: 5 sec sleep + ~3 sec (1 success) = ~8 sec floor (pre-#2176 was 5 sec, plus the install overhead).
- *2× H100 MCDM*: 5 sec sleep + ~6 sec (2 successes) = ~11 sec floor (pre-#2176 was 5 sec).
- *Pathological case*: 60-sec deadline still caps total wait, with a loud failure message instead of a silent ride into the next step.

## Checklist

### DIFF
--- a/ci/tools/configure_driver_mode.ps1
+++ b/ci/tools/configure_driver_mode.ps1
@@ -31,7 +31,6 @@ function Set-DriverMode {
     }
 
     # Only restart NVIDIA display adapters, not other display devices (e.g. QEMU VGA).
-    # @(...) forces an array even when there's a single device, so .Count works.
     $nvidia_devices = @(Get-PnpDevice -Class Display -FriendlyName "NVIDIA*")
     $gpu_count = $nvidia_devices.Count
     foreach ($device in $nvidia_devices) {
--- a/ci/tools/configure_driver_mode.ps1
+++ b/ci/tools/configure_driver_mode.ps1
@@ -40,20 +40,11 @@ function Set-DriverMode {
         pnputil /enable-device "$($device.InstanceId)"
     }
 
-    # Restore the unconditional 5-sec settle that the pre-#2176
-    # install_gpu_driver.ps1 had after the cycle. With #2176, on
-    # DRIVER=latest rows we no longer run an install (which previously
-    # acted as additional warm-up), so the floor had effectively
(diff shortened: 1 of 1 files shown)


leofang:
/ok to test [6067a78](https://github.com/NVIDIA/cuda-python/pull/2195/commits/6067a7871c68b0919b669f10ea777eea042c7f44)

leofang:
/ok to test [4909e11](https://github.com/NVIDIA/cuda-python/pull/2195/commits/4909e110f3bc2819c18ce5ff95410d31537eba59)

rwgk (approved):

leofang:
Thanks, Ralf!

PR #2196, opened 2026-06-10
feat(pathfinder): add CTK-root canary fallback for NVIDIA binaries

### LINKED ISSUE #2119: find_nvidia_binary_utility() violates deterministic-search contract on Windows (CWD leakage via shutil.which)
## Summary

`cuda.pathfinder.find_nvidia_binary_utility()` currently delegates to
`shutil.which(name, path=trusted_dirs)` after assembling a list of trusted
search directories (NVIDIA wheel `bin/`, `CONDA_PREFIX`, `CUDA_PATH` /
`CUDA_HOME`, etc.).

On Windows, CPython's `shutil.which()` may prepend the **current working
directory** to executable resolution even when the caller supplies an
explicit `path=`. As a result, a binary located in the process CWD can be
returned in preference to the trusted CUDA / Conda / wheel binary that
pathfinder is supposed to discover.

This violates the pathfinder contract — callers expect a deterministic
lookup against a documented, bounded set of trusted roots, not a
shell-style lookup that depends on ambient process state.

## Reproduction (conceptual)

1. On Windows, install `nvcc` via the NVIDIA wheel (so it lives under a
   `site-packages/.../bin` directory pathfinder considers trusted).
2. Set CWD to a directory containing a different `nvcc.exe` (e.g.
   `%CONDA_PREFIX%\Library\bin`, or any attacker-writable location).
3. Call `cuda.pathfinder.find_nvidia_binary_utility("nvcc")`.
4. **Expected:** the wheel-resolved `nvcc.exe` (deterministic, contract-
   honoring).
   **Actual:** the CWD `nvcc.exe` is returned because `shutil.which`
   prepended CWD ahead of the supplied `path=`.

Leo's framing (#2111 comment):
> Assuming a pathological case where my CWD is `%CONDA_PREFIX%\Library\bin`
> [...] Following the pathfinder contract I would expect to get the latter
> NVCC. But because `shutil.which` prepends CWD (TIL!), I would actually
> get the former.

## Impact

- **Contract correctness:** the public API silently returns a binary that
  was never in the documented trust set.
- **Security-adjacent:** a planted `nvcc.exe` (or other supported utility)
  in an attacker-controlled CWD is executed downstream by callers that
  invoke the returned path. Originally flagged by an internal security
  scan.
- **Cross-platform inconsistency:** Linux/macOS do not exhibit the CWD
  prepend, so behavior diverges by platform in a way users cannot predict.

## Root cause

`shutil.which()` is a "what would the shell run?" primitive, not a "search
exactly these directories" primitive. It carries ambient OS/Python policy:

- Windows: CWD prepended in certain Python versions (observed on 3.12).
- `PATHEXT` expansion is implicit and version-dependent.
- Empty/relative entries in `path=` become CWD-sensitive.
- Return value is not guaranteed to be an absolute, normalized path.

None of these are appropriate for a deterministic discovery API.

## Proposed direction

Replace the `shutil.which` delegation with an explicit, bounded directory
walk (see PR #2111 for a first cut). Key properties:

- Search **only** directories pathfinder constructs:
  1. NVIDIA Python wheel `bin/` under `site-packages`
  2. `CONDA_PREFIX` bin layout
  3. Explicit `CUDA_PATH` / `CUDA_HOME`
  4. CTK root resolved via the existing canary mechanism, then CTK `bin`
- Preserve Windows extension normalization (`.exe`, `PATHEXT` handling
  done explicitly, not inherited from the environment).
- Preserve Unix executable-bit checks.
- Return absolute, normalized paths.
- Do **not** consult ambient `PATH` or CWD by default.
- Document the search order and the non-inclusion of `PATH`/CWD.

## Open design question — ambient `PATH`

Keith raised (#2111 comment) that the official Linux CUDA install guide
instructs users to add the CTK `bin/` to `PATH`:
https://docs.nvidia.com/cuda/cuda-installation-guide-linux/#environment-setup

Users who follow that guide — setting `LD_LIBRARY_PATH` for libraries and
`PATH` for executables, without setting `CUDA_PATH`/`CUDA_HOME` — would
need to set an explicit env var for pathfinder to discover their CTK
binaries once `shutil.which` is removed.

Options to weigh:

1. **No `PATH` fallback** (PR #2111 as written). Maximally deterministic;
   requires affected users to set `CUDA_PATH`/`CUDA_HOME`. Cleanest
   contract.
2. **`PATH` fallback on Linux/macOS only.** Sidesteps the Windows CWD
   issue (the actual bug) while preserving the documented install path
   for the platform where the docs apply. Asymmetric but pragmatic.
3. **Opt-in `PATH` search** via parameter or dedicated env var (e.g.
   `CUDA_PATHFINDER_ALLOW_PATH=1`). Explicit, non-default, auditable.

Leo's call requested before landing the fix.

## Regression coverage

A test that plants a Windows CWD binary and asserts it is **not** returned
is required (already drafted in PR #2111).

## References

- PR: #2111
- Affected file: `cuda_pathfinder/cuda/pathfinder/_binaries/find_nvidia_binary_utility.py`
- Linux install guide: https://docs.nvidia.com/cuda/cuda-installation-guide-linux/#environment-setup

aryanputta:
Fixes #2119, with one accepted caveat: issue 2119 explicitly lists "do not consult ambient PATH or CWD" as a desired property. However, the added CTK-root canary fallback (see below) inherits the existing platform-native dynamic-loader semantics. In particular, the current Windows `LoadLibraryExW(..., 0)` path may include `PATH`.

## Summary

This PR removes `shutil.which` from NVIDIA binary discovery. `shutil.which` is a shell-oriented executable lookup primitive and can apply platform-specific search behavior outside Pathfinder's configured roots. In particular, on Windows it can prepend the process current working directory even when an explicit `PATH=` is supplied, allowing an unrelated binary to shadow the CUDA, Conda, or wheel binary Pathfinder was intended to find.

Replacing shutil.which with only explicit directory probes would close that loophole, but it could also make previously discoverable CUDA Toolkit binaries unavailable to Pathfinder. To preserve the user experience in most environments, this PR extends `find_nvidia_binary_utility()` with the existing CTK-root canary mechanism. If the configured binary locations all miss, Pathfinder resolves `cudart` through the existing dynamic-library discovery machinery in an isolated subprocess, derives the CUDA Toolkit root from the resolved library path, and searches that root's platform-specific `bin` layout.

## Discovery order

1. NVIDIA wheel `bin/` directories from site-packages.
2. The active Conda environment from `CONDA_PREFIX`.
3. The CUDA Toolkit root from `CUDA_HOME` or `CUDA_PATH`.
4. CTK-root canary fallback: resolve `cudart` through the OS dynamic loader,
   derive its toolkit root, and search that root's `bin` directories.

## Behavior and compatibility

The executable finder no longer delegates to `shutil.which`, directly searches the ambient `PATH`, or implicitly prepends the caller's current working directory. The canary fallback does, however, inherit the platform's existing dynamic-library search semantics. It may be influenced by `LD_LIBRARY_PATH` on Linux and by the native DLL search path on Windows; in particular, the current Windows `LoadLibraryExW(..., 0)` path may indirectly consult `PATH`. This PR does not change those loader semantics.

The fallback is intentionally not identical to searching `PATH` for the executable. When multiple CUDA Toolkit installations are visible, Pathfinder returns the binary associated with the toolkit whose `cudart` is selected by the dynamic loader, which may differ from the binary a shell would select.

For most users, toolkit discovery should continue to work without additional configuration. Because the resolution behavior changes subtly in multi-toolkit environments, this change is released with the `cuda-pathfinder` 1.6 minor version rather than as a patch release.

## Tests

- Reworked search-path tests to assert deterministic probe order.
- Added CWD-isolation coverage for the core #2119 regression.
- Added first-match, empty/duplicate directory, and POSIX non-executable rejection coverage.
- Added CTK-root canary fallback coverage, including Windows `bin/x64`, `bin/x86_64`, and `bin` layout.
- Added coverage that the canary is not consulted when an earlier trusted directory hits.
- Added focused tests for shared `resolve_ctk_root_via_canary(...)`.

Local checks: `ruff check` and `ruff format --check` pass on the touched files.


rwgk:
Hi @aryanputta, this is great, but in isolation it'll probably break some users. This was pointed out by @kkraus14 here before:

* https://github.com/NVIDIA/cuda-python/pull/2111#issuecomment-4493908367

> If someone follows this installation guide and sets `LD_LIBRARY_PATH` for libraries and `PATH` for executables ...

I ran the context and my thoughts by GPT around the same time, which lead to the conclusion at the end of this comment:

* https://github.com/NVIDIA/cuda-python/pull/2111#issuecomment-4492659662

> Use a custom bounded finder, add CTK-root canary fallback, and do not implicitly search `PATH`.

I still think that's the right compromise:
* We will find the binary utility via `LD_LIBRARY_PATH` if set. It's still an attack vector, but significantly more difficult to exploit than hacking `PATH`.
* Users following the installation guide will not see a difference in most cases, and get a more self-consistent outcome in the rare case that the searches via `PATH` and `LD_LIBRARY_PATH` produce diverging results.

The behavior difference is significant enough though that I'd want to bump the pathfinder minor version, which we can do; I don't have concerns about that.

How do you feel about expanding this PR to fold in the CTK-root canary fallback? — I believe the code changes will still be quite modest.

aryanputta:
Thanks @rwgk, that compromise makes sense and I've folded it in (pushed `c3d39f0`, no force-push). Agreed on the reasoning: the bounded finder shouldn't regress users who follow the Linux install guide and only set `LD_LIBRARY_PATH` + `PATH`, but reaching them via `LD_LIBRARY_PATH` instead of `PATH` is the safer trade.

**What the new commit adds**

Search order is unchanged for the explicit trusted dirs, with the canary as a strict fallback:

1. NVIDIA wheel `bin/` (site-packages)
2. `CONDA_PREFIX`
3. `CUDA_HOME` / `CUDA_PATH`
4. **CTK-root canary** (new): resolve `cudart` through the OS dynamic loader, derive the CTK root from its absolute path, search that root's `bin` layout

Key points:
- The canary runs **only** when steps 1-3 miss, so the common wheel/conda/`CUDA_HOME` cases never spawn the resolver subprocess.
- It reuses the existing canary toolbox. I factored the `cudart` -> CTK-root resolution into `resolve_ctk_root_via_canary(...)` in `load_nvidia_dynamic_lib.py` and had the existing dynamic-library `_try_ctk_root_canary` call it too, so lib and binary discovery share one path. The resolver runs in the isolated subprocess that honors `LD_LIBRARY_PATH` on Linux and the native DLL search on Windows.
- `PATH` and the process CWD are still never consulted, so the `LD_LIBRARY_PATH` surface is the only added vector, exactly as you described.

**Version**

Added `docs/source/release/1.6.0-notes.rst` for the minor bump (version is tag-driven, so this is the doc side of it). Happy to renumber if you'd rather it land under a different version.

**Tests**

Added coverage for the fallback in `tests/test_find_nvidia_binaries.py` (found via canary, ordering vs `CUDA_HOME`, Windows `bin/x64`/`bin/x86_64`/`bin` layout, not consulted when an earlier dir hits, cached) and a focused `resolve_ctk_root_via_canary` unit test in `tests/test_ctk_root_discovery.py`. `ruff check` and `ruff format --check` are clean.

cuda_pathfinder/cuda/pathfinder/_binaries/find_nvidia_binary_utility.py
```diff
@@ -3,13 +3,17 @@
...
 
 from cuda.pathfinder._binaries import supported_nvidia_binaries
 from cuda.pathfinder._utils.env_vars import get_cuda_path_or_home
 from cuda.pathfinder._utils.find_sub_dirs import find_sub_dirs_all_sitepackages
 from cuda.pathfinder._utils.platform_aware import IS_WINDOWS
 
+# CUDA Toolkit canary library used to derive the toolkit root when it is only
+# visible through the dynamic loader. ``cudart`` always ships with the CTK and
+# matches the anchor used by the dynamic-library CTK-root canary flow.
+_CTK_ROOT_CANARY_ANCHOR_LIBNAME = "cudart"
```
rwgk: Headers, dynamic-library descriptors, and now binary discovery all rely on the same `"cudart"` choice. Could you please centralize the anchor in a neutral shared module, e.g. `_utils/ctk_root_canary.py`?

I think this will go well with the existing code:

```
CTK_ROOT_CANARY_ANCHOR_LIBNAMES = ("cudart",)
```

cuda_pathfinder/cuda/pathfinder/_binaries/find_nvidia_binary_utility.py
```diff
@@ -28,6 +32,71 @@ def _normalize_utility_name(utility_name: str) -> str:
...
+    if IS_WINDOWS:
+        return True
+    return os.access(path, os.X_OK)
+
+
+def _ctk_bin_subdirs(root: str) -> list[str]:
+    """Return the bin directories to search under a CUDA Toolkit ``root``.
+
+    On Windows the CTK ships binaries under ``bin/x64`` (CTK 13), ``bin/x86_64``,
+    and ``bin`` (CTK 12); on Linux they live in ``bin``.
```
rwgk: Please remove: it essentially just explains the code immediately below, which is actually easier to read than this prose; and it's a private helper function.

cuda_pathfinder/cuda/pathfinder/_binaries/find_nvidia_binary_utility.py
```diff
@@ -28,6 +32,71 @@ def _normalize_utility_name(utility_name: str) -> str:
...
+    return [os.path.join(root, "bin")]
+
+
+def _resolve_ctk_root_via_canary() -> str | None:
+    """Derive the CUDA Toolkit root from the ``cudart`` canary library.
+
+    ``cudart`` is resolved by the OS dynamic loader, which honors
+    ``LD_LIBRARY_PATH`` on Linux and the native DLL search on Windows, and the
+    toolkit root is derived from its absolute path. The ambient ``PATH`` is
+    never consulted. The loader module is imported lazily to avoid pulling the
```
rwgk: Please remove the first two sentences here: they only explain code elsewhere; for a private helper function that's more noise than signal.

Also throughout: Please avoid all mentions like "unlike shutil.which" and "ambient PATH". That only makes sense in the context of this particular PR, but will be puzzling later. Explaining the which/PATH/CWD pitfalls is suitable in the PR description and release notes, but not in code comments (maybe in general terms, in one or two places).

aryanputta:
Quick check-in: is there anything else you would like changed here after the CTK-root canary fallback update, or is this mainly waiting on review / runner validation now? Happy to adjust if the version note or fallback behavior should be shaped differently.

cuda_pathfinder/cuda/pathfinder/_binaries/find_nvidia_binary_utility.py
```diff
@@ -28,6 +32,71 @@ def _normalize_utility_name(utility_name: str) -> str:
...
+    dynamic-library machinery in at import time.
+    """
+    from cuda.pathfinder._dynamic_libs.load_nvidia_dynamic_lib import resolve_ctk_root_via_canary
+
+    ctk_root: str | None = resolve_ctk_root_via_canary(_CTK_ROOT_CANARY_ANCHOR_LIBNAME)
+    return ctk_root
+
+
+def _resolve_in_trusted_dirs(normalized_name: str, dirs: list[str]) -> str | None:
+    """Resolve ``normalized_name`` against ``dirs`` only, in order.
```
rwgk:
```
Resolve ``normalized_name`` against ``dirs`` in order.
```

("only" will be puzzling later)

cuda_pathfinder/cuda/pathfinder/_binaries/find_nvidia_binary_utility.py
```diff
@@ -28,6 +32,71 @@ def _normalize_utility_name(utility_name: str) -> str:
...
+
+    Unlike ``shutil.which``, this never consults the current working directory
+    or the ambient ``PATH``. On Windows ``shutil.which`` prepends the process
+    CWD to the search even when an explicit ``path=`` is supplied, which lets a
+    binary sitting in an arbitrary CWD shadow the trusted CUDA / Conda / wheel
+    binary that pathfinder is contracted to discover. Searching the trusted
+    directories explicitly keeps the lookup deterministic and bounded.
```
rwgk: See above: please delete this entire paragraph.

cuda_pathfinder/cuda/pathfinder/_binaries/find_nvidia_binary_utility.py
```diff
@@ -28,6 +32,71 @@ def _normalize_utility_name(utility_name: str) -> str:
...
+    """
+    seen: set[str] = set()
+    for directory in dirs:
+        if not directory or directory in seen:
+            continue
```
rwgk: Skipping over empty strings can easily mask bugs (i.e. string slicing or split bug). `None` doesn't make sense here, too.

A very easy way to be stricter/safer:

```
        if directory in seen:
            continue
        assert directory
```

cuda_pathfinder/cuda/pathfinder/_binaries/find_nvidia_binary_utility.py
```diff
@@ -65,6 +134,15 @@ def find_nvidia_binary_utility(utility_name: str) -> str | None:
...
              or just ``bin`` on Linux.
 
+        4. **CTK-root canary fallback**
+
+           - Only when steps 1-3 miss: resolve the ``cudart`` library through the
+             OS dynamic loader (which honors ``LD_LIBRARY_PATH`` on Linux and the
+             native DLL search on Windows), derive the CUDA Toolkit root from it,
+             and search that root's bin layout. This finds the utility for users
+             who follow the CUDA install guide and set ``LD_LIBRARY_PATH`` for
+             libraries without also setting ``CUDA_HOME`` / ``CUDA_PATH``.
```
rwgk: I think it's better to delete the last sentence: `LD_LIBRARY_PATH` is mentioned in the previous sentence already, and the second part will just make people wonder unnecessarily/distractingly ("why would one do that?").

cuda_pathfinder/cuda/pathfinder/_binaries/find_nvidia_binary_utility.py
```diff
@@ -73,6 +151,10 @@ def find_nvidia_binary_utility(utility_name: str) -> str | None:
         (``.exe``, ``.bat``, ``.cmd``). On Unix-like systems, executables
         are identified by the ``X_OK`` (execute) permission bit.
 
+        Lookup is restricted to the trusted directories and the canary-derived
+        CTK root listed above; the process working directory and the ambient
+        ``PATH`` are never consulted.
```
rwgk: See above: delete part after semicolon.

cuda_pathfinder/cuda/pathfinder/_binaries/find_nvidia_binary_utility.py
```diff
@@ -98,10 +180,17 @@ def find_nvidia_binary_utility(utility_name: str) -> str | None:
...
 
     normalized_name = _normalize_utility_name(utility_name)
-    return shutil.which(normalized_name, path=os.pathsep.join(dirs))
+    found = _resolve_in_trusted_dirs(normalized_name, dirs)
+    if found is not None:
+        return found
+
+    # 4. CTK-root canary fallback: only when the explicit trusted dirs above
+    #    miss. Resolve cudart via the dynamic loader (honors LD_LIBRARY_PATH),
+    #    derive the toolkit root, and search its bin layout. PATH is never used.
```
rwgk: See above: delete the last line here.

cuda_pathfinder/docs/source/release/1.6.0-notes.rst
```diff
@@ -0,0 +1,27 @@
...
+
+* :func:`find_nvidia_binary_utility` gains a CTK-root canary fallback. When the
+  NVIDIA wheel, ``CONDA_PREFIX``, and ``CUDA_HOME`` / ``CUDA_PATH`` directories
+  all miss, ``cudart`` is resolved through the OS dynamic loader, which honors
+  ``LD_LIBRARY_PATH`` on Linux and the native DLL search on Windows. The CUDA
+  Toolkit root is derived from that path and its ``bin`` layout is searched.
+  This locates the utility for users who follow the CUDA installation guide and
+  set ``LD_LIBRARY_PATH`` for libraries without also setting ``CUDA_HOME`` /
+  ``CUDA_PATH``, while still never falling back to ``PATH``.
+  (`PR #2196 <https://github.com/NVIDIA/cuda-python/pull/2196>`_)
```
rwgk: Could you please combine this into one bullet point? I think it's best to also delete the "This locates the utility for users who follow..." sentence here.

rwgk (commented):
Generally, in private functions I'd add docstrings only to explain things that are not already obvious. Simple comments are often most efficient.

cuda_pathfinder/cuda/pathfinder/_binaries/find_nvidia_binary_utility.py
```diff
@@ -28,6 +28,51 @@ def _normalize_utility_name(utility_name: str) -> str:
     return utility_name
 
 
+def _is_executable_file(path: str) -> bool:
+    """Return True if ``path`` is a file the OS would run as an executable.
+
+    On Windows executability is determined by the file extension (the
+    candidate name already carries one), so existence is sufficient. On POSIX
+    the execute permission bit must be set.
+    """
```
rwgk: Could we rename this to `_is_executable_candidate`? The helper determines whether a path is a candidate for the finder to return; it does not establish that the OS can actually execute the file. With that name, I think this private function is clear enough without a docstring.

Side note: While looking into this, I noticed that the existing `.bat`/`.cmd` claim below is inconsistent with the implementation. Specifically, `if utility_name not in supported_nvidia_binaries.SUPPORTED_BINARIES` rejects input names ending in `.bat` or `.cmd`, so the current public call path reaches this helper only with `.exe` candidates on Windows, but let's clean up the `.bat`/`.cmd` claim separately in another PR.

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py
```diff
@@ -136,13 +136,26 @@ def _loadable_via_canary_subprocess(libname: str, *, timeout: float = _CANARY_PR
...
 
 
+def resolve_ctk_root_via_canary(canary_libname: str) -> str | None:
+    """Resolve the CUDA Toolkit root from a system-loadable canary library.
+
+    The canary library's absolute path is resolved by the OS dynamic loader in
+    an isolated subprocess, which honors ``LD_LIBRARY_PATH`` on Linux and the
+    native DLL search on Windows. The toolkit root is then derived from that
+    path. Returns ``None`` if the canary cannot be resolved or no root can be
+    derived. The ambient ``PATH`` is never consulted.
```
rwgk: Please delete the last sentence; currently it's not universally true. See the revised PR description.

rwgk (commented):
Hi @aryanputta, could you please make the two small suggested changes?

After that I'll re-trigger the CI and set it to auto-merge.

I already expanded the 1.6.0 release notes, updated the PR title, and the description. I'll make the 1.6.0 release right after this PR is merged.

aryanputta:
> Hi @aryanputta, could you please make the two small suggested changes?
> 
> After that I'll re-trigger the CI and set it to auto-merge.
> 
> I already expanded the 1.6.0 release notes, updated the PR title, and the description. I'll make the 1.6.0 release right after this PR is merged.

will do this now thanks for checking this, learning alot!

aryanputta:
@rwgk Done, both changes are pushed:

1. Renamed `_is_executable_file` to `_is_executable_candidate` and dropped its docstring since the name now carries the intent. Updated the test patch target to match.
2. Removed the "ambient `PATH` is never consulted" sentence from `resolve_ctk_root_via_canary`.

I left the `.bat`/`.cmd` claim untouched for the separate PR as you suggested. Ready for CI whenever you are.

rwgk (approved):
Thanks a lot!

Issue #2200, opened 2026-06-11
[FEA]: Add SM arch detection to CUDA-Pathfinder for LLVM bitcode files


benhg:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

cuda.pathfinder

### Is your feature request related to a problem? Please describe.

Since 3.7.0, NVSHMEM must ship arch-specific bitcode files (libnvshmem_device_smXX.bc) instead of libnvshmem_device.bc. The reason for this is that libnvshmem_device.bc was silently an SM90-only file, and that's not really an acceptable solution.

We attempted to ship a libnvshmem_device.bc which is a symlink to SM90 file just for compatibility, but I was unaware that symlinks are not supported in wheels. 

So this is really an NVSHMEM Bug, but I think longer-term, there is a valid use case for arch-specific bitcode files.

Note that this is not as much of an issue for LTOIR files as they can be built into Fatbins for many architectures.

### Describe the solution you'd like

Add an optional "sm_arch" kwarg to `find_bitcode_lib`:

```
cuda.pathfinder.find_bitcode_lib(name: str, sm_arch: str="") → str
```

If sm_arch is not set, it will find the corresponding .bc file for the library only. If it is set, it will find {lib name}_{sm_arch}.bc

rwgk:
Looks good to me, but a small API nit: I’d prefer `sm_arch: str | None = None` over `sm_arch: str = ""`, since absence seems semantically cleaner than an empty architecture string.

One thing worth thinking about before this becomes a public API: should `sm_arch` be validated/normalized, e.g. accept only `sm_90`/`sm90`/`90` in a documented form? Deciding that now is easier than hardening the API later.

benhg:
That's fine  to me. 

> One thing worth thinking about before this becomes a public API: should sm_arch be validated/normalized, e.g. accept only sm_90/sm90/90 in a documented form? Deciding that now is easier than hardening the API later.

I agree. I don't have a preference. what NVSHMEM does today is sm90, but of course we can change the API surface to wrap around variants for libraries if we need to.

rwgk:
> I agree. I don't have a preference. what NVSHMEM does today is sm90, but of course we can change the API surface to wrap around variants for libraries if we need to.

[...]

PR #2199, opened 2026-06-11
Exclude NVSHMEM 3.7.0 from pathfinder test dependencies and fix rlcompleter test

rwgk:
## Summary
- Add a `!=3.7.0` exclusion to the CUDA 12 and CUDA 13 NVSHMEM test dependencies.
- Keep pathfinder CI from selecting an NVSHMEM package release that does not provide the expected `libnvshmem_device.bc` file.

PR #2201, opened 2026-06-11
Add arch-specific bitcode file search to cuda.pathfinder

benhg:
## Description
Implement #2200. Adds an `sm_arch` argument to CUDA-Pathfinder's bitcode file search.

## Checklist

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_static_libs/find_bitcode_lib.py
+++ b/cuda_pathfinder/cuda/pathfinder/_static_libs/find_bitcode_lib.py
@@ -63,6 +63,7 @@ SUPPORTED_BITCODE_LIBS: tuple[str, ...] = tuple(
         name for name, info in _SUPPORTED_BITCODE_LIBS_INFO.items() if not IS_WINDOWS or info["available_on_windows"]
     )
 )
+_SM_ARCH_PATTERN = re.compile(r"sm[0-9]+[a-z]?")
 
 
 def _no_such_file_in_dir(dir_path: str, filename: str, error_messages: list[str], attachments: list[str]) -> None:
@@ -79,8 +80,8 @@ def _filename_with_sm_arch(filename: str, sm_arch: str | None) -> str:
     if sm_arch is None:
         return filename
 
-    if not re.match(r"^sm[0-9]+[a-z]?$", sm_arch):
-        raise ValueError(f"Invalid sm_arch: '{sm_arch}' must match 'sm[0-9]+[a-z]?'")
+    if not _SM_ARCH_PATTERN.fullmatch(sm_arch):
+        raise ValueError(f"Invalid sm_arch: {sm_arch!r} must match {_SM_ARCH_PATTERN.pattern!r}")
 
     stem, ext = os.path.splitext(filename)
     return f"{stem}_{sm_arch}{ext}"
--- a/cuda_pathfinder/cuda/pathfinder/_static_libs/find_bitcode_lib.py
+++ b/cuda_pathfinder/cuda/pathfinder/_static_libs/find_bitcode_lib.py
@@ -3,6 +3,7 @@
 
 import functools
 import os
+import re
 from dataclasses import dataclass
 from typing import NoReturn, TypedDict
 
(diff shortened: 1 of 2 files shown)


cuda_pathfinder/tests/test_find_bitcode_lib.py
```diff
@@ -23,8 +23,14 @@ def _bitcode_lib_info(libname: str):
     return find_bitcode_lib_module._SUPPORTED_BITCODE_LIBS_INFO[libname]
 
 
-def _bitcode_lib_filename(libname: str) -> str:
-    return _bitcode_lib_info(libname)["filename"]
+def _bitcode_lib_filename(libname: str, sm_arch: str = "") -> str:
+    return find_bitcode_lib_module._filename_with_sm_arch(_bitcode_lib_info(libname)["filename"], sm_arch)
```
benhg: This seems unnecessary.

cuda_pathfinder/tests/test_find_bitcode_lib.py
```diff
@@ -23,8 +23,14 @@ def _bitcode_lib_info(libname: str):
...
+
+
+def _arch_specific_libname() -> str:
```
benhg: This also seems unnecessary.

cuda_pathfinder/cuda/pathfinder/_static_libs/find_bitcode_lib.py
```diff
@@ -130,14 +142,23 @@ def raise_not_found_error(self) -> NoReturn:
         raise BitcodeLibNotFoundError(f'Failure finding "{self.filename}": {err}\n{att}')
 
 
-def locate_bitcode_lib(name: str) -> LocatedBitcodeLib:
+def locate_bitcode_lib(name: str, sm_arch: str = "") -> LocatedBitcodeLib:
```
leofang: Not to agree with the design or if we should start accepting PRs that differentiate SM versions at the pathfinder level, but if we do please make them kw-only args here for maximum backward compatibility 
```suggestion
def locate_bitcode_lib(name: str, *, sm_arch: str = "") -> LocatedBitcodeLib:
```

rwgk: Agreed. @benhg did you intentionally pick `str = None` over `str = None`? I'd still prefer the latter, semantically clearer, less likely to mask subtle bugs.

benhg: Nope, no reason. Will update to that.

cuda_pathfinder/cuda/pathfinder/_static_libs/find_bitcode_lib.py
```diff
@@ -74,13 +75,24 @@ def _no_such_file_in_dir(dir_path: str, filename: str, error_messages: list[str]
         attachments.append(f'  Directory does not exist: "{dir_path}"')
 
 
+def _filename_with_sm_arch(filename: str, sm_arch: str) -> str:
```
rwgk: Small cleanup: switching to `fullmatch`, compile the SM arch pattern once, and reuse `.pattern` in the error message. That keeps the validation contract in one place instead of repeating the regex in code and text.

```python
_SM_ARCH_PATTERN = re.compile(r"sm[0-9]+[a-z]?")


def _filename_with_sm_arch(filename: str, sm_arch: str) -> str:
    if not sm_arch:
        return filename

    if not _SM_ARCH_PATTERN.fullmatch(sm_arch):
        raise ValueError(f"Invalid sm_arch: {sm_arch!r} must match {_SM_ARCH_PATTERN.pattern!r}")

    stem, ext = os.path.splitext(filename)
    return f"{stem}_{sm_arch}{ext}"
```

cuda_pathfinder/cuda/pathfinder/_static_libs/find_bitcode_lib.py
```diff
@@ -170,11 +194,22 @@ def locate_bitcode_lib(name: str) -> LocatedBitcodeLib:
 
 
 @functools.cache
-def find_bitcode_lib(name: str) -> str:
+def find_bitcode_lib(name: str, sm_arch: str | None = None) -> str:
```
rwgk: Since `find_bitcode_lib()` is also public API, can we make `sm_arch` keyword-only there too?

```suggestion
def find_bitcode_lib(name: str, *, sm_arch: str | None = None) -> str:
```

It would also be good to add a small test that the positional form fails, so this API shape does not regress later:

```python
def test_find_bitcode_lib_sm_arch_is_keyword_only():
    with pytest.raises(TypeError):
        find_bitcode_lib("nvshmem_device", "sm90")
```

rwgk:
Hi @benhg, I went ahead and added a commit to parametrize the `sm_arch_is_keyword_only` test, so that we also have it for `locate_bitcode_lib`. Running the CI now.

rwgk (approved):
Thanks @benhg, I'll set this to auto-merge. 🤞

rwgk:
Sorry I overlooked before that this PR was behind `main`. It was missing PR #2199. I clicked the Update branch button and will retrigger the CI.

rwgk:
Converting to draft, so this doesn't get merged accidentally, pending offline nvshmem design/packaging discussions.

benhg:
Closing in favor of new approach mentioned in issue #2200

PR #2207, opened 2026-06-12
Restart sccache to avoid timeouts

mdboom:
I have been see a number of sccache timeout failures with out 3.15 builds.  What makes them special is they have to build numpy from source.  This is my agent's suggestion to help with that -- I have no idea why it would fail and why this would help, but thought I would at least run the experiment.

### DIFF
--- a/.github/workflows/build-wheel.yml
+++ b/.github/workflows/build-wheel.yml
@@ -456,6 +456,17 @@ jobs:
           path: ${{ env.CUDA_CORE_CYTHON_TESTS_DIR }}/test_*${{ env.PY_EXT_SUFFIX }}
           if-no-files-found: error
 
+      # Restart sccache before the second cuda.core build. For py3.15t, the
+      # preceding numpy-from-source build can take 10+ minutes during which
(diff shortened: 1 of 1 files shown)


leofang:
@trxcllnt would you mind taking a look? Is it something you've seen before? This timeout error seems to randomly appear in the past few days.

FWIW, we use GHA Cache as the sccache bucket and in parallel we just reported a network issue between GitHub and IPP to IT. Could also be related.

trxcllnt:
I have observed the GHA cache service has been flaky the past few weeks. Feel free to switch to the `rapids-sccache-devs` bucket we use in RAPIDS/CCCL (e.g. like [this](https://github.com/NVIDIA/cccl/blob/61493f8d7e3046b5aef9fe58b60a75df142ffb65/.github/actions/workflow-run-job-linux/action.yml#L69-L75)).

I don't believe this PR does what the agent says it will, since the sccache daemon runs inside the cibuildwheel container. This PR is stopping and starting the daemon on the runner, but the cibuildwheel container will start and use its own daemon.

mdboom:
Seems like whatever the fix is is not this, so closing.

PR #2209, opened 2026-06-13
[DO-NOT-MERGE] Demonstrate CI gating-check bug: build fail -> Check job status green

leofang:
**Do not merge.** This is a demonstration of the bug described in #2208.

Injects `exit 7` early in `ci/tools/env-vars` so every `Build *` matrix entry fails at the env-vars step (before cibuildwheel runs). Tests downstream will be `skipped` due to needs-failure propagation.

Expected outcome:
- `Build linux-64`, `Build linux-aarch64`, `Build win-64` → failure
- `Test linux-64`, `Test linux-aarch64`, `Test win-64`, `Test sdist *` → skipped
- **`Check job status` → success** ← the bug being demonstrated

This PR will be closed once CI confirms the reproduction.

Refs #2208

PR #2210, opened 2026-06-13
ci: fix Check job status to detect skipped results (fixes #2208)

### LINKED ISSUE #2208: CI: "Check job status" aggregator passes when build jobs fail (skipped-as-success regression)
The required status check `Check job status` reports `success` even when one or more `Build *` jobs fail. The merge gate on `main` is effectively bypassable today.

## Symptom

Recent CI runs on `main` where build jobs failed but `Check job status` succeeded:

- Run https://github.com/NVIDIA/cuda-python/actions/runs/27442561109 (head of PR-branch pull-request/2161, "CI: fetch tags for release dry-run validation"): `Build linux-64, CUDA 13.3.0 / py3.13` failed → `Check job status = success`.
- Run https://github.com/NVIDIA/cuda-python/actions/runs/27434626922 (head of pull-request/2206, ancestor of c12838169 currently HEAD of main): 24 `Build linux-*` failures across the matrix → `Check job status = success`.

Both runs are on real PRs (not `[no-ci]` / `[doc-only]`). The repository's `Prevent committing without PR` ruleset on `main` lists only `Check job status` + `pre-commit.ci - pr` as required status checks. Nothing else gates the merge.

## Root cause

`Check job status` is defined here (current main, `c12838169e`):
https://github.com/NVIDIA/cuda-python/blob/c12838169e7637815e06ca2746dfb1ab727ec13a/.github/workflows/ci.yml#L419-L483

The `Exit` step body only catches `'cancelled'` or `'failure'` results on its `test-*`, `doc`, `test-sdist-*` dependencies — see the inline `if` chain:
https://github.com/NVIDIA/cuda-python/blob/c12838169e7637815e06ca2746dfb1ab727ec13a/.github/workflows/ci.yml#L460-L483

In particular, the test-job clause (lines 474–479) reads:
https://github.com/NVIDIA/cuda-python/blob/c12838169e7637815e06ca2746dfb1ab727ec13a/.github/workflows/ci.yml#L474-L479

When `build-linux-64` (or `build-windows`) fails, GitHub Actions sets the dependent `test-linux-64.result` to `'skipped'` (default needs-failure propagation). `'skipped'` is not in the catch list, so the aggregator exits 0.

GitHub's "skipped checks count as success for required status checks" behavior — explicitly called out in CCCL's `ci:` gate — makes this a silent merge gate bypass:
https://github.com/NVIDIA/cccl/blob/8c0e6cb1b6412d7bd0070f9ac3f55fa80231961a/.github/workflows/ci-workflow-pull-request.yml#L463-L526

## Regression history

PR #1008 (merged 2025-10-01) replaced `poseidon/wait-for-status-checks` with a `needs:`-only aggregator and an `exit 0` step. The previous design was here, at the pre-merge base commit:
https://github.com/NVIDIA/cuda-python/blob/c4bb6230eb448a77b38b707cfb4c3be24808cd6f/.github/workflows/status-check.yml#L14-L20

The original action would have failed on any non-success check (including `skipped`); the replacement does not. Subsequent additions of `if: always()` + the `cancelled || failure` chain partially restored intent but did not cover the `skipped` case.

## Proposed fix

Adopt CCCL's `check_result` pattern: explicit `expected="success"` per dependency, with `expected="skipped"` for legitimate `[doc-only]` skips. Reference implementation (CCCL):
https://github.com/NVIDIA/cccl/blob/8c0e6cb1b6412d7bd0070f9ac3f55fa80231961a/.github/workflows/ci-workflow-pull-request.yml#L463-L526

Sketch adapted to cuda-python:

```
- name: Check results
  run: |
    if [[ "${{ needs.should-skip.outputs.skip }}" == "true" ]]; then
      echo "[no-ci] - skipping aggregator checks"
      exit 0
    fi
    doc_only="${{ needs.should-skip.outputs.doc-only }}"
    status="success"
    check_result() {
      name=$1; expected=$2; result=$3
      if [[ "$result" \!= "$expected" ]]; then
        echo "::error::$name did not match expected result"
        status="failed"
      fi
    }
    check_result "detect-changes" "success" "${{ needs.detect-changes.result }}"
    check_result "doc"            "success" "${{ needs.doc.result }}"
    if [[ "$doc_only" == "true" ]]; then expected="skipped"; else expected="success"; fi
    check_result "test-sdist-linux"   "$expected" "${{ needs.test-sdist-linux.result }}"
    # ...etc for every test-*
    [[ "$status" == "success" ]]
```

## Demonstration

A draft PR will be opened against `main` that intentionally fails the Build stage. It should reproduce: Build jobs go red, `Check job status` goes green.

## Related

NVIDIA/numba-cuda-mlir shares the same aggregator design and was bitten by this in NVIDIA/numba-cuda-mlir#125 (auto-merged with failing Windows builds; reverted in NVIDIA/numba-cuda-mlir#134). Fix should be ported there once shape is agreed.

leofang:
Fixes #2208.

Replaces the `cancelled || failure` predicate in `.github/workflows/ci.yml`'s `checks:` job with CCCL's `check_result` pattern: per-dep `expected="success"`, with `expected="skipped"` for legitimate `[doc-only]` skips, and an early short-circuit for `[no-ci]`.

Reference: https://github.com/NVIDIA/cccl/blob/8c0e6cb1b6412d7bd0070f9ac3f55fa80231961a/.github/workflows/ci-workflow-pull-request.yml#L463-L526

End-to-end validation (pre-fix vs post-fix reproducer): https://github.com/NVIDIA/cuda-python/pull/2210#issuecomment-4714140531

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -432,52 +432,42 @@ jobs:
     steps:
       - name: Exit
         run: |
-          # if any dependencies were cancelled or failed, that's a failure
-          #
-          # see https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#always
-          # and https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks
-          # for why this cannot be encoded in the job-level `if:` field
-          #
-          # TL; DR: `$REASONS`
-          #
-          # The intersection of skipped-as-success and required status checks
-          # creates a scenario where if you DON'T `always()` run this job, the
-          # status check UI will block merging and if you DO `always()` run and
-          # a dependency is _cancelled_ (due to a critical failure, which is
-          # somehow not considered a failure ¯\_(ツ)_/¯) then the critically
-          # failing job(s) will timeout causing a cancellation here and the
-          # build to succeed which we don't want (originally this was just
-          # 'exit 0')
-          #
-          # Note: When [doc-only] is in PR title, test jobs are intentionally
-          # skipped and should not cause failure.
-          #
-          # detect-changes gates whether heavy test matrices run at all; if it
-          # does not succeed, downstream test jobs are skipped rather than
-          # failed, which would otherwise go unnoticed here. Require its
-          # success explicitly so a broken gating step cannot masquerade as a
-          # green CI run.
-          doc_only=${{ needs.should-skip.outputs.doc-only }}
-          if ${{ needs.detect-changes.result != 'success' }}; then
-            exit 1
+          # GitHub treats `result == 'skipped'` as success for required
+          # status checks (see CCCL gate comment + cccl#605). The previous
+          # `cancelled || failure` predicate let upstream build failures
+          # propagate as `skipped` on downstream test jobs and silently
+          # pass this aggregator. Adopt CCCL's `check_result` pattern:
+          # require an explicit `expected` status per dependency, where
+          # anything else (including `skipped` from a failed upstream)
+          # fails the gate. `if: always()` on the job still ensures this
+          # step runs even when needs are skipped.
+          if [[ "${{ needs.should-skip.outputs.skip }}" == "true" ]]; then
+            echo "[no-ci] - skipping aggregator checks"
+            exit 0
           fi
-          if ${{ needs.doc.result == 'cancelled' || needs.doc.result == 'failure' }}; then
-            exit 1
-          fi
-          if ${{ needs.test-sdist-linux.result == 'cancelled' ||
-                 needs.test-sdist-linux.result == 'failure' ||
-                 needs.test-sdist-windows.result == 'cancelled' ||
-                 needs.test-sdist-windows.result == 'failure' }}; then
-            exit 1
-          fi
-          if [[ "${doc_only}" != "true" ]]; then
-            if ${{ needs.test-linux-64.result == 'cancelled' ||
-                   needs.test-linux-64.result == 'failure' ||
-                   needs.test-linux-aarch64.result == 'cancelled' ||
-                   needs.test-linux-aarch64.result == 'failure' ||
-                   needs.test-windows.result == 'cancelled' ||
-                   needs.test-windows.result == 'failure' }}; then
-              exit 1
+
+          doc_only="${{ needs.should-skip.outputs.doc-only }}"
+          status="success"
+          check_result() {
+            name=$1; expected=$2; result=$3
+            echo "Checking $name: result='$result' (expected '$expected')"
+            if [[ "$result" != "$expected" ]]; then
+              echo "::error::$name did not match expected result"
+              status="failed"
             fi
-          fi
-          exit 0
+          }
+
+          # always expected to succeed (even in [doc-only] mode)
+          check_result "should-skip"    "success" "${{ needs.should-skip.result }}"
+          check_result "detect-changes" "success" "${{ needs.detect-changes.result }}"
+          check_result "doc"            "success" "${{ needs.doc.result }}"
+
+          # [doc-only] flips these from 'success' to 'skipped'
+          if [[ "$doc_only" == "true" ]]; then expected="skipped"; else expected="success"; fi
+          check_result "test-sdist-linux"   "$expected" "${{ needs.test-sdist-linux.result }}"
+          check_result "test-sdist-windows" "$expected" "${{ needs.test-sdist-windows.result }}"
+          check_result "test-linux-64"      "$expected" "${{ needs.test-linux-64.result }}"
+          check_result "test-linux-aarch64" "$expected" "${{ needs.test-linux-aarch64.result }}"
+          check_result "test-windows"       "$expected" "${{ needs.test-windows.result }}"
+
+          [[ "$status" == "success" ]]
--- a/ci/tools/env-vars
+++ b/ci/tools/env-vars
@@ -15,16 +15,6 @@ if [[ ${#} -ne 1 ]]; then
   exit 1
 fi
 
-# [demo] Intentional failure to validate the gating-check fix in this PR.
-# Mirrors the #2209 reproducer: every Build matrix entry should fail here,
-# downstream Test jobs should be skipped, and -- with the new aggregator --
-# Check job status should now report failure (whereas it reported success
-# in #2209 before the fix). REMOVE THIS BLOCK before merging.
-if [[ "${1}" == "build" ]]; then
-  echo "::error::INTENTIONAL FAILURE validating gating-check fix on PR #2210. See https://github.com/NVIDIA/cuda-python/issues/2208" >&2
-  exit 7
-fi
-
 PYTHON_VERSION_FORMATTED=$(echo "${PY_VER}" | tr -d '.')
 
 if [[ "${HOST_PLATFORM}" == linux* ]]; then
--- a/ci/tools/env-vars
+++ b/ci/tools/env-vars
@@ -15,6 +15,16 @@ if [[ ${#} -ne 1 ]]; then
   exit 1
 fi
 
+# [demo] Intentional failure to validate the gating-check fix in this PR.
+# Mirrors the #2209 reproducer: every Build matrix entry should fail here,
+# downstream Test jobs should be skipped, and -- with the new aggregator --
+# Check job status should now report failure (whereas it reported success
+# in #2209 before the fix). REMOVE THIS BLOCK before merging.
+if [[ "${1}" == "build" ]]; then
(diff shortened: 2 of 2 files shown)


leofang:
### Gating-check fix validated end-to-end

The new aggregator catches the exact failure mode that motivated #2208. Demonstrated by running the #2209 reproducer (intentional `exit 7` in `ci/tools/env-vars`) against this branch:

- **Pre-fix** (PR #2209, head [`deddae4794`](https://github.com/NVIDIA/cuda-python/commit/deddae4794)): https://github.com/NVIDIA/cuda-python/actions/runs/27451928558 — 24 Build failures, all `Test *` wheels `skipped`, **`Check job status = success`** (the bug).
- **Post-fix** (this branch, head [`811388c99e`](https://github.com/NVIDIA/cuda-python/commit/811388c99e)): https://github.com/NVIDIA/cuda-python/actions/runs/27588163601 — same 24 Build failures, same `Test *` wheels `skipped`, **`Check job status = failure`** (the fix working).

Same reproducer, opposite verdict. The skipped-as-success hole is closed.

The demo commit [`811388c99e`](https://github.com/NVIDIA/cuda-python/commit/811388c99e) will be reverted next so this PR returns to a mergeable state.

rwgk (approved):
Looks good to me and codex gpt-5.5:

The new `check_result` approach is a clear improvement over the previous `cancelled || failure` checks: it closes the skipped-as-success hole while keeping the intentional `[doc-only]` and `[no-ci]` cases explicit.

The one subtle behavior change I noticed is that fork or non-NVIDIA repo runs may now report a red final aggregator where jobs are skipped by owner-gated conditions. I agree that is not a concern for this PR, and it is preferable to avoid weakening the aggregator in a way that could reintroduce false-green CI.

.github/workflows/ci.yml
```diff
@@ -432,52 +432,42 @@ jobs:
     steps:
       - name: Exit
         run: |
```
u7k4rs6, line 434: **[STYLE]** Redundant comments explaining the `always()` condition.

**Why it matters:** The extensive comments at the beginning of the job are no longer relevant to the new logic and add unnecessary noise. The new logic is self-explanatory with the `check_result` function.

**Suggested fix:** Remove the block of comments starting from line 434 down to line 473 (inclusive) that explains the `always()` condition and the previous logic.

---
*Explanation for juniors: The comments describe a complex reasoning for using `always()` and the previous failure condition. With the introduction of the `check_result` function and a more explicit status checking mechanism, these comments are outdated and can be removed to improve readability.*

u7k4rs6, line 434: **[BUG]** Incorrect handling of skipped jobs in CI aggregation.

**Why it matters:** The previous logic `cancelled || failure` would cause jobs that were intentionally skipped (e.g., due to `doc-only` flag) to be treated as failures if their upstream dependencies were cancelled. This prevented the CI from passing when it should have.

**Suggested fix:** Replace the old `if` conditions checking for `cancelled || failure` with a new `check_result` function that explicitly checks for the expected status ('success' or 'skipped').

---
*Explanation for juniors: The new `check_result` function, along with the `expected` variable, provides a more robust way to handle job statuses. It allows for defining what status is expected for each job, including 'skipped' when the `doc-only` flag is true. This ensures that intentionally skipped jobs do not cause the CI to fail.*

u7k4rs6, line 434: **[PERFORMANCE]** Potentially unnecessary execution of CI jobs.

**Why it matters:** The `if: always()` on the job ensures that this aggregation step runs even if upstream jobs are skipped. While this is necessary for the aggregation logic to function, it might lead to this job running when it's not strictly needed if the `should-skip` condition is met.

**Suggested fix:** Add an explicit check at the beginning of the job to exit early if `needs.should-skip.outputs.skip` is true.

---
*Explanation for juniors: The new logic includes an early exit condition: `if [[ "${{ needs.should-skip.outputs.skip }}" == "true" ]]; then echo "[no-ci] - skipping aggregator checks"; exit 0; fi`. This prevents the rest of the aggregation logic from running when the build is intended to be skipped, improving efficiency.*

.github/workflows/ci.yml
```diff
@@ -432,52 +432,42 @@ jobs:
...
+          # propagate as `skipped` on downstream test jobs and silently
+          # pass this aggregator. Adopt CCCL's `check_result` pattern:
+          # require an explicit `expected` status per dependency, where
+          # anything else (including `skipped` from a failed upstream)
+          # fails the gate. `if: always()` on the job still ensures this
+          # step runs even when needs are skipped.
+          if [[ "${{ needs.should-skip.outputs.skip }}" == "true" ]]; then
+            echo "[no-ci] - skipping aggregator checks"
+            exit 0
           fi
```
u7k4rs6, line 447: **[STYLE]** Inconsistent variable assignment for `doc_only`.

**Why it matters:** The `doc_only` variable was previously assigned using `${{ needs.should-skip.outputs.doc-only }}` directly within the `if` condition. Now, it's assigned to a shell variable `doc_only` using `doc_only="${{ needs.should-skip.outputs.doc-only }}"` before being used.

**Suggested fix:** Ensure consistent variable assignment. The new assignment `doc_only="${{ needs.should-skip.outputs.doc-only }}"` is correct for shell scripting.

---
*Explanation for juniors: The change from using the GitHub Actions expression directly in the `if` statement to assigning it to a shell variable `doc_only` is a stylistic improvement for clarity and consistency within the shell script block. The new approach is more readable.*

u7k4rs6, line 447: **[STYLE]** Complex and repetitive `if` conditions.

**Why it matters:** The original `if` conditions checking for multiple job results (`cancelled || failure`) were long and repetitive, making them hard to read and maintain.

**Suggested fix:** Replace the complex `if` conditions with calls to the `check_result` function.

---
*Explanation for juniors: The introduction of the `check_result` function encapsulates the logic for checking job statuses. This makes the code much cleaner, more readable, and easier to extend or modify in the future. Each job's status is now checked individually with a clear expected outcome.*

u7k4rs6, line 447: **[STYLE]** Lack of explicit success status tracking.

**Why it matters:** The previous logic relied on `exit 1` to indicate failure. There was no explicit variable to track the overall success status of the aggregation.

**Suggested fix:** Introduce a `status` variable initialized to 'success' and update it to 'failed' within the `check_result` function if any check fails.

---
*Explanation for juniors: The new `status` variable provides a clear way to track the overall outcome of the CI aggregation. The `check_result` function modifies this variable upon encountering an unexpected job result, and the final `[[ "$status" == "success" ]]` expression determines the job's overall success or failure.*

.github/workflows/ci.yml
```diff
@@ -432,52 +432,42 @@ jobs:
...
+
+          doc_only="${{ needs.should-skip.outputs.doc-only }}"
+          status="success"
+          check_result() {
+            name=$1; expected=$2; result=$3
+            echo "Checking $name: result='$result' (expected '$expected')"
+            if [[ "$result" != "$expected" ]]; then
+              echo "::error::$name did not match expected result"
+              status="failed"
             fi
```
u7k4rs6, line 457: **[STYLE]** Redundant `exit 0`.

**Why it matters:** The `exit 0` at the end of the script was redundant because the script would naturally exit with a success code if no `exit 1` was encountered. The new logic uses a final expression to determine the exit code.

**Suggested fix:** Remove the `exit 0` statement.

---
*Explanation for juniors: The new logic concludes with `[[ "$status" == "success" ]]`. This expression evaluates to true (exit code 0) if `status` is 'success', and false (exit code 1) otherwise. This is a more idiomatic way to handle script exit codes based on a condition and makes the redundant `exit 0` unnecessary.*

u7k4rs6 (commented):
## PR Risk Summary
**Quality Score:** 3/10
**Risk Level:** high
**Merge Recommendation:** Do not merge
**Rationale:** The CI workflow file has several issues. Most critically, there's an incorrect handling of skipped jobs in CI aggregation, which could lead to inaccurate reporting of build statuses. Additionally, there are concerns about potentially unnecessary job executions, impacting CI performance. Several style issues, including redundant comments, inconsistent variable assignment, complex `if` conditions, lack of explicit success tracking, and redundant `exit 0`, further reduce the quality of the changes. These issues collectively pose a significant risk to the CI pipeline's reliability and efficiency.

**High Risk Changes:**
- Incorrect handling of skipped jobs in CI aggregation.
- Potentially unnecessary execution of CI jobs.

u7k4rs6 (commented):
## PR Risk Summary
**Quality Score:** 10/10
**Risk Level:** low
**Merge Recommendation:** Safe to merge
**Rationale:** The only file changed is a GitHub Actions CI workflow file. No functional code changes were introduced, and the review found no issues. This change is low risk.

leofang:
@u7k4rs6 please do not fire up nonsense agents against our repo and add noise. This is not a welcomed behavior.

u7k4rs6:
@leofang Apologies, that was an automated agent I should not have run against your repo, especially on an already-merged PR. It added noise and got the core point wrong. I've turned it off. Thanks for the patience.

PR #2073, opened 2026-05-12
[no-ci] Prepare for cuda.core v1.0.1 release

leofang:
## Description

closes 

## Checklist

- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #2213, opened 2026-06-14
ci: remove release-time self-test of release-notes checker

### LINKED ISSUE #2062: [BUG] CI: "Self-test release-notes checker" fails due to root conftest.py import error
## Description

The `check-release-notes` job in the `CI: Release` workflow (`release.yml`) fails at the "Self-test release-notes checker" step with pytest exit code 4 (collection error).

Example failure: https://github.com/NVIDIA/cuda-python/actions/runs/25701496246/job/75462415468

## Root cause

The step runs:

```yaml
- name: Self-test release-notes checker
  run: |
    pip install pytest
    pytest ci/tools/tests
```

When pytest launches from the repo root, it discovers the root-level `conftest.py`, which imports `cuda.pathfinder`. Since the job only installs `pytest` (not the project packages), the import fails and pytest aborts before any tests run.

The actual release-notes file (`cuda_core/docs/source/release/1.0.0-notes.rst`) is fine — the job never reaches the "Check versioned release notes exist" step.

## Suggested fix

Remove the self-test step from the release workflow entirely. Running infra/tooling tests during a release is not the right place — they should be validated separately (e.g. in a CI workflow that runs on PRs).

For reference, this was already worked around by commenting out the step in a release branch: https://github.com/NVIDIA/cuda-python/commit/40508c5e0c356476e82e843ad8c9606633d57ac6

aryanputta:
## Description

Fixes #2062.

The `check-release-notes` job in `release.yml` ran `pytest ci/tools/tests` from the repo root. Pytest then discovers the root `conftest.py`, which imports `cuda.pathfinder`. The job installs only `pytest` (not the project packages), so collection aborted with exit code 4 before any test ran, and the job never reached the actual `Check versioned release notes exist` step.

As noted in the issue, running infra/tooling tests during a release is the wrong place for them. The step was already disabled (commented out) and carried to `main` via #2073. This removes the dead block so the release workflow only validates that the versioned release notes exist.

## Change

- Remove the commented-out `Self-test release-notes checker` step from the `check-release-notes` job in `.github/workflows/release.yml`.

The `Set up Python` step is retained because `Check versioned release notes exist` still runs `python ci/tools/check_release_notes.py`.

## Follow-up (out of scope)

The tooling tests under `ci/tools/tests` can be exercised in a PR-triggered CI workflow so they stay covered without the release-time package-import problem. Happy to do that as a separate PR if maintainers want it; running them cleanly needs to avoid loading the root `conftest.py` (e.g. `--confcutdir`), so it is kept separate from this minimal fix.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -164,11 +164,6 @@ jobs:
         with:
           python-version: "3.12"
 
-      # - name: Self-test release-notes checker
-      #   run: |
-      #     pip install pytest
-      #     pytest ci/tools/tests
-
       - name: Check versioned release notes exist
         run: |
           python ci/tools/check_release_notes.py \


aryanputta:
The `CI/CD` label was applied automatically. The metadata check is still red on milestone and assignee, which I can't set as an external contributor. Could a maintainer add a milestone (e.g. `cuda.bindings next`, matching recent CI/CD PRs) and an assignee? Happy to adjust anything else.

leofang:
Sorry, Aryan, I am afraid I need to close your PR. We'll do trivial clean-ups like this ourselves. Plus, as per [our contributor guideline](https://github.com/NVIDIA/cuda-python/blob/main/CONTRIBUTING.md) we need contributors to discuss with us in the target issue first before we decide if an external PR could be accepted. In this case, the issue already has an assignee.

PR #2215, opened 2026-06-15
Add critical section guards to accessors

seberg:
Protect accessors with cython critical sections so free-threaded execution avoids races when reading cached state.

`@cython.critical_section` effectively takes a lock based on the first argument passed in.  (Although that lock _can_ be released to avoid deadlocks, it is as safe or safer as the GIL, but that doesn't make it 100% in all situations.)

At least one of these was noticed by free-threaded test runs.  I am not certain that they will guarantee identity returns, but pretty sure this will make them thread-safe.
(The reason being that at least on non free-threaded builds they may release the GIL or, on free-threaded, do API calls that release the critical section temporarily.)

About the `.pyi` fixes: I think these are incorrect but `cython` is already used in other places and shouldn't be a runtime/typing requirement at all.

Towards gh-2194 which will add some testing although not strictly guaranteed for all paths covered here.


seberg:
/ok to test [652a3e7](https://github.com/NVIDIA/cuda-python/pull/2215/commits/652a3e7480bc58cd07907080088af4cf2e563489)

mdboom:
> About the .pyi fixes: I think these are incorrect but cython is already used in other places and shouldn't be a runtime/typing requirement at all.

We probably need to submit a fix to `stubgen-pyx` to not do this.  I think it's probably just being "literal" about what's there.  I don't think it "breaks" our `.pyi` files, just unnecessary.

There are obviously a lot more properties/accessors in the codebase.  Besides seeing multi-threaded test failures, are there other criteria we should follow about when to apply these decorators?

seberg:
> We probably need to submit a fix to stubgen-pyx to not do this. I think it's probably just being "literal" about what's 

Yeah, right now it seems you need `cython` and will just get `Any` (because Cython isn't fully typed I think).  A simple "ignore these" feature would be good.

> There are obviously a lot more properties/accessors in the codebase.

Whenever we mutate a `cdef object` (global or class instance) then there can be a race that breaks the `INCREF`/`DECREF` balancing.
That is even if we do _not_ care about calls being strictly idempotent (once cached a cache value can't be replaced).  And that incref/decref balancing issue is strictly a free-threading issue.

I would hope that Cython will provide things to make this easier in the future (maybe `atomic object`).

You are not wrong, in that I should make another pass and maybe TSAN testing could flush out some as well.  In practice hitting these issues is very unlikely but if you do...

cuda_core/cuda/core/_memory/_buffer.pyx
```diff
@@ -445,6 +446,7 @@ cdef class Buffer:
 
 # Memory Attribute Query Helpers
 # ------------------------------
+@cython.critical_section
```
seberg: Whoops, I solved this with the atomic not sure why it came back.
```suggestion
```

leofang (approved):

seberg:
/ok to test [d750365](https://github.com/NVIDIA/cuda-python/pull/2215/commits/d7503654d61fcb0e52b0c83eb2e4c9cdc77899ff)

seberg:
/ok to test [1d5c0ce](https://github.com/NVIDIA/cuda-python/pull/2215/commits/1d5c0ce8406e2b236c3e9d5ca31788121c276291)

PR #2216, opened 2026-06-15
Use atomic flag for buffer memory-attribute initialization

seberg:
Store and read `Buffer._mem_attrs_inited` via std::atomic using acquire/release semantics to avoid races during lazy pointer memory- attribute initialization under free-threaded execution.

At least strictly speaking, `_mem_attrs_inited` is a flag that guards access to other memory attributes.
My understanding is that on ARM processors with weak memory ordering this can be a problem even beyond compiler optimizations making use of the fact that order isn't strictly guaranteed (without the light use of atomics).

I.e.:
thread1 -> init _mem_attrs
thread1 -> sets _mem_attrs_inited
thread2 -> reads _mem_attrs_inited
thread2 -> reads _mem_attrs

Does *not* guarantee that thread2 doesn't read the uninitialized `_mem_attrs`

**This is probably not sufficiently covered by pytest-run-parallel, so if desired I/we should add an explicitly threaded test for it. (Although I have doubts we'll be able to fail the test even without the changes)**

### DIFF
--- a/cuda_core/cuda/core/_memory/_buffer.pxd
+++ b/cuda_core/cuda/core/_memory/_buffer.pxd
@@ -3,6 +3,8 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from libc.stdint cimport uintptr_t
+from libcpp cimport bool as cpp_bool
+from libcpp.atomic cimport atomic as std_atomic, memory_order_acquire, memory_order_release
 
 from cuda.bindings cimport cydriver
 from cuda.core._resource_handles cimport DevicePtrHandle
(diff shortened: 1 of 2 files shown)


seberg:
/ok to test [7c1c215](https://github.com/NVIDIA/cuda-python/pull/2216/commits/7c1c21578d17a353ca515eabcd357034b33206c1)

leofang (approved):
The change seems fine to me. Though I wonder if there'd be a significant perf hit. Originally `Buffer` was meant to be a thin wrapper over the raw pointer, but we've grown the complexity over time. 

Might be a good idea checking with @danielfrg and see if the micro benchmarks can tell us the perf difference. I dunno if/where it runs TBH.

In any case, approving. Thanks, Sebastian!

seberg:
> if there'd be a significant perf hit

I doubt this'll even register, I think the critical sections may be measurable but still so fast that it shouldn't really matter.
(I don't think atomics have much of an impact on non-threaded code, and even if threaded this should scale well, I believe.)

PR #2217, opened 2026-06-15
Use setdefault for internal cache to ensure identity caching

seberg:
Switch selected cache write sites to setdefault/get-based patterns and keep explicit comments where setdefault is intentionally not used for primitive value caches.
(I just thought having `setdefault` mentioned might be nice in case of cargo-culting.)

Towards gh-2194; pytest-run-parallel does have a specific test for the `GraphMemoryResource` change.  I could add an identity test for `ComputeCapability` probably (I doubt it actually matters for it, the use of `setdefault` just feels like a good pattern to me).

### DIFF
--- a/cuda_core/cuda/core/_device.pyx
+++ b/cuda_core/cuda/core/_device.pyx
@@ -85,9 +85,12 @@ cdef class DeviceProperties:
 
     cdef inline int _get_cached_attribute(self, attr, default=0) except? -2:
         """Retrieve the attribute value, using cache if applicable."""
-        if attr not in self._cache:
-            self._cache[attr] = self._get_attribute(attr, default)
-        return self._cache[attr]
+        cached = self._cache.get(attr)
+        if cached is not None:
+            return cached
+        cdef int value = self._get_attribute(attr, default)
+        self._cache[attr] = value  # setdefault not needed for ints
+        return value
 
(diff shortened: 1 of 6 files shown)


seberg:
/ok to test [3497ccf](https://github.com/NVIDIA/cuda-python/commit/3497ccfaa002ab285ccc7d04db57c2f57a8483d3)

mdboom (approved):
LGTM.  

Do we want to add any AGENTS.md content about this pattern being preferred over `functools.cache` or the non-setdefault caching pattern?

seberg:
> Do we want to add any AGENTS.md content about this pattern being preferred over functools.cache or the non-setdefault caching pattern?

Not sure `AGENTS.md` has such code styles yet?  But if we want to add that I can follow-up.
FWIW, in most cases `@functools.cache` is nicer, but if it matters to return an idempotent result strictly then it doesn't (at the moment).

However, @rwgk: When scanning for `@cache` I didn't modify `cuda-pathfinder` since almost all places were just strings, etc. where it is good.
The one odd-one-out that caught my eye is `load_nvidia_dynamic_lib` if loading twice from two threads or the function being _strictly_ idempotent can matter, then we need to fix it (I can do that in a follow-up if it seems important to you.)

Issue #1307, opened 2025-12-03
cufile tests don't pass on ext4 filesystems in CI


cpcloud:
After https://github.com/NVIDIA/cuda-python/pull/1271, we have to xfail a number of cufile tests in CI that run on an ext4 filesystem even though that is nominally supported.

It would be good to figure out what's going on here and why these don't work in CI even though they are running on an ext4 filesystem.

cpcloud:
cc @sourabgupta3

rwgk:
I was curious and looked at the CI run on `main` after #1271 was merged:

https://github.com/NVIDIA/cuda-python/actions/runs/19907928749

After unpacking the log archive locally:

```
$ grep 'Z XFAIL' Test*.txt | wc -l
169
```

Showing only the first few:

```
$ grep 'Z XFAIL' Test*.txt | head -n 20
Test_linux-64___py3.10__13.0.2__wheels__l4.txt:2025-12-03T20:40:43.0964244Z XFAIL tests/test_cufile.py::test_handle_register - handle_register call fails in CI for unknown reasons
Test_linux-64___py3.10__13.0.2__wheels__l4.txt:2025-12-03T20:40:43.0966208Z XFAIL tests/test_cufile.py::test_cufile_read_write - handle_register call fails in CI for unknown reasons
Test_linux-64___py3.10__13.0.2__wheels__l4.txt:2025-12-03T20:40:43.0967481Z XFAIL tests/test_cufile.py::test_cufile_read_write_host_memory - handle_register call fails in CI for unknown reasons
Test_linux-64___py3.10__13.0.2__wheels__l4.txt:2025-12-03T20:40:43.0969007Z XFAIL tests/test_cufile.py::test_cufile_read_write_large - handle_register call fails in CI for unknown reasons
Test_linux-64___py3.10__13.0.2__wheels__l4.txt:2025-12-03T20:40:43.0970180Z XFAIL tests/test_cufile.py::test_cufile_write_async - handle_register call fails in CI for unknown reasons
Test_linux-64___py3.10__13.0.2__wheels__l4.txt:2025-12-03T20:40:43.0971317Z XFAIL tests/test_cufile.py::test_cufile_read_async - handle_register call fails in CI for unknown reasons
Test_linux-64___py3.10__13.0.2__wheels__l4.txt:2025-12-03T20:40:43.0972477Z XFAIL tests/test_cufile.py::test_cufile_async_read_write - handle_register call fails in CI for unknown reasons
[...]

PR #2218, opened 2026-06-15
TST: Use tempdir for cufile tests (and mark some as thread unsafe) 

seberg:
## Description

Split out, since it also seems that changing these to use a temp_dir (necessary to run them
in parallel in multiple threads) also fixes the CI failures.

This PR also adds `thread_unsafe` markers to a few tests here (and to the `pytest.ini`, I have used the other one for other tests in the full changeset)

Towards gh-2194

## Checklist

### DIFF
--- a/cuda_bindings/tests/test_cufile.py
+++ b/cuda_bindings/tests/test_cufile.py
@@ -1,8 +1,7 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: LicenseRef-NVIDIA-SOFTWARE-LICENSE
 
 import ctypes
-import errno
 import logging
 import os
 import pathlib
@@ -118,12 +117,6 @@ pytestmark = [
     ),
 ]
 
-xfail_handle_register = pytest.mark.xfail(
-    condition=isSupportedFilesystem() and os.environ.get("CI") is not None,
-    raises=cufile.cuFileError,
-    reason="handle_register call fails in CI for unknown reasons",
-)
-
 
 def test_cufile_success_defined():
     """Check if CUFILE_SUCCESS is defined in OpError enum."""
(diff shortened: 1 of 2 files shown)


seberg:
/ok to test [b9f4c54](https://github.com/NVIDIA/cuda-python/pull/2218/commits/b9f4c54ffa4bf3b7efea15b3ddb9512b5b0a7856)

mdboom (approved):
LGTM, but also pinging @sourabgupta3 for viz.

seberg:
@sourabgupta3 if you have any thoughts, happy to follow up.  But I think it's low-risk and I would like to not pile up related PRs.

Beyond using `tmp_dir`, it marks tests as thread-unsafe that use stats counters as those are global and it seems even `stats_start()/stats_stop()` aren't actually thread-safe but that seems unsurprising (i.e. if anything a documentation thing).

cuda_bindings/tests/test_cufile.py
```diff
@@ -118,12 +117,6 @@ def get_tegra_kind():
     ),
 ]
 
-xfail_handle_register = pytest.mark.xfail(
-    condition=isSupportedFilesystem() and os.environ.get("CI") is not None,
-    raises=cufile.cuFileError,
-    reason="handle_register call fails in CI for unknown reasons",
-)
```
leofang, line 125: Sorry for catching up slowly. IIRC this decorator was first added in #1271 (there was a long discussion that seems not decipherable to me now) and is still tracked in #1307. Not sure if this rings any bell to @rwgk? Is there any QA system that would require this decorator?

leofang: My bot told me that if @sourabgupta3 or @rsarpangalav could confirm that the actual root cause of us having to add `xfail_handle_register` was because the CWD being on a non-ext4 mount inside the container, then this PR would allow us to declare #1307 as resolved and close it.

seberg: Sorry for being impatient.  This only affects `os.environ.get("CI") is not None` so I would think QA doesn't matter.

However...  I glanced over `isSupportedFilesystem()` and that was silly.  The temporary folder is not necessarily on the same file system after all.  I'll follow up.

seberg: I opened gh-2233 but maybe with some input (or revisiting) we can also figure out which check is missing from the supported filesystem check that would have made it reliable on CI.

PR #2221, opened 2026-06-15
ci: disable sccache preprocessor cache mode

### LINKED ISSUE #2220: [CI] Preprocessor cache hits = 0% across all Linux wheel builds (sccache + PEP 517 build isolation)
## Summary

`SCCACHE_GHA_USE_PREPROCESSOR_CACHE_MODE=true` is enabled in [build-wheel.yml#L58](https://github.com/NVIDIA/cuda-python/blob/3833b931e534f5af98fbae910a461f77b9ef9b27/.github/workflows/build-wheel.yml#L58), but the preprocessor cache has **never** served a hit in any CI run I've sampled (PR 2218 today, main today 00:23 UTC, main on 2026-05-27). All compiled translation units register as `Preprocessor cache misses`. The downstream object cache still hits ~100% via GHAC, so wheel build correctness is fine — but the preprocessor stage is pure overhead.

## Evidence

From job [`Build linux-64, CUDA 13.3.0 / py3.12`](https://github.com/NVIDIA/cuda-python/actions/runs/27540895877/job/81402152477) (PR #2218):

```
                       cuda.bindings   cuda.core   cuda.core (prev CTK)
Preprocessor hits                  0           0                     0
Preprocessor misses               29          42                    42
Cache hits                        29          42                    40
Cache hits rate              100.00 %    100.00 %               95.24 %
Avg preprocessor miss        0.094 s    17.993 s               14.148 s
```

Same pattern in main run [27516906885 / job 81327125521](https://github.com/NVIDIA/cuda-python/actions/runs/27516906885/job/81327125521) and main run [26482898891 / job 77984135920](https://github.com/NVIDIA/cuda-python/actions/runs/26482898891/job/77984135920) (3 weeks ago) — 0 hits in every case.

## Root cause

pip's per-run build-isolation overlay lives at `/tmp/pip-build-env-<RANDOM>/overlay/...` and that random path lands in `-I` arguments of every `c++` invocation, e.g.:

```
c++ ... -I/tmp/build-env-_283dvkj/include ... -c build/cython/cuda/bindings/runtime.cpp
```

rapidsai/sccache hashes the *raw* compiler arguments into the preprocessor cache key without any basedir stripping for flag values — see [`preprocessor_cache_entry_hash_key` in preprocessor_cache.rs#L493-L497](https://github.com/rapidsai/sccache/blob/v0.14.0-rapids.18/src/compiler/preprocessor_cache.rs#L493-L497). Different random `-I` path each CI run → different preprocessor key → guaranteed miss. The object cache survives because its key is derived from the preprocessor *output*, which is byte-identical between runs (the overlay dir contributes no headers to the expansion).

`SCCACHE_BASEDIRS` would not help either: it strips paths from the *input file* only, not from flag values. (`ccache`'s `CCACHE_BASEDIR` does both — sccache lacks the equivalent.)

## Perf implication

Lookups aren't free. In the linked job:
- `cuda.core`: 42 misses × ~18 s avg lookup ≈ **~12.6 min** wasted
- `cuda.core (prev CTK)`: 42 × ~14 s ≈ ~9.9 min

So preprocessor cache mode currently costs roughly **20 min per Linux job** without delivering a single hit. Multiplied across the build matrix (6 Python versions × 2 arches × 2 CTKs in the wheel jobs), this is non-trivial CI cost.

## Proposed fix

Quick win: stop setting `SCCACHE_GHA_USE_PREPROCESSOR_CACHE_MODE` until the underlying issue is addressed. PR to follow.

Real fix (separate work): switch cibuildwheel to `--no-build-isolation`, so pip's overlay dir never appears in `-I` to begin with. That also unlocks reusing the scikit-build / Cython build dir across runs.

leofang:
Closes #2220.

Preprocessor cache hits are 0% in every CI run sampled, and the lookups cost ~20 min per Linux wheel job. Disable until the underlying pip-build-env path leak is fixed at the source. The object cache via GHAC is unaffected (still hits ~100%).

### DIFF
--- a/.github/workflows/build-wheel.yml
+++ b/.github/workflows/build-wheel.yml
@@ -55,7 +55,6 @@ jobs:
           curl -fsSL "https://github.com/rapidsai/sccache/releases/latest/download/sccache-$(uname -m)-unknown-linux-musl.tar.gz" \
             | sudo tar -C /usr/local/bin -xvzf - --wildcards --strip-components=1 -x '*/sccache'
           echo "SCCACHE_PATH=/usr/local/bin/sccache" >> "$GITHUB_ENV"
-          echo "SCCACHE_GHA_USE_PREPROCESSOR_CACHE_MODE=true" >> "$GITHUB_ENV"
 
       # xref: https://github.com/orgs/community/discussions/42856#discussioncomment-7678867
       - name: Adding addtional GHA cache-related env vars
@@ -190,7 +189,6 @@ jobs:
             ACTIONS_RESULTS_URL=${{ env.ACTIONS_RESULTS_URL }}
             ACTIONS_CACHE_URL=${{ env.ACTIONS_CACHE_URL }}
             ACTIONS_CACHE_SERVICE_V2=${{ env.ACTIONS_CACHE_SERVICE_V2 }}
-            SCCACHE_GHA_USE_PREPROCESSOR_CACHE_MODE=${{ env.SCCACHE_GHA_USE_PREPROCESSOR_CACHE_MODE }}
             SCCACHE_DIR=/host/${{ env.SCCACHE_DIR }}
             SCCACHE_CACHE_SIZE=${{ env.SCCACHE_CACHE_SIZE }}
           CIBW_ENVIRONMENT_WINDOWS: >
@@ -261,7 +259,6 @@ jobs:
             ACTIONS_RESULTS_URL=${{ env.ACTIONS_RESULTS_URL }}
             ACTIONS_CACHE_URL=${{ env.ACTIONS_CACHE_URL }}
             ACTIONS_CACHE_SERVICE_V2=${{ env.ACTIONS_CACHE_SERVICE_V2 }}
-            SCCACHE_GHA_USE_PREPROCESSOR_CACHE_MODE=${{ env.SCCACHE_GHA_USE_PREPROCESSOR_CACHE_MODE }}
             SCCACHE_DIR=/host/${{ env.SCCACHE_DIR }}
             SCCACHE_CACHE_SIZE=${{ env.SCCACHE_CACHE_SIZE }}
           CIBW_ENVIRONMENT_WINDOWS: >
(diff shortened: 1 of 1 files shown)


leofang:
Observed effect on `Average preprocessor cache miss` (same `Build linux-64, CUDA 13.3.0 / py3.12` ↔ `py3.11` slot before/after; this PR's job ran `py3.11`):

| build step | before (`main` @ `ae1617d`, py3.12) | after (this PR, py3.11) |
|---|---|---|
| Build cuda.bindings wheel | [0.065 s](https://github.com/NVIDIA/cuda-python/actions/runs/27516906885/job/81327125521#step:18:2382) | [0.005 s](https://github.com/NVIDIA/cuda-python/actions/runs/27564848120/job/81486057633#step:18:2378) |
| Build cuda.core wheel | [13.780 s](https://github.com/NVIDIA/cuda-python/actions/runs/27516906885/job/81327125521#step:23:1381) | [0.003 s](https://github.com/NVIDIA/cuda-python/actions/runs/27564848120/job/81486057633#step:23:1378) |
| Build cuda.core wheel (prev CTK) | [7.476 s](https://github.com/NVIDIA/cuda-python/actions/runs/27516906885/job/81327125521#step:44:1387) | [0.002 s](https://github.com/NVIDIA/cuda-python/actions/runs/27564848120/job/81486057633#step:44:1381) |

`Preprocessor cache hits` stays at 0 across the board — preprocessor cache mode is still on for the local-disk tier (sccache's default), but local-only misses are essentially free. The GHA preprocessor tier (which never hit and accounted for all the network round-trips) is what's removed here.

mdboom:
> and the lookups cost ~20 min per Linux wheel job

I think you mean ~20 sec?  Still very worthwhile, just wanted to make sure I understand correctly.

mdboom (approved):
LGTM

leofang:
> I think you mean ~20 sec?  Still very worthwhile, just wanted to make sure I understand correctly.

According to my bot, the unit is minutes (so not a typo for seconds), but "~20 min" was cumulative CPU-time (sum of avg miss × misses across the 3 build steps), though it missed that we build in parallel so the wall-clock save per Linux job is more like ~1–2 min.

Another thing I noticed is that we seem to have a bad day again when moving data from github to on-prem CI...

PR #2222, opened 2026-06-15
cuda.core: validate accessed_by kinds before cuMemAdvise (Glasswing V19.1)

Andy-Jost:
## Summary

Addresses Glasswing finding V19.1 (NVBUG 6268913): the `ManagedBuffer.accessed_by` setter validated element types up front but deferred location-kind checks to `_advise_one` inside the update loop. A bulk assignment containing an invalid kind (e.g. `Host(numa_id=...)`, which maps to `host_numa`) could unset valid entries before raising, leaving torn driver state.

## Changes

- `cuda_core/cuda/core/_memory/_managed_buffer.py`: dry-run `_coerce_location` and reject unsupported kinds for `CU_MEM_ADVISE_SET_ACCESSED_BY` before any `cuMemAdvise` calls
- `cuda_core/tests/memory/test_managed_ops.py`: add `test_accessed_by_set_assignment_validates_kind_before_mutation`

## Test Coverage

- `test_accessed_by_set_assignment_validates_kind_before_mutation` — invalid bulk assignment raises without removing previously applied `accessed_by` entries

## Related Work

- NVIDIA/cuda-python-private#380 (Glasswing V19.1, NVBugs [6268913](https://nvbugspro.nvidia.com/bug/6268913))
- Part of Glasswing audit umbrella NVIDIA/cuda-python-private#358

### DIFF
--- a/cuda_core/cuda/core/_memory/_managed_buffer.py
+++ b/cuda_core/cuda/core/_memory/_managed_buffer.py
@@ -9,6 +9,7 @@ from typing import TYPE_CHECKING, Any
 from cuda.core._device import Device
 from cuda.core._host import Host
 from cuda.core._memory._buffer import Buffer
+from cuda.core._memory._managed_location import _coerce_location
 from cuda.core._memory._managed_memory_ops import (
     _advise_one,
     _do_single_discard_prefetch_py,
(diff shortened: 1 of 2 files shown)


cuda_core/cuda/core/_memory/_managed_buffer.py
```diff
@@ -227,6 +228,11 @@ def accessed_by(self, locations: Iterable[Device | Host]) -> None:
             if not isinstance(loc, (Device, Host)):
                 raise TypeError(f"accessed_by entries must be Device or Host, got {type(loc).__name__}")
             target.add(loc)
+        for loc in target:
+            spec = _coerce_location(loc)
+            assert spec is not None
```
mdboom, line 233: Will Glasswing complain that this goes away in `-O` mode (as it did in another issue)?

mdboom (approved):
LGTM

cuda_core/cuda/core/_memory/_managed_buffer.py
Andy-Jost: It appears that Glasswing only flags issues when it can put together a trace that involves crossing a security boundary, so it won't flag every use of `assert`.

In the other issue (cuda-python-private#382 -- Glasswing V3.2, NVBUG 6268893), it flagged that assert because the data it checked came in through pickle and was therefore untrusted.

PR #2229, opened 2026-06-16
tests: Mark tests as unsafe or limit number of threads.

seberg:
This is straight forward (but that doesn't mean we may not want to skip some of) and does two things:
1. Running with many threads, the IPC tests run out of fd's and get very slow, so limit them to some arbitrary value 4/8.
2. Mark tests as thread-unsafe (I'll add some inline comments).

For thread-unsafe markers should sanity check if it looks like it should be thread-safe we may need either explicit tests or fix things.

### DIFF
--- a/cuda_bindings/tests/nvml/test_device.py
+++ b/cuda_bindings/tests/nvml/test_device.py
@@ -120,6 +120,7 @@ def test_read_prm_counters(all_devices):
         assert len(read_counters) == 5
 
 
+@pytest.mark.thread_unsafe(reason="API appears to be thread-unsafe (2026-06)")
 def test_read_write_prm(all_devices):
     for device in all_devices:
         # Docs say supported in BLACKWELL or later
--- a/cuda_bindings/tests/nvml/test_init.py
+++ b/cuda_bindings/tests/nvml/test_init.py
@@ -42,6 +42,7 @@ def test_devices_are_the_same_architecture(all_devices):
 
 
 @pytest.mark.skipif(sys.platform == "win32", reason="Test not supported on Windows")
+@pytest.mark.thread_unsafe(reason="nvml init affects other threads")
 def test_init_ref_count():
     """
     Verifies that we can call NVML shutdown and init(2) multiple times, and that ref counting works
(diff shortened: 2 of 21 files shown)


cuda_bindings/tests/nvml/test_device.py
```diff
@@ -120,6 +120,7 @@ def test_read_prm_counters(all_devices):
         assert len(read_counters) == 5
 
 
+@pytest.mark.thread_unsafe(reason="API appears to be thread-unsafe (2026-06)")
```
seberg, line 123: This felt fine to me, as it writes the counter but if it doesn't seem good it could be an upstream issue.  (CI ran into a segfault, IIRC.)

cuda_pathfinder/tests/test_find_nvidia_binaries.py
```diff
@@ -173,6 +173,7 @@ def test_find_binary_cache_negative_result(monkeypatch, mocker):
 
 
 @pytest.mark.usefixtures("clear_find_binary_cache")
+@pytest.mark.thread_unsafe(reason="functools.cache may replace entry.")
```
seberg, line 176: This tests identity on `Path` objects, which isn't a given with `functools.cache` right now.  But it felt harmless enough to just ignore here.

cuda_bindings/tests/nvml/test_init.py
```diff
@@ -42,6 +42,7 @@ def get_architecture_name(arch):
 
 
 @pytest.mark.skipif(sys.platform == "win32", reason="Test not supported on Windows")
+@pytest.mark.thread_unsafe(reason="nvml init affects other threads")
```
mdboom, line 45: Definitely true.

mdboom (approved):

seberg:
/ok to test [1f7783f](https://github.com/NVIDIA/cuda-python/pull/2229/commits/1f7783f9f3d0a581553fabfde47cdc6a34680f4d)

seberg:
Putting this in then, I looked through once more and to me the skips seem reasonable:
* `nvml.device_read_write_prm_v1` probably being thread-unsafe seems not so weird.
* OpenGL tests also feel OK to (as the context is global)

PR #2230, opened 2026-06-16
TST: Pin test dependencies to exact versions

mdboom:
This is in response to #2227 causing an unwelcome surprise when pytest was updated.  It was fine at the time, but this kind of thing gets really annoying during release time.

There is no reason not to pin our test dependencies, and let dependabot handle updating them so we can deal with any breakages when we want to, not to urgently unblock CI.

## Summary

- Pin all test `[dependency-groups]` specifiers to exact `==` versions in `cuda_bindings`, `cuda_core`, and `cuda_pathfinder` so CI runs against a known-good set of packages
- Add three `pip` ecosystem entries to `.github/dependabot.yml` (one per subproject) so Dependabot opens PRs when newer PyPI versions are available

## Notes

- `cython` and `setuptools` are intentionally left as range specifiers (`>=3.2,<3.3` and `>=80`) since they are build-time dependencies with looser compatibility constraints
- `ml-dtypes` in `cuda_core` is left as `>=0.5.4,<0.6.0` for the same reason

## Test plan

- [ ] Verify CI passes with pinned versions
- [ ] Confirm dependabot config parses cleanly (check Actions tab after merge)

### DIFF
--- a/cuda_bindings/pyproject.toml
+++ b/cuda_bindings/pyproject.toml
@@ -48,7 +48,7 @@ test = [
     "setuptools>=80.0.0",
     # TODO: remove the Python 3.15 guard once 3.15 is officially supported
     "matplotlib==3.10.9; python_version < '3.15'",
-    "numpy==2.4.6",
+    "numpy==2.2.6",
     "pytest==9.1.0",
     "pytest-benchmark==5.2.3",
     "pytest-repeat==0.9.4",
--- a/cuda_bindings/pyproject.toml
+++ b/cuda_bindings/pyproject.toml
@@ -48,7 +48,7 @@ test = [
     "setuptools>=80.0.0",
     # TODO: remove the Python 3.15 guard once 3.15 is officially supported
     "matplotlib>=3.5.0,<=3.10.9; python_version < '3.15'",
-    "numpy>=1.21.1,<=2.2.6",
+    "numpy>=1.21.1,<=2.5.0",
     "pytest==9.1.0",
     "pytest-benchmark==5.2.3",
     "pytest-repeat==0.9.4",
--- a/cuda_bindings/pyproject.toml
+++ b/cuda_bindings/pyproject.toml
@@ -47,8 +47,8 @@ test = [
     "cython>=3.2,<3.3",
     "setuptools>=80.0.0",
     # TODO: remove the Python 3.15 guard once 3.15 is officially supported
-    "matplotlib==3.10.9; python_version < '3.15'",
-    "numpy==2.2.6",
+    "matplotlib; python_version < '3.15'",
+    "numpy",
     "pytest==9.1.0",
     "pytest-benchmark==5.2.3",
     "pytest-repeat==0.9.4",
--- a/cuda_bindings/pyproject.toml
+++ b/cuda_bindings/pyproject.toml
@@ -47,7 +47,7 @@ test = [
     "cython>=3.2,<3.3",
     "setuptools>=80.0.0",
     # TODO: remove the Python 3.15 guard once 3.15 is officially supported
-    "matplotlib==3.11.0; python_version < '3.15'",
+    "matplotlib==3.10.9; python_version < '3.15'",
     "numpy==2.4.6",
     "pytest==9.1.0",
     "pytest-benchmark==5.2.3",
--- a/cuda_bindings/pyproject.toml
+++ b/cuda_bindings/pyproject.toml
@@ -47,8 +47,8 @@ test = [
     "cython>=3.2,<3.3",
     "setuptools>=80.0.0",
     # TODO: remove the Python 3.15 guard once 3.15 is officially supported
-    "matplotlib; python_version < '3.15'",
-    "numpy",
+    "matplotlib>=3.5.0,<=3.10.9; python_version < '3.15'",
+    "numpy>=1.21.1,<=2.2.6",
     "pytest==9.1.0",
     "pytest-benchmark==5.2.3",
     "pytest-repeat==0.9.4",
--- a/cuda_bindings/pyproject.toml
+++ b/cuda_bindings/pyproject.toml
@@ -45,15 +45,15 @@ all = [
 [dependency-groups]
 test = [
     "cython>=3.2,<3.3",
-    "setuptools>=77.0.0",
+    "setuptools>=80.0.0",
     # TODO: remove the Python 3.15 guard once 3.15 is officially supported
-    "matplotlib>=3.5.0; python_version < '3.15'",
-    "numpy>=1.21.1",
-    "pytest>=6.2.4",
-    "pytest-benchmark>=3.4.1",
-    "pytest-repeat",
-    "pytest-randomly",
-    "pyglet>=2.1.9",
+    "matplotlib==3.11.0; python_version < '3.15'",
+    "numpy==2.4.6",
+    "pytest==9.1.0",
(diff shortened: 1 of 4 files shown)


cuda_bindings/pyproject.toml
```diff
@@ -45,15 +45,15 @@ all = [
...
+    "setuptools>=80.0.0",
     # TODO: remove the Python 3.15 guard once 3.15 is officially supported
-    "matplotlib>=3.5.0; python_version < '3.15'",
-    "numpy>=1.21.1",
-    "pytest>=6.2.4",
-    "pytest-benchmark>=3.4.1",
-    "pytest-repeat",
-    "pytest-randomly",
-    "pyglet>=2.1.9",
+    "matplotlib==3.10.9; python_version < '3.15'",
```
rwgk: WDYT about e.g.

```
"matplotlib>=3.5.0,<=3.10.9; python_version < '3.15'",
```

so that we don't lock in other packages to one very specific version? — I realize it's only for testing, but I wonder if e.g. numba or rapids testing could be affected by our choices here.

mdboom: Why would they test in the same venv as ours?  If they wanted to test against a prerelease of cuda_bindings, only the library dependencies would apply, not the test dependencies.

mdboom: I think the only concern here is that /this/ list is self-consistent, for example, the numpy and matplotlib versions specified here are compatible (which they are).

rwgk: > Why would the test in the same venv as ours?

I think I've seen that (SWQA testing), although digging up what I have in mind isn't easy.

I'm OK to be simple/strict first and see what happens.

cuda_bindings/pyproject.toml
```diff
@@ -45,15 +45,15 @@ all = [
...
     # TODO: remove the Python 3.15 guard once 3.15 is officially supported
-    "matplotlib>=3.5.0; python_version < '3.15'",
-    "numpy>=1.21.1",
-    "pytest>=6.2.4",
-    "pytest-benchmark>=3.4.1",
-    "pytest-repeat",
-    "pytest-randomly",
-    "pyglet>=2.1.9",
+    "matplotlib; python_version < '3.15'",
+    "numpy",
```
rwgk: Could we keep the previous lower bounds here rather than making these bare dependencies?

What I would do:

- Restore the lower bounds, because they encode the minimum supported/tested floor.
- Add upper bounds to prevent surprise future releases from breaking CI.
- Split NumPy by Python version if one cap cannot span the full Python matrix.
- Add a short comment for any Python-version split so it is clear this is compatibility-driven, not arbitrary.

Policy-wise, I think lower bounds and upper bounds serve different purposes: lower bound = minimum supported/tested floor; upper bound or exact pin = CI stability cap. I do not think we should lose the former while adding the latter.

For example, roughly this shape, informed by the CI/branch history so far:

```
"matplotlib>=3.5.0,<=3.10.9; python_version < '3.15'",
"numpy>=1.21.1,<=2.2.6; python_version < '3.11'",
"numpy>=1.21.1,<=2.4.6; python_version >= '3.11'",
```

rwgk (commented):
I think we also need to update the `pixi.toml` files as part of this PR, or at least explicitly call them out as follow-up work.

Right now this PR freezes the test dependencies in the package `pyproject.toml` dependency groups, but the pixi environments still have broad specs for the same test tools. That means some CI/developer test paths can still float even after this PR lands.

The main ones I noticed are:

- `cuda_bindings/pixi.toml`
- `cuda_core/pixi.toml`
- `cuda_pathfinder/pixi.toml`

If the goal is to stop unexpected pytest-side dependency changes from putting us into fire-fighting mode, the pixi-managed test environments should follow the same pinning policy as the `pyproject.toml` test groups, or we should document why they intentionally remain floating.

mdboom:
> If the goal is to stop unexpected pytest-side dependency changes from putting us into fire-fighting mode, the pixi-managed test environments should follow the same pinning policy as the pyproject.toml test groups, or we should document why they intentionally remain floating.

One of the reasons I'm sort of anti-pixi is that it uses the conda (forge) packaging ecosystem which lags behind PyPI.  It's not actually possible to pin these to the same versions right now, for example, because many of them have yet to be released.  This creates a whole other set of fire-fighting surprises.  We should either (a) use pixi everywhere, including in CI or (b) move to PyPI-based packaging everywhere and use `uv`'s tooling rather than pixi.  Supporting both is a major pain point -- one of them is always broken and they can't match.  I'm personally more familiar with (b) and that is where the community is trending, but could be convinced otherwise.

This is a separate issue from needing to support both ecosystems for our users -- we do need to test on both there.

cuda_bindings/pyproject.toml
mdboom: In general, this makes sense. There is no need for the two numpy entries, since the lack of availability of anything greater than 2.2.6 on Python 3.10 will automatically enforce that first rule.

rwgk:
@mdboom I merged main to resolve the two sdist test failures (your PR #2241).

rwgk (approved):
Approving, assuming the CI passes now.

I think the pins solve the immediate CI-stability problem.

One thing we should confirm after this lands is whether Dependabot actually updates PEP 735 `[dependency-groups]` entries. The GitHub docs gpt-5.5 found for pip mention `.txt` files and PEP 621 `pyproject.toml`, but not dependency groups, so I’d consider the automatic-update path as unproven until we see an actual Dependabot PR.

xref: https://docs.github.com/en/code-security/reference/supply-chain-security/supported-ecosystems-and-repositories#pip-and-pip-compile

leofang:
> One thing we should confirm after this lands is whether Dependabot actually updates PEP 735 `[dependency-groups]` entries. The GitHub docs gpt-5.5 found for pip mention `.txt` files and PEP 621 `pyproject.toml`, but not dependency groups, so I’d consider the automatic-update path as unproven until we see an actual Dependabot PR.

I manually triggered the dependabot updates (#2246, #2247). I think we need to update `.github/dependabot.yml` to tell it to monitor the pip packaging ecosystem as well. Right now it only cares about GHA updates.

mdboom:
> > One thing we should confirm after this lands is whether Dependabot actually updates PEP 735 `[dependency-groups]` entries. The GitHub docs gpt-5.5 found for pip mention `.txt` files and PEP 621 `pyproject.toml`, but not dependency groups, so I’d consider the automatic-update path as unproven until we see an actual Dependabot PR.
> 
> I manually triggered the dependabot updates (#2246, #2247). I think we need to update `.github/dependabot.yml` to tell it to monitor the pip packaging ecosystem as well. Right now it only cares about GHA updates.

I think this PR does that already:

https://github.com/NVIDIA/cuda-python/pull/2230/changes#diff-dd4fbda47e51f1e35defb9275a9cd9c212ecde0b870cba89ddaaae65c5f3cd28R8

mdboom:
I'm going to merge this.  @leofang might be right that the dependabot config isn't complete -- but I don't think there is any way to know other than merging and testing.  That's separate from pinning the dependencies itself, of course, and we can always roll-back the dependabot changes later if we need to.

mdboom:
By the flood of dependabot PRs today, it does look like it is working.  I'll handle them -- not all of them are updates that we want.

leofang:
> I think this PR does that already:

Ah sorry, I missed it!

PR #2231, opened 2026-06-16
pathfinder: add sm_arch bitcode lookup

### LINKED ISSUE #2200: [FEA]: Add SM arch detection to CUDA-Pathfinder for LLVM bitcode files

aryanputta:
## Summary

- add optional `sm_arch` lookup to `locate_bitcode_lib()` and `find_bitcode_lib()`
- normalize accepted SM forms like `sm90`, `sm90a`, `sm_90`, and `90` to the package filename suffix
- enable arch-specific lookup for `nvshmem_device` as `libnvshmem_device_{sm_arch}.bc`
- cover normalization, search order, invalid input, and unsupported-library behavior in pathfinder tests

Fixes #2200

## Validation

- `python3 -m py_compile cuda/pathfinder/_static_libs/find_bitcode_lib.py tests/test_find_bitcode_lib.py`
- `git diff --check`
- `awk 'length($0) > 120 { print FILENAME ":" FNR ":" length($0) }' cuda_pathfinder/cuda/pathfinder/_static_libs/find_bitcode_lib.py cuda_pathfinder/tests/test_find_bitcode_lib.py`
- `pre-commit run ruff-check --files cuda_pathfinder/cuda/pathfinder/_static_libs/find_bitcode_lib.py cuda_pathfinder/tests/test_find_bitcode_lib.py`
- `pre-commit run ruff-format --files cuda_pathfinder/cuda/pathfinder/_static_libs/find_bitcode_lib.py cuda_pathfinder/tests/test_find_bitcode_lib.py`
- `graphify update .`

`python3 -m pytest tests/test_find_bitcode_lib.py` is blocked in this local macOS environment during collection before these tests run. Importing `cuda.pathfinder` selects the Linux dynamic-loader module and fails on missing `dlinfo` from `/usr/lib/libdl.dylib`.

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_static_libs/find_bitcode_lib.py
+++ b/cuda_pathfinder/cuda/pathfinder/_static_libs/find_bitcode_lib.py
@@ -76,15 +76,13 @@ def _normalize_sm_arch(sm_arch: str | None) -> str | None:
 
     if not isinstance(sm_arch, str):
         raise ValueError(
-            "Invalid sm_arch value. Expected None or an NVIDIA SM architecture like "
-            "'sm90', 'sm90a', 'sm_90', or '90'."
+            "Invalid sm_arch value. Expected None or an NVIDIA SM architecture like 'sm90', 'sm90a', 'sm_90', or '90'."
         )
 
     match = re.fullmatch(r"(?:sm_?)?([0-9]{2,3}[a-z]?)", sm_arch)
     if match is None:
         raise ValueError(
-            "Invalid sm_arch value. Expected None or an NVIDIA SM architecture like "
-            "'sm90', 'sm90a', 'sm_90', or '90'."
+            "Invalid sm_arch value. Expected None or an NVIDIA SM architecture like 'sm90', 'sm90a', 'sm_90', or '90'."
         )
     return f"sm{match.group(1)}"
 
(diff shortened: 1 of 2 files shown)


rwgk:
There is already #2201. It's currently on hold, pending internal discussions. I'll close this PR.

BTW: I started reviewing your #2196 but was too busy to finish the review. One high-level observation: it's a good start, but fixing up the Gen-AI-isms (e.g. over-explaining things that are not relevant beyond the scope of the PR) directly is far easier these days than crafting review comments. I'm still figuring out how to deal with that best.

PR #2232, opened 2026-06-16
[no-ci] Add test authorship tags to AGENTS.md

rwgk:
## Summary

This adds repository-wide `AGENTS.md` guidance for marking the provenance of newly added unit tests with lightweight comment tags.

The tag system is intentionally minimal. The design goal is to keep agents and humans choosing from no more than three clear states:

- `AGENT-AUTHORED <model>` for tests authored by an agent and not yet materially reviewed or rewritten by a human.
- `HUMAN-REVIEWED` for agent-authored tests that a human has materially reviewed or rewritten.
- `HUMAN-AUTHORED` for tests authored by a human, or rewritten enough that the authorship is primarily human.

The guidance keeps tags local to the tests they describe: tags should be placed immediately above each test function, with class-level tags only when every test method in the class has the same provenance. It also says to use at most one authorship tag per test and to treat missing tags as legacy or unknown provenance rather than implicit human authorship.

## Relationship to #2245

An alternative pytest marker-based version is available in #2245. That version uses formal `pytest.mark` markers and registers them in the pytest config roots so tests can be selected by provenance. This PR keeps the simpler comment-tag approach for direct comparison.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -116,9 +116,11 @@ repos/{owner}/{repo}/milestones --jq '.[].title'`, and pick the best match.
 ## Test authorship tags
 
 Use these tags to make the provenance of newly added unit tests explicit. Use
-the host language's comment syntax, and place the tag immediately above the
-test function, test class, or at the top of a new test file when the whole file
-has the same provenance.
+the host language's comment syntax, and place the tag immediately above each
+test function. A class-level tag is acceptable only when every test method in
+the class has the same provenance. Do not use file-level tags for authorship
+provenance; they are too easy to miss in large files and make later per-test
+provenance changes unclear.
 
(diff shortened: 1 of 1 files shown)


mdboom (changes requested):
I love this concept, of course.

Rather than comments, we could use `pytest.mark`, which would then easily provide tools for running a subset of these tests etc.

Here's the info about adding custom markers:

https://docs.pytest.org/en/stable/example/markers.html#registering-markers

rwgk:
> Rather than comments, we could use `pytest.mark`, which would then easily provide tools for running a subset of these tests etc.

That's now implemented as an alternative under PR #2245.

I have a slight preference for #2245 over this PR, mostly because the decorators are formally attached.

rwgk:
Closing after the alternative #2245 was merged.

PR #2233, opened 2026-06-17
test(bindings): Fix unsupported FS check in cufile

seberg:
This is a follow-up from making cufile tests use temporary directories as noted by Leo

I don't think QA is a problem, because the previous xfail was guarded by having `CI` in the environment variables.

However, the `isSupportedFilesystem` was using the wrong directory now as we are now running the test in the temporary directory.

My suspicion is that there is some additional check that would be strictly needed (e.g. to check that it isn't just ext4 but also directly mounted on a local nvme device) but I have not figured out a check for that.

(I don't love `pytest.skip()` use, but no good way around as we need the tempdir.)


cuda_bindings/tests/test_cufile.py
```diff
@@ -78,15 +78,20 @@ def cufileVersionLessThan(target):
         return True  # Assume old version if any error occurs
 
 
-@cache
-def isSupportedFilesystem():
-    """Check if the current filesystem is supported (ext4 or xfs).
+@pytest.fixture(scope="session")
+def skipIfUnsupportedFilesystem(tmpdir_factory):
+    """Fixture that skips if the current filesystem is supported (ext4 or xfs).
```
leofang: Docstring is inverted — body skips when the FS is *not* supported.

```suggestion
    """Fixture that skips if the current filesystem is not supported (ext4 or xfs).
```

cuda_bindings/tests/test_cufile.py
```diff
@@ -78,15 +78,20 @@ def cufileVersionLessThan(target):
...
+
+    The actual requirements are probably both stricter (ext4 was not working on CI previously)
+    and possibly also less strict.
 
     This uses `findmnt` so the kernel's mount table logic owns the decoding of the filesystem type.
     """
-    fs_type = subprocess.check_output(["findmnt", "-no", "FSTYPE", "-T", os.getcwd()], text=True).strip()  # noqa: S603, S607
+    cmd = ["findmnt", "-no", "FSTYPE", "-T", tmpdir_factory.getbasetemp()]
+    fs_type = subprocess.check_output(cmd, text=True).strip()  # noqa S603, S607
```
leofang: Lost the colon during the refactor. Without it, ruff parses this as a bare `# noqa` (suppressing all rules on the line) and the rule codes become trailing text — `RUF100`/strict-noqa configs will flag this.

```suggestion
    fs_type = subprocess.check_output(cmd, text=True).strip()  # noqa: S603, S607
```

cuda_bindings/tests/test_cufile.py
```diff
@@ -195,8 +200,7 @@ def driver(ctx):
     cufile.driver_close()
 
 
-@pytest.mark.skipif(not isSupportedFilesystem(), reason="cuFile handle_register requires ext4 or xfs filesystem")
-@pytest.mark.usefixtures("driver")
+@pytest.mark.usefixtures("driver", "skipIfUnsupportedFilesystem")
```
leofang, line 211: `usefixtures` string order is *not* an execution-order guarantee — pytest orders fixtures by scope and dependency graph, not by position in the marker. With `("driver", "skipIfUnsupportedFilesystem")` `driver` can (and likely will) set up first, so we'll run `cufile.driver_open()` on every test we then skip, and if `driver` setup itself touches the FS on an unsupported mount it'll error instead of cleanly skipping.

Suggest making the skip a *dependency* of `driver` (and `stats`) so it always resolves first:

```python
@pytest.fixture
def driver(ctx, skipIfUnsupportedFilesystem):
    ...

@pytest.fixture
def stats(..., skipIfUnsupportedFilesystem):
    ...
```

Then `skipIfUnsupportedFilesystem` can be dropped from each test's `usefixtures` list — the dependency carries it transitively.

cuda_bindings/tests/test_cufile.py
```diff
@@ -78,15 +78,20 @@ def cufileVersionLessThan(target):
...
+def skipIfUnsupportedFilesystem(tmpdir_factory):
+    """Fixture that skips if the current filesystem is supported (ext4 or xfs).
+
+    The actual requirements are probably both stricter (ext4 was not working on CI previously)
+    and possibly also less strict.
 
     This uses `findmnt` so the kernel's mount table logic owns the decoding of the filesystem type.
     """
-    fs_type = subprocess.check_output(["findmnt", "-no", "FSTYPE", "-T", os.getcwd()], text=True).strip()  # noqa: S603, S607
+    cmd = ["findmnt", "-no", "FSTYPE", "-T", tmpdir_factory.getbasetemp()]
```
leofang, line 90: Nit: `tmpdir_factory.getbasetemp()` returns `py.path.local`. `subprocess` coerces via `os.fspath`, so it works today, but explicit `str(...)` is more defensive against future `py` deprecation.

```suggestion
    cmd = ["findmnt", "-no", "FSTYPE", "-T", str(tmpdir_factory.getbasetemp())]
```

cuda_bindings/tests/test_cufile.py
```diff
@@ -78,15 +78,20 @@ def cufileVersionLessThan(target):
...
+    fs_type = subprocess.check_output(cmd, text=True).strip()  # noqa S603, S607
     logging.info(f"Current filesystem type (findmnt): {fs_type}")
-    return fs_type in ("ext4", "xfs")
+    if fs_type not in ("ext4", "xfs"):
+        pytest.skip("cuFile handle_register requires ext4 or xfs filesystem")
```
leofang: Nit: worth a sentence in the PR body or docstring that on dev machines where the pytest basetemp lands on `tmpfs` (a common default for `/tmp`), the cuFile suite will now silently skip — previously it would attempt to run from CWD. Correct behavior, but a noticeable change for local devs that's worth flagging so nobody is puzzled by 100% skipped tests.

leofang (changes requested):
A few items — main concern is fixture execution ordering (inline below): `usefixtures` order is not an execution-order guarantee, so `driver` can set up before the skip resolves and we'd hit `cufile.driver_open()` on tests we're about to skip (or error in driver setup on an unsupported FS instead of cleanly skipping). Docstring inversion and `# noqa` colon are quick fixes. Two minor nits.

Doesn't address #1307 itself — fine, that's a separate root-cause question we can keep open.

-- Leo's bot

cuda_bindings/tests/test_cufile.py
seberg: while I don't care, I really don't see why it makes sense to be "defensive" about this here.

cuda_bindings/tests/test_cufile.py
seberg: This seems like the bot is not thinking it through?

`driver` is used in basically *all* tests, this just skips some.  If it depended on it, sure we skip some setup work but we would stop running tests that can run just fine?!

cuda_bindings/tests/test_cufile.py
seberg: Hmmmm, but this is a point because the tests *did* pass, but it might be that they are now indeed skipped (sometimes/always?).

I.e. things are working (maybe in cufile compat mode) here and skipping them would be just because of the FS check not being very good.

**EDIT: Sorry, let me just run and see what is the actual situation and then rethink...**

cuda_bindings/tests/test_cufile.py
leofang: It's been a while since I last read this file. My take is that for the tests that do need the `driver` fixture, we want to ensure the execution order to be "check if the test needs to be skipped" -> "driver setup" -> "test" -> "driver teardown", and the bot was cautiously calling out that this order may or may not be preserved by `usefixtures` with multiple args?

cuda_bindings/tests/test_cufile.py
seberg: Hmmmm, OK, maybe mostly for posterity.  The **old** path that was using the current working directory directly failed in this way: https://github.com/NVIDIA/cuda-python/actions/runs/28160946473/job/83403807062?pr=2233#step:31:1364
```
 25-06-2026 09:55:05:66 [pid=4215 tid=4215] ERROR  cufio-udev:67 udev property not found: ID_FS_USAGE vda1
 25-06-2026 09:55:05:66 [pid=4215 tid=4215] ERROR  cufio-fs:970 error getting volume attributes error for device: dev_no: 253:1
 25-06-2026 09:55:05:66 [pid=4215 tid=4215] DEBUG  cufio_core:1350 cuFile DIO status for file descriptor 31 DirectIO not supported
 25-06-2026 09:55:05:66 [pid=4215 tid=4215] DEBUG  cufio:311 cuFileHandleRegister GDS not supported or disabled by config, using cuFile posix read/write with compat mode enabled

** Note: Now trying the compat mode: **

 25-06-2026 09:55:05:67 [pid=4215 tid=4215] DEBUG  cufio-udev:131 numanode string -1
 25-06-2026 09:55:05:67 [pid=4215 tid=4215] DEBUG  cufio-udev:131 numanode string -1
 25-06-2026 09:55:05:67 [pid=4215 tid=4215] DEBUG  cufio-udev:148 device pci path string : 0000:09:00.0->0000:00:02.7
 25-06-2026 09:55:05:67 [pid=4215 tid=4215] DEBUG  cufio-udev:94 sysfs attribute found integrity/device_is_integrity_capable vda1
 25-06-2026 09:55:05:67 [pid=4215 tid=4215] DEBUG  cufio-fs:374 block device vda1 drive integrity check capability not present. Ok
 25-06-2026 09:55:05:67 [pid=4215 tid=4215] INFO   cufio-fs:449 Block dev: /dev/vda1 numa node: -1 pci bridge: 0000:00:02.7 bdf: 0000:09:00.0
 25-06-2026 09:55:05:67 [pid=4215 tid=4215] INFO   cufio-udev:99 sysfs attribute not found device/transport vda1
 25-06-2026 09:55:05:67 [pid=4215 tid=4215] INFO   cufio-udev:99 sysfs attribute not found wwid vda1
 25-06-2026 09:55:05:67 [pid=4215 tid=4215] DEBUG  cufio-udev:94 sysfs attribute found queue/logical_block_size vda1

** NOTE: And this seems to be the problem (But locally I think `ID_FS_USAGE` doesn't exist either, but not 100% sure. **

 25-06-2026 09:55:05:67 [pid=4215 tid=4215] ERROR  cufio-udev:67 udev property not found: ID_FS_USAGE vda1
 25-06-2026 09:55:05:67 [pid=4215 tid=4215] ERROR  cufio-fs:970 error getting volume attributes error for device: dev_no: 253:1
 25-06-2026 09:55:05:67 [pid=4215 tid=4215] ERROR  cufio-obj:251 unable to get volume attributes for fd 31
 25-06-2026 09:55:05:67 [pid=4215 tid=4215] ERROR  cufio:333 cuFileHandleRegister error, failed to allocate file object
 25-06-2026 09:55:05:67 [pid=4215 tid=4215] ERROR  cufio:363 cuFileHandleRegister error: file descriptor is not registered
 25-06-2026 09:55:05:104 [pid=4215 tid=4215] TRACE  cufio:895 cuFileDriver closing
```

So what that gives?  I suspect the FS based skipped was always shady because `tmpfs`, etc. work just fine in compat mode and the whole FS based logic makes sense without compat mode?
However, compat mode is tried here and fails in `cwd` (but not in the temp folder which actually reports as `overlayfs`(!?), although I would have thought it is just `/tmp/<something>` but I didn't check yet.)

seberg: I guess ping @sourabgupta3, maybe you have some relatively quick insights?  TBH, I wonder if we shouldn't just delete the check entirely and assume that the CI filesystem failing even with compat mode allowed is a weird niche thing that is hard to check for...

cuda_bindings/tests/test_cufile.py
leofang: Retracting this — I was overcautious. `skipIfUnsupportedFilesystem` is `scope="session"` and `driver` is default function-scope, and pytest guarantees higher-scoped fixtures set up first regardless of `usefixtures` string order. So the skip *does* fire before `driver` setup — no re-ordering needed. Sorry for the noise.

cuda_bindings/tests/test_cufile.py
seberg: I had talked with Sourab a while ago already but didn't immediately follow up. I completely misunderstood the situation.

I'll try to add a new option (just to see if it might work even outside `tmpfs` in which case it seems fine to keep) but the real reason that this works is that CUDA (probably 13.2) simply supports tmpfs, so these changes were fine on those grounds.

seberg:
Ok, not sure what's best.  The temp folder is reported as `overlay` and to get at what's below may be possible but also a bit more complex.

So I just added a "overlay + CI" is OK path, maybe that is fine?  That way the tests are run and pass, but of the check is still what it was.
(On newer versions `tmpfs` may actually work, and it may be related to that.  So I am not quite sure how far back this will work, but I guess CI would fail if that is a problem.)

seberg:
OK, a last update to make sure this should work if backported to 12.9 (by just not running there, because we will see `tmpfs` and then skip as it is unsupported which should be right -- note that it previously never ran, `main` is in a state where things might fail if backported?).

PR #2234, opened 2026-06-17
cuda.core: document the canonical resource-lifetime pattern in AGENTS.md

rparolin:
## Summary

Documents the canonical CUDA resource-lifetime pattern in `cuda_core/AGENTS.md`,
so future contributions route resource ownership through the existing C++
`std::shared_ptr` handle layer instead of re-implementing raw-handle +
`__dealloc__` ownership.

This codifies the implementation guidance from the texture/surface review
(#2095, design issue #2188) — @Andy-Jost's note that the new types had diverged
from the handle architecture. The fix landed in #2095; this captures the rule so
it isn't re-learned.

## Changes

- New **"Resource lifetime management"** section: store a `*Handle` (not a raw
  handle + `_owning` flag); `close()` is `self._handle.reset()`; no
  `__dealloc__`/`__del__` that calls `cu*Destroy`; structural dependencies
  embedded in the C++ box (not `_parent_ref`/`_source_ref` Python keepalives);
  `create_*_handle` + `get_last_error()`; `TaggedHandle` for identical integer
  handle types (`CUdeviceptr`/`CUtexObject`/`CUsurfObject`). Points at
  `_cpp/DESIGN.md`, `_cpp/REGISTRY_DESIGN.md`, and `_stream`/`_event`/`_buffer`
  as reference consumers.
- Amends the **"Semantics"** guidance, which previously said "use `__dealloc__`
  where possible" (the instruction that led to the divergent pattern), to scope
  `__dealloc__`/`__del__` to non-resource finalization and defer resource
  cleanup to the handle deleter.

Docs only; no code or behavior change.

### DIFF
--- a/cuda_core/AGENTS.md
+++ b/cuda_core/AGENTS.md
@@ -24,6 +24,49 @@ This file describes `cuda_core`, the high-level Pythonic CUDA subpackage in the
 - **Build backend**: `build_hooks.py` handles Cython extension setup and build
   dependency wiring.
 
+## Resource lifetime management
+
+CUDA resources (contexts, streams, events, memory pools, device pointers,
+libraries, kernels, graphs, arrays, textures/surfaces, etc.) are owned through a
+single C++ `std::shared_ptr`-based handle layer -- never by raw driver handles
+stored in Python objects. This is the canonical pattern; see
+`cuda/core/_cpp/DESIGN.md` and `cuda/core/_cpp/REGISTRY_DESIGN.md`, with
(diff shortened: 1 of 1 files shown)


mdboom:
As a test of this content, I had Claude implement this approach for the handling of "event sets" in `cuda.core.system` and put up an entirely edited draft here: https://github.com/NVIDIA/cuda-python/pull/2240

I suspect it is not 100% how we want it.

We should review that PR, and tweak the content in this PR to reflect the feedback, then try again and see if the agent does a better job.

PR #2236, opened 2026-06-17
Include cuda.core C++ .h files in packages

fallintoplace:
## Summary
- Include `_cpp/*.h` in the cuda-core source distribution manifest.
- Include `_cpp/*.h` in setuptools package data so wheel builds carry the C++ helper header.

## Why
`cuda/core/_tensor_map.pyx` includes `_cpp/tensor_map_cccl.h`, but the `_cpp` packaging metadata only included `.hpp` headers. Builds from packaged artifacts could miss the required header.

## Checks
- `cd cuda_core && uv run --no-project --with build python -m build --sdist --outdir ../cuda-python-package-check`
- `tar -tzf ../cuda-python-package-check/cuda_core-*.tar.gz | rg "cuda/core/_cpp/tensor_map_cccl\.h$"`
- `cd cuda_core && uv run --no-project --with "setuptools>=80" --with "setuptools-scm[simple]>=8" --with "Cython>=3.2,<3.3" --with "cuda-pathfinder>=1.5" python setup.py build_py --build-lib ../cuda-python-build-py-check`
- `test -f ../cuda-python-build-py-check/cuda/core/_cpp/tensor_map_cccl.h`
- `git diff --check -- cuda_core/MANIFEST.in cuda_core/pyproject.toml`


mdboom:
Makes sense to me.  Why did our `sdist` testing not catch this?  Is there something we should improve there?

leofang (approved):
Approving. Quick survey while checking whether this is a real gap (and to answer @mdboom's sdist question).

**What's in `cuda-core 1.0.1` wheel today** (`cp312`, `manylinux_x86_64`; same shape across builds): 0 `.pyx`, 0 `.cpp` (stripped — implementations live in the `.so`); 67 `.pxd`, 15 `.h`/`.hpp` (kept — declaration surface for downstream `cimport`/`#include`). `_cpp/resource_handles.hpp` is shipped; `_cpp/tensor_map_cccl.h` is silently missing.

**Why it hasn't failed:** `test-sdist-linux` builds sdist → wheel-from-sdist, but `setuptools-scm` auto-bundles every git-tracked file into the sdist regardless of `MANIFEST.in`, so the wheel build never sees a missing header. And at runtime the compiled `_tensor_map*.so` already contains the symbol; `_tensor_map.pxd` has no `cdef extern from`, so no downstream `cimport` chain reaches this header either. Purely a packaging inconsistency, not a broken wheel.

**De facto rule this PR upholds:** for `_cpp/` and `_include/`, ship declaration files (`.pxd`, `.h`, `.hpp`) and strip implementations (`.pyx`, `.cpp`). `main` had `"cuda.core._cpp" = ["*.hpp"]`, inconsistent with the sibling `"cuda.core._include" = ["*.h", "*.hpp"]` and dropping `.h` files from `_cpp/`. This PR restores the pattern.

leofang:
/ok to test [fe08af4](https://github.com/NVIDIA/cuda-python/pull/2236/commits/fe08af468ea83f2b12546d79d31209f7ce45ad65)

PR #2240, opened 2026-06-22
Use shared_ptr for NVML event_set resource management

mdboom:
This is a proof-of-concept of the proposed instructions in #2234.  I asked Claude to use those instructions to port the management of NVML event sets to use shared_ptr.  I didn't edit its output in any way.  We should review this and point out any mistakes, and then use that to improve the instructions in #2234.

### DIFF
--- a/cuda_core/cuda/core/_cpp/resource_handles.cpp
+++ b/cuda_core/cuda/core/_cpp/resource_handles.cpp
@@ -1338,6 +1338,77 @@ FileDescriptorHandle create_fd_handle_ref(int fd) {
 }
 
 // ============================================================================
+// NVML event set function pointers and registration
+// ============================================================================
+
+NvmlEventSetFreeFn p_nvmlEventSetFree = nullptr;
+NvmlSysEventSetFreeFn p_nvmlSysEventSetFree = nullptr;
+
+void register_nvml_event_set_fn_pointers(intptr_t event_set_free_fn,
+                                         intptr_t sys_event_set_free_fn) noexcept {
+    p_nvmlEventSetFree = reinterpret_cast<NvmlEventSetFreeFn>(event_set_free_fn);
+    p_nvmlSysEventSetFree = reinterpret_cast<NvmlSysEventSetFreeFn>(sys_event_set_free_fn);
+}
+
+// ============================================================================
+// NVML Event Set Handles (device-scope)
+// ============================================================================
+
+namespace {
+struct NvmlEventSetBox {
+    NvmlEventSetValue resource;
+};
+}  // namespace
+
+NvmlEventSetHandle create_nvml_event_set_handle(intptr_t handle) {
+    if (!p_nvmlEventSetFree) {
+        return NvmlEventSetHandle{};
+    }
+    auto box = std::shared_ptr<NvmlEventSetBox>(
+        new NvmlEventSetBox{{handle}},
+        [](NvmlEventSetBox* b) {
+            if (p_nvmlEventSetFree && b->resource.raw) {
+                p_nvmlEventSetFree(reinterpret_cast<void*>(b->resource.raw));
+            }
+            delete b;
+        }
+    );
+    return NvmlEventSetHandle(box, &box->resource);
+}
+
+// ============================================================================
+// NVML System Event Set Handles (system-scope)
+// ============================================================================
+
+namespace {
+struct NvmlSysEventSetBox {
+    NvmlSysEventSetValue resource;
+};
+}  // namespace
+
+NvmlSysEventSetHandle create_nvml_sys_event_set_handle(intptr_t handle) {
+    if (!p_nvmlSysEventSetFree) {
+        return NvmlSysEventSetHandle{};
+    }
+    auto box = std::shared_ptr<NvmlSysEventSetBox>(
+        new NvmlSysEventSetBox{{handle}},
+        [](NvmlSysEventSetBox* b) {
+            if (p_nvmlSysEventSetFree && b->resource.raw) {
+                // Matches NVML_STRUCT_VERSION(SystemEventSetFreeRequest, 1):
+                // version = sizeof(struct) | (1 << 24). Both our struct and the
+                // NVML header struct have the same layout ({unsigned int, void*}).
+                NvmlSysEventSetFreeRequest req;
+                req.set = reinterpret_cast<void*>(b->resource.raw);
+                req.version = (unsigned int)(sizeof(NvmlSysEventSetFreeRequest) | (1u << 24u));
+                p_nvmlSysEventSetFree(&req);
+            }
+            delete b;
+        }
+    );
+    return NvmlSysEventSetHandle(box, &box->resource);
+}
+
+// ============================================================================
 // SM resource split wrapper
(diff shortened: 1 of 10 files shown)


cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -1337,6 +1337,77 @@ FileDescriptorHandle create_fd_handle_ref(int fd) {
...
+// NVML Event Set Handles (device-scope)
+// ============================================================================
+
+namespace {
+struct NvmlEventSetBox {
+    NvmlEventSetValue resource;
+};
+}  // namespace
+
+NvmlEventSetHandle create_nvml_event_set_handle(intptr_t handle) {
```
Andy-Jost, line 1363: An open design question: some `create_*` functions take an already-created handle (as this one does), meaning allocation is done in Cython. Others take construction arguments and create the handle inside.

I have not been able to settle on one approach, but it seems like something we could standardize.

Here, if `p_nvmlEventSetFree` is NULL, or `new` fails, the handle would be leaked unless the caller guards it (granted, those seem like remote possibilities). Construction parameters would resolve that and also enable possible caching or recycling.

On the other hand, accepting construction parameters makes a fatter interface, with potentially many functions that just wrap existing CUDA APIs, 

I don't have a coherent analysis, just a few disconnected thoughts like this.

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -1337,6 +1337,77 @@ FileDescriptorHandle create_fd_handle_ref(int fd) {
...
+    if (!p_nvmlEventSetFree) {
+        return NvmlEventSetHandle{};
+    }
+    auto box = std::shared_ptr<NvmlEventSetBox>(
+        new NvmlEventSetBox{{handle}},
+        [](NvmlEventSetBox* b) {
+            if (p_nvmlEventSetFree && b->resource.raw) {
+                p_nvmlEventSetFree(reinterpret_cast<void*>(b->resource.raw));
```
Andy-Jost, line 1371: Missing `GILReleaseGuard` here.

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -1337,6 +1337,77 @@ FileDescriptorHandle create_fd_handle_ref(int fd) {
...
+        new NvmlSysEventSetBox{{handle}},
+        [](NvmlSysEventSetBox* b) {
+            if (p_nvmlSysEventSetFree && b->resource.raw) {
+                // Matches NVML_STRUCT_VERSION(SystemEventSetFreeRequest, 1):
+                // version = sizeof(struct) | (1 << 24). Both our struct and the
+                // NVML header struct have the same layout ({unsigned int, void*}).
+                NvmlSysEventSetFreeRequest req;
+                req.set = reinterpret_cast<void*>(b->resource.raw);
+                req.version = (unsigned int)(sizeof(NvmlSysEventSetFreeRequest) | (1u << 24u));
+                p_nvmlSysEventSetFree(&req);
```
Andy-Jost, line 1403: ditto regarding `GILReleaseGuard`

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -1337,6 +1337,77 @@ FileDescriptorHandle create_fd_handle_ref(int fd) {
...
+// ============================================================================
+
+NvmlEventSetFreeFn p_nvmlEventSetFree = nullptr;
+NvmlSysEventSetFreeFn p_nvmlSysEventSetFree = nullptr;
+
+void register_nvml_event_set_fn_pointers(intptr_t event_set_free_fn,
+                                         intptr_t sys_event_set_free_fn) noexcept {
+    p_nvmlEventSetFree = reinterpret_cast<NvmlEventSetFreeFn>(event_set_free_fn);
+    p_nvmlSysEventSetFree = reinterpret_cast<NvmlSysEventSetFreeFn>(sys_event_set_free_fn);
+}
```
Andy-Jost, line 1351: Can/should the registration of these NVML symbols be made to match the driver pattern, where `p_*` symbols are resolved via `cuda.bindings` in `_resource_handles.pyx`? The NVML initialization adds a wrinkle so it's not clear to me what's best.

cuda_core/cuda/core/_cpp/resource_handles.hpp
```diff
@@ -661,6 +715,14 @@ inline std::intptr_t as_intptr(const FileDescriptorHandle& h) noexcept {
...
 }
 
+inline std::intptr_t as_intptr(const NvmlEventSetHandle& h) noexcept {
+    return h ? h->raw : 0;
+}
+
+inline std::intptr_t as_intptr(const NvmlSysEventSetHandle& h) noexcept {
+    return h ? h->raw : 0;
+}
+
```
Andy-Jost, line 725: I was expecting `as_cu`, `as_intptr`, and `as_py` to always come as a set. The NVML bindings don't appear to follow that pattern. When all three definitions collapse to the same thing, I wonder if it's better to still define each of the accessor functions. That's a genuine question: I don't know the answer.

cuda_core/cuda/core/system/_event.pxi
```diff
@@ -116,14 +114,15 @@ cdef class DeviceEvents:
...
         self._device_handle = device_handle
-        self._event_set = nvml.event_set_create()
-        # If this raises, the event needs to be freed and this is handled by
-        # this class's __dealloc__ method.
-        nvml.device_register_events(self._device_handle, event_bitmask, self._event_set)
+        cdef intptr_t raw_set = nvml.event_set_create()
+        # If device_register_events raises, create_nvml_event_set_handle already
+        # owns the handle and its shared_ptr deleter will free it.
+        self._h_event_set = create_nvml_event_set_handle(raw_set)
+        nvml.device_register_events(self._device_handle, event_bitmask, raw_set)
```
Andy-Jost, line 121: This is where I would expect to see `as_py(self._h_event_set)` in place of `raw_set`. Not because it changes anything functionally but because it promotes a consistent rule: "to dereference a handle for passing to a Python function, use `as_py`."

cuda_core/cuda/core/system/_event.pxi
```diff
@@ -167,4 +166,4 @@ cdef class DeviceEvents:
         :class:`cuda.core.system.GpuIsLostError`
             If the GPU has fallen off the bus or is otherwise inaccessible.
         """
-        return EventData(nvml.event_set_wait_v2(self._event_set, timeout_ms))
+        return EventData(nvml.event_set_wait_v2(as_intptr(self._h_event_set), timeout_ms))
```
Andy-Jost, line 169: Similar to the above comment. My mental model:

- for hash, printf, repr, etc. --> `as_intptr`
- for passing to Python or returning to a user --> `as_py`
- for passing to Cython --> `as_cu`

Maybe it makes sense to do something different for NVML; or, is it better to have consistency at every call site?

mdboom:
Closing as WONTFIX.  The generator has moved a long way since this.

Issue #1719, opened 2026-03-04
[FEA]: Add context managers to objects where resource cleanup is required


mdboom:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

cuda.core

### Is your feature request related to a problem? Please describe.

[#1711 illustrates](https://github.com/NVIDIA/cuda-python/pull/1711#pullrequestreview-3890441760) a number of places where users are required to perform explicit cleanup of objects, by updating the examples to use `try`/`finally` and performing cleanup within the `finally`.

#1586 is related and probably a part of this work, but not as broad as this.

### Describe the solution you'd like

We should implement context managers on any objects that need an explicit `close` or other resource revocation.

This includes at least Graph, GraphBuilder, Stream, possibly others.

Once these context managers are added, we should update any examples/docs/tests to use them.

### Describe alternatives you've considered

I have read of resistance to this idea elsewhere based on the performance issues with `contextlib.contextmanager`, and if we provide them, users will use them over manual resource deallocation and pay the performance penalty everywhere.

However, context managers as a general concept (implemented with `__enter__` and `__exit__`) do not have a major performance problem relative to correctly written code with a `try`/`finally`, as far as I know.  I'm happy to be corrected on that, if there is in fact a performance problem vs. /correct/ code.

I think supporting context managers would be a major ergonomic improvement over manual tracking of context or even `try`/`finally` (which doesn't compose as well as `with`).

leofang:
~~There are a lot of reference materials that did not get copied into #1586.~~ (**Update:** Bradley actually captured the key issues, my bad.) Let me ping you internally. The tl;dr is we need a design doc and a comprehensive test plan before proceeding.

PR #2243, opened 2026-06-23
Fully clean up buffers/memory resources in tests

seberg:
Maybe the final (small) split-out for the free-threading test PR.

Add explicit sync points and notes around `mr.close()` making sure that there is no MR shutdown with pending async buffer freeing. These can currently cause occasional dead-locks when we have many MRs in parallel.  I believe this is a small upstream issue (625181).

### DIFF
--- a/cuda_core/tests/conftest.py
+++ b/cuda_core/tests/conftest.py
@@ -290,6 +290,9 @@ def ipc_memory_resource(request, ipc_device):
     assert mr.is_ipc_enabled
     yield mr
     mr.close()
+    # TODO(seberg): Make sure the `mr` and it's buffers are fully torn down.
+    # May be unnecessary as `mr.close()` is not parallel with other work.
+    ipc_device.sync()
 
 
 @pytest.fixture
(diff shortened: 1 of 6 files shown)


mdboom:
Somewhat adjacent issue: https://github.com/NVIDIA/cuda-python/issues/1719

mdboom (approved):
I see no reason not to merge this.

As an aside, though...  How did you find all of these and is there a way to enforce that this kind of thing doesn't creep back in in the future?

seberg:
> is there a way to enforce that this kind of thing doesn't creep back in in the future?

That'll be the full changes at https://github.com/NVIDIA/cuda-python/pull/2194 to run with pytest-run-parallel, I was just splitting things out since it seemed a bit easier to move forward.

To be fair, cleanup eventually happens so with single threaded (no pytest-run-parallel) I am not sure this can fail (at least in practice).

/ok to test [b478199](https://github.com/NVIDIA/cuda-python/pull/2243/commits/b478199ef0888a60ac0a2296e190807578d456fe)

seberg:
/ok to test [b478199](https://github.com/NVIDIA/cuda-python/pull/2243/commits/b478199ef0888a60ac0a2296e190807578d456fe)

PR #2244, opened 2026-06-23
Remove upper limit on setuptools-scm

mdboom:
Yesterday, in #2241, an upper limit on the version of setuptools-scm was added to avoid the broken 10.1.1 version.  Since then, 10.1.2 was released which reportedly fixes the bug.  This makes our workaround more specific.


mdboom:
Closing.  This demonstrates that the setuptools-scm bug is still not fixed upstream.

PR #2264, opened 2026-06-25
Update setuptools-scm to use latest, but avoid the broken 10.1 release

mdboom:
Follow-on to PR #2241


mdboom:
The sdist configs are passing here, so this indicates the `setuptools-scm` bug was fixed (in 10.2.0).

rwgk (approved):
Nice!

I added a reference to #2241 in the PR description.

The CI failure appears to be a flake.

Issue #2267, opened 2026-06-25
CI: All linux-aarch64 configurations with CTK 13.x fail the TestPrefetchBatch.test_per_buffer_location test


mdboom:
This breakage happened externally to any change in the codebase.  This run worked yesterday, but re-running it today causes it to fail: https://github.com/NVIDIA/cuda-python/actions/runs/28138378349

It is also failing on most open PRs today.

```
__________________ TestPrefetchBatch.test_per_buffer_location __________________

self = <test_managed_ops.TestPrefetchBatch object at 0xf0e37fc9b880>
location_ops_device = <Device 0 (NVIDIA L4)>
location_ops_mr = <cuda.core._memory._managed_memory_resource.ManagedMemoryResource object at 0xf0e375abf340>

    def test_per_buffer_location(self, location_ops_device, location_ops_mr):
        from cuda.core.utils import prefetch_batch

        device = location_ops_device
        bufs = [location_ops_mr.allocate(_MANAGED_TEST_ALLOCATION_SIZE, stream=device.default_stream) for _ in range(2)]
        stream = device.create_stream()

        prefetch_batch(stream, bufs, [Host(), device])
        stream.sync()

        last0 = _last_prefetch_location(bufs[0])
        last1 = _last_prefetch_location(bufs[1])
>       assert last0 == _HOST_LOCATION_ID
E       assert 0 == -1

bufs       = [<Buffer ptr=0x1910200000 size=4096>, <Buffer ptr=0x1910201000 size=4096>]
device     = <Device 0 (NVIDIA L4)>
last0      = 0
last1      = 0
location_ops_device = <Device 0 (NVIDIA L4)>
location_ops_mr = <cuda.core._memory._managed_memory_resource.ManagedMemoryResource object at 0xf0e375abf340>
prefetch_batch = <cyfunction prefetch_batch at 0xf0e39d903f40>
self       = <test_managed_ops.TestPrefetchBatch object at 0xf0e37fc9b880>
stream     = <Stream handle=0xc60a5a102770 context=0xc60a5789cb50>

tests/memory/test_managed_ops.py:226: AssertionError
```

leofang:
This should be fixed by #2268. What happened yesterday was that all linux-aarch64 runners have a 4K page size on host but the size in the VM was bumped to 64K. Though before 2268 landed the size bump was reverted, so the verification will have to wait until it's bumped next time.

PR #2268, opened 2026-06-25
cuda.core: make managed-prefetch test page-size aware

rparolin:
## Summary

`TestPrefetchBatch.test_per_buffer_location` started failing on the CTK 13.x linux-aarch64 CI configs after the runners moved to the `nvidia-64k` (64 KB page-size) kernel:

```
>       assert last0 == _HOST_LOCATION_ID
E       assert 0 == -1
```

### Root cause

The test hardcoded a 4096-byte allocation and assumed two pooled buffers landed on **separate physical pages**. Managed-memory prefetch and `CU_MEM_RANGE_ATTRIBUTE_LAST_PREFETCH_LOCATION` operate at **page granularity**, and `ManagedMemoryResource` is a pool, so the two `allocate(4096)` calls are packed adjacently (pointers 4 KB apart).

- **4 KB pages:** each 4 KB buffer is its own page → per-buffer prefetch is independent → passes.
- **64 KB pages:** both buffers share one 64 KB page → prefetching `bufs[1]` to the device migrates the whole shared page → querying `bufs[0]` reports device 0 instead of host (-1) → `assert 0 == -1`.

The prefetch itself worked correctly; the test's premise (sub-page allocations are independently prefetchable) only holds when buffer size ≥ page size. The failure is latent on any genuine 64 KB-page platform (Grace / Grace-Blackwell), so reverting the runner kernel only masks it.

### Fix

- Derive `_MANAGED_TEST_ALLOCATION_SIZE` from `mmap.PAGESIZE` so each buffer occupies a full page on every platform (no hardcoded page size).
- Add a precondition asserting the two buffers sit on distinct physical pages, so a future pool-packing change fails loudly instead of silently migrating a shared page.

### Verification

`pytest tests/memory/test_managed_ops.py` → `34 passed, 1 skipped` on an x86 (4 KB page) RTX 5880 Ada box, including the guarded `test_per_buffer_location`.

### DIFF
--- a/cuda_core/tests/memory/test_managed_ops.py
+++ b/cuda_core/tests/memory/test_managed_ops.py
@@ -1,6 +1,8 @@
 # SPDX-FileCopyrightText: Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+import mmap
+
 import pytest
 from helpers.buffers import DummyDeviceMemoryResource, DummyUnifiedMemoryResource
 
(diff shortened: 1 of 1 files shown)


mdboom:
Should fix #2267.

kkraus14 (approved):

PR #2274, opened 2026-06-29
Add Windows arm64 discovery for cuDLA

nikshayshrivastava:
## Description

Adds Windows (arm64) support for `cuda.bindings.cudla`, which was
previously available only on Linux/aarch64 (Tegra).Changes:

- Add Windows discovery for the cuDLA runtime in pathfinder's descriptor
  catalog so `load_nvidia_dynamic_lib("cudla")` can locate `cudla.dll`
- Relax `test_supported_libnames_windows_site_packages_libdirs_ctk_consistency`
  to a subset check: every site-packages entry must map to a supported Windows
  libname, but a supported library may legitimately omit `site_packages_windows`
  (as cuDLA now does).

## Checklist

### DIFF
--- a/cuda_pathfinder/tests/test_load_nvidia_dynamic_lib.py
+++ b/cuda_pathfinder/tests/test_load_nvidia_dynamic_lib.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
 import os
@@ -44,9 +44,13 @@ def test_supported_libnames_linux_site_packages_libdirs_ctk_consistency():
 
 
 def test_supported_libnames_windows_site_packages_libdirs_ctk_consistency():
-    assert tuple(sorted(supported_nvidia_libs.SUPPORTED_LIBNAMES_WINDOWS)) == tuple(
-        sorted(supported_nvidia_libs.SITE_PACKAGES_LIBDIRS_WINDOWS_CTK.keys())
-    )
+    # Not every Windows CTK library ships in a pip wheel (e.g. cudla is loaded
+    # from the local CUDA Toolkit only), so a library may legitimately omit
+    # site_packages_windows. Only assert that every site-packages entry maps to
+    # a supported Windows libname, not the other way around.
+    site_packages_libnames = set(supported_nvidia_libs.SITE_PACKAGES_LIBDIRS_WINDOWS_CTK.keys())
+    supported_libnames = set(supported_nvidia_libs.SUPPORTED_LIBNAMES_WINDOWS)
+    assert site_packages_libnames <= supported_libnames, sorted(site_packages_libnames - supported_libnames)
 
 
 @pytest.mark.parametrize("dict_name", ["SUPPORTED_LINUX_SONAMES", "SUPPORTED_WINDOWS_DLLS"])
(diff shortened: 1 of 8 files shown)


leofang:
I wonder why the path alarm CI would trigger for Nikshay... 🤔

rwgk:
> I wonder why the path alarm CI would trigger for Nikshay... 🤔

**Collaborator permission:** read

is not a trust signal because any github member has read access to this public repo.

For **Author association** we could have:
- `OWNER`: owns the repository
- `MEMBER`: member of the org that owns the repository
- `COLLABORATOR`: invited collaborator on the repository
- `CONTRIBUTOR`: previously committed to the repository
- `NONE`: no detected association with the repository

With authenticated `gh` I can determine conclusively that Nikshay is a `MEMBER`, but unfortunately for our purposes, CONTRIBUTOR clobbers that. — That clobbering behavior could be non-deterministic, although I don't have enough evidence to be sure.

Bottom line: we don't get a lot of false positives, but this is one. Unfortunately I don't have new ideas; we explored a lot already (#2104).

cuda_pathfinder/tests/test_load_nvidia_dynamic_lib.py
```diff
@@ -44,9 +44,13 @@ def test_supported_libnames_linux_site_packages_libdirs_ctk_consistency():
...
-    assert tuple(sorted(supported_nvidia_libs.SUPPORTED_LIBNAMES_WINDOWS)) == tuple(
-        sorted(supported_nvidia_libs.SITE_PACKAGES_LIBDIRS_WINDOWS_CTK.keys())
-    )
+    # Not every Windows CTK library ships in a pip wheel (e.g. cudla is loaded
+    # from the local CUDA Toolkit only), so a library may legitimately omit
+    # site_packages_windows. Only assert that every site-packages entry maps to
+    # a supported Windows libname, not the other way around.
+    site_packages_libnames = set(supported_nvidia_libs.SITE_PACKAGES_LIBDIRS_WINDOWS_CTK.keys())
+    supported_libnames = set(supported_nvidia_libs.SUPPORTED_LIBNAMES_WINDOWS)
+    assert site_packages_libnames <= supported_libnames, sorted(site_packages_libnames - supported_libnames)
```
rwgk: Nit: the explicit assertion message after the comma can be dropped. We already run pytest with `--showlocals`, so `site_packages_libnames` and `supported_libnames` will be shown on failure, and pytest’s set assertion introspection should show the subset mismatch.

rwgk (commented):
This looks good, but did we ever test real cuDLA API calls through `cuda.bindings.cudla` on WoA?

The cuda_bindings unit tests import the module and check enums/POD/API surface, but they don't call wrappers that trigger the lazy `_init_cudla()` path. That's not different from Linux coverage on the available GHA runners here, but we do have internal Orin testing that exercises `_init_cudla()` against real `libcudla.so`.

The WoA evidence so far shows that pathfinder can locate `cudla.dll`, which is useful, but I'd feel much more comfortable merging this PR if we had at least internal evidence that the Windows binding loader path has coverage comparable to the Orin testing.

@isVoid could you maybe work with @nikshayshrivastava on more complete WoA testing.

leofang:
> Bottom line: we don't get a lot of false positives, but this is one. Unfortunately I don't have new ideas; we explored a lot already (#2104).

Yeah having false positives is fine, we mainly don't want false negatives. Let's not sweat on this because it's not worth more time IMHO, also Keith is working on something that could change the situation soon-ish.

cuda_pathfinder/tests/test_load_nvidia_dynamic_lib.py
nikshayshrivastava: fixed in the latest patch

isVoid:
@rwgk 

I validated this PR with @nikshayshrivastava provided test. The bindings builds, and the test passes without issue.

rwgk (approved):
Awesome, thanks both for the validation!

PR #1969, opened 2026-04-23
Find tests that are skipped in every configuration

mdboom:
See #1903

EDIT: Since we now have new configurations with more dependencies and hardware coverage as part of nightly jobs, this task can no longer be integrated into a single workflow as it was before.  It is now just an offline script that anyone can run at any time.  It grabs the latest CI main and CI nightly and analyses the tests from there.  Requires `gh` to be installed and authenticated.  The usual user API token should work -- all of the logs are public and it's not so many requests that rate limiting should matter.

PR #2275, opened 2026-06-29
BUG: Fix test name parsing in find_skipped_tests.py

mdboom:
This regex was changed at the last minute to this PR, but it broke the test name parsing, as @leofang [discovered](https://github.com/NVIDIA/cuda-python/pull/1969#issuecomment-4813266244).  This reverts it back to a working state.


seberg (approved):
Not sure what's up with CI, but I'll just approve if it works for you.
(I guess it might miss that one test that a space in it's parametrization again, but it's also not a big deal.)

PR #2276, opened 2026-06-29
Add pathfinder support for `cupti64_2026.2.1.dll` (released with CTK 13.3.1)

rwgk:
- Add `cupti64_2026.2.1.dll` to the pathfinder descriptor catalog so `load_nvidia_dynamic_lib("cupti")` recognizes the DLL shipped with Windows CTK 13.3.1.
- Add the `cuda-pathfinder` `1.5.6` release notes and version-switcher entry.

___

Minor note: this PR is very similar to PR #1906.

PR #2278, opened 2026-06-29
Fix small bug in nvml.unit_get_devices()

mdboom:
This fixes a small bug in the hand-written wrappers for NVML, that was discovered while rebuilding `cuda.bindings.nvml` by including the upstream header while investigating #1663.  (A nice side-effect of that is that it finds real bugs!).

Unfortunately, I don't know if this test can/will run on any of our hardware.


leofang (approved):

leofang:
Well... I clicked merge on the cybind MR 🤷  Since the Window CI failures are known (due to the new CUPTI release which https://github.com/NVIDIA/cuda-python/pull/2276 fixed), let me just admin-merge (and try to backport).

PR #2280, opened 2026-06-30
cuda.core: graph slot table for node attachment lifetimes

Andy-Jost:
## Summary

CUDA graph nodes frequently reference Python-owned resources — kernel-argument buffers, host-callback functions and their user data, and memcpy/memset operands — but the driver does not keep those resources alive. The driver copies argument *values* into a node at add time; it does not retain the Python wrappers or the device allocations they point at. If the owning object is collected (or a `Buffer` is closed) before the graph is instantiated or launched, the node is left referencing freed memory.

This PR introduces a per-graph *slot table* that binds the lifetime of these attachments to the lifetime of the `CUgraph` itself. Each node gets a small, fixed set of slots; an owner placed in a slot is held until the graph is destroyed. The table is stored on the graph as a CUDA user object, so destroying (or cloning) the graph propagates ownership correctly with no per-node bookkeeping in Python.

It also exposes `GraphBuilder.graph_definition`, completing step 3 of #1330 by giving users an explicit `GraphDefinition` view of a captured graph.

## Design

- **OpaqueHandle**: a type-erased owning handle with two factories — `make_opaque_py` retains a `PyObject*` (incref/decref), and `make_opaque_malloc` owns a `malloc`'d buffer. A slot can therefore hold any kind of owner uniformly.
- **Per-graph slot table**: a map from `CUgraphNode` to a fixed-size array of `OpaqueHandle` slots. It is created lazily on first attachment and retained on the `CUgraph` as a user object (`cuUserObjectCreate` + `cuGraphRetainUserObject` with `MOVE`). When the graph is destroyed, the user object's destructor frees the table and releases every owner it holds. Conditional-branch body graphs (ref handles) get their own table the same way.
- **Attachment API**: `graph_set_slot(graph, node, slot, owner)` installs an owner into a node slot and returns `CUresult` for `HANDLE_RETURN`-style error handling. This replaces the previous approach of attaching an ad-hoc CUDA user object per resource at each node, along with its per-type heap-copy deleters.

## Changes

- Slot-table infrastructure in `resource_handles.{hpp,cpp}` (`OpaqueHandle`, the slot table, lazy creation, `graph_set_slot`) with its Cython surface in `_resource_handles.{pxd,pyx}`.
- Graph nodes in `_graph_node.pyx` — kernel, event-record, event-wait, host-callback, memcpy, and memset — now store their owning handles in node slots. Stream-captured callbacks recover the just-captured host node from `cuStreamGetCaptureInfo` and use the same path; forked builders share the primary builder's graph handle, so their attachments land in the same table.
- Kernel nodes retain the kernel-argument tuple (slot 1) so the Python objects backing the arguments — notably device `Buffer`s — outlive the graph. This is the slot-table port of the user-object fix from #2041.
- `GraphNode.memcpy`/`memset` (and the `GraphDefinition` pass-throughs) now accept either a `Buffer` or a raw `int` address for each operand. A `Buffer` operand is retained at the *allocation* level (its `DevicePtrHandle`), so `close()`/reset cannot free memory the graph still references; a raw `int` behaves exactly as before (caller owns the lifetime), keeping the change backward compatible. Keyword-only `dst_owner`/`src_owner` arguments let callers attach an arbitrary owner to a raw-pointer operand; combining an owner with a `Buffer` operand is rejected.
- `graph/_utils` is renamed to `graph/_host_callback` now that it holds only host-callback machinery, and a shared `_attach_host_callback_owners` unifies the eager (`GN_callback`) and capture (`add_callback`) attachment paths.
- `GraphBuilder.graph_definition` exposes the captured graph as a `GraphDefinition` that shares ownership of the underlying `CUgraph`. State-guard rules: valid on a primary builder only after `end_building()`; valid on a conditional body both before `begin_building()` and after `end_building()`; never valid on a forked builder (access through the primary instead).
- A test-only `_utils/_weak_handles` module provides `weak_handle()` for deterministic, refcount-free lifetime assertions.

## Stream-capture lifetime contract

Operations recorded during stream capture reference caller-owned memory and are not retained, unlike explicit `GraphDefinition` construction. Host callbacks are the one exception: they are retained on both the capture and explicit paths. This contract is documented on `GraphBuilder`.

## Test coverage

- Slot-table lifetime tests for `Buffer` memcpy/memset operands (including clone), and for capture host callbacks retained after dropping their Python references.
- `dst_owner`/`src_owner` retention verified with weakrefs, plus rejection of `Buffer`+owner combinations.
- Device-allocation lifetime tests using `weak_handle()` to confirm an allocation survives `reset`/`close` while a graph still references it.
- `graph_definition` tests: happy path, both hybrid conditional-body flows (populate-via-explicit-API and capture-then-augment), the three error states (forked, capturing, primary pre-capture), and the shared-ownership guarantee (the `GraphDefinition` survives the builder's `close()`).

## Related work

- Builds on the handle-layer plumbing from #2008.
- Ports and generalizes the kernel-argument user-object fix from #2041.
- Completes step 3 of #1330.


rwgk:
I'm trying to help a little bit using gpt-5.5 on my side.

It'd be great to get into the habit of letting our agents add the authorship markers, at least to agent-generated tests.

**Findings**
- P2: `memset` is no longer backward-compatible for existing positional `height`/`pitch` calls. Both `GraphDefinition.memset(...)` and `GraphNode.memset(...)` now insert `*` before `height`, so calls like `g.memset(dst, value, width, height, pitch)` start raising `TypeError`. That is a public API break unless intentional and documented. See `cuda_core/cuda/core/graph/_graph_definition.pyx:158` and `cuda_core/cuda/core/graph/_graph_node.pyx:285`.
- P2: `GraphBuilder.graph_definition` returns a `GraphDefinition` wrapping an empty/reset graph handle after `gb.close()`. `close()` resets `_h_graph` and sets `CLOSED`, but the property does not call `GB_check_open()` or check `CLOSED`, so later `nodes()`, `instantiate()`, etc. hit CUDA with a null graph instead of raising a clear builder-closed error. This only affects accessing the property after close; a view obtained before close is intended to remain valid. See `cuda_core/cuda/core/graph/_graph_builder.pyx:280` and `cuda_core/cuda/core/graph/_graph_builder.pyx:339`.
- P3: The newly added tests lack the repo’s explicit authorship markers. `pytest.ini` registers `agent_authored`, `human_reviewed`, and `human_authored`, and the repo guidance says newly added unit tests should carry one marker immediately above each test. This PR adds many tests without one, starting at `cuda_core/tests/graph/test_graph_builder.py:295`, `cuda_core/tests/graph/test_graph_builder.py:444`, and `cuda_core/tests/graph/test_graph_definition_lifetime.py:634`.

cuda_core/cuda/core/graph/_graph_node.pyx
```diff
@@ -648,14 +694,16 @@ cdef inline KernelNode GN_launch(GraphNode self, LaunchConfig conf, Kernel ker,
...
+    # the device memory they reference.
+    owner = ker._h_kernel
+    HANDLE_RETURN(graph_set_slot(h_graph, new_node, 0, owner))
+    kernel_args = ker_args.kernel_args
     if kernel_args is not None:
-        Py_INCREF(kernel_args)
-        _attach_user_object(as_cu(h_graph), <void*>kernel_args,
-                            <cydriver.CUhostFn>py_object_user_object_destroy)
+        owner = make_opaque_py(kernel_args)
+        HANDLE_RETURN(graph_set_slot(h_graph, new_node, 1, owner))
```
mdboom: One of the things Claude identified here is that if either of these `graph_set_slot` calls fail, the node created with `cuGraphAddKernelNode` is still in the graph, and pointing at Python memory that could be freed from underneath it (since the slot setting thing here failed).

The fix is probably to have a `try/except` that deletes the node and then reraises in the except block.

But whether this is worth the squeeze I leave up to you.

====

cuda_core/cuda/core/graph/_graph_node.pyx:694-706 (and the analogous memset/memcpy/event/callback sites) — a graph node is committed to the driver before its owner is retained, with no rollback on failure.
cuGraphAddKernelNode (or cuGraphAddHostNode/memcpy/memset equivalents) runs first and permanently commits the node — for host-callback nodes this means the driver now holds a raw pointer into Python-owned memory. Only afterward does graph_set_slot(...) via HANDLE_RETURN attempt to retain the owner. If that call fails (ensure_slot_table returns null because the driver lacks cuUserObjectCreate/cuGraphRetainUserObject/cuUserObjectRelease, or either call errors for any other reason, e.g. OOM), an exception is raised — but the node is already permanently in the graph with no owner slot attached. If the caller's only Python reference to the callback/argument tuple is later dropped, a subsequent launch()/instantiate() dereferences a freed pointer. The "old driver" branch of this is unlikely in practice since cuUserObjectCreate (CUDA 11.3+) is present on every driver supporting the package's declared CUDA 12/13 dependency, but genuine transient failures from the two driver calls are not otherwise guarded against, and there's no cleanup/rollback path in any of the call sites.

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -1114,12 +1120,92 @@ LibraryHandle get_kernel_library(const KernelHandle& h) noexcept {
...
+                                 1, CU_USER_OBJECT_NO_DESTRUCTOR_SYNC) != CUDA_SUCCESS) {
+            delete table;  // no user object created; nothing else owns the table
+            return nullptr;
+        }
+        if (p_cuGraphRetainUserObject(box->resource, user_obj, 1,
+                                      CU_GRAPH_USER_OBJECT_MOVE) != CUDA_SUCCESS) {
+            p_cuUserObjectRelease(user_obj, 1);  // drops refcount to 0 -> frees table
+            return nullptr;
+        }
+    }
```
mdboom, line 1193: Identified by Claude (full output below, my commentary above):

If two threads where to act on the same GraphBox here, there is a race condition on setting up its slot_table.  Holding the GIL is probably a simple solution -- locking on the individual `GraphBox` would be more complex.  It might also be possible to recheck that `box->slot_table` is still `NULL` after the GIL is re-aquired here and only overwrite if so.

This is only an issue if two threads were going to work on building the same graph in parallel.  Given how complex graphs are, it's probably ok to say concurrent writing to them isn't supported, but I don't think we actually enforce that anywhere.

====

 cuda_core/cuda/core/_cpp/resource_handles.cpp:1171-1196 (ensure_slot_table) — unsynchronized check-then-act on GraphBox::slot_table racing across threads.
The function checks box->slot_table for null, then releases the GIL around cuUserObjectCreate/cuGraphRetainUserObject, then writes box->slot_table. If two threads add the first node to the same graph concurrently, both can see slot_table == nullptr, both release the GIL, and both race to create and retain a separate GraphSlotTable/user object on the same graph — an unsynchronized concurrent write to a plain pointer (a data race under the C++ memory model) that leaves one of the two tables orphaned as the "cached" pointer while both are still retained on the graph. No lock guards slot_table. This requires genuinely concurrent graph-building from multiple threads on the same builder/graph, which may already be outside the library's supported usage, but nothing documents or enforces that.

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -1114,12 +1120,92 @@ LibraryHandle get_kernel_library(const KernelHandle& h) noexcept {
...
+
+// Slot table layout (internal). Each graph maps CUgraphNode -> a fixed-size
+// array of type-erased owners. The width is the most any single node needs: a
+// kernel node holds its kernel and its packed arguments; a host node holds its
+// callback and the userData. The table is heap-allocated and retained on the
+// graph as a user object, so the driver frees it -- and every owner in it --
+// when the graph is destroyed.
+constexpr std::size_t SLOTS_PER_NODE = 2;
+using NodeSlots = std::array<OpaqueHandle, SLOTS_PER_NODE>;
+using GraphSlotTable = std::map<CUgraphNode, NodeSlots>;
```
mdboom, line 1132: Claude suggested that `std::unordered_map` might be a small performance win here.

====

cuda_core/cuda/core/_cpp/resource_handles.cpp:1132 — GraphSlotTable uses std::map instead of std::unordered_map.
<unordered_map> is already included and used elsewhere in the same file (HandleRegistry, ipc_ptr_cache). graph_set_slot does a red-black-tree lookup/insert (with a heap allocation per node) on every single node addition; an unordered_map<CUgraphNode, NodeSlots> would be O(1) amortized with no custom hash needed (CUgraphNode is a plain pointer). Cheap, low-risk swap for graphs with many nodes.

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -1114,12 +1120,92 @@ LibraryHandle get_kernel_library(const KernelHandle& h) noexcept {
...
+        }
+    }
+    box->slot_table = table;  // non-owning cache; the user object owns it
+    return table;
+}
+
 }  // namespace
 
+OpaqueHandle make_opaque_py(PyObject* obj) {
+    Py_INCREF(obj);
```
mdboom, line 1201: Should we aquire the GIL here (or at least assert that it's already held?).  Usage looks fine at the moment, but it might be easy to misuse.

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -1137,6 +1223,19 @@ GraphHandle create_graph_handle_ref(CUgraph graph, const GraphHandle& h_parent)
     return GraphHandle(box, &box->resource);
 }
 
+CUresult graph_set_slot(const GraphHandle& h_graph, CUgraphNode node,
```
mdboom, line 1226: Most (but not all) of the places where this is called look like:

```
if owner:
    graph_set_slot(graph, node, 0, owner)
```

I wonder if the places where there isn't a NULL check is a bug.

We could modify this function so that `owner == NULL` is accepted, but is a no-op.  That would prevent such bugs in the future.  (On the other hand it may read as less explicit at the call site, so I'm 50/50 on this suggestion).

cuda_core/cuda/core/graph/_graph_builder.pyx
```diff
@@ -282,6 +295,64 @@ cdef class GraphBuilder:
...
+        GraphDefinition
+            A view of the graph being built.
+
+        Raises
+        ------
+        RuntimeError
+            If the builder is forked, currently building, or (for primary
+            builders) has not started building yet.
+        """
+        if self._kind == FORKED:
```
mdboom, line 345: Minor performance nit: If you refactor this as `if/elif` (with the last `and` as a nested `if`) Cython will generate a switch statement.

cuda_core/cuda/core/graph/_graph_definition.pyi
```diff
@@ -85,7 +86,7 @@ class GraphDefinition:
         See :meth:`GraphNode.deallocate` for full documentation.
         """
 
-    def memset(self, dst: int, value, width: int, height: int=1, pitch: int=0) -> MemsetNode:
+    def memset(self, dst: Buffer | int, value, width: int, *, height: int=1, pitch: int=0, dst_owner=None) -> MemsetNode:
```
mdboom: This is technically an API breakage.  Before this change `GraphDefinition.memset(buffer, value, width, height)` would work.  After, this would have to be `GraphDefinition.memset(buffer, value, width, height=height)`.

(Even though this is in a private module, `GraphDefinition` is re-exposed to the public API, so this is public API).

Personally, I think this is a good change and small/obscure enough to not matter much in practice.  At a minimum it needs a "Breaking change" release note.  If we want to be really pedantic, it probably needs a deprecation cycle.

cuda_core/cuda/core/graph/_graph_definition.pyi
```diff
@@ -120,7 +121,7 @@ class GraphDefinition:
             A new EmptyNode that depends on all input nodes.
         """
 
-    def memcpy(self, dst: int, src: int, size: int) -> MemcpyNode:
+    def memcpy(self, dst: Buffer | int, src: Buffer | int, size: int, *, dst_owner=None, src_owner=None) -> MemcpyNode:
```
mdboom, line 124: This one only affects new arguments, so isn't API breakage.

cuda_core/cuda/core/graph/_graph_node.pyi
```diff
@@ -178,13 +179,16 @@ class GraphNode:
             A new FreeNode representing the free operation.
         """
 
-    def memset(self, dst: int, value, width: int, height: int=1, pitch: int=0) -> MemsetNode:
+    def memset(self, dst: Buffer | int, value, width: int, *, height: int=1, pitch: int=0, dst_owner=None) -> MemsetNode:
```
mdboom: Ditto as above: This is technically API breakage.

mdboom (changes requested):
The review is a combination of some things Claude found and some things I found.

I'm putting this as request changes mainly for the API breakage -- we should make a call on how to handle that.

The rest of the comments are "real, but probably very hard to reproduce bugs" or random performance suggestions that I would have approved the PR even if unaddressed.

cuda_core/cuda/core/_cpp/resource_handles.cpp
Andy-Jost: AFAIK all of the entry points in resource_handles are called from Cython with the GIL held. Destroy callbacks acquire the GIL because they can be run at arbitrary times (we also check for interpreter shutdown in those). My feeling is that it's probably not worth the extra complexity to add GIL acquisition to all of the entry points.

cuda_core/cuda/core/graph/_graph_definition.pyi
Andy-Jost: Good catch. For this relatively new API, I'm leaning towards just adding an item in the release note.

leofang: The ship has sailed. We must not break across 1.x.

leofang (changes requested):

leofang:
> * P2: `memset` is no longer backward-compatible for existing positional `height`/`pitch` calls. Both `GraphDefinition.memset(...)` and `GraphNode.memset(...)` now insert `*` before `height`, so calls like `g.memset(dst, value, width, height, pitch)` start raising `TypeError`. That is a public API break unless intentional and documented. See `cuda_core/cuda/core/graph/_graph_definition.pyx:158` and `cuda_core/cuda/core/graph/_graph_node.pyx:285`.

We need to set up our agents to be ultra-sensitive about breaking changes. I am glad that everyone's agent caught this, but it should not be labelled as P2, but instead P0 or blocking.

mdboom:
> I am glad that everyone's agent caught this, but it should not be labelled as P2, but instead P0 or blocking.

For the record, it's the `.pyi` files that make this easy to spot manually.  I keep meaning to add `griffe` to our automation, too, so we would have deterministic API breakage checking, and not rely on our agents to get lucky.

cuda_core/cuda/core/_cpp/resource_handles.cpp
Andy-Jost: Agreed that concurrent graph modifications are outside of the supported concurrency model, so locking is overkill. The CUDA graph-construction calls are already documented as not thread safe (in a dedicated section “Graph object thread safety”) and the custom in Python numeric/GPU libraries like NumPy, PyTorch, CuPy is that synchronization for concurrent mutation falls to the user. Beyond that, cuda.core does not make any concurrency guarantees that I'm aware of.

My understanding, then, is that when making concurrent modifications to cuda.core objects, applications are responsible for providing synchronization.

cuda_core/cuda/core/_cpp/resource_handles.cpp
Andy-Jost: I doubt it. For small maps (N<100) with integer keys, `std::map` wins in my experience.

Andy-Jost:
Thanks for the feedback everyone. The latest changes should address all of the issues.

cuda_core/cuda/core/_cpp/resource_handles.cpp
mdboom: Makes sense.  Maybe an AGENTS.md content would keep flagging this stuff.  But nothing to do on that here.

mdboom (approved):
LGTM

cuda_core/cuda/core/graph/_graph_builder.pyx
```diff
@@ -819,21 +833,27 @@ cdef class GraphBuilder:
...
+        _resolve_host_callback(fn, user_data, &c_fn, &c_user_data, &fn_owner, &data_owner)
 
         with nogil:
             HANDLE_RETURN(cydriver.cuLaunchHostFunc(c_stream, c_fn, c_user_data))
 
+        # Capturing the host function added a node to the graph; it is now the
+        # stream's sole capture dependency. Key the callback's owners to it so
+        # they live in the graph's slot table like any explicitly-added node.
+        cdef cydriver.CUgraphNode host_node = _capture_tail_node(c_stream)
+        _attach_host_callback_owners(self._h_graph, host_node, fn_owner, data_owner)
```
leofang, line 855: Different path from the rollback thread on `_graph_node.pyx:694-706`: the explicit path now has the `try/except: cuGraphDestroyNode` in [`GN_callback`](https://github.com/NVIDIA/cuda-python/blob/eb4991c0dd2277895cfa06a0e4a5ce49500362bf/cuda_core/cuda/core/graph/_graph_node.pyx#L1099-L1110), but the *stream-capture path* here is still bare. `cuLaunchHostFunc` commits the node first, then `_capture_tail_node` (raises on `num_deps != 1`) and `_attach_host_callback_owners` run without rollback. If either throws, the graph now holds a host node whose C fn pointer references Python objects going out of scope → dangling on the next launch. Wrap in a matching `try/except: cuGraphDestroyNode(host_node)` for symmetry?

cuda_core/cuda/core/graph/_graph_builder.pyx
```diff
@@ -913,6 +933,28 @@ cdef inline int _get_capture_info(
...
+    with nogil:
+        IF CUDA_CORE_BUILD_MAJOR >= 13:
+            HANDLE_RETURN(cydriver.cuStreamGetCaptureInfo(
+                stream, &status, NULL, NULL, &deps, NULL, &num_deps))
+        ELSE:
+            HANDLE_RETURN(cydriver.cuStreamGetCaptureInfo(
+                stream, &status, NULL, NULL, &deps, &num_deps))
+    if num_deps != 1:
+        raise RuntimeError(
+            f"expected exactly one capture dependency after a host callback, got {num_deps}")
```
leofang, line 954: Is `num_deps == 1` after `cuLaunchHostFunc` a hard driver invariant across every capture state (post-join forked stream, mid-conditional-body, capture with side deps, ...), or is there a stress case that trips this `RuntimeError`? Asking because raising after the node is already captured lands us in the same dangling-C-pointer state as the rollback question above.

leofang (approved):
LGTM — nice generalization of the slot-table pattern beyond the ad-hoc user-object approach. Two follow-up questions inline; none need to block merge before 1.1.0.

-- Leo's bot

cuda_core/cuda/core/_cpp/resource_handles.cpp
leofang: Coming back to note: we also ship FT builds, so we should expect FT users to come and ask questions here. Is same-graph concurrent mutation being user-responsibility still the intended contract under FT? If yes, we should probably qualify the FT declaration or land it in docs (Mike's AGENTS.md suggestion works) rather than only in review threads. Happy either way — just want it on the record before 1.1.0.

-- Leo's bot

leofang:
Follow-up on deadlock-safety — worth surfacing publicly since we've hit the GIL ↔ CUDA driver lock ordering elsewhere (see [numba-cuda#321](https://github.com/NVIDIA/numba-cuda/pull/321) for the canonical fix pattern).

The pattern we want to avoid: a host callback (via `cuLaunchHostFunc` or `cuGraphAddHostNode`) runs while the driver holds an internal lock, then needs the GIL to call into Python; if another thread is holding the GIL and calling into the driver *without* releasing it first, they deadlock waiting on each other. The same shape applies to any Python `__del__` that fires from a driver callback context.

Checked the PR against that pattern:

- Every driver call the PR touches — `cuLaunchHostFunc`, `cuStreamGetCaptureInfo`, `cuUserObjectCreate`, `cuGraphRetainUserObject`, `cuGraphDestroyNode` — is inside a `with nogil:` block or a `nogil` C helper. So the library never holds the GIL while entering the driver.
- The Python host-callback trampoline (`_py_host_trampoline` with `noexcept with gil`) is preserved verbatim from the prior `_utils.pyx`, just relocated to [`_host_callback.pyx#L23-L28`](https://github.com/NVIDIA/cuda-python/blob/eb4991c0dd2277895cfa06a0e4a5ce49500362bf/cuda_core/cuda/core/graph/_host_callback.pyx#L23-L28). No new callback-crossing shape.
- `cuUserObjectCreate` uses `CU_USER_OBJECT_NO_DESTRUCTOR_SYNC` at [`resource_handles.cpp#L1182-L1184`](https://github.com/NVIDIA/cuda-python/blob/eb4991c0dd2277895cfa06a0e4a5ce49500362bf/cuda_core/cuda/core/_cpp/resource_handles.cpp#L1182-L1184), so `cuGraphDestroy` doesn't synchronize on the slot-table destructor — that avoids the obvious "destroy holds a lock, destructor needs GIL held by a caller waiting on destroy" wedge.
- `py_deleter` at [`resource_handles.cpp#L1138-L1143`](https://github.com/NVIDIA/cuda-python/blob/eb4991c0dd2277895cfa06a0e4a5ce49500362bf/cuda_core/cuda/core/_cpp/resource_handles.cpp#L1138-L1143) acquires the GIL with an interpreter-shutdown check — same discipline as the pre-existing stream/deviceptr deleters.

Where the PR does *modestly* widen the surface: the set of Python objects whose `__del__` fires from a driver callback thread at graph destroy now includes `Buffer` operands passed to `memcpy` / `memset` and any user-supplied `dst_owner` / `src_owner`, on top of the previously-retained kernel-arg tuples and host-callback closures. The user-side contract this creates — *any CUDA work reachable from a retained object's `__del__` must release the GIL before entering the driver* — is the same one that already applies to host callbacks; nothing new in kind, just a slightly larger set of user objects it applies to.

Worth spelling that out in the docstring of `_py_host_trampoline` and/or in an `AGENTS.md` per @mdboom's earlier [suggestion](https://github.com/NVIDIA/cuda-python/pull/2280#discussion_r3516597169), so future contributors (and users retaining custom objects into a graph) hit the guidance without needing to reconstruct the lock-ordering rationale. Not blocking on this PR — happy to open a follow-up issue.

-- Leo's bot

cuda_core/cuda/core/graph/_graph_builder.pyx
Andy-Jost: Good catch — you're right, the capture path is missing the rollback that `GN_callback` has. If `_attach_host_callback_owners` throws, the host node stays in the graph but its owners get dropped, so the callback's C pointer dangles. In practice this only triggers on allocation failure or a driver without user-object support, so I don't think it blocks 1.1.0. I'd also rather not just mirror the `cuGraphDestroyNode` rollback here, since destroying a node mid-capture isn't clearly legal — the cleaner fix is to build the slot table *before* `cuLaunchHostFunc` so the post-commit step can't fail. Tracked in #2317; deferring rather than re-spinning CI this close to release.

cuda_core/cuda/core/graph/_graph_builder.pyx
Andy-Jost: I'm fairly confident `num_deps == 1` holds: right after a single `cuLaunchHostFunc`, the stream's tail becomes exactly that host node, even after a join or inside a conditional body (a later op consumes the prior deps and becomes the sole tail). So I don't expect a stress case to trip the `RuntimeError`. But you're right that the failure mode is bad — raising here lands in the same dangling state as the rollback issue above, so it's really the same root cause. Folded into the same follow-up (#2317): make the post-commit path non-throwing instead of relying on the invariant plus a raise.

Andy-Jost:
Thanks for the careful review, Leo. Both host-callback points are real but low-risk (narrow trigger, and the `num_deps` invariant holds in practice). Since the PR is approved and we're right at the 1.1.0 cut, I've tracked both in a single follow-up issue (#2317) — they share one root cause (post-commit steps on the capture path can throw and leave a committed, unowned node) — rather than start another review+CI cycle. Neither affects the common path.

PR #2282, opened 2026-06-30
test: patch pytest-run-parallel worker context setup

rwgk:
Alternative to #2228 (so we can decide which direction we prefer)

## Summary

Add a cuda-python-local workaround for pytest-run-parallel's current fixture-sharing model so CUDA context setup can run inside each worker thread.

The shared helper patches `pytest_run_parallel.plugin.wrap_function_parallel` when pytest-run-parallel is installed, attaches optional per-worker context factories to selected pytest items, and keeps the behavior a no-op when pytest-run-parallel is absent. The bindings and core test conftests use that helper to set up CUDA context state in the worker thread before the test body runs.

## Background

pytest-run-parallel executes the same pytest item in multiple worker threads, but fixture values are created by pytest before the worker calls and are shared across those threads. That is problematic for CUDA tests because the current CUDA context is thread-local. Until pytest-run-parallel has an upstream extension hook for per-worker setup, this PR monkeypatches the wrapper function as a localized workaround.

This is intended as an alternative to wrapping CUDA tests directly in each conftest. The monkeypatch centralizes the pytest-run-parallel-specific behavior and keeps package conftests focused on package-specific CUDA setup.

## Changes

- Add `cuda_python_test_helpers.pytest_run_parallel` with:
  - idempotent installation of the pytest-run-parallel wrapper patch
  - signature validation for the upstream wrapper entry point
  - isolated kwargs copies for worker and iteration calls
  - per-worker context factory support attached to pytest item functions
  - no-op behavior when pytest-run-parallel is not importable
- Wire `cuda_bindings/tests/conftest.py` to run per-worker CUDA driver/context setup for CUDA-context-dependent bindings tests.
- Exclude cuDLA tests from the bindings worker-context selection because they intentionally override `ctx` as a no-op to avoid CUDA initialization.
- Wire `cuda_core/tests/conftest.py` to run per-worker `init_cuda` setup and replace direct CUDA fixture kwargs that need worker-local values.
- Add unit coverage using a fake pytest-run-parallel module so the monkeypatch mechanics can be tested without requiring pytest-run-parallel or CUDA hardware.

## Notes

This workaround should be removable once pytest-run-parallel exposes an upstream hook that runs inside each worker thread before the test function is invoked.

### DIFF
--- a/ci/tools/tests/test_pytest_run_parallel_patch.py
+++ b/ci/tools/tests/test_pytest_run_parallel_patch.py
@@ -1,0 +1,121 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import sys
+import threading
+import types
+from contextlib import contextmanager
+from pathlib import Path
+
+import pytest
+
+_TEST_HELPERS_ROOT = Path(__file__).resolve().parents[3] / "cuda_python_test_helpers"
+sys.path.insert(0, str(_TEST_HELPERS_ROOT))
+
+from cuda_python_test_helpers.pytest_run_parallel import (
+    install_run_parallel_worker_context_patch,
+    mark_item_for_worker_context,
+)
+
+
+def _install_fake_pytest_run_parallel(monkeypatch):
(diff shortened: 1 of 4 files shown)


rwgk:
Closing in favor of #2228

PR #2284, opened 2026-07-01
Use cuda.pathfinder to locate nvcc for object-code fixture

rwgk:
## Description

This fixes the cuda.core object-code test fixture in environments where `nvcc` is available through `cuda.pathfinder`, but is not directly on `PATH`.

Previously, `get_saxpy_object()` checked whether `cuda.pathfinder` could locate `nvcc`, but then invoked `build_test_binaries.sh` without passing that resolved path through. The shell script still called bare `nvcc`, so the fallback build failed when `nvcc` was discoverable but not on `PATH`.

This PR:

- resolves `nvcc` once with `find_nvidia_binary_utility("nvcc")`;
- passes the resolved path to `build_test_binaries.sh` through `NVCC`;
- keeps the shell script usable by hand by defaulting `NVCC` to `nvcc`.

### DIFF
--- a/cuda_core/tests/test_module.py
+++ b/cuda_core/tests/test_module.py
@@ -189,15 +189,18 @@ def get_saxpy_object():
     obj_path = binaries_dir / "saxpy.o"
 
     if not obj_path.is_file():
-        if find_nvidia_binary_utility("nvcc") is None:
+        nvcc_path = find_nvidia_binary_utility("nvcc")
+        if nvcc_path is None:
             pytest.skip(
                 f"saxpy.o not found at {obj_path} and nvcc is unavailable. "
                 "In CI this is downloaded from the build stage."
             )
+        env = os.environ.copy()
+        env["NVCC"] = nvcc_path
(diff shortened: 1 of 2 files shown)


cuda_core/tests/test_module.py
```diff
@@ -189,15 +189,18 @@ def get_saxpy_object():
...
     if not obj_path.is_file():
-        if find_nvidia_binary_utility("nvcc") is None:
+        nvcc_path = find_nvidia_binary_utility("nvcc")
+        if nvcc_path is None:
             pytest.skip(
                 f"saxpy.o not found at {obj_path} and nvcc is unavailable. "
                 "In CI this is downloaded from the build stage."
             )
+        env = os.environ.copy()
+        env["NVCC"] = nvcc_path
```
leofang, line 199: 👍

lijinf2 (approved):
Good catch. Passing the pathfinder-resolved nvcc through NVCC closes the gap with the shell script's bare nvcc call.

PR #2285, opened 2026-07-01
[codex] Add Apache license compliance metadata

kkraus14:
## What changed

- add a repository-level Apache 2.0 license copy and NVIDIA copyright preambles to the Apache-licensed component license files
- document the DCO sign-off workflow, including `git commit -s` and the resulting `Signed-off-by` trailer
- add DLPack and Deprecated attributions to a `cuda.core` NOTICE file and include that file in source distributions
- clarify the repository's mixed-license scope and link to the third-party notices

## Why

The repository distributes Apache-licensed components alongside proprietary components and includes vendored third-party source in `cuda.core`. The existing files did not centralize all associated license metadata, attribution notices, or the contributor sign-off procedure.

## Impact

This changes documentation and packaging metadata only; runtime behavior is unchanged. Built `cuda-core` packages now carry both `LICENSE` and `NOTICE` under their distribution license metadata.

## Validation

- `git diff --check`
- repository pre-commit hooks; all applicable hooks pass
- verified the Pixi `cuda-core` package build contains both `LICENSE` and `NOTICE` under `dist-info/licenses`

The `lychee` hook was skipped because the current repository configuration references a non-release `rev` and exits before checking links.

### DIFF
--- a/LICENSE-Apache2.txt
+++ b/LICENSE-Apache2.txt
@@ -1,4 +1,6 @@
-Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-FileCopyrightText: Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
 
                                  Apache License
                            Version 2.0, January 2004
--- a/LICENSE-Apache2.txt
+++ b/LICENSE-Apache2.txt
@@ -1,6 +1,4 @@
 Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
-SPDX-FileCopyrightText: Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
-SPDX-License-Identifier: Apache-2.0
 
                                  Apache License
                            Version 2.0, January 2004
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,6 +1,11 @@
-This repository is structured in a way that files are licensed differently
+This repository contains components under different licenses. The
+component-specific license files listed below govern each component. A
+repository-level copy of the Apache 2.0 license is available in
+[`LICENSE-Apache2.txt`](./LICENSE-Apache2.txt) for Apache-licensed content.
+
    - [`cuda.python`](./cuda_python/LICENSE): NVIDIA Software License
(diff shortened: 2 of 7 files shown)


LICENSE-Apache2.txt
leofang, line 1: Q: Don't we have this license for both cuda-core and cuda-pathfinder?
https://github.com/NVIDIA/cuda-python/blob/main/cuda_core/LICENSE
Maybe we should just move it to the top-level and make every project point to the same license?

LICENSE-Apache2.txt
```diff
@@ -0,0 +1,180 @@
+Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-FileCopyrightText: Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
```
leofang: Q: Do we need SPDX headers in a license file? lol

LICENSE.md
```diff
@@ -1,6 +1,11 @@
-This repository is structured in a way that files are licensed differently
+This repository contains components under different licenses. The
+component-specific license files listed below govern each component. A
+repository-level copy of the Apache 2.0 license is available in
+[`LICENSE-Apache2.txt`](./LICENSE-Apache2.txt) for Apache-licensed content.
```
leofang, line 4: ditto

cuda_core/NOTICE
```diff
@@ -0,0 +1,42 @@
...
+
+DLPack
+Copyright (c) 2017 by Contributors
+Licensed under the Apache License, Version 2.0.
+Source: https://github.com/dmlc/dlpack
+
+Deprecated
+Copyright (c) 2017 Laurent LAPORTE
+Licensed under the MIT License.
+Source: https://github.com/tantale/deprecated
```
leofang, line 18: btw I've been meaning to ask... @mdboom @kkraus14 have we filed an OSRB for this? If not, we should do so. For OSS-licensed projects it's quick.

LICENSE-Apache2.txt
kkraus14: I asked the same question internally and the guidance is yes they want this, though we only need the Copyright and not the SPDX identifiers, so I'll fix that.

LICENSE-Apache2.txt
kkraus14: This will be cleaned up in the future regardless, so I'd be more keen to just keep this for now and clean it up when we make the future changes.

cuda_core/NOTICE
kkraus14: We need to do this still.

kkraus14:
Sorry, this was an accidental merge. @leofang if you'd like me to revert, just let me know.

PR #2283, opened 2026-07-01
CI: nightly-cuda-core + nightly-numba-cuda-mlir + PTDS fix

leofang:
## Summary

- Add `nightly-cuda-core`: released cuda-core from PyPI against main-built pathfinder + bindings. Fills the "core released x bindings main" quadrant discussed here (https://github.com/NVIDIA/cuda-python/issues/1955#issuecomment-4296771068).
- Add `nightly-numba-cuda-mlir`: MLIR-backend companion to `nightly-numba-cuda`.
- Both fetch the release tag's test suite from git — neither wheel ships `test_*.py`.
- Fix `CUDA_PYTHON_PER_THREAD_DEFAULT_STREAM` → `CUDA_PYTHON_CUDA_PER_THREAD_DEFAULT_STREAM` typo in `ci/test-matrix.yml` (https://github.com/NVIDIA/cuda-python/issues/971#issuecomment-4849789246).

Refs #971, #1955.

PR #2286, opened 2026-07-01
cuda_pathfinder: pin nvshmem to <3.7 (was previously excluding only 3.7.0)

leofang:
## Summary

`nvidia-nvshmem-cu{12,13}` 3.7.x is breaking main, not just 3.7.0. Widen the exclusion from `\!=3.7.0` to `<3.7` so all 3.7.x versions are avoided until we can move forward.

Extracted from #2283 to unblock CI ASAP.

See issue #2200 (and PR #2201) for related discussions.

### DIFF
--- a/cuda_pathfinder/pyproject.toml
+++ b/cuda_pathfinder/pyproject.toml
@@ -29,7 +29,7 @@ cu12 = [
     "nvidia-cusparselt-cu12",
     "nvidia-libmathdx-cu12",
     "nvidia-nccl-cu12; sys_platform != 'win32'",
-    "nvidia-nvshmem-cu12!=3.7.0; sys_platform != 'win32'",
+    "nvidia-nvshmem-cu12<3.7; sys_platform != 'win32'",
 ]
 cu13 = [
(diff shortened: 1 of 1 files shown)


mdboom (approved):
👍

cuda_pathfinder/pyproject.toml
```diff
@@ -29,7 +29,7 @@ cu12 = [
     "nvidia-cusparselt-cu12",
     "nvidia-libmathdx-cu12",
     "nvidia-nccl-cu12; sys_platform != 'win32'",
-    "nvidia-nvshmem-cu12!=3.7.0; sys_platform != 'win32'",
+    "nvidia-nvshmem-cu12<3.7; sys_platform != 'win32'",
```
rwgk, line 32: For the future: best to create an issue to track such pinning, and add a `# issue #1234` comment here. 

Otherwise this might get buried under the pile of other things we're dealing with.

leofang: It was already done: https://github.com/NVIDIA/cuda-python/issues/2200

leofang:
Since most linux-64 CIs are green, let me admin-merge to unblock everyone. The CI queue seems long today.

PR #2289, opened 2026-07-01
[doc-only] cuda_core: add "10 minutes to cuda.core" tutorial to docs

sri-koundinyan:
This adds a beginner-friendly "10 minutes to cuda.core" guide to the `cuda.core` docs. It walks through the core workflow with small, runnable snippets: selecting a device, compiling a CUDA C++ kernel, allocating memory, copying, launching, timing with events, using multiple streams, capturing a CUDA graph, and interoperating with CuPy and PyTorch. The new page lives at `cuda_core/docs/source/10_minutes_to_cuda_core.rst` and is added to the docs table of contents between the Installation and Examples sections. All code snippets were verified to run end-to-end on CUDA 13.

### DIFF
--- a/cuda_core/docs/source/10_minutes_to_cuda_core.rst
+++ b/cuda_core/docs/source/10_minutes_to_cuda_core.rst
@@ -47,7 +47,7 @@ a kernel" using small, runnable snippets. For the full reference, see the
 
    .. code-block:: console
 
-      $ pip install cuda-core[cu13]
+      $ pip install "cuda-core[cu13]"
 
    The :doc:`install page <install>` covers driver requirements, conda, CUDA 12,
    and installing from source.
@@ -69,17 +69,16 @@ on the current host thread. Always call it before doing GPU work.
 
 .. code-block:: python
 
-   dev = Device()        # device 0 by default; Device(1) selects another GPU
+   dev = Device()        # current device (device 0 if none is current); Device(1) picks a specific GPU
    dev.set_current()     # initialize CUDA and make this device current
 
    print(dev.name)                 # e.g. 'NVIDIA GB10'
    print(dev.compute_capability)   # ComputeCapability(major=12, minor=1)
    print(dev.arch)                 # '121'  (handy for building kernels below)
 
-:class:`Device` objects are thread-local singletons: ``Device(0)`` always hands
-you back the same object for device 0 on that thread, so libraries sharing your
-process see and use the same GPU. Rich device attributes live under
-:attr:`Device.properties` (for example
+:class:`Device` is a thread-local singleton: on the same thread, ``Device(0)``
+always returns the same object, so repeated calls are cheap. Rich device
+attributes live under :attr:`Device.properties` (for example
 ``dev.properties.multiprocessor_count``).
 
 
@@ -121,16 +120,17 @@ exception carries the compiler's log, so you can see exactly what went wrong
    try:
        Program(bad, code_type="c++", options=opts).compile("cubin")
    except Exception as e:
-       print(e)        # includes: error: identifier "not_a_real_symbol" is undefined
+       # the exception carries the compiler log
+       assert 'identifier "not_a_real_symbol" is undefined' in str(e)
 
 
 Streams
 -------
 
 A :class:`Stream` is an ordered queue of GPU work. Operations on the same stream
-run in order; separate streams may overlap. Most ``cuda.core`` operations that
-touch the GPU take a stream so you stay in control of ordering. Create one from
-the device:
+run sequentially; operations on separate streams may run concurrently. Most
+``cuda.core`` operations that touch the GPU take a stream so you stay in control
+of ordering. Create one from the device:
 
 .. code-block:: python
 
@@ -193,8 +193,8 @@ gives a CuPy array backed by the device buffer.
 Copying and launching
 ---------------------
 
-Copy host → device with :meth:`Buffer.copy_from`, run the kernel with
-:func:`launch`, then copy device → host with :meth:`Buffer.copy_to`. These are
+Copy host to device with :meth:`Buffer.copy_from`, run the kernel with
+:func:`launch`, then copy device to host with :meth:`Buffer.copy_to`. These are
 all stream-ordered and asynchronous; :meth:`Stream.sync` blocks until the queued
 work finishes.
 
@@ -227,7 +227,7 @@ Timing with events
 
 An :class:`Event` marks a point in a stream. Create timing-enabled events,
 :meth:`record <Stream.record>` them around some work, synchronize, and subtract
-to get elapsed GPU time in milliseconds.
+to get the elapsed time between them in milliseconds.
 
 .. code-block:: python
 
@@ -290,10 +290,11 @@ independent and lets the runtime overlap it when it can. The
 Capturing work in a CUDA graph
 ------------------------------
 
-When you launch the same sequence of operations repeatedly, per-launch CPU
-overhead adds up. A CUDA graph lets you record that sequence once and replay it
-with a single launch. Use the stream's graph builder: begin building, issue the
-launches into the *builder* instead of the stream, then complete the graph.
+Repeating the same GPU work many times means paying the CPU cost of submitting
+each operation on every repeat. A CUDA graph lets you record that work once and
+then launch the whole graph repeatedly, instead of re-issuing each operation
+every time. Use the stream's graph builder: begin building, issue your
+operations into the *builder* instead of the stream, then complete the graph.
 
 .. code-block:: python
 
@@ -306,11 +307,18 @@ launches into the *builder* instead of the stream, then complete the graph.
    graph = gb.end_building().complete()
 
    graph.upload(stream)
-   graph.launch(stream)     # replay the whole sequence in one shot
+   graph.launch(stream)     # replay the whole graph in one shot
    stream.sync()
 
 See :cuda-core-example:`cuda_graphs.py <cuda_graphs.py>`
 for a complete capture-and-replay example with a measured speedup.
+
+Beyond the stream capture shown here, ``cuda.core`` also provides an explicit
+graph interface (the :mod:`cuda.core.graph` module) for building, inspecting, and
+modifying graphs directly, including graphs first produced by capture. The
+project's `graph tests
+<https://github.com/NVIDIA/cuda-python/tree/main/cuda_core/tests/graph>`_ have
+many usage examples.
 
 
 Working with CuPy and PyTorch
@@ -343,25 +351,20 @@ array of ones:
 
 .. code-block:: python
 
-   # PyTorch: wrap torch's stream via the __cuda_stream__ protocol
+   # PyTorch streams implement the __cuda_stream__ protocol natively
    import torch
 
-   class PyTorchStreamWrapper:
-       def __init__(self, pt_stream):
-           self.pt_stream = pt_stream
-       def __cuda_stream__(self):
-           return (0, self.pt_stream.cuda_stream)
-
    t = torch.ones(n, dtype=torch.float32, device="cuda")
-   ts = dev.create_stream(PyTorchStreamWrapper(torch.cuda.current_stream()))
+   ts = dev.create_stream(torch.cuda.current_stream())
    launch(ts, config, scale, t.data_ptr(), np.float32(2.0), np.uint64(t.numel()))
    ts.sync()
-   assert torch.allclose(t, torch.full_like(t, 2.0))
+   assert bool((t == 2).all())
 
-The ``__cuda_stream__`` protocol is how any object advertises that it represents
-a CUDA stream; :meth:`Device.create_stream` accepts such objects so you can drive
-``cuda.core`` work on another library's stream. See the
-:doc:`interoperability guide <interoperability>` for details.
+The ``__cuda_stream__`` protocol is how an object advertises that it represents a
+CUDA stream, so :meth:`Device.create_stream` can accept it and drive ``cuda.core``
+work on another library's stream. PyTorch implements it natively as of version
+2.10. See the :doc:`interoperability guide <interoperability>` for the protocol
+details and how to support it in your own types.
 
 **Array-library-agnostic views.** To accept *any* CuPy/PyTorch/NumPy-like array
 that supports DLPack or the CUDA Array Interface, decorate a function with
@@ -399,7 +402,7 @@ Cleaning up
 Buffers, streams, events, graphs, and graph builders hold CUDA resources. They
 are released when garbage-collected, but you can release them explicitly with
 ``close()``, which the :cuda-core-examples:`cuda.core examples </>` do in a
-``finally`` block. Buffers are also context managers.
+``finally`` block.
 
 .. code-block:: python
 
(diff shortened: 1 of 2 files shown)


cuda_core/docs/source/10_minutes_to_cuda_core.rst
```diff
@@ -0,0 +1,438 @@
...
+
+Without the :meth:`Stream.wait`, the copy on ``stream_b`` could race ahead of the
+kernel on ``stream_a``. Whether the two independent launches actually overlap is
+up to the GPU scheduler, which can run them together only when each leaves the
+device underutilized; using separate streams expresses that the work is
+independent and lets the runtime overlap it when it can. The
+:doc:`examples <examples>` show this scaled up across multiple GPUs.
+
+
+Capturing work in a CUDA graph
```
Andy-Jost, line 290: It's probably worth adding a sentence to note that cuda.core also provides an explicit graph interface. It can be used to build, inspect, and modify graphs (including graphs built via stream capture). There are lots of examples under `cuda_core/tests/graph`.

cuda_core/docs/source/10_minutes_to_cuda_core.rst
```diff
@@ -0,0 +1,438 @@
...
+
+.. note::
+
+   These snippets target CUDA 13 and need a working CUDA 13 installation plus
+   ``cuda.core`` and a matching ``cuda.bindings`` (13.x). The quickest way to get
+   both from PyPI is:
+
+   .. code-block:: console
+
+      $ pip install cuda-core[cu13]
```
Andy-Jost: Square brackets require being quoted in most shells.

cuda_core/docs/source/10_minutes_to_cuda_core.rst
```diff
@@ -0,0 +1,438 @@
...
+------------------
+
+:class:`Device` is your entry point. Creating one does **not** initialize the
+GPU; calling :meth:`Device.set_current` is what sets up a `CUDA context
+<https://docs.nvidia.com/cuda/cuda-programming-guide/03-advanced/driver-api.html#context>`_
+on the current host thread. Always call it before doing GPU work.
+
+.. code-block:: python
+
+   dev = Device()        # device 0 by default; Device(1) selects another GPU
```
Andy-Jost: The comment is misleading. `Device()` returns the current device, while falling back to device 0 if none is current.

cuda_core/docs/source/10_minutes_to_cuda_core.rst
```diff
@@ -0,0 +1,438 @@
...
+   dev.set_current()     # initialize CUDA and make this device current
+
+   print(dev.name)                 # e.g. 'NVIDIA GB10'
+   print(dev.compute_capability)   # ComputeCapability(major=12, minor=1)
+   print(dev.arch)                 # '121'  (handy for building kernels below)
+
+:class:`Device` objects are thread-local singletons: ``Device(0)`` always hands
+you back the same object for device 0 on that thread, so libraries sharing your
+process see and use the same GPU. Rich device attributes live under
```
Andy-Jost: This is a false cause-and-effect. Thread-locality is not the reason why libraries can share the same device.

What users may observe is that on a given thread `Device(0)` always hands back the same object for device 0, so its context, default memory resource, and cached attributes are reused rather than rebuilt on each call.

On different threads, `Device(0)` returns different objects bound to distinct contexts. This makes it possible to interact with CUDA from multiple threads in cuda.core.

cuda_core/docs/source/10_minutes_to_cuda_core.rst
```diff
@@ -0,0 +1,438 @@
...
+exception carries the compiler's log, so you can see exactly what went wrong
+(other errors raise an exception with the CUDA error name and description).
+
+.. code-block:: python
+
+   bad = r'extern "C" __global__ void k() { not_a_real_symbol; }'
+   try:
+       Program(bad, code_type="c++", options=opts).compile("cubin")
+   except Exception as e:
+       print(e)        # includes: error: identifier "not_a_real_symbol" is undefined
```
Andy-Jost: Could say `assert 'identifier "not_a_real_symbol" is undefined' in str(e)`

cuda_core/docs/source/10_minutes_to_cuda_core.rst
```diff
@@ -0,0 +1,438 @@
...
+
+
+Streams
+-------
+
+A :class:`Stream` is an ordered queue of GPU work. Operations on the same stream
+run in order; separate streams may overlap. Most ``cuda.core`` operations that
```
Andy-Jost: More precisely: operations on the same stream run sequentially; operations on separate streams may run concurrently.

cuda_core/docs/source/10_minutes_to_cuda_core.rst
```diff
@@ -0,0 +1,438 @@
...
+   host_np[:] = np.arange(n, dtype=np.float32)
+
+The same trick works on the *device* side with CuPy: ``cp.from_dlpack(dbuf)``
+gives a CuPy array backed by the device buffer.
+
+
+Copying and launching
+---------------------
+
+Copy host → device with :meth:`Buffer.copy_from`, run the kernel with
```
Andy-Jost: UNICODE symbol here. Not sure that's intended.

cuda_core/docs/source/10_minutes_to_cuda_core.rst
```diff
@@ -0,0 +1,438 @@
...
+That is the core ``cuda.core`` workflow: **select a device, compile, allocate,
+copy, launch, sync.** Everything below builds on it.
+
+
+Timing with events
+------------------
+
+An :class:`Event` marks a point in a stream. Create timing-enabled events,
+:meth:`record <Stream.record>` them around some work, synchronize, and subtract
+to get elapsed GPU time in milliseconds.
```
Andy-Jost: This is confusing because "elapsed" and "CPU/GPU" are different timing categories. The events measure "elapsed" or "wallclock" time.

cuda_core/docs/source/10_minutes_to_cuda_core.rst
```diff
@@ -0,0 +1,438 @@
...
+up to the GPU scheduler, which can run them together only when each leaves the
+device underutilized; using separate streams expresses that the work is
+independent and lets the runtime overlap it when it can. The
+:doc:`examples <examples>` show this scaled up across multiple GPUs.
+
+
+Capturing work in a CUDA graph
+------------------------------
+
+When you launch the same sequence of operations repeatedly, per-launch CPU
```
Andy-Jost: We can't use "launch" to refer to an arbitrary CUDA operation. CUDA graphs capture many types of operations, and kernel launches are only one type.

cuda_core/docs/source/10_minutes_to_cuda_core.rst
```diff
@@ -0,0 +1,438 @@
...
+overhead adds up. A CUDA graph lets you record that sequence once and replay it
```
Andy-Jost: "Sequence" here is potentially confusing. In general, a graph is not a sequence.

cuda_core/docs/source/10_minutes_to_cuda_core.rst
```diff
@@ -0,0 +1,438 @@
...
+with a single launch. Use the stream's graph builder: begin building, issue the
```
Andy-Jost: The graph lets you replay a computation many times without incurring the cost of CUDA API calls each time.

cuda_core/docs/source/10_minutes_to_cuda_core.rst
```diff
@@ -0,0 +1,438 @@
...
+       def __init__(self, pt_stream):
+           self.pt_stream = pt_stream
+       def __cuda_stream__(self):
+           return (0, self.pt_stream.cuda_stream)
+
+   t = torch.ones(n, dtype=torch.float32, device="cuda")
+   ts = dev.create_stream(PyTorchStreamWrapper(torch.cuda.current_stream()))
+   launch(ts, config, scale, t.data_ptr(), np.float32(2.0), np.uint64(t.numel()))
+   ts.sync()
+   assert torch.allclose(t, torch.full_like(t, 2.0))
```
Andy-Jost: Exact comparison is appropriate here. It could match line 342 exactly or at least use `torch.equal` instead of `allclose`.

cuda_core/docs/source/10_minutes_to_cuda_core.rst
```diff
@@ -0,0 +1,438 @@
...
+
+   graph.close()
+   gb.close()
+   dbuf.close(stream=stream)
+   host.close(stream=stream)
+   stream.close()
+
+   # or scope a buffer with a with-statement:
+   with dev.allocate(nbytes, stream=stream) as tmp:
+       ...  # tmp is freed at the end of the block
```
Andy-Jost: @leofang I'm surprised to see context manager support in `Buffer`. I'm for it, but I thought we were avoiding context managers in cuda.core. FYI, it was added in PR #1701 ("Make GraphicsResource inherit from Buffer").

Users reading this might expect other cuda.core objects with `close` members to support this as well. Is that something we can add?

Andy-Jost (changes requested):
This looks like a great addition to the docs. Blocking merge for a few minor correctness issues but mostly it looks fine to me.

cuda_core/docs/source/10_minutes_to_cuda_core.rst
```diff
@@ -0,0 +1,438 @@
...
+.. code-block:: python
+
+   start = dev.create_event({"timing_enabled": True})
+   end = dev.create_event({"timing_enabled": True})
+
+   stream.record(start)
+   for _ in range(100):
+       launch(stream, config, scale, dbuf, np.float32(1.0), np.uint64(n))
+   stream.record(end)
+   end.sync()
```
juenglin, line 241: Should this be `stream.sync()`?

cuda_core/docs/source/10_minutes_to_cuda_core.rst
```diff
@@ -0,0 +1,438 @@
...
+   launch(stream, config, scale, a.data.ptr, np.float32(2.0), np.uint64(a.size))
+   stream.sync()
+   assert bool((a == 2).all())
+
+.. code-block:: python
+
+   # PyTorch: wrap torch's stream via the __cuda_stream__ protocol
+   import torch
+
+   class PyTorchStreamWrapper:
```
juenglin: This adapter is no longer needed. You can use
```
ts = dev.create_stream(torch.cuda.current_stream())
```
below.

juenglin (approved):

cuda_core/docs/source/10_minutes_to_cuda_core.rst
leofang: I thought there was a special reason why graphic resource works differently. Are we teaching it here? If not, we should not teach about context managers.

cuda_core/docs/source/10_minutes_to_cuda_core.rst
leofang: But note my comment here: https://github.com/NVIDIA/cuda-python/pull/2305#discussion_r3538554463

cuda_core/docs/source/10_minutes_to_cuda_core.rst
sri-koundinyan: Good call, added a short paragraph pointing to the explicit `cuda.core.graph` interface with a link to the graph tests.

cuda_core/docs/source/10_minutes_to_cuda_core.rst
sri-koundinyan: Done

cuda_core/docs/source/10_minutes_to_cuda_core.rst
sri-koundinyan: Fixed: now "current device (device 0 if none is current)".

cuda_core/docs/source/10_minutes_to_cuda_core.rst
sri-koundinyan: Got it. Updated to include a short note (same-thread Device(0) returns the same object, so repeat calls are cheap) and left the primary-context sharing content to the interop section.

cuda_core/docs/source/10_minutes_to_cuda_core.rst
sri-koundinyan: Done

cuda_core/docs/source/10_minutes_to_cuda_core.rst
sri-koundinyan: Reworded to "operations" rather than "launches". "launch" now only refers to launching the graph itself.

cuda_core/docs/source/10_minutes_to_cuda_core.rst
sri-koundinyan: Done. Added the same sentence: "operations on the same stream run sequentially; operations on separate streams may run concurrently."

cuda_core/docs/source/10_minutes_to_cuda_core.rst
sri-koundinyan: Thanks for catching this. Not intended.

cuda_core/docs/source/10_minutes_to_cuda_core.rst
sri-koundinyan: Fixed to "the elapsed time between them in milliseconds."

cuda_core/docs/source/10_minutes_to_cuda_core.rst
sri-koundinyan: Got it. Removed "sequence" from the prose and the code comment.

cuda_core/docs/source/10_minutes_to_cuda_core.rst
sri-koundinyan: A slight reframing: "record that work once and then launch the whole graph repeatedly, instead of re-issuing each operation every time."

cuda_core/docs/source/10_minutes_to_cuda_core.rst
sri-koundinyan: Yes, done

cuda_core/docs/source/10_minutes_to_cuda_core.rst
sri-koundinyan: Removed the context-manager example

cuda_core/docs/source/10_minutes_to_cuda_core.rst
sri-koundinyan: Thanks. I kept `end.sync()`. It matches the pattern in the `Event` docstring. But `stream.sync()` would also work since end is the last op.

cuda_core/docs/source/10_minutes_to_cuda_core.rst
sri-koundinyan: Dropped the wrapper. Now `dev.create_stream(torch.cuda.current_stream())` directly.

Andy-Jost (approved):
Looks great, thanks for the updates.

PR #2295, opened 2026-07-02
Chore: Make installing the pre-commit hook "required"

mdboom:
I think we should start encouraging /all/ developers to install the pre-commit hook, not just running it manually.  I expect some controversy on this one.

Even though we do "squash on merge" and having the last commit in a PR is usually good enough, the copyright year updater can not handle this correctly (since it has no way of otherwise knowing when a file was updated). The `.pyi` generator can create merge conflicts (in the ctk-next workflow) if it is not always self-consistent with every commit.

I think we should stay true to the phrase "pre-commit" and ensure this is run "pre- every single commit".


rwgk (approved):
Nice!

Did you already look into running a one-time backfill, to make all copyright years correct based on git history? I think otherwise merge commits might have spurious year updates for a while, until we organically got everything.

mdboom:
> Did you already look into running a one-time backfill, to make all copyright years correct based on git history? I think otherwise merge commits might have spurious year updates for a while, until we organically got everything.

I did this when I first introduced the copyright year pre-commit check.  But it's possible some errors have been introduced since.

CONTRIBUTING.md
```diff
@@ -61,21 +61,18 @@ This project uses [pre-commit.ci](https://pre-commit.ci/) with GitHub Actions. A
 To set yourself up for running pre-commit checks locally and to catch issues before pushing your changes, follow these steps:
 
 * Install pre-commit with: `pip install pre-commit`
+* Run this once per checkout: `pre-commit install`
```
leofang, line 64: Non-blocking: I thought `pre-commit -a` below would take care of this step?

leofang:
I think requiring pre-commit to be installed and run is a no-brainer. But isn't it a tautology that we add a pre-commit check to see if pre-commit is installed? If pre-commit is not run and files drift, wouldn't our pre-commit.ci catch it and fail? Not sure how this PR would help (apart from the doc change)...

CONTRIBUTING.md
mdboom: It only takes care of it if you remember to run it before every single commit.

The driver for this change is that some people forget to run it with every commit (easy to do), and CI only checks the last commit.  Most of the time, that's equivalent, but with copyright year handling (which needs to know what changed) and `.pyi` handling (which can introduce conflicts) it isn't equivalent.

mdboom:
> I think requiring pre-commit to be installed and run is a no-brainer. But isn't it a tautology that we add a pre-commit check to see if pre-commit is installed? If pre-commit is not run and files drift, wouldn't our pre-commit.ci catch it and fail? Not sure how this PR would help (apart from the doc change)...

There is, unfortunately, no way to insist that everyone has it installed before making a commit.

This PR helps for the case of people who have been running `pre-commit` manually -- they will now get a nudge to `pre-commit install`.  Once they do that, the nudge will go away.  (It was @Andy-Jost's suggestion -- It took me a while to understand why it would be useful, too).

If the config change seems too niche, we could just merge the documentation (policy) change here instead.

leofang:
> If the config change seems too niche, we could just merge the documentation (policy) change here instead.

It'd be my preference. Sometimes I do web-based development through GitHub UI, and there is no way for me to run pre-commit locally. As long as pre-commit.ci still runs and catches issues, I can just leave a comment and let it fix for me.

mdboom:
>> If the config change seems too niche, we could just merge the documentation (policy) change here instead.

> It'd be my preference. Sometimes I do web-based development through GitHub UI, and there is no way for me to run pre-commit locally. As long as pre-commit.ci still runs and catches issues, I can just leave a comment and let it fix for me.

To be clear, this doesn't affect that workflow at all.  There's no way to truly enforce -- this is just a nudge.  The pre-commit CI won't ever give the warning.

toolshed/check_precommit_installed.py
juenglin, line 1: Nice, thanks.

juenglin (approved):

rwgk:
> As long as [pre-commit.ci](http://pre-commit.ci/) still runs and catches issues

That can be "too late", unless we can teach pre-commit.ci to run individually for each commit that's on top of `main`.

I think this is a useful PR, at least for our team. I'd expect that everybody runs pre-commit at least sometimes, in most environments. That's when the nudge will fire.

If pre-commit isn't available, there will be no nudge.

> Sometimes I do web-based development through GitHub UI

Not in this PR for sure, but a belts-and-suspenders workflow that ensures the copyright years are accurate, and ideally also checks the generated .pyi stubs, would be great.

mdboom:
> > As long as [pre-commit.ci](http://pre-commit.ci/) still runs and catches issues

> That can be "too late", unless we can teach pre-commit.ci to run individually for each commit that's on top of main.

For @leofang's use case of quick edits in the web API, this will mean copyright years will not update (and the CI can't catch that).  Any necessary changes to `.pyi` files that weren't made will be errors in the CI (similar to any ruff formatting violations).  Aside from the copyright year issue, quick edits are less problematic because they are single commits, not a PR of potentially multiple commits.

> a belts-and-suspenders workflow that ensures the copyright years are accurate

We might be able to do that as a nightly check (which could flag the errors, but not edit the files).  I don't think it can be done from pre-commit itself, since it can't reliably determine the merge base (in order to see what has changed) prior to creating a PR.

PR #2293, opened 2026-07-02
Relicense CUDA Bindings and CUDA Python under Apache 2.0

kkraus14:
## What

- relicense cuda-bindings and the cuda-python metapackage under Apache 2.0
- make the root LICENSE canonical and package Apache 2.0 license copies in all independently built distributions while preserving package-specific copyright spans
- set License-Expression metadata and package the canonical license in wheels and sdists
- convert generated and maintained source headers without functional generated-code changes
- remove the remaining proprietary EULA text and enable cuda-bindings contributions through the DCO workflow

## Dependency

This draft is based on the post-PR-2285 main branch and is blocked on the internal cybind MR.

## Validation

- clean modern and legacy generation for CTK 13.3 and 12.9 emitted Apache-2.0 for all 68 CUDA outputs while preserving the proprietary default for unrelated outputs
- cuda-python and cuda-bindings wheels and sdists report License-Expression: Apache-2.0 and License-File: LICENSE with exact canonical license bytes
- local/path-based cuda-bindings lockfile entries report Apache-2.0 while historical released-package entries retain their original metadata
- CUDA 13.3 bindings wheel built successfully
- 378 bindings tests passed with 64 expected skips
- 9 Cython interoperability tests passed
- bindings documentation built successfully
- all repository pre-commit hooks passed

PR #2297, opened 2026-07-02
[no-ci] CI: remove restricted paths guard

rwgk:
## Description

Remove the obsolete `CI: Restricted Paths Guard` workflow now that CUDA Python is being relicensed under Apache-2.0.

The guard existed to label external PRs that touched `cuda_bindings/` or `cuda_python/` for special review while those paths had license-specific contribution restrictions. With the relicensing work in PR #2285 and PR #2293, that path-specific automation is no longer needed.

(I also checked for non-obvious references before removing the workflow. Tracked repository files no longer reference `CI: Restricted Paths Guard`, `restricted-paths-guard`, or `Needs-Restricted-Paths-Review`, and the GitHub ruleset/branch-rule metadata I inspected did not show any required-check reference to the guard.)

## Follow-up (done)

- Remove the `Needs-Restricted-Paths-Review` label from the repository after this workflow removal lands.


leofang:
@kkraus14 raised the concern that even with re-licensing we still don't want users to touch the generated code (because we need to update the codegen first). @kkraus14 @mdboom, we need to a new practice that can still be automated. Code review can be conducted by anyone on the team, and we've seen human errors before (that led to this path guard).

leofang (changes requested):
(see above, blocking merge until we have a conclusion.)

rwgk:
> @kkraus14 raised the concern that even with re-licensing we still don't want users to touch the generated code (because we need to update the codegen first). @kkraus14 @mdboom, we need to a new practice that can still be automated. Code review can be conducted by anyone on the team, and we've seen human errors before (that led to this path guard).

I think it's a great idea to have a "generated files guard", but I think we should start such a workflow from scratch. E.g. it doesn't need any of the "is NVIDIA employee" logic, which is what made the restricted paths guard workflow so complex. The "generated files guard" workflow should be developed in tandem with cybind tweaks: small changes in cybind that make the guard workflow easier to implement (@mdboom for visibility). — I'm interested in working on that: if done right, it'll make the cuda-python-private workflows easier, too, potentially much easier.

leofang:
> but I think we should start such a workflow from scratch. E.g. it doesn't need any of the "is NVIDIA employee" logic

I think we do mean that we don't want non-NVIDIANs to touch these files.

rwgk:
> > but I think we should start such a workflow from scratch. E.g. it doesn't need any of the "is NVIDIA employee" logic
> 
> I think we do mean that we don't want non-NVIDIANs to touch these files.

Oh, I didn't explain:

Because determining "is NVIDIA employee" is troublesome/unreliable, I was thinking of adding a (simple) heuristic: does it look like cybind was run?

So if someone only changes one, or a small subset, of the generated files, the alarm would trigger.

It seems very unlikely to me that anyone would change all generated files not through cybind.

rwgk:
> Because determining "is NVIDIA employee" is troublesome/unreliable, I was thinking of adding a (simple) heuristic: does it look like cybind was run?
> 
> So if someone only changes one, or a small subset, of the generated files, the alarm would trigger.

@leofang see cybind issue 192

rwgk:
> > Because determining "is NVIDIA employee" is troublesome/unreliable, I was thinking of adding a (simple) heuristic: does it look like cybind was run?
> > So if someone only changes one, or a small subset, of the generated files, the alarm would trigger.

With PR #2293 merged, the existing `restricted-paths-guard.yml` no longer matches the policy or mechanism we need. Its detection logic, generated PR comment, workflow name, and label all encode the former license-based restrictions. For generated-file integrity, PR #2310 instead uses the existing merge-blocking pre-commit CI, so no separate GitHub Actions workflow is needed.

If we separately decide to enforce a policy based on contributor affiliation, that should be treated as a distinct requirement and implemented and named accordingly, rather than carried forward through the old restricted-paths guard.

leofang:
Yes, my understanding from the meetings yesterday/today is we'll merge this and then #2310 to replace the guarding mechanism.

leofang (approved):

rwgk:
I removed the Needs-Restricted-Paths-Review label.

PR #2300, opened 2026-07-02
Add griffe API checking to cuda_core

mdboom:
This adds a CI job to run `griffe` over `cuda_core` to detect API breakage.

It checks both against the merge base of the PR and against the latest `cuda-core` tag.

~This PR also deliberately makes an API breakage (the same one found in #2280) so we can confirm it's working.~

One side effect of this is that griffe /requires/ the top-level `__init__.py` to have an `__all__` so it's explicit about what the public API is.  This is a minor ongoing maintenance burden, but probably worth the effort.

### DIFF
--- a/.github/actions/griffe-api-check/action.yml
+++ b/.github/actions/griffe-api-check/action.yml
@@ -1,0 +1,44 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+name: griffe API check
+
+description: >-
+  Check a package's public API (as defined by `__all__`) for changes using
+  griffe.
+
+inputs:
+  package-name:
+    description: "Importable package name to check, e.g. cuda.core"
+    required: true
+  package-dir:
+    description: "Directory to search for the package sources, e.g. cuda_core"
+    required: true
+  merge-base:
+    description: >-
+      Git ref/sha to compare the current code against, typically the PR's
+      merge-base with its target branch.
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Install uv
+      uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39  # v8.2.0
+      with:
+        enable-cache: false
+
+    - name: Check API
+      shell: bash --noprofile --norc -euo pipefail {0}
+      env:
+        PACKAGE_NAME: ${{ inputs.package-name }}
+        PACKAGE_DIR: ${{ inputs.package-dir }}
+        MERGE_BASE: ${{ inputs.merge-base }}
+      run: |
+        uvx griffe check "$PACKAGE_NAME" \
+          --search "$PACKAGE_DIR" \
+          --find-stubs-packages \
+          --against "$MERGE_BASE" \
+          --format github \
+          2>&1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,6 +103,7 @@ jobs:
       test_bindings: ${{ steps.compose.outputs.test_bindings }}
       test_core: ${{ steps.compose.outputs.test_core }}
       test_pathfinder: ${{ steps.compose.outputs.test_pathfinder }}
+      pr_merge_base: ${{ steps.filter.outputs.merge_base }}
     steps:
       - name: Checkout repository
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
@@ -156,6 +157,7 @@ jobs:
             echo "python_meta=$(has_match '^cuda_python/')"
             echo "test_helpers=$(has_match '^cuda_python_test_helpers/')"
             echo "shared=$(has_match '^(\.github/|ci/|scripts/|toolshed/|conftest\.py$|pyproject\.toml$|pixi\.(toml|lock)$|pytest\.ini$|ruff\.toml$)')"
+            echo "merge_base=${base}"
           } >> "$GITHUB_OUTPUT"
 
       - name: Compose gating outputs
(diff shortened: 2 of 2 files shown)


mdboom:
image

.github/workflows/ci.yml
```diff
@@ -228,6 +230,50 @@ jobs:
             echo "test_pathfinder=${test_pathfinder}"
           } >> "$GITHUB_OUTPUT"
 
+  # Compares the public API of cuda_core (as defined by `__all__` in its
+  # modules/subpackages) against the PR's base branch and the latest
+  # cuda-core release tag, using griffe (see .github/actions/griffe-api-check).
+  # Findings fail this job (shows red in the UI), but it never blocks merging
+  # the PR: it's intentionally excluded from `checks:`'s `needs`, so it isn't
+  # part of the required aggregator status.
+  api-check-core:
```
leofang: Need to wire this job up in the check status job.

cuda_core/cuda/core/__init__.py
```diff
@@ -68,6 +68,48 @@ class _PatchedProperty(metaclass=_PatchedPropMeta):
 del _patch_rlcompleter_for_cython_properties
 
 
+__all__ = [
```
leofang, line 101: Q: If this is a must to make griffe happy, should we just unify the style and always use `__all__` in every submodule, private or public?

leofang:
Very nice! It shows warnings right at the offending diff! 

截圖 2026-07-02 下午5 57 03

cuda_core/cuda/core/__init__.py
mdboom: It's just needed in the public ones.  I think that's a reasonable convention.

mdboom:
> Very nice! It shows warnings right at the offending diff!

Yep.  Someday maybe we can get it to put it in the `.pyx` file, not the generated `.pyi` file, but this is more than good enough to catch breakages.

.github/workflows/ci.yml
mdboom: Yes, I think we should eventually.  I was thinking we could make it a "soft fail" to start with while we learn about whether it produces any false positives.

mdboom:
Note: The changes that this is highlighting against the `cuda-core-v1.0.1` tag are coming from the fact that `cuda-core-v1.0.1` predates having `.pyi` files.  So that's not anything to be concerned about.

This PR should work as-is for checking for changes in a specific PR.  For changes against the latest release, we will need to wait until the next release of cuda-core (which will have `.pyi` files) and then we can do that as well.  (That's not really useful for review, but it's useful to have all the changes in one place so that /when/ we make a breaking release we know exactly what all of them are for release notes purposes).

.github/workflows/ci.yml
mdboom: Actually, I'm not sure we ever want this to fail.  There will be times in our release cycle where we are ok with breaking the API (once something has been deprecated for some time).  We could maybe convince griffe to understand that someday, but in the meantime, having human-override judgement in the loop is probably good enough.

leofang: In that case, we could admin-merge maybe? Or is there a way to skip griffe on a per-line basis (like `# noqa`)?

rwgk: We could introduce a `griffe-override` label or similar to enable merging. Or alternatively `[API-BREAK]` in the PR title, to make it totally obvious. Or both, so we can easily deal with griffe false positives.

rwgk (commented):
gpt-5.5 (with manual edits/additions)

## Findings

- Low: `cuda_core/cuda/core/__init__.py` adds `DeviceResources` to top-level `__all__`, but it is documented in `cuda_core/docs/source/api_private.rst`, not `cuda_core/docs/source/api.rst`. Since `cuda_core/AGENTS.md` defines public API by `__all__`, this promotes a returned helper into the explicit public surface. I would either remove it from top-level `__all__` or move it to the main API page.
 
- Low: `.github/workflows/ci.yml` runs `api-check-core` only when `cuda_core/` changes. A follow-up that edits only `.github/actions/griffe-api-check/action.yml` will not exercise the action. Consider gating on `core || shared`, or a narrower action-specific change flag.
 
- Low: `.github/actions/griffe-api-check/action.yml` invokes unpinned `uvx griffe` in both comparisons. That is fine for a prototype, but if this becomes a trusted signal, pin griffe or route it through the repo-managed environment.

## Maintenance Risk

A maintenance risk from adding `__all__` is source-of-truth drift. New APIs can be imported and documented but omitted from `__all__`, stale names can remain after removals (although `import *` would catch those), and returned/private helper types can accidentally become SemVer-public. The `DeviceResources` mismatch is a concrete example: it is imported at the top level and now listed in `__all__`, but the docs currently treat it as a private/returned helper.

That said, an explicit `__all__` is probably the right convention if griffe is going to define and compare the supported public API from it. The burden is manageable as long as there is a consistency check that catches drift early.

A simple guard against stale names could be:

```python
def test_cuda_core_all_exports_resolve():
    import cuda.core

    missing = [name for name in cuda.core.__all__ if not hasattr(cuda.core, name)]
    assert missing == []
```

## Suggested Follow-On Idea

Add a follow-on docs/API consistency test that compares the public symbols in `cuda.core.__all__` against the public entries in `cuda_core/docs/source/api.rst`, with deliberate exceptions for submodule namespaces and returned-helper/private API docs.

It should parse the Sphinx `autosummary` and `data` entries, then account for known surfaces such as `cuda.core.graph`, `cuda.core.texture`, `cuda.core.utils`, `cuda.core.checkpoint`, and entries intentionally documented in `api_private.rst`. The useful failure mode is: “this symbol is public by `__all__` but missing from public docs” or “this symbol is documented as public but not exported by `__all__`.”

.github/workflows/ci.yml
mdboom: The label thing might work.  I don't know if we want to rely on admin merge given not many of us can do that.

I don't think per-line would help because we want to say "just this once it's ok to break API, but not after this point".

There are two important scenarios I have in mind where this is necsesary:

1) Right before a major release, we break all the APIs we already told our users (through deprecation) that we were are going to break

2) Between releases it is perfectly ok to break APIs of things that have been added since the last release.  I think it's pretty common that new APIs get introduced, merged and then "revised" before an actual release.  This is why I implemented comparing to the previous released version in this PR -- we could see that "even though this API broke against main, it didn't break against the latest release (because it's on new APIs).  That feature of this PR won't actually work until the next cuda-core release, unfortunately, since 1.0.1 didn't have `.pyi` files (and it has a ton of type annotation bugs that that work revealed).

mdboom:
@rwgk: Do you want to file an issue for keeping `__all__` and the docs in sync?  (That seems totally unrelated to this PR).

mdboom:
Heads up -- I am going to wait for the cuda.core 1.1.0 release to come back and finish this PR.  Things will work much better when the latest tagged release as `.pyi` files.  (1.0.1 predates shipping `.pyi` files).

rwgk:
> @rwgk: Do you want to file an issue for keeping `__all__` and the docs in sync?

Done: #2326

> (That seems totally unrelated to this PR).

Yes, absolutely.

cuda_core/cuda/core/__init__.py
juenglin: Do we need to add something about this to AGENTS.md?

mdboom: #2347 added test infrastructure to make sure `__all__` is kept in sync.  Personally, I always prefer that kind of thing to AGENTS.md when possible.

.github/actions/griffe-api-check/action.yml
```diff
@@ -0,0 +1,44 @@
...
+        enable-cache: false
+
+    - name: Check API
+      shell: bash --noprofile --norc -euo pipefail {0}
+      env:
+        PACKAGE_NAME: ${{ inputs.package-name }}
+        PACKAGE_DIR: ${{ inputs.package-dir }}
+        MERGE_BASE: ${{ inputs.merge-base }}
+      run: |
+        uvx griffe check "$PACKAGE_NAME" \
```
juenglin, line 39: If this is use of a new third-pary tool, do we need to get it approved first?

Secondly, you are not pinning a version but we are pinning versions of other tools (like github actions). What's the thinking or general rule of thumb for when to pin?

juenglin (approved):
Left a question about tool version pinning.

PR #2302, opened 2026-07-06
Add exact-name bitcode library lookup

benhg:
## Description
 
Addresses new design of #2200.

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_static_libs/find_bitcode_lib.py
+++ b/cuda_pathfinder/cuda/pathfinder/_static_libs/find_bitcode_lib.py
@@ -182,23 +182,23 @@ def locate_bitcode_lib(name: str) -> LocatedBitcodeLib:
     return _locate_bitcode_lib(name)
 
 
-def locate_bitcode_lib_by_name(name: str, filename: str) -> LocatedBitcodeLib:
+def locate_bitcode_lib_by_name(libname: str, filename: str) -> LocatedBitcodeLib:
     """Locate an exact bitcode filename using a supported library's search paths.
 
     This lets a library define its own filename convention, including any
     architecture or other attributes encoded in the filename.
 
     Args:
-        name: Supported bitcode library whose configured directories to search.
+        libname: Supported bitcode library whose configured directories to search.
         filename: Exact file name to locate. Directory components are not allowed.
 
     Raises:
         TypeError: If ``filename`` is not a string.
-        ValueError: If ``name`` is unsupported or ``filename`` is not a file name.
+        ValueError: If ``libname`` is unsupported or ``filename`` is not a file name.
         BitcodeLibNotFoundError: If ``filename`` cannot be found.
     """
     _validate_filename(filename)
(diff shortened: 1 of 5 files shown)


rwgk:
@benhg, @leofang: I checked CUDA 12.0–13.3 and NVSHMEM cu12 wheels 3.6.5, 3.7.0, and 3.7.1. All actual bitcode filenames end in `.bc`, but NVSHMEM’s bitcode directory also contains `.a` archives, the host `.so`, and bootstrap/transport plugins. As written, `find/locate_bitcode_lib_by_name()` can return any of those by basename. Should the API require a `.bc` filename, or is generic exact-name lookup within the configured directory intentional? Would suffix validation be the right contract, or do we want content-level bitcode validation?

benhg:
> @benhg, @leofang: I checked CUDA 12.0–13.3 and NVSHMEM cu12 wheels 3.6.5, 3.7.0, and 3.7.1. All actual bitcode filenames end in `.bc`, but NVSHMEM’s bitcode directory also contains `.a` archives, the host `.so`, and bootstrap/transport plugins. As written, `find/locate_bitcode_lib_by_name()` can return any of those by basename. Should the API require a `.bc` filename, or is generic exact-name lookup within the configured directory intentional? Would suffix validation be the right contract, or do we want content-level bitcode validation?

I could go either way. The argument in favor of looser checking is we may have new formats (e.g. .bc.fatbin or .bc.gz or something), and could use the same function to support .ltoir. The argument against is more risk of bugs

cuda_pathfinder/cuda/pathfinder/_static_libs/find_bitcode_lib.py
```diff
@@ -169,6 +172,35 @@ def locate_bitcode_lib(name: str) -> LocatedBitcodeLib:
...
+    """Locate a bitcode library by name.
+
+    Raises:
+        ValueError: If ``name`` is not a supported bitcode library.
+        BitcodeLibNotFoundError: If the bitcode library cannot be found.
+    """
+    return _locate_bitcode_lib(name)
+
+
+def locate_bitcode_lib_by_name(name: str, filename: str) -> LocatedBitcodeLib:
```
leofang: `name` and `filename` are quite confusing IMHO. Is `name` the directory name that contains the bitcode library?

benhg: Name is the library name (e.g. "nvshmem"). Filename is the bitcode filename (e.g. "libnvshmem_device.bc")

benhg: agree it's confusing

leofang: How about replacing `name` by `libname` or `library_name`? I know it's deviating from the existing convention in other APIs... but probably acceptable given that this is our first time adding a by-name API?

benhg: Works for me.

PR #2303, opened 2026-07-06
[no-ci] expand context7.json

uday1o1:
Summary

Adds context7.json so Context7 can use CUDA Python docs from the repository.

This sets up:

- docs folders for indexing
- exclusions for non-user-facing docs content

After merge, we should refresh/check the Context7 page to confirm the expected documentation is available.

### DIFF
--- a/context7.json
+++ b/context7.json
@@ -1,4 +1,32 @@
 {
+  "$schema": "https://context7.com/schema/context7.json",
   "url": "https://context7.com/nvidia/cuda-python",
-  "public_key": "pk_gupaHhsdvsuT1j3BZpb7i"
+  "public_key": "pk_gupaHhsdvsuT1j3BZpb7i",
+  "folders": [
+    "cuda_python/docs/source",
+    "cuda_bindings/docs/source",
+    "cuda_core/docs/source",
+    "cuda_pathfinder/docs/source"
+  ],
+  "excludeFolders": [
+    "**/_templates"
+  ],
+  "excludeFiles": [
+    "AGENTS.md",
+    "CLAUDE.md",
+    "CHANGELOG.md",
+    "CHANGELOG.mdx",
+    "changelog.md",
+    "changelog.mdx",
+    "CONTRIBUTING.md",
+    "LICENSE.md",
+    "license.md",
+    "CODE_OF_CONDUCT.md",
+    "code_of_conduct.md",
+    "SECURITY.md",
+    "api_private.rst",
+    "conduct.rst",
+    "contribute.rst",
+    "license.rst"
+  ]
 }


danielfrg (approved):

context7.json
```diff
@@ -1,4 +1,79 @@
...
+    "license.md",
+    "CODE_OF_CONDUCT.md",
+    "code_of_conduct.md",
+    "SECURITY.md",
+    "api_private.rst",
+    "conduct.rst",
+    "contribute.rst",
+    "license.rst"
+  ],
+  "previousVersions": [
```
leofang: We might need to auto-generate this part at least. Each sub-folder already has a file like this and it's been very painful to keep up, e.g.
https://github.com/NVIDIA/cuda-python/blob/main/cuda_python/docs/nv-versions.json

rwgk (changes requested):
Thank you for putting this together. I support the goal of improving CUDA Python's Context7 entry, including retaining the ownership claim and adding stable indexing controls such as focused documentation folders and targeted exclusions.

[Context7's documentation](https://context7.com/docs/library-owners) describes `context7.json` as optional configuration for parsing and presentation. Claiming a library also provides useful administrative capabilities, including metrics, refresh controls, and version management. These are valuable, bounded reasons to maintain the file. However, I am requesting changes because the proposed hand-maintained `previousVersions` array creates an additional release-maintenance obligation without a clear source of truth or demonstrated need.

CUDA Python already maintains documentation version lists in:

- `cuda_python/docs/nv-versions.json`
- `cuda_bindings/docs/nv-versions.json`
- `cuda_core/docs/nv-versions.json`
- `cuda_pathfinder/docs/nv-versions.json`

Adding a separately curated list to `context7.json` gives every release another place where an update can be missed or become inconsistent. After overlapping CUDA Python and cuda.bindings releases are deduplicated, the existing files represent 58 distinct historical versions, while Context7 [permits at most 20 configured versions](https://context7.com/docs/howto/claiming-libraries#managing-versions). The 14 versions selected here therefore encode a retention policy, but that policy is neither documented nor consistently derivable: the CUDA Python and cuda.pathfinder entries generally retain the latest patch from each minor line, while cuda.core includes every 1.x release.

There is also a representation mismatch. CUDA Python, cuda.bindings, cuda.core, and cuda.pathfinder have independent release streams, whereas each Context7 version maps to a complete repository snapshot. For example, `cuda-core-v1.1.0` also captures whichever cuda.bindings and cuda.pathfinder revisions existed at that commit. In addition, `v12.9.7`, `v12.8.0`, and `v12.6.2` do not contain the configured `cuda_pathfinder/docs/source` folder. A single historical-version list therefore cannot accurately represent the component-version matrix.

The repository's array is not needed to expose the latest documentation. Context7's [public search result](https://context7.com/api/v2/libs/search?libraryName=cuda-python&query=CUDA%20Python) already exposes `v13.3.1`, even though `context7.json` on `main` currently contains only `url` and `public_key`. Whether Context7 detected that version automatically or it was configured through the administration interface, the result shows that `previousVersions` in the repository is not a prerequisite for publishing the current documentation. Its incremental benefit is specifically older-version coverage, and this PR does not establish that demand for such coverage justifies the additional maintenance burden.

### Required change

Please remove `previousVersions` from this PR. The durable ownership, folder, and exclusion configuration can still land in this PR without it. If historical-version indexing is later demonstrated to be necessary, it can be designed separately around an appropriate source of truth and maintenance model.

### Non-blocking suggestion

To keep this optional, Context7-specific configuration minimal, I would also consider omitting `branch: "main"`, which duplicates Context7's default-branch behavior, and `projectTitle` / `description`, which are fallback metadata. The focused `folders` and exclusions are the expanded fields most likely to improve retrieval quality.

uday1o1:
Thanks, removed `previousVersions`.

rwgk (approved):
Thanks for keeping this up to date!

PR #2305, opened 2026-07-06
cuda.core: update PyTorch example to use native __cuda_stream__

### LINKED ISSUE #1083: Update PyTorch example to skip using a custom stream adaptor
The upstream has implemented `__cuda_stream__` support 🎉 
https://github.com/pytorch/pytorch/pull/163614
Once a new release of PyTorch is out, we can update our example and check if the custom adapter is still needed, or if we can use the built-in support:
https://github.com/NVIDIA/cuda-python/blob/279c94351bd81ac807ab1ecd47644dfdce917c1a/cuda_core/examples/pytorch_example.py#L40-L50

juenglin:
Signed-off-by: Ralf Juengling <rjuengling@nvidia.com>

## Description

closes #1083

Drop the `PyTorchStreamWrapper` helper and pass the PyTorch stream
directly to `Device.create_stream()` now that PyTorch implements
`__cuda_stream__` natively.

## Checklist

### DIFF
--- a/cuda_core/examples/pytorch_example.py
+++ b/cuda_core/examples/pytorch_example.py
@@ -32,16 +32,6 @@ __global__ void saxpy_kernel(const T* a, const T* x, const T* y, T* out, size_t 
 """
 
 
-# Create a wrapper class that implements __cuda_stream__
-class PyTorchStreamWrapper:
-    def __init__(self, pt_stream):
-        self.pt_stream = pt_stream
-
-    def __cuda_stream__(self):
-        stream_id = self.pt_stream.cuda_stream
-        return (0, stream_id)  # Return format required by CUDA Python
-
-
 def main():
     dev = Device()
     dev.set_current()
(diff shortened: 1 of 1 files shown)


Andy-Jost (approved):
LGTM!

cuda_core/examples/pytorch_example.py
```diff
@@ -32,24 +32,14 @@
...
-
 def main():
     dev = Device()
     dev.set_current()
 
     pt_stream = torch.cuda.current_stream()
     print(f"PyTorch stream: {pt_stream}", file=sys.stderr)
 
-    stream = dev.create_stream(PyTorchStreamWrapper(pt_stream))
+    stream = dev.create_stream(pt_stream)
```
leofang, line 54: Thanks, @juenglin. Could you check starting with which PyTorch version (I guess 2.9 or 2.10, not very ancient) this was supported? We probably should do a version check in the example and do early return

leofang (changes requested):

cuda_core/examples/pytorch_example.py
juenglin: 2.10 introduced support for this. I will create a commit to check or the torch version.

leofang (approved):

Issue #2248, opened 2026-06-25
[DOC]: cuda.core docs/type hints understate option params since dict is accepted at runtime but not documented


fnrizzi:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Is this for new documentation, or an update to existing docs?

Update

### Describe the incorrect/future/missing documentation

Several `cuda.core` public APIs document their options/config parameter as accepting only a dataclass (or `None`), but at runtime these parameters also accept a plain `dict` since the argument is normalized by `check_or_create_options`, converting a `dict` into the dataclass via `cls(**options)`. 
As a result, the documented signatures and the shipped `.pyi` type stubs are narrower than the actual accepted types.

Affected APIs:
- `Device.create_context` — `ContextOptions | None` (also accepts `dict`)
- `Device.create_event` — `EventOptions | None` (also accepts `dict`)
- `Stream.record` — `EventOptions | None` (also accepts `dict`)
- `launch` — `LaunchConfig` (also accepts `dict`)
- `Program.__init__` — `ProgramOptions | None` (also accepts `dict`)
- `Linker.__init__` — `LinkerOptions | None` (also accepts `dict`)
- `SMResource.split` — `SMResourceOptions` (also accepts `dict`)
- `WorkqueueResource.configure` — `WorkqueueResourceOptions` (also accepts `dict`)
- `VirtualMemoryResource.__init__` — `VirtualMemoryResourceOptions | None` (also accepts `dict`)

This also seems inconsistent with `PinnedMemoryResource.__init__`, which already annotates its parameter as `PinnedMemoryResourceOptions | dict[str, object] | None`, so the convention for accepting a `dict` already seems to exist in the codebase.

### If this is a correction, please provide a link to the incorrect documentation. If this is a new documentation request, please link to where you have looked.

for example, https://nvidia.github.io/cuda-python/cuda-core/latest/generated/cuda.core.Program.html#cuda-core-program. 

i have started a fix for this in my fork: https://github.com/fnrizzi/cuda-python/commit/80dc3ec28e9a817cc1f309109414dfda420a1b1c
but not yet opened MR to give precedence to the issue first as per contributing guidelines.

leofang:
[...]

PR #2307, opened 2026-07-07
cuda.core.texture: API-style follow-up before v1.1.0 (#2292)

### LINKED ISSUE #2292: cuda.core.texture: API-style follow-up before v1.1.0 release
## Background

PR #2095 (design issue #2188) merged the texture/surface stack on 2026-06-27. I was cc'd on #2188 but didn't get a chance to review before it was merged/closed, my apology. The functional design (5 types, factory set, folded `format` + `num_channels`, copy-only interop, graph-capture parity, resource-handle lifetime rewrite) is sound. This issue captures API-style items to address before v1.1.0 ships (milestone due 2026-07-08).

### 1. Allocation shape: `Device.create_*(options=…)` instead of classmethod factories

Every other user-allocatable resource in `cuda.core` is created via `Device.create_*` (`create_stream`, `create_event`, …) or `mr.allocate(size, stream=…)`. `OpaqueArray.from_descriptor(…)` / `MipmappedArray.from_descriptor(…)` introduce a new per-type-classmethod pattern.

Move to:

- `Device.create_opaque_array(options=OpaqueArrayOptions(…)) -> OpaqueArray`
- `Device.create_mipmapped_array(options=MipmappedArrayOptions(…)) -> MipmappedArray`
- `Device.create_texture_object(*, resource, options=TextureObjectOptions(…)) -> TextureObject`
- `Device.create_surface_object(*, resource) -> SurfaceObject`

Ties the resource to whichever device/context is current at construction, mirroring `Stream` / `Event`. Removes the classmethod entry points from the types themselves (the private `_from_handle` graphics-interop escape hatch stays).

Current entry point:
https://github.com/NVIDIA/cuda-python/blob/2780efd4733e43ea5e40287e1a27feb119b3a568/cuda_core/cuda/core/texture/_array.pyi#L58-L79

### 2. Missing `XxxOptions` dataclasses

`cuda.core` convention: multi-field creation goes through an `XxxOptions` dataclass — `StreamOptions`, `EventOptions`, `LaunchConfig`, `DeviceMemoryResourceOptions`, `PinnedMemoryResourceOptions`, `ManagedMemoryResourceOptions`, `VirtualMemoryResourceOptions`, `TensorMapDescriptorOptions`. `OpaqueArray.from_descriptor` currently takes 4 loose kwargs; `MipmappedArray.from_descriptor` takes 5.

```python
@dataclass
class OpaqueArrayOptions:
    shape: tuple[int, ...]
    format: numpy.dtype | ArrayFormatType    # see item 3
    num_channels: int
    is_surface_load_store: bool = False

@dataclass
class MipmappedArrayOptions:
    shape: tuple[int, ...]
    format: numpy.dtype | ArrayFormatType
    num_channels: int
    num_levels: int
    is_surface_load_store: bool = False
```

Also rename `TextureDescriptor` → `TextureObjectOptions`: `CUDA_TEXTURE_DESC` is pure creation input (not queryable back), so "descriptor" was borrowing the driver naming inappropriately. `TextureObjectOptions` matches `TensorMapDescriptorOptions` / `StreamOptions` / etc.

### 3. `format` accepts `numpy.dtype` (matches `TensorMapDescriptor` precedent)

`TensorMapDescriptorOptions.data_type` already accepts NumPy / `ml_dtypes` dtype objects with `TensorMapDataType` retained for compatibility:
https://github.com/NVIDIA/cuda-python/blob/2780efd4733e43ea5e40287e1a27feb119b3a568/cuda_core/cuda/core/_tensor_map.pyi#L88-L102

The 8 current `ArrayFormat` values map 1:1 to numpy dtypes. Apply the same pattern to `format`: `numpy.dtype | ml_dtypes.* | ArrayFormatType`, prefer dtypes, keep the enum as fallback / escape-hatch for driver formats without numpy equivalents (BC1..BC7, `UNORM_INTx` in CTK 13+, `BFLOAT16` via `ml_dtypes`).

### 4. Enum style: `IntEnum` in root namespace → `StrEnum` in `cuda.core.typing`

Every other `cuda.core` enum unified pre-1.0 lives in `cuda.core.typing` as a `StrEnum` with semantic string values and a `Type` suffix (`SourceCodeType`, `GraphMemoryType`, `VirtualMemoryHandleType`, `ObjectCodeFormatType`, `PCHStatusType`, `CompilerBackendType`, `GraphConditionalType`, `ManagedMemoryLocationType`, `VirtualMemoryLocationType`, `VirtualMemoryGranularityType`, `VirtualMemoryAccessType`, `VirtualMemoryAllocationType`).

Existing pattern (`SourceCodeType`):
https://github.com/NVIDIA/cuda-python/blob/2780efd4733e43ea5e40287e1a27feb119b3a568/cuda_core/cuda/core/typing.py#L62-L72

The texture enums are `IntEnum` backed by `cydriver.CU_TR_*`, sit in `cuda.core.texture` root, and drop the `Type` suffix:
https://github.com/NVIDIA/cuda-python/blob/2780efd4733e43ea5e40287e1a27feb119b3a568/cuda_core/cuda/core/texture/_texture.pyi#L11-L33

Bring them in line:

```python
# cuda/core/typing.py
class AddressModeType(StrEnum):
    WRAP = "wrap"
    CLAMP = "clamp"
    MIRROR = "mirror"
    BORDER = "border"

class FilterModeType(StrEnum):
    POINT = "point"
    LINEAR = "linear"

class ReadModeType(StrEnum):
    ELEMENT_TYPE = "element_type"
    NORMALIZED_FLOAT = "normalized_float"

# Optional if item 3's numpy path covers everything:
class ArrayFormatType(StrEnum):
    UINT8 = "uint8"
    ...
```

### 5. `ResourceDescriptor.from_array` → `from_opaque_array`

The type rename `Array` → `OpaqueArray` (item 1 of #2188) was to remove "array" ambiguity; leaving `ResourceDescriptor.from_array(array: OpaqueArray)` puts it right back:
https://github.com/NVIDIA/cuda-python/blob/2780efd4733e43ea5e40287e1a27feb119b3a568/cuda_core/cuda/core/texture/_texture.pyi#L58-L60

- `from_array` → `from_opaque_array`
- `from_mipmapped_array` — unchanged
- `from_linear` — unchanged (matches `CUresourceType.LINEAR`; renaming to `from_buffer` collides with `from_pitch2d`, which also takes a `Buffer`)
- `from_pitch2d` — unchanged

### 6. Drop `SurfaceObject.from_array` sugar

`TextureObject` has only `from_descriptor`. `SurfaceObject` has both `from_array(array)` and `from_descriptor(resource=…)`:
https://github.com/NVIDIA/cuda-python/blob/2780efd4733e43ea5e40287e1a27feb119b3a568/cuda_core/cuda/core/texture/_surface.pyi#L32-L37

`from_array` collapses `from_descriptor(resource=ResourceDescriptor.from_array(a))` into one call — same speculative-sugar rule that killed the `arr.to_buffer(mr, stream)` helper in #2188 decision 2. Remove.

If item 1 lands, both go away (single entry point becomes `Device.create_surface_object(*, resource)`).

### 7. `OpaqueArray.element_size` → `element_bytes`

Byte-count accessors elsewhere on the type spell `_bytes` explicitly (`size_bytes`, `pitch_bytes`). `element_size` returns a byte count and should follow:
https://github.com/NVIDIA/cuda-python/blob/2780efd4733e43ea5e40287e1a27feb119b3a568/cuda_core/cuda/core/texture/_array.pyi#L107-L109

## Not addressed here

- Descriptor round-trip introspection (`cuTexObjectGetResourceDesc` etc.) — deferred in the PR.

-- Leo's bot

Andy-Jost:
## Summary

Addresses all seven API-style items in #2292 to settle the `cuda.core.texture` public surface before it ships in v1.1.0. These are naming/shape refinements to the texture/surface stack merged in #2095 (design #2188); there are no changes to the underlying CUDA behavior. Because the surface first ships in v1.1.0, the renames need no deprecation shims and the new/renamed entry points carry `.. versionadded:: 1.1.0`.

Developed as a sequence of reviewable commits (one per item group); the PR is intended to squash-merge.

## Changes

- **Allocation shape → `Device.create_*` + `XxxOptions` (items 1, 2):** replace the per-type classmethod factories with `Device.create_opaque_array`, `Device.create_mipmapped_array`, `Device.create_texture_object`, and `Device.create_surface_object`, each taking an options dataclass (`OpaqueArrayOptions`, `MipmappedArrayOptions`, `TextureObjectOptions`). This matches the `create_stream`/`create_event` + `StreamOptions`/`EventOptions` convention. The private `_from_handle` graphics-interop escape hatch is retained.
- **Rename `TextureDescriptor` → `TextureObjectOptions` (item 2):** `CUDA_TEXTURE_DESC` is pure creation input, so the type follows the `XxxOptions` convention rather than the driver's "descriptor" naming.
- **`format` accepts NumPy dtypes (item 3):** array/mipmap creation and `ResourceDescriptor.from_linear`/`from_pitch2d` now accept an `ArrayFormatType`, a plain `str`, or a NumPy dtype object, mirroring `TensorMapDescriptorOptions.data_type`.
- **Enums → `cuda.core.typing` as `StrEnum` (item 4):** `ArrayFormat`/`AddressMode`/`FilterMode`/`ReadMode` become `ArrayFormatType`/`AddressModeType`/`FilterModeType`/`ReadModeType` in `cuda.core.typing`, with semantic string values and the `Type` suffix, matching every other `cuda.core` enum. Plain strings are accepted anywhere an enum is expected.
- **`ResourceDescriptor.from_array` → `from_opaque_array` (item 5):** removes the "array" ambiguity that the `Array` → `OpaqueArray` rename was meant to eliminate.
- **Drop `SurfaceObject.from_array` sugar (item 6):** a surface is now built from an explicit `ResourceDescriptor.from_opaque_array(...)`, consistent with the single-entry-point rule.
- **`OpaqueArray.element_size` → `element_bytes` (item 7):** the accessor returns a byte count, matching `size_bytes` / `pitch_bytes`.

Docs (`api.rst`, `api_private.rst`), the texture tests, and the four texture examples are updated throughout.

## Test Coverage

- Updated `tests/test_texture_surface.py` for the new API; negative-path tests reflect that validation now fires at options construction (`ValueError`) rather than at object creation.
- Added tests for the NumPy-dtype and `str` `format` paths (marked `@pytest.mark.agent_authored`).

## Related Work

Closes #2292. Follow-up to #2095 (design #2188).

## Notes

CI will build and run the tests. Local build/run was not possible on the dev machine (no GPU); `pre-commit run --all-files` passes.

### DIFF
--- a/cuda_core/cuda/core/_device.pyx
+++ b/cuda_core/cuda/core/_device.pyx
@@ -1456,6 +1456,62 @@ class Device:
         self._check_context_initialized()
         return GraphBuilder._init(self.create_stream())
 
+    def create_opaque_array(self, options: object = None):
+        """Create an :obj:`~cuda.core.texture.OpaqueArray` on this device.
+
+        Allocates an opaque, hardware-laid-out CUDA array for texture/surface
+        access. The array is bound to the device/context that is current at
+        construction.
+
+        Note
+        ----
+        Device must be initialized.
+
+        Parameters
+        ----------
+        options : :obj:`~cuda.core.texture.OpaqueArrayOptions`
+            Allocation options (shape, format, channels, surface flag).
+
+        Returns
+        -------
+        :obj:`~cuda.core.texture.OpaqueArray`
+            Newly created opaque array.
+
+        .. versionadded:: 1.1.0
+        """
+        from cuda.core.texture._array import _create_opaque_array
+
+        self._check_context_initialized()
+        return _create_opaque_array(options)
+
+    def create_mipmapped_array(self, options: object = None):
+        """Create a :obj:`~cuda.core.texture.MipmappedArray` on this device.
+
+        Allocates a mipmapped CUDA array for texture/surface access across
+        levels. The array is bound to the device/context that is current at
+        construction.
+
+        Note
+        ----
+        Device must be initialized.
+
+        Parameters
+        ----------
+        options : :obj:`~cuda.core.texture.MipmappedArrayOptions`
+            Allocation options (shape, format, channels, levels, surface flag).
+
+        Returns
+        -------
+        :obj:`~cuda.core.texture.MipmappedArray`
+            Newly created mipmapped array.
+
+        .. versionadded:: 1.1.0
+        """
+        from cuda.core.texture._mipmapped_array import _create_mipmapped_array
+
+        self._check_context_initialized()
+        return _create_mipmapped_array(options)
+
 
 cdef inline int Device_ensure_cuda_initialized() except? -1:
     """Initialize CUDA driver and check version compatibility (once per process)."""
--- a/cuda_core/cuda/core/_device.pyx
+++ b/cuda_core/cuda/core/_device.pyx
@@ -44,6 +44,16 @@ from typing import TYPE_CHECKING
 if TYPE_CHECKING:
     import cuda.core.system  # no-cython-lint
     from cuda.core.graph import GraphBuilder
+    from cuda.core.texture import (
+        MipmappedArray,
+        MipmappedArrayOptions,
+        OpaqueArray,
+        OpaqueArrayOptions,
+        ResourceDescriptor,
+        SurfaceObject,
+        TextureObject,
+        TextureObjectOptions,
+    )
 
 # TODO: I prefer to type these as "cdef object" and avoid accessing them from within Python,
 # but it seems it is very convenient to expose them for testing purposes...
@@ -1456,12 +1466,13 @@ class Device:
         self._check_context_initialized()
         return GraphBuilder._init(self.create_stream())
 
-    def create_opaque_array(self, options: object = None):
-        """Create an :obj:`~cuda.core.texture.OpaqueArray` on this device.
+    def create_opaque_array(self, options: OpaqueArrayOptions) -> OpaqueArray:
+        """Create an :obj:`~cuda.core.texture.OpaqueArray` on the current device.
 
         Allocates an opaque, hardware-laid-out CUDA array for texture/surface
-        access. The array is bound to the device/context that is current at
-        construction.
+        access. The array is created in the current CUDA context, so make this
+        device current with :meth:`set_current` before calling (mirroring
+        :meth:`create_stream` / :meth:`create_event`).
 
         Note
         ----
@@ -1484,12 +1495,13 @@ class Device:
         self._check_context_initialized()
         return _create_opaque_array(options)
 
-    def create_mipmapped_array(self, options: object = None):
-        """Create a :obj:`~cuda.core.texture.MipmappedArray` on this device.
+    def create_mipmapped_array(self, options: MipmappedArrayOptions) -> MipmappedArray:
+        """Create a :obj:`~cuda.core.texture.MipmappedArray` on the current device.
 
         Allocates a mipmapped CUDA array for texture/surface access across
-        levels. The array is bound to the device/context that is current at
-        construction.
+        levels. The array is created in the current CUDA context, so make this
+        device current with :meth:`set_current` before calling (mirroring
+        :meth:`create_stream` / :meth:`create_event`).
 
         Note
         ----
@@ -1512,15 +1524,18 @@ class Device:
         self._check_context_initialized()
         return _create_mipmapped_array(options)
 
-    def create_texture_object(self, *, resource, options: object = None):
-        """Create a :obj:`~cuda.core.texture.TextureObject` on this device.
+    def create_texture_object(
+        self, *, resource: ResourceDescriptor, options: TextureObjectOptions | None = None
+    ) -> TextureObject:
+        """Create a :obj:`~cuda.core.texture.TextureObject` on the current device.
 
         Binds a resource (an :obj:`~cuda.core.texture.OpaqueArray` /
         :obj:`~cuda.core.texture.MipmappedArray` / linear or pitch2d
         :obj:`~cuda.core.Buffer`, wrapped in a
         :obj:`~cuda.core.texture.ResourceDescriptor`) as a bindless texture for
-        kernel-side sampled reads. The object is bound to the device/context
-        that is current at construction.
+        kernel-side sampled reads. The object is created in the current CUDA
+        context, so make this device current with :meth:`set_current` before
+        calling (mirroring :meth:`create_stream` / :meth:`create_event`).
 
         Note
         ----
@@ -1545,14 +1560,16 @@ class Device:
         self._check_context_initialized()
         return _create_texture_object(resource, options)
 
-    def create_surface_object(self, *, resource):
-        """Create a :obj:`~cuda.core.texture.SurfaceObject` on this device.
+    def create_surface_object(self, *, resource: ResourceDescriptor) -> SurfaceObject:
+        """Create a :obj:`~cuda.core.texture.SurfaceObject` on the current device.
 
         Binds an :obj:`~cuda.core.texture.OpaqueArray` (via a
         :obj:`~cuda.core.texture.ResourceDescriptor`) as a bindless surface for
         kernel-side typed load/store. The backing array must have been created
-        with ``is_surface_load_store=True``. The object is bound to the
-        device/context that is current at construction.
+        with ``is_surface_load_store=True``. The object is created in the
+        current CUDA context, so make this device current with
+        :meth:`set_current` before calling (mirroring :meth:`create_stream` /
+        :meth:`create_event`).
 
         Note
         ----
--- a/cuda_core/cuda/core/_device.pyx
+++ b/cuda_core/cuda/core/_device.pyx
@@ -1512,6 +1512,70 @@ class Device:
         self._check_context_initialized()
         return _create_mipmapped_array(options)
 
+    def create_texture_object(self, *, resource, options: object = None):
+        """Create a :obj:`~cuda.core.texture.TextureObject` on this device.
+
+        Binds a resource (an :obj:`~cuda.core.texture.OpaqueArray` /
+        :obj:`~cuda.core.texture.MipmappedArray` / linear or pitch2d
+        :obj:`~cuda.core.Buffer`, wrapped in a
+        :obj:`~cuda.core.texture.ResourceDescriptor`) as a bindless texture for
+        kernel-side sampled reads. The object is bound to the device/context
+        that is current at construction.
+
+        Note
+        ----
+        Device must be initialized.
+
+        Parameters
+        ----------
+        resource : :obj:`~cuda.core.texture.ResourceDescriptor`
+            The memory backing the texture.
+        options : :obj:`~cuda.core.texture.TextureObjectOptions`
+            Sampling state (address/filter/read modes, normalization, etc.).
+
+        Returns
+        -------
+        :obj:`~cuda.core.texture.TextureObject`
+            Newly created texture object.
+
+        .. versionadded:: 1.1.0
+        """
+        from cuda.core.texture._texture import _create_texture_object
+
+        self._check_context_initialized()
+        return _create_texture_object(resource, options)
+
+    def create_surface_object(self, *, resource):
+        """Create a :obj:`~cuda.core.texture.SurfaceObject` on this device.
+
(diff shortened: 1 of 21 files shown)


cuda_core/cuda/core/_device.pyx
```diff
@@ -1456,6 +1466,133 @@ class Device:
         self._check_context_initialized()
         return GraphBuilder._init(self.create_stream())
 
+    def create_opaque_array(self, options: OpaqueArrayOptions) -> OpaqueArray:
+        """Create an :obj:`~cuda.core.texture.OpaqueArray` on the current device.
+
+        Allocates an opaque, hardware-laid-out CUDA array for texture/surface
+        access. The array is created in the current CUDA context, so make this
+        device current with :meth:`set_current` before calling (mirroring
+        :meth:`create_stream` / :meth:`create_event`).
```
Andy-Jost, line 1475: See #2311. I think this behavior (using the current context rather than the device specified by `self`) is buggy, but it matches other methods like `Device.create_stream` and `Device.create_event`. Fixing it is out of scope for this PR.

cuda_core/tests/test_texture_surface.py
```diff
@@ -59,36 +63,74 @@ def test_array_2d_create_and_properties(init_cuda):
...
     try:
         assert arr.shape == (8, 8, 4)
         assert arr.is_surface_load_store is True
-        assert arr.element_size == 4
+        assert arr.element_bytes == 4
+    finally:
+        arr.close()
+
+
+@pytest.mark.agent_authored(model="claude-opus-4.8")
```
juenglin, line 82: What is the purpose of these markers (`agent_authored`)?

cuda_core/cuda/core/texture/_texture.pyx
```diff
@@ -547,9 +433,12 @@ cdef class TextureObject:
         return self._source_ref
 
     @property
-    def texture_descriptor(self):
-        """The :class:`TextureDescriptor` this texture was built from."""
-        return self._texture_desc
+    def options(self):
```
rwgk, line 439: Informational only:

Initially codex flagged this as: Exposes the caller’s mutable options. Mutating it later changes reported state without changing the immutable CUDA object. Return an immutable snapshot or remove the property.

When prompted, codex confirmed that there is no code that actually mutates the object. It think it's an acceptable risk.

cuda_core/tests/test_texture_surface.py
Andy-Jost: @rwgk can answer that

cuda_core/cuda/core/texture/_texture.pyi
```diff
@@ -174,38 +156,43 @@ class TextureDescriptor:
...
     border_color : tuple of float or None
         4-tuple used when ``address_mode`` includes ``BORDER``; ``None`` means
         zero.
+
+    .. versionadded:: 1.1.0
     """
-    address_mode: AddressMode | tuple[AddressMode, ...] = AddressMode.CLAMP
-    filter_mode: FilterMode = FilterMode.POINT
-    read_mode: ReadMode = ReadMode.ELEMENT_TYPE
+    address_mode: AddressModeType | tuple[AddressModeType, ...] = AddressModeType.CLAMP
```
rwgk: Suggested change:

Include `str` in all four enum-bearing field annotations:

```python
address_mode: AddressModeType | str | tuple[AddressModeType | str, ...] = AddressModeType.CLAMP
filter_mode: FilterModeType | str = FilterModeType.POINT
read_mode: ReadModeType | str = ReadModeType.ELEMENT_TYPE
mipmap_filter_mode: FilterModeType | str = FilterModeType.POINT
```

Rationale (gpt-5.6-sol):

Before this PR, `TextureDescriptor` used `IntEnum` fields and the runtime rejected strings, so the enum-only annotations were accurate. This PR deliberately changes the runtime behavior: `__post_init__` and `_normalize_address_modes` now accept and coerce plain strings, as required by item 4. Keeping enum-only annotations therefore makes the generated stub reject calls that the implementation supports, for example `TextureObjectOptions(filter_mode="linear")`.

This is a compatible typing-only widening, not another runtime behavior change. It also matches existing `cuda.core` convention, including `Program(..., code_type: SourceCodeType | str)`, `Program.compile(target_type: ObjectCodeFormatType | str)`, `Linker.link(target_type: ObjectCodeFormatType | str)`, and `ManagedMemoryResourceOptions.preferred_location_type: ManagedMemoryLocationType | str | None`. Although `__post_init__` normalizes three fields back to enum instances, the pragmatic field annotations should describe the inputs accepted by the generated dataclass constructor.

cuda_core/cuda/core/typing.py
```diff
@@ -219,4 +223,69 @@ class VirtualMemoryAllocationType(StrEnum):
     MANAGED = "managed"
 
 
+class ArrayFormatType(StrEnum):
```
rwgk, line 226: This lacks `.. versionadded:: 1.1.0`

cuda_core/tests/test_texture_surface.py
rwgk: @juenglin see AGENTS.md

rwgk (commented):
@Andy-Jost what's your opinion regarding the suggested `str` type annotations?

cuda_core/tests/test_texture_surface.py
rwgk: @juenglin your question made me realize that I forgot to add a Purpose section to the PR #2245 description. That's fixed now. Please look there.

cuda_core/cuda/core/texture/_texture.pyx
Andy-Jost: I'd prefer if the `*Options` classes were immutable but that's definitely out of scope here

cuda_core/cuda/core/texture/_texture.pyi
Andy-Jost: Agreed — done. I applied the `str` widening to the source annotations in `_texture.pyx` (`TextureObjectOptions`: `address_mode`, `filter_mode`, `read_mode`, `mipmap_filter_mode`) rather than the generated stub, since the `.pyi` is auto-regenerated by the stubgen pre-commit hook and would otherwise be clobbered. The regenerated stub now reflects it. As you noted, this matches existing convention (`Program.compile(target_type: ObjectCodeFormatType | str)`, etc.) and is a typing-only widening.

cuda_core/cuda/core/typing.py
Andy-Jost: Fixed — added `.. versionadded:: 1.1.0` here. Your comment made me audit the rest: the other three new enums (`AddressModeType`, `FilterModeType`, `ReadModeType`) had the same gap, and several texture classes (`ResourceDescriptor`, `TextureObject`, `OpaqueArray`, `MipmappedArray`, `SurfaceObject`) lacked a class-level marker. Added class-level markers to all and removed the redundant method-level ones. Pushed in fa7cb36.

Andy-Jost:
@rwgk Thanks for the review. I agree with the `str` annotations — since this PR makes `TextureObjectOptions` coerce plain strings at runtime, the annotations should describe what the constructor accepts, and it matches existing convention (`Program.compile`, `Linker.link`, `ManagedMemoryResourceOptions`). Pushed in fa7cb36:

- Widened `address_mode`, `filter_mode`, `read_mode`, `mipmap_filter_mode` to include `str` (in `_texture.pyx`; stub regenerated) and noted "Plain strings are accepted" on each.
- Added `.. versionadded:: 1.1.0` to all four new texture enums and the texture classes that were missing a class-level marker (your `ArrayFormatType` comment prompted the wider audit).

No runtime behavior change beyond what the PR already introduced. Ready for another look.

rwgk (approved):
Looks great to me!

cuda_core/cuda/core/texture/_texture.pyi
leofang: This discussion came after I reviewed the PR last night. I think based on our Tuesday meeting discussion for https://github.com/NVIDIA/cuda-python/issues/2248 we don't want to do such widening ?

PR #2281, opened 2026-06-30
cuda.core: add tests for ObjectCode.from_library

lijinf2:
## Description

closes #663.

Adds tests for `ObjectCode.from_library` and extends the saxpy test-binary infrastructure from #2193.
- `build_test_binaries.sh`: package `saxpy.o` into `saxpy.a` for Linux / `saxpy.lib` for Windows via `nvcc -lib`
- CI (`build-wheel.yml`): upload library artifacts alongside object fixtures
- `test_module.py`: new `get_saxpy_library` fixture and three tests mirroring `from_object` coverage (bytes, file path, linker integration)

## Checklist

- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #2310, opened 2026-07-07
Add content seals and validation for generated CUDA Bindings files

rwgk:
## Summary

- Regenerate CUDA Bindings with cybind MR 460, adding format 1 content seals to 68 generated Python, Cython, template, and RST files.
- Add `toolshed/check_generated_file_seals.py` to validate seal syntax, format, comment style, and the recorded SHA-256 digest.
- Run the validator from `pre-commit` for text files under `cuda_bindings/`.

Each seal contains a machine-readable generated-file marker and a SHA-256 digest of the complete file content excluding the seal line itself. This makes ordinary manual edits detectable while also providing a consistent marker for generated-file discovery. The digest is an integrity check for accidental changes, not a cryptographic attestation of provenance.

## Dependency

The generated files in this PR were produced by cybind MR 460, which implements the producer side of the seal format. This PR implements the corresponding CUDA Python consumer-side validation and should remain a draft until that format is finalized.

## Relation to PR 2297

[PR 2297](https://github.com/NVIDIA/cuda-python/pull/2297) proposes removing the license-era `CI: Restricted Paths Guard`; its discussion also asks what should replace that workflow's incidental protection against manual edits to generated files.

For that generated-file protection, a separate GitHub Actions workflow is unnecessary. This PR puts the check in the repository's existing pre-commit suite. A manual edit changes the content without updating its seal, the seal hook fails, and the required pre-commit CI check blocks the PR from merging. The same rule applies to every contributor, without employee-detection logic or path-specific review labels.

The seal is intentionally an integrity check, not an authorization boundary. If the project retains a separate policy that only NVIDIA employees may modify particular paths, that policy would require separate enforcement; it is not needed to detect ordinary manual changes to cybind-generated files.

## Example failure

After manually modifying a sealed generated file, pre-commit reports:

```text
Check generated-file seals...............................................Failed
- hook id: check-generated-file-seals
- exit code: 1

Manual changes detected in 'cuda_bindings/cuda/bindings/cynvrtc.pyx'. It is a generated file and should not be edited directly.
Recorded content SHA-256: 9725757222bfc514253ab6a6113b4e6b1c33fab841fe2fff8ebb1012ecf7715b
Computed content SHA-256: f752f676ca6b4a5a99608c86c4581ca78958a0b0d233b2a931db201bfdc6aefd

```

## Unrelated piggy-backed tiny fix (to avoid making a separate PR)

- commit d8733610ed9e7c41cb6226aa585dda9a4f16c435: Make the RDC test-binary helper consistently use its resolved NVCC command for both compilation and library creation.


mdboom:
I'll come back to review this in a little bit, but let's merge https://github.com/NVIDIA/cuda-python/pull/2316 to catch us up to the `main` branch of the generator before merging this.

toolshed/check_cybind-generated_file_seals.py
```diff
@@ -0,0 +1,129 @@
...
+        print(f"INVALID cybind-generated file seal comment prefix in {filepath!r}")
+        return False
+
+    unsealed_blob = b"".join(lines[:marker_index] + lines[marker_index + 1 :])
+    computed_digest = hashlib.sha256(unsealed_blob).hexdigest().encode("ascii")
+    recorded_digest = match.group("digest")
+    if recorded_digest != computed_digest:
+        print(
+            f"MISMATCHED cybind-generated file seal in {filepath!r}: "
+            f"recorded {recorded_digest.decode()}, computed {computed_digest.decode()}"
```
mdboom: This message might be a bit confusing for someone who doesn't understand the motivation here.  Maybe:

"Manual changes detected in {filepath!r}.  It is a generated file and should not be edited directly."

And then show the hashes if you want (though I don't know how useful they are to the user).

toolshed/check_generated_file_seals.py
```diff
@@ -0,0 +1,129 @@
...
+from pathlib import Path, PureWindowsPath
+
+GENERATED_FILE_SEAL_TOKEN = "CYBIND-GENERATED-DO-NOT-MODIFY-THIS-FILE:"  # noqa: S105
+SUPPORTED_GENERATED_FILE_SEAL_FORMATS = frozenset({1})
+
+_TOKEN_BYTES = GENERATED_FILE_SEAL_TOKEN.encode("ascii")
+_MARKER_REGEX = re.compile(
+    rb"^(?P<prefix>#|\.\.) "
+    + re.escape(_TOKEN_BYTES)
+    + rb" format=(?P<format>[0-9]+); content-sha256=(?P<digest>[0-9a-f]{64})\n$"
```
mdboom, line 22: I think the `\n$` here means it might fail on Windows where the files could have `\r\n` line endings.  Would this work just as well?

```suggestion
    + rb" format=(?P<format>[0-9]+); content-sha256=(?P<digest>[0-9a-f]{64})$"
```

mdboom, line 22: Not strictly necessary because we control both ends, but users would probably expect arbitrary amounts of space between the keys.

```suggestion
    + rb" *format=(?P<format>[0-9]+) *; *content-sha256=(?P<digest>[0-9a-f]{64})\n$"
```

toolshed/check_generated_file_seals.py
```diff
@@ -0,0 +1,129 @@
...
+)
+_SOURCE_SUFFIXES = frozenset({".py", ".pxd", ".pxi", ".pyx"})
```
mdboom: This is basically used to get a comment character for specific file types.  To make this more extensible for more filetypes, I'd just be super explicit:

```
_COMMENT_CHARS = {
   ".py": b"#",
   ".pxd": b"#",
   ".pxi": b"#",
   ".pyx": b"#",
   ".pyx.in": b"#",
   ".pxd.in": b"#",
   ".pxi.in": b"#",
   ".rst": b"..",
   ".c": b"//",
   ".cpp": b"//",
   ".h": b"//",
}
```

toolshed/check_generated_file_seals.py
```diff
@@ -0,0 +1,129 @@
...
+        blob = Path(filepath).read_bytes()
+    except OSError as error:
+        print(f"ERROR reading {filepath!r}: {error}")
+        return False
+
+    lines = blob.splitlines(keepends=True)
+    marker_indexes = [index for index, line in enumerate(lines) if _TOKEN_BYTES in line]
+
+    if not marker_indexes:
+        if normalized_path in previously_sealed_paths:
```
mdboom, line 92: It's probably not worth fixing, but worth noting that if a user renames and edits a file in the same commit, it will not get flagged.

toolshed/check_cybind-generated_file_seals.py
```diff
@@ -0,0 +1,129 @@
...
+    if suffix in _SOURCE_SUFFIXES:
+        return b"#"
+    if suffix == ".rst":
+        return b".."
+    return None
+
+
+def load_previously_sealed_paths():
+    process = subprocess.run(  # noqa: S603
+        ["git", "grep", "-l", "-I", "-F", GENERATED_FILE_SEAL_TOKEN, "HEAD", "--"],  # noqa: S607
```
mdboom: No point in grepping outside of `cuda_bindings` (since the pre-commit config limits it there as well).

```suggestion
        ["git", "grep", "-l", "-I", "-F", GENERATED_FILE_SEAL_TOKEN, "HEAD", "--", "cuda_bindings"],  # noqa: S607
```

mdboom (changes requested):
Let's improve the message -- the rest is just suggestions.

Also, let's merge #2316 first.

cuda_bindings/cuda/bindings/cudla.pyx
```diff
@@ -3,6 +3,7 @@
 
 # <<<< PREAMBLE CONTENT >>>>
 
+# CYTHON-BINDINGS-GENERATED-DO-NOT-MODIFY-THIS-FILE: format=1; content-sha256=6e7ac86c22e602c08df8250f3b50c135945378aa8ae4ddb4e10174fd979c4aa5
```
mdboom: Slightly weird placement here (but not worth blocking on).

cuda_core/tests/test_binaries/build_test_binaries.sh
```diff
@@ -20,9 +20,9 @@ NVCC="${NVCC:-nvcc}"
     -o "${SCRIPTPATH}/saxpy.o" "${SCRIPTPATH}/saxpy.cu"
 
 if [[ "${OS:-}" == "Windows_NT" ]]; then
-    nvcc -lib -o "${SCRIPTPATH}/saxpy.lib" "${SCRIPTPATH}/saxpy.o"
+    "${NVCC}" -lib -o "${SCRIPTPATH}/saxpy.lib" "${SCRIPTPATH}/saxpy.o"
```
mdboom, line 23: Seems unrelated?  Maybe accidentally brought over from another branch?

mdboom (changes requested):

toolshed/check_generated_file_seals.py
mdboom: I tested this on Windows, and it's actually fine.  (However a bunch of other things in our pre-commit hooks are broken on Windows -- I'll file a separate issue for that).

toolshed/check_cybind-generated_file_seals.py
rwgk: Done. The failure now explains that a manual change was detected in a generated file and that the file should not be edited directly. I kept the recorded and computed hashes as secondary diagnostic information (e.g. if someone accidentally cuts off characters in the seal line, or parsing the seal line goes subtly wrong, that'd be pretty obvious).

toolshed/check_generated_file_seals.py
rwgk: I think we should keep the explicit `\n`. Cybind normalizes generated content to LF before sealing and writes the sealed output as bytes, and cuda-python enforces `* text eol=lf` in `.gitattributes`.

Also, because the checker uses `fullmatch`, removing `\n` would reject the existing LF-terminated marker lines rather than make the expression line-ending agnostic. Supporting CRLF would require a coordinated change to the marker parsing and content-digest normalization, rather than this regex-only change.

rwgk: I'd prefer to keep one canonical spelling here. We control both the producer and consumer, and this is a new machine-generated format rather than something people are expected to author manually. Strict spacing catches unexpected format drift early and also keeps ad-hoc tooling, including simple `grep`-based inspection, straightforward and predictable.

toolshed/check_generated_file_seals.py
rwgk: Done, thanks!

toolshed/check_generated_file_seals.py
rwgk: As long as the renamed file is staged, a rename followed by an edit is still detected: the seal moves with the file, and the changed content no longer matches its recorded digest.

This pre-commit check is is meant to be an integrity check for ordinary accidental changes. I don't think we should expand the scope beyond that. A rename of a generated file is probably not an accident. Guarding against removing a seal or forging one is also out of scope.

toolshed/check_cybind-generated_file_seals.py
rwgk: Done.

cuda_core/tests/test_binaries/build_test_binaries.sh
rwgk: See my comment under 2281:

https://github.com/NVIDIA/cuda-python/pull/2281#discussion_r3540401109

It's a really tiny fix, I hope to avoid creating a separate PR just for that by piggy-backing here.

cuda_bindings/cuda/bindings/cudla.pyx
rwgk: I agree this isn't pretty, but yes, it'll be great to keep to ball rolling.

rwgk: FYI — I'm taking another look, because I want to retest anyway (after merging the cybind mainline into MR 460).

rwgk: I found a nice solution that's already pushed to MR 460. 6 files are changed here:

* commit 9e91f7ad5193fbcc2296275b4cc8ad427c18f83a

While working on that, I discovered that `grep -l -F CYTHON-BINDINGS-GENERATED-DO-NOT-MODIFY-THIS-FILE` matched `toolshed/check_generated_file_seals.py`, so I fixed that, too:

* commit 89bccba8f05be2dd44bf6a3fcdf4c578c4c8f9c6

I'm holding off triggering the CI again until #2323 is merged.

mdboom (approved):
LGTM.  Now that the generator side of this has been merged, we should probably regenerate one more time over here (probably a no-op) and then assuming that's clean, let's merge this.

PR #2312, opened 2026-07-07
Add Windows ARM64 build support

isVoid:
## What changed

- Select the CUDA Toolkit `lib\arm64` directory when building cuda.bindings on `win-arm64`.
- Load `nvcudaa64.dll` from the Windows CUDA loader path on ARM64 builds.
- Build the cuda.core AOTI shim import library with `/MACHINE:ARM64` for `win-arm64`.

## Why

These changes add native Windows ARM64 build support for cuda.bindings and the cuda.core tensor bridge AOTI shim while preserving existing x64 behavior.

## Validation

- `pixi run python -m py_compile cuda_bindings/build_hooks.py`
- `pixi run python -m py_compile cuda_core/setup.py cuda_bindings/build_hooks.py`
- `git diff --check -- cuda_core/setup.py cuda_bindings/build_hooks.py cuda_bindings/cuda/bindings/_bindings/loader.cpp`
- pre-commit hooks during both commits


isVoid:
This PR is to additionally verify that the WoA build changes does not affect the existing build system by running on the full CI matrix.

rwgk (approved):

PR #2316, opened 2026-07-07
Chore: Catch up generated code

mdboom:
This catches up our generated code to the current version of the generator.

The two changes are: 

- Dropping information about the generator version
- The new preamble-based approach to importing external things and generating helper functions.


rwgk (approved):
Thanks @mdboom!

PR #2318, opened 2026-07-07
[doc-only] docs: clarify cuda.core concurrency and thread-safety policy

Andy-Jost:
## Summary

Writes down cuda.core's concurrency and thread-safety expectations, which were previously agreed only in review threads (#2280) and a follow-up Slack discussion, but never documented. Adds a short user-facing policy page and a matching set of contributor-facing invariants. Docs only; no code or behavior changes.

The policy is intentionally minimal, per the discussion: state what users can rely on today without overcommitting to guarantees we would have to maintain long-term.

## Policy

- Concurrent **reads** of a cuda.core object from multiple host threads are supported.
- Concurrent **mutation** of the same object is not supported; the application must provide external synchronization (as is customary for NumPy, CuPy, and PyTorch).
- The library protects the integrity of its own internal state, so concurrent reads of one object and concurrent use of distinct objects cannot corrupt the interpreter. This is an integrity guarantee only.
- Distinct objects can still collide through shared CUDA driver/context state (e.g. changing peer device access while another thread accesses affected memory); synchronizing those cases is also the caller's responsibility.

## Changes

- `docs/source/concurrency.rst`: new user-facing "Concurrency and Thread Safety" page, linked into the docs toctree (`docs/source/index.rst`).
- `cuda_core/AGENTS.md`: new "Concurrency and free-threading" section for contributors and reviewers -- reads-safe/mutation boundary, prefer immutable designs, shared-driver-state collisions, guarding internal cached state only for legitimately-shared objects (not thread-local ones like `Device`, cf. #2321), GIL-held entry points, and releasing the GIL before entering the driver from callback/`__del__` paths.

## Related Work

- Consolidates decisions from #2280 (graph slot-table review) and the follow-up thread-safety discussion.
- Reflects the free-threading hardening in #2162, #2215, #2216, #2217.
- Part of EPIC #350.

### DIFF
--- a/cuda_core/AGENTS.md
+++ b/cuda_core/AGENTS.md
@@ -63,6 +63,44 @@ This file describes `cuda_core`, the high-level Pythonic CUDA subpackage in the
 - Prefer explicit error propagation over silent fallback paths.
 - If you change public behavior, update tests and docs under `docs/source/`.
 
+## Concurrency and free-threading
+
+`cuda.core` ships free-threaded (no-GIL) wheels and builds with
+`freethreading_compatible=True`. The user-facing policy lives in
+`docs/source/concurrency.rst`; the invariants below are for contributors. Reviewers
+and agents should flag violations.
+
+- **Reads are safe; mutation is the boundary**: concurrent reads of an object are
+  supported, but concurrent mutation of the same public object (e.g., building one
+  graph from two threads, or `close()` racing another call) is the caller's
+  responsibility -- do not add locks to make it safe. Prefer immutable designs to
+  keep the mutable, thread-unsafe surface small. Protecting library-internal state
+  *is* in scope.
+- **Distinct objects can still collide via shared driver state**: operating on
+  separate objects is not automatically safe when they share driver or context
+  state (e.g., changing peer device access while another thread touches affected
+  memory). Synchronizing these cases is the caller's responsibility; do not try to
+  lock around them internally.
+- **Protect internal cached/module-level state**: guard lazily-populated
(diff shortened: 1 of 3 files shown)


cuda_core/AGENTS.md
```diff
@@ -63,6 +63,44 @@ This file describes `cuda_core`, the high-level Pythonic CUDA subpackage in the
...
+  lock around them internally.
+- **Protect internal cached/module-level state**: guard lazily-populated
+  `cdef object` caches and module-level state so concurrent access cannot corrupt
+  interpreter state (CPython reference counts -- a strictly free-threading hazard).
+  Established patterns are `@cython.critical_section` on accessors (#2215), an
+  atomic initialization flag (#2216), and `dict.setdefault` for identity caches
+  (#2217). Guard state only on objects that are legitimately shared between threads;
+  objects that are not meant to be shared (e.g., the thread-local `Device`) do not
+  need such guards (see #2321). Reference-count integrity is guaranteed; cache
+  value-identity/idempotency is not.
```
seberg, line 92: Yeah, I think there are a few objects where we want idempotency.  But not everwhere (places that use `dict.setdefault()` to ensure this).

seberg (approved):
Thanks, much nicer than what I could have come up with! And I don't have a concrete improvement thought.

(My small thoughts are at best if we are comfortable to say that e.g. memory pools is, I believe, safe and might seem mutable. And I suspect most things are thread-safe in the sense of not corrupting state, unless it's a green context or so, but I got a feeling that information likely wouldn't add much even if we were comfortable to note it.)

PR #2320, opened 2026-07-08
Check nvJitLink driver major compatibility

### LINKED ISSUE #712: Re-work on `Linker` dispatching logic

isVoid:
## Summary

  This PR improves linker backend selection and fixes a cuLink storage-lifetime regression.

  - Handle systems where the CUDA driver major version is newer than the available nvJitLink major version. In this case, `cuda.core` emits a warning and falls back to the cuLink driver APIs instead of selecting an incompatible nvJitLink backend.
  - Keep the option arrays and log buffers passed to `cuLinkCreate` alive for the full lifetime of the `CUlinkState`. The state is now destroyed with `cuLinkDestroy` before its referenced storage is released, preventing use-after-free, heap corruption, and possible segmentation faults.

  ## Testing

  - Added coverage for nvJitLink and driver major-version compatibility.
  - Added cuLink regression coverage for repeated link/close operations and log access after closing the linker.

closes #712 
supersedes #1911


cuda_core/tests/test_program_cache.py
```diff
@@ -390,6 +396,17 @@ def test_make_program_cache_key_ptx_linker_equivalent_options_hash_same(a, b, mo
     assert k_a == k_b
 
 
+@pytest.mark.agent_authored(model="gpt-5")
+def test_make_program_cache_key_ptx_nvjitlink_time_values_hash_same(require_nvjitlink_backend, monkeypatch):
```
lijinf2: nit: could avoid defining `require_nvjitlink_backend` and use
`@pytest.mark.skipif(is_culink_backend, reason="test requires nvJitLink backend")`
instead, consistent with `test_linker.py`.

cuda_core/cuda/core/_linker.pyx
```diff
@@ -155,10 +155,21 @@ cdef class Linker:
...
             self._nvjitlink_handle.reset()
         else:
+            if self._drv_log_bufs is not None:
+                if self._info_log is None:
+                    self._info_log = self.get_info_log()
+                if self._error_log is None:
+                    self._error_log = self.get_error_log()
+            # Destroy the CUlinkState before releasing storage referenced by it.
             self._culink_handle.reset()
+            self._drv_jit_keys.swap(empty_keys)
```
lijinf2, line 170: one question (not blocking): is syntax `self._drv_jit_keys.swap(vector[cydriver.CUjit_option]())` supported in .pyx? C++ should support temporary object created as a function argument and I get used to it.

isVoid: > C++ should support temporary object created as a function argument and I get used to it.

Yes C++ support temp vars as function argument, but only if the function declare to support rvalue reference. `swap(vector& __x) _GLIBCXX_NOEXCEPT` is declared with lvalue reference so you can't pass a temporary object to argument there. [godbolt](https://godbolt.org/z/rMPEvE4os)

However, in cython both works. In either way presented cython will lower the arguments into a named local to support various c++ / python interfacing shenanigans. 

By cythonizing both and comparing the lowered code, the inlined version generate a slightly more complex C++ with additional assignment like below. So I suggest we keep the current declare-then-use form, not the inline construction form.

```c++
  try {
      __pyx_t_5 = std::vector<CUjit_option>();
  } catch (...) {
      __Pyx_CppExn2PyErr();
      goto error;
  }
  __pyx_v_self->_drv_jit_keys.swap(__pyx_t_5);

  try {
      __pyx_t_6 =
          std::vector<__pyx_t_4core_7_linker_void_ptr>();
  } catch (...) {
      __Pyx_CppExn2PyErr();
      goto error;
  }
```

lijinf2 (approved):
Looks pretty good. Thank you!

Also got the code reviewed by agent. Did not find any must-fix comments by examining the agent's suggestions.

cuda_core/cuda/core/_linker.pyx
lijinf2: Good findings. Was not aware that swap takes only lvalue as argument, and that cython is doing the same declare-then-use under the hood with rvalue. That makes sense. I learned a lot from the answer!

leofang: Couldn't we just shrink the vector sizes to zero and let the elements go out of scope?

isVoid: > Couldn't we just shrink the vector sizes to zero and let the elements go out of scope?

Both handles to the vectors now lives with the `Linker` class so they don't go out of scope as long as the Linker instance stays alive.

`clear()` and `resize(0)` both destroys the elements in the held vector but does not necessarily deallcoate the backing allocation just in case we need to reuse them. This could increase unwanted host side usage.

PR #2321, opened 2026-07-08
Revert SMResource splitting fix, the problem was the test

seberg:
## Description

I got things wrong, I think.  While completely harmless to have the lock, I think we can consider `SMResource` as thread-unsafe as it is based of `Device()` and the green-context where `Device()` is thread-local and green-contexts are by definition thread unsafe to share.

There is probably some documentation that would be good that sharing device objects across threads is unsafe (to the point of crashes).  But I now think the correct fix, at least for now, is to just fix the test and I have included that here.

@Andy-Jost sorry about this but FYI since you merged the other PR for me.

### DIFF
--- a/cuda_core/cuda/core/_device_resources.pxd
+++ b/cuda_core/cuda/core/_device_resources.pxd
@@ -2,8 +2,6 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-cimport cython
-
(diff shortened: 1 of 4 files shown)


Andy-Jost (approved):
The changes look good. Under the concurrency rules, the mutex is not needed here.

PR #2323, opened 2026-07-08
Fix pre-commit CI

mdboom:
Numpy 2.5.0 introduced new syntax in their `.pyi` file that is not supported on Pythons prior to 3.12.0.  We tell mypy that we must support all the way back to Python 3.10, so this fails with:

```
/pc/clone/iiMN7YpuTkSqCPCVYmy4Hg/py_env-python3/lib/python3.14/site-packages/numpy/__init__.pyi:737: error: Type statement is only supported in Python 3.12 and greater  [syntax]
```

Pinning to unblock CI.

This doesn't seem to be a bug on Numpy's part, since they only claim to support Python 3.12 currently.


leofang (approved):

leofang:
Thanks, Mike! 

Weird that the assignee check failed but the PR is merged... 🤔

rwgk:
> Weird that the assignee check failed but the PR is merged... 🤔

It was never made merge-blocking. The history is that Rob started it, I helped a little with follow-up work, and I think we both wanted to let it run for a while and confirm it was stable before making it required. We never got back to that last step. Should I create an issue so we don’t forget?

leofang:
Ah I just made it required and was about to circle back and leave a paper trail here.

PR #2324, opened 2026-07-08
Remove vendored deprecated library

mdboom:


cuda_core/cuda/core/system/_device.pyi
```diff
@@ -792,7 +790,6 @@ class MigInfo:
 class _NvlinkInfoMeta(type):
 
     @property
-    @deprecated(version='1.1.0', reason='Use Device.get_nvlink_count instead to get the actual number of Nvlinks available on a specific device.')
     def max_links(cls):
```
leofang, line 793: Q: Maybe we should update the docstring?

leofang (approved):

cuda_core/cuda/core/system/_device.pyi
mdboom: We could.  I didn't think there was much point without the warning being emitted -- but I suppose I can add that manually as well.

PR #2327, opened 2026-07-08
Make pre-commit work on Windows

### LINKED ISSUE #2322: pre-commit hooks are broken on Windows
Running `pre-commit run --all-files` on a clean checkout on Windows fails in a number of ways.

This prevents developers on Windows (non-WSL) from effectively working on this codebase.

We should fix any bugs in the pre-commit checks, and then add a job to CI to run `pre-commit run --all-files` on Windows (or add a config option for `pre-commits` built-in CI to get it to run on Windows as well, if that's possible).

mdboom:
Our pre-commit config didn't work on Windows, which makes things harder than it already is for Windows development.

- The `.pyi` generation step would create line ending churn.  This adds a wrapper around our use of `stubgen-pyx` so normalize line endings.  (Maybe someday that can be an upstream contribution to `stubgen-pyx` but this is good enough for now).

- `lychee` doesn't run on Windows.  Unfortunately, there is no way to config a task to only run on certain platforms.  The best I or my agent was able to come up with was to set an `envvar` to skip it.

- Doing an end-of-file check on a symlink would fail (they are files without newlines at the end on Windows).  This updates the pre-commit config to ignore all symlinks for that task.

- `mypy` has never run on Windows before, and it found a few typing bugs, fixed here.

- Lastly, this adds pre-commit CI on Windows so these fixes will stay in place.  (And having Windows mypy checking seems useful).


.pre-commit-config.yaml
```diff
@@ -68,6 +68,7 @@ repos:
     hooks:
       - id: lychee
         args:
+
```
leofang: ?

CONTRIBUTING.md
```diff
@@ -74,6 +75,14 @@ between commits, leaving stale headers or out-of-date stubs in the history.
...
 warning reminding you to run `pre-commit install`.
 
+### Pre-commit on Windows
+
+For development on Windows (not WSL), the `lychee` pre-commit task will not work.
+You will need to set the following environment variable to make it pass:
+
+```
+SKIP=lychee
+```
```
leofang: Q: Is it not possible to add it to the pre-commit config?

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
leofang, line 1: Touching this file seems unnecessary, why does pre-commit care on Windows?

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_windows.py
```diff
@@ -21,7 +21,10 @@
 POINTER_ADDRESS_SPACE = 2 ** (struct.calcsize("P") * 8)
 
 # Set up kernel32 functions with proper types
-kernel32 = ctypes.windll.kernel32  # type: ignore[attr-defined]
+windll = getattr(ctypes, "windll", None)
+if windll is None:
+    raise RuntimeError("ctypes.windll is required on Windows")
```
leofang, line 28: This seems... dumb to me TBH. We guaranteed this file is only executed on Windows, no? How is it possible that `windll` is not available on Windows...?

CONTRIBUTING.md
leofang: I see this was answered in the PR description... alright

.github/workflows/ci.yml
```diff
@@ -416,6 +416,35 @@ jobs:
     with:
       is-release: ${{ github.ref_type == 'tag' }}
 
+  precommit-windows:
```
rwgk, line 504: This job executes PR-controlled pre-commit hooks but inherits the repository's default token permissions. Because this workflow runs on upstream `push` refs (`pull-request/*`), the live job received broad write scopes, and `actions/checkout` persists that credential by default. A modified local hook could therefore write through Git. Could we add job-level `permissions: { contents: read }` and set `persist-credentials: false` on the checkout?

.pre-commit-config.yaml
```diff
@@ -88,6 +89,7 @@ repos:
     - id: check-yaml
     - id: debug-statements
     - id: end-of-file-fixer
+      exclude_types: [symlink]
```
rwgk: `exclude_types: [symlink]` only works when Git materializes real filesystem symlinks. With `core.symlinks=false`, Git for Windows checks these out as regular one-line files; `identify` tags them as `text`, so `end-of-file-fixer` still rewrites them. I reproduced this with `CLAUDE.md`, which changed from `AGENTS.md` to `AGENTS.md\n`. Could we exclude the tracked symlink paths (or use an index-mode-aware wrapper) and add a `core.symlinks=false` CI case?

CONTRIBUTING.md
```diff
@@ -74,6 +75,14 @@ between commits, leaving stale headers or out-of-date stubs in the history.
 If the hook isn't installed, `pre-commit run` (and CI) will print a visible
 warning reminding you to run `pre-commit install`.
 
+### Pre-commit on Windows
+
+For development on Windows (not WSL), the `lychee` pre-commit task will not work.
+You will need to set the following environment variable to make it pass:
+
+```
+SKIP=lychee
```
rwgk: This is not directly usable in native Windows shells: PowerShell needs `$env:SKIP = "lychee"`, and cmd needs `set SKIP=lychee`; as a standalone Git Bash assignment it is not exported to a later `pre-commit` process. Could we provide shell-specific commands (or an inline Git Bash example) so the workaround can be followed literally?

CONTRIBUTING.md
```diff
@@ -74,6 +75,14 @@ between commits, leaving stale headers or out-of-date stubs in the history.
 If the hook isn't installed, `pre-commit run` (and CI) will print a visible
 warning reminding you to run `pre-commit install`.
 
+### Pre-commit on Windows
+
+For development on Windows (not WSL), the `lychee` pre-commit task will not work.
+You will need to set the following environment variable to make it pass:
```
rwgk: Lychee itself supports Windows, and the Windows CI run passed under Git Bash. Could we narrow this wording to the hook's actual limitation? Suggested text:

> The auto-installing `lychee` hook uses a Bash wrapper. On native Windows, run pre-commit from Git Bash. If Bash is unavailable, skip the hook by setting `SKIP=lychee` using your shell's environment-variable syntax:

rwgk (commented):
Some gpt-5.6-sol generated comments.

It figured out that lychee actually runs on Windows, but only with a suitable `bash`:

https://github.com/NVIDIA/cuda-python/actions/runs/28974154791/job/85977090993#step:5:1

.pre-commit-config.yaml
mdboom:
```suggestion
```

CONTRIBUTING.md
mdboom: Yeah.  Seems like a big hole in pre-commit.

The other downside is that `lychee` doesn't run in our CI (because it's slow).  So patches that come from Windows won't get `lychee`-tested.

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
mdboom: Because `mypy` is now checking this on Windows, where things like `os.RTLD_NOW` don't exist.

`mypy` doesn't/can't know that this is only imported on Linux.  It does have special support for parsing `if sys.platform == "linux"` (which is why this works), but it can't follow that between modules.  (And we use slightly different logic there anyway).

IME, this stuff is necessary to make type-checking complete, annoying as it is.

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_windows.py
mdboom: In the type information, it's specified as `| None`, because it doesn't exist in all places.  `mypy` is a static analyzer, so it knows nothing about the runtime state of the program.

.pre-commit-config.yaml
mdboom: Hmm...  This worked for me locally, but I will have another look.

CONTRIBUTING.md
mdboom: I feel like asking developers to work in Git Bash on Windows is a non-starter.  They would have to to development in PowerShell and only git commands in Git Bash.  However, let me see if Git Bash gets used implicitly when pre-commit is installed as a hook -- that would be workable.

.pre-commit-config.yaml
mdboom: I see -- it works as part of the installed hook (which only looks at changed fails) but fails for `pre-commit run --all-files`.  I think just adding the names of the symlinks to the exclude pattern should be good enough (though will break when we add symlinks in the future).

CONTRIBUTING.md
mdboom: As @rwgk pointed out below, this works correctly when the `pre-commit` hook is installed (which we want to encourage anyway, so the problem isn't as big a deal as first thought).

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_windows.py
mdboom: I've refactored this to use `sys.platform == "win32"` (which mypy recognizes) rather than `IS_WINDOWS` (which it can't).  It makes things simpler.  The linux module changes are still required but are less dynamic and "dumb".

mdboom: I have simplified the type-checking somewhat, but this specific thing is required.  You have to imagine "what happens if we load this code on any platform" and accommodate all cases.

CONTRIBUTING.md
mdboom: When the pre-commit hook is installed, this works correctly.  So the problem isn't as bad as it first appeared.  I've tweaked the instructions to be more specific.

.pre-commit-config.yaml
```diff
@@ -88,7 +88,7 @@ repos:
     - id: check-yaml
     - id: debug-statements
     - id: end-of-file-fixer
-      exclude: &gen_exclude '^(?:cuda_python/README\.md|cuda_bindings/cuda/bindings/.*\.in?|cuda_bindings/docs/source/module/.*\.rst?|.*\.pyi)$'
+      exclude: &gen_exclude '^(?:cuda_python/README\.md|(?:^|/)CLAUDE\.md|(?:^|/)\.git_archival\.txt|cuda_bindings/cuda/bindings/.*\.in?|cuda_bindings/docs/source/module/.*\.rst?|.*\.pyi)$'
```
rwgk: This still misses eight nested symlink placeholders. Because the complete expression is anchored with `^`, the inner `(?:^|/)` alternatives are evaluated at offset 0 and cannot match paths such as `cuda_core/CLAUDE.md` or `cuda_pathfinder/.git_archival.txt`. In a `core.symlinks=false` checkout, `end-of-file-fixer` rewrote all eight nested placeholders. Could we use something like `(?:.*/)?(?:CLAUDE\.md|\.git_archival\.txt)` inside the outer anchored alternation (or list the tracked paths explicitly), and ideally exercise a `core.symlinks=false` checkout in CI?

cuda_pathfinder/cuda/pathfinder/_utils/driver_info.py
```diff
@@ -63,10 +63,8 @@ def query_driver_cuda_version() -> DriverCudaVersion:
 def _query_driver_cuda_version_int() -> int:
     """Return the encoded CUDA driver version from ``cuDriverGetVersion()``."""
     loaded_cuda = _load_nvidia_dynamic_lib("cuda")
-    if IS_WINDOWS:
-        # `ctypes.WinDLL` exists on Windows at runtime. The ignore is only for
-        # Linux mypy runs, where the platform stubs do not define that attribute.
-        loader_cls: Callable[[str], ctypes.CDLL] = ctypes.WinDLL  # type: ignore[attr-defined]
+    if sys.platform == "win32":
```
rwgk: The direct `sys.platform` check is useful here because mypy understands the platform narrowing, but removing `IS_WINDOWS` leaves the existing tests patching a deleted attribute (`test_utils_driver_info.py:49` and `:96`), which is why every current Windows test job fails with `AttributeError`.

I asked my agent about teaching tests to keep `IS_WINDOWS` and `sys.platform` synchronized, but it said that's brittle. As a safer alternative, could we make the selected implementation the injectable seam?

```python
if sys.platform == "win32":
    _DRIVER_LIB_LOADER: Callable[[str], ctypes.CDLL] = ctypes.WinDLL
else:
    _DRIVER_LIB_LOADER = ctypes.CDLL
```

and then:

```python
driver_lib = _DRIVER_LIB_LOADER(loaded_cuda.abs_path)
```

Unit tests can patch `_DRIVER_LIB_LOADER` directly, while a small native-platform test can assert that it is `ctypes.WinDLL` when `IS_WINDOWS` and `ctypes.CDLL` otherwise. This keeps direct `sys.platform` confined to the mypy-recognized selection boundary, avoids globally pretending the process is another OS, and lets Linux/Windows CI verify the real selection.

CONTRIBUTING.md
```diff
@@ -74,6 +75,17 @@ between commits, leaving stale headers or out-of-date stubs in the history.
...
+
+For development on Windows (not WSL), the `lychee` pre-commit task will not work
+when running `pre-commit run --all-files`.  This problem does not occur if you
+install the pre-commit hook and run it automatically as part of your `git
+commit` workflow.  To resolve this, you can either:
+
+1. Run `pre-commit` it in Git Bash, rather directly in PowerShell or cmd
+
+2. Skip it by setting the environment variable `SKIP` to `lychee`.  This would
+   be `$env:SKIP = "lychee"` in PowerShell or `SKIP=lychee` in cmd.
```
rwgk: Command Prompt requires `set SKIP=lychee` (plain `SKIP=lychee` is not cmd syntax).

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_windows.py
```diff
@@ -44,6 +48,14 @@
 ]
 kernel32.GetModuleFileNameW.restype = ctypes.wintypes.DWORD
 
+# AddDllDirectory (Windows 7+)
+kernel32.AddDllDirectory.argtypes = [ctypes.wintypes.LPCWSTR]
+kernel32.AddDllDirectory.restype = ctypes.c_void_p  # DLL_DIRECTORY_COOKIE
```
rwgk: These lines should be removed. PR #2399 intentionally switched to `os.add_dll_directory()` and removed this raw binding after @leofang's review (https://github.com/NVIDIA/cuda-python/pull/2399#discussion_r3633890088). This PR's merge commit 2cc0f30 retained the new stdlib call at line 89 but resurrected the old declaration; it looks like a merge artifact.

CONTRIBUTING.md
```diff
@@ -74,6 +75,17 @@ between commits, leaving stale headers or out-of-date stubs in the history.
...
 warning reminding you to run `pre-commit install`.
 
+### Pre-commit on Windows
+
+For development on Windows (not WSL), the `lychee` pre-commit task will not work
+when running `pre-commit run --all-files`.  This problem does not occur if you
+install the pre-commit hook and run it automatically as part of your `git
+commit` workflow.  To resolve this, you can either:
+
+1. Run `pre-commit` in Git Bash, rather directly in PowerShell or cmd
```
rwgk:
```suggestion
1. Run `pre-commit` in Git Bash, rather than directly in PowerShell or cmd
```

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_windows.py
mdboom:
```suggestion
```

rwgk (approved):
Thanks @mdboom for tackling this!

toolshed/run_stubgen_pyx.py
juenglin, line 1: Windows-style line breaks is still one difference I see when running this on Windows.

```
$ git diff
warning: in the working copy of 'cuda_core/cuda/core/_context.pyi', CRLF will be replaced by LF the next time Git touches it
warning: in the working copy of 'cuda_core/cuda/core/_device.pyi', CRLF will be replaced by LF the next time Git touches it
...
```

PR #2330, opened 2026-07-09
cuda.core: expose driver-populated fields on WorkqueueResource

leofang:
## Summary

Add read-only :attr:`sharing_scope`, :attr:`concurrency_limit`, and :attr:`device` properties on `WorkqueueResource` so users can inspect the driver-populated workqueue config without dropping to raw `cydriver`. Also adds `WorkqueueSharingScopeType` StrEnum in `cuda.core.typing` for a typed return; raw strings matching the enum member values remain accepted at runtime for backward compatibility.

Follows section 5(d) — _"Resources should be opaque but round-trippable"_ — from the [green ctx design doc](https://docs.google.com/document/d/1fNlXLJw_oy4GLyUhzxlDO5aC_q2GYbekJs-DN3y-hWg/edit). That principle was written into `SMResource` (which exposes `sm_count`, `min_partition_size`, etc.) but not applied to `WorkqueueResource` — this PR closes that asymmetry.

## Rationale

Before this change, `dev.resources.workqueue` returned an object whose driver-populated fields (`sharingScope`, `wqConcurrencyLimit`, `device`) were only reachable by:
- `ctypes.cast`-ing the raw `handle` pointer back to a `CUdevResource*` and reading the struct manually, or
- Re-issuing `cuDeviceGetDevResource` via `cuda.bindings.cydriver`.

Both routes require users to touch the driver ABI that `cuda.core` exists to hide.

## API surface

```python
wq = dev.resources.workqueue

wq.sharing_scope        # -> WorkqueueSharingScopeType (StrEnum, str-compatible)
wq.concurrency_limit    # -> int (driver cap, typically CUDA_DEVICE_MAX_CONNECTIONS)
wq.device               # -> cuda.core.Device

# Configure still accepts raw strings (backward compatible)
wq.configure(WorkqueueResourceOptions(sharing_scope=\"green_ctx_balanced\"))

# ...or the new StrEnum
from cuda.core.typing import WorkqueueSharingScopeType
wq.configure(WorkqueueResourceOptions(
    sharing_scope=WorkqueueSharingScopeType.GREEN_CTX_BALANCED,
    concurrency_limit=4,
))
```

The `WorkqueueResourceOptions.sharing_scope` type annotation is now `WorkqueueSharingScopeType | None` (was `str | None`). Since 1.1.0 hasn't shipped yet this is a pre-release refactor; runtime behavior is preserved (StrEnum members are `str`-equal to their value).

## Test coverage

Extended existing `TestWorkqueueResource` tests without touching the ones that exercise raw-string inputs (to preserve visible string-input coverage):
- `test_query` — now also asserts `isinstance(wq.sharing_scope, WorkqueueSharingScopeType)`, `wq.concurrency_limit >= 1`, `wq.device.device_id == init_cuda.device_id`.
- `test_configure_concurrency_limit` — now round-trips `concurrency_limit == 4` via the new property.
- Added `test_configure_scope_with_enum` — configures via `WorkqueueSharingScopeType.GREEN_CTX_BALANCED`, verifies the property returns that same enum member.
- Left alone: `test_configure_valid_scope`, `test_configure_concurrency_and_scope`, `test_invalid_scope_raises_at_construction`, both `_raises_at_construction` tests, `test_configure_none_is_noop` — these keep raw-string inputs exercised.

Full `test_green_context.py` locally: 36 passed, 2 skipped (arch), zero regressions.

## Related

- Follows up on #2329 (`concurrency_limit` write-side).
- Design doc will be updated to reflect the addition of read-only properties on `WorkqueueResource` (per 5(d)) and the new `WorkqueueSharingScopeType` StrEnum.

-- Leo's bot


cuda_core/cuda/core/_device_resources.pyi
```diff
@@ -44,9 +47,13 @@ class WorkqueueResourceOptions:
 
     Attributes
     ----------
-    sharing_scope : str, optional
-        Workqueue sharing scope. Accepted values: ``"device_ctx"``
-        or ``"green_ctx_balanced"``. (Default to ``None``)
+    sharing_scope : :class:`~cuda.core.typing.WorkqueueSharingScopeType`, optional
```
mdboom: I'm surprised this was missed by my sweep for "enum-like-things" (though I used an agent -- I guess it missed it).  Anyway, it's good to come back to it now and this isn't API-breaking.

cuda_core/cuda/core/_device_resources.pyx
```diff
@@ -126,9 +131,13 @@ cdef class WorkqueueResourceOptions:
...
-    sharing_scope : str, optional
-        Workqueue sharing scope. Accepted values: ``"device_ctx"``
-        or ``"green_ctx_balanced"``. (Default to ``None``)
+    sharing_scope : :class:`~cuda.core.typing.WorkqueueSharingScopeType`, optional
+        Workqueue sharing scope. Accepted values:
+        :attr:`~cuda.core.typing.WorkqueueSharingScopeType.DEVICE_CTX` or
+        :attr:`~cuda.core.typing.WorkqueueSharingScopeType.GREEN_CTX_BALANCED`.
+        Raw strings equal to the enum member values (``"device_ctx"``,
+        ``"green_ctx_balanced"``) are accepted at runtime for backward
+        compatibility. (Default to ``None``)
```
mdboom: Elsewhere in `cuda-core` we:

- Treat these as "strings first" and leave the enum just in the type annotation: https://github.com/NVIDIA/cuda-python/blob/main/cuda_core/cuda/core/_memory/_managed_memory_resource.pyi#L25

- Unless there is further documentation about the required values, we don't list them, since they are listed in the enum: https://github.com/NVIDIA/cuda-python/blob/main/cuda_core/cuda/core/_memory/_virtual_memory_resource.py#L46

I definitely don't think we need to describe the enum/str flexibility here (since it's standard across the library).  Maybe we should mention that somewhere centrally.

If we want to list the accepted values here, I think it should be as strings.

Also "Default to ``None`` is redundant because it says "optional" above.

```suggestion
        Workqueue sharing scope. Accepted values: ``"device_ctx"`` or 
        ``"green_ctx_balanced"``.
```

cuda_core/cuda/core/_device_resources.pyx
```diff
@@ -554,6 +563,57 @@ cdef class WorkqueueResource:
...
+        :attr:`WorkqueueResourceOptions.sharing_scope`.
+        """
+        IF CUDA_CORE_BUILD_MAJOR >= 13:
+            from cuda.core.typing import WorkqueueSharingScopeType
+            cdef object scope = self._wq_config_resource.wqConfig.sharingScope
+            if scope == cydriver.CUdevWorkqueueConfigScope.CU_WORKQUEUE_SCOPE_DEVICE_CTX:
+                return WorkqueueSharingScopeType.DEVICE_CTX
+            elif scope == cydriver.CUdevWorkqueueConfigScope.CU_WORKQUEUE_SCOPE_GREEN_CTX_BALANCED:
+                return WorkqueueSharingScopeType.GREEN_CTX_BALANCED
+            raise RuntimeError(f"Unknown sharing scope enum value: {scope}")
```
mdboom, line 579: If sharing scope is passed in as `None` (or not provided) this will raise.  Does that mean it is therefore required (or should have a default value of `device_ctx` maybe?)

cuda_core/cuda/core/_device_resources.pyx
mdboom, line 155: We really should use the enum itself to validate the options (so we don't have the list of acceptable options duplicated all over the place).  I thought it would be as easy as:

```
self.sharing_scope = WorkqueueSharingScopeType(self.sharing_scope)
```

which validates as excepted, but doesn't present the options.

I think we should add a "nice" enum validator and then do a sweep to use it everywhere (but outside of the scope for this PR).

cuda_core/cuda/core/_device_resources.pyx
```diff
@@ -554,6 +563,57 @@ cdef class WorkqueueResource:
...
+        :meth:`configure` with
+        :attr:`WorkqueueResourceOptions.sharing_scope`.
+        """
+        IF CUDA_CORE_BUILD_MAJOR >= 13:
+            from cuda.core.typing import WorkqueueSharingScopeType
+            cdef object scope = self._wq_config_resource.wqConfig.sharingScope
+            if scope == cydriver.CUdevWorkqueueConfigScope.CU_WORKQUEUE_SCOPE_DEVICE_CTX:
+                return WorkqueueSharingScopeType.DEVICE_CTX
+            elif scope == cydriver.CUdevWorkqueueConfigScope.CU_WORKQUEUE_SCOPE_GREEN_CTX_BALANCED:
+                return WorkqueueSharingScopeType.GREEN_CTX_BALANCED
```
mdboom, line 578: nit: Could use a `match` statement here.

cuda_core/docs/source/release/1.1.0-notes.rst
```diff
@@ -106,9 +106,19 @@ New features
...
+  configuring the expected maximum concurrent stream-ordered workloads,
+  and read-only :attr:`WorkqueueResource.sharing_scope`,
+  :attr:`~WorkqueueResource.concurrency_limit`, and
+  :attr:`~WorkqueueResource.device` properties for round-tripping the
+  driver-populated values. Added
+  :class:`~cuda.core.typing.WorkqueueSharingScopeType` StrEnum accepted
+  by :attr:`WorkqueueResourceOptions.sharing_scope` in addition to raw
+  strings.
+  (`#2329 <https://github.com/NVIDIA/cuda-python/pull/2329>`__,
+  `#2330 <https://github.com/NVIDIA/cuda-python/pull/2330>`__)
```
mdboom, line 121: I think this needs to go into cuda-core 1.2.0, unless we really plan to ship this in 1.1.0.

mdboom (changes requested):

cuda_core/cuda/core/_device_resources.pyx
```diff
@@ -138,7 +147,7 @@ cdef class WorkqueueResourceOptions:
         submission remains non-overlapping. (Default to ``None``)
     """
 
-    sharing_scope: str | None = None
+    sharing_scope: WorkqueueSharingScopeType | None = None
```
mdboom:
```suggestion
    sharing_scope: WorkqueueSharingScopeType | str | None = None
```

cuda_core/cuda/core/_device_resources.pyi
leofang: Yeah I think the agent missed it (and my agent, too) most likely because there was no read-only property access (that would surface the need of adding an enum type) until this PR.

cuda_core/cuda/core/_device_resources.pyx
leofang: xref https://github.com/NVIDIA/cuda-python/pull/2307#discussion_r3546833760.

@mdboom @Andy-Jost as per our discussion for #2248 on Tuesday, we decided to **not** widen the type annotation? Do I remember it correctly?

cuda_core/docs/source/release/1.1.0-notes.rst
leofang: The product team (@sri-koundinyan) suggested to include this in v1.1.0 to lay the ground work for a blog post based on green ctx. I discussed with @Andy-Jost and decided to squeeze it in.

leofang: (Unfortunate last minute struggle that I hate, yeah...)

cuda_core/cuda/core/_device_resources.pyx
leofang: This is the getter, not setter. This is coming from the driver and it should always be one of the supported values. 
- If started with device resource, this is the device's default device sharing scope
- If started with an existing `Context`, this is the value that the user set to construct the context instance

cuda_core/cuda/core/_device_resources.pyx
mdboom: #2248 was for option classes, not enums.  And this change is narrowing the type (so would be a breaking API change).

leofang: > #2248 was for option classes, not enums. 

Does it matter, though? This is still applied to the option class constructor.

> And this change is narrowing the type (so would be a breaking API change).

Yes 😞 Alright.

leofang: We really could use griffe to detect this breakage 🙂

cuda_core/cuda/core/_device_resources.pyx
leofang: I feel I am missing something. Maybe this comment makes perfect sense but is left at the wrong line? 😛

cuda_core/cuda/core/_device_resources.pyx
leofang: Good catch, thanks Mike! Done in https://github.com/NVIDIA/cuda-python/pull/2330/commits/be917b62399f573e047038145f8622bf65f79d6e

cuda_core/cuda/core/_device_resources.pyi
```diff
@@ -44,9 +47,9 @@ class WorkqueueResourceOptions:
...
```
mdboom:
```suggestion
    sharing_scope : :class:`~cuda.core.typing.WorkqueueSharingScopeType` | str, optional
```

Sorry, missed this in the first review.

cuda_core/cuda/core/_device_resources.pyx
```diff
@@ -126,9 +131,9 @@ cdef class WorkqueueResourceOptions:
 
     Attributes
     ----------
-    sharing_scope : str, optional
-        Workqueue sharing scope. Accepted values: ``"device_ctx"``
-        or ``"green_ctx_balanced"``. (Default to ``None``)
+    sharing_scope : :class:`~cuda.core.typing.WorkqueueSharingScopeType`, optional
```
mdboom:
```suggestion
    sharing_scope : :class:`~cuda.core.typing.WorkqueueSharingScopeType` | str, optional
```

cuda_core/cuda/core/_device_resources.pyx
mdboom: Oh, somehow it is at the wrong line.  I was just pointing out that the validator here could just use the enum validator (so we don't have to duplicate the list of correct enum values every where that we validate them).  We could almost do that now, but the error message wouldn't be as helpful.  I'll work on that separately -- not worth fixing in this PR.

mdboom (approved):
Approving, but we should make these two changes so we aren't narrowing the types in the docs.

leofang:
Thanks, Mike! Since CI is almost green, what I will do is to wait the CI to finish, and then push the typing only changes to this PR, wait for the pre-commit CI to finish, and then admin merge.

PR #2335, opened 2026-07-09
Build cuda.core RDC test fixtures without Bash

rwgk:
This changes the `cuda.core` RDC test fixture build from a Bash script to a small Python driver around `nvcc`.

Why this change:
- The tests still require `nvcc`, but they no longer require Bash. That avoids Windows-specific ambiguity between Git Bash, WSL Bash, path conversion behavior, and environment propagation.
- The fixture builder now calls `nvcc` directly from Python with `subprocess`, using the platform's normal `PATH` lookup.
- The pathfinder lookup is removed intentionally. For the environments we care about, requiring `nvcc` to be on `PATH` is straightforward, matches normal CUDA command-line expectations, and avoids passing a full `nvcc` path through another layer.
- If local fixture files are absent, tests now distinguish between `nvcc` being unavailable, the host compiler environment being unavailable, and other `nvcc` failures. Only the first two become skips; unexpected `nvcc` failures still fail the test.

Suggested review path:
1. Review `cuda_core/tests/test_binaries/build_test_binaries.py` first. It is the Python replacement for the deleted Bash script.
2. Review `_build_saxpy_rdc()` in `cuda_core/tests/test_module.py` next. That is where the local fallback behavior and skip/fail policy lives.
3. Review the small CI workflow change last; it now runs `nvcc --version` to aid troubleshooting and then the Python fixture builder.


cuda_core/tests/test_module.py
```diff
@@ -193,22 +193,77 @@ def _read_saxpy_rdc(kind: str) -> bytes:
...
-        subprocess.run(  # noqa: S603
-            ["bash", str(binaries_dir / "build_test_binaries.sh")],  # noqa: S607
-            check=True,
-            env=env,
-        )
+        _build_saxpy_rdc(binaries_dir)
     return rdc_path.read_bytes()
 
 
+def _subprocess_output(result: subprocess.CompletedProcess[str]) -> str:
```
lijinf2, line 200: Could we delete this function _subprocess_output and just default to system stdout/stderr behaviors?

cuda_core/tests/test_module.py
```diff
@@ -193,22 +193,77 @@ def _read_saxpy_rdc(kind: str) -> bytes:
...
+        sections.append(f"stderr:\n{result.stderr.rstrip()}")
+    return "\n".join(sections) or "<no output>"
+
+
+def _host_compiler_is_unavailable(output: str) -> bool:
+    normalized = output.lower()
+    # Keep these patterns narrow. Unknown nvcc failures should fail the test and
+    # expose their diagnostics, not be silently reclassified as environment skips.
+    windows_compiler_missing = (
+        "cannot find compiler" in normalized and "cl.exe" in normalized and "in path" in normalized
```
lijinf2, line 214: Is "cannot find compiler" keyword stable enough and will not be modified in future release of nvcc logging? 

I got the following from agent and it works on linux but not sure on windows.  

```python
    for line in p.stdout.splitlines():
        if line.startswith("#$"):
            cmd = line[3:].strip().split()
            if cmd:
                return cmd[0]  # gcc / g++ / clang++ / cl.exe
```

cuda_core/tests/test_module.py
```diff
@@ -193,22 +193,77 @@ def _read_saxpy_rdc(kind: str) -> bytes:
...
+    )
```
lijinf2, line 215: On windows is "cl.exe" the only host compiler available? What if nvcc is using another host compiler?

cuda_core/tests/test_binaries/build_test_binaries.py
```diff
@@ -0,0 +1,56 @@
...
+        nvcc_extra_flags.extend(["-Xcompiler", "/Zc:preprocessor"])
+
+    with tempfile.TemporaryDirectory(prefix="build_test_binaries-", dir=script_dir) as temp_dir:
+        temp_dir_path = Path(temp_dir)
+        object_path = temp_dir_path / final_object_path.name
+        library_path = temp_dir_path / final_library_path.name
+
+        _run(
+            [
+                "nvcc",
```
lijinf2, line 37: Your previous PR uses NVCC environmental variable and pathfinder instead of bare nvcc command. Do we consider that here?

cuda_core/tests/test_module.py
rwgk: I would prefer to keep this helper. We intentionally capture stdout/stderr here because we need to inspect the `nvcc` output to distinguish known host-compiler setup failures from unexpected `nvcc` failures.

For unexpected failures, `pytest.fail(...)` includes the captured stdout/stderr in one self-contained failure message. That makes the CI/test log easier to interpret, especially when pytest output capture or parallel test output would otherwise split the context.

On success we still print the captured output back to stdout/stderr, so the normal diagnostic trail is preserved.

cuda_core/tests/test_module.py
rwgk: Good question. I do not want to rely on this message being stable forever. The intent is almost the opposite: keep the recognized skip cases narrow.

If `nvcc` changes this diagnostic in a future release, this helper will stop recognizing it as a known environment setup issue. The test will then fail and include the full `nvcc` output, which gives us the evidence needed to decide whether to add another narrowly scoped skip pattern.

The `#$ ...` parsing approach can identify commands emitted by `nvcc -v`/dry-run style output, but it does not directly answer the question we need here: did the actual fixture build fail because the host compiler environment is missing, or did it fail for some other reason? For this test, it is safer to classify only the exact known failure modes after the real build attempt, and let everything else fail visibly.

cuda_core/tests/test_module.py
rwgk: For the path this PR exercises, yes, the Windows host compiler we expect is MSVC `cl.exe`. The fixture builder invokes bare `nvcc` and does not pass `--compiler-bindir` / `-ccbin`, so it is intentionally using the normal Windows `nvcc` setup.

If someone has a custom Windows `nvcc` host-compiler setup and it fails with a different diagnostic, this code will not silently skip it. It will fail the test and expose the actual output. That is intentional: the skip classification is only for the known default Windows setup failure where `nvcc` reports that it cannot find `cl.exe` in `PATH`.

In other words, this PR does intentionally not try to generalize all possible Windows host compiler configurations. It makes the supported test contract (see ctk-next PR 436)explicit: `nvcc` on `PATH`, with the normal host compiler environment already configured.

cuda_core/tests/test_binaries/build_test_binaries.py
rwgk: Yes, this was considered. Now that the builder is Python instead of Bash, a full `nvcc.exe` path would be much less problematic than it was in the earlier Bash-based version.

I still prefer bare `nvcc` here because it keeps the test contract simple and explicit: the test environment should put the intended CUDA toolkit's `bin` directory on `PATH`, and the host compiler environment should already be configured. That is easy to satisfy in CI, runbooks, and local development.

Using pathfinder or an `NVCC` override inside this unit test would add a second toolkit-selection mechanism. That can create mismatches where the test discovers one `nvcc`, while `CUDA_PATH`, `PATH`, runtime libraries, or the surrounding runbook point at another toolkit. For this fixture build, relying on the environment's normal executable lookup is simpler and easier to reason about.

lijinf2 (approved):
Thanks for clarifying. I’m okay with keeping the current approach: bare nvcc from PATH, narrow skip classification for known host compilers, and fail visibly for unknown nvcc failures and log message updates.

One minor suggestion: please consider adding a short code comment/doc note that this test intentionally assumes nvcc is on PATH and a working host compiler environment is configured.

rwgk:
Thanks @lijinf2!

> One minor suggestion: please consider adding a short code comment/doc note that this test intentionally assumes nvcc is on PATH and a working host compiler environment is configured.

Done: commit aea8bb561468e74754513d769b62f53d13fa41a9

leofang:
Thanks guys!

Issue #2179, opened 2026-06-08
Fix the handling of the `CUcheckpointRestoreArgs_st` struct


leofang:
This seems very complicated. I don't think we ever have code that does conditional logic for struct members? Is it possible to keep it simple, to make it easier to reason about what code would be eventually used to generate a binary?
1. cython-gen does not use megaheader, so whether or not we expose the complete member list in a `cdef extern from` block does not matter, the C compiler will do the right job
2. the tempita + pyclibrary template approach will be killed soon-ish

_Originally posted by @leofang in https://github.com/NVIDIA/cuda-python/pull/2144#discussion_r3325433666_

lijinf2:
Closed because target file `cuda_bindings/cuda/bindings/cydriver.pxd.in` has been deleted and has become `cuda_bindings/cuda/bindings/cydriver.pxd`.

PR #2339, opened 2026-07-10
Add ObjectCode.get_module() for legacy driver API integration

### LINKED ISSUE #489: Exposing `cuLibraryGetModule` in `cuda.core`
For legacy library, it takes time to migrate from module dependent APIs to library dependent APIs. In the transition time, `cuLibraryGetModule` is a very useful API to allow switching from context independent module loading to old style module management, providing backward compatibility. Since `cuda.core` operates with context independent module loading, ti would be helpful to expose this API in `cuda.core` to support better integration with legacy libraries.

lijinf2:
## Description

closes #489 

Add `ObjectCode.get_module()` to expose the `CUmodule` for the current context via `cuLibraryGetModule`, so code using legacy driver APIs can interoperate with `cuda.core`'s context-independent `CUlibrary` loading.

## Checklist

### DIFF
--- a/cuda_core/cuda/core/_module.pyi
+++ b/cuda_core/cuda/core/_module.pyi
@@ -458,6 +458,16 @@ class ObjectCode:
 
         """
 
+    def get_module(self) -> object:
+        """Return the :obj:`~driver.CUmodule` associated with this object code.
+
+        Returns
+        -------
+        :obj:`~driver.CUmodule`
+            Module handle for the current CUDA context, suitable for legacy
+            driver APIs that accept ``CUmodule``.
+        """
+
     @property
     def code(self) -> CodeTypeT:
         """Return the underlying code object."""
--- a/cuda_core/cuda/core/_module.pyi
+++ b/cuda_core/cuda/core/_module.pyi
@@ -459,7 +459,11 @@ class ObjectCode:
         """
 
     def get_module(self) -> object:
-        """Return the :obj:`~driver.CUmodule` associated with this object code.
+        """Return a context-dependent :obj:`~driver.CUmodule` for legacy interop.
+
+        Bridges the native :obj:`~driver.CUlibrary` (see :attr:`handle`) to a
+        ``CUmodule`` via ``cuLibraryGetModule``, for use with legacy driver APIs
+        that only accept ``CUmodule``.
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


cuda_core/cuda/core/_module.pyx
```diff
@@ -796,6 +797,21 @@ cdef class ObjectCode:
...
+        -------
+        :obj:`~driver.CUmodule`
+            Module handle for the current CUDA context, suitable for legacy
+            driver APIs that accept ``CUmodule``.
+        """
+        self._lazy_load_module()
+        cdef cydriver.CUmodule mod
+        with nogil:
+            HANDLE_RETURN(cydriver.cuLibraryGetModule(&mod, as_cu(self._h_library)))
+        return driver.CUmodule(<intptr_t>mod)
```
leofang, line 817: Q: Shouldn't we use `as_py()` to do the return value conversion?

Q: How do we teach the difference between `.get_module()` and `.handle`?

lijinf2: Sorry for not getting back sooner. I shifted focus to issue #2179, for which I need to set up cuda bindings development environment for the first time. 

It looks `as_py()` does not provide support for CUmodule type yet. I am not sure if we should add the support there per CUmodule is deemed as legacy API. But let me know if you find the support favorable. Happy to add that. 

For .handle vs .get_module(): I'll clarify in the docstring.

lijinf2: Done

isVoid (approved):
I think we can create a new `as_py` function in `_resource_handle.hpp` that accpets cumodule type. But this can be set as a follow up issue.

cuda_core/cuda/core/_module.pyi
```diff
@@ -458,6 +458,20 @@ class ObjectCode:
 
         """
 
+    def get_module(self) -> object:
```
juenglin, line 461: Did you purposefully chose `object` as return type instead of `CUModule`? If so, please help me understand why.

isVoid: I would just assume the agent follows the return typing in `.handle`. Which is also `object`.

lijinf2: Yes, agent follows .handle return type. 

I will try CUModule type and see what happens in the followup issue https://github.com/NVIDIA/cuda-python/issues/2415

lijinf2: CUModule is more accurate here. Thank you for pointing this out! @juenglin

PR #2347, opened 2026-07-12
test(core): add cuda.core.__all__ vs public docs consistency check

### LINKED ISSUE #2326: [FEA]: Add a consistency check for cuda.core.__all__ vs cuda_core/docs/source/api.rst
xref: https://github.com/NVIDIA/cuda-python/pull/2300#pullrequestreview-4622238712

Copy-pasting the relevant part for easy referrence:

Add a follow-on docs/API consistency test that compares the public symbols in `cuda.core.__all__` against the public entries in `cuda_core/docs/source/api.rst`, with deliberate exceptions for submodule namespaces and returned-helper/private API docs.

It should parse the Sphinx `autosummary` and `data` entries, then account for known surfaces such as `cuda.core.graph`, `cuda.core.texture`, `cuda.core.utils`, `cuda.core.checkpoint`, and entries intentionally documented in `api_private.rst`. The useful failure mode is: “this symbol is public by `__all__` but missing from public docs” or “this symbol is documented as public but not exported by `__all__`.”

aryanputta:
## Description

Closes #2326.

Adds `cuda_core/tests/test_api_docs_consistency.py`, the follow-on consistency check requested in the #2300 review.

### What it checks

- `test_all_exports_resolve`: every name in `cuda.core.__all__` resolves via `hasattr` (the stale-name guard suggested in the review).
- `test_public_symbols_are_documented`: fails with "public by `cuda.core.__all__` but missing from public docs (api.rst)" plus the offending names.
- `test_documented_symbols_are_exported`: fails with "documented as public in api.rst but not exported by `cuda.core.__all__`" plus the offending names.

### How the docs are parsed

- `.. module::` and `.. currentmodule::` directives are tracked, so only `autosummary` entries and `.. data::` directives active under `cuda.core` count. Entries under `cuda.core.utils` or `cuda.core.texture` are excluded automatically.
- Dotted entries such as `graph.Graph` or `checkpoint.Process` are submodule namespaces, not flat `cuda.core` names, and are excluded.
- Symbols documented in `api_private.rst` are accepted as documented by their trailing name, so returned helpers such as `_device_resources.DeviceResources` do not fail the public-docs direction.

### Ordering with #2300

The tests skip when `cuda.core.__all__` is not defined, so this is safe to land before #2300 and activates once #2300 merges. The docs fixture also skips when docs sources are unavailable (installed-wheel test runs).

### Validation

I do not have a CUDA-capable machine available, so I validated with a stubbed `cuda.core` module against the real RST files in this repo:

- With `__all__` set to the list from #2300: all 3 tests pass (the current `api.rst`/`api_private.rst` and the #2300 `__all__` are exactly consistent).
- Without `__all__` (current main): all 3 tests skip.
- Negative tests: removing a name from the stub `__all__` and adding a bogus name each fail the corresponding direction with the intended message.

`ruff check` and `ruff format` pass.


cuda_core/tests/test_api_docs_consistency.py
```diff
@@ -0,0 +1,118 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""Consistency check between ``cuda.core.__all__`` and the public API docs.
+
+Compares the public symbols exported by ``cuda.core.__all__`` against the
+public entries in ``docs/source/api.rst``. Symbols documented in
+``docs/source/api_private.rst`` (returned helpers users cannot instantiate)
+are accepted as documented. Dotted entries such as ``graph.Graph`` or
+``checkpoint.Process`` describe submodule namespaces, not the flat
+``cuda.core`` namespace, and are excluded from the comparison.
```
mdboom: These should be included (as well as `cuda.core.system`).

Any `__init__.py` inside of a public (non `_`-prefixed) submodule is public API.  Currently, `graph`, `system`, `texture` and `utils`, though more may be added in the future so the test should adapt automatically.

Today, only `system` has its own separate API `.rst` page (which is fine).  I would argue the other submodules should have their own `.rst` page as well.  (And we could do that refactor here).

mdboom (changes requested):
The main change required is that the scope of this is too limited.

It would also be worthwhile verifying that this tests what we think it does by (in another PR) adding some changes that this is expected to catch to confirm that they are in fact caught.

I also think this should land the `__all__` into cuda.core here (again, to confirm it's working) rather than waiting for #2300 to add it.

aryanputta:
> The main change required is that the scope of this is too limited.
> 
> 
> 
> It would also be worthwhile verifying that this tests what we think it does by (in another PR) adding some changes that this is expected to catch to confirm that they are in fact caught.
> 
> 
> 
> I also think this should land the `__all__` into cuda.core here (again, to confirm it's working) rather than waiting for #2300 to add it.

Can you assign this to me, but will work on this today.

aryanputta:
Hello @mdboom, thank you for the review. I have addressed both points.

- Landed `__all__` into `cuda.core` in this PR so the check runs against a real export surface instead of skipping. I also added an aggregated `__all__` to `cuda.core.graph` (built from its star-imported submodules, so it cannot drift); `system`, `texture`, and `utils` already defined one.
- Widened the scope: the test now auto-discovers public subpackages from `cuda.core.__path__` (the internal `cuNN` wheel shims are excluded), asserts each defines a fully resolvable `__all__`, and cross-checks each documented subpackage against `api.rst`, handling both the dotted (`graph.Graph`) and flat (`currentmodule`) doc conventions. `system` is documented in `api_nvml.rst` and has a runtime-conditional `__all__`, so its doc cross-check is skipped while it is still verified for a well-formed `__all__`.

On confirming the check actually catches regressions: I prepared a separate branch that injects one deliberate discrepancy (a symbol documented in `api.rst` that nothing exports). `test_documented_symbols_are_exported` fails on it as expected while the rest of the suite stays green: https://github.com/aryanputta/cuda-python/tree/test/api-docs-consistency-demo-2326 . I can open that as a follow-up PR once this one lands, or fold a permanent negative test in here if you prefer. Please let me know which you would like.

cuda_core/tests/test_api_docs_consistency.py
```diff
@@ -0,0 +1,186 @@
...
+            module = argument
+            continue
+        body = []
+        while i < len(lines):
+            body_line = lines[i]
+            if body_line.strip() and not body_line.startswith(" "):
+                break
+            body.append(body_line)
+            i += 1
+        yield module, directive, argument, body
```
mdboom: This sort of string parsing can become very brittle to slightly different uses.

I think it would be worth experimenting with using the `docutils` rst parser directly here to get this information.  If it becomes unwieldy we can fall back to this, but I think it's worth a shot.

For example, my agent found these bugs with this.  Using the upstream `docutils` parser would work identically to how the docs are built so should be resistant to those kinds of things:

====

_iter_directive_blocks swallows the body of .. module:: / .. currentmodule:: directives silently. When those directives are encountered the function does continue and skips the body scan. If the RST has :synopsis: or other options under .. module:: cuda.core, the outer loop's next iteration sees those indented lines as ordinary text (they don't start with .. ) and skips them — which is the right behaviour here. However, if someone puts a body directive immediately below .. module:: (no blank line separator), the first non-blank indented line will be silently dropped from the outer loop. Not a current issue but fragile.

cuda_core/tests/test_api_docs_consistency.py
```diff
@@ -0,0 +1,186 @@
...
+    assert hasattr(module, "__all__"), f"cuda.core.{sub} does not define __all__"
+    missing = [name for name in module.__all__ if not hasattr(module, name)]
+    assert missing == [], f"cuda.core.{sub}.__all__ lists names that do not resolve: {missing}"
+
+
+@pytest.mark.parametrize("sub", PUBLIC_SUBPACKAGES)
+def test_subpackage_exports_match_docs(sub, docs_dir):
+    documented = _documented_subpackage_names(docs_dir / "api.rst", sub)
+    if not documented:
+        pytest.skip(f"cuda.core.{sub} is not documented in api.rst (its reference lives elsewhere)")
```
mdboom: Unless I am mistaken, this will cause a subpackage that is documented in another file (such as `cuda.core.system` is today) to not be tested.  Is my understanding correct?

We need to make sure that for every exported public symbol, it is in the documentation /somewhere/.  I think better logic would be to collect all of the documented symbols in /all/ of the doc files, and then iterate over all of the publicly exposed symbols and make sure they appear in those doc files.

Treating `api.rst` as special doesn't really make sense.

mdboom (changes requested):

mdboom:
@aryanputta: Thanks a lot for working on the `docutils`-based parser here.  I think that will be much more robust.

In the interest of expediency, I pushed some commits to this PR.

- Added the `docutils` dependency to `pyproject.toml`

- Updated the cuda_core code itself to fix issues found by the new tests

The biggest change, however, was removing the check that everything that is documented is public.  That added a lot of complexity (since it needed to handle `api-private.rst` specially).  I also think it's not something that we want to enforce.  It is perfectly reasonably to document something private from any page (think, for example, a document about cuda-core internals).  All we really want to enforce is that every public API is included in the docs /somewhere/.  In practice, that's usually one of the `api-*.rst` pages, but there's no reason it has to remain that way for all time.  This significantly simplifies the tests and removes a lot of special-casing that was there.

cuda_core/tests/test_api_docs_consistency.py
```diff
@@ -0,0 +1,223 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""Consistency checks between the public ``__all__`` surface and the API docs.
+
+Covers the flat ``cuda.core`` namespace and every public subpackage
+(``graph``, ``system``, ``texture``, ``utils``, and any added later) discovered
+automatically from ``cuda.core.__path__``. For each public namespace, exported
+``__all__`` names must appear somewhere in ``cuda_core/docs/source``.
+"""
+
```
rwgk, line 29: Document what we're intentionally not checking for here.

aryanputta: I expanded the module docstring to spell out the intentional scope limits: the check is one-way (public export -> documented) and does not enforce the reverse, since documenting a private or internal symbol on any page is allowed. It is also a name-presence check only, so it does not verify signatures, docstrings, or rendered output, treats `:no-index:`/deprecated entries as documented, and does not descend into class members or non-top-level docs pages. Pushed in 39f9843.

Andy-Jost:
Two findings:

  • Medium — `cuda_core/tests/test_api_docs_consistency.py:49` filters on `ispkg`, so public modules such as `cuda.core.checkpoint` and `cuda.core.typing` are never checked. This notably misses `checkpoint.Process`, despite issue #2326 explicitly naming checkpoint.
  • Medium — aggregating `_memory_all` into top-level `cuda.core.__all__` promotes `IPCAllocationHandle` and `IPCBufferDescriptor` to public API. They were absent from the curated #2300 list and previously omitted from top-level imports.

mdboom:
> Two findings:
> 
> • Medium — `cuda_core/tests/test_api_docs_consistency.py:49` filters on `ispkg`, so public modules such as `cuda.core.checkpoint` and `cuda.core.typing` are never checked. This notably misses `checkpoint.Process`, despite issue #2326 explicitly naming checkpoint. • Medium — aggregating `_memory_all` into top-level `cuda.core.__all__` promotes `IPCAllocationHandle` and `IPCBufferDescriptor` to public API. They were absent from the curated #2300 list and previously omitted from top-level imports.

I've addressed these and pushed a new commit.

cuda_core/cuda/core/__init__.py
```diff
@@ -69,45 +69,51 @@ class _PatchedProperty(metaclass=_PatchedPropMeta):
...
+    + _graphics_all
+    + _host_all
+    + _launch_config_all
+    + _launcher_all
+    + _linker_all
+    + _memory_all
+    + _module_all
+    + _program_all
+    + _stream_all
+    + _tensor_map_all
```
Andy-Jost: nit: use `__all__ = [ *a, *b, ... ]` rather than repeated addition on lists.
nit: `__all__` is usually a list.

Suggested:

```
from cuda.core._context import *
from cuda.core._context import __all__ as _context_all
from cuda.core._device import *
from cuda.core._device import __all__ as _device_all
__all__ = [
    *_context_all,
    *_device_all,
]
del _context_all, _device_all
```

aryanputta: Sorry question - what does nit mean? Do you guys want me to work on this or would you guys rather work and finish this off internally, so commits don't get messy? @mdboom @Andy-Jost

Andy-Jost: "nit" means small and nice-to-fix but not important enough to block the merge. Sometimes it can also be used to express a small preference that's not worth arguing over if there's any disagreement.

In this case, since CI hasn't been run it's probably worth fixing.

aryanputta: Okay thank you, should I fix this or let micheal finish this off and have him just resolve this issue?

Andy-Jost: It looks like Mike already addressed this

aryanputta:
Hello @mdboom, the test matrix on `565479b` was red because of one real failure, not the guard itself.

`tests/test_memory.py::test_package_contents` still listed `IPCAllocationHandle` and `IPCBufferDescriptor` in its `expected` set. Setting `__all__ = []` in `_memory/_ipc.pyx` (the fix for Andy-Jost's second finding) means `from cuda.core._memory import *` no longer binds them, so the assertion compared 15 names against the 13 that are actually exported. I have dropped the two entries in `f768d40`. Everything else in that run passed (1 failed, 3476 passed).

One job, `Test linux-64 / Python 3.11, CUDA 13.3.0 (local), GPU l4`, failed earlier than the tests with `Unable to download artifact(s): No server is currently available`. That is a GitHub artifact service outage and unrelated.

Current head is `f768d4052b`, so a fresh `/ok to test f768d40` is needed when you get a chance.

Andy-Jost (approved):

mdboom (approved):

PR #2348, opened 2026-07-13
Generator changes

mdboom:
This brings in a few recent changes in the generator:

- Generating platform-specific files, rather than using templates
- Dropping the gil for all API calls (now for functions returning `char *`)
- Not including version information in `fast_enum.py`


rwgk (approved):
I'll go ahead and merge this.

It'd be great if you could add the MR numbers to the bullet points in the PR description. That'd make it much easier for us to make the connections back in the future, and will also make it easier to review update PRs like this one.

PR #2355, opened 2026-07-14
Revert nvJitLink driver-major compatibility fallback

### LINKED ISSUE #2351: cuda.core Linker rejects LTO-IR (regression from #2320): falls back to cuLink* when CUDA driver major > nvJitLink major
### Summary
#2320 ("Check nvJitLink driver major compatibility") added a gate to `_decide_nvjitlink_or_driver()` that drops the `Linker` from nvJitLink to the driver `cuLink*` APIs whenever `driver_major > nvjitlink_version[0]`. `cuLink*` cannot link LTO-IR, so any LTO-IR link in that configuration now fails with:

```
ValueError: Unknown code_type associated with ObjectCode: ltoir
```

### Root cause
Gate added by #2320:
https://github.com/NVIDIA/cuda-python/blob/c000331de6c37aa4565af74b001271ffcf6d5c99/cuda_core/cuda/core/_linker.pyx#L698-L708
`cuLink*` "do not support ... linking LTO IRs":
https://github.com/NVIDIA/cuda-python/blob/c000331de6c37aa4565af74b001271ffcf6d5c99/cuda_core/cuda/core/_linker.pyx#L683-L687
Raised here:
https://github.com/NVIDIA/cuda-python/blob/c000331de6c37aa4565af74b001271ffcf6d5c99/cuda_core/cuda/core/_linker.pyx#L556-L558

Before #2320, nvJitLink was used whenever present and `>=12.3`, so nvJitLink 12.x linked LTO-IR fine even under a newer driver.

### Impact
Dominant cause of the current `CI: Nightly optional-deps` failures on `main`. Any `cuda.core.Linker` user with LTO-IR inputs where installed nvJitLink major < CUDA driver major is hit — e.g. cu12 wheels (nvJitLink 12.9) on a 13.x driver:

- `nightly-numba-cuda` cu12.9.1 (all 3 platforms): 2043&times; `Unknown code_type ... ltoir` of 2045 errors. numba-cuda links via cuda.core: `numba_cuda/.../cudadrv/driver.py:2473` &rarr; `cuda/core/_linker.pyx:{82,536,558}`. [Failing job log](https://github.com/NVIDIA/cuda-python/actions/runs/29221596287/job/86727688621)
- `nightly-numba-cuda-mlir` cu12.9.1: +90&times; the same error.
- cu13 rows (nvJitLink 13 + driver 13) are unaffected.

Env: cuda-core `1.1.1.dev` (main), `nvidia-nvjitlink-cu12 12.9.86`, driver 13.x.

### Releases
Not in `cuda-core v1.1.0` (tagged ~80 min before #2320 merged; its `_linker.pyx` has none of the gate). Main-only; would ship in the next release unless fixed first.

### Fix
When any input is LTO-IR, the linker must **not** fall back to the driver `cuLink*` APIs under any circumstances — they cannot link LTO-IR. `_decide_nvjitlink_or_driver()` must not force the driver backend when LTO-IR inputs are present: use nvJitLink regardless, and if nvJitLink is genuinely unusable, raise a clear error instead of silently degrading to `cuLink*` and failing with `Unknown code_type ... ltoir`.

cc @isVoid (#2320 author), @lijinf2 (reviewer)

-- Leo's bot

isVoid:
## Summary

- Revert the linker backend-selection portion of #2320.
- Stop falling back to cuLink solely because the CUDA driver major version is newer than the available nvJitLink major version.
- Retain #2320's cuLink option-array and log-buffer lifetime fix, including its regression tests.

## Rationale

The backend-selection policy and the cuLink storage-lifetime fix are independent. This change rolls back the compatibility fallback while preserving the memory-safety fix that keeps storage referenced by `CUlinkState` alive until `cuLinkDestroy`.

## User impact

`cuda.core` will use nvJitLink whenever it is available and exposes the required version symbol, even when the CUDA driver has a newer major version. When the cuLink backend is used, its option arrays and log buffers remain valid for the full lifetime of the link state.

## Testing

- `pixi run --as-is -e cu13 pytest tests/test_linker.py tests/test_optional_dependency_imports.py tests/test_program_cache.py -q` (279 passed, 3 skipped)
- Simulated cuLink selection for the two affected test groups (4 passed, 3 skipped with `-time is not supported in cuLink`)
- `git diff --check upstream/main...HEAD`

Partially reverts NVIDIA/cuda-python#2320.

### DIFF
--- a/cuda_core/tests/test_linker.py
+++ b/cuda_core/tests/test_linker.py
@@ -8,6 +8,7 @@ import pytest
 
 from cuda.core import Device, Linker, LinkerOptions, Program, ProgramOptions, _linker
 from cuda.core._module import ObjectCode
+from cuda.core._program import _can_load_generated_ptx
 from cuda.core._utils.cuda_utils import CUDAError
 
 ARCH = "sm_" + "".join(f"{i}" for i in Device().compute_capability)
@@ -232,6 +233,47 @@ def test_linker_handle(compile_ptx_functions):
     handle = linker.handle
     assert handle is not None
     assert int(handle) != 0
+
+
+@pytest.mark.agent_authored(model="gpt-5")
+@pytest.mark.skipif(not is_culink_backend, reason="driver backend regression test")
+def test_driver_linker_lifetime_no_heap_corruption(compile_ptx_functions):
+    if not _can_load_generated_ptx():
+        pytest.skip("PTX version too new for current driver")
+
+    linker = Linker(*compile_ptx_functions, options=LinkerOptions(arch=ARCH))
+    linker.link("cubin")
+    linker.close()
+    del linker
+
+    obj_a = Program(kernel_a, "c++", ProgramOptions(relocatable_device_code=True)).compile("ptx")
+    obj_b = Program(device_function_b, "c++", ProgramOptions(relocatable_device_code=True)).compile("ptx")
+    obj_c = Program(device_function_c, "c++", ProgramOptions(relocatable_device_code=True)).compile("ptx")
+    linker = Linker(obj_a, obj_b, obj_c, options=LinkerOptions(arch=ARCH))
+    linker.link("cubin")
+    linker.close()
+
+
+@pytest.mark.agent_authored(model="gpt-5")
+@pytest.mark.skipif(not is_culink_backend, reason="driver backend regression test")
+def test_driver_linker_preserves_error_log_after_close(init_cuda):
+    if not _can_load_generated_ptx():
+        pytest.skip("PTX version too new for current driver")
+
+    bad_kernel = """
+extern __device__ int Z();
+__global__ void A() { int r = Z(); }
+"""
+    bad_obj = Program(bad_kernel, "c++", ProgramOptions(relocatable_device_code=True)).compile("ptx")
+    linker = Linker(bad_obj, options=LinkerOptions(arch=ARCH))
+    with pytest.raises(CUDAError):
+        linker.link("cubin")
+
+    error_log = linker.get_error_log()
+    assert error_log
+    linker.close()
+    assert linker.get_error_log() == error_log
+    assert isinstance(linker.get_info_log(), str)
 
 
 @pytest.mark.skipif(is_culink_backend, reason="nvjitlink options only tested with nvjitlink backend")
--- a/cuda_core/tests/test_linker.py
+++ b/cuda_core/tests/test_linker.py
@@ -8,7 +8,6 @@ import pytest
 
 from cuda.core import Device, Linker, LinkerOptions, Program, ProgramOptions, _linker
 from cuda.core._module import ObjectCode
-from cuda.core._program import _can_load_generated_ptx
 from cuda.core._utils.cuda_utils import CUDAError
 
 ARCH = "sm_" + "".join(f"{i}" for i in Device().compute_capability)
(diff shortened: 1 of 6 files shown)


cuda_core/tests/test_linker.py
```diff
@@ -197,12 +197,19 @@ def test_linker_options_as_bytes_nvjitlink():
...
-        options.as_bytes(backend)
+        options.as_bytes("invalid")
+
+
+@pytest.mark.skipif(not is_culink_backend, reason="driver backend test")
+def test_linker_options_as_bytes_driver_not_supported():
+    """Test that as_bytes() is not supported for driver backend"""
+    options = LinkerOptions(arch="sm_80")
+    with pytest.raises(RuntimeError, match="as_bytes\\(\\) only supports 'nvjitlink' backend"):
+        options.as_bytes("driver")
```
leofang: nit: this resurrects a pre-#2320 test bug: `as_bytes("driver")` raises `ValueError` (the backend-string check fires before the availability check), so `pytest.raises(RuntimeError)` won't catch it — the test fails whenever it isn't skipped:

https://github.com/NVIDIA/cuda-python/blob/f835032f979f7de13f576219b0cdd706a20a4466/cuda_core/cuda/core/_linker.pyx#L465-L469

#2320's parametrization was correct on both backends:

```suggestion
@pytest.mark.parametrize("backend", ("invalid", "driver"))
def test_linker_options_as_bytes_invalid_backend(backend):
    """Test LinkerOptions.as_bytes() with invalid backend"""
    options = LinkerOptions(arch="sm_80")
    with pytest.raises(ValueError, match="only supports 'nvjitlink' backend"):
        options.as_bytes(backend)
```

cuda_core/tests/test_program_cache.py
```diff
@@ -375,6 +371,9 @@ def test_make_program_cache_key_ignores_name_expressions_for_non_nvrtc(code_type
             {"link_time_optimization": None},
             id="lto_false_eq_none",
         ),
+        # ``time`` is a presence gate: the linker emits ``-time`` for any
+        # non-None value, so True / "path" produce the same flag.
+        pytest.param({"time": True}, {"time": "timing.csv"}, id="time_true_eq_path"),
```
leofang: nit: on a cuLink-backend host, `_LinkerBackend.validate()` rejects non-None `time`, so this case fails there:

https://github.com/NVIDIA/cuda-python/blob/f835032f979f7de13f576219b0cdd706a20a4466/cuda_core/cuda/core/utils/_program_cache/_keys.py#L509-L522

The monkeypatch in this test only pins the version probe, not the backend decision. Pinning the decision too (same idiom as `test_make_program_cache_key_ptx_driver_ignored_fields_collapse`) keeps the case portable:

```python
    from cuda.core import _linker

    monkeypatch.setattr(_linker, "_decide_nvjitlink_or_driver", lambda: False)  # nvJitLink
```

cuda_core/tests/test_program_cache.py
```diff
@@ -1014,9 +1001,9 @@ def test_make_program_cache_key_rejects_side_effect_options_nvrtc(option_kw, ext
         pytest.param({"time": "whatever.csv"}, id="time_path"),
     ],
 )
-@pytest.mark.skipif(is_culink_backend, reason="test requires nvJitLink backend")
 def test_make_program_cache_key_accepts_side_effect_options_for_ptx(option_kw):
-    """nvJitLink ``-time`` writes only to its info log, not the filesystem."""
+    """The side-effect guard is NVRTC-specific: PTX (linker) and NVVM must
+    not be blocked by options whose side effects only apply under NVRTC."""
     _make_key(code=".version 7.0", code_type="ptx", options=_opts(**option_kw))  # no raise
```
leofang, line 1017: nit: same as the comment above — without the skipif, `time_true`/`time_path` hit the driver-unsupported-options rejection on cuLink hosts. Pinning the backend avoids both the failure and the skip:

```suggestion
def test_make_program_cache_key_accepts_side_effect_options_for_ptx(option_kw, monkeypatch):
    """The side-effect guard is NVRTC-specific: PTX (linker) and NVVM must
    not be blocked by options whose side effects only apply under NVRTC."""
    from cuda.core import _linker

    monkeypatch.setattr(_linker, "_decide_nvjitlink_or_driver", lambda: False)  # nvJitLink
    _make_key(code=".version 7.0", code_type="ptx", options=_opts(**option_kw))  # no raise
```

leofang (approved):
LGTM. The inline comments below are nitpicks about tests for the cuLink backend path, which does not appear to have CI coverage today — feel free to defer.

-- Leo's bot

leofang (approved):

cuda_core/tests/test_program_cache.py
isVoid: I kinda feel like instead of explicitly pinning, we should mark the test with `time` argument with `skipif` when the backend is driver. Even though currently we don't have a setup in the CI matrix that touches this test, in the future if we augment the matrix and this path hits, we can at least see the skip reason in the test, which means the un-reverted change is better.

Issue #2352, opened 2026-07-13
cuda.core: add atomic update methods for graph definition nodes


Andy-Jost:
## Background

Part of #1330. Depends on #2350.

`GraphNode` subclasses currently expose node parameters as read-only properties. The original plan called for property setters, but CUDA updates most node types by replacing one complete parameter structure. Several parameters and their ownership requirements are also coupled—for example, a host callback function with its `userData`, or a kernel with its arguments.

Add atomic, subclass-specific `update()` methods rather than independent property setters.

This feature requires CUDA 12.2+ and uses the generic `cuGraphNodeSetParams` API. Supporting CUDA 12.0/12.1 through the older typed setters is intentionally out of scope.

## Proposed API

Properties remain read-only for inspection. Mutation uses one method call and one CUDA parameter update.

```python
kernel.update(config=new_config)
kernel.update(args=new_args)
kernel.update(kernel=new_kernel, args=new_args)
kernel.update(config=new_config, kernel=new_kernel, args=new_args)

memcpy.update(dst=new_dst, src=new_src, size=new_size)
memset.update(dst=new_dst, value=new_value, width=new_width)

host.update(new_callback, user_data=new_user_data)
child.update(new_child_graph)
event_record.update(new_event)
event_wait.update(new_event)
```

Omitted keyword arguments preserve their current values where doing so is coherent.

Kernel-specific rules:

- Arguments may be updated while retaining the existing kernel.
- Changing the kernel requires explicitly supplying its arguments, including `args=()` for a no-argument kernel.
- Configuration, kernel, and arguments may be replaced atomically in one call.
- Cluster dimensions and cooperative-launch state remain outside this API unless they can be represented correctly by the supported graph-node APIs.

Memcpy/memset rules:

- Multiple fields may be replaced atomically.
- `Buffer` operands retain their allocation handles.
- Raw integer addresses remain caller-owned.
- Explicit `dst_owner` and `src_owner` are accepted only alongside their corresponding raw-address operands.
- Supplying a new raw address without an owner removes that slot's ownership from the updated definition.

Host callbacks always replace the callback and user-data binding together; updating only one is not supported.

## Implementation

Each method combines supplied values with the node's current state to construct one fully initialized, zero-padded `CUgraphNodeParams`. It sets the appropriate node-type member and invokes `cuGraphNodeSetParams` exactly once through shared update-and-rollback plumbing.

When ownership changes:

[...]

PR #2357, opened 2026-07-14
cuda.core: retain graph attachments with per-node user objects

### LINKED ISSUE #2350: cuda.core: retain graph node attachments as per-node CUDA user objects
## Background

Part of #1330.

CUDA graph user objects let applications associate externally owned resources with a `CUgraph`. CUDA retains those user-object references when a graph is cloned or instantiated, allowing the resources to outlive the source graph objects that first referenced them.

cuda-core currently collects every graph-node dependency—such as kernels, argument buffers, events, callbacks, and memory operands—in one mutable `GraphSlotTable`. It attaches that table to the graph as a single CUDA user object.

Because clones and executable graphs retain the same user object, they all share the same mutable table payload with the source graph definition. Updating or deleting a source node can replace or remove a table entry and release a resource that an existing `CUgraphExec` still references, corrupting the already-created executable.

The proposed design moves the CUDA user-object attachment point from the mutable table to the table’s individual node entries. Each entry has its own immutable `NodeAttachments` bundle and user object, so a clone or executable instance retains the user objects for the entries it inherited even if the source definition later changes or removes those entries from its table.

## Design

Keep a node attachment table as metadata owned by `GraphBox`, but do not retain the table itself as a CUDA user object. Instead, each resource-bearing node has an immutable `NodeAttachments` bundle retained through its own `CUuserObject`.

Conceptually:

```text
GraphBox
  CUgraph
  map<CUgraphNode, AttachmentRecord>

AttachmentRecord
  CUuserObject
  NodeAttachments*  # non-owning; lifetime transfers from CUuserObject to the cleanup queue

NodeAttachments
  collection of OpaqueHandle owners
```

Examples include:

- Kernel node: kernel handle and kernel-argument holder.
- Host node: callback owner and copied user-data owner.
- Memcpy node: source and destination owners.
- Memset node: destination owner.
- Event node: event owner.

CUDA automatically propagates these user objects when graphs are cloned, instantiated, or used for whole-graph updates. Releasing a definition's reference therefore does not invalidate clones, executable graphs, or in-flight launches that still require the bundle.

All wrappers for the same underlying `CUgraph`, including child and conditional graph views, must share the same attachment metadata rather than creating independent tables.

## User-object destructor safety

CUDA may invoke a user-object destructor asynchronously on a shared internal thread. The destructor must not call CUDA APIs and should not block.

The destructor must not directly delete `NodeAttachments`. Its `OpaqueHandle` entries are type-erased owning references, and releasing the final reference may invoke CUDA-backed deleters such as `cuMemFree`, `cuMemFreeAsync`, `cuEventDestroy`, or `cuLibraryUnload`. Releasing a Python-backed owner may also execute arbitrary finalizer code.

### Deferred cleanup mechanism

The CUDA callback transfers each intact `NodeAttachments` payload to a process-lifetime deferred-cleanup queue. Queue insertion must not allocate, release an `OpaqueHandle`, or invoke CUDA. The queue state must remain valid for as long as CUDA may invoke an outstanding user-object destructor.

One possible wakeup mechanism is `Py_AddPendingCall`. A proposed implementation is:

1. Intrusively push the payload onto a lock-free MPSC queue.
2. Use an atomic flag to ensure that at most one drain call is scheduled.
3. Call `Py_AddPendingCall` to schedule a C drain callback.
4. The drain callback atomically detaches the queued payloads and deletes them from a normal Python thread, where CUDA-backed and Python-backed deleters may run.
5. Clear the scheduled flag and reschedule if payloads arrived concurrently with the drain.
6. If scheduling fails, leave the payloads queued and retry from a later CUDA callback or safe cuda-core entry point. Never retry in a blocking loop or reclaim on CUDA's callback thread.

`Py_AddPendingCall` avoids introducing and shutting down a dedicated worker thread, but it has important limitations:

- Its main-thread queue is bounded (32 entries in CPython 3.14), so scheduling one pending call per payload is unsafe. cuda-core must coalesce cleanup behind one pending call.
- It returns `-1` when the queue is full without setting an exception.
- CPython currently takes an internal mutex, so the call is not strictly non-blocking.
- Execution may be delayed while the main thread is outside the bytecode evaluator.
- It targets the main interpreter, requiring an explicit decision about subinterpreter support.
- Behavior must be tested on both GIL-enabled and free-threaded Python.
- Pending cleanup must stop during Python finalization; payloads that cannot be reclaimed safely are intentionally leaked.

A dedicated host worker remains an alternative if `Py_AddPendingCall` cannot satisfy callback latency, blocking, subinterpreter, or shutdown requirements. The implementation choice should be validated with stress tests covering more than 32 simultaneous user-object destructions, scheduling failure, concurrent enqueue/drain races, free-threaded Python, and interpreter finalization.

## Attachment transactions

Provide internal operations for creating, installing, replacing, and removing immutable node attachment bundles.

Replacement follows a two-phase transaction:

1. Create the replacement bundle and retain its user object on the graph.
2. Perform the CUDA node-parameter update.
3. On success, publish the replacement in `GraphBox` and release the graph's old user-object reference.
4. On failure, release the replacement and leave the old record unchanged.

Node deletion releases its attachment only after `cuGraphDestroyNode` succeeds. Python handle invalidation and registry removal must also occur after driver success.

The public node-update APIs are out of scope for this issue, but this internal transaction support will be used by the follow-up definition-level update work.

## Implementation scope

- Replace the user-object-owned `GraphSlotTable` in `_cpp/resource_handles.*`.
- Introduce immutable per-node attachment payloads and attachment records.
- Add dynamic loading and wrappers for `cuGraphReleaseUserObject`.
- Replace sequential `graph_set_slot()` calls with one attachment installation per resource-bearing node.
- Keep graph attachment metadata available across equivalent `GraphHandle` wrappers.
- Correct `GraphNode.destroy()` ordering and remove attachments after successful deletion.
- Add a process-lifetime deferred-cleanup queue for `NodeAttachments` payloads.
- Make the CUDA user-object destructor enqueue the intact payload without releasing any `OpaqueHandle`.
- Prototype and select a wakeup mechanism, using coalesced `Py_AddPendingCall` scheduling as the primary candidate and a dedicated host worker as the fallback.
- Define retry, initialization, and shutdown behavior, including outstanding callbacks and queued payloads during Python finalization.

## Acceptance criteria

- Existing graph attachment lifetime tests continue to pass.
- An executable graph remains valid after its source node is deleted and source owners are dropped.
- A cloned graph retains attachments independently of source mutation or destruction.
- Reuse of a deleted `CUgraphNode` handle cannot invalidate an older executable graph.
- Failed node deletion leaves the Python node and its attachments valid.
- Nodes with multiple owners retain and release the complete attachment bundle together.
- Attachments are released after the last graph, executable graph, or in-flight launch reference disappears.
- The CUDA callback releases no `OpaqueHandle` and invokes no CUDA API. If `Py_AddPendingCall` is selected, it is the only permitted Python C API call from that callback.
- When a graph owns the final reference to a `Buffer`, its allocation remains valid through all dependent launches and is reclaimed from the deferred-cleanup worker.
- Final release of event, kernel/library, and Python-backed attachments occurs only from the selected deferred-cleanup context.
- Destroying an executable graph with an in-flight launch does not enqueue its attachments until CUDA determines that the launch no longer requires them.
- Stress coverage exceeds CPython's 32-entry pending-call capacity and exercises scheduling failure and concurrent enqueue/drain races.
- Deferred-cleanup shutdown cannot cause a callback to access destroyed queue state or invoke Python after finalization.

## Non-goals

- Public `GraphNode.update()` methods.
- Individual executable-graph node updates.
- Importing metadata for graphs cloned or modified outside `cuda.core`.
- Changing the existing raw-handle lifetime contract.

Andy-Jost:
## Summary

Fix the graph attachment lifetime problem described in #2350. Each resource-bearing node version now has an immutable attachment owned through its own CUDA user object, so source-graph mutation cannot release resources still used by clones, executable graphs, or in-flight launches.

Implementing that ownership change also required cuda-core's metadata to follow CUDA's graph topology and mutation boundaries. Each graph hierarchy keeps root, child, and conditional graph state coherent; clone and embed paths copy and rekey that metadata, and prepare/commit operations publish it only after the corresponding CUDA mutation succeeds. Together, these pieces make cuda-core's bookkeeping match the lifetimes CUDA actually enforces.

## Changes

- Replace the mutable graph-wide slot table with complete per-node attachment bundles and user objects.
- Track canonical root, child, and conditional graphs in a shared hierarchy while keeping attachment and node-handle metadata scoped to each graph.
- Copy and rekey attachment metadata when CUDA embeds cloned child graphs, and invalidate borrowed graph/node handles when CUDA destroys their backing objects.
- Preserve captured host callback owners when CUDA commits a callback but its node cannot be recovered.
- Keep graph mutation and metadata updates failure-aware via prepare/commit attachment transactions and failure-atomic deletion.
- Add contributor design documentation, release notes, and regressions across graph definitions, clones, executable updates, and launches.

## Review guide

1. Start with `cuda_core/cuda/core/_cpp/GRAPH_ATTACHMENTS.md` for the ownership model, graph hierarchy, common operations, and invariants.
2. Review `resource_handles.cpp/.hpp` for per-node user objects, canonical graph boxes, clone metadata import, handle invalidation, and deferred cleanup.
3. Review `_resource_handles.*` and `graph/*.pyx` for the Cython integration and CUDA-operation ordering.
4. Review the graph lifetime, mutation, builder, and update tests for correctness evidence and failure coverage.

## Test coverage

- Full local cuda-core test suite on GPU

## Related Work

- Closes #2350.
- Builds on callback-safe user-object cleanup from #2371.
- Part of #1330.
- Provides the ownership foundation for #2352.


cuda_core/cuda/core/_cpp/GRAPH_ATTACHMENTS.md
```diff
@@ -0,0 +1,263 @@
...
+- `invalidate_child_graph_state`: invalidate boxes and node handles after CUDA
+  destroys child graphs
+
+Private helpers `rekey_attachments` and `copy_attachments` implement recursive
+clone mapping.
+
+The Cython graph code creates CUDA nodes and then calls these operations to keep
+the C++ metadata synchronized with the driver.
+
+## Common operations
```
mdboom: Can we put some of this concrete procedural content as comments next to the relevant code?  It's sort of something you need to flip back-and-forth between to understand...

cuda_core/cuda/core/_cpp/GRAPH_ATTACHMENTS.md
```diff
@@ -0,0 +1,263 @@
+# Graph Node Attachments
+
+## Purpose
+
+CUDA user objects tie externally managed resource lifetimes to CUDA graphs.
+cuda-core uses them to keep resources referenced by graph node parameters
+alive. It also keeps non-owning metadata that maps each node to the
+user-object attachment backing its current parameters. This mapping lets
+cuda-core update the correct user-object references during graph changes and
+reproduce the node associations in graph clones.
```
mdboom: I feel like the motivation is missing here, and it jumps right into the solution.

The motivation is that we must keep all resources used by graph parameters alive for the lifetime of any graph that may use them.

The solution, is therefore to use CUDA user objects as a way to keep user resources alive.

Do I have that right?

cuda_core/cuda/core/_cpp/GRAPH_ATTACHMENTS.md
```diff
@@ -0,0 +1,263 @@
...
+complete parameter set that may refer to kernels, memory, events, callbacks, or
+other external resources. CUDA does not normally take ownership of those
+resources.
+
+A definition can be cloned, embedded as a child graph, or instantiated as a
+`CUgraphExec`. Each operation copies the definition's current nodes and
+parameters. The source and copy then have independent lifetimes: changing the
+source does not change an existing clone or executable. An embedded clone is
+owned by its parent node and is destroyed when that node is removed or
+replaced.
```
mdboom: Are we talking about definition in an abstract sense here, or `CUgraph` specifically?  If the latter, maybe:


```suggestion
A `CUgraph` can be cloned, embedded as a child graph, or instantiated as a
`CUgraphExec`. Each operation copies the definition's current nodes and
parameters. The source and copy then have independent lifetimes: changing the
source does not change an existing clone or executable. An embedded clone is
owned by its parent node and is destroyed when that node is removed or
replaced.
```

cuda_core/cuda/core/_cpp/GRAPH_ATTACHMENTS.md
```diff
@@ -0,0 +1,263 @@
...
+replaces their inherited state, while failure leaves the old state intact. An
+in-flight launch may continue using an older parameter set even after its
+source node or executable has changed or been destroyed. Direct
+executable-node updates require separate executable-specific ownership and are
+not tracked by this definition metadata.
+
+Several versions of one logical node's parameters may therefore remain live at
+the same time. Without version-specific ownership, replacing or deleting a node
+could release memory, kernels, callbacks, or other resources that an existing
+clone, executable, or launch still uses.
```
mdboom: What is meant by "version" here?  Isn't it just different "instances"?  If not, maybe define "version" here.

In compiler runtimes, versioned data structures are usually used for caching -- you can compare a version number on a dictionary, for example, to see if it's been mutated since the last time you cached something about it.  I don't see anything like that happening in the code (unless I missed it), so I feel like "instance" is a more precise term.

cuda_core/cuda/core/_cpp/GRAPH_ATTACHMENTS.md
```diff
@@ -0,0 +1,263 @@
...
+
+cuda-core prevents this by giving each parameter version an immutable
+`NodeAttachment` owned through a graph-retained CUDA user object. Non-owning
```
mdboom: For clarity -- this is immutable on the C++ side, but may reference mutable data on the Python side (most Python data is).  I don't think that matters to the implementation, I just found it confusing as to what layer is being talked about.

cuda_core/cuda/core/_cpp/GRAPH_ATTACHMENTS.md
mdboom: A mermaid graph would go a long way here, showing all of the objects and how the relate with owning and non-owning pointers.

cuda_core/cuda/core/_cpp/GRAPH_ATTACHMENTS.md
```diff
@@ -0,0 +1,263 @@
...
+- the `CUgraph`
+- its parent box and owning node, when it is a child graph
+- a `GraphAttachmentMap` from `CUgraphNode` to `NodeAttachment*`
+- a per-graph weak registry of canonical `GraphNodeHandle` objects
+
+`GraphHierarchy::graphs` stores live boxes in parent-before-child order. When
+CUDA destroys a child graph, cuda-core unregisters its raw handle, nulls the
+resource, clears its non-owning metadata, and moves the box to
+`GraphHierarchy::graveyard`. Keeping this tombstone at a stable address lets
+existing aliasing handles safely observe that the graph is invalid. Removing
```
mdboom: Where does this happen?  I only see graphs being put in the graveyard.  Must be something more subtle going on?

cuda_core/cuda/core/graph/_graph_definition.pyi
```diff
@@ -56,6 +56,12 @@ class GraphDefinition:
     A GraphDefinition is used to construct a graph explicitly by adding nodes
     and specifying dependencies. Once construction is complete, call
     instantiate() to obtain an executable Graph.
+
+    Notes
+    -----
+    Definitions that view the same root, child, or conditional graph hierarchy
+    share underlying graph state. Mutations anywhere in that hierarchy must be
+    externally synchronized.
```
mdboom, line 64: The only real issue my agent found was that self-referencing a graph may create a cycle that the Python garbage collector may be unable to break.  We could add a comment to that effect, but I sort of think it seems very unlikely someone would do that.

mdboom (commented):
I have some clarifying questions about the design doc.  Once I get my head around that I'll come back and look more closely at the code.

cuda_core/cuda/core/_cpp/GRAPH_ATTACHMENTS.md
Andy-Jost: clarified

cuda_core/cuda/core/_cpp/GRAPH_ATTACHMENTS.md
Andy-Jost: Added a new section to address this. I also added a simple Mermaid diagram showing the Hierarchy/Box/Handle relationship and cleaned up the text in that section.

cuda_core/cuda/core/_cpp/GRAPH_ATTACHMENTS.md
Andy-Jost: I added an introduction that should clarify

cuda_core/cuda/core/_cpp/GRAPH_ATTACHMENTS.md
Andy-Jost: Good call. I removed this stuff, except for the section on deferred cleanup. Key procedural details were moved into brief code comments. The document is overall much shorter and more focused.

cuda_core/cuda/core/_cpp/GRAPH_ATTACHMENTS.md
Andy-Jost: Clarified to
> a `NodeAttachment` whose owner bundle is immutable after publication (although the resources it keeps alive may themselves be mutable).

Transitive mutability is a real problem without a good solution I can see. Example: someone could embed a `cuda.core.Buffer` in a kernel parameter bundle and then close the buffer.

cuda_core/cuda/core/_cpp/GRAPH_ATTACHMENTS.md
Andy-Jost: This occurs in `invalidate_child_graph_state`. The object moved to the graveyard is a `GraphBox`, which contains the destroyed child graph’s `CUgraph` and related state. Each `GraphHandle` aliases that `CUgraph` field while sharing ownership of the containing `GraphHierarchy`.

When a parent node containing a child graph is destroyed, CUDA also destroys the embedded child graph, invalidating its raw handle. This PR sets the `GraphBox`’s `CUgraph` field to null so existing aliases reflect that invalidation, then moves the retired box to the graveyard so the aliased field’s address remains valid.

This machinery is necessary because CUDA couples removal of an embedded child graph from its parent with destruction of that graph. If the API allowed the graph to be detached and its ownership transferred to the caller, cuda.core could use an ordinary owning handle and call `cuGraphDestroy` after its final reference was released.

cuda_core/cuda/core/graph/_graph_definition.pyi
Andy-Jost: Thanks for pointing this out. The general problem is that Python's cyclic garbage collector cannot trace references through the driver, so reference cycles involving retained user objects result in leaks. This limitation is very difficult to remove, so for this PR I opted to update the docs.

I noted the limitation in `GRAPH_ATTACHMENTS.md` and the docstrings (example [here](https://github.com/Andy-Jost/cuda-python/blob/c2759d269d8c6eb2605b7562b02afb01f1cf3212/cuda_core/cuda/core/graph/_graph_node.pyx#L500-L504)) for API functions that retain Python objects on graphs:

  • `GraphNode.callback` and `GraphBuilder.callback`
  • `GraphNode.launch` for retained kernel arguments
  • `GraphNode.memset` and `GraphNode.memcpy` for retained buffers and explicit owners

cuda_core/cuda/core/_cpp/GRAPH_ATTACHMENTS.md
mdboom: Yes.  Transitive mutability is a big problem in Python in general -- see this PEP: https://pep-previews--4468.org.readthedocs.build/pep-0795/

It may be years until we can rely on that, unfortunately.

cuda_core/cuda/core/graph/_graph_definition.pyi
mdboom: > so reference cycles involving retained user objects result in leaks

Yes, I agree.  I think document and move on is the right choice.  I can't conceive of a legit use case to need to do that.

cuda_core/tests/graph/test_graph_definition_mutation.py
```diff
@@ -307,6 +310,42 @@ def test_destroyed_node(init_cuda):
...
+    assert node.is_valid
+    assert node in graph_def.nodes()
+    assert edge in graph_def.edges()
+    assert callback_weak() is not None
+
+    graph = graph_def.instantiate()
+    stream = Device().create_stream()
+    graph.launch(stream)
+    stream.sync()
+    assert called[0]
```
mdboom, line 346: Is it worth checking here (as the above test), that deleting the graph ultimately ends up in the callback being deleted?

mdboom (approved):
Thanks, the design docs look much better now.

I read through the tests carefully this time, and they seem to be testing what we need to.

The C++ basically makes sense to me, but my C++ is rusty enough that I'm not sure I would spot any best practice problems that my agent wouldn't catch.

One non-blocking suggestion: This is fiddly enough that I wonder if it would be helpful to add assertions that confirm the invariants hold at key points.  These would compile away in a standard build, but maybe we could turn them on in at least one CI config to catch things earlier.  (Those sorts of things have been helpful in the CPython codebase).

cuda_core/cuda/core/_cpp/GRAPH_ATTACHMENTS.md
Andy-Jost: This is great, thanks for sharing. My spiciest opinion that I came to after writing a Haskell compiler long ago is that everything should be immutable always.

cuda_core/tests/graph/test_graph_definition_mutation.py
Andy-Jost: I don't think so. The existing tests `test_callback_survives_source_node_deletion` and `test_attachment_released_only_after_last_executable_is_destroyed` already cover it.

PR #2371, opened 2026-07-14
cuda.core: defer graph user-object payload cleanup

Andy-Jost:
## Summary

Move graph attachment payload destruction off CUDA's internal user-object callback thread. The callback now performs only an intrusive queue handoff and one coalesced `Py_AddPendingCall`; resource handles and Python owners are released later from Python's main thread, where their deleters may safely call CUDA.

## Changes

- Add a process-lifetime lock-free deferred-cleanup queue with one pending-call wakeup regardless of payload count.
- Preserve queued payloads when pending-call scheduling fails and retry from later safe graph-attachment entries.
- Stop cleanup during interpreter finalization and intentionally leak intact payloads when safe cleanup is no longer possible.
- Initialize shutdown state explicitly during module import and document the callback/threading contract.
- Add stress coverage for more than 32 callbacks, pending-call queue saturation, main-thread finalization, and interpreter shutdown.

## Related Work

- Part of #2350.
- Prerequisite for #2357.

### DIFF
--- a/cuda_core/cuda/core/_cpp/resource_handles.cpp
+++ b/cuda_core/cuda/core/_cpp/resource_handles.cpp
@@ -202,6 +202,7 @@ class DeferredReclaimer;
 
 using DeferredDestroyFn = void (*)(void*) noexcept;
 
+// Preallocated envelope that transfers one payload from CUDA to the reclaimer.
 struct DeferredDelete {
     std::atomic<DeferredDelete*> next{nullptr};
     void* payload;
@@ -213,8 +214,10 @@ struct DeferredDelete {
         : payload(payload_), destroy(destroy_), reclaimer(reclaimer_) {}
 };
 
+// Process-lifetime MPSC queue that drains payloads from Python's main thread.
 class DeferredReclaimer {
 public:
+    // Transfer one preallocated payload envelope from a producer to the queue.
     void retire(DeferredDelete* item) noexcept {
         DeferredDelete* head = head_.load(std::memory_order_relaxed);
         do {
@@ -224,20 +227,24 @@ public:
         try_schedule();
     }
 
+    // Permanently disable pending-call scheduling during interpreter shutdown.
     void stop() noexcept {
         accepting_.store(false, std::memory_order_release);
     }
 
+    // Reattempt scheduling for payloads left queued after an earlier failure.
     void retry() noexcept {
         try_schedule();
     }
 
 private:
+    // Adapt the reclaimer drain to CPython's int (*)(void*) callback ABI.
     static int pending_call(void* arg) noexcept {
         static_cast<DeferredReclaimer*>(arg)->drain();
         return 0;
     }
 
+    // Coalesce all queued work behind at most one CPython pending call.
     void try_schedule() noexcept {
         if (!accepting_.load(std::memory_order_acquire) ||
             !head_.load(std::memory_order_acquire)) {
@@ -256,6 +263,7 @@ private:
         }
     }
 
+    // Detach and destroy all queued payloads from Python's main thread.
     void drain() noexcept {
         if (!Py_IsInitialized() || py_is_finalizing()) {
             stop();
@@ -280,8 +288,11 @@ private:
         }
     }
 
+    // Head of the intrusive multi-producer, single-consumer payload stack.
     std::atomic<DeferredDelete*> head_{nullptr};
+    // True while one cuda-core drain callback is pending or executing.
     std::atomic<bool> scheduled_{false};
+    // False once shutdown begins, causing later payloads to be leaked safely.
     std::atomic<bool> accepting_{true};
 };
 
@@ -303,11 +314,13 @@ DeferredDelete* make_deferred_delete(
     return new DeferredDelete(payload, destroy, reclaimer);
 }
 
+// Destroy an envelope synchronously before CUDA has acquired ownership.
 void destroy_deferred_delete_now(DeferredDelete* item) noexcept {
     item->destroy(item->payload);
     delete item;
 }
 
+// CUDA's CUhostFn ABI is void (*)(void*); recover the typed envelope and queue it.
 void enqueue_deferred_delete(void* item) noexcept {
     auto* deferred = static_cast<DeferredDelete*>(item);
     deferred->reclaimer->retire(deferred);
--- a/cuda_core/cuda/core/_cpp/resource_handles.cpp
+++ b/cuda_core/cuda/core/_cpp/resource_handles.cpp
@@ -204,14 +204,12 @@ using DeferredDestroyFn = void (*)(void*) noexcept;
 
 // Preallocated envelope that transfers one payload from CUDA to the reclaimer.
 struct DeferredDelete {
-    std::atomic<DeferredDelete*> next{nullptr};
+    DeferredDelete* next = nullptr;
     void* payload;
     DeferredDestroyFn destroy;
-    DeferredReclaimer* reclaimer;
 
-    DeferredDelete(void* payload_, DeferredDestroyFn destroy_,
-                   DeferredReclaimer* reclaimer_) noexcept
-        : payload(payload_), destroy(destroy_), reclaimer(reclaimer_) {}
+    DeferredDelete(void* payload_, DeferredDestroyFn destroy_) noexcept
+        : payload(payload_), destroy(destroy_) {}
 };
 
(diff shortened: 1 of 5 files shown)


Andy-Jost:
PR #2371 separates “CUDA says this payload is no longer needed” from actually cleaning it up.

## Parts

### 1. `DeferredCleanupItem`

A small queue item allocated before CUDA sees the user object. It contains:

- The payload, currently `GraphSlotTable*`
- The function that cleans up that payload
- A `next` pointer for the intrusive queue

The process-wide cleanup queue pointer is published atomically once during module initialization and loaded by the CUDA callback; it is not repeated in each item. Because the item is preallocated, CUDA’s callback does not allocate memory.

### 2. `DeferredCleanupQueue`

A process-lifetime object containing:

- `head_`: atomic head of an intrusive multiple-producer, single-consumer queue
- `scheduled_`: whether a Python pending call is already scheduled
- `accepting_`: disabled during Python finalization

It is intentionally never freed, so a late CUDA callback can never access destroyed queue state.

### 3. CUDA callback

CUDA invokes `enqueue_cleanup()` when the user-object reference count reaches zero.

The callback only:

1. Atomically enqueues its `DeferredCleanupItem`.
2. Attempts to schedule one pending call.

It does not release a handle, call CUDA, allocate, or wait for cleanup.

### 4. Pending-call scheduling

`schedule()` atomically changes `scheduled_` from false to true. Therefore, many CUDA callbacks still produce at most one outstanding `Py_AddPendingCall`.

If CPython’s pending-call queue is full:

- Scheduling returns `-1`.
- `scheduled_` returns to false.
- The payload remains intact in cuda-core’s cleanup queue.
- A later enqueue or safe graph-attachment entry retries scheduling.

Retrying schedules cleanup; it does not enqueue the item again.

### 5. Main-thread drain

The pending callback runs on Python’s main thread. It atomically exchanges `head_` with null, taking exclusive ownership of the entire queued list.

It then invokes each item’s cleanup function and deletes the item. Any resulting Python finalizer or CUDA-backed handle deleter runs from this permitted host context.

Items arriving during a drain are either picked up by its next loop iteration or cause another pending call to be scheduled after `scheduled_` is cleared.

### 6. Shutdown

A `Py_AtExit` callback sets `accepting_` false. The drain also checks whether Python is finalizing.

After that point:

- CUDA callbacks may still enqueue safely because the cleanup queue lives for the process lifetime.
- No new Python callback is scheduled.
- Queued payloads are intentionally leaked rather than risking Python/CUDA use during shutdown.

## Why cleanup happens once

Each payload follows one of two mutually exclusive paths:

```text
CUDA user-object creation fails:
    item allocated → cleanup executed directly

CUDA user-object creation succeeds:
    CUDA-owned → callback → enqueued → atomically detached → cleanup executed
```

After successful `cuUserObjectCreate`, cuda-core never executes the item’s cleanup directly. CUDA invokes its destructor callback once, when the user object’s total reference count reaches zero.

Within the cleanup queue:

- The CUDA callback enqueues that item once.
- Scheduling retries never enqueue it again.
- `head_.exchange(nullptr)` transfers each queued item to the single consumer exactly once.
- The cleanup function is `noexcept`, so processing cannot abandon an item halfway through.
- The item is deleted immediately after its payload is cleaned up.

Thus safety relies on CUDA’s user-object contract providing one final destructor invocation. The cleanup queue then preserves that exactly-once transition through atomic ownership transfer; it does not need a separate “already cleaned” flag.

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -185,6 +186,162 @@ class GILAcquireGuard {
...
+
+void initialize_deferred_cleanup() {
+    if (deferred_cleanup_queue.load(std::memory_order_acquire)) {
+        return;
+    }
+    auto* queue = new DeferredCleanupQueue();
+    deferred_cleanup_queue.store(queue, std::memory_order_release);
+    if (Py_AtExit(stop_deferred_cleanup) != 0) {
+        queue->stop();
+    }
```
mdboom, line 342: This is thread-safe only as long as the GIL is held -- it is in practice (also it's run once at import under normal conditions) but maybe just add a comment.

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -185,6 +186,162 @@ class GILAcquireGuard {
...
+
+    // Permanently disable pending-call scheduling during interpreter shutdown.
+    void stop() noexcept {
+        accepting_.store(false, std::memory_order_release);
+    }
+
+    // Reattempt scheduling for payloads left queued after an earlier failure.
+    void retry_schedule() noexcept {
+        schedule();
+    }
```
mdboom, line 236: Maybe add a little context about why retrying would be necessary?  I don't think it's necessarily unsound, just could use a little motivating statement.

mdboom (approved):
My initial reaction was that this might be more complex than it needs to be.  `Py_AddPendingCall` already functions as a queue (you can call it multiple times and it will run all of the calls in order the next chance it calls any of them), so it might not be necessary to also have a queue on our side.

/However/ `Py_AddPendingCall` is a FIFO queue, with a maximum of 300 elements (and it appears to silently evict elements if it's more than 300).  All that is "probably fine", but it's an implementation detail that could change.

So I think what's here is preferable -- we can be in charge of our own destiny on the queueing.  (Say, for example, we discover that the ordering needs to change, this gives us a place to do so).

My other comments are just around improving comments and docs.

Andy-Jost:
Thanks for the review, @mdboom. I'll wrap the touch-ups into the next PR (#2357).

As for the complexity, I largely agree. My initial idea was very similar to yours: simply make a call to `Py_AddPendingCall` from CUDA's destructor thread for each user object. That ran into a few snags:

- `Py_AddPendingCall` only accepts 32 queued callbacks. I had my agent follow up after you mentioned 300 entries and it seems to stand by that number (details below).
- When `Py_AddPendingCall` fails for any reason on the CUDA thread, we're totally stuck without any recourse other than to leak that object. Introducing a queue we manage where insertion never fails fixes that.

So I agree, the complexity is higher than anticipated, but I don't think it can be reduced.

## Queue limit details

>   My “32” was substantively right for Py_AddPendingCall; Mike found the newer 300-entry backing array but not the effective limit of this public API. Exact answer:
> 
>   • Python 3.10–3.12: array size 32, but the ring buffer reserves one slot, so only 31 callbacks fit.
>   • Python 3.13–3.15, including free-threaded builds: backing array size is 300, but Py_AddPendingCall targets pending_mainthread, whose configured maximum remains 32.
>   • Full queues do not evict callbacks. Py_AddPendingCall returns -1.
> 
>   Source findings
>   • Python 3.10 defines NPENDINGCALLS 32 (https://github.com/python/cpython/blob/3.10/Include/internal/pycore_interp.h) and its push function returns -1 when advancing would meet first (https://github.com/python/cpython/blob/3.10/Python/ceval.c#L584-L598). That leaves 31 usable entries.
>   • Python 3.13 introduces PENDINGCALLSARRAYSIZE 300, but also explicitly defines MAXPENDINGCALLS_MAIN 32 (https://github.com/python/cpython/blob/3.13/Include/internal/pycore_ceval_state.h).
>   • The runtime initializes pending_mainthread.max from that 32-entry limit in pycore_runtime_init.h (https://github.com/python/cpython/blob/3.13/Include/internal/pycore_runtime_init.h).
>   • Py_AddPendingCall (https://github.com/python/cpython/blob/3.13/Python/ceval_gil.c#L805-L819) explicitly selects _Py_PENDING_MAINTHREADONLY; when that queue is full, it returns -1.
>   • Python 3.15 retains MAXPENDINGCALLS_MAIN 32 (https://github.com/python/cpython/blob/3.15/Include/internal/pycore_ceval_state.h).
>   • The public documentation (https://docs.python.org/3.14/c-api/threads.html#c.Py_AddPendingCall) promises -1 on failure but intentionally does not document a capacity.

rwgk:
I'm seeing what must be a flake here (first attempt):

* https://github.com/NVIDIA/cuda-python/actions/runs/29451966527/job/87480738831?pr=2374

```
tests\graph\test_graph_definition_lifetime.py
```
```
SystemError: D:\a\1\s\Python\ceval_gil.c:1017: bad argument to internal function
```

I'll try a rerun.

rwgk:
The rerun succeeded:

https://github.com/NVIDIA/cuda-python/actions/runs/29451966527/job/87494564108?pr=2374

PR #2374, opened 2026-07-15
Pathfinder: add support on cutensorMp library loading

yangcal:
## Description

This PR enables pathfinder to support cutensorMp library loading

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/descriptor_catalog.py
+++ b/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/descriptor_catalog.py
@@ -376,6 +376,13 @@ DESCRIPTOR_CATALOG: tuple[DescriptorSpec, ...] = (
         dependencies=("cutensor", "cublasLt"),
     ),
     DescriptorSpec(
+        name="cutensorMp",
+        packaged_with="other",
+        linux_sonames=("libcutensorMp.so.2",),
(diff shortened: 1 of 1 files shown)


yangcal:
@b-kloss @ChrisPsa for vis, for cutensorMp bindings, we need to first have library loading support from pathfinder

rwgk:
QQ (I only have a minute right now):

Did you already look at `cuda_pathfinder/pyproject.toml`?

Do we need to add `cutensorMp` to the test dependencies? (Do we need to distinguish cu12, cu13?)

yangcal:
@rwgk good question.

`cutensorMp` is shipped as part of the `cutensor-cu12/13` package and the `cutensor-cuXX` wheels are already included in the `cu12/13` dependency groups so there is no action required there. For the so path, there is no difference between `cu12/cu13`, both under `site-packages/cutensor/lib` (same as the existing base `cutensor` library)

Re: tests, I believe the existing parameterized tests automatically include `cutensorMp` as part of descriptor tests and linux dynamic loading tests. Therefore no action should be needed there either.

rwgk (approved):
Codex figured out that `cudart` was missing in the `dependencies` list. I went ahead an added it, then triggered the CI.

yangcal:
> Codex figured out that `cudart` was missing in the `dependencies` list. I went ahead an added it, then triggered the CI.

I can confirm `cudart` is indeed a required dependency. Thanks for putting up the fix

PR #2378, opened 2026-07-16
[no-ci] Support organization-owned forks in AGENTS.md

rwgk:
## Description

The original goal of this PR is narrow: **when a user directly asks an agent to create a CUDA Python pull request, the agent should automatically know how to use the contributor's approved fork, whether that fork is owned by a personal account or an organization.**

This PR does not tell agents when to create pull requests. A direct user request to create one is the policy boundary. The repository guidance and skill supply only the technical information needed after that request.

The existing guidance assumes that every contributor uses a personal fork and that `gh pr create` can always identify it. That breaks down for some organization-owned forks whose repository name differs from the base repository. [GitHub CLI issue cli/cli#10093](https://github.com/cli/cli/issues/10093) tracks that limitation; GitHub's GraphQL API supports the topology through `headRepositoryId`.

The implementation separates the shared repository invariant from the
on-demand procedure:

- `AGENTS.md` contains only the shared repository invariant: the canonical upstream repository is read-only, and pull-request branches must be pushed to an approved personal or organization-owned fork associated with the contributor.
- The repository skill `$create-cuda-python-pull-request` contains the procedural details for inspecting fork topology, preparing required PR metadata, using `gh pr create` when possible, and applying the GraphQL fallback when necessary.
- The skill's automatic trigger is narrowly scoped to a user request that directly asks to create or open a CUDA Python pull request. Its trigger explicitly excludes implementation, commits, pushes, branch preparation, PR advice, and general GitHub work.

This preserves the original automatic behavior without putting PR-creation procedure in always-loaded repository instructions. The user's request determines whether a PR should be created; the skill determines how to create the requested PR correctly.


AGENTS.md
```diff
@@ -14,26 +14,103 @@ guide for package-specific conventions and workflows.
...
+Do not push to the canonical repository unless the user explicitly requests,
+in the current conversation, a push to a specific branch in that repository.
+This can be necessary, for example, to test workflows that run only for
+upstream branches. Before such a push, verify the exact `OWNER/REPOSITORY`,
+remote, and destination branch. Authorization to push a test branch does not
+authorize pushing to a default or protected branch, creating tags,
+force-pushing, or deleting branches; those actions require separate explicit
+authorization.
+
+Use `gh pr create` when it can identify the fork unambiguously. Always select
```
mdboom: So if the agent can identify the fork unambiguously, it is free to create a PR?  Certainly that's not what's intended, but 100% can imagine the agent doing the wrong thing here.

AGENTS.md
```diff
@@ -14,26 +14,103 @@ guide for package-specific conventions and workflows.
...
+
+```
+gh pr edit <pr-number> --repo "${BASE_REPO}" \
+  --add-assignee "<assignee>" --add-label "<label>" \
+  --milestone "<milestone>"
+```
+
+If you are unsure which label or milestone to use, inspect the base repository
+with `gh label list --repo "${BASE_REPO}"` and `gh api
+"repos/${BASE_REPO}/milestones" --jq '.[].title'`, then pick the best match.
```
mdboom:
```suggestion
If you are unsure which label or milestone to use, don't guess. Flag it for the user and have them fill it in.
```

AGENTS.md
```diff
@@ -14,26 +14,103 @@ guide for package-specific conventions and workflows.
...
+- Run `git remote -v` and resolve the complete `OWNER/REPOSITORY` names of the
+  base repository and intended fork.
+- Confirm that the push target is a fork of the base repository and is not the
+  base repository itself. Compare complete repository names; the owner alone
+  is not sufficient to distinguish them.
+- Push using the explicit fork remote and branch, for example
+  `git push <fork-remote> <branch>`.
+
+Do not push to the canonical repository unless the user explicitly requests,
+in the current conversation, a push to a specific branch in that repository.
```
mdboom: This opens up a window for mistakes.  What's wrong with saying "Never push to the canonical repository"?  The user can always do it by hand anyway.  None of us should be running with write access to the canonical repository in our local checkouts -- that's too dangerous.

AGENTS.md
```diff
@@ -14,26 +14,103 @@ guide for package-specific conventions and workflows.
...
-(`git push <personal-fork-remote> <branch>`). Open the PR from that fork with
-`gh pr create`. Do not use `git push upstream` or any command that writes to
-the `upstream` remote.
-
-When creating pull requests with `gh pr create`, always assign at least one
-label and a milestone. CI enforces this via the `pr-metadata-check` workflow
-and will block PRs that are missing labels or a milestone. Use `--label` and
-`--milestone` flags, for example:
+Treat the canonical repository used as the pull-request base as read-only by
+default. Push branches and commits to an approved fork, which may be owned by
```
mdboom:
```suggestion
Treat the canonical repository used as the pull-request base as read-only. Push branches and commits to an approved fork, which may be owned by
```

AGENTS.md
```diff
@@ -14,26 +14,103 @@ guide for package-specific conventions and workflows.
...
+        headRefName: $headRefName
+        title: $title
+        body: $body
+        draft: $draft
+      }) {
+        pullRequest { number url }
+      }
+    }' \
+  --jq '.data.createPullRequest.pullRequest'
+```
```
mdboom: What's the rationale for including detailed information about the GraphQL API here?  Certainly the agent has access to the docs, and I worry this would bitrot.

mdboom (changes requested):
I've had enough run-ins with my agent pushing branches and creating pull requests before they were ready that I've worked really hard to get my agent to /never/ write anything in a remote location without my consent (and I prefer to be in the driver's seat even now for those steps).

I'm worried that a lot of the language here will encourage my agent to act too autonomously again. We should instead frame it in terms of things we want it /not/ to do (don't push to a fork that can't be the source of the PR you are trying to create), rather than things /to do/ (use `gh pr` to create PRs).  To put it another way -- this is shared context we want everyone to agree on to collaborate well together.  It's not intended to prescribe /how/ to accomplish that.  A pre-AI analogy would be: we all agree to write standard text files and share them with git.  We don't have to agree on what editor we are using or what git tool we use.

Therefore, I think the only required change here should be:

```
All branch creation and pushes must go to the contributor's
personal fork.
```

changes to:

```
Branch creation and pushes must not go directly to the canonical upstream fork. Use the user's personal fork, or if that is not supported as an official fork of upstream, use a intra-repo fork "owned" by the user.  Confirm that the push target is a fork of the base repository and is not the base repository itself.
```

AGENTS.md
rwgk: Addressed: PR creation now requires an explicit user request.

AGENTS.md
rwgk: Addressed: Agents must ask rather than guess metadata.

AGENTS.md
rwgk: "By default" was removed.

AGENTS.md
rwgk: Addressed with a compromise based on the team discussion: the full mutation example was replaced with the essential, more durable API details. The CLI issue is now linked so maintainers can determine when the workaround is no longer needed.

AGENTS.md
rwgk: Addressed: The canonical repository is unconditionally read-only to agents.

The wording is very strong now. Let's see how it'll work in practice for situations that require pushes to upstream (i.e. for testing changes to GHA workflows; or stacked PRs).

rwgk (commented):
@mdboom Could you please take another look?

The section now separates remote-write policy from technical procedure. It explicitly requires direct user authorization, states that a push request does not authorize PR creation, and clarifies that the technical instructions grant no permission by themselves.

mdboom:
> @mdboom Could you please take another look?
> 
> The section now separates remote-write policy from technical procedure. It explicitly requires direct user authorization, states that a push request does not authorize PR creation, and clarifies that the technical instructions grant no permission by themselves.

I'm still in favor of just fixing the bug (that it mentions personal forks and intra-org forks are also ok).  The rest of this feels like unnecessary prescription.  Any of us can put it in our local instructions if we want this, but I don't think we all will.

Can we move the PR creation stuff to a skill such that I could easily tell my agent to ignore it?

rwgk:
Thanks @mdboom. I took your latest comment to mean that the shared repository invariant belongs in `AGENTS.md`, while the on-demand PR-creation mechanics should not be part of those always-loaded instructions.

I have reworked the PR around that distinction. `AGENTS.md` now says only that the canonical upstream repository is read-only and that pull-request branches go to an approved personal or organization-owned fork associated with the contributor.

I also want to preserve the original goal of this PR: when a user directly asks an agent to create a PR, the agent should automatically know how to handle either fork topology, including the current `gh pr create` limitation. Those mechanics now live in `$create-cuda-python-pull-request`. Its automatic trigger is narrowly scoped to direct PR-creation requests and explicitly excludes implementation, commits, pushes, branch preparation, PR advice, and general GitHub work.

In other words, the user's direct request determines whether a PR should be created; the skill only supplies the mechanics for creating that requested PR correctly.

Could you please take another look?

If the current PR does not sufficiently capture what you have in mind, please feel free to push edits directly to this PR branch.

mdboom (approved):

PR #2379, opened 2026-07-16
Fix potential data race in bindings init functions

mdboom:
Corresponds to generator MR 480


leofang:
cybind MR 480 was already merged. Let's keep things in sync.

leofang (approved):

leofang:
It seems the runner is stuck. Let me admin-merge.

PR #2383, opened 2026-07-17
cuda.core: refresh frozen CUresult explanation table for CUDA 13.3

### LINKED ISSUE #2370: [MNT] cuda.core: refresh the frozen CUresult explanation table for CUDA 13.3
## Summary

[`_utils/driver_cu_result_explanations_frozen.py`](https://github.com/NVIDIA/cuda-python/blob/c000331de6c37aa4565af74b001271ffcf6d5c99/cuda_core/cuda/core/_utils/driver_cu_result_explanations_frozen.py#L4)
is a fallback table of `CUresult` explanation strings snapshotted at **CUDA Toolkit v13.1.1**.

CUDA 13.3 added `CUDA_ERROR_GRAPH_RECAPTURE_FAILURE` (13.2 added no new driver error codes), so
the new code currently has no fallback explanation text. The error will start appearing once
graph re-capture lands (#2359).

## Proposed approach

- Regenerate the snapshot against CUDA 13.3 using the same process that produced the 13.1.1
  table (helpers are exercised by `tests/test_utils_enum_explanations_helpers.py` /
  `tests/test_cuda_utils.py`).
- If not already covered there: add a tripwire test comparing `cuda.bindings` `CUresult`
  members against the table keys, so future toolkit bumps flag missing entries automatically.
- Consider adding the refresh to the release checklist for new CTK support.

## References

- Found during the CUDA 12.8 → 13.3 bindings vs. cuda.core gap sweep (2026-07-14)

-- Leo's bot

Symbioose:
## Description

closes #2370 

This regenerates the frozen `CUresult` fallback explanation table
(`driver_cu_result_explanations_frozen.py`) for CUDA 13.3, which added
`CUDA_ERROR_GRAPH_RECAPTURE_FAILURE` (code 918). This table is only used
as a fallback for `cuda-bindings` versions that don't yet expose usable
enum `__doc__` text (see `enum_explanations_helpers.py`), but it needs to
stay in sync with new CUDA Toolkit error codes so those users still get
an informative error message.

Changes:
- Added the `918: "..."` entry to `_FALLBACK_EXPLANATIONS` and bumped the
  version comment to CUDA Toolkit v13.3.
- Added a tripwire test (`test_frozen_driver_table_covers_all_curesult_members`)
  that compares `cuda.bindings.driver.CUresult` members against the frozen
  table keys, so future toolkit bumps that add new driver error codes are
  caught automatically.
- Documented the refresh step in the cuda.core release checklist
  (`RELEASE-core.md`).

## Checklist


Symbioose:
Hi @leofang, this is ready for review. First-time contributor, let me know if I'm missing anything 👍, and could you add an assignee/milestone so the metadata check passes? Thanks!

mdboom (approved):

kkraus14 (changes requested):
These fallback explanations are only used when we didn't have a richer explanation of the result code available as a `__doc__` property, which started with cuda-bindings 12.9.6 / 13.2.0. Since the result code being added is specific to new CUDA 13.3 features it can't be returned in situations where the fallback explanations are being used.

We could still merge this to keep the fallbacks up to date in case a new CUDA release somehow returns a new error code from an existing API, but if we wish to merge this both the driver and runtime explanation need to be updated.

leofang:
This seems to make sense to me. I'll run CI and then admin-merge.

leofang:
/ok to test [82e40fc](https://github.com/NVIDIA/cuda-python/pull/2383/commits/82e40fcb92c184dfa74538db73f304eb8be0a4f8)

leofang (approved):

rwgk:
To mention this here: PR #2792 reverted this PR, while also adding a source comment to explain better.

PR #2384, opened 2026-07-17
Fix #2377: Reorganize the test helpers (where appropriate) to cuda_python_test_helpers

### LINKED ISSUE #2377: MNT: Consolidate test helpers

mdboom:
This moves test helpers that are generically useful across the subprojects to `cuda_python_test_helpers`.  It fixes some corner cases in the `sys.path` hack to import `cuda_python_test_helpers`.  It doesn't install it as a proper package -- I had hoped that would be possible, but upstream `pip` does not support `path` dependencies in a dependency group.  Using that would require `pip` to be improved or require everyone to use `uv`.


cuda_python_test_helpers/cuda_python_test_helpers/__init__.py
```diff
@@ -26,6 +27,7 @@ def _detect_wsl() -> bool:
 
 IS_WSL: bool = _detect_wsl()
 IS_WINDOWS: bool = platform.system() == "Windows" or sys.platform.startswith("win")
+IS_LINUX: bool = not IS_WINDOWS and not IS_WSL and platform.system() == "Linux"
```
juenglin, line 30: Should be just `platform.system() == "Linux"` imho.

juenglin (approved):
LGTM

cuda_python_test_helpers/cuda_python_test_helpers/__init__.py
mdboom: On WSL, `platform.system()` is `"Linux"`, so you still need to do the WSL check.  (Though the WINDOWS check is probably redundant).  In any event, this was pre-existing and this PR was intended as just a re-organization.

rwgk:
My codex thinks this failure

https://github.com/NVIDIA/cuda-python/actions/runs/30987658226/job/92246084554?pr=2509

is due to this PR. See below. Is this already on your radar?

```
• Confirmed: this is unrelated to PR 2509 and is deterministic, not a flake.

  - The job passes all 1,276 Pathfinder tests, then fails on the first bindings import with ModuleNotFoundError: No module named 'cuda'.
  - Pixi never builds cuda-bindings, so PR 2509’s modified build hooks are never executed.
  - The same failure already occurred in PR 2384, which was merged as 19e66499c2.
  - PR 2384 changed both pixi.toml files without updating their lockfiles. With Pixi 0.73, the first automatic lock refresh produces an
    environment missing the local cuda-bindings and cuda-pathfinder packages.

  - I reproduced the exact failure from a clean checkout using the CI’s Pixi 0.73. A second invocation in the same checkout converges and
    succeeds, but rerunning GitHub Actions starts fresh and repeats the first-pass failure.

  - The new TOML table placement also appears to unintentionally move several existing test dependencies from Conda to PyPI.

  I recommend fixing this separately from PR 2509:

  1. Keep the existing test dependencies under [feature.test.dependencies].
  2. Put only cuda-python-test-helpers under [feature.test.pypi-dependencies], assuming that dependency is intentional.
  3. Regenerate both cuda_bindings/pixi.lock and cuda_core/pixi.lock with Pixi 0.73 until they are stable and pixi lock --check passes from
     a clean checkout.

  4. Exercise both source-build/import commands.

  The same stale-lock problem exists in cuda_core, so fixing only the bindings lock would likely expose the core failure next. Optional
  hardening would be to make CI use locked mode so future manifest/lock mismatches fail explicitly instead of silently rewriting the lock.
```

juenglin:
@rwgk - Fix here: https://github.com/NVIDIA/cuda-python/pull/2532

PR #2389, opened 2026-07-19
Fix SM resource alignment discovery test

rwgk:
## Description

Update the SM-resource discovery test for the intentional CUDA 13.4 driver semantic change reported under nvbug 6464867, and document its user-visible effect.

### Analysis (gpt-5.6-sol)

`SMResourceOptions(count=None)` leaves `coscheduledSmCount` at 0. CUDA continues to use its default co-scheduled group size internally while preserving 0 in `groupParams`. Starting with CUDA 13.4, however, the driver no longer guarantees that the discovered `smCount` is a multiple of that default or of the device's `smCoscheduledAlignment`. An unconstrained request may therefore return every available SM; on B100 it returns 148 with an empty remainder even though `smCoscheduledAlignment` is 8. The previous `148 % 8 == 0` assertion tested a guarantee CUDA 13.4 intentionally removed.

This change is entirely driver-side. `cuda.bindings` forwards the resource-split request and the driver-populated resource structures without calculating either value. The CUDA 13.4 changes to the relevant structures add `localityDomainId` within existing reserved or padded space; they preserve the ABI and do not affect this path. This is not a `cuda.bindings` regression.

### Fix

- Exercise the retained alignment guarantee by passing `coscheduled_sm_count=sm_resource.coscheduled_alignment` explicitly before checking the returned count.
- Gate the CUDA 13 structured-split behavior on driver version, and skip cleanly when structured split is unavailable in the build or the topology rejects the explicit resource configuration.
- Keep separate coverage for the CUDA 12 `SplitByCount` discovery path, which continues to return a group size aligned to its granularity.
- Add a `cuda.core` 1.2.0 release note explaining that unconstrained discovery on CUDA 13.4+ may span the full GPU and leave an empty remainder.


cuda_core/tests/test_green_context.py
```diff
@@ -405,11 +405,23 @@ def test_discovery_mode(self, sm_resource):
...
-    def test_discovery_respects_alignment(self, sm_resource):
-        groups, _ = sm_resource.split(SMResourceOptions(count=None))
+    def test_discovery_respects_explicit_coscheduled_sm_count(self, sm_resource):
+        """Constrain discovery explicitly because unconstrained discovery may use all SMs."""
+        if binding_version() < (13, 1, 0) or driver_version() < (13, 1, 0):
+            pytest.skip("explicit co-scheduled SM discovery requires CUDA 13.1+")
+
+        alignment = sm_resource.coscheduled_alignment
+        if alignment <= 0:
+            pytest.skip("device does not report a co-scheduled SM alignment")
```
leofang: This guard can't fire, and I'd drop it.

- Negative is impossible by construction: [`_coscheduled_alignment` is `unsigned int`](https://github.com/NVIDIA/cuda-python/blob/5e0631c8705ff02e8ce5798185b783793f586224/cuda_core/cuda/core/_device_resources.pxd#L14), so the property never returns a negative. `<= 0` is really `== 0`.
- Zero is impossible for a device-level resource, which is what `sm_resource` is. [`_from_dev_resource`](https://github.com/NVIDIA/cuda-python/blob/5e0631c8705ff02e8ce5798185b783793f586224/cuda_core/cuda/core/_device_resources.pyx#L433-L447) takes `res.sm.smCoscheduledAlignment` straight from `cuDeviceGetDevResource` on 13.x builds; on 12.x builds it comes from [`_sm_resource_granularity()`](https://github.com/NVIDIA/cuda-python/blob/5e0631c8705ff02e8ce5798185b783793f586224/cuda_core/cuda/core/_device_resources.pyx#L399-L411), a hardcoded 8 or 2. The 13.4 header documents the device-level value as 2 (CC 7.x/8.x/Tegra) or 8 (CC 9.0+), with `CUdevSmResource` marked OUTPUT-only.
- The same file already asserts this three times with no version gate: [L248](https://github.com/NVIDIA/cuda-python/blob/5e0631c8705ff02e8ce5798185b783793f586224/cuda_core/tests/test_green_context.py#L248), [L259](https://github.com/NVIDIA/cuda-python/blob/5e0631c8705ff02e8ce5798185b783793f586224/cuda_core/tests/test_green_context.py#L259) and [L265](https://github.com/NVIDIA/cuda-python/blob/5e0631c8705ff02e8ce5798185b783793f586224/cuda_core/tests/test_green_context.py#L265). Any configuration where this test runs also runs those — so if the guard ever fired, those three would already be failing, and this test would skip silently while they fail loudly.

I suspect it came from the neighbouring convention, where 0 *is* real: in the request struct `coscheduledSmCount == 0` means "unconstrained" on 13.4, and a split-*result* `CUdevSmResource` can come back with `smCoscheduledAlignment == 0` — which is why [`_from_split_resource` guards it with `_fallback_if_zero`](https://github.com/NVIDIA/cuda-python/blob/5e0631c8705ff02e8ce5798185b783793f586224/cuda_core/cuda/core/_device_resources.pyx#L458-L461). That convention belongs to split results and request params, not to a device query.

If you'd rather keep the invariant stated, `assert alignment > 0` would match `test_properties`; a `skip` hides it.

cuda_core/tests/test_green_context.py
```diff
@@ -405,11 +405,23 @@ def test_discovery_mode(self, sm_resource):
         assert len(groups) == 1
         assert groups[0].sm_count >= sm_resource.min_partition_size
 
-    def test_discovery_respects_alignment(self, sm_resource):
-        groups, _ = sm_resource.split(SMResourceOptions(count=None))
+    def test_discovery_respects_explicit_coscheduled_sm_count(self, sm_resource):
+        """Constrain discovery explicitly because unconstrained discovery may use all SMs."""
+        if binding_version() < (13, 1, 0) or driver_version() < (13, 1, 0):
+            pytest.skip("explicit co-scheduled SM discovery requires CUDA 13.1+")
```
leofang: Two things here.

First, the bindings half of this gate measures the wrong thing for this fix. Nothing that matters to this test changed on the header or bindings side between 13.3 and 13.4: `CU_DEV_SM_RESOURCE_GROUP_PARAMS` gained `localityDomainId` carved out of `reserved[]` (same size, and cuda.core zero-fills the struct either way), and `CUdevSmResource` gained the same field inside a union already padded to `RESOURCE_ABI_BYTES` — both ABI-compatible, neither on this path. The behavior difference is entirely driver-side, and `driver_version()` already covers it. So the bindings check isn't what makes this test meaningful.

Second, it isn't redundant either — it's incidentally guarding API *availability* rather than behavior, and that guard is load-bearing: on cuda-core-cu12 with a 13.4 driver, `driver_version()` passes but [`_can_use_structured_sm_split()`](https://github.com/NVIDIA/cuda-python/blob/5e0631c8705ff02e8ce5798185b783793f586224/cuda_core/cuda/core/_device_resources.pyx#L233-L245) returns False (compile-time `CUDA_CORE_BUILD_MAJOR == 12`), `split()` falls through to the ByCount path, and a non-`None` `coscheduled_sm_count` [raises `RuntimeError`](https://github.com/NVIDIA/cuda-python/blob/5e0631c8705ff02e8ce5798185b783793f586224/cuda_core/cuda/core/_device_resources.pyx#L254-L257).

Which suggests expressing the two concerns separately: keep `driver_version() < (13, 1, 0)` as the semantic gate, since that is the thing being tested, and handle availability by wrapping the `split()` call in `except RuntimeError: pytest.skip(...)`. That also avoids hand-rolling a second copy of `_can_use_structured_sm_split()`, which the current gate reproduces only partially — it drops both `has_sm_resource_split()` and the `CUDA_CORE_BUILD_MAJOR >= 13` check.

cuda_core/tests/test_green_context.py
```diff
@@ -405,11 +405,23 @@ def test_discovery_mode(self, sm_resource):
...
+        alignment = sm_resource.coscheduled_alignment
+        if alignment <= 0:
+            pytest.skip("device does not report a co-scheduled SM alignment")
+
+        groups, _ = sm_resource.split(
+            SMResourceOptions(
+                count=None,
+                coscheduled_sm_count=alignment,
+            )
+        )
```
leofang: With an explicit `coscheduledSmCount`, the header says the split returns `CUDA_ERROR_INVALID_RESOURCE_CONFIGURATION` if `smCount` ends up 0 or below `coscheduledSmCount`. Discovery on an unusual topology could hit that. The file already has the pattern ([`_is_invalid_resource_configuration`](https://github.com/NVIDIA/cuda-python/blob/5e0631c8705ff02e8ce5798185b783793f586224/cuda_core/tests/test_green_context.py#L101-L102) → skip, as used in [`_try_sm_split`](https://github.com/NVIDIA/cuda-python/blob/5e0631c8705ff02e8ce5798185b783793f586224/cuda_core/tests/test_green_context.py#L115-L121)); reusing it here would keep things consistent.

leofang (approved):
Confirmed the root cause against the raw headers: this is an intentional driver semantic change in CUDA 13.4, not a stale test. Diffing the `cuDevSmResourceSplit` documentation between the CUDA 13.2 and 13.4.46 `cuda.h`:

- CC 9.0+ guidance changed from *"The `smCount` must be a multiple of 8, or `coscheduledSmCount` if provided"* to *"The `smCount` must be a multiple of `coscheduledSmCount` if provided"*.
- *"The alignment (and default value of `coscheduledSmCount`) is 8"* → *"The alignment is 8"* — the parenthetical was dropped.
- New line: *"On all architectures, portable code should set `smCount` to a multiple of the device's alignment granularity (`sm.smCoscheduledAlignment`)"* — the responsibility moved to the caller.
- Most directly: the doc's own Example 1, all-zero `groupParams`, which is exactly the call the old test made, had its comment rewritten from *"Will discover how many SMs there are, that are co-scheduled in groups of `smCoscheduledAlignment`"* to *"Will discover how many SMs are available without imposing a co-scheduled group-size constraint."*

So 148 SMs with an empty remainder is the intended 13.4+ behavior, and asserting only the guarantee the API still makes is the right fix.

This is scoped to discovery mode (`count=None`), which is why the diff touches exactly one test: there the driver chooses the SM count, and 13.4 stopped constraining that choice. For explicit requests (`count=N`) the caller supplies the count and is responsible for aligning it, so the change doesn't apply — the explicit-split tests in this file are untouched and unaffected.

Three things I'd like reflected before merge:

1. The description says the alignment *"is not imposed unless the request supplies a co-scheduling constraint."* The header is more specific: the default is still applied internally (*"A `coscheduledSmCount` of 0 uses the default value internally while preserving 0 in `groupParams`"*) — what was dropped is the guarantee on the *returned* `smCount`. Worth stating precisely so nvbug 6464867 closes as "intentional driver semantic change, test updated" rather than "test bug".

2. This is a user-visible change for `SMResourceOptions(count=None)`: on 13.4+ it can claim every SM and leave an empty remainder, where 13.1-13.3 returned an aligned subset. Nothing in cuda.core's docs promised alignment here (I grepped docstrings, `docs/source`, and release notes — `count` only says "discovery mode (auto-detect)"), so no doc is wrong today, but a release note seems warranted. The `green_ctx` fixture now spans the whole GPU on 13.4+ as a side effect.

3. The old test also covered unconstrained-discovery alignment on 12.x via the ByCount path, where 12.x does still round up to granularity. That now skips entirely. Presumably deliberate — flagging that the 12.x rounding behavior is no longer covered anywhere.

-- Leo's bot

rwgk:
> The description says the alignment "is not imposed unless the request supplies a co-scheduling constraint."

Thanks - updated the analysis to distinguish the internally applied default from the returned-`smCount` guarantee. It now describes this as an intentional CUDA 13.4 driver semantic change that required a test update: `coscheduledSmCount=0` still selects the driver's default internally while remaining 0 in `groupParams`, but CUDA 13.4 no longer guarantees that the returned `smCount` is aligned. It also makes clear that this is not a `cuda.bindings` regression.

rwgk:
> 2. This is a user-visible change for `SMResourceOptions(count=None)`: on 13.4+ it can claim every SM and leave an empty remainder, where 13.1-13.3 returned an aligned subset.

Added a `cuda.core` 1.2.0 release note covering the CUDA 13.4 behavior, including the possible whole-GPU group and empty remainder, the internally applied default, and the need to set `coscheduled_sm_count` explicitly when an aligned result is required.

Implemented by 94c09e6efb2.

rwgk:
> 3. The old test also covered unconstrained-discovery alignment on 12.x via the ByCount path, where 12.x does still round up to granularity. That now skips entirely. Presumably deliberate — flagging that the 12.x rounding behavior is no longer covered anywhere.

Good catch. I restored a dedicated CUDA 12 `SplitByCount` discovery test, so its rounding-to-granularity behavior remains covered independently of the CUDA 13.1+ structured-split test.

Implemented by abf5fea8ffd.

cuda_core/tests/test_green_context.py
rwgk: Agreed. I removed the skip. `sm_resource` is a device-level resource and the suite already asserts that `coscheduled_alignment` is positive, so an invariant violation will fail visibly instead of being hidden.

Implemented by abf5fea8ffd.

cuda_core/tests/test_green_context.py
rwgk: Agreed. I kept only the driver-version gate for the structured-split semantics and moved API-availability handling to the `split()` call. A `RuntimeError` now skips configurations where the structured API is unavailable, without duplicating `_can_use_structured_sm_split()` in the test.

Implemented by abf5fea8ffd.

cuda_core/tests/test_green_context.py
rwgk: Agreed. The explicit split now reuses `_is_invalid_resource_configuration()` to skip the documented topology-dependent `CUDA_ERROR_INVALID_RESOURCE_CONFIGURATION`, while re-raising every other `CUDAError`.

Implemented by abf5fea8ffd.

PR #1898, opened 2026-04-13
Skip the tests when the requested NVML function is unavailable

rluo8:
Skip the tests when nvml function couldn't be found in the installed driver.  

FAILED tests/system/test_system_device.py::test_addressing_mode - cuda.bindings._internal.utils.FunctionNotFoundError: function nvmlDeviceGetAddressingMode is not found
FAILED tests/system/test_system_device.py::test_repair_status - cuda.bindings._internal.utils.FunctionNotFoundError: function nvmlDeviceGetRepairStatus is not found
FAILED tests/system/test_system_device.py::test_clock - cuda.bindings.nvml.NotFoundError: Not Found

No output when using driver 576.57. 
dumpbin /exports C:\Windows\System32\nvml.dll |findstr /i "nvmlDeviceGetAddressingMode nvmlDeviceGetRepairStatus"

After the fix, the tests could be skipped instead of failure.

collected 3 items

tests/system/test_system_device.py::test_clock SKIPPED (Unsupported call for device architecture B...)
tests/system/test_system_device.py::test_repair_status SKIPPED (Unsupported call for device archit...)
tests/system/test_system_device.py::test_addressing_mode SKIPPED (Unsupported call for device arch...)

PR #2390, opened 2026-07-20
test: skip system NVML tests when the symbol is absent in older bindings

rluo8:
## Description

This is to fix bug 6179497, and it's a test only issue.
Tests failed when using cuda.core v1.1.0 and cuda bindings 12.9.7.

  FAILED system/test_system_device.py::test_addressing_mode
         FunctionNotFoundError: function nvmlDeviceGetAddressingMode is not found
  FAILED system/test_system_device.py::test_repair_status
         FunctionNotFoundError: function nvmlDeviceGetRepairStatus is not found

This issue was originally fixed in pr #1898, but the fix in cuda.bindings wasn't inculded in cuda 12.9.x. 
This fix back-stops the skip on the cuda.core test side (independent of the cuda.bindings version): a missing NVML symbol is skipped rather than failed.

## Checklist


mdboom (changes requested):
`unsupported_before` is meant to indicate that a feature isn't available on a specific hardware architecture.  It's not meant to indicate that a function isn't available on a specific CTK version.

We should skip each of the failing tests individually, based on the CTK version.

For example:

```
@pytest.mark.skipif(get_user_mode_driver_version() < (13, 0))
```

leofang:
I think we should check the KMD version not UMD version?

rluo8:
Thanks @mdboom , @leofang. I updated the patch to skip the two cases when kernel mode driver < R580.  Verified on a forward compatibility system (UMD 13.0 + KMD 575), the tests would still fail with FunctionNotFoundError if using UMD check. After switching to use KMD version check, the two tests could be skipped correctly.

mdboom (approved):
Thanks, @rluo8.  LGTM.

mdboom (approved):

rluo8:
Hi @mdboom , the CI check have all passed now. It seems that there's something wrong with Github and it disabled the auto-merge. Could you please help merge this PR? 
Thanks!

rluo8:
Closing as no longer needed.
PR #1898 already made `unsupported_before` skip on a missing NVML symbol, but the helper shipped inside the cuda.bindings wheel and not included in version 12.9.7.
After #2384 moved `arch_check.py` into the standalone `cuda_python_test_helpers` package, which travels with the test suite rather than with the installed cuda.bindings. Both `test_addressing_mode` and `test_repair_status` now skip on their own.

PR #2009, opened 2026-05-01
demo: use pytest `subtests` in cuda.core system device loops to ensure full test coverage

rwgk:
Keep per-device skips and xfails from aborting the rest of the GPU system coverage.

Require pytest >= 9.0 in cuda.core test environments so the built-in subtests fixture is available.

To see how simple this change is, it's best to review with "Hide whitespace":

Screenshot 2026-05-01 at 15 51 57

___

```bash
git diff -b -w main -- cuda_core/tests/
```

```diff
@@ -341,8 +341,9 @@ def test_c2c_mode_enabled():


 @pytest.mark.skipif(helpers.IS_WSL or helpers.IS_WINDOWS, reason="Persistence mode not supported on WSL or Windows")
-def test_persistence_mode_enabled():
+def test_persistence_mode_enabled(subtests):
     for device in system.Device.get_all_devices():
+        with subtests.test(device_index=device.index):
             is_enabled = device.is_persistence_mode_enabled
             assert isinstance(is_enabled, bool)
             try:
@@ -615,14 +616,16 @@ def test_clock_event_reasons():
         assert all(isinstance(reason, system.ClocksEventReasons) for reason in reasons)


-def test_fan():
+def test_fan(subtests):
     for device in system.Device.get_all_devices():
         # The fan APIs are only supported on discrete devices with fans,
         # but when they are not available `device.num_fans` returns 0.
         if device.num_fans == 0:
+            with subtests.test(device_index=device.index):
                 pytest.skip("Device has no fans to test")

         for fan_idx in range(device.num_fans):
+            with subtests.test(device_index=device.index, fan_idx=fan_idx):
                 fan_info = device.get_fan(fan_idx)
                 assert isinstance(fan_info, _device.FanInfo)

@@ -655,8 +658,9 @@ def test_fan():
                     fan_info.set_default_speed()


-def test_cooler():
+def test_cooler(subtests):
     for device in system.Device.get_all_devices():
+        with subtests.test(device_index=device.index):
             # The cooler APIs are only supported on discrete devices with fans,
             # but when they are not available `device.num_fans` returns 0.
             if device.num_fans == 0:
```

PR #2391, opened 2026-07-20
Use subtests where appropriate everywhere

mdboom:
This is carrying out the demo in #2009 to use subtests everywhere.


rwgk:
Hi @mdboom - I asked Codex to turn all of the findings from its review of #2391 into concrete fixes in the experimental draft PR #2479. My plan was to review the changes myself first, then offer the individual commits as possible cherry-picks rather than add a long list of abstract suggestions here. Please feel free to cherry-pick, adapt, or discard each item independently.

The complete set passed [this CI run](https://github.com/NVIDIA/cuda-python/actions/runs/30758661345?pr=2479). That run was before I reverted 13440641046ca53b951b55ca5d495e7810a44ef8. I pushed the revert afterward only to make it clear that I do not want `ci/` to depend on code in `toolshed/`; it was not a response to a CI failure.

## The three changes I looked at more closely

### 1. Recognize native pytest subtest outcomes in the skipped-test report

Commit e7f801734922bff8fd0ef5d8185a239cd4d3abfe teaches `toolshed/find_skipped_tests.py` about `SUBPASSED`, `SUBFAILED`, `SUBSKIPPED`, and the other native subtest outcomes.

The important case is an all-skipped subtest: pytest emits one or more `SUBSKIPPED` lines and then reports the parent test as `PASSED`. Treating that final parent pass as evidence that the test executed would incorrectly remove it from the "always skipped" report. The change tracks whether subtest outcomes were seen, ignores the synthetic parent pass in that case, and still treats a mixed skipped/passed set as having executed.

There is one test-ownership question here. The parser belongs to `toolshed/`, and I think `ci/` should remain independent of it. Commit 13440641046ca53b951b55ca5d495e7810a44ef8 moved the small regression test into `ci/tools/tests` so normal CI would collect it; I reverted that move in d6376d52eec665d1f8782bd2d8104d3cb5f69c6d, leaving the test back under `toolshed/` and outside normal CI collection.

My inclination is to delete this small regression test. If you think it is worth retaining, I think it should live under `toolshed/tests/`; we could then decide whether `ci-nightly.yml` should explicitly run:

```bash
pytest -v --noconftest toolshed/tests
```

I would be interested in which of those two directions you prefer. The parser fix itself is independent of that choice.

### 2. Use the documented Fermi boundary for the power-limit getter

Commit 57e1a4c42d1e9b523277cccca90e1d41bee36bfb changes only the getter test from Kepler to Fermi.

The native NVML documentation for [`nvmlDeviceGetPowerManagementLimit`](https://docs.nvidia.com/deploy/nvml-api/group__nvmlDeviceQueries.html) says the getter is supported on Fermi or newer fully supported devices. The nearby default-limit and constraints queries, and the setters, are documented as Kepler or newer. Using the setter's Kepler boundary for the getter would therefore skip a valid Fermi case.

### 3. Use readable device labels in subtest reports

Commit 978d282d769b0c31bb440b0412123c7ccb28d697 changes report metadata such as:

```text
SUBSKIPPED(device=<cuda.core.system._device.Device object at 0x...>)
SUBSKIPPED(device=125060957155008)
```

to:

```text
SUBSKIPPED(device_index=0)
```

`subtests.test()` uses its keyword arguments as report descriptions, not as parametrization. Python still evaluates `device.index` before entering the context, just as it would any argument; after that lookup, the value only controls the description. It does not create a different pytest node ID or replace the outer `device` that the test body uses. I think the index is better metadata because object addresses and raw NVML handles are opaque and process-specific, while `device_index=0` is concise and directly comparable across jobs. The index is not intended as a persistent hardware identity across reboots; the improvement is clearer, repeatable CI output.

## The remaining findings

### 4. Give every fan its own subtest

Commits 0d565c1a943b5f9cfc81a48fe2b125baf9fcc630 and 3e8d151f6a5d1ad508db506ed8416bf315eb1e82 are best considered together.

There is a concrete coverage bug in the original fan loop: it obtains each fan in turn, but the assertions and setter/getter checks are outside the loop, so they run only once against the final fan. The first commit moves the complete check inside the loop and gives each `(device, fan_index)` pair its own subtest, so every fan is actually exercised and one fan's result does not suppress the others. The follow-up gives the `num_fans` query its own per-device subtest, so an unsupported or zero-fan device is reported for that device without preventing later devices from being considered.

### 5. Isolate independent inner loop cases

Commit 69009b041092d158f0d0d1025bb898e628c1cdc7 applies the same principle more broadly: a broad per-device subtest does not provide completeness when several independent, potentially skipping cases are nested inside it.

It splits or flattens the relevant boundaries for affinity scopes/APIs, CPU-affinity results, compute modes, clock types, temperature thresholds and sensors, and NVLink links/APIs. A skip or failure in one independent item is then reported for that item while the remaining items can still run. This seems like a useful middle ground: keep the loops simple, but put the subtest boundary around the unit of coverage we actually want to preserve.

### 6. Narrow the cooler unsupported guard

Commit 07e493b68939fb11b7dc07eb0fe2add75ebd4373 limits `unsupported_before(...)` to the `device.cooler` lookup it is intended to guard.

Keeping the subsequent type checks and `signal_type`/`target` property validation outside that guard prevents an unexpected `NotSupported` from those later operations from being converted into a skip. In other words, known architectural absence remains a skip, while a regression in the returned object remains visible as a failure.

## Cherry-pick notes

- 0d565c1a94 and 3e8d151f6a belong together.
- 2adcc74eb5518a9b7bfc1f297d08fbb935af9691 is only the Ruff `SIM117` cleanup needed after the related context-manager changes; it is not a separate behavioral proposal.
- 978d282d76 touches many of the same subtest lines as the other changes, so it will likely apply most cleanly after whichever behavioral commits you choose.
- 1344064104 and d6376d52ee record the test-location experiment and its revert; they cancel each other and neither is intended for cherry-picking.
- The merge commits in #2479 are branch maintenance and are not intended for cherry-picking.

If you want the complete set, the original order of the proposed commits is:

```text
0d565c1a94 e7f8017349 69009b0410 57e1a4c42d 07e493b689 3e8d151f6a 978d282d76 2adcc74eb5
```

mdboom:
> 1. Recognize native pytest subtest outcomes in the skipped-test report

I think updating the skipped test finder makes sense.  I think adding testing to `toolshed` utilities is out-of-scope here, though.  If we want to do that, let's do it as follow-on work.

> 2. Use the documented Fermi boundary for the power-limit getter

I don't think we should change any of the limits from the existing code on `main`.  The documentation about when features are available is largely wrong, and most of these limits were discovered through trial-and-error by submitting to QA.  I worry that messing them will cause unnecessary cycles with QA again.

3, 4, 5, and 6 seem fine.

Why not just push the changes to this PR and I will give it a final review?

rwgk:
Hi @mdboom — thanks for the detailed reply on the suggestions from #2479.

I have now applied the agreed work directly to this branch. I used #2479 as a scratch area, reviewed each proposed change, and then cherry-picked or adapted the useful parts here after merging the current `main`. I kept the logical changes separate so they are easy to inspect, adjust, or drop independently.

### Merge from `main`

- b9ecde4b6501ba7fb96a0b17231cf832e486f6af merges the current `main` into this branch before applying the follow-up changes. The conflict resolutions keep this PR's pytest 9 minimum, `main`'s new `cuda-python-test-helpers` dependency and platform helpers, `main`'s newer system-device tests, and `main`'s removal of the obsolete `NVLINK_MAX_LINKS` test.

### Follow-up commits

- 910462a2cd16cb78bbb59674fe920433c9c7f3b7 gives every physical fan its own subtest, so a failure for one fan does not prevent the remaining fans from being exercised.

- 9dd3c7067775e3e9afe7b84a0754fa4bd010553b teaches `toolshed/find_skipped_tests.py` about pytest's `SUB*` outcomes. In particular, a parent reported as `PASSED` no longer hides the fact that all of its subtests skipped. Following your preference, the final branch does not add a toolshed test under `ci/` (or elsewhere).

- b7f469c98a44f08baa1a439279b70a15fe01d80c separates independent inner cases — including compute modes, affinity APIs/scopes, clocks, temperature queries, and NVLink queries — so one skip or failure does not truncate unrelated coverage.

- 35ca79949edb5a94428e6a3a9d526708faf9b50a narrows the cooler `unsupported_before` scope to the API call itself, leaving the result validation outside that context.

- 7ccd34e34e140b84ac4652a975fb586fb7db05f2 queries and validates the fan count in its own subtest. A failure there remains local to that device instead of aborting the rest of the test.

- 0f67e6b45d3491466034cdb0ec21283c98eec5d2 replaces raw device objects/handles in subtest metadata with scalar device indices. A raw object or handle representation can include a process-specific address; `device_index=0` is concise, searchable, and much easier to compare across jobs.

- a014155ba114ab8d3669d3e4c37aa823d35ae594 combines adjacent context managers when they intentionally cover the same region.

- 2adfe8ea949a5a5d9d662a6f5bdea45f2620ef02 preserves the support behavior already present on `main`: the power-limit getter keeps `unsupported_before(device, None)`, while the setter remains guarded as Kepler-or-newer.

- e08703805405043613806f39f2a0682965a821a9 adds an explicit Kepler guard to the newly independent memory-affinity subtest. Before the split it implicitly stopped when the preceding CPU-affinity call skipped; after the split it needs its own guard.

- a3dc33db70060a763d6a534e16935d791281e8bc validates temperature, thermal-settings, and NVLink-count results before assigning them to values consumed outside their subtests. This prevents a failed type assertion from leaking an invalid value into later cases.

### Validation

- A local clean CUDA 13.3 build completed successfully.
- The full test run passed for pathfinder, both bindings configurations, bindings Cython tests, and core Cython tests.
- The core Python run had one expected failure (due to my local driver), everything else passed.
- After the last two audit fixes, the focused affinity/temperature/NVLink selection passed: 8 tests passed, 8 architecture-dependent subtests skipped, and 204 subtests passed.
- The parser smoke checks and all non-network pre-commit hooks pass.

Please feel free to adapt or discard any individual commit if you prefer a different shape.

rwgk (approved):
LGTM. All tests pass, except the pixi smoke test, which has been failing for a few days already is is definitely unrelated.

rwgk:
> LGTM. All tests pass, except the pixi smoke test, which has been failing for a few days already is is definitely unrelated.

I just merged main, which should resolve the pixi smoke test failure (see PR #2532).

PR #2392, opened 2026-07-20
Get tracebacks from CI (on Linux)

mdboom:
This will run the tests inside of gdb on Linux, so if a test segfaults (or sends some other signal), we will get a traceback in the logs.


mdboom:
Marking as a draft because it seems to fail on Python 3.12.

mdboom:
Current issues with this approach:

- Unexplained breakage on Python 3.12
- We don't use debug builds of Python, so tracebacks are less useful than they could be
- Using debug builds might have other side effects in the non-segfaulting case, so not clear there's a way forward.

Given all that, closing this as a dead experiment, but if someone wants to move forward with a better solution, feel free.

PR #2393, opened 2026-07-20
Make Windows pathfinder dynamic library searches architecture-aware

isVoid:
## What changed

This PR makes `cuda-pathfinder` dynamic-library discovery architecture-aware on Windows, including the CUDA Toolkit 13.4 Windows-on-Arm layouts.

### Architecture-aware discovery and loading

- Add the immutable `WindowsSearchDirs` descriptor type, with independent ordered `x64` and `arm64` paths for wheel and local-toolkit discovery.
- Record `supported_windows_arch` per library descriptor and select Windows paths using the Python process architecture rather than the host architecture.
- Prefer CUDA 13 aggregate-wheel directories (`nvidia/cu13/bin/x86_64` and `nvidia/cu13/bin/arm64`), while retaining CUDA 12 component-wheel locations as x64-only fallbacks.
- Use `SUPPORTED_NVIDIA_LIBNAMES` as the public current-platform/current-architecture CTK projection. Loader availability uses the internal `ALL_AVAILABLE_LIBNAMES` set so supported third-party and driver libraries remain loadable.
- Add the public `UnsupportedArchError` for unsupported Windows Python platform tags.
- Preserve the legacy unsuffixed Windows lookup tables as x64 compatibility aliases and add explicit x64 and ARM64 projections.

### Special library layouts

- **CUPTI:** Search `extras/CUPTI/lib/x64`, legacy `extras/CUPTI/lib64`, then `bin` for x64; search only `extras/CUPTI/lib/arm64` for ARM64.
- **cuDLA:** Treat Windows cuDLA as ARM64-only and search only the local-toolkit `bin/arm64` location; no Windows wheel currently provides `cudla.dll`.
- **cuSPARSELt:** Use the CUDA 13 architecture-specific aggregate-wheel paths and retain the CUDA 12 component-wheel location for x64 only.
- **NVVM:** Search `nvvm/bin/x64` followed by legacy `nvvm/bin` for x64, and `nvvm/bin` for ARM64. Because `nvvm/bin` contains x64 NVVM before CTK 13.4 but ARM64 NVVM in CTK 13.4 Windows-on-Arm, validate the candidate DLL's PE/COFF `Machine` field with `windows_pe_matches_arch` before accepting it.

### Supporting changes

- Extend search-step handling and CTK-root discovery for architecture-specific Windows directories.
- Derive legacy lookup tables from the descriptor catalog and expand tests for descriptors, search steps, CTK roots, loaders, supported library names, and PE architecture validation.
- Remove the obsolete catalog writer and generated-table update scripts; retain the site-packages collection scripts and author the Windows mappings directly in `descriptor_catalog.py`.
- Restore the 1.6.0 notes to their `main` contents and document these changes in the cuda-pathfinder 1.6.1 release notes.

## Why

CUDA Toolkit 13.4 installs Windows binaries in architecture-specific directories such as `bin/x64` and `bin/arm64`. CUDA 13 wheels likewise use architecture-specific directories under `nvidia/cu13/bin`, while CUDA 12 component wheels retain their existing package-specific layouts. Windows installers can contain cross-architecture packages, so pathfinder must choose and validate libraries using the Python process architecture.

## Validation

- Full cuda-pathfinder suite on Windows: 1,243 passed, 8 skipped.
- Full cuda-pathfinder suite on Linux: 1,280 tests collected with a successful exit.
- `ruff check` and `ruff format --check` passed for the changed Python files.


cuda_pathfinder/cuda/pathfinder/_dynamic_libs/supported_nvidia_libs.py
```diff
@@ -60,11 +60,16 @@
 }
 SITE_PACKAGES_LIBDIRS_LINUX = SITE_PACKAGES_LIBDIRS_LINUX_CTK | SITE_PACKAGES_LIBDIRS_LINUX_OTHER
 
+# Historical table exports represent the original x64 catalog.
```
isVoid: Q: What should we list here? Historically it's only listing the x64 catalog. But should we instead list all available items?

cuda_pathfinder/cuda/pathfinder/_utils/windows_arch.py
```diff
@@ -0,0 +1,30 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+import sysconfig
+
+
+class UnsupportedWindowsArchError(RuntimeError):
```
leofang: We can just call it `UnsupportedArchError` in case we want to use it elsewhere.

rwgk:
```
Ultimately, the only thing that matters is that `SUPPORTED_LIBNAMES` in `cuda/pathfinder/__init__.py` correctly reflects what libs are
supported on the current platform.

To define the boundary conditions: we could completely change everything else here and adjust the unit tests accordingly.

I don't want to go to that extreme, but keep this in mind with the goal to cleanly accommodate WoA, without making the code in cuda_pathfinder/cuda/pathfinder/_dynamic_libs/supported_nvidia_libs.py awkward.

So the questions become:
  * What is the `SUPPORTED_LIBNAMES` we are aiming for?
  * What is a healthy way to achieve that?

Could you please analyze and let me know your recommendation?
```

```
My recommendation is to make SUPPORTED_NVIDIA_LIBNAMES a capability list for the current process ABI—not an OS-wide union and not an
inventory of installed files.

Target Contract

“Current platform” should mean:

- Linux, following the existing behavior.
- Windows/x64 when the running Python reports win-amd64, including x64 Python emulated on WoA.
- Windows/ARM64 when Python reports win-arm64.

“Supported” should mean Pathfinder knows how to load that library for the process ABI:

- Supported but not installed → present in SUPPORTED_NVIDIA_LIBNAMES; loading raises DynamicLibNotFoundError.
- Known but unsupported on this OS/architecture → absent from the tuple; loading raises DynamicLibNotAvailableError without searching.
- Installed wheels, conda, CUDA_PATH, and the filesystem must not affect the tuple.

For this PR, I recommend preserving the established CTK-only boundary of the public tuple. Expanding it to third-party and driver libraries
would be a separate API-semantic change, although the loader’s broader internal supported set should use the same platform predicate.

Target Sets

Based on the available CTK and prior-PR evidence, I would aim for:

 Process platform    CTK names
━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 Linux               Existing 29-name Linux tuple
──────────────────  ─────────────────────────────────────────────────
 Windows x64         Existing 28-name Windows tuple minus cudla
──────────────────  ─────────────────────────────────────────────────
 Windows ARM64       Existing 28-name Windows tuple, including cudla

The 27-name Windows common set is therefore:

cudart, nvfatbin, nvJitLink, nvrtc, nvvm,
cublas, cublasLt, cufft, cufftw, curand,
cusolver, cusolverMg, cusparse,
nppc, nppial, nppicc, nppidei, nppif, nppig,
nppim, nppist, nppisu, nppitc, npps,
nvblas, nvjpeg, cupti

I would not copy PR 440’s nvjpeg ARM64-only classification. NVIDIA’s current CUDA component matrix lists nvJPEG for x86-64, ARM64, and
Windows, and the current NVIDIA nvJPEG package includes a Windows x86-64 wheel. CUDA release matrix
(https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/), NVIDIA nvJPEG package (https://pypi.org/project/nvidia-nvjpeg/).

cudla is the one boundary I would confirm with Kitmaker before freezing the expected tuple. The private history consistently describes its
Windows addition as ARM64-specific. If an x64 Windows cudla.dll is officially supported, both Windows tuples would instead contain the same
28 names.

Healthy Implementation

I would keep architecture policy out of cuda_pathfinder/cuda/pathfinder/_dynamic_libs/supported_nvidia_libs.py:16. That module should remain
a compatibility projection, not become the place where OS, architecture, paths, and package history are recombined.

A relatively small clean design would be:

1. Add explicit descriptor availability, such as windows_supported_arches.
2. Reshape Windows paths by architecture at the collection level:

@dataclass(frozen=True)
class WindowsSearchDirs:
    x64: tuple[str, ...] = ()
    arm64: tuple[str, ...] = ()

This is healthier than pairing every individual x64 path with an ARM64 path. It naturally supports different path counts and x64-only
entries without fake “common” paths.

3. Add one pure registry query:

supported_libnames(platform, *, packaged_with=None)

4. Use that same query for both:

SUPPORTED_NVIDIA_LIBNAMES = supported_libnames(
    current_process_platform(),
    packaged_with="ctk",
)

_ALL_SUPPORTED_LIBNAMES = frozenset(
    supported_libnames(current_process_platform())
)

This prevents the current divergence between the public CTK tuple and cuda_pathfinder/cuda/pathfinder/_dynamic_libs/
load_nvidia_dynamic_lib.py:45.

5. Keep legacy SITE_PACKAGES_LIBDIRS_WINDOWS* constants as explicitly x64 compatibility projections—or introduce clearly named _X64/_ARM64
   projections. Never concatenate the two architectures.

The tests should assert the exact three capability matrices, nvjpeg on x64, the final cudla decision, unavailable-versus-missing exception
behavior, dependency closure per architecture, and catalog-writer round-tripping. Tests comparing one catalog-derived legacy table to
another catalog-derived table do not establish the desired contract.

I based this on the local branch, the private PR/issue history, and current public CUDA metadata. The internal Glean connector failed during
this analysis, so the final CTK 13.4 cudla x64/ARM64 boundary deserves explicit confirmation.
```

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/descriptor_catalog.py
```diff
@@ -304,9 +371,6 @@ class DescriptorSpec:
         site_packages_linux=("nvidia/cu13/lib",),
         # No Windows pip wheel ships cudla.dll today; it is loaded from the local
         # CUDA Toolkit only, so site_packages_windows is intentionally left empty.
-        # The Windows CUDA Toolkit ships cudla.dll under per-architecture bin
-        # subdirs (e.g. bin/arm64 on N1X); search those ahead of the defaults.
-        anchor_rel_dirs_windows=("bin/arm64", "bin/x64", "bin"),
```
isVoid, line 309: Shouldn't we use `WindowsSearchDir` here to specialize search paths on WoA? Also only `arm64` exists on CTK13.4.

isVoid: Done.

rwgk (approved):
@isVoid I added two simple commits and reran the CI. LGTM, thanks!

Could you please make a pass updating the PR Description before you merge?

isVoid:
PR description is updated. Thanks @rwgk !

Issue #2353, opened 2026-07-13
cuda.core: add append-only attachment ownership for executable graphs


Andy-Jost:
## Background

Part of #1330. Depends on #2350.

CUDA propagates user objects from a `CUgraph` when it creates or updates a `CUgraphExec`, but provides no API for directly retaining or releasing user objects on an existing executable graph. Individual executable-node updates may introduce new callbacks, buffers, events, kernels, or argument owners whose lifetimes must extend through future launches.

Add one exec-private, append-only attachment accumulator that CUDA retains as a user object. Whole-graph updates replace it with a fresh accumulator, providing a safe reclamation boundary.

## Design

```text
GraphExecBox
  CUgraphExec
  CUuserObject attachment_object     # non-owning handle
  ExecAttachments* attachments       # non-owning pointer

ExecAttachments
  vector<OpaqueHandle>
```

The CUDA user object directly owns `ExecAttachments`. `GraphExecBox` stores a raw pointer because its `CUgraphExec` guarantees the user object remains alive while the box is usable. The box must not access the pointer after calling `cuGraphExecDestroy`.

The exec does not retain its source `CUgraph`. Source identity and node-view lifetime are part of #2354. This preserves normal CUDA behavior: after instantiation or whole-graph update returns, callers may destroy the source graph without affecting the executable.

## Instantiation

Every managed instantiation creates a unique accumulator:

1. Allocate an empty `ExecAttachments` and create its user object.
2. Temporarily retain that user object on the source `CUgraph`.
3. Instantiate the graph; CUDA propagates the reference to the new `CUgraphExec`.
4. Release the source graph's temporary reference.
5. Store the user-object handle and accumulator pointer in `GraphExecBox`.

If attachment or instantiation fails, release the temporary graph reference and destroy any partially created exec. Multiple execs instantiated from one definition receive independent accumulators.

## Whole-graph update

Every `Graph.update(source)` uses a fresh accumulator:

1. Create a new empty accumulator user object.
2. Temporarily retain it on the update source graph.
3. Call `cuGraphExecUpdate`.
4. On success, replace the box's accumulator handle and pointer.
5. Release the source graph's temporary reference.

CUDA replaces the exec's old user-object set with the source graph's objects and the fresh accumulator. It retains the old accumulator until all earlier launches finish, then destroys it asynchronously.

If the update fails, release the fresh accumulator and leave the exec and its existing accumulator unchanged.

## Internal append transaction

[...]

PR #2395, opened 2026-07-20
cuda.core: add graph definition node updates

### LINKED ISSUE #2352: cuda.core: add atomic update methods for graph definition nodes

Andy-Jost:
## Summary

Add subclass-specific `update()` methods for mutable graph definition nodes. The new API supports kernel, memcpy, memset, child-graph, event-record, event-wait, and host-callback nodes while leaving inspection properties read-only.

Updates replace one complete CUDA parameter structure through `cuGraphNodeSetParams`. Optional keyword arguments preserve existing values where coherent, coupled values such as kernels and arguments are replaced together, and existing executable graphs continue using their previous parameters and retained resources. This feature requires CUDA driver and `cuda.bindings` versions 12.2 or newer.

## Changes

- Add a shared graph-definition node update path with version checks, zero-initialized generic parameters, and failure-aware attachment prepare/commit handling.
- Add `update()` methods for `KernelNode`, `MemcpyNode`, `MemsetNode`, `ChildGraphNode`, `EventRecordNode`, `EventWaitNode`, and `HostCallbackNode`.
- Make partial kernel, memcpy, and memset updates keyword-only; changing a kernel requires explicitly supplying its arguments, including `args=()` for a no-argument kernel.
- Preserve or replace retained kernels, arguments, buffers, explicit pointer owners, events, callbacks, and user data in step with the corresponding CUDA parameter update.
- Stage, rekey, and publish replacement child-graph hierarchy metadata while invalidating views backed by the old embedded clone.
- Add extensible parameterized behavior/failure tests, focused child-hierarchy lifetime tests, public API documentation, and release notes.

## Review guide

1. Review `cuda_core/cuda/core/graph/_subclasses.pyx` and `_subclasses.pyi` for the public update signatures and node-specific parameter construction.
2. Review `_set_definition_node_params` for the common CUDA 12.2 gate and attachment transaction ordering.
3. Review `resource_handles.cpp/.hpp` for staged child-hierarchy metadata replacement and shared clone/rekey primitives.
4. Review `test_graph_node_update.py` for old-versus-new executable behavior and failure preservation, then `test_graph_definition_lifetime.py` for child-view invalidation and attachment lifetime coverage.

## Related Work

- Closes #2352.
- Builds on the graph attachment ownership foundation from #2357.
- Executable-graph node updates remain tracked by #2353.
- Part of #1330.

### DIFF
--- a/cuda_core/cuda/core/graph/_subclasses.pyx
+++ b/cuda_core/cuda/core/graph/_subclasses.pyx
@@ -481,12 +481,12 @@ cdef class MemsetNode(GraphNode):
 
     def update(
         self,
+        *,
         dst: Buffer | int | None = None,
         value=None,
         width: int | None = None,
         height: int | None = None,
         pitch: int | None = None,
-        *,
         dst_owner=None,
     ) -> None:
         """Replace selected memset parameters.
@@ -640,10 +640,10 @@ cdef class MemcpyNode(GraphNode):
 
     def update(
         self,
+        *,
         dst: Buffer | int | None = None,
         src: Buffer | int | None = None,
         size: int | None = None,
-        *,
         dst_owner=None,
         src_owner=None,
     ) -> None:
--- a/cuda_core/cuda/core/graph/_subclasses.pyx
+++ b/cuda_core/cuda/core/graph/_subclasses.pyx
@@ -8,6 +8,7 @@ from __future__ import annotations
 
 from libc.stddef cimport size_t
 from libc.stdint cimport uintptr_t
+from libc.string cimport memset as c_memset
 
 from cuda.bindings cimport cydriver
 
@@ -19,14 +20,18 @@ from cuda.core.graph._graph_node cimport GraphNode
 from cuda.core._resource_handles cimport (
     EventHandle,
     GraphHandle,
-    KernelHandle,
     GraphNodeHandle,
+    KernelHandle,
+    OpaqueHandle,
+    PreparedAttachment,
     as_cu,
     as_intptr,
-    create_event_handle_ref,
     create_child_graph_handle,
+    create_event_handle_ref,
     create_kernel_handle_ref,
+    graph_commit_attachment,
     graph_node_get_graph,
+    graph_prepare_attachment,
 )
 from cuda.core._utils.cuda_utils cimport HANDLE_RETURN
 
@@ -56,6 +61,23 @@ __all__ = [
 
 cdef bint _has_cuGraphNodeGetParams = False
 cdef bint _version_checked = False
+
+
+cdef void _set_definition_node_params(
+        const GraphNodeHandle& h_node,
+        cydriver.CUgraphNodeParams* params,
+        OpaqueHandle owner0,
+        OpaqueHandle owner1=OpaqueHandle()) except *:
+    cdef GraphHandle h_graph = graph_node_get_graph(h_node)
+    cdef cydriver.CUgraphNode node = as_cu(h_node)
+    cdef PreparedAttachment prepared
+
+    HANDLE_RETURN(graph_prepare_attachment(
+        h_graph, owner0, owner1, &prepared))
+    with nogil:
+        HANDLE_RETURN(cydriver.cuGraphNodeSetParams(node, params))
+    HANDLE_RETURN(graph_commit_attachment(prepared, node))
+
 
 cdef bint _check_node_get_params():
     global _has_cuGraphNodeGetParams, _version_checked
@@ -515,6 +537,19 @@ cdef class EventRecordNode(GraphNode):
     def __repr__(self) -> str:
         return (f"<EventRecordNode handle=0x{as_intptr(self._h_node):x}"
                 f" event=0x{as_intptr(self._h_event):x}>")
+
+    def update(self, event: Event) -> None:
+        """Replace the event recorded by this node."""
+        cdef OpaqueHandle event_owner = event._h_event
+        cdef cydriver.CUgraphNodeParams params
+
+        c_memset(&params, 0, sizeof(params))
+        params.type = cydriver.CU_GRAPH_NODE_TYPE_EVENT_RECORD
+        params.eventRecord.event = as_cu(event._h_event)
+
+        _set_definition_node_params(
+            self._h_node, &params, event_owner)
+        self._h_event = event._h_event
 
     @property
     def event(self) -> Event:
--- a/cuda_core/cuda/core/graph/_subclasses.pyx
+++ b/cuda_core/cuda/core/graph/_subclasses.pyx
@@ -30,15 +30,18 @@ from cuda.core._resource_handles cimport (
     KernelHandle,
     OpaqueHandle,
     PreparedAttachment,
+    PreparedChildGraphUpdate,
     as_cu,
     as_intptr,
     create_child_graph_handle,
     create_event_handle_ref,
     create_kernel_handle_ref,
     graph_commit_attachment,
+    graph_commit_child_graph_update,
     graph_get_attachment,
     graph_node_get_graph,
     graph_prepare_attachment,
+    graph_prepare_child_graph_update,
     make_opaque_py,
 )
 from cuda.core._utils.cuda_utils cimport HANDLE_RETURN, _parse_fill_value
@@ -75,11 +78,7 @@ cdef bint _has_cuGraphNodeGetParams = False
 cdef bint _version_checked = False
 
 
-cdef void _set_definition_node_params(
-        const GraphNodeHandle& h_node,
-        cydriver.CUgraphNodeParams* params,
-        OpaqueHandle owner0,
-        OpaqueHandle owner1=OpaqueHandle()) except *:
+cdef void _require_graph_node_update_support() except *:
     cdef tuple version = cy_driver_version()
     if version < (12, 2, 0):
         raise RuntimeError(
@@ -92,6 +91,14 @@ cdef void _set_definition_node_params(
             "Graph node mutation requires cuda.bindings 12.2 or newer; "
             f"using cuda.bindings version {'.'.join(map(str, version))}"
         )
+
+
+cdef void _set_definition_node_params(
+        const GraphNodeHandle& h_node,
+        cydriver.CUgraphNodeParams* params,
+        OpaqueHandle owner0,
+        OpaqueHandle owner1=OpaqueHandle()) except *:
+    _require_graph_node_update_support()
 
     cdef GraphHandle h_graph = graph_node_get_graph(h_node)
     cdef cydriver.CUgraphNode node = as_cu(h_node)
@@ -748,6 +755,37 @@ cdef class ChildGraphNode(GraphNode):
     def __repr__(self) -> str:
         return (f"<ChildGraphNode handle=0x{as_intptr(self._h_node):x}"
                 f" child=0x{as_intptr(self._h_child_graph):x}>")
+
+    def update(self, child: GraphDefinition) -> None:
+        """Replace the embedded graph with a clone of ``child``.
+
+        ``child`` must belong to an independent graph hierarchy.
+        """
+        cdef GraphHandle h_parent = graph_node_get_graph(self._h_node)
+        cdef GraphHandle h_replacement
+        cdef cydriver.CUgraphNode node = as_cu(self._h_node)
+        cdef cydriver.CUgraphNodeParams params
+        cdef cydriver.CUresult commit_status
+        cdef PreparedChildGraphUpdate prepared
+
+        _require_graph_node_update_support()
+        c_memset(&params, 0, sizeof(params))
+        params.type = cydriver.CU_GRAPH_NODE_TYPE_GRAPH
+        params.graph.graph = as_cu(child._h_graph)
+
+        HANDLE_RETURN(graph_prepare_child_graph_update(
+            h_parent, self._h_child_graph, node,
+            child._h_graph, &prepared))
+        with nogil:
+            HANDLE_RETURN(cydriver.cuGraphNodeSetParams(
+                node, &params))
+        try:
+            commit_status = graph_commit_child_graph_update(
+                prepared, &h_replacement)
+        finally:
+            if h_replacement:
+                self._h_child_graph = h_replacement
+        HANDLE_RETURN(commit_status)
 
     @property
     def child_graph(self) -> GraphDefinition:
--- a/cuda_core/cuda/core/graph/_subclasses.pyx
+++ b/cuda_core/cuda/core/graph/_subclasses.pyx
@@ -34,6 +34,7 @@ from cuda.core._resource_handles cimport (
     graph_prepare_attachment,
 )
 from cuda.core._utils.cuda_utils cimport HANDLE_RETURN
+from cuda.core._utils.version cimport cy_binding_version, cy_driver_version
 
 from cuda.core.graph._host_callback cimport (
     _is_py_host_trampoline,
@@ -71,6 +72,19 @@ cdef void _set_definition_node_params(
         cydriver.CUgraphNodeParams* params,
         OpaqueHandle owner0,
         OpaqueHandle owner1=OpaqueHandle()) except *:
+    cdef tuple version = cy_driver_version()
+    if version < (12, 2, 0):
+        raise RuntimeError(
+            "Graph node mutation requires CUDA driver 12.2 or newer; "
+            f"using driver version {'.'.join(map(str, version))}"
+        )
+    version = cy_binding_version()
+    if version < (12, 2, 0):
+        raise RuntimeError(
+            "Graph node mutation requires cuda.bindings 12.2 or newer; "
+            f"using cuda.bindings version {'.'.join(map(str, version))}"
+        )
+
     cdef GraphHandle h_graph = graph_node_get_graph(h_node)
     cdef cydriver.CUgraphNode node = as_cu(h_node)
     cdef PreparedAttachment prepared
--- a/cuda_core/cuda/core/graph/_subclasses.pyx
+++ b/cuda_core/cuda/core/graph/_subclasses.pyx
@@ -13,6 +13,7 @@ from libc.string cimport memset as c_memset
 from cuda.bindings cimport cydriver
 
 from cuda.core._event cimport Event
+from cuda.core._kernel_arg_handler cimport ParamHolder
 from cuda.core._launch_config cimport LaunchConfig
 from cuda.core._memory._buffer cimport Buffer
 from cuda.core._module cimport Kernel
@@ -38,6 +39,7 @@ from cuda.core._resource_handles cimport (
     graph_get_attachment,
     graph_node_get_graph,
     graph_prepare_attachment,
+    make_opaque_py,
 )
 from cuda.core._utils.cuda_utils cimport HANDLE_RETURN, _parse_fill_value
 from cuda.core._utils.version cimport cy_binding_version, cy_driver_version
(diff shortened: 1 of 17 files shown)


cuda_core/cuda/core/graph/_subclasses.pyx
```diff
@@ -340,6 +385,77 @@ cdef class MemsetNode(GraphNode):
...
+    def update(
+        self,
+        dst: Buffer | int | None = None,
+        value=None,
+        width: int | None = None,
+        height: int | None = None,
+        pitch: int | None = None,
+        *,
+        dst_owner=None,
+    ) -> None:
```
mdboom, line 561: One minor point (which really hinges on how we expect these `update` methods to be used).

If the idea is that users will most often be updating a subset of these values, it might be clearer to make them all kwarg only (i.e. put the `*,` at the beginning).

It's a minor point, but something we won't be able to change after the API freezes.

mdboom (commented):
The general organization and approach looks fine to me.

One small suggestion (that I can't really evaluate because I don't deeply understand the use cases).

cuda_core/cuda/core/graph/_subclasses.pyx
Andy-Jost: Good suggestion. I think kw-only is the better option here.

cuda_core/tests/graph/test_graph_node_update.py
```diff
@@ -0,0 +1,728 @@
...
+    case = definition_update_case
+    old_graph = case.graph_def.instantiate()
+
+    case.update(case.replacement)
+    case.assert_current(case.replacement)
+
+    new_graph = case.graph_def.instantiate()
+    case.assert_exec_uses(old_graph, case.original)
+    case.assert_exec_uses(new_graph, case.replacement)
+
```
juenglin, line 701:
```suggestion
assert case.original != case.replacement
assert old_graph != new_graph
```

cuda_core/tests/graph/test_graph_node_update.py
```diff
@@ -0,0 +1,728 @@
...
+    case = definition_update_case
+
+    assert case.invalid_update is not None
+    assert case.invalid_exception is not None
+    with pytest.raises(case.invalid_exception):
+        case.invalid_update()
+
+    case.assert_current(case.original)
+    graph = case.graph_def.instantiate()
+    case.assert_exec_uses(graph, case.original)
```
juenglin, line 731:
```suggestion
    case = definition_update_case

    old_graph = case.graph_def.instantiate()
    assert case.invalid_update is not None
    assert case.invalid_exception is not None
    with pytest.raises(case.invalid_exception):
        case.invalid_update()

    case.assert_current(case.original)
    new_graph = case.graph_def.instantiate()
    case.assert_exec_uses(new_graph, case.original)
    assert case.original != case.replacement
    assert old_graph == new_graph
```

cuda_core/tests/graph/test_graph_node_update.py
```diff
@@ -0,0 +1,728 @@
...
+        graph_def=graph_def,
+        node=node,
+        original=original,
+        replacement=replacement,
+        update=update,
+        assert_current=lambda _expected: _assert_equal(node.callback, None),
+        assert_exec_uses=assert_exec_uses,
+        invalid_update=invalid_update,
+        invalid_exception=ValueError,
+        invalid_argument_update=None,
```
juenglin, line 215:
```suggestion
        invalid_argument_update=lambda: None,
```

cuda_core/tests/graph/test_graph_node_update.py
```diff
@@ -0,0 +1,728 @@
...
+        invalid_exception=ValueError,
+        invalid_argument_update=lambda: node.update(object()),
+        cleanup=_noop,
+    )
+
+
+def _host_callback_ctypes_case(device):
+    """Use ctypes callbacks and copied payloads to distinguish each exec."""
+    callback_type = ctypes.CFUNCTYPE(None, ctypes.c_void_p)
+    called = []
```
juenglin, line 173:
```suggestion
    invalid_callback_type = ctypes.CFUNCTYPE(ctypes.c_int, ctypes.c_void_p)
    called = []
```

cuda_core/tests/graph/test_graph_node_update.py
```diff
@@ -0,0 +1,728 @@
...
+        return ctypes.cast(data, ctypes.POINTER(ctypes.c_uint8))[0]
+
+    @callback_type
+    def original_fn(data):
+        called.append((original_fn, read_byte(data)))
+
+    @callback_type
+    def replacement_fn(data):
+        called.append((replacement_fn, read_byte(data)))
+
```
juenglin, line 185:
```suggestion
    @invalid_callback_type
    def invalid_fn(data):
        return 0
```

cuda_core/tests/graph/test_graph_node_update.py
```diff
@@ -0,0 +1,728 @@
...
+    def assert_exec_uses(graph, expected):
+        called.clear()
+        stream = device.create_stream()
+        graph.launch(stream)
+        stream.sync()
+        assert called == [(expected[0], expected[1][0])]
+
+    def invalid_update():
+        node.update(lambda: None, user_data=b"not valid for a Python callback")
```
juenglin, line 203:
```suggestion
        node.update(invalid_fn, user_data=b"not valid for a Python callback")
```

cuda_core/tests/graph/test_graph_node_update.py
```diff
@@ -0,0 +1,728 @@
...
+        assert_current=lambda expected: _assert_equal(node.callback, expected),
+        assert_exec_uses=assert_exec_uses,
+        invalid_update=lambda: node.update(replacement, user_data=b"not valid for a Python callback"),
+        invalid_exception=ValueError,
+        invalid_argument_update=lambda: node.update(object()),
+        cleanup=_noop,
+    )
+
+
+def _host_callback_ctypes_case(device):
```
juenglin, line 170: This test could be improved to handling of update with wrong `CFUNCTYPE` as well.

cuda_core/tests/graph/test_graph_node_update.py
```diff
@@ -0,0 +1,728 @@
...
+
+        source_value = 0x11 if expected["src"] is original_src else 0x22
+        expected_data = [source_value] * expected["size"]
+        expected_data.extend([0] * (4 - expected["size"]))
+        assert list(as_bytes(expected["dst"])) == expected_data
+        if replacement_dst is not original_dst:
+            unexpected_dst = replacement_dst if expected["dst"] is original_dst else original_dst
+            assert list(as_bytes(unexpected_dst)) == [0] * 4
+
+    def cleanup():
```
juenglin: I am surprised to see all this explicit cleanup. Is it necessary and `cuda.python` applications need to do this as well?

cuda_core/tests/graph/test_graph_node_update.py
```diff
@@ -0,0 +1,728 @@
...
+        stream.sync()
+
+        assert as_int(expected["output"]).value == expected["expected"]
+        if replacement_buffer is not original_buffer:
+            unexpected = replacement_buffer if expected["output"] is original_buffer else original_buffer
+            assert as_int(unexpected).value == 0
+
+    def invalid_update():
+        if replace == "kernel":
+            node.update(kernel=replacement_kernel)
```
juenglin, line 450: Shouldn't we also test updating a kernel with wrong args?
For this case, `empty_kernel` does not take any args, so I would expect updating with `kernel=empty_kernel, args=original_args` to also fail.

cuda_core/tests/graph/test_graph_node_update.py
```diff
@@ -0,0 +1,728 @@
...
+        node.update(config=cooperative)
+    with pytest.raises(NotImplementedError, match="clustered or cooperative"):
+        graph_def.launch(cooperative, kernel)
+
+    node.destroy()
+
+
+@pytest.mark.agent_authored(model="gpt-5.6")
+def test_partial_memory_updates_are_keyword_only(init_cuda):
+    memory_resource = LegacyPinnedMemoryResource()
```
juenglin, line 630: The tests only use `LegacyPinnedMemoryResource`. Is there a gap in test coverage?

cuda_core/tests/graph/test_graph_node_update.py
juenglin, line 1: Should we have a test for, respectively:
1. trying to update node on an invalid graph
2. changing the type of memory in a `MemcpyNode` (e.g. from host->device to device->device)
3. if it's possible to clone a node or a graph: clone a graph G to produce H, update node in G and observe that nodes in H are unchanged

cuda_core/cuda/core/graph/_subclasses.pyx
```diff
@@ -428,6 +745,127 @@ cdef class MemcpyNode(GraphNode):
...
+            raise NotImplementedError(
+                "updating multidimensional, pitched, offset, or array-backed "
+                "memcpy nodes is not supported")
+
+        c_dst_type = params.memcpy.copyParams.dstMemoryType
+        c_src_type = params.memcpy.copyParams.srcMemoryType
+        if c_dst_type == cydriver.CU_MEMORYTYPE_HOST:
+            c_dst = <cydriver.CUdeviceptr><uintptr_t>(
+                params.memcpy.copyParams.dstHost)
+        elif c_dst_type != cydriver.CU_MEMORYTYPE_ARRAY:
```
juenglin: Would be safer to just make this `else` (same on line 824 below). Doesn't make a difference at the moment because of line 810. Otherwise follow the `elif ...:` with
```
   else:
        raise NotImplementedError("implement me")
```

cuda_core/cuda/core/graph/_subclasses.pyx
```diff
@@ -340,6 +549,114 @@ cdef class MemsetNode(GraphNode):
...
+        if pitch is not None:
+            c_pitch = pitch
+
+        params.memset.dst = c_dst
+        params.memset.value = c_value
+        params.memset.elementSize = c_element_size
+        params.memset.width = c_width
+        params.memset.height = c_height
+        params.memset.pitch = c_pitch
+
```
juenglin:
```suggestion
        cdef OpaqueHandle dst_attachment_owner
        cdef GraphHandle h_graph
        cdef cydriver.CUgraphNode node = as_cu(self._h_node)
        cdef cydriver.CUcontext ctx = NULL
        cdef cydriver.CUDA_MEMSET_NODE_PARAMS current
        cdef cydriver.CUgraphNodeParams params
        cdef object queried

        if dst is None and dst_owner is not None:
            raise ValueError("dst_owner requires dst")
        if (dst is None and value is None and width is None and
                height is None and pitch is None):
            return

        c_memset(&params, 0, sizeof(params))
        params.type = cydriver.CU_GRAPH_NODE_TYPE_MEMSET
        with nogil:
            HANDLE_RETURN(cydriver.cuGraphMemsetNodeGetParams(
                node, &current))
        if _check_node_get_params():
            queried = handle_return(driver.cuGraphNodeGetParams(
                <uintptr_t>node))
            ctx = <cydriver.CUcontext><uintptr_t>int(queried.memset.ctx)
        else:
            with nogil:
                HANDLE_RETURN(cydriver.cuCtxGetCurrent(&ctx))

        cdef cydriver.CUdeviceptr c_dst = current.dst
        cdef unsigned int c_value = current.value
        cdef unsigned int c_element_size = current.elementSize
        cdef size_t c_width = current.width
        cdef size_t c_height = current.height
        cdef size_t c_pitch = current.pitch

        if dst is None:
            h_graph = graph_node_get_graph(self._h_node)
            HANDLE_RETURN(graph_get_attachment(
                h_graph, node,
                &dst_attachment_owner, NULL))
        else:
            dst_attachment_owner = _resolve_memcpy_operand(
                dst, dst_owner, "dst", &c_dst)

        if value is not None:
            c_value, c_element_size = _parse_fill_value(value)
        if width is not None:
            c_width = width
        if height is not None:
            c_height = height
        if pitch is not None:
            c_pitch = pitch

        params.memset.dst = c_dst
        params.memset.value = c_value
        params.memset.elementSize = c_element_size
        params.memset.width = c_width
        params.memset.height = c_height
        params.memset.pitch = c_pitch
        params.memset.ctx = ctx

```

cuda_core/cuda/core/graph/_subclasses.pyx
```diff
@@ -478,6 +916,37 @@ cdef class ChildGraphNode(GraphNode):
...
+            h_parent, self._h_child_graph, node,
+            child._h_graph, &prepared))
+        with nogil:
+            HANDLE_RETURN(cydriver.cuGraphNodeSetParams(
+                node, &params))
+        try:
+            commit_status = graph_commit_child_graph_update(
+                prepared, &h_replacement)
+        finally:
+            if h_replacement:
```
juenglin, line 940:
```suggestion
            if commit_status == CUDATA_SUCCESS:
```

Andy-Jost: `graph_commit_child_graph_update` returns CUDA errors through the status, but C++ errors are reported by throwing, so `commit_status` is not correct here.

cuda_core/tests/graph/test_graph_node_update.py
Andy-Jost: 1. There's no way to create an invalid `GraphDefinition` through cuda.core's public API, but we can test calling `update` on destroyed nodes (I added some).
2. Added a test.
3. cuda.core does not expose a graph clone API. Filed #2438.

cuda_core/tests/graph/test_graph_node_update.py
Andy-Jost: Is the concern that the `CFUNCTYPE` could have the wrong signature? cuda.core currently does not validate that. Rather than squeeze it into this PR, I filed #2439.

cuda_core/tests/graph/test_graph_node_update.py
Andy-Jost: You're right, it should not be needed. This was added to work around a teardown error (below). I will continue working on that.

> Warning: mr.deallocate() failed during Buffer destruction: CUDA_ERROR_INVALID_CONTEXT: This most frequently indicates that there is no context bound to the current thread. This can also be returned if the context passed to an API call is not a valid handle (such as a context that has had cuCtxDestroy() invoked on it). This can also be returned if a user mixes different API versions (i.e. 3010 context with 3020 API calls). See cuCtxGetApiVersion() for more details. This can also be returned if the green context passed to an API call was not converted to a CUcontext using cuCtxFromGreenCtx API.

cuda_core/tests/graph/test_graph_node_update.py
Andy-Jost: cuda.core does not currently validate kernel arguments for ordinary launches either. `cuKernelGetParamInfo`, which enables count and size validation, requires CUDA 12.4. This should be addressed consistently across launch, graph construction, and update paths as separate work rather than only testing updates here.

cuda_core/tests/graph/test_graph_node_update.py
Andy-Jost: I don't think we need to test multiple memory allocators here.

cuda_core/tests/graph/test_graph_node_update.py
Andy-Jost: I don't think `old_graph == new_graph` is correct here? Because they are separate instances.

juenglin: Ok, so there is no equality comparison for graph instances?

cuda_core/tests/graph/test_graph_node_update.py
juenglin: Yes, exactly.

cuda_core/tests/graph/test_graph_node_update.py
Andy-Jost: There should be an equality based on the underlying `CUgraphExec` handle, but not a structural equality.

cuda_core/cuda/core/graph/_subclasses.pyx
juenglin: Hm, the `h_replacement` handle is not initialized on line 919. Why does this work, i.e., why would the `if h_replacement:` path not always be taken?

The function throws when a CUDA error occurred [here](https://github.com/NVIDIA/cuda-python/pull/2395/changes#diff-e772efc4150f80731482db13a9cfea60a0b34ff91c9c93d889f45f4b36e24f22R1695), btw.

Andy-Jost: > the h_replacement handle is not initialized on line 919

`GraphHandle` is a class type (`std::shared_ptr`), so it is default initialized (`graph_commit_child_graph_update` also `resets` it).

> why would the if h_replacement: path not always be taken?

`graph_commit_child_graph_update` has early return paths at the beginning.

> The function throws when a CUDA error occurred here, btw.

cuda.core raises CUDA errors as exceptions. Here, at the internal Cython/C++ interface, we are juggling multiple systems, each with its own error reporting mechanism so it's messy; ultimately, all error paths need to terminate with Python exceptions. I see at least three paths:

- `CUresult` returned: use `HANDLE_RETURN` to convert it to a Python error; reuses the standard mechanism in ordinary failure cases
- `CUresult` converted to a C++ exception internally: used at the line you called out because there is additional information to report (internal state is corrupted); Cython converts to a Python exception
- exception raised: a C++ exception occurred, e.g., `std::bad_alloc`; Cython converts

This is reasonable because we want to use the standard mechanism where we can, add important info where we must, and convert C++ exceptions automatically through Cython.

Andy-Jost:
This is ready for final review. The asynchronous cleanup exposed a latent bug, which is the source of the warnings mentioned [here](https://github.com/NVIDIA/cuda-python/pull/2395#discussion_r3668875913). I'd like to merge this PR as-is and fix that in a followup.

juenglin (approved):
LGTM

Issue #2397, opened 2026-07-21
Disallow importing of legacy modules outside of the `legacy_tests` subdir


mdboom:
As we start to bring over the new implementations of driver, runtime and nvrtc, the existing tests are getting copied into a `legacy_tests` subdir.  We should then disallow the use of the legacy API outside of that directory.

This is blocked on #2384.

PR #2398, opened 2026-07-21
New lowpp layer for nvrtc

mdboom:
Adds a new modern-generated version of the lowpp layer for nvrtc, and then updates all library code to use it.

As per the design document, this lands as `cuda.bindings._v2.nvrtc`, and then will be moved to `cuda.bindings.v2.nvrtc` once we are comfortable the API is stable.

The test update procedure is:

- Copy any test files that reference `nvrtc` as-is to a new `legacy_tests` directory (and pytest also requires that we rename the file).
- Update the existing test files to use the new API.

We should also disallow the use of the legacy API in the non-legacy_tests directory.  That's currently blocked on #2384, so I created #2397 to follow-up on that later.

====

Porting cuda_core to use these new APIs will need to be follow-on work.  Our dependency specs say that cuda_core has to work with older cuda_bindings, and of course they don't have this new API yet.  So we will need to stage that appropriately.


cuda_bindings/docs/source/release/13.4.0-notes.rst
```diff
@@ -13,6 +13,14 @@ Deprecation Notices
   removed in a future version.  Python 3.10 reaches end of life in October 2026
   per the `CPython support cycle <https://devguide.python.org/versions/>`_.
 
+Prerelease feature
+------------------
+
+A new version of the ``nvrtc`` API is available as ``cuda.bindings._v2.nvrtc``.  The
+primary improvements are: (1) raising exceptions rather than returning error
+codes, (2) uses PEP8-compliant naming, and (3) is more performant.  This API is
```
mdboom:
```suggestion
codes, (2) uses PEP8-compliant naming, and (3) more performance.  This API is
```

isVoid (approved):

mdboom:
Thanks for the review, @isVoid.  Marking as draft so we don't accidentally merge this until the generator-side is merged.

PR #2399, opened 2026-07-21
Harden program-cache permissions, binary-path resolution, and coredump helper lifetime

rparolin:
Small safety fixes across three areas, each with a test. All hardening — no intentional behavior change.

## What changed

**cuda.core**
- The program cache writes compiled GPU code to disk. The `tmp/` staging area (where in-flight writes land before the atomic rename into `entries/`) is now created owner-only (`0o700`) so other local users can't read or replace device code mid-write. Per review, `root/` and `entries/` intentionally inherit the umask / any pre-existing permissions so a deliberately shared kernel cache (e.g. group-writable on a compute cluster) keeps working — an existing directory is used as-is, never re-tightened.

**cuda.pathfinder**
- `find_nvidia_binary_utility` now always returns a full absolute path, as the docs promise (it could return a relative one before).
- On Windows, if `AddDllDirectory` fails, we now warn instead of failing silently.

**cuda.bindings**
- Fix two memory bugs in the coredump-settings helper (`_HelperCUcoredumpSettings`): it kept a pointer to bytes it didn't own (dangling pointer), and never freed a 1 KB buffer (leak).

## Security scope
This started as "make the whole cache tree owner-only regardless of umask." Per review it was narrowed to keep shared caches working, so the residual posture is:
- **Confidentiality — closed in practice.** Committed entries are written via `mkstemp` (`0o600`) and `os.replace` preserves that mode, so cached kernel *contents* stay owner-readable even though `entries/` inherits the umask. Other local users can list entry filenames (key hashes) but can't read the code.
- **Tampering on a normal single-user box — closed.** Only the owner can write `entries/`, so no one can plant a kernel this process later loads.
- **Tampering in a deliberately shared (group-writable) cache — accepted trade-off, NOT closed.** A group member with write access to `entries/` could replace a cached kernel. Tamper-proofing that path needs load-time integrity checks (a separate change), so it's out of scope here. Flagging for sign-off.

## Tests
Added tests for the `tmp/` permissions (and that a shared root is preserved), the absolute path, and the coredump helper. Directory-permission logic verified locally; full suite on RTX 5880.

Reviewers: @leofang (core/bindings), @rwgk (pathfinder).

### DIFF
--- a/cuda_bindings/cuda/bindings/_internal/utils.pyx
+++ b/cuda_bindings/cuda/bindings/_internal/utils.pyx
@@ -42,7 +42,18 @@ cdef bint is_nested_sequence(data):
 
 
 cdef void* get_buffer_pointer(buf, Py_ssize_t size, readonly=True) except*:
-    """The caller must ensure ``buf`` is alive when the returned pointer is in use."""
+    """Return a raw pointer into ``buf``.
+
+    The caller must ensure ``buf`` is alive AND not reallocated/resized while
+    the returned pointer is in use. For buffer-protocol objects the export lock
+    (``PyBuffer_Release``) is dropped before this function returns, so the
+    stack-local ``Py_buffer`` view cannot keep the export alive past return.
+    Concurrently resizing a mutable exporter (e.g. ``bytearray``, ``array.array``,
+    a growable NumPy view) during a ``nogil`` native call that dereferences the
+    pointer is therefore a use-after-free (#366). Prefer immutable buffers
+    (``bytes``), or hold the buffer export for the pointer's full lifetime at the
+    call site.
+    """
     cdef void* bufPtr
     cdef int flags = cpython.PyBUF_ANY_CONTIGUOUS
     if not readonly:
--- a/cuda_bindings/cuda/bindings/_internal/utils.pyx
+++ b/cuda_bindings/cuda/bindings/_internal/utils.pyx
@@ -42,13 +42,7 @@ cdef bint is_nested_sequence(data):
 
 
 cdef void* get_buffer_pointer(buf, Py_ssize_t size, readonly=True) except*:
-    """Return a raw pointer into ``buf``.
-
-    The caller must keep ``buf`` alive AND not resize it while the pointer is in
-    use. The buffer export is released before this returns, so resizing a mutable
-    buffer (e.g. ``bytearray``) during a ``nogil`` call is a use-after-free.
-    Prefer immutable ``bytes``.
-    """
+    """The caller must ensure ``buf`` is alive when the returned pointer is in use."""
     cdef void* bufPtr
     cdef int flags = cpython.PyBUF_ANY_CONTIGUOUS
     if not readonly:
--- a/cuda_bindings/cuda/bindings/_internal/utils.pyx
+++ b/cuda_bindings/cuda/bindings/_internal/utils.pyx
@@ -44,15 +44,10 @@ cdef bint is_nested_sequence(data):
 cdef void* get_buffer_pointer(buf, Py_ssize_t size, readonly=True) except*:
     """Return a raw pointer into ``buf``.
 
-    The caller must ensure ``buf`` is alive AND not reallocated/resized while
-    the returned pointer is in use. For buffer-protocol objects the export lock
-    (``PyBuffer_Release``) is dropped before this function returns, so the
-    stack-local ``Py_buffer`` view cannot keep the export alive past return.
-    Concurrently resizing a mutable exporter (e.g. ``bytearray``, ``array.array``,
-    a growable NumPy view) during a ``nogil`` native call that dereferences the
-    pointer is therefore a use-after-free (#366). Prefer immutable buffers
-    (``bytes``), or hold the buffer export for the pointer's full lifetime at the
-    call site.
+    The caller must keep ``buf`` alive AND not resize it while the pointer is in
+    use. The buffer export is released before this returns, so resizing a mutable
+    buffer (e.g. ``bytearray``) during a ``nogil`` call is a use-after-free.
+    Prefer immutable ``bytes``.
     """
     cdef void* bufPtr
     cdef int flags = cpython.PyBUF_ANY_CONTIGUOUS
--- a/cuda_core/cuda/core/utils/_program_cache/_file_stream.py
+++ b/cuda_core/cuda/core/utils/_program_cache/_file_stream.py
@@ -399,13 +399,21 @@ class FileStreamProgramCache(ProgramCacheResource):
         self._max_size_bytes = max_size_bytes
         self._root.mkdir(parents=True, exist_ok=True)
         self._entries.mkdir(exist_ok=True)
-        # ``tmp`` stages in-flight writes before they are atomically renamed into
-        # ``entries``. Create it owner-only (0o700) to keep other local users from
-        # reading or replacing compiled device code mid-write. ``root``/``entries``
-        # intentionally inherit the umask / any pre-existing permissions so a
-        # deliberately shared cache directory keeps working (see PR #2399 review).
-        # mkdir's mode is not chmod'd afterward, so an existing ``tmp`` is left
-        # as-is rather than re-tightened.
+        # Security scope (glasswing #359/#375, narrowed per PR #2399 review):
+        #  - Confidentiality: committed entries are written via ``mkstemp`` (0o600)
+        #    and ``os.replace`` preserves that mode, so cached kernel *contents*
+        #    stay owner-readable even though ``entries/`` itself inherits the umask.
+        #    Other local users can list entry filenames (key hashes) but not read
+        #    the compiled code.
+        #  - ``tmp`` stages in-flight writes before the atomic rename; create it
+        #    owner-only (0o700) so no one can read or swap device code mid-write.
+        #  - ``root``/``entries`` intentionally inherit the umask / any pre-existing
+        #    permissions so a deliberately shared cache (e.g. group-writable on a
+        #    cluster) keeps working. Accepted trade-off: in that shared setup a
+        #    group member with write access to ``entries/`` could replace a cached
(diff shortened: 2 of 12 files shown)


cuda_bindings/cuda/bindings/_internal/utils.pyx
```diff
@@ -42,7 +42,13 @@ cdef bint is_nested_sequence(data):
...
 
 cdef void* get_buffer_pointer(buf, Py_ssize_t size, readonly=True) except*:
-    """The caller must ensure ``buf`` is alive when the returned pointer is in use."""
+    """Return a raw pointer into ``buf``.
+
+    The caller must keep ``buf`` alive AND not resize it while the pointer is in
+    use. The buffer export is released before this returns, so resizing a mutable
+    buffer (e.g. ``bytearray``) during a ``nogil`` call is a use-after-free.
+    Prefer immutable ``bytes``.
+    """
```
leofang: This code is part of cybind (which @mdboom is improving). Let's not touch it.

cuda_core/cuda/core/utils/_program_cache/_file_stream.py
```diff
@@ -397,9 +397,16 @@ def __init__(
...
+        # Cache holds compiled device code loaded via cuLibraryLoadData, so keep
+        # it owner-only (0o700) so other local users can't read or replace it.
+        # mkdir's mode is umask-masked and exist_ok keeps an existing dir's
+        # perms, so re-chmod on POSIX.
+        self._root.mkdir(parents=True, exist_ok=True, mode=0o700)
+        self._entries.mkdir(exist_ok=True, mode=0o700)
+        self._tmp.mkdir(exist_ok=True, mode=0o700)
+        if os.name != "nt":
+            for d in (self._root, self._entries, self._tmp):
+                os.chmod(d, 0o700)
```
leofang: I am not sure this is acceptable. One can imagine in a compute cluster a group of users want to share the same kernel cache. If the directory already exists, we should use it as-is without changing the permission.

leofang: `self._tmp` can certainly be made owner-only.

cuda_bindings/cuda/bindings/_internal/utils.pyx
rparolin: Reverted — dropped the `get_buffer_pointer` docstring change in `_internal/utils.pyx` in 6fd6b6cb45 so cybind is untouched.

cuda_core/cuda/core/utils/_program_cache/_file_stream.py
rparolin: Done in 6fd6b6cb45. Now only `tmp/` (the in-flight staging dir) is created owner-only via `mkdir(mode=0o700)`; `root/` and `entries/` inherit the umask / any pre-existing permissions, so a deliberately shared cache directory is used as-is. I also dropped the post-mkdir `chmod` that would have re-tightened an existing shared dir. Tests updated accordingly (asserts only `tmp` is 0o700, and that a pre-existing 0o777 shared root is preserved).

rparolin:
Heads-up on the security posture after narrowing the cache-permission fix (details in the updated PR description under **Security scope**):

- Cached kernel **contents** stay owner-only (`0o600` via `mkstemp`+`os.replace`), and `tmp/` is owner-only, so confidentiality and in-flight tampering are covered.
- On a normal single-user box, planting a kernel into `entries/` isn't possible either.
- In a **deliberately shared, group-writable** cache, a group member could replace a cached entry that this process later loads. That's the accepted trade-off for keeping shared caches working — permissions alone can't both share-write and lock out writers; closing it would need load-time integrity checks (separate change).

@leofang flagging for sign-off on accepting that residual risk (mitigated for single-user; shared-cache tampering accepted).

rparolin:
Ready for another look — pushed changes addressing the review:

@leofang (cuda.core / cuda.bindings)
- Cache permissions narrowed to your guidance: only `tmp/` is created owner-only (`0o700`); `root/` and `entries/` inherit the umask / any pre-existing perms, so a deliberately shared cache is used as-is and never re-tightened. Dropped the post-`mkdir` `chmod`.
- Reverted the `get_buffer_pointer` docstring change in `_internal/utils.pyx` — that's cybind, left untouched.
- The `_HelperCUcoredumpSettings` dangling-pointer + 1 KB leak fix is unchanged.
- Please also sign off on the documented residual risk under **Security scope** (shared group-writable cache tampering — accepted trade-off).

@rwgk (cuda.pathfinder)
- `find_nvidia_binary_utility` now always returns an absolute path (honors the documented contract).
- Windows `AddDllDirectory` failures now warn instead of being swallowed.

Thanks!

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_windows.py
```diff
@@ -72,7 +73,17 @@ def add_dll_directory(dll_abs_path: str) -> None:
...
+    if not cookie:
+        # Warn instead of failing silently; the PATH update below is a weaker
+        # fallback that newer loaders may ignore.
+        error_code = ctypes.GetLastError()  # type: ignore[attr-defined]
+        warnings.warn(
+            f"AddDllDirectory({dirpath!r}) failed (error code: {error_code}); "
+            "falling back to process-global PATH mutation for dependent-DLL resolution.",
+            RuntimeWarning,
+            stacklevel=2,
+        )
```
leofang, line 85: Q: Why wouldn't we just call `os.add_dll_directory()`? What is (was, apparently it's already there) the reason that we call win32 API directly?

leofang:
cuda.core LGTM!

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_windows.py
rparolin: Good call — migrated to `os.add_dll_directory()` in f7b16841f7. Both wrap the same Win32 `AddDllDirectory`, but the stdlib call lets us drop the hand-maintained `kernel32.AddDllDirectory` argtypes/restype binding and reports failure as `OSError` instead of a NULL-cookie + `GetLastError()` check. Behavior is unchanged: the returned handle is intentionally discarded (no finalizer, so the directory stays on the search path for the process lifetime — verified against CPython `main`), and the PATH fallback + failure warning are preserved.

cuda_pathfinder/cuda/pathfinder/_binaries/find_nvidia_binary_utility.py
```diff
@@ -63,7 +63,9 @@ def _resolve_in_trusted_dirs(normalized_name: str, dirs: list[str]) -> str | Non
         seen.add(directory)
         candidate = os.path.join(directory, normalized_name)
         if _is_executable_candidate(candidate):
-            return candidate
+            # Return an absolute path, as the docstring promises (a relative
+            # search dir would otherwise leak a relative result).
+            return os.path.abspath(candidate)
```
mdboom, line 68: We really should migrate from `os.path` to `pathlib.Path` (which has far fewer foot-guns).  https://github.com/NVIDIA/cuda-python/issues/2410

mdboom (approved):
LGTM

cuda_pathfinder/cuda/pathfinder/_binaries/find_nvidia_binary_utility.py
rparolin: @mdboom Can you create a ticket for that work? We can then prioritize it and remove it from the code base completely. Should we also include a pre-commmit check that prevents people from creating new code with `os.path` in it?

cuda_core/cuda/core/utils/_program_cache/_file_stream.py
leofang: For posterity: CUDA driver uses `644` instead of `700` for its kernel cache, so there is absolutely no reason we want to be stricter than the driver.
```shell
$  ls -lt ~/.nv/ComputeCache/0/0/ | awk '{k=0;for(i=0;i<=8;i++)k+=((substr($1,i+2,1)~/[rwx]/) \
             *2^(8-i));if(k)printf("%0o ",k);print}'
total 760
644 -rw-r--r-- 1 leof hardware  30062 Aug 11 05:19 2c8b950ddfa198
644 -rw-r--r-- 1 leof hardware  27662 Aug 11 05:16 af0ebff995d7a5
644 -rw-r--r-- 1 leof hardware  19299 Aug 11 05:09 e07c0ce7bd1690
644 -rw-r--r-- 1 leof hardware  21162 Aug 11 05:09 954c4b9174072a
644 -rw-r--r-- 1 leof hardware  21114 Aug 11 05:08 7142b32ac469c2
644 -rw-r--r-- 1 leof hardware  32099 Aug 11 05:07 7ac86f08382b04
644 -rw-r--r-- 1 leof hardware 383310 Aug 11 04:56 3f4fa57c1fd952
644 -rw-r--r-- 1 leof hardware  23974 Aug 11 04:51 946c2c627e7c2a
644 -rw-r--r-- 1 leof hardware  17912 Aug 11 04:51 4d4dd9a1bf938e
644 -rw-r--r-- 1 leof hardware  20102 Aug 11 04:46 0d95387a0f7373
644 -rw-r--r-- 1 leof hardware  16081 Aug 11 04:44 3f569d46506964
644 -rw-r--r-- 1 leof hardware  21318 Aug 11 04:43 5a42bf0e30108d
644 -rw-r--r-- 1 leof hardware  17830 Aug 11 04:38 e1aff3c0057887
644 -rw-r--r-- 1 leof hardware  16265 Aug 11 04:37 39438a0c49a09a
644 -rw-r--r-- 1 leof hardware  18870 Aug 11 04:36 5a3f5e8ca7ee34
644 -rw-r--r-- 1 leof hardware  27154 Aug 11 04:36 c533bf5678f515
```

PR #2402, opened 2026-07-21
Raise OverflowError on out-of-range c_int/c_byte kernel arguments

rparolin:
> _Prepared with the assistance of an automated agent (Claude Code / Claude Opus 4.8). Please review carefully._

## Summary

An out-of-range Python int passed as a `c_int` / `c_byte` kernel argument was **silently narrowed** — e.g. `2**32 + 5` was stored as `5`, so the kernel ran with a value the caller never intended, with no error or warning.

The `c_int`/`c_byte` param feeders now range-check the value and **raise `OverflowError`** instead. Implementation notes:
- A single code path across all supported Pythons via `PyLong_AsLongAndOverflow` (flags out-of-`long` values through its `overflow` out-param) plus an explicit 32-bit / 8-bit bounds check.
- `feed()` is declared `except? -1` in the `.pxd` so Cython actually propagates the exception — without that, the raise would be silently swallowed (the same class of bug).

## Behavior change (please confirm)

This is intentionally **stricter than `ctypes`' own behavior**, which silently wraps out-of-range ints. Raising was chosen deliberately over matching `ctypes`.

## Tests

`test_kernelParams_c_int_out_of_range_raises` (GPU): an in-range value still launches; `2**32+5` (c_int) and `200` (c_byte) now raise `OverflowError`. Verified on Python 3.14.

## Context

Addresses Glasswing finding V9.1. Behavior and implementation reviewed/shaped by @leofang (the single-code-path `PyLong_AsLongAndOverflow` form was his call). Split out from the hardening batch in #2399 so it can be reviewed on its own.

### DIFF
--- a/cuda_bindings/cuda/bindings/_lib/param_packer.h
+++ b/cuda_bindings/cuda/bindings/_lib/param_packer.h
@@ -10,6 +10,26 @@
 #include <climits>
 #include <cstdint>
 
+// PyLong_AsInt entered the public/stable CPython API in 3.13. cuda.bindings
+// supports Python 3.10+, so provide a file-local backport for older builds.
+// This is a copy of the CPython implementation; it is `static` (unlike the
+// original) because this header is compiled into every extension module that
+// includes it, mirroring the other helpers below.
+#if PY_VERSION_HEX < 0x030D0000
+static int
+PyLong_AsInt(PyObject *obj)
+{
+    int overflow;
+    long result = PyLong_AsLongAndOverflow(obj, &overflow);
+    if (overflow || result > INT_MAX || result < INT_MIN) {
+        PyErr_SetString(PyExc_OverflowError,
+                        "Python int too large to convert to C int");
+        return -1;
+    }
+    return (int)result;
+}
+#endif
+
 static PyObject* ctypes_module = nullptr;
 
 static PyTypeObject* ctypes_c_char = nullptr;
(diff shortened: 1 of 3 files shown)


cuda_bindings/cuda/bindings/_lib/param_packer.h
```diff
@@ -69,7 +71,24 @@ static void populate_feeders(PyTypeObject* target_t, PyTypeObject* source_t)
...
+                long v = PyLong_AsLongAndOverflow(value, &overflow);
+                if (overflow == 0 && v == -1 && PyErr_Occurred())
+                    return -1;  // non-overflow conversion error; exception already set
+                if (overflow != 0 || v < INT_MIN || v > INT_MAX)
+                {
+                    PyErr_SetString(PyExc_OverflowError,
+                        "Python int is out of range for a c_int (32-bit) kernel argument");
+                    return -1;
+                }
+                *((int*)ptr) = (int)v;
```
mdboom: Python 3.13 and later have `PyLong_AsInt`, which already does exactly what we want.  The usual way to handle this is to put a backport copy of the function here, and then gate on the Python version.

Below is the implementation directly from CPython, wrapped in a version guard.  This should be added to the top of the file.

```
#if PY_VERSION_HEX < 0x030D0000
int
PyLong_AsInt(PyObject *obj)
{
    int overflow;
    long result = PyLong_AsLongAndOverflow(obj, &overflow);
    if (overflow || result > INT_MAX || result < INT_MIN) {
        PyErr_SetString(PyExc_OverflowError,
                        "Python int too large to convert to C int");
        return -1;
    }
    return (int)result;
}
#endif
```

```suggestion
                v = PyLong_AsInt(value);
                if (v == -1 && PyErr_Occurred())
                    return -1;
                *((int*)ptr) = (int)v;
```

cuda_bindings/cuda/bindings/_lib/param_packer.h
```diff
@@ -89,7 +108,19 @@ static void populate_feeders(PyTypeObject* target_t, PyTypeObject* source_t)
...
+                long v = PyLong_AsLongAndOverflow(value, &overflow);
+                if (overflow == 0 && v == -1 && PyErr_Occurred())
+                    return -1;  // non-overflow conversion error; exception already set
+                if (overflow != 0 || v < INT8_MIN || v > INT8_MAX)
+                {
+                    PyErr_SetString(PyExc_OverflowError,
+                        "Python int is out of range for a c_byte (8-bit) kernel argument");
+                    return -1;
+                }
+                *((int8_t*)ptr) = (int8_t)v;
```
mdboom: This one we do need to keep because it doesn't have an equivalent in CPython.

mdboom (changes requested):

cuda_bindings/cuda/bindings/_lib/param_packer.h
rparolin: We support Python 3.10+, correct? So wouldn't we need an alternative implementation when not using 3.13?

mdboom: Yes.  That's why we would need the code in the comment added to the top of the file.  (It's a 100% identical copy of what's in CPython, gated to only appear in Python builds < 3.13).

rparolin: Done in c45304d1ac. Added a file-local `PyLong_AsInt` backport gated on `PY_VERSION_HEX < 0x030D0000` (copy of the CPython impl, made `static` since this header compiles into driver/runtime/nvrtc), and the c_int feeder now calls `PyLong_AsInt`. The c_byte feeder keeps the explicit bounds check per your note, with a self-contained comment. The existing GPU test still passes.

cuda_bindings/cuda/bindings/_lib/param_packer.h
rparolin: Kept as-is (no CPython equivalent for the 8-bit slot). Updated its comment to be self-contained instead of referencing the c_int feeder, since that feeder now uses `PyLong_AsInt`.

mdboom (approved):

PR #2404, opened 2026-07-22
coverage: add cuda.core tests for system device, checkpoint, memory, stream, green context, and tensor map

rluo8:
## Description

Adds focused unit tests across `cuda.core`, targeting input-validation branches, error paths, and object-lifecycle behavior that were previously uncovered. All tests are real-path: host-only tests exercise pure helpers / argument validation directly, and GPU-required tests use the existing fixtures and skips. No production changes.

Measured effect on `cuda/core/` line coverage : 84.34% → ~85.6%, with `system/_device.pyx` crossing 85%. Largest gains in `checkpoint.py`, `_stream.pyx`, `_memory/_memory_pool.pyx`, `_tensor_map.pyx`, and`system/_device.pyx`.
**Note** that the coverage numbers are for `cuda/core/` only, combined across the local systems Linux (A10), Windows (A10), and Linux (H200).

| File | Tests |
|---|---|
| `graph/test_graph_definition.py` | host-callback node reconstruction via `_create_from_driver` through an embedded child graph |
| `system/test_system_device.py` | temperature / constructor-selector / affinity·topology·P2P argument validation (rejected before any NVML call) |
| `test_checkpoint.py` | `TestCheckpointHelpers` (host-only `_check_timeout_ms` / `_make_restore_args` / `_as_cuuuid` marshalling + validation) and `TestCheckpointDriverDispatch` (`_call_driver` result-code / missing-symbol translation via a dependency-injected boundary fake) |
| `test_green_context.py` | closed-context handle/`is_green`/`resources` degradation, empty/wrong-type `count` split rejection, primary-context stream SM-resource path |
| `test_launcher.py` | `test_launch_scalar_argument_subclass_fallback` expanded to all ctypes scalar widths/signs + numpy, with kernel read-back asserting correct C-type marshalling (guards e.g. `uint64 → uint32_t` truncation) |
| `test_memory.py` | mempool `DeviceMemoryResource` handle ownership, idempotent close, deallocate-frees-pointer (verified via `used_mem_current` drop, not just `handle == 0`), IPC-unsupported raise, `_MemPoolAttributes` non-instantiable |
| `test_stream.py` | `__cuda_stream__` protocol validation (malformed value / wrong length / bad version / deprecated attribute form), borrow-from-existing-Stream, from_handle flag/priority query, per-thread vs legacy default-stream env selection |
| `test_tensor_map.py` | TMA descriptor arg rejection branches — box_dim/enum via the public `TensorMapDescriptorOptions(...)` constructor, dtype and `box_dim`-required via the host-only marshalling helpers |


mdboom:
The `system` tests look good to me.  This should probably get reviewed by specialists in other areas of the code as well.

rluo8:
Thanks @mdboom! @leofang could you please help suggest the right experts to review the other areas (checkpoint, memory, stream, green context, and tensor map)?

Andy-Jost:
I was able to review everything except the tensor map changes. Looks good to me.

Update: given how straightforward the tensor map change appear, I'm comfortable approving.

Andy-Jost (approved):
LGTM!

PR #2405, opened 2026-07-22
feat(security): onboard pre-commit and Pulse secret scanning

gmanal:
## Description

Onboard NVIDIA secret scanning to cuda-python using the centrally maintained NVIDIA/security-workflows surfaces. Two complementary controls:

- **Local (advisory):** `secret-scan-trufflehog` pre-commit hook — catches credentials before commit.
- **CI (enforcement):** the Pulse reusable workflow — server-side scan on trusted branches that blocks on verified secrets.

## Changes

- Add `.github/workflows/secret-scan-pulse.yml` calling the centralized Pulse reusable workflow, on `main`, `ctk-next`, and copy-pr-bot `pull-request/*` branches; runs on `nv-cpu-general` (onboarded for this repo in the enterprise runner config).
- Add the `secret-scan-trufflehog` pre-commit hook (pinned to `security-workflows` `v0.1.0`), skipped on hosted pre-commit.ci which lacks the trufflehog binary.
- Set `failure_policy: unverified` explicitly (fail on verified/live secrets, warn on unverified); fails closed on scanner/infra errors.
- Pin per policy: pre-commit `rev` to release tag `v0.1.0`, workflow `uses:` to a reviewed commit SHA.
- Document local setup and testing in CONTRIBUTING.md.

## Checklist


CONTRIBUTING.md
```diff
@@ -75,6 +75,16 @@ If the hook isn't installed, `pre-commit run` (and CI) will print a visible
 warning reminding you to run `pre-commit install`.
 
 
+## Secret Scanning
+
+Two complementary controls prevent accidental credential leakage:
+
+- **Local pre-commit** (`secret-scan-trufflehog`, via `.pre-commit-config.yaml`): catches credentials before commit. Requires `trufflehog` on your PATH (the wrapper prints install commands if missing). Never commit a flagged secret; check with maintainers on false positives, and only use `git commit --no-verify` in exceptional, agreed cases.
```
mdboom: This means that [encouraging installation of `pre-commit`](https://github.com/NVIDIA/cuda-python/pull/2295) is even more important, as is making sure [our pre-commit config works on Windows](https://github.com/NVIDIA/cuda-python/pull/2327) (which it currently doesn't).

.pre-commit-config.yaml
```diff
@@ -9,12 +9,20 @@ ci:
...
     autoupdate_schedule: quarterly
-    skip: [lychee, check-precommit-installed]
+    skip: [lychee, check-precommit-installed, secret-scan-trufflehog]
     submodules: false
 
 # Please update the rev: SHAs below with this command:
 # pre-commit autoupdate --freeze
 repos:
+  # Runs first so a leaked credential blocks the commit before any formatter runs.
+  # Requires `trufflehog` on $PATH; the wrapper prints install commands if missing.
```
mdboom: Will this fail if not installed, or just warn (and continue to the other checks)?

The installation instructions mention an [insecure installation method](https://github.com/NVIDIA/security-workflows/blob/main/hooks/trufflehog.sh#L35):

```
  Install on macOS or Linux (no package manager required):
    curl -sSfL https://raw.githubusercontent.com/trufflesecurity/trufflehog/main/scripts/install.sh \
        | sh -s -- -b /usr/local/bin
```

Maybe this tool could install it in a safe way to the local pre-commit environment (the way other pre-commit hooks work).

.github/workflows/secret-scan-pulse.yml
```diff
@@ -0,0 +1,43 @@
...
+
+jobs:
+  secret-scan:
+    name: Secret Scan
+    # Pulse needs nv-gha-runners + Vault/nvcr vars; skip on forks.
+    if: github.repository == 'NVIDIA/cuda-python'
+    uses: NVIDIA/security-workflows/.github/workflows/secret-scan-pulse.yml@69032f641c3c34e0c0b46f636b54ed2b101b7aa4
+    with:
+      runs-on: linux-amd64-cpu4
+      # Set failure_policy explicitly so enforcement can't drift with upstream defaults.
```
kkraus14, line 39: Is the Linux-only execution context intentional? The called workflow relies on POSIX shell syntax, `docker`, and `python3`, while the local pre-commit hook is also Bash-only and documents no Windows installation path. If Windows developers are expected to use this hook, please add a supported Windows implementation/instructions; otherwise, please document that this control is Linux/macOS-only.

.github/workflows/secret-scan-pulse.yml
```diff
@@ -0,0 +1,43 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-on-${{ github.event_name }}-from-${{ github.ref_name }}
+  cancel-in-progress: true
+
+# Caller permissions must cover the reusable workflow (intersection applies).
+permissions:
+  contents: read
+  id-token: write          # OIDC -> Vault -> nvcr.io image pull
+  security-events: write   # publish redacted SARIF to code scanning
```
kkraus14, line 28: `NVIDIA/cuda-python` is public. GitHub documents `actions: read` as required by `upload-sarif` only for private repositories; this workflow otherwise needs `contents: read`, `id-token: write`, and `security-events: write`. Could we drop `actions: read` here to keep the caller token at the minimum scope?

CONTRIBUTING.md
gmanal: Agreed on both. As of security-workflows v0.2.0 (now pinned in this PR), the secret-scan-trufflehog hook is self-installing and cross-platform, including Windows via Git Bash/MSYS (which bundles bash/curl/tar/sha256sum). So the only setup a contributor needs is pre-commit install.

.pre-commit-config.yaml
gmanal: - Fail vs warn: the hook is fail-closed — it blocks the commit on a verified/live secret or a real scanner error. It no longer errors on "binary missing" because it now installs itself. (Unverified findings don't block locally by design — verified-only local policy — and the broader set is caught server-side by the Pulse CI workflow.)
- Insecure install: removed. It no longer uses curl | sh. It downloads the pinned official release archive over HTTPS and verifies it against a per-platform SHA-256 (pinned from TruffleHog's cosign-signed checksums.txt) before executing, failing closed on mismatch — installed into pre-commit's per-user cache. That's exactly the "install safely into the local pre-commit environment" approach you suggested.

The fix landed in the shared hook (https://github.com/NVIDIA/security-workflows/pull/12, released as [v0.2.0](https://github.com/NVIDIA/security-workflows/releases/tag/v0.2.0))

.github/workflows/secret-scan-pulse.yml
gmanal: - Local pre-commit hook: now cross-platform (Linux/macOS/Windows via Git Bash) as of [v0.2.0](https://github.com/NVIDIA/security-workflows/releases/tag/v0.2.0).
- CI reusable workflow (Pulse): intentionally Linux. It runs the NVIDIA-licensed Pulse scanner as a Docker container on nv-gha-runners with OIDC→Vault, so it's a server-side control on Linux CI runners — OS-independent from the contributor's side (a Windows dev's PR is still scanned in CI). I'll add a note that the CI control runs on Linux runners by design, while the developer hook is cross-platform.

.github/workflows/secret-scan-pulse.yml
gmanal: done

gmanal: Correction to my earlier "done" — I reverted this; `actions: read` is required here after all, though not for the private-repo reason.

The mechanism is reusable-workflow permission validation, not an intersection. The called workflow (`NVIDIA/security-workflows/.github/workflows/secret-scan-pulse.yml`, SHA-pinned) declares `actions: read` in its own `permissions:` block, and GitHub requires the **caller to grant at least every permission the reusable workflow declares**. With `actions: read` removed, the caller offered `actions: none` and the run failed at load time:

> Error calling workflow '…/secret-scan-pulse.yml@69032f64…'. The workflow is requesting 'actions: read', but is only allowed 'actions: none'.

Dropping it from the caller would only be valid if we also dropped it from the reusable definition — but that workflow is generic and also consumed by private repos, where `upload-sarif` genuinely needs `actions: read`. So the reusable correctly keeps declaring it, and public callers grant it (harmless on public). Restored in 94851cd.

mdboom:
Blocked by #2327 so that Windows developers can install `pre-commit` locally (and get the advantage of this here).

mdboom:
> Blocked by #2327 so that Windows developers can install `pre-commit` locally (and get the advantage of this here).

Given an offline discussion, this does not need to be blocked on #2327.

.pre-commit-config.yaml
```diff
@@ -9,12 +9,20 @@ ci:
...
     submodules: false
 
 # Please update the rev: SHAs below with this command:
 # pre-commit autoupdate --freeze
 repos:
+  # Runs first so a leaked credential blocks the commit before any formatter runs.
+  # Self-installing: the hook downloads a pinned, checksum-verified trufflehog on
+  # first use (no manual install). Skipped on pre-commit.ci; Pulse CI enforces server-side.
+  - repo: https://github.com/NVIDIA/security-workflows
+    rev: v0.2.0
```
mdboom: We should freeze this to a hash.

mdboom (changes requested):

.pre-commit-config.yaml
gmanal: as per the best practices, we are using hash for CI and version for pre-commit.
do you suggest we should have hash for pre-commit hook as well?

mdboom: Yes.  We use hashes for `pre-commit` as well.

gmanal: done.

mdboom (approved):

CONTRIBUTING.md
```diff
@@ -75,6 +75,11 @@ If the hook isn't installed, `pre-commit run` (and CI) will print a visible
 warning reminding you to run `pre-commit install`.
 
 
+## Secret Scanning
+
+The `secret-scan-trufflehog` pre-commit hook scans staged files and installs TruffleHog on first run (use Git Bash on Windows). If it flags a secret, remove it before committing, or contact a maintainer if it's a false positive. Secrets are also scanned server-side in CI.
```
kkraus14, line 172: Can we not require Git Bash on Windows? Given the pre-commit hook is mostly just downloading the correct platform binary to use, I imagine we can implement it in a way that doesn't add any extra dependencies for developers?

rparolin (approved):

CONTRIBUTING.md
gmanal: Good call. 
as hook detects OS/arch, downloads a pinned TruffleHog release, verifies its SHA-256, caches it, and execs it. None of that needs a shell and can be achieved using python script as well.

The fix will be to move the wrapper to `language: python`. pre-commit is itself Python, so an interpreter is guaranteed on every platform, and the whole thing is stdlib: `platform` for arch detection, `urllib.request` to download, `hashlib` to verify, `tarfile`/`zipfile` to extract. 
No Git Bash, no curl, no extra dependencies on any OS.

Do you want to hold this PR for this change? or can we move and add it later.

kkraus14: I've opened https://github.com/NVIDIA/security-workflows/pull/14 to address this upstream

gmanal:
Closing in favor of #2589, which supersedes this PR.

#2589 does everything here — the `secret-scan-trufflehog` pre-commit hook and the server-side Pulse secret scan with `failure_policy: unverified` — but routes them through the `security-workflows` **security suite**: one pinned reference instead of one caller per scan. Beyond that it enables CodeQL SAST and retires this repo's `codeql.yml`, because both publish under `/language:python` and two analyses on the same commit overwrite each other's alerts. The suite runs the same analysis the retired workflow did: `python`, `build-mode: none`, `security-extended`.

@kkraus14 — your Git Bash question is resolved. The hook moved to `language: python` and installs the pinned, SHA-256-verified binary into pre-commit's own environment (NVIDIA/security-workflows#14, shipped in v0.3.0). No Git Bash, no curl, no extra developer dependencies; verified on Linux, macOS, and Windows across x86_64 and arm64. #2589 pins v0.3.0 and drops the Git Bash line from CONTRIBUTING.

@mdboom @rparolin — thanks for the reviews here. The SHA-pinning and CONTRIBUTING feedback both carried into #2589, and I've left a comment there summarising exactly what differs from this PR.

Issue #2388, opened 2026-07-18
[BUG]: VirtualMemoryResource: four pre-existing defects (grow-rollback access loss, dead fast path, finalizer warnings, handle_type docstring)


leofang:
### Component

cuda.core

### What happened?

Four pre-existing defects in `VirtualMemoryResource`, all found while verifying #2235 (verification details in https://github.com/NVIDIA/cuda-python/pull/2235#pullrequestreview-4728931983); none are introduced by that PR.

1. **Rollback of a failed grow loses access grants** — if `modify_allocation()` fails after the old range has been remapped (slow path), the [`_remap_old`](https://github.com/NVIDIA/cuda-python/blob/4ab061fcc9c5286a4d527967e53681a3ea4e5c39/cuda_core/cuda/core/_memory/_virtual_memory_resource.py#L400-L409) rollback restores the mapping at the original address but never re-applies the access descriptors, so the rolled-back buffer faults on its next access until `cuMemSetAccess` is re-run. Reproduced on `main` by forcing `cuMemSetAccess` to fail during a grow. Likely fix: `_remap_old` should re-apply the resource's access descriptors to the old range (best-effort, matching the remap itself).

2. **The grow fast path is dead code** — [this check](https://github.com/NVIDIA/cuda-python/blob/4ab061fcc9c5286a4d527967e53681a3ea4e5c39/cuda_core/cuda/core/_memory/_virtual_memory_resource.py#L268) compares a `CUdeviceptr` against a plain `int`, and `CUdeviceptr(x) == x` is always `False`, so `modify_allocation()` always takes the slow path (full re-reserve + remap, base pointer changes) even when the driver granted the exact contiguous extension address. Fix: compare `int(new_ptr)`.

3. **Warning spam after every slow-path grow** — the slow path calls [`buf._clear()`](https://github.com/NVIDIA/cuda-python/blob/4ab061fcc9c5286a4d527967e53681a3ea4e5c39/cuda_core/cuda/core/_memory/_virtual_memory_resource.py#L450) so the old buffer's destructor won't double-free, but the destructor still calls `deallocate()`, emitting `Warning: mr.deallocate() failed during Buffer destruction: CUDA_ERROR_INVALID_VALUE` at GC after each grow. (Adjacent to the existing TODO referencing #2049.)

[...]

PR #2407, opened 2026-07-22
fix(core): compare CUdeviceptr as int so the VMM grow fast path can run

aryanputta:
## Description

Addresses defect 2 of #2388 (the dead grow fast path). The other three defects in that issue are not touched here, so this does not close it.

`modify_allocation()` asks `cuMemAddressReserve` for the address immediately after the current range, then checks whether the driver granted that exact address before taking the fast path:

```python
res, new_ptr = driver.cuMemAddressReserve(..., int(buf.handle) + aligned_prev_size, 0)

if res != driver.CUresult.CUDA_SUCCESS or new_ptr != (int(buf.handle) + aligned_prev_size):
```

`new_ptr` is a `CUdeviceptr`. That type defines `__int__` and `__repr__` but no `__eq__` or `__richcmp__` (`cuda_bindings/cuda/bindings/driver.pyx`), so comparing it against a plain int falls back to identity and is unconditionally unequal. The guard is therefore always taken, `_grow_allocation_fast_path` is unreachable, and every grow runs the slow path: a full re-reserve plus remap, with the buffer's base pointer changing even when the driver did grant a contiguous extension.

The fix is the `int()` conversion suggested in the issue.

### Behavior change, not a pure cleanup

This makes a previously unreachable path live. After it, a grow the driver can satisfy contiguously preserves the base pointer instead of moving it. That is the documented intent of the fast path, but it has never actually executed, so it deserves explicit attention in review. Added to the 1.2.0 release notes for that reason. The existing test asserting a grown buffer's pointer may change (`test_vmm_allocator_grow_allocation`) still passes, since it only requires a non-null handle.

Since the fast path becomes reachable for the first time, I checked the one line in it that looked risky. `buf._size = new_size` still carries a `TODO: #2049 This is a real bug, accessing _size which doesn't exist`. That TODO is stale on the Cython side: `_size` was moved into a `cdef public` block in `_buffer.pxd` by #2216 and #2049 is closed, so the assignment works on a real `Buffer`. The `# type: ignore[attr-defined]` is still required because `_buffer.pyi` does not expose `_size`, so I left both the ignore and the TODO alone rather than widening this PR. Happy to clean that up separately if you want it.

### Testing

`test_vmm_allocator_grow_dispatches_to_fast_path` stubs `cuMemAddressReserve` to grant exactly the address that was requested and asserts the dispatch reaches the fast path. It should fail on current `main`, recording `"slow"`. It follows the stubbing style of the adjacent `test_vmm_allocator_grow_allocation_fast_path`.

That existing test covered `_grow_allocation_fast_path` by calling it directly, which is why the broken dispatch was invisible: it stays green whether or not `modify_allocation` can actually reach the helper. The gap was coverage of the decision that reaches it.

## Checklist


aryanputta:
I have been assigned #2388, so I am taking this out of draft. It still covers defect 2 only and does not close the issue; I can follow up on defects 1, 3, and 4 in separate PRs if that is the split you prefer.

One thing for a reviewer: workflows on this PR are still waiting on vetter validation, and I do not have an NVIDIA GPU, so `test_vmm_allocator_grow_dispatches_to_fast_path` has not run anywhere yet. It is expected to fail on current `main` by recording `"slow"`.

aryanputta:
Current head is `24f3f6c`, rebased onto `main`. Changes since I opened this:

- Dropped the explanatory comment above the guard. It described why the old comparison was wrong, which belongs in the PR description rather than in the source.
- Dropped `test_cudeviceptr_is_never_equal_to_plain_int`. It only asserted that a bindings type has no `__eq__`, which is verifiable by reading `driver.pyx`, so it was not earning its place in the suite.
- Added a bullet to the `cuda.core` 1.2.0 release notes under "Fixes and enhancements", since this changes observable behavior of a stable API rather than being an internal cleanup.
- The remaining test is unchanged in substance and the one-line `int()` fix is unchanged.

The PR description also records something I checked while reviewing this: the `TODO: #2049` above `buf._size = new_size` in `_grow_allocation_fast_path` is stale, since `_size` became `cdef public` in #2216 and #2049 is closed. I left the TODO and the `type: ignore` in place because `_buffer.pyi` still does not expose `_size`, but it is worth knowing that the assignment does work on a real `Buffer`, given this PR is what makes that line reachable.

aryanputta:
Head is now `ae5b4e5d`. Merged current `main` to clear the conflict; the PR is mergeable again.

The only conflict was in `cuda_core/docs/source/release/1.2.0-notes.rst`, where entries for #2517, #2439, and #2389 landed in the same list. Resolution keeps all of them and places this PR's entry last. The one-line `int()` fix and the dispatch test are unchanged.

This still covers defect 2 of #2388 only. Workflows here have never run, so `test_vmm_allocator_grow_dispatches_to_fast_path` has not executed anywhere; it is expected to fail on unpatched `main` by recording `"slow"`.

Could someone run CI on this when convenient?

Andy-Jost:
Thanks for the PR. VMM work is consolidated under #2906.

This makes the same one-line change as #2237, which was opened first, and the fast path has to stay off until #2887 is fixed: a fast-path grow leaks its extension. Closing in favor of #2237. A driver-backed version of the dispatch test belongs with the redesign in #2906, which lists what is open for contribution.

PR #2411, opened 2026-07-23
ci: drop custom NumPy and use beta 4 for Python 3.15

### LINKED ISSUE #2401: Drop NumPy builds for Python 3.15 from CI
[NumPy just had 3.15 nightlies](https://anaconda.org/channels/scientific-python-nightly-wheels/packages/numpy/files?file_q=cp315). (Our 3.15 works actually started before NumPy was even ready, so [we had to build NumPy in our CI](https://github.com/NVIDIA/cuda-python/blob/76bac4e1186a1c035a006d3f2dec679690bc9984/.github/workflows/build-wheel.yml#L370-L422) and we can finally remove them now.)

tirthpatel90:
## Description

Fixes #2401

This PR removes custom NumPy build and wheel installation steps for Python 3.15 across CI workflow files (`build-wheel.yml`, `test-wheel-linux.yml`, and `test-wheel-windows.yml`) now that pre-built NumPy 3.15 nightly wheels are available.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/cuda-python/blob/main/CONTRIBUTING.md).


mdboom:
/ok to test [b33bb7c](https://github.com/NVIDIA/cuda-python/pull/2411/commits/b33bb7cece963b91c1467c95ed0fccc9648ce076)

mdboom:
It's going to be hard to evaluate this PR until the Python 3.15 is green on main again.  Our self-built Numpy is segfaulting on import on main, and this upstream build of Numpy also is.  It's great that the behavior matches, but we can't really evaluate that Numpy is working in CI until Numpy and/or CPython fixes the bug that is causing the segfault.

So we might just need to sit on this one for a while.

seberg (changes requested):
Thanks, but as the issue says, this should use the scientific-python-nightly-wheels not build from scratch.
Effectively, this just moved the build from build-time to test time.  (Which pointed to it not being NumPy but rather an ABI change in CPython, but that is beside the point.)

Updating this will indeed *not* fix things since NumPy is also build with b2 but the main blocker for merging is that we should use the nightlies and not build from scratch.

tirthpatel90:
Hi @seberg, I have updated the workflows to pass `PIP_EXTRA_INDEX_URL` for Python 3.15 so `pip` fetches pre-built nightly wheels from the `scientific-python-nightly-wheels` index instead of building from source. Thanks for pointing that out!

.github/workflows/test-wheel-windows.yml
```diff
@@ -283,6 +283,11 @@ jobs:
           freethreaded: ${{ endsWith(matrix.PY_VER, 't') }}
           allow-prereleases: ${{ startsWith(matrix.PY_VER, '3.15') }}
 
+      - name: Enable Scientific Python Nightly Wheels for Python 3.15
+        if: ${{ startsWith(matrix.PY_VER, '3.15') }}
+        shell: bash --noprofile --norc -xeuo pipefail {0}
+        run: echo "PIP_EXTRA_INDEX_URL=https://pypi.anaconda.org/scientific-python-nightly-wheels/simple" >> "$GITHUB_ENV"
```
seberg: Unfortunately, I don't think this'll work, for two reasons:
1. We actually have narrow version pins `numpy<2.5`
2. I believe pip will still prefer the 2.5.x sdist and compile from scratch.

So, I think we need to relax the pin to `<=2.6.0.dev0` (all numpy pre-releases).  We could do that in `env-vars` via env vars by setting up what you have here plus `PIP_ONLY_BINARY=numpy` (to avoid pip prefering to build `2.5.x`).
(We don't need `PIP_PRE=1` if we change the pin, pip should eventually find that version if no other version has a wheel that works.)

Sorry for the creep, maybe I'll push it or follow-up too:  The windows Python 3.15 jobs were currently deleted, can you try re-adding them now that we don't have to build NumPy anymore?
(if not that is fine, I'll follow up).

tirthpatel90: Hi @seberg, thanks for the detailed feedback! I have updated the workflows for Python 3.15 to pass `PIP_ONLY_BINARY=numpy` and set `NUMPY_VERSION=<=2.6.0.dev0` via environment variables so `pip` fetches pre-built nightly wheels directly without compiling from source.

seberg: Thanks, the NumPy version had to be changed in the `pyproject.toml` (or another work-around needed).
But it's pretty simple, so pushed that (and a try to re-activate windows python 3.15 jobs).

seberg:
/ok to test [b5984f8](https://github.com/NVIDIA/cuda-python/pull/2411/commits/b5984f84293d001607260519ffb8bed040a35bd6)

seberg (approved):

seberg:
/ok to test [77b9e9d](https://github.com/NVIDIA/cuda-python/pull/2411/commits/77b9e9d7508898621207088a8defe40d42c565cc)

seberg:
/ok to test [32c03f6](https://github.com/NVIDIA/cuda-python/pull/2411/commits/32c03f66e375575444dcd6756af10a920752f0f7)

seberg:
/ok to test [749bef1](https://github.com/NVIDIA/cuda-python/pull/2411/commits/749bef1c10a05e3630c0874241518858ec6d2f89)

seberg:
OK, hopefully things will auto-merge now, thanks @tirthpatel90, sorry that I took over largely, but this is tedious without CI iteration.  Changes things so that:
* We allow installing numpy dev wheels for 3.15.
* Re-enables the windows 3.15 job (but not 3.15t, due to missing psutils and maybe more)
* Includes the `cibuildwheel` bump, because otherwise 3.15 betas will mismatch and lead to segfaults.

We should not have further ABI issues after this, just cleanups when proper 3.15 packages arrive.  (Once NumPy releases 2.5.x wheels those should be picked up, though.)

tirthpatel90:
Thanks @seberg for guiding me through this and completing the CI adjustments! Glad to contribute to the project.

seberg:
/ok to test [d51c44f](https://github.com/NVIDIA/cuda-python/pull/2411/commits/d51c44f6c48bd714d27cadd7e2f49b7dc725e15a)

PR #2414, opened 2026-07-23
Fix OOM state-desync in generated array-member setters

rparolin:
> _Prepared with the assistance of an automated agent (Claude Code / Claude Opus 4.8). Please review carefully._

## Summary

The generated setters for list-valued struct members (`CUlaunchConfig.attrs`, `CUDA_MEM_ALLOC_NODE_PARAMS.accessDescs`, and ~27 siblings) freed the old heap buffer **before** allocating the new one. On an allocation failure they raised `MemoryError` *before* updating the cached length and the C-struct pointer, leaving the wrapper with a freed buffer but a **stale length** and a **dangling** `_pvt_ptr[0].<m>`:

- a later same-length assignment took the no-resize path and `memcpy`'d through the freed/NULL buffer -> near-NULL **segfault**, and
- `getPtr()`-then-CUDA read the dangling pointer -> **use-after-free**.

## Fix

Allocate and fill the new buffer first, and free/swap the old one **only once the resize is known to succeed** -- a strong exception guarantee. A failed (over)allocation now leaves the object completely unchanged.

## Regeneration

The change is in the cybind generator template (`legacy_cython_gen`); this PR is the full regenerated output of:

```
python -m cybind --generate-module "cuda.*" --output-dir <checkout>/cuda_bindings --ctk-target-version 13.3
```

Besides the setter fix (18 `driver` + 11 `runtime` setters), the regen also picks up **docstring-only cleanups** that have landed in cybind since the last sync (e.g. `nvml.pyx`, `runtime.pxd`, `cyruntime.pxd`, `runtime.rst`). These are formatting-only -- **no API, signature, or behavior changes** -- and are included because they are part of what the canonical regeneration command produces.

## Verification

- Pre-fix build: reproducer SIGSEGVs on the same-length re-set after a failed resize.
- Post-fix (regenerated) build: survives; object state intact.
- New subprocess regression test `test_array_setter_preserves_state_on_failed_resize` forces the failed resize via an overflowing `__len__` (so `calloc` overflows `size_t` -> NULL, no real host OOM needed). Passes; the two existing #2112 array-setter regression tests still pass.

## Context

Addresses NVBUG 6268871. Companion cybind template change is under review separately; this PR exists so the regenerated bindings get full CI.

### DIFF
--- a/cuda_bindings/cuda/bindings/cyruntime.pxd
+++ b/cuda_bindings/cuda/bindings/cyruntime.pxd
@@ -4,7 +4,7 @@
 #
 # This code was automatically generated across versions from 12.9.0 to 13.3.0. Do not modify it directly.
 
-# CYTHON-BINDINGS-GENERATED-DO-NOT-MODIFY-THIS-FILE: format=1; content-sha256=2b62a3b56226f5cf6953f578057af792369a100b7775328c5aa66d81d780d2ac
+# CYTHON-BINDINGS-GENERATED-DO-NOT-MODIFY-THIS-FILE: format=1; content-sha256=a7205ec1f1749acd8f9e32e85660d77d8ccb60a0d038bb63fb4b015496f23d07
 from libc.stdint cimport uint32_t, uint64_t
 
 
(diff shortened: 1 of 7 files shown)


cuda_bindings/cuda/bindings/cyruntime.pxd
```diff
@@ -1254,6 +1254,20 @@ cdef extern from 'library_types.h':
         CUDA_EMULATION_STRATEGY_EAGER
     ctypedef cudaEmulationStrategy_t cudaEmulationStrategy
 
+cdef extern from 'library_types.h':
```
mdboom, line 1257: Looks like maybe there is a non-deterministic ordering bug in the generator here.  Not to fix now, but I will file a bug over there.

cuda_bindings/cuda/bindings/runtime.pyx
```diff
@@ -17242,21 +17296,30 @@ cdef class cudaExternalSemaphoreWaitNodeParamsV2:
...
+                free(self._paramsArray)
+                self._paramsArray = _paramsArray_new
                 self._paramsArray_length = <size_t>len(val)
-                self._pvt_ptr[0].paramsArray = self._paramsArray
-            for idx in range(len(val)):
-                string.memcpy(&self._paramsArray[idx], (<cudaExternalSemaphoreWaitParams>val[idx])._pvt_ptr, sizeof(cyruntime.cudaExternalSemaphoreWaitParams))
+                self._pvt_ptr[0].paramsArray = _paramsArray_new
+            else:
+                for idx in range(len(val)):
+                    string.memcpy(&self._paramsArray[idx], (<cudaExternalSemaphoreWaitParams>val[idx])._pvt_ptr, sizeof(cyruntime.cudaExternalSemaphoreWaitParams))
```
mdboom, line 17322: We really should refactor this out into a function to remove all this duplication, but the legacy generator makes that hard and probably not worth the effort for something that's being deprecated.

mdboom (approved):

mdboom (approved):

rparolin:
Closing as superseded by #2423.

The setter fix in this PR already landed on `main` via #2423 ("Prepare for CTK 13.4"), which regenerated the bindings with the same allocate-fill-swap ordering. Verified identical coverage on `main`: 14 of 18 resize branches in `driver.pyx` and 7 of 11 in `runtime.pyx` now carry the strong-exception-guarantee pattern -- the same counts this PR produced.

This branch is a CTK 13.3 regen, so rebasing it onto 13.4 `main` would revert the 13.4 prep. Nothing here is worth carrying forward.

PR #2417, opened 2026-07-23
fix(cuda.bindings): Make param_packer.feed() free-threading-safe via import-time init

### LINKED ISSUE #2416: Umbrella issues with free-threading and Cython bindings
*(Global, non-restarting numbering. "Agents/methods" shows cross-validation: `SS`=shared-state,
`AT`=atomics, `UA`=unsafe-apis, `LD`=lock-discipline, `MAN`=manual review, `HIST`=git history.)*

### RACE Findings (fix immediately) — 2

| # | Finding | File:Line | Severity | Agents/methods |
|---|---------|-----------|----------|--------|
| 1 | **Global `std::map m_feeders` mutated with no lock** — `feed()` calls `.find()` while `populate_feeders()` calls `operator[]`. Two threads launching **distinct** kernels concurrently corrupt the map (rehash during concurrent insert → heap corruption). Violates the "distinct objects cannot corrupt the interpreter" guarantee. | `param_packer.h:62` (decl), `:135` (find), `:138` (insert) | **CRITICAL** | SS, AT, UA, MAN |
| 2 | **Unsynchronized lazy-init of 18 ctypes globals** — `feed()` guard `if (ctypes_c_int==nullptr) fetch_ctypes();` lets two threads both run `fetch_ctypes()`, tearing writes to `ctypes_module` + 17 `ctypes_c_*` pointers and **leaking** a strong ref from `PyImport_ImportModule`. Also violates the codebase's own documented static-init/GIL deadlock rule if "fixed" naively with a magic static. | `param_packer.h:133` (guard), `:11-29` (globals), `:31-58` (`fetch_ctypes`) | **HIGH** | SS, AT, MAN |

**Reachability (verified — Findings 1 & 2 are live, not theoretical):** `param_packer.pxd:7` exposes
`int feed(...)`; it is called from the **kernel-launch parameter-packing loop** at
`cuda_bindings/cuda/bindings/_lib/utils.pxi:89`
(`size = param_packer.feed(self._ckernelParams[idx], value, ctype)`); the module is Cythonized with
**`freethreading_compatible=True`** (`cuda_bindings/build_hooks.py:227` → `Py_MOD_GIL_NOT_USED`), and no
lock is held around `feed()`. Launching kernels from multiple host threads (a normal CUDA multi-stream
pattern) therefore drives concurrent `feed()` calls with **no GIL serialization** → the races are live in
the shipped `…-cpython-315t` wheels. **Structural aggravator:** `param_packer.pxd` states the header "gets
compiled into **every** Cython extension module that depends on `param_packer.pxd`," so its `static` state
exists as **independent racy copies in each consuming module** (e.g. the driver/runtime kernel-launch
extensions), not one shared instance.

**Dedup note (Finding 2):** the shared-state scanner reported the 17 `ctypes_c_*` pointers as 19
`static_type_object` / **MIGRATE** items. That label is **imprecise** — these are *cached pointers to
`ctypes`' type objects*, not user-defined `static PyTypeObject` instances needing heap-type conversion.
The real defect is **unsynchronized lazy-init of shared state**, so all 19 are collapsed here into
Finding 2 (and the separate MIGRATE section is consequently empty).

### UNSAFE Findings (fix before declaring free-threading support) — 1

| # | Finding | File:Line | Severity |
|---|---------|-----------|----------|
| 3 | **Borrowed references cached long-term + module ref leaked.** `PyDict_GetItemString` (`:41-57`) returns **borrowed** refs cast to `PyTypeObject*` and cached for process lifetime; `PyImport_ImportModule` (`:33`) stores a strong ref in `ctypes_module` that is **never released**. Safe *only* while `ctypes` stays imported; under free-threading this compounds with Finding 2 (the racy lazy path can leak/duplicate the module ref). | `param_packer.h:33`, `:41-57` | **MEDIUM** |

### PROTECT Findings (add synchronization / verify) — 3

| # | Finding | File:Line | Severity |
|---|---------|-----------|----------|
| 4 | **`mr_dealloc_cb` is a non-atomic global function pointer** — set by `register_mr_dealloc_callback` (`:1036`), read/called from `shared_ptr` deleters on arbitrary threads (`:1054-1055`). Safe *iff* registered exactly once at import (before any MR-backed device pointer exists) — **verify**. Best practice under FT: make it `std::atomic<MRDeallocCallback>` for a well-formed happens-before. | `resource_handles.cpp:1033` | **LOW–MEDIUM** |
| 5 | **`GraphNodeBox::slot_table` is a `mutable`, lazily-created cache** (`:1313`, populated `:1358`). If any **read-path** accessor triggers the lazy creation, two concurrent *reads* of the **same** graph node race on it — breaking the "concurrent reads are safe" guarantee. If it is only populated while *mutating* the graph (adding nodes — explicitly out-of-contract), it is policy-permitted. **Verify the trigger path**; if read-reachable, guard with once-init/atomic. | `resource_handles.cpp:1313` | **LOW–MEDIUM** |
| 6 | **`initialize_deferred_cleanup` uses check-then-act, not CAS** — `if (queue.load(acquire)) return; … queue.store(release)` (memory ordering is correct). Safe *iff* invoked once at module init; if reachable concurrently, two threads each allocate a `DeferredCleanupQueue` (one leaks). **Verify single-init at import.** | `resource_handles.cpp:334-340` | **LOW** |

From https://gist.github.com/clin1234/6cdb5150ba3298576ca54fa07c017517#file-cuda-python-md 

### Must Fix (FIX) — 2

| # | Finding | File:Line | Agents/Evidence |
|---|---------|-----------|-----------------|
| 1 | **C++ exception across a non-`except+` Cython boundary → `std::terminate`.** `param_packer.pxd` declares `int feed(void*, object, object)` with **no `except +`**; Cython docs (verified via context7): *extern `cdef` functions are implicitly `noexcept`* — no try/catch is generated. Yet `fetch_ctypes()` (called from `feed()` on first use) **deliberately throws `std::runtime_error`** when `import ctypes` or its dict lookup fails (`param_packer.h:35,39`), and `m_feeders[...]`/`std::function` construction can throw `std::bad_alloc`. Any such throw unwinds into Cython-generated frames with no handler → `std::terminate()` → hard interpreter abort (during e.g. interpreter finalization, embedded/restricted envs, or OOM) instead of a Python exception. The project's own `_resource_handles.pxd:173,225` shows the correct in-repo idiom (`except+ nogil`). **Fix (1 line):** declare `feed` as `except +` — or make `feed` `noexcept` by catching in C++ and returning `-1`. Note: on the throw path the pending `ImportError` from `PyImport_ImportModule` is left set; with `except +` translation it would be replaced by `RuntimeError` — prefer catching in C++ and letting the original Python error propagate. | `param_packer.pxd:7`; `param_packer.h:33-39,62,70` | error-path scanner (adjacent hit) + manual + context7 `/cython/cython` |
| 2 | **`param_packer.h` shared-state races under free-threading** (live in shipped 3.15t wheels): unsynchronized `std::map m_feeders` read/write and racy lazy-init of 18 `static` ctypes globals, plus the never-released `ctypes_module` strong ref and long-term caching of `PyDict_GetItemString` borrowed refs. This run's scanners re-confirmed via 37 `scan_module_state` hits + 3 `scan_gil_usage` hits + 1 `scan_refcounts` `potential_leak` (:33) — **all deduplicated into this one finding.** Full analysis, interleavings, and the `PyMutex`-vs-`std::mutex` deadlock discussion: **FT report Findings 1–3**. **Fix:** populate the ctypes pointers and the (finite, 6-pair) feeder table once at module import; `feed()`'s existing `size == 0` ctypes fallback (`utils.pxi:91-95`) makes prepopulation behavior-neutral. | `param_packer.h:11-29,62,133-138` | module-state (37) + gil (3) + refcounts (1) scanners + FT report |

### Should Consider (CONSIDER) — 4

| # | Finding | File:Line |
|---|---------|-----------|
| 3 | **`_ref` creator asymmetry breaks the registry design contract on the miss path.** `create_context_handle_ref` and `create_stream_handle_ref` register their boxes (with unregistering deleters); `create_event_handle_ref` and `create_kernel_handle_ref` **look up but never register**, and their miss-path boxes are **dependency-less**: `KernelBox{kernel, {}}` drops the library dependency REGISTRY_DESIGN.md says KernelBox exists to carry; `EventBox{event, false,false,false,-1, {}}` drops the context dependency and zeroes timing/IPC metadata. Reachable from graph round-trips (`graph/_subclasses.pyx:121,511,549`): after the original owned handle dies, a round-tripped event misreports `timing_enabled=false`, and a round-tripped kernel no longer keeps its library loaded. *Mitigations found:* registry-hit path preserves everything; `Kernel.from_handle` (`_module.pyx:538-556`) explicitly accepts `mod` to re-supply library lifetime — the API anticipates the gap. *Recommendation:* align the four creators (either register miss-boxes for identity dedup, or document why events/kernels deliberately don't cache degraded boxes) and state the degraded-metadata contract in REGISTRY_DESIGN.md. | `resource_handles.cpp:769-775, 1262-1268` vs `:431-447, :622-635` |
| 4 | **Tautological PyCapsule validation in the driver-API capsule extraction.** `PyCapsule_GetPointer(capsule, PyCapsule_GetName(capsule))` passes the capsule's *own* name as the expected name — the check can never fail, so a cuda_core wheel paired with an ABI-drifted cuda_bindings wheel casts a mismatched function pointer silently (call-time UB instead of import-time error). DESIGN.md documents the convenience rationale ("query it rather than hardcoding signatures"), so this is a known trade-off — but comparing against the expected signature string (even just in debug builds) would turn silent drift into a clean import failure. Same pattern at the `cynvrtc`/`cynvvm`/`cynvjitlink` extraction sites. | `_resource_handles.pyx:352-353, 358-361, 472-473, 489-490, 506+` |
| 5 | **Advertised-but-unimplemented surfaces (parity).** `Buffer.__buffer__` / `__release_buffer__` raise `NotImplementedError("WIP…")` — defining the slots makes `memoryview(buf)` fail with `NotImplementedError` rather than `TypeError`, and stubs/docs imply buffer support; also `_memoryview.copy_from` (:446), a texture path (`texture/_texture.pyx:521`), managed-memory ops on unsupported platforms, and win32 virtual memory (`_virtual_memory_resource.py:135`). Track as explicit WIP or remove the slots until implemented. | `_memory/_buffer.pyx:436,440` et al. |
| 6 | **Maintainer-flagged open threads from #2162** (commit message, Sebastian Berg): (a) `SMResource` split-mutex — fix added `self._split_mutex` but the author records residual uncertainty ("This is thread-unsafe. But I am honestly not sure…") about CUDA-internal safety; (b) explicit TSan gap: "Testing with thread-sanitizer might flush out some…"; (c) `cuda-pathfinder`'s `load_nvidia_dynamic_lib` `functools.cache` flagged for review. These are the maintainers' own TODOs — carrying them here so they don't get lost. | `_device_resources.pyx:498-505`; pathfinder |

### Tensions

- **Register-on-miss vs. don't-cache-degraded-boxes (Finding 3):** registering event/kernel miss-boxes would restore identity-map dedup but would *cache wrong metadata* that later lookups would trust; not registering (current behavior) creates a fresh degraded box per round-trip but leaves the registry clean for a future owned registration (`map_[key] = h` overwrites). Neither is strictly better — which is exactly why the choice should be written down; today the code silently disagrees with REGISTRY_DESIGN.md's stated purpose.
- **Capsule convenience vs. ABI safety (Finding 4):** the tautological name check is what lets cuda_core avoid hardcoding cydriver's signature strings; fixing it costs a maintained signature table. Debug-only validation is the compromise.
- **cppcheck style vs. C++ idiom:** cppcheck wants constructors on `EventBox` (4 warnings); the codebase's aggregate-init-only style is safer than a constructor here (verified all 3 sites fully initialize). Dismissed rather than "fixed."

### Policy Decisions (POLICY) — 3

| # | Finding |
|---|---------|
| 7 | **abi3/Limited API is not claimed and should not be pursued now.** The Cython layer + per-version wheels (incl. `cp315t`) are incompatible with today's limited API anyway (free-threaded builds have no stable ABI yet); revisit only if wheel-matrix cost forces it. |
| 8 | **Subinterpreters are unsupported by design.** Process-wide C++ singletons (six `HandleRegistry` instances, `DeferredCleanupQueue`, driver fn-pointer capsules) are shared across interpreters; per-interpreter isolation would require moving them into per-module state. No action unless subinterpreter support becomes a goal. |
| 9 | **Code-removal opportunities: none at the current `>=3.10` floor.** When the floor reaches 3.13/3.14: `PyDict_GetItemRef` (3.13) and `PyImport_ImportModuleAttrString` (3.14) replace exactly `param_packer.h`'s `PyImport_ImportModule` + `PyDict_GetItemString`-borrowed-ref chain (~5-8 lines saved and the borrowed-ref hazard of Finding 2 eliminated structurally). `PyModule_AddType` (3.10) is already available but the project defines no hand-written types. |

From https://gist.github.com/clin1234/2469758160721f4f991110167017e4d3#file-cuda-python-md

clin1234:
## Description

Restructure param_packer.h so the ctypes type pointers and the feeder table are built once in a new init_param_packer(), called at module import (from utils.pxi) while single-threaded. feed() becomes a read-only lookup on a never-mutated std::map, so concurrent kernel launches from multiple host threads no longer race the map or the ctypes lazy-init -- races that were live in the cp315t (Py_MOD_GIL_NOT_USED) wheels. init_param_packer is declared 'except +' so a failed 'import ctypes' surfaces as a Python exception at import instead of std::terminate across the Cython boundary, and feed() is now non-throwing. The ctypes module strong ref is kept deliberately (it keeps the borrowed type pointers valid).

~Harden resource_handles.cpp: make mr_dealloc_cb std::atomic with acquire/release ordering, and document the verified single-init of initialize_deferred_cleanup() and the mutation-only lazy path for GraphBox::slot_table (both already safe; no behavior change).~

closes #2416


cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -1072,8 +1083,9 @@ DevicePtrHandle deviceptr_create_with_mr(CUdeviceptr ptr, size_t size, PyObject*
         [mr, size](DevicePtrBox* b) {
             GILAcquireGuard gil;
             if (gil.acquired()) {
-                if (mr_dealloc_cb) {
-                    mr_dealloc_cb(mr, b->resource, size, b->h_stream);
+                if (MRDeallocCallback cb =
+                        mr_dealloc_cb.load(std::memory_order_acquire)) {
```
seberg: I am not sure I am following, this seems like unnecessary hardening?

If this was a general pattern, I might not mind, but I think we have many of these globals that we assume are initialized once at module init time and if that is the only time it is set, it is fine?

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -1375,6 +1387,16 @@ GraphBox* get_box(const GraphHandle& h) noexcept {
...
+struct GraphBox {
+    CUgraph resource;
+    GraphHandle h_parent;                  // Keeps parent alive for child/branch graphs
+    // Lazily created; owned by the graph's user object. The lazy creation is
+    // reached only from graph_set_slot() (a graph *mutation*), never from a
+    // read-only accessor, so two concurrent *reads* of the same graph cannot
+    // race on it -- concurrent mutation of one graph is the user's
+    // responsibility per docs/source/concurrency.rst. No guard needed.
+    mutable GraphSlotTable* slot_table = nullptr;
+};
```
seberg: Looks like a merge conflict.  Whatever you wanted to fix doesn't exist anymore (or the comment added maybe?).

cuda_bindings/cuda/bindings/_lib/param_packer.pxd
```diff
@@ -4,4 +4,10 @@
...
 # Cython extension module that depends on param_packer.pxd.
 cdef extern from "param_packer.h":
-    int feed(void* ptr, object o, object ct) except? -1
+    # Populates ctypes pointers + the feeder table once, at import. `except +`
+    # translates a C++ throw (e.g. failed `import ctypes`) into a Python
+    # exception during module import, preserving any already-set Python error.
+    void init_param_packer() except +
+    # Hot path: read-only lookup, does not throw -> intentionally left as an
+    # implicit-noexcept extern (no `except +` needed).
+    int feed(void* ptr, object o, object ct)
```
seberg: This looks like a regression?

You are right, this function can get away without the `except +` (if we can, I wouldn't mind encoding that on the C++ side for clarity, though).

But, it clearly needs to have the `except -1`!  (you can drop the `?` not that it matters)
I don't know if your analysis was based on some old branch where Python errors were not anticipated?!

cuda_bindings/cuda/bindings/_lib/param_packer.pxd
```diff
@@ -4,4 +4,10 @@
 # Include "param_packer.h" so its contents get compiled into every
 # Cython extension module that depends on param_packer.pxd.
 cdef extern from "param_packer.h":
-    int feed(void* ptr, object o, object ct) except? -1
+    # Populates ctypes pointers + the feeder table once, at import. `except +`
+    # translates a C++ throw (e.g. failed `import ctypes`) into a Python
+    # exception during module import, preserving any already-set Python error.
+    void init_param_packer() except +
```
seberg:
```suggestion
    void init_param_packer() except +*
```
(For me, I would personally like if it actually propagated python errors right away.)

cuda_bindings/cuda/bindings/_lib/param_packer.h
seberg: This is missing the error check :/!

cuda_bindings/cuda/bindings/_lib/param_packer.h
```diff
@@ -30,6 +30,11 @@ PyLong_AsInt(PyObject *obj)
 }
 #endif
 
+// Strong reference to the ctypes module, acquired once at import in
+// init_param_packer() and intentionally never released. This ref is
+// load-bearing: the ctypes_c_* pointers below are *borrowed* references into
+// the ctypes module dict, and stay valid only while this strong ref keeps the
+// module (and therefore its type objects) alive. Do not Py_DECREF it.
 static PyObject* ctypes_module = nullptr;
```
seberg: If we re-wire everything here, can we just remove this global and make all type objects strong references instead?  That seems much clearer.

(Unfortunately, I guess we are not vendoring pythoncapi-compat, so we don't have `PyDict_GetItemStringRef`, here it's fine, so let's not worry, although I could see that we'll just need that in the future...)

cuda_bindings/cuda/bindings/_lib/param_packer.h
seberg, line 89: I guess you didn't touch it, so maybe... but all of this feels rather ugly, we could just propagate the Python error (there is absolutely no reason not to).

Since there is a new init time helper in cython, that could also just have `except +*`.

(I.e. also then we can get rid of that long new comment...)

seberg (commented):
Thanks, for finding/fixing these issues.  Moving the `ctypes` import/populating to import time seems OK to me (at worst 5% seems like additional import time).
I hate to ask it, but touching this, it would be nice to clean up the param packer initialization a bit more.

That said, the other two changes seem a bit strange to me and maybe because of merge conflicts?  (I.e. the graphbox clearly won't compile)

clin1234:
pre-commit.ci autofix

seberg:
@clin1234 gentle ping just in case. But the new push has just new sets of merge issues unfortunately (with important changes now also vanishing?). Could you clean it up and look at the other comments?

clin1234:
> @clin1234 gentle ping just in case. But the new push has just new sets of merge issues unfortunately (with important changes now also vanishing?). Could you clean it up and look at the other comments?

Made a messy rebase, but the current rebase is squashed to only one commit. Mind reviewing again?

cuda_bindings/cuda/bindings/_lib/param_packer.pxd
seberg: Still applies once you remove the blanket `PyErr_Clear()`.

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -1060,10 +1063,14 @@ DevicePtrHandle deviceptr_create_mapped_graphics(
 // MemoryResource-owned Device Pointer Handles
 // ============================================================================
 
-static MRDeallocCallback mr_dealloc_cb = nullptr;
+// Registered exactly once, from the module body of _memory/_buffer.pyx (at
+// import, single-threaded), but read from shared_ptr deleters running on
+// arbitrary threads. Atomic with release/acquire ordering so those readers have
+// a well-formed happens-before edge to the registration rather than a data race.
+static std::atomic<MRDeallocCallback> mr_dealloc_cb{nullptr};
```
seberg: I am still not following: I can't really see a world where we can create a MR before importing (and thus setting this).  So I don't think this needs an atomic, it is just like all other "initialized at import time" variables.

cuda_core/cuda/core/_cpp/resource_handles.cpp
```diff
@@ -318,8 +318,11 @@ void enqueue_cleanup(void* item) noexcept {
 
 }  // namespace
 
-// Module initialization calls this once with the GIL held, which serializes
-// the check, allocation, and publication below.
+// Called exactly once, from the module body of _resource_handles.pyx (at
+// import, single-threaded). The check-then-act below is therefore an
+// idempotency guard, not a racy lazy init -- concurrent callers would each
+// allocate a queue and all but one would leak. Do not add a second call site
+// without converting this to a compare-exchange.
```
seberg: Hmmmm, do you know what happens right now if we reload the Cython module?  Does Cython run this again?
I am somewhat tempted to just make it an `assert();` if we really are sure it is called exactly once (even in ridiculous "should not happen" reload scenarios -- where I personally wouldn't care about leaking until we approach subinterpreter readiness).

cuda_bindings/cuda/bindings/_lib/param_packer.h
```diff
@@ -50,138 +67,165 @@ static PyTypeObject* ctypes_c_float = nullptr;
...
+// and never mutated afterwards; see the thread-safety contract above.
+static std::map<std::pair<PyTypeObject*,PyTypeObject*>, std::function<int(void*, PyObject*)>> m_feeders;
+
+// Look a type up in the ctypes module dict and return a *strong* reference.
+// PyDict_GetItemString returns a borrowed reference; we upgrade it to a strong
+// one so the cached PyTypeObject* stays valid for the process lifetime even if
+// the ctypes module itself is later dropped from sys.modules.
+static PyTypeObject* fetch_ctypes_type(PyObject* ctypes_dict, const char* name)
+{
+    PyObject* type_obj = PyDict_GetItemString(ctypes_dict, name);  // borrowed
```
seberg: Actually, why not just pass the module and use `PyObject_GetAttrString()`, then we also get rid of the `INCREF` below.

cuda_bindings/cuda/bindings/_lib/param_packer.h
```diff
@@ -50,138 +67,165 @@ static PyTypeObject* ctypes_c_float = nullptr;
...
+    if (type_obj == nullptr)
+        throw std::runtime_error(std::string("Cannot find ctypes.") + std::string(name));
```
seberg: FWIW, it was an existing pattern, but in this instance, I would prefer to just return directly and *not* set a C++ exception.
To not spread over multiple comments, chain the thing in `fetch_ctypes`:
```
    bool success = (
        ctypes_c_char = fetch_ctypes_types() &&
        ctypes_c_bool = ... &&
        ...);
    Py_DECREF(ctypes_module);
    return success;
```
in `init_param_packer` make it:
```
if (!fetch_ctypes()) {
    return;
}
```
and on the Cython side, type it as `init_param_packer() except +*` (but you already did that).
(Same for the `ctypes_module` import to avoid the Python exception -> C++ exception -> Python exception.  I'll admit, I don't hate the pattern, but seems like it was more useful for the deferred init design.)

cuda_bindings/cuda/bindings/_lib/param_packer.h
```diff
@@ -30,7 +31,23 @@ PyLong_AsInt(PyObject *obj)
...
+// Python's import machinery guarantees a module body runs to completion before
+// any other thread can call into the module. Afterwards the state is read-only,
+// which is what makes the hot feed() path safe to call concurrently without a
+// lock: concurrent std::map::find() on a map that is never mutated is safe by
+// the C++ standard.
+//
+// Do not reintroduce lazy initialization here. This header is textually
+// compiled into every extension module that cimports param_packer.pxd, so each
+// such module owns an independent copy of this state and would need its own
+// racy lazy path.
```
seberg: Maybe it's just me, but can you cut this down to just the bare minimum 1-2 (feels like an agent "explain the change" to me).  I am not sure, but maybe just: Statics must be initialized at Python import time via `init_param_packer()` which happens when including `utils.pxi`.
This includes the `m_feeders` maps as it must not be mutated from threads.

cuda_bindings/cuda/bindings/_lib/param_packer.h
```diff
@@ -50,138 +67,165 @@ static PyTypeObject* ctypes_c_float = nullptr;
...
 static PyTypeObject* ctypes_c_void_p = nullptr;
 
+// (target type, source type) -> writer. Built in full by init_param_packer()
+// and never mutated afterwards; see the thread-safety contract above.
+static std::map<std::pair<PyTypeObject*,PyTypeObject*>, std::function<int(void*, PyObject*)>> m_feeders;
+
+// Look a type up in the ctypes module dict and return a *strong* reference.
+// PyDict_GetItemString returns a borrowed reference; we upgrade it to a strong
+// one so the cached PyTypeObject* stays valid for the process lifetime even if
+// the ctypes module itself is later dropped from sys.modules.
```
seberg: Same with this comment, the function is practically self explanatory, one sentence makes sense: Helper to fetch a strong reference of the ctypes type.
(or similar)?

cuda_bindings/cuda/bindings/_lib/param_packer.h
```diff
@@ -50,138 +67,165 @@ static PyTypeObject* ctypes_c_float = nullptr;
...
-                int v = PyLong_AsInt(value);
-                if (v == -1 && PyErr_Occurred())
-                    return -1;
-                *((int*)ptr) = v;
-                return sizeof(int);
-            };
-            return;
+            PyErr_Clear();
+            return 0;  // decline: let the ctypes fallback define the result
         }
```
seberg: Sorry, but can we split this out into a new issue? This is an API change and I doubt it is user-friendly.  I realize that for many integer types we _don't_ have a strict check, however, I suspect that actual kernels defined here are basically limited to this list, so in _practice_ we have strict integer conversion and we probably want that, IMO.

So this is an issue, but I would rather fix it in the opposite direction also matching NumPy, etc. (For raw Python integers coerce safely, if it's already typed, e.g. NumPy int scalar that could go either way, but in this case errors also seem good.)
The one actual annoyance with this is passing `-1` for an unsigned value as a special value.

cuda_bindings/cuda/bindings/_lib/param_packer.h
```diff
@@ -50,138 +67,165 @@ static PyTypeObject* ctypes_c_float = nullptr;
...
+// convert faithfully, which routes the caller to its ctypes fallback and gets
+// ctypes' own answer. Returning a negative size instead would be unsafe: the
+// caller tests only `if size == 0` and then does `data_idx += size`, so a -1
+// would skip the fallback and rewind the pack buffer, corrupting the next
+// argument. Any error the probe set is cleared -- feed() must not leave an
+// exception pending, since it is declared implicitly noexcept to Cython.
+static void populate_feeders()
 {
-    if (target_t == ctypes_c_int)
+    m_feeders[{ctypes_c_int, &PyLong_Type}] = [](void* ptr, PyObject* value) -> int
```
seberg, line 95: Just a thought, but I think we can add `noexcept` here (even if `std::function` will erase it).  That'll remind us of it and maybe also warn us if we should change it.
(I guess `feed` itself maybe cannot be marked, because of the `std::function` type erasure.)

(I honestly don't see much of a reason to not do `except -1*` on the Python side, but as these are all trivial it's probably OK as is.)

seberg (commented):
Thanks, similar comments as before and I would like to clean things up to avoid the C++ exceptino dance.
For the conversion helpers, I am not comfortable doing this here (and I don't think it should be done at all).  It's OK to use `int feed() except -1` and just keep what we had

(Some of the comments feel a bit long to me personally.)

clin1234:
@seburg, I implemented most of your suggestions on `param_packer.h`

clin1234:
@seberg Ping

seberg:
/ok to test [1fad6c4](https://github.com/NVIDIA/cuda-python/pull/2417/commits/1fad6c49a15aefb73dad72961f9e98a776a75a3c)

seberg (approved):
Sorry, I was out on vacation.  Your changes accidentally removed actually assigning the ctypes type objects entirely (which would lead to always using the fallback for those, I think).

Pushed a fix for that and a few other smaller cleanups (as well as merged main).

If you have a brief look that would be nice, but I think it should be good to go now.

clin1234:
TYSM, LGTM

seberg:
Thanks @clin1234 for finding and fixing this.  This closed your issue, I _think_ that was correct, because I believe other issues were either things to confirm being OK, or already fixed (as evidenced by how hard it was to resolve conflicts).

But if you have similar reports/finds I am all ears!

PR #2424, opened 2026-07-25
[doc-only] docs: document setuptools-scm clone requirements for source builds

### LINKED ISSUE #2400: Include requirements for setuptools-scm success in the CONTRIBUTING docs
This amounts to "you need a clone with tags, at least as far back as the latest tag".  We should write a recommendation about how to get there in the CONTRIBUTING guide.

bharatr21:
## Description

Closes #2400  

Every package in this repo derives its version from git tags via `setuptools-scm`, 
but there's no mention in the docs, and the `git clone` commands in our own install guides 
that are intended to be copied by contributors produce broken builds.

Two of the three bad-clone cases fail **silently**:

| Clone | Resolved version |
| --- | --- |
| `--depth 1` | `0.1.dev1+g0d22cb444` — no error |
| `--no-tags` | `0.1.dev2114+g0d22cb444` — no error |
| source zip (no `.git`) | `LookupError` — the only case that errors |

A partial `--depth` is worse than none. With the nearest `cuda-pathfinder-v*` tag
10 commits back, a `--depth 20` clone gives:
| Package | Version | Outcome |
| --- | --- | --- |
|`cuda_pathfinder` | 1.6.1.dev10+g0d22cb4 | Correct, tag falls inside depth |
|`cuda_core`       | 0.1.dev20+g0d22cb444 |  Wrong, no error |
| `cuda_bindings`  |  0.1.dev20+g0d22cb444| Wrong, no error |

So "it built and the version looked right" is not evidence the clone is sound,
and no fixed `--depth` stays correct as commits land between releases.

### Changes

- **`CONTRIBUTING.md`**: new "Cloning the repository" section (plus TOC entries)
  covering the per-package tag patterns, why `root = ".."` means the whole repo is
  needed, the recommended treeless clone, recovery for shallow clones and stale
  forks, and the symptoms of each failure mode.
- **Install guides**: updated the four `git clone` commands in the `cuda_core` and
  `cuda_pathfinder` guides to `git clone --filter=blob:none`, and added a pointer to
  the new section from all three. `cuda_bindings` has no clone command, so it gets a
  third requirement bullet alongside the existing CTK header/static-lib ones.

The recommended clone matches what CI already does (`fetch-depth: 0` +
`filter: blob:none` in `build-wheel.yml`): full commit graph and tags, no historical
blobs. It took 6.2s against `github.com`.

### Verification

- Treeless clone from GitHub fetches all 87 tags, is not shallow, and has the full
  2114-commit graph; all four packages resolve correct versions
  (`13.3.2.dev128`, `1.1.1.dev32`, `1.6.1.dev10`) via `python -m setuptools_scm`,
  so each package's real `pyproject.toml` config is exercised.
- `--depth 1`, `--depth 20`, `--no-tags`, and no-`.git` failure modes all reproduced
  against the real remote.
- `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_CUDA_CORE` and `git fetch --unshallow --tags`
  recovery paths confirmed working before documenting them.
- Full `pre-commit` hook set passes on all four files with no reformatting; RST
  parses clean and lychee reports no broken links.

## Checklist


CONTRIBUTING.md
```diff
@@ -34,6 +38,113 @@ Thank you for your interest in contributing to CUDA Python! Based on the type of
...
+
+This makes a partial `--depth` actively dangerous, because it can succeed for one
+package while quietly failing for the others. A `--depth 20` clone taken while
+the nearest `cuda-pathfinder-v*` tag was 10 commits back gives:
+
+```text
+cuda_pathfinder    -> 1.6.1.dev10+g0d22cb4     # correct
+cuda_core          -> 0.1.dev20+g0d22cb444     # wrong, no error
+cuda_bindings      -> 0.1.dev20+g0d22cb444     # wrong, no error
+```
```
mdboom: This is way too much detail.


```suggestion
```

CONTRIBUTING.md
```diff
@@ -34,6 +38,113 @@ Thank you for your interest in contributing to CUDA Python! Based on the type of
...
+$ git clone --filter=blob:none https://github.com/NVIDIA/cuda-python.git
+```
+
+This is a *treeless* clone: it fetches the complete commit graph and all tags
+(which is what `setuptools-scm` needs) while skipping historical file contents
+(which it does not). It is substantially faster than a plain clone and is
+exactly what CI uses — see the `fetch-depth: 0` plus `filter: blob:none`
+checkout settings in `.github/workflows/build-wheel.yml`.
+
+Do **not** use `--depth`, `--shallow-since`, or `--no-tags`.
```
mdboom: I think we just recommend the defaults, tbh.
```suggestion
```

mdboom (changes requested):
The `--filter=blob:none` is not necessary -- it just reduces the size of the clone while still making it buildable.  It's perfectly fine to just use the "default" clone behavior of `git`, so I think `--filter=blob:none` just confuses things and should not be mentioned.

If we are concerned about repository size, we should periodically flatten the history of the `gh-pages` branch.  (That is something we would do in automation -- no need for most developers to know about it).

bharatr21:
Applied the suggestions. Still leaving in the actual fixes for fetching upstream tags etc. as I believe they will still be useful for contributors

mdboom (approved):

PR #2426, opened 2026-07-27
coverage: capture what the Windows coverage crash actually is

rluo8:
## Description

This is a draft pr for diagnostics.  Please do not merge.

The `Run cuda.core tests (with 8MB stack)` step of `CI: Coverage` always exited 139. Its log is three lines — the traced command, `Segmentation fault`, exit 139 — with no pytest header, no `collected N items`, no test ids and no faulthandler traceback. I couldn't reproduce this failures on my local systems. 

Three diagnostics, no behaviour change:

- **`PYTHONFAULTHANDLER`** / **`PYTHONUNBUFFERED`** — names the exception and dumps every thread's stack, and unlike pytest's built-in plugin it reaches tests that spawn their own `python -c` children.
- **`--isolate`** — runs pytest in a child process so a surviving parent reports the raw NTSTATUS instead of the shell's translation of it (Git Bash renders an access violation as `Segmentation fault` / 139).  The child's status is passed through unchanged.
- **`pytest_crashlog`** — line-buffered progress log whose last line survives a native fault, localising it to interpreter startup, conftest import, a module's collection, a test, or shutdown. Opened before `pytest.main`, since pytest imports the conftests — which pull in  `cuda.core` — before `pytest_configure`. A no-op unless  `PYTEST_CRASHLOG` is set.


rluo8:
Hi @mdboom , could you please help trigger a coverage run with this draft PR ?  The Windows coverage run always failed when running cuda.core tests. This is to add some debug logs for triage. 
Thanks!

rluo8:
Investigation done. Close.

PR #2434, opened 2026-07-28
chore: refactor pyx files for readability

juenglin:
Refactor: tidy cdef declarations and struct init in .pyx files

Declare cdef locals at their point of initialization rather than at the
top of the function, and replace field-by-field struct setup with
Cython's struct-initializer syntax.

Only complete initializers are converted. Cython does not zero-fill
omitted members, so a partial initializer would leave them holding stack
garbage; sites that depend on a preceding memset are left unchanged.
Where every member is now supplied, the redundant memset is dropped.

No behavior change.

### DIFF
--- a/cuda_bindings/tests/cython/test_ccudart.pyx
+++ b/cuda_bindings/tests/cython/test_ccudart.pyx
@@ -59,11 +59,8 @@ cdef extern from *:
 
 def test_ccudart_interoperable():
     # struct
-    cdef dim3 oldDim, newDim
-    oldDim.x = 1
-    oldDim.y = 2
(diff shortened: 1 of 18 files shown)


Andy-Jost (commented):
Thanks, I prefer the new style.

Someone following the new pattern without understanding Cython's partial initialization limitation could accidentally leave junk memory. Should we add `-Werror` to the Cython build to guard against that?

It appears there is currently only one Cython warning:

```
  cuda/core/system/_process.pxi:10:31:
  Strings should no longer be used for type declarations.
```

Andy-Jost (approved):

Issue #2439, opened 2026-07-28
cuda.core: validate ctypes host callback signatures


Andy-Jost:
## Background

cuda.core accepts any `ctypes._CFuncPtr` as a CUDA host callback and casts its address to `CUhostFn` without validating the declared ctypes signature. CUDA requires a callback compatible with `void (*)(void*)`; passing an incompatible return type, argument list, or calling convention can cause undefined behavior when CUDA invokes it.

The shared callback resolver is used by both host callback node construction and `HostCallbackNode.update()`.

## Scope

- Validate ctypes callback argument and return types against the `CUhostFn` ABI.
- Handle platform-specific calling-convention requirements, including Windows.
- Define behavior for untyped or explicitly cast `_CFuncPtr` objects without rejecting compatible existing use cases unnecessarily.
- Apply the same validation to construction and update paths.
- Add positive and negative tests, including an incompatible `CFUNCTYPE`.

## Acceptance criteria

- Clearly incompatible ctypes callback signatures are rejected before being passed to CUDA.
- Valid callbacks continue to work on supported platforms.
- The error identifies the required callback signature.

mdboom:
`cuda-bindings` have a related issue.  Callbacks usually take an `intptr_t` as a pointer to a callback function, and there is no type-checking done.

PR #2440, opened 2026-07-28
fix(core): restore access descriptors when a VMM grow rolls back

aryanputta:
## Description

Defect 1 of #2388, continuing the per-defect split from #2407 (defect 2) and #2418 (defect 4).

`_grow_allocation_slow_path()` unmaps the old VA range before remapping the physical memory into the new reservation. Access permissions live on the mapping, so that unmap drops the resource's access descriptors. The `_remap_old` rollback restored the mapping but not the descriptors, so a buffer recovered from a failed grow was mapped and yet not accessible: it faulted on its next access until `cuMemSetAccess` was run again.

`_remap_old` now reapplies the access descriptors to the old range after the remap succeeds. It stays inside the same best-effort `try`, so a failure while restoring access cannot mask the error that triggered the rollback.

The descriptor list is now built once above the transaction and shared by the forward path and the rollback, rather than built only at the point of use near the end. The rollback restores exactly the descriptors the grow was in the middle of applying, and there is one source of truth for them.

## Test

`test_vmm_allocator_grow_slow_path_rollback_restores_access` fails `cuMemSetAccess` over the new range to drive the transaction into rollback, then asserts that the old range is remapped, that its access is reapplied, that the two happen in that order, and that the failed reservation is still freed. Reverting the three added lines fails the test on the restored-access assertion.

## Checks

- `ruff check`, `ruff format`
- `mypy-cuda-core`
- `cython-lint`, `stubgen-pyx`
- `lychee`, SPDX headers, rst hooks


aryanputta:
Head is now `a9be4049`. Merged current `main` to clear the conflict; the PR is mergeable again.

The only conflict was in `cuda_core/docs/source/release/1.2.0-notes.rst`, where entries for #2517, #2439, and #2389 landed in the same list. Resolution keeps all of them and places this PR's entry last. The fix in `_virtual_memory_resource.py` and the regression test are unchanged.

This still covers defect 1 of #2388 only. Workflows here have never run, so `test_vmm_allocator_grow_slow_path_rollback_restores_access` has not executed anywhere. It is a monkeypatched driver test, so it does not need a GPU beyond the `virtual_memory_management_supported` guard.

Could someone run CI on this when convenient?

Andy-Jost:
Thanks for the PR. VMM work is consolidated under #2906.

#2880 removes the code this edits: the slow path no longer unmaps the old range before the new mapping exists, so a failed grow leaves the old buffer untouched and there is no access to restore. Closing as superseded. #2388 item 1 will be closed with a note that credits this report. #2906 lists what is open for contribution.

PR #2441, opened 2026-07-28
chore: cython warnings as errors in cuda.core

juenglin:
This change follows up on [review feedback](https://github.com/NVIDIA/cuda-python/pull/2434#pullrequestreview-4800625822) in an earlier PR. The suggestion was to treat cython warnings as errors.

### DIFF
--- a/cuda_core/cuda/core/system/_process.pxi
+++ b/cuda_core/cuda/core/system/_process.pxi
@@ -7,7 +7,7 @@ class ProcessInfo:
     """
     Information about running compute processes on the GPU.
     """
-    def __init__(self, device: "Device", process_info: nvml.ProcessInfo):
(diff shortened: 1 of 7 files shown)


cuda_core/cuda/core/system/_process.pxi
```diff
@@ -7,7 +7,7 @@ class ProcessInfo:
     """
     Information about running compute processes on the GPU.
     """
-    def __init__(self, device: "Device", process_info: nvml.ProcessInfo):
+    def __init__(self, device: object, process_info: nvml.ProcessInfo):
```
mdboom: Does this make our type annotations less helpful?

juenglin: Should be `str`, not `object`.

Andy-Jost (approved):

PR #2442, opened 2026-07-29
pixi: add root lint tasks

rparolin:
## What

The root `pixi.toml` declared `ruff` as a dependency but no task ever used it, so there was no repo-wide entry point for linting. Contributors had to know to invoke `pre-commit` directly, and `AGENTS.md` tells agents to prefer pixi tasks over raw tool invocations.

Adds a dedicated `lint` feature/environment with three tasks:

```
pixi run lint          # pre-commit run --all-files (the pre-commit.ci equivalent)
pixi run lint-staged   # pre-commit run (staged files only)
pixi run lint-install  # install the git hook
```

## Why route through pre-commit instead of calling ruff directly

This was the interesting part. `.pre-commit-config.yaml` applies per-hook `exclude` patterns — generated `*.pyi` files and `cuda/bindings/_internal/_fast_enum.py`. A bare `ruff check --fix .` at the root does **not** honour those, and reformats 232 files that CI intentionally leaves alone (I hit this while developing the change). Routing every task through `pre-commit` keeps the task and pre-commit.ci in agreement by construction, with no second copy of the exclude list to drift.

`ruff` stays in the environment for ad-hoc use (`pixi run -e lint ruff format <file>`).

## Notes for reviewers

- The feature uses `no-default-feature = true`, so the linters are not solved into the `cu12`/`cu13`/`docs` environments.
- Side effect: moving `ruff` out of `[dependencies]` leaves those three environments with an empty package set. They exist only to dispatch into the sub-package manifests, so this is intentional — but it does mean `pixi run -e cu13 ruff ...` no longer resolves. Happy to keep `ruff` in `[dependencies]` instead if you'd rather preserve that.
- `pixi.lock` migrates from format v6 to v7 as a side effect of re-locking with pixi 0.73. The other three workspace lockfiles are already on v7, so this brings the root in line.

## Verification

`pixi run lint` runs the full suite green with zero file modifications — SPDX, generated-file seals, cuda-version pins, mypy, cython-lint, lychee, actionlint all pass.

---

| Check | Status | Tool | Details |
|-------|--------|------|---------|
| Secrets | SKIP | Pulse Secret Scanner | Image pull denied — not authenticated to `gitlab-master.nvidia.com:5005` (needs a PAT with `read_registry`) |
| SAST | PASS | Semgrep 1.157.0 | 0 high / 0 medium / 0 low over the changed files (bundled ruleset, severity filter ERROR) |
| License | SKIP | pip-licenses | `pixi.lock` is a conda+PyPI lockfile; pip-licenses reads only installed Python distribution metadata and has no `pixi.lock` parser. BlackDuck + OSRB remain authoritative |
| Container | N/A | — | No container files in this change |
| Security Review | PASS | inline | No auth, crypto, input-validation, or privilege-boundary code in this change |

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -1,300 +1,164 @@
-version: 6
+version: 7
+platforms:
+- name: linux-64
+  virtual-packages:
+  - __unix=0=0
+  - __linux=4.18
+  - __glibc=2.28
+  - __archspec=0=x86_64
+- name: linux-aarch64
+  virtual-packages:
+  - __unix=0=0
+  - __linux=4.18
+  - __glibc=2.28
+  - __archspec=0=aarch64
+- name: win-64
+  virtual-packages:
+  - __win=10.0
+  - __archspec=0=x86_64
 environments:
   cu12:
     channels:
     - url: https://conda.anaconda.org/conda-forge/
-    options:
-      pypi-prerelease-mode: if-necessary-or-explicit
-    packages:
-      linux-64:
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/_openmp_mutex-4.5-20_gnu.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/bzip2-1.0.8-hda65f42_9.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/ca-certificates-2026.6.17-hbd8a1cb_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/ld_impl_linux-64-2.46.1-default_hbd61a6d_102.conda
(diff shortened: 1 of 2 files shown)


rparolin:
Verified the empty-dispatcher-environment concern raised in the description: with `ruff` moved out of `[dependencies]`, `pixi run -e cu13 echo ok` still activates cleanly and `pixi run -e cu13 test-pathfinder` dispatches into the sub-manifest and runs green (1095 passed, 4 skipped). The `[activation.env]` `PIXI_ENVIRONMENT_NAME` rewrite is unaffected by the environments having an empty package set.

mdboom (changes requested):
In the interest of reducing complexity, what's the advantage of any of this once the `pre-commit` hook is installed (which should be the default for the highest safety of slipping by pre-commit)?

PR #2451, opened 2026-07-29
Fix version parsing in cuda_core; Allow new enums to be added to cuda_bindings

mdboom:
This fixes 2 bugs found when testing cuda_core against the cuda_bindings 13.4.0b1 release.

- The parsing of the cuda_bindings version assumed the version number was full of integers.  It is actually ~full of eels~ PEP 440-compliant
- The test that ensures that cuda_bindings and cuda_core enums are in sync wasn't forward compatible with the case where cuda_bindings may add more values.  This test has been fixed.

### DIFF
--- a/cuda_core/cuda/core/_utils/enum_explanations_helpers.py
+++ b/cuda_core/cuda/core/_utils/enum_explanations_helpers.py
@@ -31,6 +31,18 @@ _ExplanationTable = dict[int, str | tuple[str, ...]]
 _ExplanationTableLoader = Callable[[], _ExplanationTable]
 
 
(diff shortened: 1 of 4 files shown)


Andy-Jost (approved):
Looks good!

For the future:
The duplication of `_parse_version_triple` is due to an existing import cycle that need not be fixed in this PR.

```
  version.pyx
    → cuda_utils.pyx
      → driver_cu_result_explanations.py / runtime_cuda_error_explanations.py
        → enum_explanations_helpers.py
          → version.pyx
```

PR #2453, opened 2026-07-29
Add identity preserving pattern to critical_sections

seberg:
Split from gh-2194 just to shrink it a bit.

These commits are straight-forward and an attempt to guarantee identity results (not a single creation of the internal object). One of these ran into an issue in the tests.  It probably doesn't matter for most, but it also doesn't hurt.

To explain what is going on: Critical sections are like the GIL.  When we call arbitrary Python code they can be _suspended_ (they definitely will be suspended in `with nogil:`, i.e. when the Python thread-state is detached!).

So, the callbacks can run in multiple threads, but after that we double-check and `if self.attr is None: self.attr = obj` is all pure C code, so only one thread can hold the critical section lock for that chunk of code.

### DIFF
--- a/cuda_core/cuda/core/_memory/_buffer.pyx
+++ b/cuda_core/cuda/core/_memory/_buffer.pyx
@@ -272,8 +272,11 @@ cdef class Buffer:
     @cython.critical_section
     def ipc_descriptor(self) -> IPCBufferDescriptor:
         """Descriptor for sharing this buffer with other processes."""
+        cdef object ipc_data
         if self._ipc_data is None:
-            self._ipc_data = IPCDataForBuffer(_ipc.Buffer_get_ipc_descriptor(self), False)
+            ipc_data = IPCDataForBuffer(_ipc.Buffer_get_ipc_descriptor(self), False)
+            if self._ipc_data is None:
+                self._ipc_data = ipc_data
         return self._ipc_data.ipc_descriptor
 
     def close(self, stream: Stream | GraphBuilder | None = None) -> None:
--- a/cuda_core/cuda/core/_module.pyx
+++ b/cuda_core/cuda/core/_module.pyx
@@ -7,6 +7,7 @@ from __future__ import annotations
 cimport cython
 from libc.stddef cimport size_t
 from libc.stdint cimport intptr_t
+from libcpp.mutex cimport py_safe_call_once
 
 from collections import namedtuple
 from os import fsencode, fspath, PathLike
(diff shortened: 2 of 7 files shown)


cuda_core/cuda/core/_module.pyx
```diff
@@ -748,24 +754,28 @@ cdef class ObjectCode:
...
+            _h_library = create_library_handle_from_file(<const char*>path_bytes)
         else:
             assert_type_str_or_bytes_like(module)
             raise_code_path_meant_to_be_unreachable()
             return -1
-        if not self._h_library:
+        if not _h_library:
             HANDLE_RETURN(get_last_error())
+
+        if not self._h_library:
```
kkraus14: `critical_section` is only a publication guard here, not an exactly-once initializer. `create_library_handle_from_*()` releases the Python thread state via `GILReleaseGuard`, so two callers can both reach `cuLibraryLoad*()` before either publishes `_h_library`. This second check preserves one handle identity, but the losing `LibraryHandle` is discarded while its deleter currently does not call `cuLibraryUnload`, leaving an extra CUDA library loaded for the process. Could we use a true single-flight primitive for `_lazy_load_module` (for example Cython's `py_safe_call_once`/`py_safe_once_flag`, or an equivalent persistent lock) and add a free-threaded concurrent-first-load regression test?

kkraus14 (commented):
One concurrency concern in the lazy module loader.

cuda_core/cuda/core/_module.pyx
seberg: Thanks, good point.  I used the `safe_call_once` (I wish it was a bit prettier and didn't require `except *` but it seemed not too bad here).

The other places seem OK as is, although call-once could still make sense as a general pattern.

While checking, there were two places that had an incorrect order which I applied the minimal fix for (these use a pure-python pattern in Cython code where we could use atomics, but it didn't seem too important; but we could follow up with a more general audit).

cuda_core/cuda/core/_module.pyx
```diff
@@ -746,26 +778,8 @@ cdef class ObjectCode:
...
-        elif isinstance(module, PathLike):
-            path_bytes = fsencode(module)
-            self._h_library = create_library_handle_from_file(<const char*>path_bytes)
-        else:
-            assert_type_str_or_bytes_like(module)
-            raise_code_path_meant_to_be_unreachable()
-            return -1
-        if not self._h_library:
-            HANDLE_RETURN(get_last_error())
+        py_safe_call_once(self._load_once, _lazy_load_module_once, <void *>self)
```
kkraus14, line 782: Minor, non-blocking: this introduces `py_safe_call_once`, but the build requirement still permits Cython 3.2.0–3.2.4. This helper had a compile failure fixed in Cython 3.2.5 ([changelog](https://cython.readthedocs.io/en/latest/src/changes.html)). Could we raise the lower bound to `Cython>=3.2.5` and align the matching pixi constraints?

kkraus14 (approved):

cuda_core/cuda/core/_memory/_buffer.pyx
```diff
@@ -272,8 +272,11 @@ cdef class Buffer:
     @cython.critical_section
     def ipc_descriptor(self) -> IPCBufferDescriptor:
         """Descriptor for sharing this buffer with other processes."""
+        cdef object ipc_data
```
rparolin, line 275: Can we create a primitive that holds the double checked publishing pattern? I imagine there is going to be more places we would wish to add this in the future.

cc: @seberg

seberg: Let me think about it... I am also a bit underwhelmed with these patterns, including wondering about an `@call_once` in cython (but I am not sure that works so nicely to be worth it).

PR #2454, opened 2026-07-29
Ft testing test fixes

seberg:
Split from https://github.com/NVIDIA/cuda-python/pull/2194 with all _test_ changes.  Split out for two reasons:
1. A bit shorter.
2. By the time this is merged, the original PR may need further fixups anyway...

(adding a few inline comments)

CC @Andy-Jost


cuda_bindings/tests/test_examples.py
```diff
@@ -20,6 +20,7 @@ def test_example(example):
 
     env = os.environ.copy()
     env["CUDA_BINDINGS_SKIP_EXAMPLE"] = "100"
+    env["MPLBACKEND"] = "Agg"  # avoid plt.show() from blocking
```
seberg, line 22: I dunno if others get a pop-up for these tests, I did and with parallel testing, it might be a 100 windows :).

cuda_core/tests/graph/test_graph_definition.py
```diff
@@ -791,14 +761,16 @@ def registered(node):
 # =============================================================================
 
 
-def test_graphdef_handle_valid(sample_graphdef):
+def test_graphdef_handle_valid(init_cuda):
     """GraphDefinition has a valid non-null handle."""
+    sample_graphdef = GraphDefinition()
```
seberg, line 821: Admittedly, we could also do the fixture hack, but somehow this felt just as well for now...

cuda_core/tests/test_object_protocols.py
```diff
@@ -527,6 +527,26 @@ def sample_switch_node_alt(sample_graphdef):
     return sample_graphdef.switch(condition, 3)
 
 
+# Indirect-parametrize helpers: request.getfixturevalue() runs here, in the
+# fixture (main thread), so the resolved object is already available when the
+# test function runs in a worker thread.
```
seberg: Admittedly, the workers may not be valid on other threads, but we don't actually use CUDA API anywhere...

cuda_core/tests/helpers/latch.py
```diff
@@ -41,6 +48,7 @@ def __init__(self, device, timeout_sec=60):
 
                        // Check for timeout
                        if (clock64() - start >= timeout_cycles) {
+                           signal.store(-1, cuda::memory_order_relaxed);
```
seberg: I think I was just debugging/trying to understand where things are hanging.

Either way, the main point here is that we need to compile the kernel only once, because otherwise compiling the kernel itself will block other things and break other threads.

cuda_core/tests/graph/test_graph_definition_lifetime.py
```diff
@@ -1319,6 +1320,7 @@ def test_memcpy_buffer_survives_close(init_cuda):
     assert list(out) == [0xCD] * 4
 
 
+@pytest.mark.thread_unsafe(reason="deferred cleanup on main thread which would wait")
```
seberg, line 1416: There are three tests here that fail with `_wait_until` @Andy-Jost.  I just skipped them, the reason seems to be that cleanup can never happen as long as the main thread is waiting in `thread.join()`?

I suspect that is a potential but very minor issue (you would think _eventually_ this cleanup happens).  But wanted to make a comment.  I didn't try to understand what is going on here exactly!

cuda_core/tests/helpers/latch.py
seberg: Nvm, I pulled out the `latch.py` fixes... They are correct, but the fact that there are allocations, makes them a bit unreliable and the only way to fix it is to introduce that `barrier_wait` fixture (I had forgotten about it).

And I don't like that fixtures quite enough yet, maybe after `pytest-run-parallel` has better infrastructure for it.

cuda_core/tests/graph/test_graph_definition_lifetime.py
Andy-Jost: The issue is that deferred cleanup requires Python's main thread.

1. pytest main thread: start workers -> join workers (blocking)
2. pytest worker thread: del graph -> _wait_until
3. CUDA calls back to Python from a separate thread and cuda.core schedules the destruction with `Py_AddPendingCall`
4. Python's pending callbacks only run on the main thread.

`pytest-run-parallel` needs to give pending calls a chance to run. Instead of a blocking `worker.join`:

```
while any(worker.is_alive() for worker in workers):
    for worker in workers:
        worker.join(timeout=0.01)
```

The test also needs to be updated to use `threading.main_thread().ident` rather than `threading.get_ident()`

cuda_core/tests/test_object_protocols.py
Andy-Jost: This seems like a better approach. Suggested comment:

```
# Resolve fixture names during pytest setup, before pytest-run-parallel starts
# worker threads. The workers then share the resolved, read-only test object.
```

Andy-Jost (approved):
Getting the graph-cleanup tests running seems to require a change to `pytest-run-parallel`. Can we merge this and follow up with a PR there?

Issue #2311, opened 2026-07-07
[BUG]: Device methods operate on the current context, not the device they're called on


Andy-Jost:
## Type of Bug

Silent Failure / Wrong Result

## Component

cuda.core

## Describe the bug

Several `Device` methods read as if they create a resource *on that `Device`*, but they actually create it on whatever CUDA context is **current on the calling thread**. The device's own context (`self._context`) is passed through only as metadata (for lifetime tracking and `device_id` labeling); it is never made current before the driver call, and the methods only guard with `_check_context_initialized()`, which checks "has `set_current()` *ever* been called on this object" — not "is this device current right now."

The result: `device1.create_stream()` can silently create a stream on device 0, and label it `device_id == 1`. The returned object's metadata disagrees with where the resource physically lives.

### Affected methods

- `Device.create_stream` (`cuda_core/cuda/core/_device.pyx`) → `cuStreamCreateWithPriority` / `cuGreenCtxStreamCreate` (no context arg; current context). Also `cuCtxGetStreamPriorityRange` in `_stream.pyx` reads the current context.
- `Device.create_event` (`cuda_core/cuda/core/_device.pyx`) → `cuEventCreate` (no context arg; current context).
- The four texture allocation methods currently in flight in #2307 — `create_opaque_array`, `create_mipmapped_array`, `create_texture_object`, `create_surface_object` — which delegate to helpers that use `_get_current_device_id()` and current-context driver calls. Their docstrings already document the footgun ("created in the current CUDA context, so make this device current with `set_current` before calling").

## How to Reproduce

```python
from cuda.core import Device

dev0 = Device(0); dev0.set_current()
dev1 = Device(1); dev1.set_current()   # dev1 marked initialized
dev0.set_current()                     # device 0 is now current

s = dev1.create_stream()               # passes the init check (dev1 was inited once)
                                       # but the stream is created on device 0!
# s._device_id reports 1, contradicting where the stream actually lives
```

## Expected behavior

`deviceN.create_*(...)` should create the resource on device N's context regardless of which context is currently active on the thread, and the returned object's `device_id` should be accurate. At minimum, if we intend to keep the "current context" semantics, the methods should assert that this device is current and raise rather than silently doing the wrong thing.

## Proposed fix

Once #1586 lands and `Device` is a context manager, each method can wrap its body in a `with self:` block, which activates this device's context and restores the previous context on exit (exception-safe). That neatly fixes all six methods with a one-line change each, e.g.:

```python
[...]

PR #2455, opened 2026-07-29
cuda.core: make buffer cleanup context-safe

Andy-Jost:
## Summary

Make device-pointer cleanup retain and activate the CUDA context required by the allocation or its deallocation stream. This prevents deferred graph attachment cleanup and ordinary `Buffer` destruction from leaking memory when no context is current on the cleanup thread.

## Changes

- Resolve the release context in Cython (`Stream_resolve_context`) and pass an explicit stream/context pair into device-pointer resource handles; C++ stores that pair and makes the context current during destruction.
- Retain allocation and deallocation-stream context handles, including owning primary-context retains when the driver permits them.
- Temporarily activate the retained context for native frees, graphics unmap, and `MemoryResource.deallocate()` callbacks, then restore the caller's prior context.
- Stop treating `CUDA_ERROR_INVALID_CONTEXT` as successful memory-pool deallocation, and report cleanup failures on `stderr` instead of discarding them.
- Add an optional keyword-only `stream=` argument to `Buffer.from_handle` and `ManagedBuffer.from_handle`, defaulting to the calling thread's default stream when a memory resource owns the pointer.
- For IPC imports that arrive with a default-stream token and no current context, bind the pool device's primary context so release still has a context to run under.

## Review guide

1. Review `_stream.pyx` / `_buffer.pyx` / `_ipc.pyx` for how Cython chooses the release context.
2. Review `resource_handles.cpp/.hpp` for storing that context as a handle dependency and restoring it during teardown.
3. Review the memory and graph lifetime tests for cleanup with no current context.

## Related Work

- Follow-up to #2395.
- Related to #2001 and #1989.


kkraus14:
@Andy-Jost is there any performance concerns here? I vaguely remember some microbenchmarking you did but I don't recall if it was related to this PR.

Andy-Jost:
> @Andy-Jost is there any performance concerns here? I vaguely remember some microbenchmarking you did but I don't recall if it was related to this PR.

@kkraus14 I think you're referring to the discussion in #2311, which showed ~9ns for a current-context check and ~37ns for a check-and-switch. That came up when discussing `Device` objects as Python context managers. I don't anticipate any performance problems from this PR, though I can check if you're concerned about it.

kkraus14:
> > @Andy-Jost is there any performance concerns here? I vaguely remember some microbenchmarking you did but I don't recall if it was related to this PR.
> 
> @kkraus14 I think you're referring to the discussion in #2311, which showed ~9ns for a current-context check and ~37ns for a check-and-switch. That came up when discussing `Device` objects as Python context managers. I don't anticipate any performance problems from this PR, though I can check if you're concerned about it.

The approach here is a similar pattern, yes? If so I think we're good. If not I think it would be good to do a minimal test just to ensure we aren't adding 100s of ns of overhead.

PR #2456, opened 2026-07-30
cuda.core: add LaunchConfig.programmatic_stream_serialization for programmatic dependent launch

### LINKED ISSUE #1334: Support programatic dependent launch (PDL)
- [ ] Users can set `cudaLaunchAttributeProgrammaticStreamSerialization` to do PDL.
- [ ] PDL launches are graph-compatible and this use case should be tested and showcased

lijinf2:
## Description

Part of #1334

Expose `LaunchConfig.programmatic_stream_serialization` to set
`CU_LAUNCH_ATTRIBUTE_PROGRAMMATIC_STREAM_SERIALIZATION`, and add mapping plus
Hopper+ primary/secondary overlap tests.

Reference: https://docs.nvidia.com/cuda/cuda-programming-guide/04-special-topics/programmatic-dependent-launch.html

Release notes / Graph showcase will be in next PR. 

## Checklist


cuda_core/cuda/core/_launch_config.pyi
```diff
@@ -35,9 +35,13 @@ class LaunchConfig:
         (Default to size 0)
     is_cooperative : bool, optional
         Whether this config can be used to launch a cooperative kernel.
+    programmatic_stream_serialization : bool, optional
+        Whether to allow programmatic stream serialization (PDL). When True,
```
isVoid, line 39: PDL is `programmatic dependent launch`, "programmatic stream serialization" is a sub-concept inside PDL?

cuda_core/cuda/core/_launch_config.pyx
```diff
@@ -204,6 +224,12 @@ cpdef object _to_native_launch_config(LaunchConfig config):
         attr.value.cooperative = 1
         attrs.append(attr)
 
+    if config.programmatic_stream_serialization:
```
isVoid, line 227: Non blocking Q: I see from the comment to this function "once all modules are cythonized, this function can be dropped in favor of the cdef method above". Are all modules cythonized?

cuda_core/tests/test_launcher.py
```diff
@@ -183,6 +183,109 @@ class _FakeDev:
...
+
+    # Without the PDL attribute, same-stream kernels stay serialized.
+    assert _run(secondary_serial_cfg) == 0, "Expected no overlap when programmatic_stream_serialization is False"
+
+    # PDL overlap is opportunistic; retry a few times on a quiet GPU.
+    saw_overlap = False
+    for _ in range(5):
+        if _run(secondary_cfg) == 1:
+            saw_overlap = True
+            break
```
isVoid, line 281: Frankly, I have huge doubt over testing a non-deterministic behavior and asserting it in the test. We shouldn't be introducing flakiness in the test ourselves.

However kernel overlapping is opportunistic in nature. I vibe coded with my agent for a few rounds and I couldn't find a way to write a kernel and turn this into a deterministic test. My agent suggested that we should actually inspect the edge data from cuda graph to validate its dependency. But we don't have PDL in graph for now. (right?)

IMO, it's probably also acceptable to not expect `saw_overlap` here to always pass. Instead, we can use `pytest.xfail` when we didn't observe an overlap run. In that case, we validate the API via inspecting over a period of CI runs, not just through a single flaky failure.

isVoid (changes requested):
I have one question regarding the test, another small comment for the documentation.

cuda_core/cuda/core/_launch_config.pyi
lijinf2: I think so. PDL looks to include both `programmatic stream serialization` and `programmatic event.`

[CUDA programming guide PDL webpage ](https://docs.nvidia.com/cuda/cuda-programming-guide/04-special-topics/programmatic-dependent-launch.html)shows an example of cudaLaunchAttributeProgrammaticStreamSerialization. 

The page also describes cudaLaunchAttributeProgrammaticEvent but no code example.

cuda_core/cuda/core/_launch_config.pyx
lijinf2: I have no answer. this cpdef function might still be needed for tests / non-Cython callers.

lijinf2: Just checked codebase. This function is used in `tests/test_launcher.py::test_to_native_launch_config_no_cluster`, where there is a python caller "from cuda.core._launch_config import _to_native_launch_config"

cuda_core/tests/test_launcher.py
lijinf2: Yes, graph support is planed in next PR.

Done. Revised to xfail.

isVoid (approved):
lgtm!

Issue #2381, opened 2026-07-17
[BUG]:  Intermittent CUDA_ERROR_OUT_OF_MEMORY cascade in cuda_core test suite


rluo8:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Runtime Error

### Component

cuda.core

### Describe the bug

When running the full `cuda_core` v1.1.0 test suite (`python -m pytest .`) repeatedly across several Python environments on a single GPU, some of the runs intermittently fails with a large, identical cascade of failures.
Most of the failures are about CUDA_ERROR_OUT_OF_MEMORY.

This is similar to internal bug 5815123.

Observations:
The failures are random when tests with Python 3.10/3.11/3.12/3.13/3.14/3.14t. 
When testing a single failed case, it could pass. If I added a sleep 2 or 3 seconds hook between the two test cases, mostly it could pass.  

Attached the logs for several runs without sleeping.

[running_tests_logs.zip](https://github.com/user-attachments/files/30119247/running_tests_logs.zip)

### How to Reproduce

pip install cuda-core[cu13]
pip install cuda-bindings==13.3.1

Checkout the source codes for tag cuda-core-v1.1.0
cd cuda_core
pytest -v tests   (Run for every python version from 3.10 to 3.14t)

### Expected behavior

The tests could all pass.

### Operating System

Windows 2025 server

Andy-Jost:
## Root cause

Each CUDA memory pool reserves virtual address space at creation. Reservations are address space only — no physical memory is committed — but the per-process VA budget is bounded, and on Windows (notably MCDM) that bound is roughly 1 TB rather than the 47+ bits available on Linux.

Three factors combine:

**1. Pools created without `max_size` get a system-dependent reservation.** That default scales with device memory and lands in the hundreds of gigabytes on large-memory GPUs. Most `cuda.core` tests pass `max_size=2097152` and are harmless; the ones that pass options without setting it (the IPC tests, primarily) reserve a full-size window for a pool that holds kilobytes. The `cuda.bindings` mempool tests construct pool properties by hand and leave `maxSize` at zero, with the same effect.

[...]

PR #2458, opened 2026-07-30
test(cuda_core): capture machine state on the first CUDA OOM

juenglin:
Issue #2381 reported intermittent runs where ~190 tests failed with
`CUDA_ERROR_OUT_OF_MEMORY`. That issue is now **closed**: the root cause was
two driver-managed memory pools (the default device pool and the CUDA graph
pool) that each permanently reserve host virtual-address space sized at
roughly 2x installed device memory, combined with Windows ASLR randomizing
whether two such reservations can find contiguous room. Turning ASLR off on
the affected machine made the suite pass deterministically; see the issue
for the full investigation.

That investigation surfaced a lesson this PR is now built around:
**`CUDA_ERROR_OUT_OF_MEMORY` does not mean the device is out of memory.** The
driver returns it whenever it cannot obtain some resource, and creating --
or even just looking up -- a memory pool reserves host virtual address space
before touching any device memory. A pytest log alone cannot distinguish
"host VA exhausted" from "device physical memory exhausted" from "another
process is holding the GPU," and reproducing the original failure required a
specific driver model, so any evidence has to come from whoever hits it next.

This PR keeps that as its purpose: a **failure-triggered OOM reason checker**
for the *next* time this error shows up, anywhere, for any reason -- not a
re-diagnosis of #2381, which is already resolved.

## What changed since the original version of this PR

- **Dropped `nvidia-smi`.** The original version shelled out to
  `nvidia-smi -q` and `--query-compute-apps`. Per review, this now uses only
  `cuda.bindings.driver` APIs -- no subprocess, no NVML.
- **Replaced the raw dump with a classifier.** Instead of printing driver
  call results and leaving the reader to interpret them, `helpers/oom_diagnostics.py`
  now runs an ordered sequence of probes into a `ProbeSnapshot` and turns it
  into a one-line verdict via a pure `classify()` function:
  physical device memory exhaustion, host VA exhaustion (at a small or a
  pool-sized granularity), default-mempool materialization failure, or
  inconclusive.
- **Portable by construction.** Every probe is a documented driver call
  against the current context/device; there are no OS-specific branches.
  Devices without mempools or VMM skip those steps rather than failing.
  Alignment comes from `cuMemGetAllocationGranularity` (falling back to 2 MiB)
  rather than a hardcoded page size.
- **Kept the pytest harness** (latch to the first OOM, artifact file via
  `terminalreporter`, `pytest_terminal_summary` pointer) essentially
  unchanged; that part of the design was already reviewed favorably.
- **Kept tests**
  They cover the harness and the `classify()` decision table with injected
  snapshots (no GPU needed) plus one live smoke test that only exercises the
  side-effect-free prefix (context, `cuMemGetInfo`, attributes) -- it never
  calls `cuDeviceGetMemPool` or creates a pool, so running the test suite
  doesn't itself materialize the ~2x-device-memory reservation this checker
  is trying to diagnose.

## Checklist

### DIFF
--- a/cuda_core/tests/helpers/oom_diagnostics.py
+++ b/cuda_core/tests/helpers/oom_diagnostics.py
@@ -1,87 +1,428 @@
 # SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-"""Machine-state capture for the first CUDA OOM of a test session (issue #2381).
+"""OOM reason checker for the cuda_core test suite (issue #2381).
 
-A failing ``cuda_core`` run reports ~190 ``CUDA_ERROR_OUT_OF_MEMORY`` failures
-that all descend from a single earlier event, so only the first is worth
-capturing; running ``nvidia-smi`` on every one would add minutes and bury the
-log. Hence the latch in :class:`OomDiagnosticsRecorder`.
+``CUDA_ERROR_OUT_OF_MEMORY`` does not mean "the device ran out of memory": the
+driver returns it whenever it could not obtain some resource, and one of
+those resources is a purely host-side virtual-address (VA) reservation.
+Creating -- or even just looking up -- a memory pool reserves a VA window
+before any device memory is touched (observed default: about 2x installed
+device memory). If that host reservation fails, the error looks identical to
+genuine physical exhaustion even though ``cuMemGetInfo`` may report the
+device is almost entirely free. This module runs a small, ordered sequence of
+``cuda.bindings.driver`` calls to tell the two apart, and turns the raw
+results into a verdict a reader does not have to reconstruct by hand.
+
+Every probe uses only documented driver APIs against the current context and
+device 0, so the same sequence runs unmodified on Linux, Windows (WDDM / TCC /
+MCDM), and WSL. There is no OS-specific branch and no external process
+(``nvidia-smi``, NVML): see the PR #2458 review for why the previous version's
+``nvidia-smi -q`` dump was dropped.
+
+Capture is latched once per session, on the first failing OOM. A failing run
+can report ~190 of these, all descending from one earlier event, and
+re-running the probes on each would add noticeable time and bury the log. The
+report is written via ``terminalreporter`` rather than ``print()`` so it
+survives the stdout redirection these runs use, and ``pytest_terminal_summary``
+points at the artifact from the end of the run, where it is actually noticed.
 """
 
 import os
 import pathlib
-import subprocess
 import sys
 import threading
+from dataclasses import dataclass
 
 from cuda.bindings import driver
+from helpers.constants import POOL_SIZE
 
 OOM_MARKER = "CUDA_ERROR_OUT_OF_MEMORY"
 DEFAULT_FILENAME = "cuda_core_oom_diagnostics.txt"
 
+GIB = 1 << 30
+# Used only if cuMemGetAllocationGranularity is unavailable; see _allocation_granularity.
+FALLBACK_ALIGNMENT = 2 * 1024 * 1024
+
 _BANNER = "=" * 78
 
+_LESSON = (
+    "CUDA_ERROR_OUT_OF_MEMORY means the driver could not obtain some resource;\n"
+    "it is not proof that device memory is exhausted. Creating -- or even just\n"
+    "looking up -- a memory pool first reserves a host virtual-address window\n"
+    "(observed default: about 2x installed device memory) before any device\n"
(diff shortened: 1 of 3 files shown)


cuda_core/tests/test_helpers.py
```diff
@@ -3,11 +3,21 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import time
+import types
 
 import pytest
 from helpers.buffers import PatternGen, compare_equal_buffers, make_scratch_buffer
 from helpers.latch import LatchKernel
 from helpers.logging import TimestampedLogger
+from helpers.oom_diagnostics import (
```
lijinf2, line 12: We could remove test functions for test helpers (actually, I am thinking whether it is possible to remove whole test_helpers.py file). There was a discussion that decided not to introduce tests for test helpers. If test helpers do not work properly, let them throw errors in the CI. There was an issue #2174 tracking the discussion.

cuda_core/tests/helpers/oom_diagnostics.py
```diff
@@ -0,0 +1,208 @@
...
+
+def run_nvidia_smi(*args):
+    cmd = ["nvidia-smi", *args]
+    printable = " ".join(cmd)
+    try:
+        proc = subprocess.run(cmd, capture_output=True, text=True, timeout=60, check=False)  # noqa: S603
+    except (OSError, subprocess.SubprocessError) as exc:
+        return f"$ {printable}\n<could not run: {exc!r}>"
+    output = proc.stdout.strip() or proc.stderr.strip() or "<no output>"
+    return f"$ {printable}\n(exit {proc.returncode})\n{output}"
```
kkraus14: Instead of externally invoking `nvidia-smi`, all of the info (and a whole lot more) that it captures is available programmatically via the NVML API which we should ideally use instead

kkraus14 (approved):

cuda_core/tests/test_helpers.py
lijinf2: Maybe we can create an issue to track if deleting test_helpers.py is possible? Or revisit the decision about not introducing tests for test helpers.

lijinf2 (approved):
This PR leverages pytest_runtest_makereport and pytest_terminal_summary to collect runtime specs of the first error in a sequence of errors of the same type. Good idea!

cuda_core/tests/test_helpers.py
lijinf2: Done #2471

cuda_core/tests/helpers/oom_diagnostics.py
juenglin, line 1: Example terminal output:
```
==============================================================================
cuda_core OOM reason checker: first CUDA_ERROR_OUT_OF_MEMORY of this session
==============================================================================
test:      tests/test_memory.py::test_device_memory_resource_default_pool
phase:     call
pid:       17012
platform:  win32
exception: CUDA_ERROR_OUT_OF_MEMORY: The API call failed because it was unable to allocate enough memory or other resources to perform the requested operation.

CUDA_ERROR_OUT_OF_MEMORY means the driver could not obtain some resource;
it is not proof that device memory is exhausted. Creating -- or even just
looking up -- a memory pool first reserves a host virtual-address window
(observed default: about 2x installed device memory) before any device
memory is touched. That reservation can fail while cuMemGetInfo still
reports most of the device free. The probes below check host VA and
physical device memory separately so the two are not confused. Note that
the '2x device memory' figure is an observation from measurement and a
driver source comment, not a documented guarantee -- it can differ across
driver versions and platforms.

--- direct driver probe (bypasses cuda.core's error reporting) ---
cuCtxGetCurrent() -> ok
cuMemGetInfo() -> free=143.50 GiB, total=178.81 GiB (80.3% free)
mempools supported: True
VMM (cuMemAddressReserve) supported: True
small cuMemAlloc(4 KiB) -> ok, freed
allocation granularity: 2097152 bytes
cuMemAddressReserve(2097152 bytes) -> ok, freed
cuMemAddressReserve(384309411840 bytes, observed default pool window, not a documented guarantee) -> <failed: CUresult.CUDA_ERROR_OUT_OF_MEMORY>
cuDeviceGetMemPool(dev 0) -> <failed: CUresult.CUDA_ERROR_OUT_OF_MEMORY>
cuDeviceGetDefaultMemPool(dev 0) -> <failed: CUresult.CUDA_ERROR_OUT_OF_MEMORY>
cuMemPoolCreate(maxSize=2097152) -> ok, destroyed

verdict: likely host VA exhaustion for the pool-sized window only: a capped memory pool (helpers.constants.POOL_SIZE) still creates fine, but a reservation the size of the observed default pool window does not
==============================================================================
(diagnostics also written to /path/to/cuda_core/cuda_core_oom_diagnostics.txt)
```

PR #2461, opened 2026-07-30
fix(cuda.core): surface real CUDA error in Device.set_current()

### LINKED ISSUE #2460: [BUG]:  `Device.set_current()` discards real CUDA error and raises misleading `ValueError`
### Is this a duplicate?

### Type of Bug

Runtime Error

### Component

cuda.core

## Summary

When `cuDevicePrimaryCtxRetain` fails inside `Device.set_current()`, the actual CUDA driver error is silently discarded and the user gets:

```
ValueError: Cannot set NULL context as current
```

The real cause (e.g. `CUDA_ERROR_DEVICE_UNAVAILABLE`, `CUDA_ERROR_OUT_OF_MEMORY`, exclusive compute mode conflict) is lost.

## Root cause

In `cuda_core/cuda/core/_device.pyx` lines 1289–1292, `get_primary_context()` returns an empty handle when the driver call fails and records the error in the thread-local slot. The caller never reads it:

```cython
h_context = get_primary_context(self._device_id)
if h_context.get() == NULL:
    raise ValueError("Cannot set NULL context as current")
```

This raises the wrong exception type (every other driver failure in `cuda.core` raises `CUDAError`) and leaves the thread-local error state polluted, potentially causing a stale error to be misreported by a later call.

### How to Reproduce

This is hit whenever cuDevicePrimaryCtxRetain fails for a valid device ordinal. One way to trigger it:

1. Set the GPU to exclusive compute mode: nvidia-smi -c EXCLUSIVE_PROCESS
2. Hold a context from another process
3. Call Device(0).set_current() from a second process

### Expected behavior

`Device.set_current()` should surface the underlying driver error as a `CUDAError`, matching the pattern used elsewhere (e.g. `_context.pyx:59–61`).

### Operating System

any

### nvidia-smi output

_No response_

juenglin:
closes #2460

Device.set_current() raised a misleading ValueError ("Cannot set NULL
context as current") when cuDevicePrimaryCtxRetain failed, discarding
the actual driver error code and leaving the thread-local error state
polluted.

Replace with HANDLE_RETURN(get_last_error()) to raise CUDAError with
the real driver status, matching the established pattern used elsewhere.

Manually verified by patching get_primary_context to return a failure;
a regression test is not practical because the failure requires
exclusive compute mode (unavailable on WDDM) or hardware error states
that cannot be simulated from Python.

### DIFF
--- a/cuda_core/cuda/core/_device.pyx
+++ b/cuda_core/cuda/core/_device.pyx
@@ -1289,7 +1289,10 @@ class Device:
             # use primary ctx
             h_context = get_primary_context(self._device_id)
             if h_context.get() == NULL:
-                raise ValueError("Cannot set NULL context as current")
+                HANDLE_RETURN(get_last_error())
+                raise RuntimeError(
+                    f"Failed to retain the primary context for device {self._device_id}"
+                )
             with nogil:
                 HANDLE_RETURN(cydriver.cuCtxSetCurrent(as_cu(h_context)))
             self._has_inited = True


kkraus14 (approved):

kkraus14:
@juenglin fix looks good, but is there a reasonable way to test this?

juenglin:
> @juenglin fix looks good, but is there a reasonable way to test this?

I couldn't find a way to test it with an automated test.

To test it I reproduced the bug with a temporary code change by which [get_primary_context](https://github.com/NVIDIA/cuda-python/blob/main/cuda_core/cuda/core/_cpp/resource_handles.cpp#L550) would always fail (setting a CUDA error). Then I confirmed that this change fixes it.

PR #2465, opened 2026-07-31
ci: add selective sdist build plumbing

kkraus14:
> [!IMPORTANT]
> **Stack dependency: PR 2 of 4.** Builds on #2464, which has merged. #2466 and #2467 depend on this PR.
>
> Merge order: **#2464 (merged) -> #2465 (this PR) -> #2466 -> #2467**. All branches live in the `kkraus14` fork and target upstream `main`.

## What

- Add default-on package build inputs to both sdist reusable workflows.
- Skip package-specific setup and sdist builds when that package is not selected.
- Hydrate unchanged cuda-pathfinder and cuda-bindings wheels from the current CI run when a selected downstream sdist needs them.
- Integrate selective downloads with #2510's exact local-wheel constraints, without `PIP_PRE` or `PIP_FIND_LINKS`.
- Preserve the current Linux and Windows compiler, cache, CTK, timeout, and path behavior.

## Why

Selective wheel builds are not enough on their own: the sdist smoke jobs also rebuild every package today. This plumbing lets the final orchestration PR apply the same dependency closure to sdist coverage.

## Behavior before activation

Every new input defaults to `true`, and this PR does not change any caller. Sdist coverage remains unchanged until the orchestration PR supplies selective values.

## Validation

- Full pre-commit hooks for both workflows
- YAML and GitHub Actions lint
- Review of all build-flag combinations and Linux/Windows artifact dependencies
- Verified the final diff contains only the two sdist workflows on top of merged #2464/current `main`

Related to #299.


mdboom (approved):
LGTM

PR #2466, opened 2026-07-31
ci: add selective wheel test plumbing

kkraus14:
> [!IMPORTANT]
> **Stack dependency: PR 3 of 4.** Builds on #2464 and #2465, which have merged. #2467 depends on this PR.
>
> Merge order: **#2464 (merged) -> #2465 (merged) -> #2466 (this PR) -> #2467**. All branches live in the `kkraus14` fork and target upstream `main`.

## What

- Replace the negative bindings-only skip switch with default-on test inputs for every package.
- Gate package wheel downloads, compiled test fixtures, sanitizer setup, and individual suites on those positive selections.
- Keep nightly callers unchanged through default-on inputs.
- Make metapackage-only smoke tests install the exact local cuda-pathfinder and cuda-bindings wheels in one resolver transaction while released dependencies continue to come from configured indexes.
- Keep Python 3.15 NumPy index setup aligned with every selected suite that can resolve NumPy, including metapackage-only tests.

## Why

The final planner needs to select each package test suite independently without losing the artifacts and setup needed by downstream tests.

## Behavior before activation

Every new test input defaults to `true`. The top-level CI caller translates its existing negative bindings selection to the equivalent positive `test-bindings` value, while nightly callers continue to omit the new inputs. Test coverage remains unchanged until the orchestration PR supplies selections for every package.

## Validation

- Rebased onto current `main` after #2465 merged; the final diff contains only the two reusable wheel-test workflows and their `ci.yml` caller update
- Full targeted pre-commit hooks for all changed workflows
- YAML and GitHub Actions lint
- Linux/Windows review of artifact, Python 3.15, backport, nightly, and metapackage-only paths

Related to #299.


mdboom (approved):
Approving.  With one naming nit that you can take or leave.

kkraus14:
> Approving. With one naming nit that you can take or leave.

I'm not seeing a review comment with a naming nit. Did it get lost somehow? Regardless, will merge this now and then address the naming nit in #2467

PR #2467, opened 2026-07-31
ci: activate dependency-aware package builds and tests

kkraus14:
> [!IMPORTANT]
> **Stack completion: PR 4 of 4.** #2464, #2465, and #2466 have merged. This is the final PR in the stack.
>
> Merge order: **#2464 (merged) -> #2465 (merged) -> #2466 (merged) -> #2467 (this PR)**. The branch lives in the `kkraus14` fork and targets upstream `main`.

## What

- Compute one structured CI workplan with a small standard-library Python planner instead of maintaining duplicated positive/negative path filters and dependency expressions in workflow YAML.
- Keep the planner focused on CI execution: it classifies `merge-base...HEAD`, applies the four-package build/test impact policy, and emits final module and job decisions as JSON.
- Pass that single workplan through the wheel-build, sdist, and wheel-test reusable workflows. An omitted workplan preserves their existing full-build/full-test behavior for nightly callers.
- Resolve a successful CI run for the PR's exact merge-base commit and reuse its complete wheel set. Missing, expired, or incomplete baseline artifacts conservatively force a full build and test.
- Use broad, maintainable path conventions: unknown paths and shared CI/workflow/action infrastructure force the full pipeline; ancillary GitHub metadata is ignored.
- Leave `pixi.toml` and `pixi.lock` changes to the dedicated source-build workflow, and ignore ordinary Markdown/SVG changes unless they are test inputs or tracked symlink consumers.
- Track symlink consumers generically across both the merge-base and head trees, including additions, deletions, replacements, and target changes.
- Preserve deletion and cross-package rename handling with NUL-delimited `git diff --no-renames`.
- Keep the universal Linux artifact path for docs and downstream jobs, while gating unnecessary platform, sdist, wheel-test, and cuda-core API jobs.
- Keep the final status gate strict: intentional selective skips pass, while skips caused by upstream failures fail CI.

## Expected package behavior

- `cuda_pathfinder` source: build and test all four packages.
- `cuda_bindings` source: build and test bindings, core, and cuda-python.
- `cuda_core` source: build core; test core and cuda-python; run cuda-core API checks.
- `cuda_python` source: build bindings and cuda-python because development cuda-python wheels exactly pin bindings; test cuda-python.
- Package tests/examples: run that package's tests without rebuilding wheels.
- Test-helper and benchmark implementation: run all four package test suites.
- Linux-only or Windows-only wheel-test infrastructure: run all module tests on the affected platform.
- Package docs, ordinary Markdown/SVG, Pixi manifests, and ancillary GitHub metadata: no conditioned package work.
- Other `ci/**`, `.github/workflows/**`, or `.github/actions/**` changes: run the full pipeline.
- Unknown paths, a non-PR event, or an unavailable baseline: run the full pipeline.

## Validation

- Added table-driven planner coverage for package impact, ignored paths, platform-specific infrastructure, mixed changes, symlink consumers, unknown-path fail-open behavior, and unavailable/incomplete baselines.
- Planner unit tests pass under the repository Pixi environment.
- Ruff check and format pass.
- Full targeted pre-commit hooks pass, including SPDX and static checks.
- Independent planner, workflow, regression-scope, symlink, and maintainability reviews found no blocking issues.
- Exact head `fba3b4dbb5` completed all 102 main CI jobs successfully; all auxiliary checks also passed.

Implements the remaining dependency-aware build/test selection requested in #299.


.github/workflows/ci.yml
```diff
@@ -136,99 +207,267 @@ jobs:
...
             exit 1
           fi
+
           base=$(git merge-base HEAD "origin/${BASE_REF}")
-          changed=$(git diff --name-only "$base"...HEAD)
+          echo "sha=${base}" >> "$GITHUB_OUTPUT"
 
-          has_match() {
-            grep -qE "$1" <<< "$changed" && echo true || echo false
+      - name: Classify changed paths
```
mdboom: The hardcoded path lists here is going to be a maintenance nightmare.

Each subsection follows a pretty similar pattern, though.  If there were a better way of specifying things to reduce duplication, I think that would be fine.  (For example, if we could globally exclude all `AGENTS.md`).

As it stands, if this is the best we can do with `dorny/paths-filter`, I think we should reach for a different tool or write our own in Python.  Upside of that, too, is we could build a "test everything we need to" local script on top of it, whereas this is stuck in the GHA silo.

.github/workflows/ci.yml
```diff
@@ -136,99 +207,267 @@ jobs:
...
+              - '!cuda_core/CLAUDE.md'
+            python_source:
+              - 'cuda_python/**'
+              - '!cuda_python/docs/**'
+              - '!cuda_python/tests/**'
+              - '!cuda_python/examples/**'
+              - '!cuda_python/AGENTS.md'
+              - '!cuda_python/CLAUDE.md'
+              # cuda_python/README.md is a symlink to this packaging input.
+              - 'README.md'
```
mdboom: Not sure I understand the comment.

mdboom (changes requested):
Sorry this review will be long, and will negate some of the earlier decisions in the stack.  I should have evaluated the whole stack at once to see where this is going.

I think this is fundamentally a maintenance nightmare and way more verbose than it needs to be.  Rather than passing 9 new parameters between jobs, there should be a single, structured JSON object that is passed around.  I am thinking something like (using YAML because it's easier to type, but it would be JSON).

```
modules:
  test_helpers:
    needs_build: false
    needs_test: false
  pathfinder:
    needs_build: false
    needs_test: false
    build_depends: []
    test_depends: ["test_helpers"]
  bindings:
    needs_build: false
    needs_test: false
    build_depends: ["pathfinder"]
    test_depends: ["test_helpers"]
  core:
    needs_build: false
    needs_test: false
    build_depends: ["bindings"]
    test_depends: ["test_helpers"]
baseline_run_id: 123456789
baseline_sha: c0ffee
```

This makes verbose things that are sensitive to modules coming and going like:

```
${{ needs.detect-changes.outputs.build_pathfinder == 'true' ||
                               needs.detect-changes.outputs.build_bindings == 'true' ||
                               needs.detect-changes.outputs.build_core == 'true' ||
                               needs.detect-changes.outputs.build_python == 'true' }}
```

into:

```
Object.values(data.modules).some(m => m.needs_test)
```

Rather than this:

```
      test_bindings: >-
        ${{ steps.baseline.outputs.available != 'true' ||
            steps.filter.outputs.changes == '' ||
            steps.filter.outputs.force_all == 'true' ||
            steps.filter.outputs.all_tests == 'true' ||
            steps.filter.outputs.pathfinder_source == 'true' ||
            steps.filter.outputs.bindings_source == 'true' ||
            steps.filter.outputs.bindings_tests == 'true' ||
            steps.filter.outputs.test_helpers == 'true' }}
```

You can do:

```
(name, modules, seen = new Set()) => {
  if (seen.has(name)) return false;
  seen.add(name);
  const m = modules[name];
  return m.needs_test || (m.depends || []).some(dep => moduleNeedsTest(dep, modules, seen));
};
```

rather than hardcoding the dependencies into a bunch of large boolean expressions.

We can pre-compute values for downstream consumers so this kind of logic is only written in one place.

```
data["needs_build"] = Object.values(data.modules).some(m => m.needs_build)

# or Python
data["needs_build"] = any(mod["needs_build"] for mod in data["modules"])
```

This object would get passed between jobs as "workplan" (or some better name).

Given my other comment that `dorny/paths-filter` seems like the wrong tool for the job, I would propose:

A standalone Python script that does the analysis of changes and builds this workplan object, including as many pre-computed dependendent information as we can to make downstream consumers of this information concise and prevent long complicated expressions being repeated everywhere.

This script has the side benefit of forming the basis of a future local helper to test only what is needed (which all of this GHA work doesn't move us toward).

(I confirmed that outputs can have a [size maximum of 1MB](https://docs.github.com/en/actions/reference/workflows-and-actions/metadata-syntax#outputs-for-composite-actions) which should be more than enough for this.)

.github/workflows/ci.yml
kkraus14: `cuda_python/README.md` symlinks the root `README.md` and it gets included in the Python package that we produce, so changes to the top level `README` need to trigger a rebuild of the metapackage.

.github/workflows/ci.yml
kkraus14: Originally my agent had implemented a Python script to encapsulate the logic of what paths / files impact which builds / tests and I found that script even more difficult to understand and maintain than maintaining a list of paths like this that an agent could easily update (and we should include in the `AGENTS.md` in hindsight). Let me see what I can do to simplify things here, because I agree this is going to be an ongoing annoyance.

For what it's worth, I think this overall work is going to be a stopgap as opposed to the end solution, where moving to something like moon (https://moonrepo.dev/moon) or bazel is probably a better longer term solution that makes things more "target" oriented as opposed to path / file oriented.

ci/tools/compute_ci_plan.py
```diff
@@ -0,0 +1,241 @@
...
+    ".gitignore",
+    ".pre-commit-config.yaml",
+    ".spdx-ignore",
+    "CONTRIBUTING.md",
+    "LICENSE",
+    "SECURITY.md",
+    "context7.json",
+    "greptile.json",
+    "pixi.lock",
+    "pixi.toml",
```
mdboom: We can probably safely put these in `IGNORED_BASENAMES`

ci/tools/compute_ci_plan.py
```diff
@@ -0,0 +1,241 @@
...
+    "python": ({"bindings", "python"}, {"python"}),
+}
+
+IGNORED_BASENAMES = {"AGENTS.md", "CLAUDE.md"}
+IGNORED_PATHS = {
+    ".coveragerc",
+    ".gitignore",
+    ".pre-commit-config.yaml",
+    ".spdx-ignore",
+    "CONTRIBUTING.md",
```
mdboom: We can probably safely ignore all `.md` files, especially if we handle files downstream of a directory called `test` specially, which it looks like we already do below.

ci/tools/compute_ci_plan.py
```diff
@@ -0,0 +1,241 @@
...
+    "context7.json",
+    "greptile.json",
+    "pixi.lock",
+    "pixi.toml",
+    "pytest.ini",
+    "ruff.toml",
+    "benchmarks/cuda_bindings/pixi.lock",
+    "benchmarks/cuda_bindings/pixi.toml",
+    "ci/.ci-pipeline-regen.md",
+    "ci/ci-pipeline.svg",
```
mdboom: We can probably safely ignore all `.svg` files.

ci/tools/compute_ci_plan.py
```diff
@@ -0,0 +1,241 @@
...
+    "ci/cleanup-pr-previews",
+    "ci/tools/check_mempool_hygiene.py",
+    "ci/tools/check_pixi_cuda_version.py",
+    "ci/tools/check_release_notes.py",
+    "ci/tools/download-wheels",
+    "ci/tools/run_pytest_with_stack.py",
+    "ci/tools/validate-release-wheels",
```
mdboom: I think never ignoring anything in `ci` would be a safer (and simpler) choice.

ci/tools/compute_ci_plan.py
```diff
@@ -0,0 +1,241 @@
...
+    ".pre-commit-config.yaml",
+    ".spdx-ignore",
+    "CONTRIBUTING.md",
+    "LICENSE",
+    "SECURITY.md",
+    "context7.json",
+    "greptile.json",
+    "pixi.lock",
+    "pixi.toml",
+    "pytest.ini",
```
mdboom: This should not be on the ignore list -- this could have an appreciable impact on tests in the whole repo.

ci/tools/compute_ci_plan.py
```diff
@@ -0,0 +1,241 @@
...
+    all_tests = False
+    force_all = not merge_base or not baseline_run_id or not baseline_sha
+
+    if not force_all:
+        for path in paths:
+            path_parts = PurePosixPath(path).parts
+            if not path_parts or path in IGNORED_PATHS or path_parts[-1] in IGNORED_BASENAMES:
+                continue
+
+            if path == "README.md":
```
mdboom: Can we just track symlinks generically (using `Path.resolve` or similar)?

ci/tools/compute_ci_plan.py
```diff
@@ -0,0 +1,241 @@
...
+                if relative[0] == "docs":
+                    continue
+                if relative[0] in {"tests", "examples"} or (module == "core" and relative == ("pytest.ini",)):
+                    test_changes.add(module)
+                else:
+                    source_changes.add(module)
+                continue
+
+            if path.startswith("cuda_python_test_helpers/cuda_python_test_helpers/"):
+                test_changes.update(("bindings", "core"))
```
mdboom: It doesn't today, but it wouldn't surprise me if cuda_pathfinder depended on test helpers someday.  Probably safer to do this:

```suggestion
                test_changes.update(MODULES)
```

ci/tools/compute_ci_plan.py
```diff
@@ -0,0 +1,241 @@
...
+    ".github/workflows/test-wheel-linux.yml",
+    ".github/workflows/test-wheel-windows.yml",
+    "ci/test-matrix.yml",
+    "ci/tools/configure_driver_mode.ps1",
+    "ci/tools/guess_latest.sh",
+    "ci/tools/install_gpu_driver.ps1",
+    "ci/tools/install_gpu_driver.sh",
+    "ci/tools/run-tests",
+    "ci/tools/setup-sanitizer",
+}
```
mdboom: I think it's safer to trigger full tests if anything in `ci` or `.github/{workflow|actions}` changes.

ci/tools/compute_ci_plan.py
```diff
@@ -0,0 +1,241 @@
...
+
+INDEPENDENT_GITHUB_PATHS = {
```
mdboom: As I said elsewhere, I think we can invert the logic and consider anything in `workflows` or `actions` to trigger a full test and everything else ignored.

ci/tools/compute_ci_plan.py
```diff
@@ -0,0 +1,241 @@
...
+                if relative[0] in {"tests", "examples"} or (module == "core" and relative == ("pytest.ini",)):
+                    test_changes.add(module)
+                else:
+                    source_changes.add(module)
+                continue
+
+            if path.startswith("cuda_python_test_helpers/cuda_python_test_helpers/"):
+                test_changes.update(("bindings", "core"))
+            elif path.startswith("benchmarks/cuda_bindings/"):
+                test_changes.add("bindings")
```
mdboom: Ditto to above.

```suggestion
                test_changes.add(MODULES)
```

mdboom (changes requested):
Some suggestions to simplify the logic.

I think the agent here is trying to use a scalpel, and thereby creating real pitfalls if assumptions that hold true today don't hold true in the future.

I think instead, we take an approach to be as simple as possible, err on the side of doing too much work, rather than missing things that legitimately should be tested.

Maybe some description similar to the above would cause the agent to come up with something similar/safer?  If not, I think my concrete suggestions should also help.

.github/workflows/ci.yml
kkraus14: Addressed in the latest revision. The duplicated `dorny/paths-filter` block is gone; a small standard-library Python planner now applies broad conventions and emits one JSON workplan. Global basename/suffix rules and conservative unknown-path fallback keep the policy much smaller and safer. It is CI-focused for now, but the classification is no longer embedded in GHA.

.github/workflows/ci.yml
kkraus14: Follow-up: this is generic now rather than a `README.md` special case. The planner reads tracked symlinks from both the merge-base and head trees and uses `Path.resolve()` to propagate target changes to consumers, including link additions, deletions, and replacements.

ci/tools/compute_ci_plan.py
kkraus14: Done. `pixi.toml` and `pixi.lock` are now in `IGNORED_BASENAMES`; the dedicated Pixi source-build workflow remains responsible for those changes.

ci/tools/compute_ci_plan.py
kkraus14: Done. Ordinary `.md` files are ignored generically. Package test directories are classified first, so Markdown test inputs still select tests, and tracked symlink consumers intentionally bypass the suffix ignore.

ci/tools/compute_ci_plan.py
kkraus14: Done. Ordinary `.svg` files are ignored generically, while test and CI infrastructure rules take precedence.

ci/tools/compute_ci_plan.py
kkraus14: Implemented conservatively: `ci/**`, `.github/workflows/**`, and `.github/actions/**` force the full pipeline by default. The only exceptions are seven OS-exclusive wheel-test files/scripts, which run all module tests on their owning platform to preserve the platform split added after this review.

ci/tools/compute_ci_plan.py
kkraus14: Done. Changes under `cuda_python_test_helpers/**` now use `test_changes.update(MODULES)`.

ci/tools/compute_ci_plan.py
kkraus14: Done. The old `ci/**` ignore inventory is gone. Other than the seven explicitly OS-exclusive wheel-test paths retained for the later Linux/Windows split, any `ci/**` change now takes the conservative full-pipeline fallback.

ci/tools/compute_ci_plan.py
kkraus14: Done. Root `pytest.ini` is no longer ignored and therefore takes the conservative full-pipeline fallback.

ci/tools/compute_ci_plan.py
kkraus14: Done. Symlink handling is generic: the planner reads link entries from both endpoint trees, expands changed targets through `Path.resolve()`, and retains base-tree link paths so deleted or replaced consumers still select work.

ci/tools/compute_ci_plan.py
kkraus14: Done. The per-workflow/action allowlists were removed: workflows/actions default to the full pipeline, while other `.github/**` paths are ignored. The only narrow exceptions are the explicitly OS-exclusive wheel-test paths noted in the related thread.

ci/tools/compute_ci_plan.py
kkraus14: Done. Benchmark implementation changes now use `test_changes.update(MODULES)`; generic documentation, Pixi, and instruction-file ignores still apply.

.github/workflows/ci.yml
```diff
@@ -62,179 +66,186 @@ jobs:
...
+          cuda_version=$(yq '.cuda.build.version' ci/versions.yml)
+          if ! python_versions=$(yq -r '.jobs.build.strategy.matrix."python-version"[]' .github/workflows/build-wheel.yml); then
+            unavailable
+          fi
+          if [[ -z "${python_versions}" ]]; then
+            unavailable
+          fi
+          while IFS= read -r python_version; do
+            python=${python_version//./}
+            for platform in linux-64 linux-aarch64 win-64; do
```
mdboom: A minor point found by my agent: this hard-coded list of platforms might be easy to miss as a place that needs an update, thereby silently missing platforms added in the future, like WoA.  Could specify the supported platforms in a new key in test-matrix.yml, for example, and use `yq` to get them over here.

mdboom (approved):
LGTM

One minor nit about the hardcoded platform list.

kkraus14:
/ok to test https://github.com/NVIDIA/cuda-python/pull/2467/commits/727ef5994e41212e09a3adaff1db8fbdf126f7cc

kkraus14:
/ok to test https://github.com/NVIDIA/cuda-python/pull/2467/commits/ceaae51ecbe7c3d761c6b63611ec7bfcad0b15d1

rwgk:
/ok to test https://github.com/NVIDIA/cuda-python/pull/2467/commits/1516ca87491ed5409ad50862d815c9a96c6b4900

rwgk:
## End-to-end selective-CI validation: passed

I validated #2467 with temporary dependent draft PR [#2735](https://github.com/NVIDIA/cuda-python/pull/2735). The probe targets `pull-request/2467` at exact SHA 1516ca87491ed5409ad50862d815c9a96c6b4900 and changes only a module docstring in `cuda_core/tests/test_typing_imports.py`, so it is a genuine `cuda_core` test-only change with no product behavior change.

**Result:** the selective path worked end to end on the first attempt. No reruns or follow-up fixes were needed.

### Baseline and planner selection

- The full baseline [CI run 33340094633](https://github.com/NVIDIA/cuda-python/actions/runs/33340094633) completed successfully at the exact merge-base SHA: 102/102 jobs passed.
- The baseline contained all 50/50 unexpired artifacts required by the resolver: two universal wheels plus bindings/core wheels for eight Python variants on `linux-64`, `linux-aarch64`, and `win-64`.
- The probe's [`detect-changes` job](https://github.com/NVIDIA/cuda-python/actions/runs/33342679442/job/99341056966) selected baseline run `33340094633` and emitted the expected workplan:

```text
pathfinder: build=false, test=false
bindings:   build=false, test=false
core:       build=false, test=true
python:     build=false, test=false
platforms:  linux=true, windows=true
sdist tests=false
core API checks=false
baseline:   run 33340094633 at 1516ca87491ed5409ad50862d815c9a96c6b4900
```

### Artifact reuse and wheel-build selection

All 24 platform/Python wheel-wrapper jobs succeeded. Across those jobs:

- Every package-wheel build step was skipped: 24 pathfinder builds, 24 bindings builds, 48 current/prior core builds, and 24 cuda-python metapackage builds.
- Baseline artifact downloads succeeded in all expected wrappers: pathfinder 24/24, bindings 24/24, core 24/24, and the single universal cuda-python download 1/1.
- The wrappers republished a complete 50-wheel set under probe SHA 351129ea2f3c5aac092974c9750697c3c4f8f3a1: one pathfinder wheel, one cuda-python wheel, 24 bindings wheels, and 24 core wheels.
- The only newly generated build products were required by the selected core tests: 24 core Cython-test archives and 24 core test-binary archives.
- The probe run published 98 CI artifacts with no duplicates, expirations, or incorrect SHA-bearing names, plus the separate `github-pages` artifact.

A representative [Linux wheel-wrapper log](https://github.com/NVIDIA/cuda-python/actions/runs/33342679442/job/99341089953) records the full workplan, downloads from `run-id: 33340094633`, and successful re-uploads under the probe SHA.

### Test selection

All 65 selected GPU matrix jobs passed:

- 26 `linux-64` jobs
- 18 `linux-aarch64` jobs
- 21 `win-64` jobs

Every job ran `Run cuda.core tests` successfully. In all 65 jobs, pathfinder, bindings, cuda-python installability, pathfinder's second test phase, and the PyTorch/Numba/released-core interoperability phases were skipped by the workplan. Thus the full platform coverage remained intact while unrelated module suites did not run.

Representative completed core-suite logs include:

- [`linux-64`, Python 3.10, CUDA 13.3, L4](https://github.com/NVIDIA/cuda-python/actions/runs/33342679442/job/99341349217): 4,016 passed, 195 skipped, 2 xfailed, and 53 subtests passed.
- [`linux-aarch64`, Python 3.11, CUDA 12.9, L4](https://github.com/NVIDIA/cuda-python/actions/runs/33342679442/job/99341350953): 3,877 passed, 333 skipped, 2 xfailed, and 53 subtests passed.
- [`win-64`, Python 3.10, CUDA 12.9, RTX 2080](https://github.com/NVIDIA/cuda-python/actions/runs/33342679442/job/99341569248): 3,666 passed, 430 skipped, 2 xfailed, and 30 subtests passed.

### Final result and strict gate

The selective [CI run 33342679442](https://github.com/NVIDIA/cuda-python/actions/runs/33342679442) completed successfully on attempt 1:

- 102 jobs total
- 98 successful
- 4 intentionally skipped
- 0 failed or cancelled

The four intentional skips were exactly the two sdist jobs and the two core API-check jobs. The strict [`Check job status` gate](https://github.com/NVIDIA/cuda-python/actions/runs/33342679442/job/99343404631) passed and logged that every actual result matched its workplan-derived expectation: all platform build/test wrappers succeeded, while both sdist and both API-check results were `skipped` as expected.

The auxiliary [Bandit run](https://github.com/NVIDIA/cuda-python/actions/runs/33342679277) and [Security Suite](https://github.com/NVIDIA/cuda-python/actions/runs/33342679697) also completed successfully. Across all checks on the probe SHA, there were 110 successes, five expected skips (the four CI skips plus Security Suite preflight), and no failures.

As an observational data point rather than a controlled benchmark, the full baseline CI took 52m31s, while this selective core-test probe took 20m04s.

### Scope

This probe validates the successful `cuda_core` test-only path: exact-SHA baseline discovery, artifact completeness checks, wheel reuse/republication, module and job selection, Linux/ARM64/Windows testing, and the strict final gate's handling of intentional skips. It does not independently exercise source-change build fanout, platform-only infrastructure changes, unavailable baseline fallback, or the final gate's negative failure-propagation path.

Within that scope, I found no unexpected behavior or merge-blocking issue in #2467's selective orchestration.

rwgk (approved):
LGTM, based on
* a codex gpt-5.6-sol ultra review,
* the testing reported under [this PR comment](https://github.com/NVIDIA/cuda-python/pull/2467#issuecomment-5473880942), and
* glancing through all changes manually.

The only observation that stood out to me:

`ci/tools/compute_ci_plan.py` mixes data and algorithms. Ideally, the section starting from `MODULES` to `TEST_INFRA_PLATFORMS` would be factored into its own config-type file.

PR #2469, opened 2026-07-31
Bump cython dependency to 2.3.5

seberg:
As Keith noted, this is needed for using the `py_safe_call_once` definitions, Cython 3.2.5 changelog:
https://cython.readthedocs.io/en/latest/src/changes.html

(I guess the bump in the pre-commit is likely not strictly needed, but there also were no stub changes.)

This did change the docs entries to pin the same range.  Not really needed, but it did seem fine to keep consistent (and my agent said it keeps the pixi locks simpler).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -75,7 +75,7 @@ repos:
         pass_filenames: false
         additional_dependencies:
           - stubgen-pyx==0.2.6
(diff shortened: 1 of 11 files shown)


rparolin (approved):
lgtm

juenglin (approved):

PR #2470, opened 2026-07-31
ci: build cuda.core against the exact same-checkout cuda-bindings wheel

### LINKED ISSUE #2468: [BUG]: CI may build cuda.core against published cuda-bindings instead of the same-commit wheel
codex gpt-5.6-sol
___

## Type of bug

Silent failure

## Component

Infrastructure

## Describe the bug

The wheel and sdist CI workflows build a `cuda-bindings` development wheel from the current checkout and expose its directory through `PIP_FIND_LINKS` when building `cuda.core`. However, the `cuda.core` build backend requests only:

```text
cuda-bindings==<CUDA major>.*
```

`PIP_FIND_LINKS` adds candidates; it does not prioritize those candidates over the package index or enable prerelease selection. When a stable release satisfies the requirement, pip can therefore select the published stable wheel instead of the same-checkout `.dev` wheel.

This means `cuda.core` can be Cythonized against declarations from an older published `cuda-bindings` package even though CI has just built a newer `cuda-bindings` wheel from the PR. The resulting green build does not validate that the current `cuda.core` and `cuda-bindings` sources are compatible.

This occurred in PR #2395 (and was then discovered while locally testing PR #2393):

- The same-checkout bindings artifact was `13.3.2.dev...`.
- The `cuda.core` build requirement was `cuda-bindings==13.*`.
- Stable `cuda-bindings 13.3.1` was available from the package index.
- The published 13.3.1 `cydriver.pxd` declared `CUDA_MEMCPY3D.srcHost` as `const void*`.
- The current in-tree `cydriver.pxd` declared the field as `void*`.
- Consequently, `params.srcHost = <const void*>...` compiled in CI but failed
  when Cython consumed the current in-tree PXD.

The failure from a same-source editable QA build is recorded in `cuda-python_qa_bindings_linux_2026-07-31+003242_build_log.txt`:

```text
Assigning to 'void *' from 'const void *' discards const qualifier
```

The standard wheel tests do not expose the problem. They install the current bindings wheel only after the core wheel has already been Cythonized. The const qualification does not change the struct layout or ABI, so runtime testing continues to pass.

The paired pixi source-build workflow could catch this class of incompatibility, but its pull-request path filter does not include ordinary `cuda_core/cuda/**/*.pyx` changes.

## How to reproduce

1. Build a `.dev` `cuda-bindings` wheel from the current checkout.
2. Leave the normal package index enabled and expose the wheel only through `PIP_FIND_LINKS`.
3. Build `cuda.core`; its backend requests `cuda-bindings==13.*`.
4. Observe that the build succeeds against the published stable bindings PXD.
5. Editable-install the current bindings source and then build the same core source.
6. Observe that Cython instead consumes the in-tree PXD and reports the discarded-const error.

## Expected behavior

When CI builds `cuda.bindings` and `cuda.core` from the same checkout, the core build must use the exact bindings artifact produced from that checkout. CI should fail if it cannot resolve that artifact, and the selected bindings version and location should be visible in the build log.

## Possible fixes

- Make the dynamic core build requirement select the exact same-checkout bindings version or artifact. A direct local requirement or exact constraint is more robust than relying on `PIP_FIND_LINKS`; merely enabling prereleases would still leave room for a different index candidate to win later.
- Add a build-time assertion that reports and validates the resolved `cuda-bindings` version and PXD location.
- Expand the paired source-build workflow's pull-request path filter to cove ordinary `cuda_core` Cython sources.

rwgk:
## Summary

The `cuda.core` build backend requests `cuda-bindings==<CUDA major>.*`. CI previously made its newly built development wheel available only through `PIP_FIND_LINKS`.

`PIP_FIND_LINKS` adds candidates without prioritizing them over the package index. A published stable wheel can therefore beat a same-checkout development wheel when prereleases are not enabled. Enabling prereleases is not sufficient either, because a different, newer index candidate could still win. Even an exact version pin would leave the artifact source ambiguous if a tagged build were rerun after publication.

As a result, CI could successfully build current `cuda.core` sources against declarations from an older or otherwise unrelated published `cuda-bindings` package. The standard wheel tests install the current bindings wheel only after the core wheel has already been Cythonized, so they did not detect this mismatch. This is the false-green behavior described in #2468.

The build backend now recognizes the CI-only `CUDA_CORE_BUILD_BINDINGS_WHEEL_DIR` environment variable:

- when unset, it preserves the existing `cuda-bindings==<CUDA major>.*` requirement
- when set, it requires the directory to contain exactly one matching `cuda_bindings-<major>.*.whl`
- it returns that wheel as a direct PEP 508 `cuda-bindings @ file://...` build requirement
- it fails with a descriptive error for an empty setting, a missing wheel, or multiple matching same-major wheels
- it reports both the selected wheel and the imported bindings version and location in the build log

Filtering by CUDA major allows current- and previous-major wheels to coexist in one artifact directory while still making the selected dependency unambiguous.

The local artifact directory is wired into every same-checkout pip/PEP 517 `cuda.core` build in CI:

- current-major production wheels on Linux and Windows
- previous-major production wheels on Linux and Windows
- Linux and Windows wheel-from-sdist tests
- Linux and Windows coverage builds

Linux coverage now retains and installs a locally built bindings wheel before building core. Existing `PIP_FIND_LINKS` and `PIP_PRE` settings remain only where they are needed to keep the same-checkout prerelease `cuda-pathfinder` eligible; bindings selection no longer depends on pip candidate ranking.

Intentional compatibility coverage is unchanged. The explicit `BINDINGS_SOURCE=published` rows continue testing released same-major, older-minor `cuda-bindings` wheels from PyPI, and isolated Pixi environments that intentionally resolve published packages are unaffected. The analogous static `cuda-pathfinder` build-dependency selection is outside this PR.

Closes #2468.

___

## Fix the graph memcpy host-source pointer cast

**Already merged under PR #2476**

The current `cuda-bindings` declaration exposes `CUDA_MEMCPY3D.srcHost` as `void *`, but a refactor in #2395 unintentionally restored a `const void *` cast in `cuda.core`. Cython correctly rejected assigning that const-qualified pointer to the mutable field because it discards the qualifier.

The assignment now casts the source address to `void *`, matching the current declaration. A mutable pointer remains assignable when building against an older declaration that accepts `const void *`, so this keeps the intentional published-bindings compatibility path valid as well.

This source incompatibility had been masked by the first issue: CI was Cythonizing against the older published declaration rather than the current same-checkout PXD. Selecting the exact local bindings wheel exposed the failure and made the one-line correction necessary. The change does not alter structure layout, ABI, ownership, or runtime copy behavior.

## Avoid SIGPIPE in the cuda.core API check

**Already merged under PR #2475**

The API check selected the newest `cuda-core-v*` release tag with:

```bash
gh api ... --paginate | head -1
```

The workflow runs with `set -o pipefail`. Once `head` received the first matching tag, it closed the pipe while `gh` was still paginating. `gh` then received SIGPIPE and returned status 141, causing the job to fail even though tag selection had succeeded.

The pipeline now uses `sed -n '1p'`. It prints the first matching tag while continuing to consume the complete paginated stream, allowing `gh` to finish normally. A genuine `gh api` failure still propagates through `pipefail`, and the existing explicit check for an empty tag remains unchanged.

## Validation

- `TestVenv/bin/python -m pytest -v --noconftest cuda_core/tests/test_build_hooks.py ci/tools/tests` (`61 passed`)
- `pre-commit run --all-files`, including ruff, mypy, YAML validation, and actionlint
- pip dry-run using the generated direct local-wheel requirement
- build-hook coverage for the unchanged fallback, a development wheel in a path containing spaces, co-located CUDA 12 and 13 wheels, an empty setting, missing artifacts, and ambiguous same-major artifacts
- live API-tag selection plus a simulated producer failure to confirm that real `gh api` errors still propagate

## Coordination with #2464

PR #2464 independently evolved to address overlapping parts of these build and CI issues as part of its broader selective-wheel-build work. The implementations were cross-checked, and the detailed comparison is recorded in the PR discussion. Whichever PR lands second can rebase and reconcile the overlapping changes.


rwgk:
Added follow-up commit 71502adb7c to fix the unrelated API-check failure exposed by this PR.

It is not a flake:

- PR #2393 skipped both cuda.core API-check jobs because it did not change cuda.core.
- PR #2395's successful CI run predates the workflow change that introduced this job.
- Every observed execution after the job was introduced—the post-merge push, the following nightly run, and this PR—failed in `Find latest release tag` with exit 141.

The command used `gh api --paginate | head -1` under `pipefail`. After reading the first tag, `head` closed the pipe while `gh` was still paginating, so `gh` received SIGPIPE and the step failed.

The commit replaces `head -1` with `sed -n '1p'`. `sed` prints only the first (newest) matching tag but continues consuming the full stream, allowing `gh` to complete normally. This remains fail-closed: a genuine `gh api` failure still propagates through `pipefail`, and the existing empty-tag check is unchanged.

I kept this as a separate commit because the same issue is also being handled in #2464; it can be dropped easily if that PR lands first.

Validation:

- the updated command selected `cuda-core-v1.1.1` from the live repository tags
- a simulated producer failure propagated its nonzero status through the pipeline
- `pre-commit run --all-files` passed, including actionlint

rwgk:
Thanks @kkraus14 for pointing me to #2464. I did a fresh cross-check before continuing with this PR, since #2464 now overlaps with #2470 in three areas:

- selecting the exact same-checkout `cuda-bindings` wheel when building `cuda.core`
- fixing the `srcHost` Cython cast
- avoiding the API-check `head -1` / SIGPIPE failure

The exact-wheel work in #2464 is a valid solution for the production wheel workflow. Its direct-file constraints prevent an index candidate from taking precedence, and an earlier CI run surfaced the original Cython mismatch, which is useful confirmation that the local development wheel was selected. That approach also fits #2464's broader selective-build and artifact-reuse work, and it additionally constrains `cuda-pathfinder`.

#2470 takes a different approach for issue #2468: the `cuda.core` build hook selects an exact local bindings wheel when the CI-only environment variable is set. That selection is currently wired into the current- and previous-major production wheel builds, the Linux and Windows sdist tests, and the Linux and Windows coverage builds. It also provides explicit selection/import diagnostics and focused unit tests, while leaving ordinary builds unchanged when the variable is unset.

The `srcHost` changes in the two PRs are equivalent. The API-check fixes differ slightly: #2470 uses `sed -n '1p'` so the complete paginated stream is consumed while a genuine `gh api` failure still propagates through `pipefail`; #2464 currently uses process substitution with `|| true`.

Based on this cross-check, I think it still makes sense to continue #2470 as the focused fix for #2468. In particular, #2464 does not currently apply its exact bindings constraint to the sdist or coverage workflows. At the same time, its constraint-based implementation remains useful within the selective-build work.

Whichever PR lands second will need a small rebase and de-duplication of the overlapping cast and API-check changes. I wanted to record the comparison here so it is clear that #2470 is continuing with the current state of #2464 taken into account.

cuda_core/tests/test_build_hooks.py
```diff
@@ -135,3 +135,58 @@ def test_missing_cuda_path_raises_error(self):
...
+
+        assert build_hooks._get_cuda_bindings_require() == [f"cuda-bindings @ {matching_wheel.resolve().as_uri()}"]
+
+    def test_local_wheel_directory_must_not_be_empty(self, monkeypatch):
+        monkeypatch.setenv("CUDA_CORE_BUILD_MAJOR", "13")
+        monkeypatch.setenv("CUDA_CORE_BUILD_BINDINGS_WHEEL_DIR", "")
+        build_hooks._determine_cuda_major_version.cache_clear()
+
+        with pytest.raises(RuntimeError, match="CUDA_CORE_BUILD_BINDINGS_WHEEL_DIR must not be empty"):
+            build_hooks._get_cuda_bindings_require()
```
juenglin, line 170: Doesn't the next test cover this case (empty is also not "exactly one")?

juenglin (approved):

rwgk:
I converted this PR to Draft, in favor of #2510 (which I still need to review more closely myself).

rwgk:
Closing, after #2510 was merged.

PR #2472, opened 2026-07-31
build: make Cython/extension artifacts CUDA-major aware

rparolin:
## What

Moving a checkout between the `cu12` and `cu13` pixi environments fails at compile time, pointing at code that is perfectly fine. Reproduced on `main` at 755cf61 — `cu13` build, then `cu12`:

```
build/cython/cuda/core/_device_resources.cpp:4143:76:
  error: 'CUdevWorkqueueConfigScope' was not declared in this scope
```

Nothing tells you the real cause: leftover generated C++ from the previous CUDA major.

This is the configuration-aware approach @kkraus14 asked for in #2443, in place of a destructive cleaner.

## Why it happens

Two caches, neither of which notices the CUDA major changed.

**Cython does not hash `compile_time_env`.** `cuda.core` feeds `CUDA_CORE_BUILD_MAJOR` through it, so a `cu13 -> cu12` switch silently reuses the `cu13` generated C++. That is the error above.

**setuptools' `build_ext` compares source mtimes against the output `.so`.** In an editable install that `.so` lives in the source tree under a name keyed by the Python ABI tag alone — there is nowhere to record the CUDA major. On a `cu12 -> cu13 -> cu12` round trip the final build finds an older generated source next to a newer `.so` and skips the rebuild entirely.

## The change

Two things, both in `cuda_core`:

- Generated sources go to `build/cython/cu<major>` instead of a single shared `build/cython`, so each major keeps its own cache.
- `build/.build-cuda-major` records the last **completed** build (written after `build_ext` succeeds, so a failed build does not claim outputs it never produced). When it differs, `setup.py` sets `build_ext.force`.

The second is what handles the return leg, where directory keying can't help.

`CUDA_PYTHON_COVERAGE` is untouched, as requested — it still generates in-tree (`build_dir="."`) so it can package the generated sources.

**Migration is self-healing.** An existing unkeyed `build/cython` is orphaned and ignored; the first build after this change regenerates into a keyed directory and overwrites the in-tree extensions. No cleanup step required.

## Verification

`cu13 -> cu12 -> cu13` with pixi on linux-64. Before this PR the `cu12` leg fails to compile. After, all three legs build, `build/cython/cu12` and `build/cython/cu13` both exist, and the extension from the second `cu13` leg is **byte-identical** to the first (`md5 1c53ab83a7201d54576175c912d03e93`).

That last point is the load-bearing assertion. Leg 1's generated `.cpp` is older than leg 2's `.so`, so without the forced rebuild `newer_group()` would have skipped and left the `cu12` artifact in place — and the build would still have "succeeded". Only the byte-identical result distinguishes a real rebuild from a wrong skip.

New `build-identity-roundtrip` job in `ci-pixi-source-test.yml` runs that sequence on every relevant PR. 18 unit tests cover the stamp logic; `pre-commit` clean.

## Scope

Deliberately narrow — the CUDA major is the only axis keyed, and only `cuda_core` is touched.

- **`cuda_bindings` is not changed.** It has no `compile_time_env`, so its generated C++ is CUDA-major independent, and it cannot be source-built against CUDA 12 at all today (its sources reference `CUatomicOperation`, `nvrtcBundledHeadersInfo`, `CUstreamCigCaptureParams`). That is a separate pre-existing problem, worth its own issue; I can file it.
- **`debug` and coverage are not keyed.** Both change the compiled output, and toggling either has a similar staleness problem — but it is a different defect from the one this PR is titled for. Worth noting for anyone tempted to add them: `gdb_debug=True` does *not* change the generated C++ (I tested it — the `.c` is byte-identical; Cython only writes a side `cython_debug/` directory), so keying a directory by it would only duplicate identical sources. `linetrace` genuinely does change the generated C++.
- **The `cu12` leg resolves published `cuda-bindings` 12.x**, not the local checkout — `cuda_core`'s `cu12` environment omits `local-deps` by design.

## Relationship to #2443

#2443 proposed `pixi run clean` as the workflow. Per review feedback, this PR makes the artifacts configuration-aware so cleanup is not the correctness boundary. #2443 can stand down to narrowed recovery tooling.

---

| Check | Status | Tool | Details | Artifact |
|-------|--------|------|---------|----------|
| Secrets | SKIP | Pulse Secret Scanner | Image pull denied — not authenticated to `gitlab-master.nvidia.com:5005` (needs a PAT with `read_registry`) | `.plc/security/pulse-secret-scan.json` (skip-stub) |
| SAST | WARN | Semgrep 1.157.0 | Findings are pre-existing and untouched by this diff: `.github/workflows/ci-pixi-source-test.yml run-shell-injection` (the existing nightly `full-test` job), plus `tempfile-without-flush` / `dangerous-subprocess-use-audit` in `cuda_bindings`, which this PR no longer modifies | `.plc/security/semgrep.json` |
| Dependencies | WARN | osv-scanner / pip-audit | Source in a dependency-bearing ecosystem changed but no manifest or lockfile is in scope. The change adds no dependencies — the new code imports only the standard library — so there is nothing to resolve; recorded rather than suppressed | `.plc/security/dependency-*.json` |
| License | SKIP | pip-licenses | No installed-distribution metadata in scope; no dependencies added. BlackDuck + OSRB remain authoritative | `.plc/security/pip-licenses.json` (skip-stub) |
| Container | N/A | — | No container files in this change | — |
| Security Review | PASS | inline | Reviewed the file-writing path: the stamp is written to a fixed relative path under `build/`, contents are a CUDA major string with no user-controlled input, and nothing is deleted. No shell invocation added | inline in PR diff |


cuda_core/build_hooks.py
```diff
@@ -119,6 +119,51 @@ def _determine_cuda_major_version() -> str:
 # used later by setup()
 _extensions = None
 
+# Records the CUDA major of the last completed build, so setup.py can force
+# build_ext when it changes. Written by record_build_major().
+_BUILD_MAJOR_STAMP = os.path.join("build", ".build-cuda-major")
```
mdboom: This will be relative to `pwd`, and you can build a project from anywhere.  It should instead be something like:

`Path(__file__).parent / "build" / ".build-cuda-major"`

(and we should use `pathlib.Path` for all new code, not the soft-deprecated `os.path` APIs).

cuda_core/build_hooks.py
```diff
@@ -119,6 +119,51 @@ def _determine_cuda_major_version() -> str:
...
+    global force_build_ext
+
+    cuda_major = _determine_cuda_major_version()
+    try:
+        with open(_BUILD_MAJOR_STAMP, encoding="utf-8") as f:
+            previous = f.read().strip()
+    except FileNotFoundError:
+        previous = None
+
+    if previous is not None and previous != cuda_major:
```
mdboom: IIUC, if `previous` is `None`, we failed to determine whether we need a full rebuild, so we should do a full rebuild.

```suggestion
    if previous != cuda_major:
```

cuda_core/build_hooks.py
```diff
@@ -119,6 +119,51 @@ def _determine_cuda_major_version() -> str:
...
+    return cuda_major
+
+
+def record_build_major() -> None:
+    """Stamp the CUDA major of the build that just completed.
+
+    setup.py calls this after build_ext succeeds, so that a build which failed
+    partway through does not claim outputs it never produced.
+    """
+    os.makedirs(os.path.dirname(_BUILD_MAJOR_STAMP), exist_ok=True)
```
mdboom: Ditto for `pathlib`.

mdboom (changes requested):

.github/workflows/ci-pixi-source-test.yml
```diff
@@ -108,12 +110,69 @@ jobs:
...
+          if [ "${stamp}" != "13" ]; then
+            echo "::error::build stamp is '${stamp}', expected 13"
+            exit 1
+          fi
+          for major in cu12 cu13; do
+            if [ ! -d "cuda_core/build/cython/${major}" ]; then
+              echo "::error::no ${major} generated-source directory"
+              exit 1
+            fi
+          done
```
mdboom, line 171: Maybe to address later -- but I worry about the time of this test for sort of a niche problem.  Is there a way we could "simulate" a build doing the wrong thing rather than doing a full build?

rparolin: I took a look at this and introduced some fast tests that just check if the build system picks up that a rebuild needs to happen (detected change from cu13 -> cu12 for example).   However, the timinings on cluster show that the round trip build smoke (cu13 -> cu12 -> cu13) runs in parallel for 12m+.  So it doesn't increase walltime waiting for a PR result.  The only concern would be load on the CI cluster which shouldn't be a major concern b/c we are using the public gh actions runners (not our GPU runners).  I recommend merging this and if we discover that we are seeing negative effects on PR walltime we can move that round trip build to run once nightly.  sg?

From Claude: 

> Good call — added those, and they found a real bug on the way.
> 
> **Fast tests** (`cuda_core/tests/test_build_hooks.py`, ~0.1s, nothing compiles): cu12/cu13 generate into different directories; that directory is anchored so it agrees with the stamp from any cwd; `build_ext.finalize_options()` picks up the force flag, and leaves it alone when clear. I confirmed each fails when I revert the matching fix.
> 
> One supporting change: `setup()` is behind `if __name__ == "__main__":` so the command classes can be imported. setuptools runs setup.py as a script, so builds are unaffected — verified with a clean rebuild.
> 
> (These turned up a hazard worth knowing: `_build_cuda_core()` permanently prepends `cuda_bindings/` to `sys.path`, so a later bare `import build_hooks` gets *cuda_bindings'* copy.)
> 
> **On the cost** — same run, cold GitHub-hosted `ubuntu-latest`:
> 
> | job | elapsed |
> |---|---|
> | `cu13 -> cu12 -> cu13 round trip` | 12m15s |
> | `build smoke (cu13)` | 23m55s |
> 
> They start together and the round trip finishes ~11m40s first, so it adds no wall-clock time — the workflow waits on build smoke regardless. It's ~1.3% of this PR's CI (102 jobs / 947 runner-minutes), fires on ~1 in 5 PRs via the `paths:` filter, and is CPU-only on GitHub's free public-repo pool, so no GPU runners are touched. If it ever is a burden, moving it to nightly is a one-line change.
> 
> **Why keep it:** the fast tests check that we do what we intended. They can't check that Cython still ignores `compile_time_env` or that setuptools still skips on mtime — and the bug happened because those assumptions were wrong. The `paths:` filter already covers `pixi.lock`/`pixi.toml`, so it fires exactly on the dependency bumps that could invalidate them.

mdboom: > So it doesn't increase walltime waiting for a PR result.

> The only concern would be load on the CI cluster which shouldn't be a major concern b/c we are using the public gh actions runners (not our GPU runners).

We are limited to a number of concurrent jobs per project, though.  I'm not sure what "tier" we are in, but I see it get bottlenecked all the time.

rparolin: Moved `build-identity-roundtrip` to nightly-only in e0bd7ffe7e5. It no longer runs on PRs, so it won't use up a job slot there. The fast unit tests in `cuda_core/tests/test_build_hooks.py` still catch the bug on every PR.

mdboom (approved):

PR #2481, opened 2026-08-03
cuda.core: return CUmodule via as_py in ObjectCode.get_module

### LINKED ISSUE #2415: Create `as_py` function for `CUModule` type
I think we can create a new `as_py` function in `_resource_handle.hpp` that accpets cumodule type. But this can be set as a follow up issue.

_Originally posted by @isVoid in https://github.com/NVIDIA/cuda-python/pull/2339#pullrequestreview-4768583702_

lijinf2:
Use the shared handle export path for legacy CUmodule interop instead of constructing driver.CUmodule directly.

## Description

closes #2415

Add an `as_py(const CUmodule&)` overload and use it in `ObjectCode.get_module()` so legacy `CUmodule` export goes through the same handle→Python wrapper path as other CUDA resources.

`CUmodule` from `cuLibraryGetModule` is not an owning `*Handle` (`shared_ptr`); its lifetime is owned by the parent `CUlibrary`, so this overload takes the regular pointer and only wraps it as `cuda.bindings.driver.CUmodule`.

## Checklist


cuda_core/cuda/core/_cpp/resource_handles.hpp
```diff
@@ -968,6 +968,12 @@ inline PyObject* as_py(const LibraryHandle& h) noexcept {
     return detail::make_py("cuda.bindings.driver", "CUlibrary", as_intptr(h));
 }
 
+// Regular pointer (CUmodule), not a shared_ptr: lifetime is owned by the
+// CUlibrary it was retrieved from.
+inline PyObject* as_py(const CUmodule& h) noexcept {
+    return detail::make_py("cuda.bindings.driver", "CUmodule", reinterpret_cast<std::intptr_t>(h));
```
isVoid: We should extend the overload for `as_intptr(h)` for CUmodule above?

lijinf2: Done.

cuda_core/cuda/core/_cpp/resource_handles.hpp
```diff
@@ -838,6 +838,12 @@ inline std::intptr_t as_intptr(const LibraryHandle& h) noexcept {
     return reinterpret_cast<std::intptr_t>(as_cu(h));
 }
 
+// Regular pointer (CUmodule), not a shared_ptr: lifetime is owned by the
+// CUlibrary it was retrieved from.
+inline std::intptr_t as_intptr(const CUmodule& h) noexcept {
+    return reinterpret_cast<std::intptr_t>(h);
```
isVoid: Yeah.. I think the agent is not quite smart here. There's a pretty obvious pattern here: `as_cu` -> `as_intptr` -> `as_py`, but the agent isn't smart enough to pick it up. Also the comment is not really needed here.

lijinf2: Yes, agents chose not to revise `as_cu` and `as_intptr` and it needs clear instructions in the prompt. 
Just revised `as_cu` to support CUModule.

isVoid (approved):

PR #2484, opened 2026-08-03
Add mitigations for intermittent test failures with CUDA OOM  

juenglin:
Addresses #2381.

Test-suite hygiene for the address-space exhaustion described in [this comment](https://github.com/NVIDIA/cuda-python/issues/2381#issuecomment-5169825707). Two of the three contributing factors are in the tests themselves, and this PR takes both.

**Deterministic release at test boundaries.** `_init_cuda_context` now collects and
drains before dropping the context. A pool's virtual address reservation is not
returned until the pool is destroyed *and* the stream-ordered frees of its outstanding
allocations retire; previously neither was forced between tests, so reservations from
earlier tests stayed live while later tests created their own.

**Right-sized pools.** Seven pool constructions passed options without setting
`max_size`, so each reserved a system-dependent window scaling with device memory while
holding at most 1 KiB:
- `test_memory_peer_access.py` — four device pools, including the `isolated_dmr_x2`
  fixture (once per test across seven tests) and a comprehension that reserved one
  window per device. All hold at most `NBYTES = 1024`.
- `test_memory.py` — three pinned pools in the `numa_id` tests, which only read
  `mr.numa_id` and never allocate.
All now pass the suite-standard `POOL_SIZE = 2097152`, newly introduced to
`test_memory_peer_access.py` with a comment explaining why a peer-access test caps a
pool it barely uses.

### Measured impact
On a 23.9 GiB device, **20** uncapped pools exhaust the process address space, versus
125 pools capped at 4 GiB — an uncapped pool costs roughly 1x device memory in
reservation, a 2 MiB cap costs 2 MiB.

### Deliberately not included
Managed pools cannot be right-sized: `cuMemPoolCreate` documents that
`CUmemPoolProps.maxSize` **must be zero** for managed pools, so
`ManagedMemoryResourceOptions` correctly has no `max_size` option.

## Checklist


Andy-Jost (approved):
LGTM!

Issue #2410, opened 2026-07-23
Migrate from os.path to pathlib.Path


mdboom:
`os.path` is an old API that is soft-deprecated in favor of the type-safe `pathlib.Path` API.

We should migrate to using `pathlib.Path` everywhere internally, with care being take to ensure that we still accept strings everywhere we currently do so as not to break public API.

mdboom:
Unassigning myself.  I think this would be a great contributor bug and I don't want to hold it while I'm busy.

LeSingh1:
@mdboom I'd like to pick this up, if it's still free.

I scoped the current surface on `main` to check how it should be broken up:

| Area | `os.path.*` call sites |
| --- | --- |
| `cuda_pathfinder` | 163 |
| `cuda_core` | 25 |
| `cuda_bindings` | 19 |
| `toolshed` | 7 |
| `ci` | 6 |
| root `conftest.py` / test helpers | 2 |
| **total** | **222 across 41 files** |

That's too much for one reviewable diff, so I'd suggest doing it as a small series scoped by package, each independently reviewable and revertable, rather than one sweeping change:

1. `cuda_pathfinder/_dynamic_libs` (`search_platform.py`, `search_steps.py`)
2. `cuda_pathfinder/_static_libs` (`find_static_lib.py`, `find_bitcode_lib.py`)
3. `cuda_pathfinder/_binaries` + `_utils` (`env_vars.py`, `find_sub_dirs.py`)
4. `cuda_pathfinder/tests`
5. `cuda_core` (incl. `build_hooks.py`)
6. `cuda_bindings` (incl. `build_hooks.py`)
7. `toolshed` / `ci` / root `conftest.py`

On the API-compatibility point you raised: I'd keep every public entry point accepting `str` exactly as it does today, converting to `Path` internally and converting back at the boundary where a function currently documents or returns `str`. `cuda.pathfinder`'s public returns look like the main risk area, so I'd treat "no public signature or return type changes" as a hard constraint and cover it with tests asserting the returned types stay as they are. Anything that would change a return type I'd leave alone and call out separately rather than fold in silently.

Happy to start with (1) as a single small PR so you can sanity-check the conversion style and compat approach before I do the rest — if that shape works for you, could you assign me? And if you'd rather this land as one big diff, or want a different split, I'll follow whatever you prefer.

LeSingh1:
One finding worth raising before the rest of the series lands, because it affects how every chunk should be written.

[...]

PR #2489, opened 2026-08-04
Migrate cuda_pathfinder/_dynamic_libs search modules from os.path to pathlib

LeSingh1:
Part 1 of the series proposed in #2410: `cuda_pathfinder/cuda/pathfinder/_dynamic_libs/` — `search_platform.py` and `search_steps.py`. Deliberately small so the conversion style and the compatibility approach can be reviewed before the remaining six parts follow the same template.

### What changed

Path construction, joining, and filesystem predicates now go through `Path`/`PurePath` instead of `os.path` string manipulation. `search_platform.py` no longer imports `os` at all.

### Compatibility

Treated as a hard constraint, per the discussion on the issue. Every entry point still accepts `str`, and every function that documents or returns `str` still returns `str` — `Path` is used strictly as the internal representation and converted back with `str()` at each return. `SearchPlatform`, `FindResult.abs_path`, and the `LoadedDL.abs_path` that reaches users are unchanged in both type and value. No signature changes.

### Deliberately not converted

Rather than folding these in silently, both are left alone with a comment saying why:

- `_find_lib_dir_using_anchor` keeps `os.path.normpath`. pathlib has no equivalent because `PurePath` intentionally does not collapse `..`, so converting would change the path reported for an anchor such as a `CUDA_PATH` containing `..`.
- `_derive_ctk_root_windows` keeps `ntpath`. `ntpath.dirname` slices its input and preserves the caller's separators; `PureWindowsPath` rewrites them to backslashes. `derive_ctk_root` also reaches this function on Linux, so converting would return an unusable root (`\opt\foo`) for a POSIX path whose parent directory happens to be named `bin`.

`glob.glob` is also retained where wildcard expansion is needed — `Path.glob` orders results differently, and the reverse-sorted newest-first policy from #1732 depends on the current string sort.

Happy to revisit any of these if you'd prefer them handled differently; that decision will set the pattern for the rest of the series.

### Verification

No behavior change intended. Beyond the test suite, I fuzzed the old and new implementations against each other: 200k randomized path shapes for the `derive_ctk_root` helpers, and 3000 randomized real directory trees (including trailing-slash, `//` and `/./` spellings) for the search helpers. The only string-level divergence is that a redundant `.` component in an input directory is now collapsed — the new string is the normalized form of the old and names the same file — and that input cannot occur, since `lib_dir` always arrives already `normpath`-ed.

`pre-commit run` passes every hook at the pinned versions; `mypy` is clean over `cuda_pathfinder/cuda`.

Since I don't have CUDA hardware, I verified the suite on Linux CI rather than locally. Against `upstream/main` the full pytest output is byte-identical apart from elapsed time — same 1188 passed, 4 skipped, and the same 88 `Not found` errors for absent NVIDIA libraries, in the same order.

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_platform.py
+++ b/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_platform.py
@@ -10,7 +10,6 @@ on OS flags like ``IS_WINDOWS``. Instead, it calls through the single
 
 from __future__ import annotations
 
-import glob
 from collections.abc import Sequence
 from dataclasses import dataclass
 from pathlib import Path, PurePath
@@ -23,13 +22,30 @@ from cuda.pathfinder._utils.platform_aware import IS_WINDOWS
 from cuda.pathfinder._utils.windows_arch import windows_pe_matches_arch, windows_python_arch
 
 
+def sorted_glob(directory: Path, pattern: str, *, reverse: bool = False) -> list[Path]:
+    """Return ``directory.glob(pattern)`` matches in a deterministic order.
+
+    The ordering is deliberately taken from the string form rather than from
+    ``Path`` comparison: ``PurePath`` ordering is case-insensitive on Windows,
+    so plain ``sorted()`` over ``Path`` objects would reorder mixed-case
+    filenames relative to the byte-wise ordering used up to now. Issue #1732
+    tracks the newest-first policy that rides on this ordering, so it is kept
+    unchanged on both platforms.
+    """
+    return sorted(directory.glob(pattern), key=str, reverse=reverse)
+
+
+def _sorted_dir_entry_names(directory: Path) -> list[str]:
+    return sorted(entry.name for entry in directory.iterdir())
+
+
 def _no_such_file_in_sub_dirs(
     sub_dirs: Sequence[str], file_wild: str, error_messages: list[str], attachments: list[str]
 ) -> None:
     error_messages.append(f"No such file: {file_wild}")
     for sub_dir in find_sub_dirs_all_sitepackages(sub_dirs):
         attachments.append(f'  listdir("{sub_dir}"):')
-        for node in sorted(node_path.name for node_path in Path(sub_dir).iterdir()):
+        for node in _sorted_dir_entry_names(Path(sub_dir)):
             attachments.append(f"    {node}")
 
 
@@ -38,7 +54,7 @@ def _find_so_in_rel_dirs(
     so_basename: str,
     error_messages: list[str],
     attachments: list[str],
-) -> str | None:
+) -> Path | None:
     sub_dirs_searched: list[tuple[str, ...]] = []
     file_wild = so_basename + "*"
     for rel_dir in rel_dirs:
@@ -50,29 +66,29 @@ def _find_so_in_rel_dirs(
             # multiple coexist, matching the newest-first bias elsewhere in pathfinder
             # (see LinuxSearchPlatform.find_in_lib_dir and load_dl_linux._candidate_sonames).
             # Issue #1732 tracks the deferred question of raising on true ambiguity.
-            so_path = Path(abs_dir, so_basename)
+            abs_dir_path = Path(abs_dir)
+            so_path = abs_dir_path / so_basename
             if so_path.is_file():
-                return str(so_path)
-            for match in sorted(glob.glob(str(Path(abs_dir, file_wild))), reverse=True):
-                so_path = Path(match)
+                return so_path
+            for so_path in sorted_glob(abs_dir_path, file_wild, reverse=True):
                 if so_path.is_file():
-                    return str(so_path)
+                    return so_path
         sub_dirs_searched.append(sub_dir)
     for sub_dir in sub_dirs_searched:
         _no_such_file_in_sub_dirs(sub_dir, file_wild, error_messages, attachments)
     return None
 
 
-def _find_dll_under_dir(dirpath: str, file_wild: str, target_arch: str | None = None) -> str | None:
-    for match in sorted(glob.glob(str(Path(dirpath, file_wild)))):
-        path = Path(match)
+def _find_dll_under_dir(dirpath: Path, file_wild: str, target_arch: str | None = None) -> Path | None:
+    for path in sorted_glob(dirpath, file_wild):
         if not path.is_file():
             continue
         if is_suppressed_dll_file(path.name):
             continue
+        # windows_pe_matches_arch() lives in _utils and is still str-typed.
         if target_arch is not None and not windows_pe_matches_arch(str(path), target_arch):
             continue
-        return str(path)
+        return path
     return None
 
 
@@ -81,14 +97,14 @@ def _find_dll_in_rel_dirs(
     lib_searched_for: str,
     error_messages: list[str],
     attachments: list[str],
-) -> str | None:
+) -> Path | None:
     sub_dirs_searched: list[tuple[str, ...]] = []
     for rel_dir in rel_dirs:
         sub_dir = PurePath(rel_dir).parts
         for abs_dir in find_sub_dirs_all_sitepackages(sub_dir):
-            dll_name = _find_dll_under_dir(abs_dir, lib_searched_for)
-            if dll_name is not None:
-                return dll_name
+            dll_path = _find_dll_under_dir(Path(abs_dir), lib_searched_for)
+            if dll_path is not None:
+                return dll_path
         sub_dirs_searched.append(sub_dir)
     for sub_dir in sub_dirs_searched:
         _no_such_file_in_sub_dirs(sub_dir, lib_searched_for, error_messages, attachments)
@@ -100,7 +116,7 @@ class SearchPlatform(Protocol):
 
     def site_packages_rel_dirs(self, desc: LibDescriptor) -> tuple[str, ...]: ...
 
-    def conda_anchor_point(self, conda_prefix: str) -> str: ...
+    def conda_anchor_point(self, conda_prefix: str) -> Path: ...
 
     def anchor_rel_dirs(self, desc: LibDescriptor) -> tuple[str, ...]: ...
 
@@ -110,16 +126,16 @@ class SearchPlatform(Protocol):
         lib_searched_for: str,
         error_messages: list[str],
         attachments: list[str],
-    ) -> str | None: ...
+    ) -> Path | None: ...
 
     def find_in_lib_dir(
         self,
-        lib_dir: str,
+        lib_dir: Path,
         desc: LibDescriptor,
         lib_searched_for: str,
         error_messages: list[str],
         attachments: list[str],
-    ) -> str | None: ...
+    ) -> Path | None: ...
 
 
 @dataclass(frozen=True, slots=True)
@@ -130,8 +146,8 @@ class LinuxSearchPlatform:
     def site_packages_rel_dirs(self, desc: LibDescriptor) -> tuple[str, ...]:
         return cast(tuple[str, ...], desc.site_packages_linux)
 
-    def conda_anchor_point(self, conda_prefix: str) -> str:
-        return conda_prefix
+    def conda_anchor_point(self, conda_prefix: str) -> Path:
+        return Path(conda_prefix)
 
     def anchor_rel_dirs(self, desc: LibDescriptor) -> tuple[str, ...]:
         return cast(tuple[str, ...], desc.anchor_rel_dirs_linux)
@@ -142,22 +158,21 @@ class LinuxSearchPlatform:
         lib_searched_for: str,
         error_messages: list[str],
         attachments: list[str],
-    ) -> str | None:
+    ) -> Path | None:
         return _find_so_in_rel_dirs(rel_dirs, lib_searched_for, error_messages, attachments)
 
     def find_in_lib_dir(
         self,
-        lib_dir: str,
+        lib_dir: Path,
         _desc: LibDescriptor,
         lib_searched_for: str,
         error_messages: list[str],
         attachments: list[str],
-    ) -> str | None:
-        lib_dir_path = Path(lib_dir)
+    ) -> Path | None:
         # Most libraries have both unversioned and versioned files/symlinks (exact match first)
-        so_path = lib_dir_path / lib_searched_for
+        so_path = lib_dir / lib_searched_for
         if so_path.is_file():
-            return str(so_path)
+            return so_path
         # Some libraries only exist as versioned files (e.g., libcupti.so.13 in conda),
         # so the glob fallback is needed
         file_wild = lib_searched_for + "*"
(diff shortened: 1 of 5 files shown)


LeSingh1:
@mdboom — this is the part 1 I offered on #2410, sent as a small sample so the conversion style and the compatibility approach can be checked before the remaining six parts follow the same template. Happy to hold the rest until you've had a look, or to drop this entirely if you'd rather assign it elsewhere.

Two notes for whoever picks it up:

`pr-metadata-check` needs a milestone and I can't set one as an outside contributor — the `cuda.pathfinder` label applied but the milestone didn't. `cuda.pathfinder next` looks like the right one.

On verification: I don't have CUDA hardware, so I ran the suite on Linux CI rather than locally. Rather than just matching the pass count, I diffed the full pytest output against `upstream/main` with timestamps stripped — it's byte-identical apart from elapsed time, including the same 88 `Not found` errors for absent NVIDIA libraries in the same order. I also fuzzed the old and new implementations against each other (200k randomized path shapes for the `derive_ctk_root` helpers, 3000 randomized directory trees for the search helpers), which is what surfaced the two sites I left on `os.path`/`ntpath` and documented in the description — the `ntpath` one is a genuine functional regression on Linux, not a style preference.

mdboom:
A few implementation comments:

> Treated as a hard constraint, per the discussion on the issue. Every entry point still accepts str, and every function that documents or returns str still returns str — Path is used strictly as the internal representation and converted back with str() at each return. SearchPlatform, FindResult.abs_path, and the LoadedDL.abs_path that reaches users are unchanged in both type and value. No signature changes.

This only needs to apply to functions in the public API.  We are free to change the implementation details of internal function calls as we see fit, so we should prefer `pathlib.Path` wherever feasible.

> glob.glob is also retained where wildcard expansion is needed — Path.glob orders results differently, and the reverse-sorted newest-first policy from https://github.com/NVIDIA/cuda-python/issues/1732 depends on the current string sort.

All uses of `glob.glob` are passed to `sorted` anyway, so why does the sorting order matter?  In the changes here we are paying a penalty of converting between `Path` and `str` objects, so it would be preferable to stay in the `pathlib.Path` space if possible.

LeSingh1:
Done in 5d5f21a.

Internals and the `SearchPlatform` protocol now pass and return `Path`. `str()` is applied only where a value leaves these modules: `FindResult.abs_path` (which surfaces unchanged as the public `LoadedDL.abs_path`) and `derive_ctk_root()`. All four `glob.glob` sites are now `Path.glob` — you were right that the `sorted()` wrapper makes the ordering argument moot.

One thing worth flagging: `sorted()` over `Path` compares `PurePath._str_normcase`, which is lowercased on Windows, so it isn't the same order as `sorted()` over `str` for mixed-case names. I kept the existing order by sorting on the string form in a small `sorted_glob()` helper, rather than change Windows behaviour in a PR that otherwise changes nothing — and I have no Windows machine to check it on. Happy to drop the key for simpler code if you'd prefer.

`os.path.normpath` in `_find_lib_dir_using_anchor` stays: no pathlib equivalent, since `PurePath` doesn't collapse `..` and `resolve()` would also follow symlinks. `_derive_ctk_root_windows` stays on `ntpath` — I re-checked, and it still matters because `derive_ctk_root` reaches it on Linux, where `PureWindowsPath` turns `/opt/foo/bin/libx.so` into `\opt\foo`.

**A correction to my earlier comment.** I said the CI baseline was "1188 passed and 88 `Not found` errors for absent NVIDIA libraries". That was wrong, and I should have looked harder before asserting it. The 88 were `fixture 'mocker' not found` — my fork-side test job was missing `pytest-mock`, so those 88 tests never ran. The `Not found: libname=` lines I took for the cause were unrelated INFO logging. I've fixed the job to install the `test` dependency group and pinned `-p no:randomly`, and the real numbers are **1276 passed, 4 skipped, 0 errors**, identical on the PR's base commit and on this revision — full pytest output byte-identical apart from elapsed time.

Differential fuzzing against the previous revision (400k randomized path shapes, 600 randomized directory trees) shows one intended difference: `conda_anchor_point()` returns a normalized `Path` rather than `$CONDA_PREFIX` verbatim. Not observable downstream, since `_find_lib_dir_using_anchor` already normpath'd its result.

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_steps.py
```diff
@@ -37,7 +37,12 @@
 
 @dataclass
 class FindResult:
-    """A library file located on disk (not yet loaded)."""
+    """A library file located on disk (not yet loaded).
+
+    ``abs_path`` is a ``str``: it is handed to the platform loaders and surfaces
+    unchanged as the public ``LoadedDL.abs_path``. Everything upstream of this
+    dataclass works with ``Path``.
```
mdboom: Unnecessary comment.

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_steps.py
```diff
@@ -138,7 +155,11 @@ def _derive_ctk_root_windows(resolved_lib_path: str) -> str | None:
 
 
 def derive_ctk_root(resolved_lib_path: str) -> str | None:
-    """Derive CTK root from a resolved canary library path."""
+    """Derive CTK root from a resolved canary library path.
+
+    Returns a ``str`` because the result crosses into ``load_nvidia_dynamic_lib``
+    and the header search, which are outside this module.
```
mdboom: Unnecessary comment

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_steps.py
```diff
@@ -37,7 +37,12 @@
...
-    """A library file located on disk (not yet loaded)."""
+    """A library file located on disk (not yet loaded).
+
+    ``abs_path`` is a ``str``: it is handed to the platform loaders and surfaces
+    unchanged as the public ``LoadedDL.abs_path``. Everything upstream of this
+    dataclass works with ``Path``.
+    """
 
     abs_path: str
     found_via: str
```
mdboom: Let's store these as `Path` with a custom field converter.  That way all users of this class don't have to convert their paths back to strings and back again.


```suggestion
abs_path: Path = field(converter=Path)
found_via: str
```

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_steps.py
```diff
@@ -147,11 +168,11 @@ def derive_ctk_root(resolved_lib_path: str) -> str | None:
 
 def find_via_ctk_root(ctx: SearchContext, ctk_root: str) -> FindResult | None:
     """Find a library under a previously derived CTK root."""
-    lib_dir = _find_lib_dir_using_anchor(ctx.desc, ctx.platform, ctk_root)
+    lib_dir = _find_lib_dir_using_anchor(ctx.desc, ctx.platform, Path(ctk_root))
     abs_path = _find_using_lib_dir(ctx, lib_dir)
     if abs_path is None:
         return None
-    return FindResult(abs_path, "system-ctk-root")
+    return FindResult(str(abs_path), "system-ctk-root")
```
mdboom:
```suggestion
    return FindResult(abs_path, "system-ctk-root")
```

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_steps.py
```diff
@@ -166,7 +187,7 @@ def find_in_site_packages(ctx: SearchContext) -> FindResult | None:
         return None
     abs_path = ctx.platform.find_in_site_packages(rel_dirs, ctx.lib_searched_for, ctx.error_messages, ctx.attachments)
     if abs_path is not None:
-        return FindResult(abs_path, "site-packages")
+        return FindResult(str(abs_path), "site-packages")
```
mdboom:
```suggestion
        return FindResult(abs_path, "site-packages")
```

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_steps.py
```diff
@@ -179,7 +200,7 @@ def find_in_conda(ctx: SearchContext) -> FindResult | None:
     lib_dir = _find_lib_dir_using_anchor(ctx.desc, ctx.platform, anchor)
     abs_path = _find_using_lib_dir(ctx, lib_dir)
     if abs_path is not None:
-        return FindResult(abs_path, "conda")
+        return FindResult(str(abs_path), "conda")
```
mdboom:
```suggestion
        return FindResult(abs_path, "conda")
```

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_steps.py
```diff
@@ -197,10 +218,10 @@ def find_in_cuda_path(ctx: SearchContext) -> FindResult | None:
     cuda_home = get_cuda_path_or_home()
     if cuda_home is None:
         return None
-    lib_dir = _find_lib_dir_using_anchor(ctx.desc, ctx.platform, cuda_home)
+    lib_dir = _find_lib_dir_using_anchor(ctx.desc, ctx.platform, Path(cuda_home))
     abs_path = _find_using_lib_dir(ctx, lib_dir)
     if abs_path is not None:
-        return FindResult(abs_path, "CUDA_PATH")
+        return FindResult(str(abs_path), "CUDA_PATH")
```
mdboom:
```suggestion
        return FindResult(abs_path, "CUDA_PATH")
```

cuda_pathfinder/tests/test_search_steps.py
```diff
@@ -682,9 +682,9 @@ def test_find_lib_dir_uses_descriptor_linux(self, tmp_path):
         (tmp_path / "nvvm" / "lib64").mkdir(parents=True)
 
         desc = _make_desc(name="nvvm", anchor_rel_dirs_linux=("nvvm/lib64",))
-        result = _find_lib_dir_using_anchor(desc, LinuxSearchPlatform(), str(tmp_path))
+        result = _find_lib_dir_using_anchor(desc, LinuxSearchPlatform(), tmp_path)
         assert result is not None
-        assert result.endswith(os.path.join("nvvm", "lib64"))
+        assert str(result).endswith(os.path.join("nvvm", "lib64"))
```
mdboom: Let's also update the uses of `os.path` in test files.

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_platform.py
```diff
@@ -24,13 +22,30 @@
...
+    tracks the newest-first policy that rides on this ordering, so it is kept
+    unchanged on both platforms.
+    """
+    return sorted(directory.glob(pattern), key=str, reverse=reverse)
+
+
+def _sorted_dir_entry_names(directory: Path) -> list[str]:
+    return sorted(entry.name for entry in directory.iterdir())
+
+
```
mdboom: > One thing worth flagging: sorted() over Path compares PurePath._str_normcase, which is lowercased on Windows, so it isn't the same order as sorted() over str for mixed-case names. I kept the existing order by sorting on the string form in a small sorted_glob() helper, rather than change Windows behaviour in a PR that otherwise changes nothing — and I have no Windows machine to check it on. Happy to drop the key for simpler code if you'd prefer.

Yes, let's switch to the simpler code.  I would consider this a fix of a latest Windows bug because it makes sorting behave the same way as it normally does on that platform.

So let's remove these helpers, take the change in Windows sorting behavior as desired (one of the reasons to move to `Path` is to get these sorts of bugfixes from upstream Python), and update all uses of these helpers to the simpler form.

mdboom (changes requested):

LeSingh1:
Done, pushed as 2bda8e8.

`FindResult.abs_path` is a `Path` now and the four `str()` wrappers are gone.

One note on the suggestion itself: `dataclasses.field()` has no `converter=` parameter — that is an `attrs` feature. On 3.13 the signature is `default, default_factory, init, repr, hash, compare, metadata, kw_only`, and `cuda-pathfinder` is `requires-python = ">=3.10"`, so it would not be usable here even if a later CPython adds it. Every construction site already holds a `Path`, so there is nothing to coerce and I annotated the field plainly. Happy to add a `__post_init__` coercion if you would rather it accept `str` too.

The one remaining `str()` is at `LOADER.load_with_abs_path()`, which is `str`-typed all the way down to `ctypes` — widening it would pull `platform_loader.py` and both `load_dl_*` modules into a PR scoped to the search modules.

Both flagged comments are deleted.

`sorted_glob()` and `_sorted_dir_entry_names()` are gone — your anchor covered both — replaced by plain `sorted(d.glob(...))` and `sorted(e.name for e in d.iterdir())`. Taking the Windows ordering change as intended.

`os.path` is out of `test_search_steps.py`, and `test_ctk_root_discovery.py` now compares `Path`s (its `abs_path == str(lib)` assertions would have broken otherwise). `import os` stays for `os.environ`.

On validation: `cuda_pathfinder` cannot import on macOS (`load_dl_linux.py` binds `dlinfo`, absent from Darwin libc). I got a local run by preloading a throwaway shim that stubs `dlinfo`/`dlerror` — nothing committed — and saw `test_search_steps.py` 63 passed, mypy and ruff clean. The full suite shows 86 failures whose list is byte-identical to the same run on the unmodified base, so no regression, but that is macOS with a faked libc symbol. Linux and especially Windows CI are the real gate, and Windows is where the sort-order change lands.

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_steps.py
LeSingh1: `FindResult.abs_path` is now `Path`, and the four `FindResult(...)` call sites drop their `str()` (your other four suggestions).

One deviation from the literal suggestion: `field(converter=...)` is an `attrs` feature, not stdlib. `dataclasses.field` on 3.13 takes only `default, default_factory, init, repr, hash, compare, metadata, kw_only`, and PEP 712 (which proposed `converter`) was rejected — so `abs_path: Path = field(converter=Path)` raises `TypeError: field() got an unexpected keyword argument`. Since every producer already builds a `Path`, I made it a plain `abs_path: Path` annotation rather than add an `attrs` dependency or hand-roll `__post_init__` coercion. Happy to add the coercion if you want it enforced at construction.

`LoadedDL.abs_path` stays `str` — it is public API — so there is now exactly one `str()` at that boundary (`load_nvidia_dynamic_lib.py:161`) instead of a conversion at every producer.

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_platform.py
LeSingh1: Agreed and done — `sorted_glob()` and the sibling sorting helper are gone, all call sites use the plain `Path.glob` / `iterdir` form, and the Windows lowercased-sort-order change is taken as the bugfix you describe.

cuda_pathfinder/tests/test_search_steps.py
LeSingh1: Done. `test_search_steps.py` and `test_ctk_root_discovery.py` are `os.path`-free, and I swept the other two `_dynamic_libs` test files — `test_driver_lib_loading.py` and `test_load_nvidia_dynamic_lib.py` each had one `os.path.isfile`, now `Path(...).is_file()`. Remaining `os.` uses in those files are `os.environ`.

The `os.path` still in `test_find_nvidia_binaries.py`, `test_find_nvidia_headers.py`, `test_find_static_lib.py`, `test_find_bitcode_lib.py` and `test_utils_find_sub_dirs.py` covers `_binaries` / `_headers` / `_static_libs` / `_utils` — other parts of #2410. Say the word if you would rather they all move at once.

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_steps.py
LeSingh1: Both removed (this one and the `find_via_*` return-type note at line 161). With `FindResult.abs_path` now typed `Path` the type commentary is redundant anyway.

mdboom:
@LeSingh1: Can you resolve the merge conflict?  Once that's done, I think this one is good to merge.

LeSingh1:
Rebased onto current main, conflict resolved. It was just colliding imports in `test_search_steps.py` — upstream added `wintypes`, this branch adds `Path`; both are kept and both are used. One of my commits dropped out as its contents had already landed upstream.

`ruff check` and `ruff format --check` are clean on the five changed files. I did not run the suite locally since this is macOS.

mdboom (approved):

mdboom (changes requested):
This has some Windows failures that will need to be addressed before merging.

PR #2490, opened 2026-08-04
cuda.core: resolve default-stream context per call instead of caching it

### LINKED ISSUE #2485: [BUG]: default_stream() singletons permanently cache the first-seen context and device
### Type of Bug

Silent Failure

### Component

cuda.core

### Describe the bug

`default_stream()` returns one of two module-level singletons — [`LEGACY_DEFAULT_STREAM` or `PER_THREAD_DEFAULT_STREAM`](https://github.com/NVIDIA/cuda-python/blob/51141c8531263c0a1fe013b82910eec16a9a2566/cuda_core/cuda/core/_stream.pyx#L422-L423) — created at import time. Both wrap a default-stream token, which means "whatever context is current." Neither carries a context of its own: [`get_stream_context`](https://github.com/NVIDIA/cuda-python/blob/51141c8531263c0a1fe013b82910eec16a9a2566/cuda_core/cuda/core/_cpp/resource_handles.cpp#L713-L727) returns empty for those handles, so [`_from_handle`](https://github.com/NVIDIA/cuda-python/blob/51141c8531263c0a1fe013b82910eec16a9a2566/cuda_core/cuda/core/_stream.pyx#L105-L113) leaves `_h_context` unset.

The first caller that needs a context fills it in. [`Stream_ensure_ctx`](https://github.com/NVIDIA/cuda-python/blob/51141c8531263c0a1fe013b82910eec16a9a2566/cuda_core/cuda/core/_stream.pyx#L450-L461) asks `cuStreamGetCtx`, gets whatever is current at that moment, and stores it on the singleton. [`Stream_ensure_ctx_device`](https://github.com/NVIDIA/cuda-python/blob/51141c8531263c0a1fe013b82910eec16a9a2566/cuda_core/cuda/core/_stream.pyx#L464-L483) then caches `_device_id` the same way. Neither field is ever cleared.

So a process-wide object ends up permanently bound to one context, even though the token it wraps isn't.

That shows up in a few places:

1. **Stale answers after a context switch.** [`Stream.context`](https://github.com/NVIDIA/cuda-python/blob/51141c8531263c0a1fe013b82910eec16a9a2566/cuda_core/cuda/core/_stream.pyx#L359-L361), [`.device`](https://github.com/NVIDIA/cuda-python/blob/51141c8531263c0a1fe013b82910eec16a9a2566/cuda_core/cuda/core/_stream.pyx#L353), [`.resources`](https://github.com/NVIDIA/cuda-python/blob/51141c8531263c0a1fe013b82910eec16a9a2566/cuda_core/cuda/core/_stream.pyx#L371-L373), and [`.record()`](https://github.com/NVIDIA/cuda-python/blob/51141c8531263c0a1fe013b82910eec16a9a2566/cuda_core/cuda/core/_stream.pyx#L281-L282) all read the cached fields. On a multi-GPU machine they keep reporting the first device; `record()` builds the event on the first context, not the current one.

2. **First-touch order decides the answer.** Even [`__repr__`](https://github.com/NVIDIA/cuda-python/blob/51141c8531263c0a1fe013b82910eec16a9a2566/cuda_core/cuda/core/_stream.pyx#L219-L220) populates the cache, so printing the stream in a log or a debugger permanently changes later behavior.

3. **A context stuck for the process lifetime.** The cached `ContextHandle` holds a reference, so a primary context that would otherwise be releasable never is.

4. **First use races.** The write happens under `nogil` on an object shared by every thread. For `PER_THREAD_DEFAULT_STREAM` the caching is wrong even in principle — each thread has its own per-thread default stream and its own current context.

Normal streams are fine. Their context is fixed at creation and `_h_context` comes from the stream box, which is what the existing note on `Stream.device` is talking about.

Spotted while reviewing #2455, which needs a context for a default-stream token and can't take it off the `Stream` for this reason. Reproduced on a local 2-GPU system.

### How to Reproduce

On a host with two GPUs:

```python
from cuda.core import Device, LEGACY_DEFAULT_STREAM

Device(0).set_current()
print(LEGACY_DEFAULT_STREAM.device.device_id)   # 0

Device(1).set_current()
print(Device().device_id)                       # 1
print(LEGACY_DEFAULT_STREAM.device.device_id)   # 0, expected 1
print(LEGACY_DEFAULT_STREAM.context)            # still device 0's context
```

Same idea with two contexts on one device (e.g. a green context made current after the first query).

Order dependence on its own:

```python
from cuda.core import Device, LEGACY_DEFAULT_STREAM

Device(0).set_current()
repr(LEGACY_DEFAULT_STREAM)                     # binds the singleton to device 0

Device(1).set_current()
LEGACY_DEFAULT_STREAM.record()                  # event created on device 0's context
```

First-touch pinning (fresh process):

```python
from cuda.core import Device, LEGACY_DEFAULT_STREAM

Device(1).set_current()
print(LEGACY_DEFAULT_STREAM.device.device_id)   # 1
Device(0).set_current()
print(LEGACY_DEFAULT_STREAM.device.device_id)   # still 1
```

(`Device(...).default_stream` is the same singleton; `default_stream()` itself is not exported from `cuda.core`.)

### Expected behavior

A default-stream token has no context of its own, so queries against the default stream should resolve against whatever is current at call time. `.context`, `.device`, `.resources`, and `.record()` should follow context switches, and a query shouldn't pin a context reference beyond the call.

Two plausible fixes: skip the cache when the handle is `CU_STREAM_LEGACY` / `CU_STREAM_PER_THREAD` and resolve per call, or have `default_stream()` return a fresh `Stream` so per-object caching is sound. The first keeps today's object identity (`__eq__` / `__hash__` already key off the handle, not identity).

### System information

Reproduced on a local 2-GPU system against an installed `cuda.core` built from recent `main`.

aryanputta:
Fixes #2485.

## Problem

`LEGACY_DEFAULT_STREAM` and `PER_THREAD_DEFAULT_STREAM` wrap default-stream tokens, which mean "whatever context is current". Both are module-level singletons created at import time, and neither carries a context of its own: `get_stream_context` returns empty for those handles, so `_from_handle` leaves `_h_context` unset.

`Stream_ensure_ctx` then stored the first context it resolved via `cuStreamGetCtx` onto the singleton, and `Stream_ensure_ctx_device` cached `_device_id` the same way. Neither field was ever cleared, so a process-wide object became permanently bound to one context even though the token it wraps is not.

## Fix

This takes the first of the two options in the issue: skip the cache for default-stream tokens and resolve per call. That keeps today's object identity, so `__eq__` and `__hash__` keying off the handle are unaffected.

`Stream_ensure_ctx` and `Stream_ensure_ctx_device` are replaced by `Stream_get_ctx` and `Stream_get_ctx_device`, which return the context and device through out-parameters and write back to the object only when `Stream_is_default_token` is false. `Stream.context`, `.device`, `.resources`, `.record()`, and `__repr__` now read the returned values rather than the cached fields.

This addresses all four symptoms in the issue: queries follow context switches, first-touch order no longer decides later answers, a query no longer pins a context reference for the lifetime of the process, and the shared singletons are no longer written to under `nogil` from multiple threads.

The predicate keys off the raw handle rather than singleton identity, so a `Stream` built over a foreign default-stream handle behaves the same way.

## Incidental correctness fix

`record()` previously called only `Stream_ensure_ctx_device`, which resolved the context only when `_device_id < 0`. A stream with a device ID already set but no context handle would have passed an empty context to `Event._init`. The new path always resolves both.

## Notes

`create_stream_handle_ref` builds `StreamBox{stream, {}}`, so a default-stream token's box context is always empty. The early return in `Stream_get_ctx` therefore cannot fire for a token, and the driver query is the only path that resolves it.

## Tests

Four tests in `cuda_core/tests/test_stream.py` covering `.device`, `.context`, `.record()`, and `repr()` first-touch behaviour, plus a counterpart asserting that a created stream still keeps the context it was made on.

They need two GPUs and skip otherwise, matching the issue's own reproducer. I looked at using a green context to get a second context on a single device, but no existing test pairs a default stream with a green context and `_stream.pyx` already notes that green contexts reject `CU_STREAM_DEFAULT` for created streams, so I did not want to assert semantics there that the repo has not established.

A release note bullet is added under "Fixes and enhancements" in `1.2.0-notes.rst` since this changes observable behaviour of a stable API.


Andy-Jost (approved):

aryanputta:
Rebased onto current `main` (`dbb2faff`) to pick up #2492.

The docs failure on the previous run was not this diff. All 83 lychee errors were the single 404 at `https://docs.nvidia.com/nvshmem/api/api/language_bindings/python/index.html`; that run finished 19:58 and #2492 repointed the link to `/api/latest/` at 20:51. The new path returns 200. The other red job, `Test win-64 ... (MCDM)`, failed at the `Ensure GPU is working` step rather than in the tests.

This needs a fresh `/ok to test dbb2faff4c` when you have a moment.

PR #2493, opened 2026-08-04
Use pathlib in cuda.pathfinder._static_libs (part 2 of #2410)

LeSingh1:
Follow-up to the review on #2489: the str-compatibility constraint applies only to the public API, so the finder internals now pass and return `Path`. `str()` is applied once, where `abs_path` is stored on `LocatedStaticLib` / `LocatedBitcodeLib`.

The rel-path constants are now forward-slash literals, matching `site_packages_dirs` in the same dicts.

One behaviour change: a `CUDA_PATH` or `CONDA_PREFIX` containing redundant separators now gives a normalized `abs_path`, because `Path` collapses them.

Verified on Linux CI: full pytest output byte-identical to the base commit apart from elapsed time (1276 passed, 4 skipped, 0 errors). Differential fuzzing against the previous revision — 16k lookups over randomized trees, comparing located paths and error text — shows no other difference.

Part 2 of #2410.

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_static_libs/find_bitcode_lib.py
+++ b/cuda_pathfinder/cuda/pathfinder/_static_libs/find_bitcode_lib.py
@@ -36,7 +36,7 @@ class _BitcodeLibInfo(TypedDict):
 _SUPPORTED_BITCODE_LIBS_INFO: dict[str, _BitcodeLibInfo] = {
     "device": {
         "filename": "libdevice.10.bc",
-        "rel_path": str(Path("nvvm", "libdevice")),
+        "rel_path": "nvvm/libdevice",
         "site_packages_dirs": (
             "nvidia/cu13/nvvm/libdevice",
             "nvidia/cuda_nvcc/nvvm/libdevice",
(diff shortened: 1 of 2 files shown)


mdboom (approved):

mdboom:
@LeSingh1: This PR has a non-trivial merge conflict.  Would you mind resolving it?

LeSingh1:
Rebased onto current `main` — conflict resolved in 11916d60.

The conflict was `_SUPPORTED_STATIC_LIBS_INFO["cudadevrt"]`. `main` replaced the static dict with `_cudadevrt_info()`, which resolves the Windows arch at call time via `windows_python_arch()` and adds the non-x64 fallbacks. I kept that version and applied only this PR's conversion to it:

```python
"ctk_rel_paths": (str(Path("lib", arch_dir)),),
"conda_rel_paths": (str(Path("lib", arch_dir)), *conda_fallback_dirs),
```

so the arch-aware behaviour from `main` is preserved and the `os.path.join` calls inside it are gone. `import os` stays — `os.environ` is still used further down.

One thing I should flag rather than let you assume otherwise: I could not run the test suite locally to confirm. `cuda_pathfinder` cannot be imported on macOS at all — `load_dl_linux.py` binds `dlinfo`, which does not exist in libc here, so collection fails before any test runs. I verified the resolution by reading it and with `py_compile`; CI is the real check.

Happy to adjust if you would rather the helper stay on `os.path.join` and this PR skip that function.

PR #2494, opened 2026-08-04
Use pathlib in cuda.pathfinder._binaries and _utils (part 3 of #2410)

LeSingh1:
`find_nvidia_binary_utility` works in `Path` internally; its public return is unchanged. `SITE_PACKAGES_BINDIRS` holds path components instead of joined strings, since the caller immediately split them again.

`find_sub_dirs_no_cache` tests directories through a small `_is_dir()` helper rather than `Path.is_dir()` directly. The two are not interchangeable: `os.path.isdir()` returns `False` for any stat error, while `Path.is_dir()` propagates the ones outside pathlib's ignore list (`EACCES`, `ENAMETOOLONG`). This walks site-packages trees we do not control, so one unreadable directory would otherwise turn a clean "not found" into a `PermissionError`. Raised separately on #2410, since it applies beyond this chunk.

`test_find_nvidia_binaries.py` is included here rather than in part 4, because it asserts on the exact values these private helpers receive and return.

Verified on Linux CI: full pytest output byte-identical to the base commit apart from elapsed time. Fuzzed `find_sub_dirs_no_cache` against the old implementation — 3000 calls comparing result order, plus 720 over trees containing unreadable directories.

Part 3 of #2410.

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_utils/find_sub_dirs.py
+++ b/cuda_pathfinder/cuda/pathfinder/_utils/find_sub_dirs.py
@@ -8,20 +8,6 @@ from collections.abc import Sequence
 from pathlib import Path
 
 
-def _is_dir(path: Path) -> bool:
-    """``path.is_dir()``, but False instead of raising on an inaccessible path.
-
-    This walks directories nobody here controls, so it has to tolerate whatever
-    it runs into. Path.is_dir() only swallows the errnos in pathlib's ignore
-    list, and raises for the rest (EACCES, ENAMETOOLONG); os.path.isdir, which
-    this replaces, returned False for all of them.
-    """
-    try:
-        return path.is_dir()
-    except OSError:
-        return False
-
-
 def find_sub_dirs_no_cache(parent_dirs: Sequence[str], sub_dirs: Sequence[str]) -> list[str]:
     # Results stay str: they are consumed by _binaries, _dynamic_libs, _headers
     # and _static_libs, so the type flip belongs in its own change.
@@ -31,7 +17,7 @@ def find_sub_dirs_no_cache(parent_dirs: Sequence[str], sub_dirs: Sequence[str]) 
         while stack:
             current_path, idx = stack.pop()
             if idx == len(sub_dirs):
-                if _is_dir(current_path):
+                if current_path.is_dir():
                     results.append(str(current_path))
                 continue
 
@@ -42,11 +28,11 @@ def find_sub_dirs_no_cache(parent_dirs: Sequence[str], sub_dirs: Sequence[str]) 
                 except OSError:
                     continue
                 for entry_path in entries:
-                    if _is_dir(entry_path):
+                    if entry_path.is_dir():
                         stack.append((entry_path, idx + 1))
             else:
                 next_path = current_path / sub
-                if _is_dir(next_path):
+                if next_path.is_dir():
                     stack.append((next_path, idx + 1))
     return results
 
--- a/cuda_pathfinder/cuda/pathfinder/_utils/find_sub_dirs.py
+++ b/cuda_pathfinder/cuda/pathfinder/_utils/find_sub_dirs.py
@@ -2,36 +2,51 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import functools
-import os
 import site
 import sys
 from collections.abc import Sequence
+from pathlib import Path
+
+
+def _is_dir(path: Path) -> bool:
+    """``path.is_dir()``, but False instead of raising on an inaccessible path.
+
+    This walks directories nobody here controls, so it has to tolerate whatever
+    it runs into. Path.is_dir() only swallows the errnos in pathlib's ignore
+    list, and raises for the rest (EACCES, ENAMETOOLONG); os.path.isdir, which
+    this replaces, returned False for all of them.
+    """
+    try:
+        return path.is_dir()
+    except OSError:
+        return False
 
 
 def find_sub_dirs_no_cache(parent_dirs: Sequence[str], sub_dirs: Sequence[str]) -> list[str]:
+    # Results stay str: they are consumed by _binaries, _dynamic_libs, _headers
+    # and _static_libs, so the type flip belongs in its own change.
     results = []
     for base in parent_dirs:
-        stack = [(base, 0)]  # (current_path, index into sub_dirs)
+        stack = [(Path(base), 0)]  # (current_path, index into sub_dirs)
         while stack:
             current_path, idx = stack.pop()
             if idx == len(sub_dirs):
-                if os.path.isdir(current_path):
-                    results.append(current_path)
+                if _is_dir(current_path):
+                    results.append(str(current_path))
                 continue
 
             sub = sub_dirs[idx]
             if sub == "*":
                 try:
-                    entries = sorted(os.listdir(current_path))
+                    entries = sorted(current_path.iterdir(), key=lambda entry: entry.name)
                 except OSError:
                     continue
-                for entry in entries:
-                    entry_path = os.path.join(current_path, entry)
-                    if os.path.isdir(entry_path):
+                for entry_path in entries:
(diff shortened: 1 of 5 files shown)


cuda_pathfinder/cuda/pathfinder/_utils/find_sub_dirs.py
```diff
@@ -2,36 +2,51 @@
...
 
 import functools
-import os
 import site
 import sys
 from collections.abc import Sequence
+from pathlib import Path
+
+
+def _is_dir(path: Path) -> bool:
```
mdboom: As discussed in #2410, let's not worry about this behavioral change.  No need for this helper function -- just use `is_dir()` directly.

mdboom (changes requested):
Looks good, other than the `_is_dir` helper not being needed.

LeSingh1:
Done — `_is_dir` is gone and the three call sites use `Path.is_dir()` directly. Understood on the behavioural change; I dropped the note arguing for it rather than re-litigate.

`cuda_pathfinder` can't be imported on my macOS box — `load_dl_linux.py` binds `dlinfo`, which is absent from macOS libc, so collection dies before any test runs. I worked around it by registering stub `cuda` / `cuda.pathfinder` / `cuda.pathfinder._utils` module objects with `__path__` pointing at the real directories, which loads `find_sub_dirs.py` from disk without executing the package `__init__`. `tests/test_utils_find_sub_dirs.py` then ran unmodified: 11 passed, including the three site-packages ordering tests. ruff check and ruff format clean.

That covers the file you commented on; CI is the real validation for the rest.

cuda_pathfinder/cuda/pathfinder/_utils/find_sub_dirs.py
LeSingh1: Done in 82f7dd3 — `_is_dir` is gone and the two call sites use `Path.is_dir()` directly. That commit predates my noticing your review, so this was already pushed; flagging it so the thread is not left hanging.

LeSingh1:
The `_is_dir` ask is addressed (reply inline).

One thing I left alone, flagging rather than deciding: `test_find_nvidia_binaries.py` still has four `os.path.abspath` calls in assertions (plus one in `test_utils_find_sub_dirs.py`). You asked on #2489 to move test files off `os.path` too, and I did that there — but these four are assertions specifically *about* absolute-path semantics, and `os.path.abspath` has no exact pathlib equivalent: `Path.resolve()` also resolves symlinks (which would change what these assert under a symlinked temp dir or conda prefix) and `Path.absolute()` does not normalise `..`. I have no Linux or conda environment to check the difference on, so I would rather not guess. Happy to convert them to `str(Path(...).absolute())` if you are confident that is the intent.

mdboom:
>  Happy to convert them to str(Path(...).absolute()) if you are confident that is the intent.

Yes, please do.

LeSingh1:
Done — the three `os.path.abspath` assertions in `test_find_nvidia_binaries.py` are now `str(...absolute())`, and I took the `os.path.isdir` in `test_utils_find_sub_dirs.py` with them so that file is `os.path`-free. `import os` stays in the binaries test because `os.sep` is still used to build the platform-neutral fake roots.

Ran both test files on macOS with a throwaway libc shim (nothing committed): same 16 pre-existing failures as the base commit, all of them the real-toolkit `test_find_binary_utilities` cases, and all three assertions I touched pass. `ruff check` and `ruff format --check` are clean on both files.

mdboom:
> Ran both test files on macOS with a throwaway libc shim

cuda-python doesn't support macOS so no need to test there.

mdboom:
@LeSingh1: Please resolve the merge conflicts, then I think this one is good-to-go.

LeSingh1:
Rebased onto current main. The conflict was more than textual, so worth spelling out what I did.

Upstream added Windows compute-sanitizer handling and moved the CUDA Toolkit search out of the early dirs-collection block, both after this branch was cut. I kept your structure in both places and re-applied the pathlib migration on top, rather than reinstating the old ordering.

That left the file mixing `str` and `Path` across the helper boundaries, which mypy caught: `_resolve_names_in_trusted_dirs` was still `list[str]` while `dirs` had become `list[Path]`, and `_resolve_candidate_paths` was passing a `str` to a now-`Path` predicate. `_resolve_in_trusted_dirs` now returns `str` so it matches the `str | None` the public function returns, which also removed three conversions at the call sites.

mypy and ruff are clean on all five files. Not run locally beyond that, since macOS is unsupported here.

mdboom (approved):

mdboom (changes requested):
Looks like there are some test failures we need to fix.

LeSingh1:
Thanks — I looked at the failures and there are two causes, both mine from the rebase. One is fixed in 7417997534, the other is not yet, so this still needs work.

Fixed: I had changed `_resolve_in_trusted_dirs` to return `str` to satisfy mypy, but this branch's tests assert it returns a `Path`. It returns `Path` again and the three call sites convert at the public boundary instead. That accounts for the `TestResolveInTrustedDirs` failures.

Not fixed: the remaining failures, such as `test_find_binary_search_path_includes_site_packages_conda_cuda`, expect `cuda_home / "bin"` in the early collected dirs. You moved the CUDA Toolkit search out of that block and into step 3.2 after this branch was cut, and I took your version during the rebase. So the branch's tests still encode the old ordering and need rewriting against the new one. That is a bigger change than a conflict fix and I would rather do it carefully than guess at it — I will push it separately.

Apologies for the noise on an already-approved PR.

mdboom:
Looks like we have merge conflicts again.  Please fix those, and I'll run the tests again.  This is getting close, though.

LeSingh1:
Rebased onto current main in 2d5b704065. The conflict this time was only the import block.

While tracing why the earlier tests failed I found the real second cause, and it is a runtime bug rather than an ordering problem as I guessed last time. At step 3.2, `cuda_path` comes from `get_cuda_path_or_home()`, which returns `str`, and it was being passed straight into `_ctk_bin_subdirs(root: Path)` — so `root / "bin"` would raise `TypeError`. The canary branch below it already wrapped in `Path(...)`; this one did not. Now both do.

Worth flagging that mypy did not catch this, so the earlier green typecheck I reported was not the assurance I treated it as. I still cannot run the suite here — `cuda.pathfinder` will not import on macOS — so CI is the real check.

LeSingh1:
Diagnosed and fixed the Windows failures from the last run.

They weren't platform-specific. `_is_executable_candidate` now takes a `Path`, but the tests that drive it through `_patch_exec_probe` still built their expected paths with `os.path.join`. A `str` never compares equal to a `Path`, so the fake probe reported every candidate as missing and each lookup returned `None`. It only showed up on Windows because the Linux test jobs were skipped in that run — the same 14 tests would have failed there.

Converted the remaining tests to `Path`, and added an assertion in `_patch_exec_probe` that `existing` holds `Path`s so the same mismatch fails loudly instead of silently returning `None`.

Verified by running the file against the previous commit and against this one (with a local stub for `load_dl_linux`, since macOS has no `dlinfo`): 30 failed / 19 passed before, 16 failed / 33 passed after. The 16 that remain are the subprocess-based `test_find_binary_utilities` and `test_caching_per_utility`, which fail identically on both commits because the child process doesn't get the stub — they pass in your CI.

Merge conflicts are resolved and the PR shows mergeable again. Ready for another run when you are.

mdboom:
Thanks.  Unfortunately merge conflicts again, if you could handle.  (Changes like this that are so broad tend to have merge conflicts).

LeSingh1:
Rebased onto current main. One conflict, in `supported_nvidia_binaries.py` — you reworded the comment to say "in search order" while this branch changed the values from `os.path.join` strings to path-component tuples. Kept both.

`ruff check` is clean and the pathfinder tests are unchanged from before the rebase (the failures I see locally are the subprocess-based ones that need `dlinfo`, which macOS doesn't have).

Ready for a run whenever you are.

PR #2496, opened 2026-08-04
Use pathlib in toolshed and ci helper scripts (part 7 of #2410)

LeSingh1:
Path joining and filesystem predicates in the helper scripts, plus `glob.glob` → `Path.glob` in `dump_cutile_b64.py`.

`os.path.isfile` stays in `check_generated_file_seals.py`: that guard skips anything that is not a readable regular file, and `Path.is_file()` raises on `EACCES`/`ENAMETOOLONG` where `os.path.isfile` returns `False`. `os.path.abspath` stays where `sys.path` needs a `str`.

Not covered by the `cuda_pathfinder` CI job, so I verified locally instead: `ci/tools/tests/test_check_release_notes.py` passes (42 tests), and both pre-commit hooks produce output identical to the pre-change scripts when run over every tracked `.py` file.

Part 7 of #2410.

Parts 5 and 6 (`cuda_core` and `cuda_bindings`, largely `build_hooks.py`) are not ready — I have no way to run a wheel build here, and given the `Path.is_file()` divergence noted on #2410 I would rather not ship unverifiable build-system path changes.


ci/tools/check_release_notes.py
```diff
@@ -85,7 +86,7 @@ def is_backport_version(version: str, backport_branch: str) -> bool:
 
 
 def notes_path(package: str, version: str) -> str:
-    return os.path.join(package, "docs", "source", "release", f"{version}-notes.rst")
+    return str(Path(package, "docs", "source", "release", f"{version}-notes.rst"))
```
mdboom: We can treat all functions in these `toolshed` helper scripts as private.  (They are generally not imported from third-party code).  So we can change the return type on this to `Path` and prevent more conversions back and forth.

toolshed/check_generated_file_seals.py
```diff
@@ -142,6 +142,8 @@ def main(args):
 
     returncode = 0
     for filepath in args:
+        # os.path.isfile, not Path.is_file: this skips anything that is not a
+        # readable regular file, and Path.is_file raises on e.g. EACCES.
         if not os.path.isfile(filepath):
```
mdboom: As mentioned elsewhere, let's not worry about this behavioral difference.  Update to `Path.is_file`

mdboom (changes requested):

LeSingh1:
Both addressed.

`notes_path` returns `Path` now, so `check_release_notes` does `Path(repo_root) / path` instead of the `Path(repo_root, str(Path(...)))` round-trip you were pointing at. The `Path` then flows unconverted into the problem tuples and out through the f-strings, so I widened three annotations from `list[tuple[str, str]]` to `list[tuple[str | Path, str]]` — that list also carries literal `<tag>` / `<component>` / `<backport-git-tag>` placeholders, so the union is honest rather than a workaround.

`check_generated_file_seals.main` uses `Path.is_file()`, the comment defending `os.path.isfile` is gone, and `import os` went with it — that was the file's last `os` use.

Left alone deliberately: `validate_generated_file_seal` still takes the raw string, because it interpolates `{filepath!r}` in about ten user-facing messages and a `Path` would print `PosixPath('foo.pyx')` into pre-commit output — that would add ten `str()` calls, not remove conversions. `repo_root: str` stays because argparse hands it over as a string.

`ci/tools/tests/test_check_release_notes.py` passes (42), including the `::warning file=...` assertion that pins the rendered path, so it renders byte-identically. I also ran both scripts by hand against real repo files.

ci/tools/check_release_notes.py
LeSingh1: Done. `repo_root` is now a `Path` end to end — `load_backport_branch`, `check_release_notes` and `validate_backport_decision` all take `Path`, and `--repo-root` parses with `type=Path`. That removed the `Path(repo_root)` re-wrap inside the functions and 19 `str(tmp_path)` conversions in the tests. The five `main()` argv lists keep `str()` since those are command-line strings, which argparse converts back to `Path`.

toolshed/check_generated_file_seals.py
LeSingh1: Done in 4385757 (pushed before I saw the review, flagging so the thread is not left hanging): the checker uses `Path(filepath).is_file()` and the comment about the `os.path.isfile` behavioural difference is gone. Also swept the rest of the PR for the same pattern — `toolshed/build_static_bitcode_input.py` still had one `os.path.abspath`, now `Path.resolve()`.

mdboom (approved):

PR #2501, opened 2026-08-04
Use pathlib in cuda.bindings build hooks and tests (part 6 of #2410)

LeSingh1:
Part 6 of #2410, and the last of the series.

Replaces `os.path`/`os.sep` with `pathlib.Path` in `cuda_bindings/build_hooks.py` and in the test modules that build paths to test data, the examples directory, and a reproducer's cwd.

Values handed to setuptools' `Extension` (`include_dirs`, `library_dirs`) and to the cuFile string parameters stay `str`; only the path construction moves to pathlib.

As in part 5, `glob.glob` is left alone and only the pattern construction moved, and I diffed the old and new extension-discovery logic against the real source tree on Linux — identical across all 39 `cuda.bindings` module names and basenames, and identical include/library dir lists.

One cosmetic note: the `test_examples.py` parametrize IDs change from `.../tests/../examples/foo.py` to `.../examples/foo.py`. Same file set — I grepped `ci/`, `.github/`, `scripts/` and `toolshed/` and nothing selects these tests by ID. Happy to preserve the literal `..` if you would rather keep the IDs stable.

These files all need a GPU or a built `cuda.bindings`, so the changes are verified by inspection plus `ruff check`, `ruff format --check` and `py_compile` rather than by a test run; `cuda_core/tests/test_build_hooks.py` still passes 14/14 on Linux CI.

NOTE: developed with the assistance of an AI coding agent. I reviewed and verified the change before submitting.

### DIFF
--- a/cuda_bindings/build_hooks.py
+++ b/cuda_bindings/build_hooks.py
@@ -85,11 +85,11 @@ def _get_cuda_path() -> str:
 
 
 def _rename_architecture_specific_files():
-    path = os.path.join("cuda", "bindings", "_internal")
+    path = Path("cuda", "bindings", "_internal")
     if sys.platform == "linux":
-        src_files = glob.glob(os.path.join(path, "*_linux.pyx"))
+        src_files = path.glob("*_linux.pyx")
     elif sys.platform == "win32":
-        src_files = glob.glob(os.path.join(path, "*_windows.pyx"))
+        src_files = path.glob("*_windows.pyx")
     else:
         raise RuntimeError(f"platform is unrecognized: {sys.platform}")
     dst_files = []
@@ -97,19 +97,22 @@ def _rename_architecture_specific_files():
         with tempfile.NamedTemporaryFile(delete=False, dir=".") as f:
             shutil.copy2(src, f.name)
             f_name = f.name
-        dst = src.replace("_linux", "").replace("_windows", "")
-        os.replace(f_name, f"./{dst}")
+        dst = src.with_name(src.name.replace("_linux", "").replace("_windows", ""))
+        os.replace(f_name, dst)
         dst_files.append(dst)
     return dst_files
 
 
 def _prep_extensions(sources, libraries, include_dirs, library_dirs, extra_compile_args, extra_link_args):
-    pattern = sources[0]
+    # sources[0] is a glob pattern (or a concrete file), not a directory, so
+    # glob.glob() is still the right tool; str() covers the Path entries that
+    # _rename_architecture_specific_files() produces.
+    pattern = str(sources[0])
     files = glob.glob(pattern)
     libraries = libraries if libraries else []
     exts = []
     for pyx in files:
-        mod_name = pyx.replace(".pyx", "").replace(os.sep, ".").replace("/", ".")
+        mod_name = ".".join(Path(pyx).with_suffix("").parts)
         exts.append(
             Extension(
                 mod_name,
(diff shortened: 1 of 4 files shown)


cuda_bindings/tests/test_examples.py
```diff
@@ -5,12 +5,13 @@
...
 import sys
+from pathlib import Path
 
 import pytest
 from cuda_python_test_helpers.pep723 import has_package_requirements_or_skip
 
-examples_path = os.path.join(os.path.dirname(__file__), "..", "examples")
-examples_files = glob.glob(os.path.join(examples_path, "**/*.py"), recursive=True)
+examples_path = Path(__file__).parents[1] / "examples"
+examples_files = glob.glob(str(examples_path / "**" / "*.py"), recursive=True)
```
mdboom: Why no `Path.glob` here?

cuda_bindings/build_hooks.py
```diff
@@ -79,11 +80,11 @@ def _get_cuda_path() -> str:
 
 
 def _rename_architecture_specific_files():
-    path = os.path.join("cuda", "bindings", "_internal")
+    path = Path("cuda", "bindings", "_internal")
     if sys.platform == "linux":
-        src_files = glob.glob(os.path.join(path, "*_linux.pyx"))
+        src_files = glob.glob(str(path / "*_linux.pyx"))
```
mdboom: Ditto

cuda_bindings/build_hooks.py
```diff
@@ -79,11 +80,11 @@ def _get_cuda_path() -> str:
...
     elif sys.platform == "win32":
-        src_files = glob.glob(os.path.join(path, "*_windows.pyx"))
+        src_files = glob.glob(str(path / "*_windows.pyx"))
```
mdboom: Ditto

cuda_bindings/build_hooks.py
```diff
@@ -103,7 +104,7 @@ def _prep_extensions(sources, libraries, include_dirs, library_dirs, extra_compi
     libraries = libraries if libraries else []
     exts = []
     for pyx in files:
-        mod_name = pyx.replace(".pyx", "").replace(os.sep, ".").replace("/", ".")
+        mod_name = ".".join(Path(pyx).with_suffix("").parts)
```
mdboom, line 114: 👍

mdboom (changes requested):

LeSingh1:
Switched both spots to `Path.glob`.

`_rename_architecture_specific_files` now yields `Path`, so three consumers moved with it: `src.replace("_linux", "")` became `src.with_name(src.name.replace(...))` — `Path.replace` is `os.replace`, so leaving that would have been a silent footgun that renames files on disk — and `f.endswith(".pyx")` / `Path(f).name` became `f.suffix` / `f.name`.

I left `_prep_extensions`' `glob.glob` and the `cuda/bindings/*.pyx` listing alone. Those take a glob *pattern* rather than a directory, and their output goes straight into `setuptools.Extension(sources=...)` and `cythonize`, which do raw string operations — Cython's `extended_iglob` in particular.

Two things I checked rather than assumed:

`Path.glob` matches dotfiles and hidden directories where `glob.glob` does not. Both example trees produce identical sets today (14 files here, 20 in cuda_core), and I diffed the `_rename_architecture_specific_files` rewrite against a synthetic `_internal/` to confirm byte-identical `dst_files` and resulting tree.

With bare `Path` values pytest degrades the parametrize IDs to `example0..13`, so I added `ids=str` and confirmed via `--collect-only` that IDs and collection order are byte-identical to the old string version.

Not run locally: the cuda_bindings suite and a real wheel build need CUDA and a toolkit. py_compile, ruff and ruff format are clean; CI is the gate.

LeSingh1:
All three `Path.glob` comments addressed (replies inline).

One `glob.glob` remains, in `_prep_extensions`. It is not a directory search: `sources[0]` is a pattern that may be rooted anywhere (`"cuda/bindings/utils/*.pyx"`) or a concrete file, so there is no `Path.glob` form for it — `Path.glob` needs the directory split out from the pattern. I trimmed the comment there to say just that. The `glob.glob("cuda/bindings/*.pyx")` call further down is untouched by this PR, so I left it rather than widen the diff; happy to convert both in a follow-up if you would like the module free of `glob`.

cuda_bindings/tests/test_examples.py
LeSingh1: Switched to `examples_path.glob("**/*.py")` in 29320f21 (pushed before I saw the review). Verified it collects the same 14 files as the old `glob.glob(..., recursive=True)` call, and the `ids=str` parametrize renders the same test IDs.

cuda_bindings/build_hooks.py
LeSingh1: Both are `path.glob("*_linux.pyx")` / `path.glob("*_windows.pyx")` as of 29320f21. `_rename_architecture_specific_files` now works in `Path` throughout, so `dst` is built with `with_name()` instead of string replacement on the full path.

cuda_bindings/build_hooks.py
```diff
@@ -85,31 +85,33 @@ def _get_cuda_path() -> str:
...
         dst_files.append(dst)
     return dst_files
 
 
 def _prep_extensions(sources, libraries, include_dirs, library_dirs, extra_compile_args, extra_link_args):
-    pattern = sources[0]
+    # sources[0] is a pattern rooted anywhere, not a directory to search, so
+    # there is no Path.glob() form for it.
+    pattern = str(sources[0])
     files = glob.glob(pattern)
```
mdboom, line 110: I don't understand this comment.  Would `Path().glob(pattern)` not work?

mdboom (changes requested):

PR #2509, opened 2026-08-05
build: report internal build dependency provenance

rwgk:
## Description

Related to #2468. The stacked follow-up, #2510, addresses dependency selection itself.

PEP 517 build hooks execute in environments where the dependency actually imported at build time is not always obvious from the surrounding workflow. A locally built wheel, an editable or source installation, and a released package can all be present or eligible. Without explicit provenance in the build log, a successful build can appear to validate one combination of sources while actually using another, and discrepancies between local and CI builds can be unnecessarily difficult to diagnose.

This PR reports the internal dependency that each build hook imported, at the point where it is used:

- The `cuda.bindings` build reports the `cuda-pathfinder` version and package directory.
- The `cuda.core` build reports the `cuda-pathfinder` version and package directory.
- The `cuda.core` build reports the `cuda-bindings` version and package directory.

The resulting messages have this form:

```text
Using cuda-pathfinder <version> from <package directory>
Using cuda-bindings <version> from <package directory>
```

Both pieces of information are useful. A version alone may not distinguish a local artifact from an index-provided installation, while a path alone does not identify which release or development version was imported. Together they make the build log direct, self-contained evidence of dependency provenance.

The messages are written to stderr so they remain visible in build logs. This PR does not change dependency resolution or validation behavior. While touching the duplicated `cuda-pathfinder` import helpers, it also keeps them equivalent and narrowly replaces their `os.path` operations with `pathlib.Path` operations.

### DIFF
--- a/cuda_bindings/build_hooks.py
+++ b/cuda_bindings/build_hooks.py
@@ -16,6 +16,7 @@ import shutil
 import sys
 import sysconfig
 import tempfile
+from pathlib import Path
 from warnings import warn
 
 from setuptools import build_meta as _build_meta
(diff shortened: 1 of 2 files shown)


juenglin (approved):

rwgk:
Thanks @juenglin!

PR #2512, opened 2026-08-05
Fix nvbug6550424: Don't check NVML init behavior on CTK >= 13.4

mdboom:
Fundamentally, this just skips a test that no longer makes sense on NVML >= 13.4 since the init behavior was changed upstream.

This also refactors the code to move `driver_version_less_than` to a shared helper in `cuda_python_test_helpers`.

### DIFF
--- a/cuda_bindings/tests/nvml/test_init.py
+++ b/cuda_bindings/tests/nvml/test_init.py
@@ -7,6 +7,7 @@ import warnings
 import pytest
 
 from cuda.bindings import nvml
+from cuda_python_test_helpers import driver_version_less_than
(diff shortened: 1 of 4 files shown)


rwgk (approved):

rwgk:
Is the

[CI / Build win-64, CUDA 13.3.0 / py3.15 (push)](https://github.com/NVIDIA/cuda-python/actions/runs/31004419739/job/92314770231?pr=2512)

failure merge-blocking now?

It thought I saw an intermittent numpy version mixup before, but I see you reran once already.

PR #1034, opened 2025-09-26
test(pathfinder): preserve logging info but require opt-in behavior

cpcloud:
This PR attempts to allow developers to opt **in** to the info level logging
when running `cuda_pathfinder` tests, instead of the logging always showing up
in a summary, including when terminating tests early using ^C.

Now, if you want the extra info you have to pass `--log-cli-level=INFO`.

The main difference now is that the logging will be **live** instead of in a summary.

I would argue this is more useful when you've opted into the logging, since the
logs will appear immediately below the test name and thus can be easily associated with
the logging invocation.

Finally, this also requires less fixture hacking and less pytest configuration.

Issue #650, opened 2025-05-20
[FEA]: Add support for user-opt-in logging in path finder


leofang:
Examples:
- https://github.com/cupy/cupy/issues/9127 (It contains logs from CuPy's internal search logic for library loading)
- nvmath-python ([docs](https://docs.nvidia.com/cuda/nvmath-python/latest/overview.html#full-logging-support))

u7k4rs6:
Hi @leofang @rparolin, I'd like to take this one. It's unassigned with no branch or PR, and I have a CUDA dev setup on Ubuntu to test against. Since it was moved out of the pathfinder backlog, could you confirm you'd still accept an external PR before I invest? Assuming yes, here's the approach I'd propose so we can lock the interface up front:

- Standard-library `logging` with a dedicated `cuda.pathfinder` logger, off by default since it sits on the import hot path.
- Opt-in via an env var, e.g. `CUDA_PATHFINDER_LOG_LEVEL` mapped to standard logging levels (happy to match whatever naming convention you prefer), plus optionally a small programmatic setter.
- On each lookup, emit a search trail: candidate paths and their order, env vars consulted (CUDA_HOME, LD_LIBRARY_PATH, conda/pip locations), the resolved path and version actually loaded, fallbacks taken, and on failure the full list of paths tried. This mirrors the CuPy (cupy#9127) and nvmath-python logging you linked.
- Tests via pytest `caplog`, most of which run without a GPU.

Does the interface (logger name, env var, optional API toggle) look right? If so, I'll scope the PR to cuda.pathfinder. Thanks!

rparolin:
@rwgk Can you comment on the proposed direction above? Thanks!

u7k4rs6:
@rwgk @rparolin friendly ping on the above. Rather than keep the interface question open, I'm happy to just open a draft PR against what I sketched (dedicated cuda.pathfinder logger, CUDA_PATHFINDER_LOG_LEVEL, off by default) so there's something concrete to react to, and adjust naming from review. Planning to start on that this week unless you'd rather settle the interface first.

u7k4rs6:
Opened a draft PR implementing the interface above, scoped to load_nvidia_dynamic_lib. Happy to extend to the other four finder families, or change the naming and levels, whatever suits.

u7k4rs6:
[...]

PR #2513, opened 2026-08-05
feat(pathfinder): opt-in diagnostic logging for load_nvidia_dynamic_lib

u7k4rs6:
## Description

Implements the interface proposed in #650. Opening as a draft: the direction was
proposed on 4 Jun and pinged since without a reply, so this is a concrete shape to
react to rather than a request to merge. Naming (logger name, env var) and level
choices are open to whatever the maintainers prefer.

Scoped to `load_nvidia_dynamic_lib` only. The same pattern extends to the other
four finder families -- `find_nvidia_binary_utility`, `find_nvidia_header_directory`,
`find_static_lib`, `find_bitcode_lib` -- and I'll extend it if you like the shape.

### What

- `cuda.pathfinder` logger, disabled by default, enabled via
  `CUDA_PATHFINDER_LOG_LEVEL` (level name or integer).
- Each find step logs whether it matched; a successful load logs the resolved
  path and `found_via`; the failure path logs the accumulated candidate list.
- Records carry `pathfinder_*` fields (`pathfinder_libname`, `pathfinder_found_via`,
  `pathfinder_abs_path`, `pathfinder_step`, `pathfinder_matched`, ...) so consumers
  can filter structurally instead of regexing messages. `found_via` already existed
  on `FindResult` and `LoadedDL`; this surfaces it rather than inventing a field.
- NullHandler only. No `basicConfig`, no root-logger configuration.
- Invalid env values warn once and leave logging disabled.

### On off-by-default

@rwgk argued against opt-in diagnostics in #1034, and that concern is worth
answering directly rather than leaving implicit.

`SearchContext.raise_not_found()` already formats the full candidate list into
`DynamicLibNotFoundError`, always on. The case being protected there -- someone
hits a failure and then needs a second run with a flag to diagnose it -- is
already covered today, and this PR does not change it. What is off by default is
success-path telemetry, where by definition nothing needs diagnosing. Because
`logging` is imported lazily, the disabled path costs nothing at all, so
off-by-default is not a tradeoff being made against visibility.

There is one real gap, and I don't want to gloss it: a load that succeeds but
resolves the wrong library. `found_via` being off by default does cost something
there.

I looked at whether `warnings.warn` -- already the house idiom, per
`env_vars.py:96` and `load_dl_windows.py:90` -- would fit a genuinely suspicious
success better than a log line. I don't think it does. Both existing uses signal
an actionable anomaly, and neither shape applies:

- The cascade returns the first hit and never enumerates alternatives. Detecting
  that a conda copy shadowed a site-packages copy would mean continuing the search
  after a hit -- real filesystem work on every successful load.
- The one suspicious-success signal already computed for free is
  `was_already_loaded_from_elsewhere`. Warning on it would be wrong: it is the
  normal case whenever two NVIDIA libraries share a dependency, so it would fire
  constantly on correct usage.

So it stays a log field. If you'd rather have an always-on warning for some
narrower condition, I'm happy to add it -- I just couldn't find one that is both
cheap to detect and reliably suspicious.

### Import cost

`import logging` pulls in seven modules (`logging`, `atexit`, `string`, `_string`,
`textwrap`, `traceback`, `_colorize`). Measured on this branch:

| | logging in `sys.modules` after `import cuda.pathfinder` |
|---|---|
| disabled | `False` |
| enabled  | `True`  |

Per call site, disabled: 29.7 ns vs 24.7 ns baseline over 500k iterations.

### Provenance

Written with AI assistance. The new tests carry
`@pytest.mark.agent_authored(model="claude-opus-5")`, the convention documented
in `AGENTS.md`; flagging it here too since that marker only covers tests, and
this PR also adds source under `cuda/pathfinder/_utils/`.

## Checklist

Docs box unticked deliberately: nothing user-facing is documented yet, and where
it belongs (`docs/source/`, the env-var list, or under `load_nvidia_dynamic_lib`)
is worth a steer first.

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py
+++ b/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py
@@ -35,6 +35,7 @@ from cuda.pathfinder._dynamic_libs.subprocess_protocol import (
     parse_dynamic_lib_subprocess_payload,
 )
 from cuda.pathfinder._dynamic_libs.supported_nvidia_libs import ALL_AVAILABLE_LIBNAMES
+from cuda.pathfinder._utils.diagnostic_log import LOGGER, search_extra
 from cuda.pathfinder._utils.platform_aware import IS_WINDOWS
 
 if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


rparolin:
@rwgk Please review this.

u7k4rs6:
@rparolin Thanks. Also flagging that copy-pr-bot is still holding workflows pending vetting, so nothing has run on this yet. Happy to un-draft once CI can report.

u7k4rs6:
@rparolin  The /ok to test came back with E1. That doc page is behind SSO so I can't read it, but the PR was still a draft when you ran it and it's out of draft now, so a retry may go through: /ok to test fbfe6de66b522446ed5c679840bc73a7fa74ee57.

Also, un-drafting made pr-metadata-check run for real and it's failing on the type label. I have cuda.pathfinder but none of feature/enhancement/etc., and I can't set labels. feature would match #650.

PR #2514, opened 2026-08-05
Add check and agent guidance about uncapped mempools

juenglin:
Follow-up to #2484, which right-sized seven memory pools that reserved an
address-space window scaling with device memory while holding at most 1 KiB.
This PR adds the guidance and the automated check to keep new pools from
reintroducing the problem.

`cuda_core/tests/AGENTS.md` (new) documents the rule: a pool created without
`max_size` reserves roughly 1x device memory of address space regardless of
what the test allocates, and the whole suite shares one process, so those
reservations accumulate. It also spells out a distinction that is easy to get
backwards: `DeviceMemoryResource(dev)` with no options does **not** create a
pool — it wraps the device's default mempool and costs nothing — so adding
`max_size` to such a call converts a free wrapper into a new pool and makes
things worse. Managed pools are documented as exempt, since `cuMemPoolCreate`
requires `maxSize == 0` for them.

A new pre-commit check `check_mempool_hygiene` enforces it. It statically parses
every `.py` under `tests/` and fails on `DeviceMemoryResourceOptions` /
`PinnedMemoryResourceOptions` constructions that omit `max_size`, including the
dict-literal options form. The check is static rather than runtime so it also
covers pools in tests that skip on the current platform. Deliberate exceptions
opt out with a `# uncapped-pool-ok: <reason>` comment.

Applying the check surfaced three call sites, all in `test_memory.py` and all
inside `pytest.raises` where validation rejects the arguments before a pool is
created; those are annotated. `test_multiprocessing_warning.py` was hardcoding
`2097152` in four places and now uses the suite-standard `POOL_SIZE` constant.
No production code changes; test-suite hygiene only.

## Checklist

### DIFF
--- a/ci/tools/check_mempool_hygiene.py
+++ b/ci/tools/check_mempool_hygiene.py
@@ -1,25 +1,26 @@
 # SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
-#
 # SPDX-License-Identifier: Apache-2.0
 
-"""Static guard against uncapped memory pools in the test suite.
+"""Check that tests do not create uncapped CUDA memory pools.
 
 A pool created without ``max_size`` reserves an address-space window sized from
-device memory rather than from what the test allocates, and the whole suite
-shares one process. Enough of those reservations exhaust the address space and
-the rest of the session fails with ``CUDA_ERROR_OUT_OF_MEMORY`` on a device with
-free physical memory (issue #2381). See AGENTS.md in this directory.
+installed device memory rather than from what the test allocates, and the whole
+cuda_core suite shares one process. Enough of those reservations exhaust the
+address space, after which the rest of the session fails with
+CUDA_ERROR_OUT_OF_MEMORY on a device with free physical memory.
 
-This check is static rather than runtime so that it also covers pools created
-by tests that are skipped on the current platform.
+See cuda_core/tests/AGENTS.md for the rule this enforces.
 """
 
+from __future__ import annotations
+
+import argparse
 import ast
-import pathlib
+import sys
+from pathlib import Path
 
-import pytest
-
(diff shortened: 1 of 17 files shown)


cuda_core/tests/AGENTS.md
```diff
@@ -0,0 +1,67 @@
+# cuda.core test suite
+
+Package-wide conventions live in `../AGENTS.md`; repository-wide ones in
+`../../AGENTS.md`. This file covers conventions specific to the tests.
+
+## Never create an uncapped memory pool
+
+A memory pool created without `max_size` reserves a virtual address window
+sized from device memory (roughly 1x device memory) regardless of what the
```
Andy-Jost: A qualifier like "installed" or "physical" (device memory) would help clarify, to me.

cuda_core/tests/AGENTS.md
```diff
@@ -0,0 +1,67 @@
...
+test actually allocates. The reservation is charged to the process address
+space even though it is not backed by physical memory, and it is not returned
+until the pool is destroyed *and* the stream-ordered frees of its outstanding
+allocations retire. The whole suite shares one process and one device, so
+these reservations accumulate across tests.
+
+When a test needs its own pool, use the suite-wide cap:
+
+```python
+POOL_SIZE = 2097152  # 2 MiB
```
Andy-Jost: We've been adding suite-wide test constants like this in ad-hoc fashion. Nothing wrong with that, but it might be worth considering a module like `helpers/constants.py`.

ci/tools/check_mempool_hygiene.py
Andy-Jost, line 1: Would this be something to put in the pre-commit hook rather than the live tests?

juenglin: Good idea.

Andy-Jost (approved):

PR #2517, opened 2026-08-06
cuda.core: accept ProgramOptions(name=None)

### LINKED ISSUE #2516: [BUG]: ProgramOptions(name=None) raises AttributeError despite the str | None annotation
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this bug and that I agree to the [Code of Conduct](https://github.com/NVIDIA/cuda-python/blob/main/CODE_OF_CONDUCT.md).

### Type of Bug

Runtime Error

### Component

cuda.core

### Describe the bug

`ProgramOptions.name` is annotated as accepting `None`:

```python
name: str | None = "default_program"
```

but `__post_init__` calls `.encode()` on it unconditionally:

```python
def __post_init__(self) -> None:
    self._name = self.name.encode()
```

So `ProgramOptions(name=None)` raises `AttributeError: 'NoneType' object has no attribute 'encode'`. The annotation and the docstring (`name : str, optional`) both read as permission to pass `None`.

This is the only field in `__post_init__` whose `None` case is not handled. `arch` normalizes two lines below:

```python
if self.arch is None:
    self.arch = f"sm_{Device().arch}"
```

The failure happens at the `.encode()` call, before `Device()` is consulted, so it reproduces without a CUDA context.

### How to Reproduce

```python
from cuda.core import ProgramOptions

ProgramOptions(name=None)
```

### Expected behavior

Either `None` is accepted and normalized to the documented default `"default_program"`, matching how `arch` is handled in the same method, or the annotation is narrowed to `str` so the type no longer advertises a value that cannot be passed.

I have a patch for the first option and will open a PR shortly.

### Additional context

The `None` path has no test coverage today: every `ProgramOptions(name=...)` in `cuda_core/tests/test_program.py` passes a string.

aryanputta:
## Description

Closes #2516.

`ProgramOptions.name` is annotated `str | None` and documented as `name : str, optional`, but `__post_init__` called `.encode()` on it unconditionally, so `ProgramOptions(name=None)` raised `AttributeError: 'NoneType' object has no attribute 'encode'`.

`name` was the only field in `__post_init__` whose `None` case was unhandled. `arch` already normalizes two lines below:

```python
if self.arch is None:
    self.arch = f"sm_{Device().arch}"
```

This applies the same treatment to `name`, so the annotation and the implementation agree.

## Behavior

`options._name` is what reaches `nvrtcCreateProgram` via `<const char*>options._name`. After this change `name=None` produces `b"default_program"`, byte-for-byte identical to what default construction produces today:

| construction | `name` | `_name` |
| --- | --- | --- |
| `ProgramOptions()` | `"default_program"` | `b"default_program"` |
| `ProgramOptions(name=None)` | `"default_program"` | `b"default_program"` |
| `ProgramOptions(name="mine")` | `"mine"` | `b"mine"` |

No existing construction changes. The only path affected is the one that previously raised before reaching any CUDA call, so nothing can be depending on the old behavior.

## Testing

Added `test_program_options_name_accepts_none`, parametrized over `None` and an explicit name, asserting both `name` and the encoded `_name`. It passes `arch="sm_90"` explicitly so it introduces no device query of its own.

The `None` path had no coverage before this: every `ProgramOptions(name=...)` in `cuda_core/tests/test_program.py` passed a string.

Scope is deliberately limited to the `None` contract. A non-`str` `name` still raises `AttributeError` rather than a typed error; that is a separate question from the annotation mismatch this fixes, and I did not want to bundle it.

## Checklist

- [x] New or existing tests cover these changes
- [x] The documentation is up to date with these changes
- [x] Release note added to `cuda_core/docs/source/release/1.2.0-notes.rst`


mdboom (approved):
Lack of type-checking in Cython strikes again 😦 

Thanks for the fix.

PR #2409, opened 2026-07-23
fix(cuda.core): fall back to driver when nvJitLink < 12.3 is installed

atiaomar1978-hub:
## Summary

Fixes #2408.

When nvJitLink 12.0–12.2 is installed, `Linker.which_backend()` and `Linker` construction raised `FunctionNotFoundError: function nvJitLinkVersion is not found` instead of warning and falling back to the driver (`cuLink`) backend.

This is a regression introduced in cuda-core 0.7.0. cuda-core 0.6.0 handled this correctly.

## Root cause

nvJitLink 12.0–12.2 only export versioned symbols (`__nvJitLink*_12_X`); the unversioned `nvJitLinkVersion` export was added in 12.3.

Since cuda-core >= 0.7.0, `_decide_nvjitlink_or_driver()` probed availability via:

```python
_optional_cuda_import(
    "cuda.bindings.nvjitlink",
    probe_function=lambda module: module.version(),
)
```

On nvJitLink 12.0–12.2, `module.version()` calls the missing unversioned symbol and raises `FunctionNotFoundError`. `_optional_cuda_import()` only treats `DynamicLibNotFoundError` as "unavailable", so the exception escapes instead of triggering driver fallback.

cuda-core 0.6.0 used `_inspect_function_pointer("__nvJitLinkVersion")` (symbol pointer inspection, no function call), which handled this correctly.

## Fix

Stop probing via `module.version()`. Import `cuda.bindings.nvjitlink` without a probe function and rely on the existing `_nvjitlink_has_version_symbol()` helper for the >= 12.3 check.

## Reproduction (before fix)

```bash
pip install "cuda-core==1.1.0" "cuda-bindings==12.9.7" "cuda-toolkit[nvjitlink]==12.0.1"
python -c "from cuda.core import Linker; print(Linker.which_backend())"
```

**Result:** `FunctionNotFoundError: function nvJitLinkVersion is not found`

Independently confirmed on **Windows x86_64** (no NVIDIA GPU) with Python 3.12.10 and `nvidia-nvjitlink-cu12==12.0.140`.

**Expected (and restored) behavior:**

```python
>>> Linker.which_backend()
RuntimeWarning: ... is too old (<12.3) ...
'driver'
```

## Tests added

- `test_decide_nvjitlink_or_driver_falls_back_when_nvjitlink_too_old`
- `test_decide_nvjitlink_or_driver_selects_nvjitlink_when_version_symbol_present`
[...]

PR #2518, opened 2026-08-06
[docs-only]: remove merge conflict markers from the cuda.core 1.2.0 release notes

aryanputta:
## Description

`cuda_core/docs/source/release/1.2.0-notes.rst` on `main` contains unresolved merge conflict markers at lines 28, 37 and 46:

```
<<<<<<< HEAD
- The default-stream singletons ...
=======
- :meth:`Linker.which_backend` ...
>>>>>>> origin/main
```

Both bullets are correct and belong in the notes. Only the three marker lines are removed; no prose is changed.

They entered in `418595b78a` (#2490), where the `#2485` default-stream bullet and the `#2409` nvJitLink bullet from `d9da44221` were left unresolved against each other.

## Why CI did not catch it

Two reasons, both worth knowing:

- The `check-merge-conflict` pre-commit hook only reports markers while the repository is mid-merge. A commit that records the markers as ordinary content passes it.
- RST renders the markers as literal text rather than failing the Sphinx build, so `Docs / Build docs` stayed green.

Without this they would ship verbatim in the published 1.2.0 release notes.

## Testing

`pre-commit run --files cuda_core/docs/source/release/1.2.0-notes.rst` passes, and the file no longer matches `<<<<<<<`, `=======` or `>>>>>>>`.


mdboom (approved):

mdboom:
I wonder how this got past pre-commit 🤔

mdboom:
Closing as already-fixed.

Issue #1824, opened 2026-03-25
Known issue: `cuda.pathfinder` cannot be imported in PEP 517 in-tree build backends without a workaround


rwgk:
### Problem

Packages that use `cuda` as a namespace package (e.g. `cuda-bindings`, `cuda-core`) and define an in-tree build backend via `backend-path=["."]` in `pyproject.toml` cannot `import cuda.pathfinder` at build time, even when `cuda-pathfinder` is listed in `build-system.requires` and correctly installed in the build environment.

### Cause

When a PEP 517 build frontend (`pip install`, `python -m build`) loads an in-tree build backend, it prepends the project directory to `sys.path`. At that point, `import cuda` resolves the `cuda` namespace package to *only* the project's `cuda/` subdirectory. Python caches a `_NamespacePath` for `cuda` in `sys.modules`. The `cuda/pathfinder/` tree installed under `site-packages` is never merged into that namespace, so `import cuda.pathfinder` raises `ModuleNotFoundError`.

`pkgutil.extend_path` does not help: `_NamespacePath._recalculate()` re-derives its path list from the import machinery's finders, discarding manual additions.

### Workaround

Replace the `_NamespacePath` with a plain `list` that includes the `site-packages/cuda/` directory:

```python
import os
import sys

def _import_get_cuda_path_or_home():
    """Import get_cuda_path_or_home, working around PEP 517 namespace shadowing.

    See https://github.com/NVIDIA/cuda-python/issues/1824 for why this helper is needed.
    """
    try:
        import cuda.pathfinder
    except ModuleNotFoundError as exc:
        if exc.name not in ("cuda", "cuda.pathfinder"):
            raise
        try:
            import cuda
        except ModuleNotFoundError:
            cuda = None

        for p in sys.path:
            sp_cuda = os.path.join(p, "cuda")
            if os.path.isdir(os.path.join(sp_cuda, "pathfinder")):
                cuda.__path__ = list(cuda.__path__) + [sp_cuda]
                break
        else:
            raise ModuleNotFoundError(
                "cuda-pathfinder is not installed in the build environment. "
                "Ensure 'cuda-pathfinder>=1.5' is in build-system.requires."
            )
        import cuda.pathfinder

    return cuda.pathfinder.get_cuda_path_or_home
```

[...]

PR #792, opened 2025-08-04
Update `cuda.bindings` to 13.0.0

leofang:
## Description

Close #791.

## Checklist

- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #2519, opened 2026-08-06
build: report cuda-bindings provenance in PEP 517 builds

rwgk:
## Description

Related to #1824 and #2468.

Follow-up to #2509. This PR is intended to merge before #2510; it is submitted separately because contributor branches must be hosted on a fork and GitHub does not currently support cross-fork stacked bases. #2510 separately enforces exact local-wheel selection, while this PR only makes the build hook's existing provenance reporting work consistently.

#2509 added a `cuda.core` build-hook message that reports the imported `cuda-bindings` version and package directory. The message appears for previous-CTK CUDA 12 builds, but not for the 24 current-CTK CUDA 13 builds.

The difference comes from call ordering in the PEP 517 in-tree backend. `cuda_core/build_hooks.py` tries to import `cuda.bindings` before calling `_get_cuda_path()`. At that point, `backend-path = ["."]` can leave the repository's `cuda_core/cuda` directory cached as the only path in the `cuda` namespace, so the installed build dependency is not visible. The import failure is then suppressed by the existing best-effort guard. `_get_cuda_path()` invokes the namespace repair documented in #1824, but currently does so only after that import attempt.

CUDA 12 wheels happen not to expose the problem because they still contain the historical `_cuda_bindings_redirector.pth`, which imports `cuda` during interpreter startup. That redirector stopped shipping with CUDA 13 as part of #792, explaining why only the previous-CTK builds currently reach the provenance print.

This PR moves the existing `_get_cuda_path()` call ahead of the `cuda.bindings` import and reuses its result later in the build. The namespace is therefore repaired before the import, allowing the existing #2509 message to report the CUDA 13 dependency's version and location and allowing the existing Cython path setup to use that imported package.

This does not change build requirements or dependency-resolution policy. `_get_cuda_path()` was already mandatory later in the same function, so the functional change is limited to performing that lookup earlier.

## Testing

- `TestVenv/bin/python -m pytest cuda_core/tests/test_build_hooks.py -v --noconftest` (15 passed)
- The current-CTK CI matrix provides end-to-end verification: its build logs should now contain `Using cuda-bindings 13...` messages.

### DIFF
--- a/cuda_core/build_hooks.py
+++ b/cuda_core/build_hooks.py
@@ -133,6 +133,9 @@ def _build_cuda_core(debug=False):
     # This function populates "_extensions".
     global _extensions
 
+    # Resolve CUDA first so the pathfinder import repairs PEP 517 namespace shadowing before importing bindings.
+    cuda_path = _get_cuda_path()
+
     # Add cuda-bindings to sys.path so Cython can find .pxd files
     # This is needed for editable installs where meta path finders don't work for Cython
     # We need to add the directory containing the 'cuda' package so Cython can resolve
@@ -178,7 +181,7 @@ def _build_cuda_core(debug=False):
 
         return sources
 
-    all_include_dirs = [os.path.join(_get_cuda_path(), "include")]
+    all_include_dirs = [os.path.join(cuda_path, "include")]
     extra_compile_args = []
     extra_link_args = []
     extra_cythonize_kwargs = {}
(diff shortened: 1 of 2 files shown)


rwgk:
Closing: I want this change to become part of the existing 2509, 2510 stack. codex and I are still learning how to do that.

rwgk:
The stacked version of this PR is #2520

PR #2522, opened 2026-08-06
"cuda.core tests: reserve driver-managed mempools up front"

### LINKED ISSUE #2381: [BUG]:  Intermittent CUDA_ERROR_OUT_OF_MEMORY cascade in cuda_core test suite

juenglin:
closes #2381

Root cause and measurements are in [this comment](https://github.com/NVIDIA/cuda-python/issues/2381#issuecomment-5207900890); in short, the driver's default memory pool and graph memory pool each permanently reserve ~2x installed device memory of virtual address space, and neither can be capped or released.

A session-scoped autouse fixture now takes both reservations up front, so they land back to back in a nearly empty address space instead of whenever test order happens to trigger them. This does not reduce the footprint; it just makes the big reservations deterministic. If either reservation is refused the session aborts with an explanation, rather than cascading ~194 identical OOM failures that say nothing about the cause.

Available address space is reported before and after via `cuMemAddressReserve` rather than an OS query, so it behaves the same on Windows and Linux. Devices without mempool support are skipped, not failed. Escape hatches: `CUDA_CORE_TEST_SKIP_EARLY_RESERVATION=1` disables the fixture (use as the control when measuring whether it helps or is still needed).

The fixture can be removed once either driver-side improvement lands:
1. the per-pool reservation is no longer ~2x device memory, or becomes cappable
   (`cuGraphAddMemAllocNode` currently accepts `poolProps.maxSize` but ignores it); or
2. the per-process address-space budget stops being the binding constraint, which the
   fixture's own `remaining headroom: N more pool-sized reservations` line reports —
   an N comfortably above 1 means ordering can no longer matter on that machine.

## Checklist


Andy-Jost (approved):

rluo8 (approved):

Andy-Jost:
The two cancelled tests appear to be a real deadlock or other hang.

PR #2523, opened 2026-08-06
ci: limit pytest duration reports to the slowest 20 tests

Andy-Jost:
## Summary
CI was passing `--durations=0`, which prints a duration line for every test phase and makes job logs hard to scan. Switch wheel-test workflows and `ci/tools/run-tests` to `--durations=20` so only the slowest 20 entries are reported.

## Changes
- `.github/workflows/test-wheel-linux.yml`
- `.github/workflows/test-wheel-windows.yml`
- `ci/tools/run-tests`

### DIFF
--- a/.github/workflows/test-wheel-linux.yml
+++ b/.github/workflows/test-wheel-linux.yml
@@ -428,7 +428,7 @@ jobs:
         if: ${{ inputs.test-mode == 'nightly-pytorch' }}
         run: |
           pushd cuda_core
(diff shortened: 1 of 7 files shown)


juenglin (approved):

mdboom (changes requested):
Could we put this in the `[tool.pytest]` section in the `pyproject.toml` files to reduce the duplication and make it apply by default for developers?

mdboom (approved):

Issue #1804, opened 2026-03-21
[BUG]: cudaGraphGetEdges() suffers use-after-free for cudaGraphEdgeData


galv:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this bug and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Type of Bug

Silent Failure

### Component

cuda.bindings

### Describe the bug

Take a look here:

https://github.com/NVIDIA/cuda-python/blob/346afc808db7f7573f9b7a7c2d297868ea7e4894/cuda_bindings/cuda/bindings/runtime.pyx.in#L35898-L35901

We are creating a python cudaGraphEdgeData object via a pointer to the underlying C cudaGraphEdgeData struct.

That underlying constructor https://github.com/galv/cuda-python/blob/346afc808db7f7573f9b7a7c2d297868ea7e4894/cuda_bindings/cuda/bindings/runtime.pyx.in#L18241 just saves that pointer, rather doing a deep copy. Thus, it relies upon that underlying pointer staying alive.

Then, cudaGraphGetEdges() immediately frees the underlying pointer that the python objects reference, https://github.com/NVIDIA/cuda-python/blob/346afc808db7f7573f9b7a7c2d297868ea7e4894/cuda_bindings/cuda/bindings/runtime.pyx.in#L35901, and returns the list of python objects which are now backed by dead pointers.

Thus, the cudaGraphEdgeData returned by cudaGraphGetEdges() always suffers a use-after-free. You need to do a deep-copy.

The driver code suffers from the same issue.

Please fix this.

### How to Reproduce

I don't think you need a repro to see the problem in this case.

### Expected behavior

Do a deep copy of the underlying data.

leofang:
@mdboom would you mind taking over this?

leofang:
Closed by #2083.

PR #2524, opened 2026-08-06
test(cuda.core): cover PDL same-stream overlap via GraphBuilder capture

### LINKED ISSUE #1334: Support programatic dependent launch (PDL)

lijinf2:
## Description
close #1334 

- Add GraphBuilder coverage for `programmatic_stream_serialization`: programmatic dependency-edge mapping and same-stream primary/secondary overlap (Hopper+, opportunistic → xfail).
- Share the overlap protocol in `helpers/pdl_kernels.py` so direct and graph-capture paths stay in sync.
- Tweak 1.2.0 notes to use the runtime PDL attribute name.

## Checklist


cuda_core/tests/helpers/pdl_kernels.py
```diff
@@ -0,0 +1,123 @@
...
+    }
+    """
+    arch = "".join(f"{i}" for i in device.compute_capability)
+    pro_opts = ProgramOptions(std="c++17", arch=f"sm_{arch}", include_path=helpers.CUDA_INCLUDE_PATH)
+    prog = Program(code, code_type="c++", options=pro_opts)
+    mod = prog.compile("cubin")
+    return mod.get_kernel("primary_kernel"), mod.get_kernel("secondary_kernel")
+
+
+def run_pdl_overlap_check(device, *, via_graph: bool = False):
```
isVoid: Why do we need to test with graph launch? This seems like an additional requirement for programmatic stream.

cuda_core/tests/graph/test_graph_builder.py
```diff
@@ -694,3 +696,71 @@ def test_graph_definition_conditional_body_during_capture_raises(init_cuda):
...
+        ed = edge_data[0]
+        # Driver (cuda.h) ↔ Runtime / Programming Guide (driver_types.h):
+        #   CU_GRAPH_DEPENDENCY_TYPE_PROGRAMMATIC  ↔ cudaGraphDependencyTypeProgrammatic
+        #   CU_GRAPH_KERNEL_NODE_PORT_PROGRAMMATIC ↔ cudaGraphKernelNodePortProgrammatic
+        assert ed.type == driver.CUgraphDependencyType.CU_GRAPH_DEPENDENCY_TYPE_PROGRAMMATIC, ed.type
+        assert ed.from_port == driver.CU_GRAPH_KERNEL_NODE_PORT_PROGRAMMATIC, ed.from_port
+
+    mod = compile_common_kernels()
+    producer = mod.get_kernel("add_one")
+    consumer = mod.get_kernel("add_one")
```
isVoid: So... I'm curious. `add_one` is a simple kernel that doesn't have programmatic dependency like the kernel `primary_kernel`. Setting `programmatic_stream_serialization` between for the second launch still create a graph dependency between these two kernels?

cuda_core/tests/helpers/pdl_kernels.py
lijinf2: Yeah, that is a request in #1334 
per "PDL launches are graph-compatible and this use case should be tested and showcased"

cuda_core/tests/graph/test_graph_builder.py
lijinf2: Yeah, here just needs two dummy kernels as input so that we can check things in _assert_programmatic_dependency_edge. 

Renamed to dummy_kernel.

isVoid (approved):

cuda_core/tests/graph/test_graph_builder.py
```diff
@@ -709,3 +711,134 @@ def test_graph_definition_conditional_body_during_capture_raises(init_cuda):
...
+        """Assert capture of ProgrammaticStreamSerialization produced a programmatic edge.
+
+        Per Programming Guide (Programmatic Dependent Launch): stream-capturing a
+        secondary launch with ``cudaLaunchAttributeProgrammaticStreamSerialization``
+        maps to a programmatic dependency edge from the programmatic kernel port.
+        """
+        from cuda.bindings import driver
+
+        h_graph = graph_definition.handle
+        err, _, _, _, num_edges = driver.cuGraphGetEdges(h_graph)
```
isVoid: This API seems to return different number of results based on the major version so you can do:

```suggestion
    if driver.CUDA_VERSION >= 13000:
        err, _, _, _, num_edges = driver.cuGraphGetEdges(h_graph)
    else:
        err, _, _, _, num_edges = driver.cuGraphGetEdges_v2(h_graph)
```

Though what stands out here is that `cuda.core` graph APIs discards the edge data to prevent implementing the whole helper solely with `cuda.core`. A feature gap we can consider adding?

lijinf2:
@isVoid Thanks for the commit that handles the versioned cuGraphGetEdges / cuGraphGetEdges_v2 API!

I dug into the remaining CI failures on the CUDA 13.0.x rows. They weren’t from the versioned-edge-API handling — those jobs install published cuda-bindings==13.0.* from PyPI, and that line still has the use-after-free in cuGraphGetEdges where the returned CUgraphEdgeData wrappers point at a freed scratch buffer (#1804). The fix only shipped in 13.3.0 (and was backported to 12.9.7 on the 12.x branch), so 13.0.x / 13.1.x / 13.2.x published wheels don’t have it. 

I pushed a follow-up that skips the edge-data assertion on those older bindings.

cuda_core/tests/graph/test_graph_builder.py
lijinf2: Good point — yeah, without edge metadata on the cuda.core graph APIs this helper has to go through cuda.bindings.

lijinf2:
pre-commit.ci run

PR #2526, opened 2026-08-06
cuda.core: capture complete Buffer deallocation recipe at creation

### LINKED ISSUE #2497: [BUG]: Buffer cleanup must capture a complete deallocation recipe at creation
## Summary

Buffer / device-pointer destruction can run with no suitable CUDA context current on the destroying thread. The free then fails (or was historically treated as success for `CUDA_ERROR_INVALID_CONTEXT`), leaking memory. This shows up especially for graph-retained buffers (deferred cleanup) and can also arise when the last reference is dropped off the allocating thread (e.g. cyclic GC under free-threading).

The fix is to **capture a complete cleanup recipe at Buffer allocation / binding time**, so teardown does not depend on ambient thread state for context.

## Problem

Owning `Buffer`s eventually call into the driver to free (pool / `cuMemFreeAsync`, MR `deallocate`, etc.). That call needs:

1. A **stream** that defines free ordering (for stream-ordered paths), and
2. A **context** that makes the driver call legal when destruction runs.

Destruction is not guaranteed to run on the allocating thread with that context still current:

- Graph attachments are released via deferred cleanup (CUDA user-object dtor cannot call CUDA; work is drained later via pending call / another thread).
- Ordinary last-ref / GC teardown can run on whichever thread drops the last reference.

Today, some owning paths store a deallocation `StreamHandle`, but default-stream **tokens** (`CU_STREAM_LEGACY` / `CU_STREAM_PER_THREAD`) are not real streams: their meaning depends on ambient state when the driver call runs. Separately, some adoption paths never record a stream at all and fall back to `default_stream()` only at destructor time (see `Buffer.from_handle` with `mr=`).

A `StreamHandle` already embeds an owning `ContextHandle` for created streams (resource-handles model). The cleanup recipe should be that handle (stream + context), not a side channel.

## Proposed direction

At allocation / binding, record a **complete** deallocation recipe: a `StreamHandle` that carries the context needed to replay the free. At free time: make that context current if needed, free on that stream, restore the prior binding.

Concrete work includes:

- Ensure every owning Buffer path captures the recipe when the Buffer is created (pool allocate, async alloc, IPC import, MR-backed adoption, etc.).
- Add missing `stream=` (keyword-only) where adoption APIs take ownership of stream-ordered teardown — notably `Buffer.from_handle` / `ManagedBuffer.from_handle` when `mr=` is set — so the destructor is not the first place a stream is chosen.
- Bind context into default-stream tokens at capture when forming the stored `StreamHandle` (legacy and PTDS tokens both need a definite context for the free call to be legal).
- Fail loudly on cleanup errors; do not treat `CUDA_ERROR_INVALID_CONTEXT` as successful deallocation.

## Per-Thread Default Stream (PTDS) policy

`CU_STREAM_PER_THREAD` is local to **both** the host thread and the context. There is no public API to promote the PTDS token to a transferable real `CUstream` (driver-internal PTDS streams are not given public handles). Therefore, when PTDS is used for destruction ordering it is impossible to capture a complete deallocation recipe. Deallocations running on a separate thread will be ordered relative to that thread's default stream rather than the default stream that was active at allocation time.

**Decision:** keep the PTDS token in the captured recipe (token + bound context). cuda.core will restore the context and pass `CU_STREAM_PER_THREAD` at free time. **Applications are responsible for ordering** if they use PTDS together with:

- free-threading / off-thread last-ref or GC destruction, or
- graph-retained buffers (deferred cleanup is not the allocating thread).

Same-thread PTDS alloc/use/free remains coherent.

**Optional UX:** detect when a Buffer whose deallocation stream is PTDS is attached to a graph, and emit a warning that deferred release will not preserve the allocating thread's PTDS ordering. (Warning only — consistent with accepting the risk. Free-threading GC has the same hazard without graphs and should be called out in docs even if we only warn on graph attach.)

Related: bare default-stream **query** caching is #2485 / #2490 (ambient token semantics for `.context` / `.device`). Complementary; does not by itself fix captured free-ordering.

## Non-goals

- Choosing the free stream only in the destructor via `default_stream()` when ownership was taken earlier.
- Replacing PTDS with a newly created stream at capture, or forbidding PTDS at Buffer creation (rejected; see policy above).
- A public "promote PTDS to real stream" API (does not exist).

## Related

- #2485 / #2490 — default-stream tokens must not sticky-cache context on the singleton.
- #2001 (closed) — explicit `stream=` on allocate/deallocate paths; adoption / `from_handle` still incomplete for recorded teardown.
- #1989 — `from_handle` / lifetime consistency more broadly.

## Acceptance sketch

- Owning Buffer creation always records a complete cleanup recipe (`StreamHandle` with embedded context for the free).
- `from_handle(..., mr=...)` (and managed counterpart) accept `stream=` and store it for teardown; no destructor-only default-stream fallback for that path.
- Deferred / no-current-context destruction can activate the bound context and free.
- PTDS + cross-thread / graph deferred free is documented as app-ordered; optional warning on graph attach of PTDS-backed buffers.

Andy-Jost:
## Summary

Closes #2497. Supersedes the approach in draft #2455.

Buffer / device-pointer destruction can run with no suitable CUDA context current (deferred graph cleanup, off-thread GC). Owning paths now capture a complete free-ordering recipe at allocation or binding time — a `StreamHandle` with a bound context for default-stream tokens — and activate that context at teardown instead of relying on ambient thread state.

## Changes

- Introduce an internal `DeallocationStream` (stream handle + optional PTDS host-thread id) and bind context into default-stream tokens when recording free ordering.
- Activate the bound context around device-pointer teardown; report cleanup failures instead of silently skipping frees.
- Add keyword-only `stream=` on `Buffer.from_handle` / `ManagedBuffer.from_handle` when `mr=` owns the pointer, so the destructor is not the first place a stream is chosen.
- Stop treating `CUDA_ERROR_INVALID_CONTEXT` as success in `_MemPool.deallocate`, and let explicit `mr.deallocate()` raise; destruction still contains errors in the MR dealloc callback.
- Document PTDS deallocation ordering on the relevant `stream` parameters and note the fix in the 1.2.0 release notes.

PTDS remains app-ordered for cross-thread / deferred free (no attach-time warning): graph kernel-arg retention cannot be detected completely without traversing argument objects, and an incomplete warning would fire on correct single-threaded PTDS usage.

## Related

- #2497
- #2455 (earlier draft; different design)
- #2485 / #2490 (default-stream token query caching; complementary)


cuda_core/cuda/core/_memory/_buffer.pyx
```diff
@@ -74,6 +75,21 @@ cdef void _mr_dealloc_callback(
...
 
 
+cdef inline void _require_deallocation_stream_context(Stream s) except *:
+    """Default-stream tokens need a current context to pin into the free recipe."""
+    cdef ContextHandle h_ctx
+    if get_stream_context(s._h_stream):
+        return
+    h_ctx = get_current_context()
+    if h_ctx:
+        return
```
juenglin: I don't understand this. Did you mean to set `s._h_stream` as a side-effect here or return a context?

juenglin:
A few gaps in the new test coverage and some refactoring opportunities:

**Missing coverage**

1. `test_from_handle_mr_records_default_stream` and `test_from_handle_mr_records_explicit_stream`
   are `Buffer`-only. The same code path runs for `ManagedBuffer.from_handle` (it delegates to
   `Buffer._init`), but if it ever regressed the tests would not catch it. Suggest parametrizing
   both with `[Buffer, ManagedBuffer]`.

2. Similarly, `test_from_handle_stream_requires_mr` only tests `Buffer`. The `stream requires a
   memory resource` guard applies equally to `ManagedBuffer.from_handle` and should be covered.

3. The `_require_deallocation_stream_context` guard is also invoked inside `Buffer_close` when a
   `stream` argument is passed. There is no test that calls `buf.close(stream=default_stream())`
   with no context current and asserts the `RuntimeError`. The equivalent test for `from_handle`
   (`test_from_handle_mr_default_stream_requires_context`) exists but the `close()` path does not.

**Refactoring**

Five near-identical inline `MemoryResource` subclasses (`StrictCapturingMR`, `CapturingMR`,
`NoopMR`) are defined across the deallocation-stream tests. Suggest extracting two module-level
helpers:
- `_make_capturing_mr(device)` → returns `(mr, captured)` where `captured["stream"]` is set on
  each `deallocate` call.
- `_make_noop_mr(device)` → returns a device-only MR with a no-op `deallocate`.

Also, `Stream_accept` and `default_stream` are re-imported locally inside several test functions.
These should be module-level imports. Two of the new tests (`test_buffer_close_default_stream_requires_context`
and `test_managed_buffer_from_handle_mr_records_default_stream`) call `default_stream()` without
importing it at all — latent `NameError`s if those lines are reached.

juenglin (approved):

cuda_core/cuda/core/_memory/_buffer.pyx
Andy-Jost: This is anticipating an error that could occur later in `make_deallocation_stream`: when no context is active, we should raise an error. I think the cleanest fix is to remove this function entirely, relying on a return value of `CUDA_ERROR_INVALID_CONTEXT` from the aforementioned function.

PR #2532, opened 2026-08-07
fix(pixi): restore conda test deps and relock after #2384

juenglin:
## Summary
- #2384 inserted `[feature.test.pypi-dependencies]` mid-table in `cuda_bindings`/`cuda_core` `pixi.toml`, accidentally moving conda test deps (`pytest-benchmark`, `numpy`, `pyglet`, …) to PyPI and leaving the lockfiles stale.
- Fresh CI installs then dropped the local `cuda-bindings`/`cuda-core` source packages, so the pixi source-build smoke failed with `ModuleNotFoundError: No module named 'cuda'`.
- Restore those deps under `[feature.test.dependencies]`, keep only `cuda-python-test-helpers` under pypi, and regenerate both lockfiles with Pixi 0.73 until `pixi lock --check` is clean.

## Test plan
- [x] `pixi lock --check --manifest-path cuda_bindings`
- [x] `pixi lock --check --manifest-path cuda_core`
- [x] Re-run `CI: pixi run test (source build)` (build smoke) on this branch and confirm `bindings import OK` / `core import OK`

### DIFF
--- a/cuda_bindings/pixi.lock
+++ b/cuda_bindings/pixi.lock
@@ -20,6 +20,8 @@ environments:
   cu12:
     channels:
     - url: https://conda.anaconda.org/conda-forge/
+    indexes:
+    - https://pypi.org/simple
     packages:
       linux-64:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/_openmp_mutex-4.5-20_gnu.conda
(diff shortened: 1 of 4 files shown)


rwgk (approved):
Thanks @juenglin! (sorry I didn't see this before in the ever swelling flood of emails)

PR #2534, opened 2026-08-07
Fix CUDA Runtime CTK test fixture on Windows ARM64

isVoid:
## What changed

Update the Windows CUDA Runtime CTK test fixture to model the current CUDA 13 layout explicitly:

- x64 creates `bin/x64/cudart64_13.dll`;
- Arm64 creates `bin/arm64/cudart64_13.dll`;
- Linux behavior is unchanged.

## Why

#2393 made Windows CTK dynamic-library searches architecture-aware. Native Arm64 Python now correctly searches `bin/arm64`, but `_create_cudart_in_ctk` still created `cudart64_12.dll` in the legacy unqualified `bin` directory. As a result, `test_try_via_ctk_root_regular_lib` failed only under native Windows Arm64.

The fixture now targets the latest CUDA layout for both Windows architectures. Production discovery continues to retain its legacy unqualified `bin` fallback for older x64 toolkits.

## Validation

- Native Windows Arm64 focused regression: 1 passed.
- `tests/test_ctk_root_discovery.py`: 37 passed.

### DIFF
--- a/cuda_pathfinder/tests/test_ctk_root_discovery.py
+++ b/cuda_pathfinder/tests/test_ctk_root_discovery.py
@@ -81,9 +81,9 @@ def _create_nvvm_in_ctk(ctk_root):
 def _create_cudart_in_ctk(ctk_root):
     """Create a fake cudart lib in the platform-appropriate CTK subdirectory."""
     if IS_WINDOWS:
-        lib_dir = ctk_root / "bin"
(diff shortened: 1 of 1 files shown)


isVoid:
Ah, closing in favor of duplicate PR: https://github.com/NVIDIA/cuda-python/pull/2528

PR #2536, opened 2026-08-09
Use pathlib in cuda.pathfinder._headers (part 8 of #2410)

LeSingh1:
Part 8 of #2410 (`os.path` -> `pathlib.Path`), covering `cuda_pathfinder/cuda/pathfinder/_headers`.

This is the last un-migrated corner of `cuda.pathfinder` runtime code that isn't already
covered by an open part of the series.

### What changed

- `_locate_in_anchor_layout()` and `resolve_conda_anchor()` build and return `Path`
  objects, and their anchor arguments accept `str` or any `os.PathLike[str]` (per your
  note on #2410 about widening what we accept).
- The `os.path.isfile(os.path.join(...))` probes become `(dir / name).is_file()`, which
  makes the `_joined_isfile` helper unnecessary.
- `LocatedHeaderDir.abs_path` stays a `str`: it is public API (`LocatedHeaderDir` is
  re-exported from `cuda.pathfinder`), so the find steps convert at the boundary, the
  same way #2493 does for `LocatedBitcodeLib`/`LocatedStaticLib`.

### Two deliberate exceptions, flagged for review

1. **`_abs_norm()` keeps `os.path.abspath()`.** pathlib has no purely lexical
   normalization: `Path.resolve()` follows symlinks, which would change the reported
   header directory under the very common `/usr/local/cuda -> /usr/local/cuda-<ver>`
   layout, and `Path.absolute()` does not collapse `..`. I'd rather leave one clearly
   commented `os.path` call here than silently change what the public API returns.

   While there, the wrapping `os.path.normpath()` is dropped as redundant: both
   `posixpath.abspath` and `ntpath.abspath` already normalize.

2. **`glob.glob()` stays** for the descriptor catalog's glob patterns. Only the
   `os.path` parts of those expressions move to `Path`. `Path.glob()` is not a drop-in
   replacement (it needs a separate base directory for absolute patterns, and it does
   not skip dot-entries the way `glob.glob()` does), so swapping it in would be a
   behavior change unrelated to this issue.

Following your call on #2410, the single-file `is_file()` probes keep pathlib's error
behavior rather than being wrapped to mimic `os.path.isfile()`'s "any stat error means
`False`".

### Verification

No behavior change is intended, so there is no new test; the existing suite is the
check. `cuda_pathfinder/tests/test_find_nvidia_headers.py` exercises all five find
steps, including the conda `targets/*/include` layout and the CTK-root canary fallback.

- `pytest cuda_pathfinder/tests` — same pass/fail set as `upstream/main` on this machine
  (the only failures are the ones that need a real CUDA install).
- `mypy cuda/pathfinder/_headers/` — no new errors (the two pre-existing
  `redundant-cast` errors in `header_descriptor.py` are untouched and out of scope).
- `ruff check` and `ruff format --check` clean on both files.

Refs #2410.

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_headers/find_nvidia_headers.py
+++ b/cuda_pathfinder/cuda/pathfinder/_headers/find_nvidia_headers.py
@@ -8,6 +8,7 @@ import glob
 import os
 from collections.abc import Callable
 from dataclasses import dataclass
+from pathlib import Path
 from typing import TYPE_CHECKING
 
 from cuda.pathfinder._dynamic_libs.load_nvidia_dynamic_lib import (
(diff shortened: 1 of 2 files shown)


cuda_pathfinder/cuda/pathfinder/_headers/find_nvidia_headers.py
```diff
@@ -95,9 +96,9 @@ def find_in_conda(desc: HeaderDescriptor) -> LocatedHeaderDir | None:
     anchor_point = resolve_conda_anchor(desc, conda_prefix)
     if anchor_point is None:
         return None
-    found_header_path = _locate_in_anchor_layout(desc, anchor_point)
-    if found_header_path:
-        return LocatedHeaderDir(abs_path=found_header_path, found_via="conda")
+    found_header_dir = _locate_in_anchor_layout(desc, anchor_point)
+    if found_header_dir is not None:
+        return LocatedHeaderDir(abs_path=str(found_header_dir), found_via="conda")
```
mdboom, line 101: Let's update `LocatedHeaderDir` to store a `Path`.  That isn't really public API.  Again, that should simplify the conversion back and forth.

mdboom (changes requested):

Issue #2313, opened 2026-07-07
[FEA]: add workaround for loading broken cufile wheel


gforsyth:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

cuda.pathfinder

### Is your feature request related to a problem? Please describe.

In https://github.com/rapidsai/kvikio/pull/1000 we ran across a bug in `nvidia-cufile`, where the upstream `cufile` wheel is improperly linked and fails to load `librt.so.1` and complains about missing `shm_copy`.

**Note:** this is definitely *NOT* a `pathfinder` bug, just reporting it in case you want to incorporate the workaround we've added in `kvikio` to help other users (while we wait to see if it's fixed and backported upstream)

### Environment

Running on an arm64 machine using this container: `rapidsai/citestwheel:26.08-cuda13.0.3-rockylinux8-py3.12`

Install `cufile` and `cuda-python`: `pip install nvidia-cufile==1.15.1.6.* cuda-python`

### Loading error

Attempting to load `cufile` errors:
```python
Python 3.12.13 (main, Jul  1 2026, 20:40:39) [GCC 8.5.0 20210514 (Red Hat 8.5.0-28)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from cuda.pathfinder import load_nvidia_dynamic_lib
>>> load_nvidia_dynamic_lib('cufile')
Traceback (most recent call last):
  File "/pyenv/versions/3.12.13/lib/python3.12/site-packages/cuda/pathfinder/_dynamic_libs/load_dl_linux.py", line 218, in load_with_abs_path
    handle = _load_lib(desc, found_path)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/pyenv/versions/3.12.13/lib/python3.12/site-packages/cuda/pathfinder/_dynamic_libs/load_dl_linux.py", line 155, in _load_lib
    return ctypes.CDLL(filename, cdll_mode)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/pyenv/versions/3.12.13/lib/python3.12/ctypes/__init__.py", line 379, in __init__
    self._handle = _dlopen(self._name, mode)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2537, opened 2026-08-09
Recover from the nvidia-cufile "undefined symbol: shm_open" load failure

LeSingh1:
Closes the failure reported in #2313.

`nvidia-cufile` wheels reference `shm_open()` but ship no `DT_NEEDED` entry for librt.
On glibc < 2.34 — where `shm_open` still lives in `librt` rather than in libc —
`dlopen()` therefore fails and `load_nvidia_dynamic_lib("cufile")` dies with:

```
RuntimeError: Failed to dlopen .../nvidia/cu13/lib/libcufile.so.0:
.../libcufile.so.0: undefined symbol: shm_open
```

Loading `librt.so.1` with `RTLD_GLOBAL` first makes the reference resolve. That is the
workaround kvikio carries today (rapidsai/kvikio#1000), and the one @gforsyth wrote up
in the issue.

### Shape of the fix

The obvious implementation is an unconditional preload in `_work_around_known_bugs()`,
next to the existing `nvrtc` entry. I did not do that, because it would put librt into
the global symbol scope of every process that loads cufile, forever, including the
overwhelming majority where nothing is broken.

Instead the workaround is strictly **reactive**: it runs only after the primary
`dlopen()` has already failed, and only when the failure carries exactly this signature
(libname `cufile` **and** `shm_open` in the dlopen error).

That buys three things:

- the normal path is byte-for-byte unchanged — no extra `dlopen` on success;
- the workaround disappears by itself once the wheel is fixed, so it cannot mask a
  future regression (the concern already written into the `nvrtc` workaround's comment);
- if the retry does not help, the **original** error is what the user sees, since it is
  the more informative one.

### Scope notes

- Only `load_with_abs_path()` is covered. That is the path taken for the wheel install
  in the report (found via site-packages). I deliberately left `load_with_system_search()`
  alone rather than widen the blast radius beyond the reported symptom — happy to extend
  it if you'd rather have both.
- No descriptor/catalog change: this is a defect in one specific wheel build, not a
  standing dependency of cufile, so it does not belong in `dependencies`.

### Tests

New `cuda_pathfinder/tests/test_load_dl_linux.py` (the Linux-side counterpart to
`test_load_dl_windows.py`). No real cufile install is needed — `_load_lib` and
`ctypes.CDLL` are stubbed, so the tests reproduce the exact failure signature
deterministically. Coverage:

- the retry succeeds after librt is loaded (**fails on `main`**);
- the original error is reported when the retry still fails (**fails on `main`**);
- the retry is skipped when librt itself cannot be loaded (**fails on `main`**);
- an unrelated cufile failure, and the same `shm_open` symptom on a *different* library,
  do **not** trigger the workaround — these two pass on `main` as well, on purpose; they
  are there to pin down that the trigger stays narrow.

Verified: 3 of the 5 fail against `upstream/main` and all 5 pass with the change; the
rest of `cuda_pathfinder/tests` has the same pass/fail set as `main`. `ruff check`,
`ruff format --check` clean, and `mypy` reports no new errors (the one pre-existing
`LoadedDL` `arg-type` error in this file is untouched).

One question for you: #2313 is assigned to @juenglin — if that work is already in
flight, say the word and I'll close this.

Refs #2313

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
+++ b/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py
@@ -208,6 +208,35 @@ def _work_around_known_bugs(libname: str, found_path: str) -> None:
                     ctypes.CDLL(dep_path, CDLL_MODE)
 
 
+def _retry_load_after_working_around_known_bugs(
+    desc: LibDescriptor, found_path: str, error: OSError
+) -> ctypes.CDLL | None:
+    """Second-chance load for known-broken NVIDIA wheels.
+
+    Returns the loaded library, or None if no workaround applies or the retry
+    also fails. Kept strictly reactive: nothing here runs unless the primary
+    load already failed with the exact signature of the known bug, so it
+    disappears by itself once the offending wheel is fixed.
+    """
+    if desc.name != "cufile" or "shm_open" not in str(error):
+        return None
+    # Work around bug in nvidia-cufile wheels (observed with
+    #   nvidia_cufile_cu13-1.15.1.6-py3-none-manylinux_2_28_aarch64.whl)
+    # Issue: libcufile.so.0 references shm_open() but has no DT_NEEDED entry
+    # for librt, so dlopen() fails with "undefined symbol: shm_open" on
+    # glibc < 2.34, where shm_open lives in librt rather than in libc.
+    # Making librt's symbols globally visible first lets the reference resolve.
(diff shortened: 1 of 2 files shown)


juenglin:
Thanks for the contribution, @LeSingh1.

Considering that 1. the chance of users encountering this issue is very low, and 2. that we have documented two workarounds [here](https://github.com/NVIDIA/cuda-python/issues/2313) I decided not to merge this.

PR #2544, opened 2026-08-09
fix(ci): fail cleanly on a malformed cuda.build.version instead of tracebacking

LeSingh1:
## Problem

`ci/tools/check_pixi_cuda_version.py` returns a deliberate diagnostic exit code for every problem it anticipates — missing `versions.yml`, missing `cuda.build.version`, missing `pixi.toml`, missing feature key — and then unpacks the version with no checking at all:

```python
try:
    build_version = yaml.safe_load(...)["cuda"]["build"]["version"]
except (KeyError, TypeError):
    print(f"error: cuda.build.version not found in {VERSIONS_FILE_PATH}", file=sys.stderr)
    return 2

major, minor, *_ = build_version.split(".")
```

YAML makes that easy to break. `ci/versions.yml` quotes the value today:

```yaml
cuda:
  build:
    version: "13.3.0"
```

Drop the quotes and a two-component value loads as a **float**, a bare number as an **int** — neither has `.split`. And a quoted-but-single-component value survives to the tuple unpacking and dies there instead.

Driving `main()` against the real `pixi.toml` files with only `ci/versions.yml` varied:

| `version:` value | loaded as | result on `main` |
| --- | --- | --- |
| `"13.3.0"` | `str` | exit 0 (correct) |
| `13.3` | `float` | **`AttributeError: 'float' object has no attribute 'split'`** |
| `13` | `int` | **`AttributeError: 'int' object has no attribute 'split'`** |
| *(empty)* | `None` | **`AttributeError: 'NoneType' object has no attribute 'split'`** |
| `[13, 3]` | `list` | **`AttributeError: 'list' object has no attribute 'split'`** |
| `"13"` | `str` | **`ValueError: not enough values to unpack (expected at least 2, got 1)`** |

All five escape as an uncaught traceback out of a pre-commit hook, pointing the contributor at this script rather than at the line they edited. The `except (KeyError, TypeError)` above shows the intent was already to fail cleanly here.

## Fix

Add `parse_build_version`, which returns `(major, minor)` only for a `<major>.<minor>[.<patch>]` string of digits and `None` otherwise, and have `main` report the bad value with the same `return 2` shape as its neighbours:

```
error: cuda.build.version=13.3 in .../ci/versions.yml is not a '<major>.<minor>[.<patch>]'
version string. Quote the value in YAML so it is not loaded as a number (13.3 becomes a
float, 13 becomes an int).
```

The message names the quoting trap because that is how the value actually goes wrong.

No behavior change for well-formed input: the current `"13.3.0"` still yields `expected="13.3.*"` and `cuda_feature="cu13"`, and running the tool against the real repo still exits 0.

## Tests

This script had no tests. Added `ci/tools/tests/test_check_pixi_cuda_version.py`:

- `parse_build_version` over the accepted shapes (`13.3.0`, `12.9.1`, `13.3`, `13.3.0.1`) and every rejected one (float, int, `None`, list, `"13"`, `""`, `"13."`, `".3"`, `"cuda.13"`).
- End-to-end tests that build a temporary repo layout (`ci/versions.yml` + both `pixi.toml` files), monkeypatch the module's path constants, and assert `main()` returns 2 with the diagnostic — and returns 0 for a well-formed version.

The module imports PyYAML at module scope (the pre-commit hook supplies it via `additional_dependencies`), so the test module opens with `pytest.importorskip("yaml")` rather than failing collection where PyYAML is absent.

## Verification

Executed in full (pure Python, no GPU):

- The six-row table above was produced by driving the **`upstream/main`** `main()` directly with monkeypatched path constants; five of the six raise uncaught.
- With the fix, all of those return exit 2 with the new message, and `pytest ci/tools/tests` is **73 passed**.
- Running the tool against the real repo checkout still prints `OK: ...` and exits 0.
- Note on the usual "restore the file from `upstream/main` and watch the tests fail" check: because the fix introduces a new function, the test module cannot even import against `main` (`ImportError: cannot import name 'parse_build_version'`). The evidence that matters is the table above, taken against unmodified `upstream/main` code.
- `ruff check` / `ruff format --check` clean on both files; `toolshed/check_spdx.py` clean on the new file; `python -m py_compile` clean. Index verified clean before committing.

## Related

Independent of #2538, #2543 (both `check_mempool_hygiene.py`) and #2496 (`check_release_notes.py`) — different files, no overlap.

### DIFF
--- a/ci/tools/check_pixi_cuda_version.py
+++ b/ci/tools/check_pixi_cuda_version.py
@@ -16,6 +16,24 @@ VERSIONS_FILE_PATH = ROOT / "ci" / "versions.yml"
 PIXI_FILES = [ROOT / d / "pixi.toml" for d in ("cuda_bindings", "cuda_core")]
 
 
+def parse_build_version(build_version: object) -> tuple[str, str] | None:
+    """Split ``cuda.build.version`` into ``(major, minor)``, or ``None``.
+
+    Returns ``None`` for anything that is not a ``<major>.<minor>[.…]`` string
+    of digits. YAML makes this easy to get wrong: an unquoted ``13.3`` loads as
+    the float ``13.3`` and an unquoted ``13`` as the int ``13``, neither of
+    which has ``.split``. Without this check those -- and a quoted but
+    single-component ``"13"`` -- escaped as a raw traceback from a pre-commit
+    hook whose every other failure path returns a diagnostic exit code.
+    """
+    if not isinstance(build_version, str):
+        return None
+    parts = build_version.split(".")
+    if len(parts) < 2 or not all(part.isdigit() for part in parts[:2]):
+        return None
+    return parts[0], parts[1]
+
+
 def main() -> int:
     """Verify cuda_bindings/cuda_core pixi pins match ci/versions.yml."""
     if not VERSIONS_FILE_PATH.is_file():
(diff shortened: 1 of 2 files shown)


mdboom:
We don't need this level of testing for automation scripts.  They will tend to fail in other ways (which is what we want) if these sorts of bugs slip through.

PR #2550, opened 2026-08-09
fix(ci): report malformed CTK redistrib metadata instead of tracebacking

LeSingh1:
## Problem

`fetch_ctk_redistrib.main()` wraps its whole body in a deliberate handler so the tool reports problems rather than crashing:

```python
except (ValueError, KeyError, OSError, urllib.error.URLError, json.JSONDecodeError) as exc:
    print(f"ERROR: {exc}", file=sys.stderr)
    return 1
```

Several malformed-manifest shapes escape it, because the guards test for **absence** rather than **type**.

**1. String-valued top-level keys are treated as components.** Every real `redistrib_*.json` carries `"release_date"`, `"release_label"`, `"release_product"` at the top level, next to the component objects. `metadata.get(component)` returns a `str` for those, and the guard only rejects `None`:

```python
component_info = metadata.get(component)
if component_info is None:
    raise KeyError(f"unknown CTK component {component!r}")

subdir_info = component_info.get(ctk_subdir)   # AttributeError on a str
```

**2. A manifest that is valid JSON but not an object.** `.github/actions/fetch_ctk/action.yml:77` downloads it with `curl -LSs "$CTK_JSON_URL" -o "$CTK_JSON_FILE"` — no `--fail` — so an HTTP error page or redirect body is written to the file. If it parses as `null`, an array, or a string, the failure surfaces several frames later.

**3. A subdir entry that is a bare string** rather than an object raises the same `AttributeError` from a different line.

Reproduced through `main(argv)` with `--metadata-path` (no network), against a manifest shaped like the real thing:

| input | result on `main` |
| --- | --- |
| `--component cuda_nvcc` (valid) | `rc=0` (correct) |
| `--component not_a_component` | `rc=1`, `ERROR: "unknown CTK component..."` (correct) |
| `--component release_label` | **`AttributeError: 'str' object has no attribute 'get'`** |
| metadata is `null` | **`TypeError: argument of type 'NoneType' is not iterable`** |
| metadata is a JSON array | **`AttributeError: 'list' object has no attribute 'get'`** |
| subdir entry is a string | **`AttributeError: 'str' object has no attribute 'get'`** |

The contrast with the absent-component row is the point: the tool already knows how to report this class of problem, and these paths bypass it.

## Fix

- Validate at load time that the manifest is a JSON object, with a message naming the source (path or URL).
- Check that a component entry and its subdir entry are objects before reaching into them, with a message that says what was found instead.
- In `filter_components`, replace `ctk_subdir in metadata.get(resolved_component, {})` with an explicit `isinstance(..., dict)` check. On a string value that `in` silently degrades from a key lookup to a **substring test** — it happens to return `False` for the real keys today, but only by luck.

Valid manifests are unaffected: the component lookup, the subdir lookup, and the `relative_path` extraction all behave exactly as before.

## Tests

This tool had no tests. Added `ci/tools/tests/test_fetch_ctk_redistrib.py` (which is already run by the nightly tooling job), driving `main(argv)` with `--metadata-path` so nothing touches the network:

- a valid component resolves to its `relative_path`;
- each of the three string-valued release keys is rejected with `rc=1`;
- an absent component keeps its existing `unknown CTK component` message;
- a non-object subdir entry is reported;
- `null` / array / string manifests are reported for **both** subcommands;
- `filter-components` skips a string-valued top-level key with the existing "Skipping unsupported CTK component" notice and `rc=0`.

## Verification

Executed in full (pure Python, stdlib only, no GPU and no network):

```
# with the fix
pytest ci/tools/tests  ->  69 passed

# with ci/tools/fetch_ctk_redistrib.py restored from upstream/main
10 failed, 3 passed
```

`ruff check` / `ruff format --check` clean on both files; `toolshed/check_spdx.py` clean on the new file. Restore done with `cp` aside + `git show upstream/main:<path> >`; index verified clean before committing.

### DIFF
--- a/ci/tools/fetch_ctk_redistrib.py
+++ b/ci/tools/fetch_ctk_redistrib.py
@@ -70,12 +70,25 @@ def load_metadata(*, metadata_path: str | None, metadata_url: str | None) -> dic
         raise ValueError("exactly one of --metadata-path or --metadata-url is required")
 
     if metadata_path is not None:
-        return json.loads(Path(metadata_path).read_text(encoding="utf-8"))
+        return _as_metadata_object(json.loads(Path(metadata_path).read_text(encoding="utf-8")), metadata_path)
 
     assert metadata_url is not None
     metadata_url = validate_metadata_url(metadata_url)
     with urllib.request.urlopen(metadata_url) as response:  # noqa: S310 - scheme is restricted to https above
-        return json.load(response)
+        return _as_metadata_object(json.load(response), metadata_url)
+
+
+def _as_metadata_object(metadata: Any, source: str) -> dict[str, Any]:
+    """Reject JSON that parsed fine but is not a redistrib manifest.
+
+    The manifest is downloaded with ``curl -LSs`` (no ``--fail``), so an error
+    page or a redirect body lands in the file and may still be valid JSON --
+    just not an object. Without this the failure surfaces several frames later
+    as ``TypeError: argument of type 'NoneType' is not iterable``.
+    """
+    if not isinstance(metadata, dict):
+        raise ValueError(f"CTK redistrib metadata from {source} must be a JSON object, got {type(metadata).__name__}")
+    return metadata
 
 
 def resolve_component_name(metadata: dict[str, Any], component: str) -> str:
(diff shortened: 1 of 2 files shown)


mdboom:
Doesn't seem worth the effort for CI automation scripts.

Issue #2077, opened 2026-05-13
EPIC: Tests that never run in CI


mdboom:
#1969 adds a step to CI that will analyze the logs to find all tests that are skipped in our CI under all configurations.  The complete list of tests (as of today) is copy-pasted below.

This issue is to start a discussion about what to do about them.

### Removal

I am not advocating a rush to remove them -- they often do run when QA runs them on different platforms than we have in CI.  However, some of these may be low-value and we could consider removing on a case-by-case basis.  Bitrotting of tests is as much a concern as bitrotting of production code.

### Annotation

I think it would be helpful to add a comment to all of these tests that they are not covered in CI.  That would be a helpful signal when editing a test, or seeing a failure in QA, that it is not well-covered.  An agent could easily take this list and create all those comments -- we can remove the comments if any of the tests are reworked to run in CI.  The existing tool that analyses logs for skipped tests could surface the fact that a test has an annotation that it always skips when that is no longer the case.

### Hardware/platform coverage

Does this list point at any hardware that would help us get better coverage?  As we determine /what kinds/ of hardware to add, this may be helpful in determining what would provide the most value.

### Software coverage

I think some of these aren't covered because they require a dependency, such as pytorch.  That's a heavy dependency we probably don't want to test with /everywhere/, but maybe we could add it in a few configurations to get more test coverage.

# Always skipped tests

- [ ] `cuda_bindings/tests/nvml/test_device.py::test_device_vgpu_get_heterogeneous_mode`
- [ ] `cuda_bindings/tests/nvml/test_device.py::test_get_nv_link_supported_bw_modes`
- [ ] `cuda_bindings/tests/nvml/test_device.py::test_read_write_prm`
- [ ] `cuda_bindings/tests/nvml/test_device.py::test_set_power_management_limit`
- [ ] `cuda_bindings/tests/nvml/test_pci.py::test_bridge_chip_info`
- [ ] `cuda_bindings/tests/nvml/test_pynvml.py::test_device_get_attributes`
- [ ] `cuda_bindings/tests/nvml/test_pynvml.py::test_device_get_cpu_affinity_within_scope[<AffinityScope.NODE:`
- [ ] `cuda_bindings/tests/nvml/test_pynvml.py::test_device_get_cpu_affinity_within_scope[<AffinityScope.SOCKET:`
- [ ] `cuda_bindings/tests/nvml/test_pynvml.py::test_device_get_memory_affinity[<AffinityScope.NODE:`
[...]

PR #2565, opened 2026-08-09
Stop _gl_context from converting test failures into skips

LeSingh1:
`_gl_context()` yields from inside a `try` whose handler turns everything into a skip:

```python
    try:
        ...create context and texture...
        yield int(tex_id.value), int(target)

    except Exception as e:
        # Convert any pyglet/GL creation failure into a clean skip
        pytest.skip(f"Could not create GL context/texture: {type(e).__name__}: {e}")
```

`@contextmanager` re-raises the with-body's exception **at the yield point**, so
`except Exception` also catches failures from the *test body*. `AssertionError` is an
`Exception`; pytest's `Skipped` is a `BaseException`, so the replacement skip escapes
cleanly and the run is reported as SKIPPED.

That means neither assertion in `test_cuda_gl_register_image_smoketest` can fail the suite:

```python
        assert name in acceptable, f"cudaGraphicsGLRegisterImage returned {name}"
        if name == "cudaSuccess":
            assert int(resource) != 0
```

A genuinely wrong `cudaGraphicsGLRegisterImage` return is reported as
`Could not create GL context/texture: AssertionError: ...`. The handler's own comment says
*"creation failure"*, which is what it was meant to cover.

Demonstrated with a minimal standalone reproduction of the same shape:

```
test_yield_in_try.py::test_body_failure_under_main_shape   SKIPPED
    Could not create GL context/texture: AssertionError: this assertion should FAIL the test
test_yield_in_try.py::test_body_failure_under_fixed_shape  FAILED
    AssertionError: this assertion should FAIL the test
```

Both tests in this module are on the always-skipped list in #2077, so this has never been
visible in CI.

### Fix

Move the context/texture creation into `_create_gl_texture()` so the handler wraps only
that, and keep the `yield` outside it. Cleanup stays in the outer `finally`, so a
partially built context is still torn down on the skip path. `tex_id` is now initialized,
which also removes an `UnboundLocalError` that the `finally` block was silently swallowing
when creation failed before the texture existed.

`pytest.importorskip` and the no-EGL `pytest.skip` inside `_create_gl_texture` still
propagate correctly — `Skipped` is a `BaseException`, so `except Exception` does not catch
them.

### Test

`test_gl_context_lets_body_failures_fail` stubs `_create_gl_texture()` and asserts that an
`AssertionError` raised inside the `with` body propagates. **No GL context, no display, no
GPU, no pyglet** — so unlike the two smoketests it actually runs in CI. I verified it
passes against the restructured contextmanager and that the pre-existing shape converts
the same failure into a skip.

`ruff check` and `ruff format --check` are clean.

### DIFF
--- a/cuda_bindings/tests/test_graphics_apis.py
+++ b/cuda_bindings/tests/test_graphics_apis.py
@@ -12,15 +12,14 @@ import pytest
 from cuda.bindings import runtime as cudart
 
 
-@contextlib.contextmanager
-def _gl_context():
-    """
-    Yield a (tex_id, tex_target) with a current GL context.
+def _create_gl_texture():
+    """Create a hidden/headless GL context plus a tiny texture to register.
+
+    Returns ``(win, tex_id, target)``. Raises on any pyglet/GL failure.
     Tries:
       1) Windows: hidden WGL window (no EGL)
       2) Linux with DISPLAY/wayland: hidden window
       3) Linux headless: EGL headless if available
-    Skips if none work.
     """
     pyglet = pytest.importorskip("pyglet")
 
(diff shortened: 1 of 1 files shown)


LeSingh1:
Closing — this is already fixed on `main`.

`_gl_context` now splits the two concerns the way this PR was asking for: `_open_gl_window()` sits in a `try` that skips only when `is_gl_context_unavailable(e)` and re-raises otherwise, and the `yield` is in a plain `try/finally` with no `except`, so a genuine test failure inside the `with` block propagates instead of being reported as a skip. The old blanket `pytest.skip(f"Could not create GL context/texture: ...")` handler is gone.

That covers the problem, and the current shape is better than what I proposed since it also distinguishes an unavailable context from any other creation error. Nothing left for this one to do.

PR #2571, opened 2026-08-09
cuda.core: emit -numba-debug to libNVVM, not --numba-debug

### LINKED ISSUE #2570: [BUG]: ProgramOptions(numba_debug=True) always fails on the NVVM backend
### Is this a duplicate?

No.

### Type of Bug

Runtime Error

### Component

cuda.core

### Describe the bug

`ProgramOptions(numba_debug=True)` never works on the NVVM backend. `cuda.core` emits the option to libNVVM as `--numba-debug`, and libNVVM accepts only the single-dashed `-numba-debug`, so every such compile raises `NVVM_ERROR_INVALID_OPTION`. libNVVM rejects the double-dashed form of every option, not just this one, so no toolkit version accepts the current spelling. The NVRTC backend tolerates both spellings and is unaffected.

This also hides itself: `test_nvvm_program_numba_debug` is gated on a probe that asks libNVVM about the same wrong spelling, so the test skips everywhere. It is one of the entries in #2077. The skip reason blames CTK 13.2, but the option is rejected on CTK 13.3 as well, so the toolkit version is not the cause.

Present in released `cuda-core` 1.1.0 and 1.1.1, and on `main`. Added in #2158, closing #1287.

### How to Reproduce

```python
from cuda.core import Device, Program, ProgramOptions
from cuda.core._program import _get_nvvm_module

nvvm = _get_nvvm_module()
ir = 'target triple = "nvptx64-unknown-cuda"\ntarget datalayout = "e-p:64:64:64-i64:64:64-n16:32:64"\ndefine void @simple() {\nentry:\n  ret void\n}\n!nvvmir.version = !{!1}\n!1 = !{i32 %d, i32 %d, i32 %d, i32 %d}\n' % nvvm.ir_version()

Device(0).set_current()
opts = ProgramOptions(arch="sm_89", debug=True, numba_debug=True)
print(opts.as_bytes("nvvm"))
Program(ir, "nvvm", opts).compile("ptx")
```

```
[b'-arch=compute_89', b'-g', b'--numba-debug']
  ...
cuda.bindings.nvvm.nvvmError: ERROR_INVALID_OPTION (7)
NVVM program log: libnvvm : error: --numba-debug is an unsupported option
```

Asking libNVVM directly which spelling it accepts:

```console
$ python -c "from cuda.bindings.utils import check_nvvm_compiler_options as c; print([(o, c([o])) for o in [\"--numba-debug\", \"-numba-debug\"]])"
[('--numba-debug', False), ('-numba-debug', True)]
```

### Expected behavior

`ProgramOptions(numba_debug=True)` should compile on the NVVM backend. The option should be emitted as `-numba-debug`, which is also what numba-cuda emits for the NVVM path.

### Operating System

Ubuntu 24.04 on WSL2, CUDA 13.3

fedonman:
## Summary

Closes #2570.

libNVVM accepts only single-dashed options, but `_prepare_nvvm_options_impl` emitted `--numba-debug`, so every `ProgramOptions(numba_debug=True)` compile on the NVVM backend failed with `NVVM_ERROR_INVALID_OPTION`. numba-cuda emits `-numba-debug` for the same path. The NVRTC backend tolerates both spellings and is left alone.

The guarding test never caught this because it probed libNVVM with the same wrong spelling, so it skipped on every configuration, including CUDA 13.3. It is one of the entries in #2077.

## Changes

- `_program.pyx`: emit `-numba-debug` in `_prepare_nvvm_options_impl`. The NVRTC call site is unchanged.
- `tests/test_program.py`: probe and assert the emitted spelling, add an assertion that the double-dashed form is absent, and drop the skip reason's claim that the option needs CTK 13.2 (libNVVM rejects the double-dashed form at any version).
- `docs/source/release/1.2.0-notes.rst`: release note. The wrong spelling is in released 1.1.0 and 1.1.1.

## Verification

`test_nvvm_program_numba_debug` previously skipped everywhere. It now runs and passes, and it fails without the production change:

```console
$ git stash push cuda_core/cuda/core/_program.pyx   # production only
$ python -m pytest tests/test_program.py -k numba_debug -v
tests/test_program.py::test_nvvm_program_options_as_bytes_numba_debug FAILED [ 50%]
tests/test_program.py::test_nvvm_program_numba_debug FAILED              [100%]
cuda.bindings.nvvm.nvvmError: ERROR_INVALID_OPTION (7)
NVVM program log: libnvvm : error: --numba-debug is an unsupported option
```

With the change:

```console
$ python -m pytest tests/test_program.py -k numba_debug -v
tests/test_program.py::test_nvvm_program_options_as_bytes_numba_debug PASSED [ 50%]
tests/test_program.py::test_nvvm_program_numba_debug PASSED              [100%]
```

Whole module, before and after:

```console
$ python -m pytest tests/test_program.py -q   # upstream/main
110 passed, 2 skipped in 2.01s
$ python -m pytest tests/test_program.py -q   # this branch
111 passed, 1 skipped in 2.61s
```

### DIFF
--- a/cuda_core/cuda/core/_program.pyx
+++ b/cuda_core/cuda/core/_program.pyx
@@ -1233,7 +1233,7 @@ cdef inline object _prepare_nvvm_options_impl(object opts, bint as_bytes):
     if opts.debug is not None and opts.debug:
         options.append("-g")
     if opts.numba_debug:
-        options.append("--numba-debug")
+        options.append("-numba-debug")
     if opts.device_code_optimize is False:
         options.append("-opt=0")
     elif opts.device_code_optimize is True:
--- a/cuda_core/docs/source/release/1.2.0-notes.rst
+++ b/cuda_core/docs/source/release/1.2.0-notes.rst
@@ -73,6 +73,13 @@ Fixes and enhancements
   Windows, both ``ctypes.CFUNCTYPE`` and ``ctypes.WINFUNCTYPE`` are accepted.
   (`#2439 <https://github.com/NVIDIA/cuda-python/issues/2439>`__)
 
+- ``ProgramOptions(numba_debug=True)`` now works on the NVVM backend. The
+  option was emitted as ``--numba-debug``, and libNVVM accepts only the
+  single-dashed ``-numba-debug``, so every such compilation failed with
+  ``NVVM_ERROR_INVALID_OPTION``. The NVRTC backend, which tolerates both
+  spellings, was unaffected.
+  (`#2570 <https://github.com/NVIDIA/cuda-python/issues/2570>`__)
+
 Deprecation Notices
 -------------------
 
--- a/cuda_core/tests/test_program.py
+++ b/cuda_core/tests/test_program.py
@@ -93,12 +93,12 @@ def _check_nvvm_arch(arch: str) -> bool:
 
 
 def _check_nvvm_supports_numba_debug() -> bool:
-    """Check if the installed libNVVM recognizes --numba-debug (CTK 13.2+)."""
+    """Check if the installed libNVVM recognizes -numba-debug."""
     if not _has_check_nvvm_compiler_options():
         return False
     from cuda.bindings.utils import check_nvvm_compiler_options
 
-    return check_nvvm_compiler_options(["--numba-debug"])
+    return check_nvvm_compiler_options(["-numba-debug"])
 
 
 @pytest.fixture(scope="session")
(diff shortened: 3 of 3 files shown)


cuda_core/cuda/core/_program.pyx
```diff
@@ -1233,7 +1233,7 @@ cdef inline object _prepare_nvvm_options_impl(object opts, bint as_bytes):
     if opts.debug is not None and opts.debug:
         options.append("-g")
     if opts.numba_debug:
-        options.append("--numba-debug")
+        options.append("-numba-debug")
```
lijinf2, line 1236: My understanding of the usual CLI convention is that `--` pairs with a full option name (e.g. `--numba-debug`), while `-` pairs with a short abbreviation (e.g. `-n`). Emitting a long name with a single dash (`-numba-debug`) goes against that convention.

Is there a reason we use a single dash "-numba-debug" here? @isVoid

isVoid: No particular reason here. `-numba-debug` is undocumented so even it's exposed in `cuda.core` it's subject to change at any time. It's only to support numba's debug mode.

@fedonman could you share what do you use it for?

fedonman: Nothing of my own to be honest, I came at this from #2077.

`test_nvvm_program_numba_debug` was on the list of tests that skip on every configuration. When I looked at why, the skip guard was probing libNVVM with the same `--numba-debug` spelling that `_prepare_nvvm_options_impl` emits, so it skipped everywhere and hid the fact that the option never worked on the NVVM backend.

So my interest is just consistency of the public surface: `ProgramOptions(numba_debug=True)` is accepted by the API today and raises `NVVM_ERROR_INVALID_OPTION` 100% of the time on that backend. Either it should work (this PR, matching what numba-cuda emits) or it should be deprecated/removed

rparolin:
@fedonman Thank you for submitting a PR. Please accept my apology but I'm going to close this PR. I submitted a similar fix for https://github.com/NVIDIA/cuda-python/issues/2570 because there was no one assigned to the ticket.  In future, please ensure that you are assigned to an issue by a member of the CUDA Python team before submitting a PR. That prevents others from doing redundant work.

Ping me if you have any questions. Again, thanks for the PR and apologies for closing.

fedonman:
@rparolin No worries at all! I opened the issue and submitted the PR directly. Thanks for clarifying the process. I’ll request assignment before working on future issues.

PR #2589, opened 2026-08-10
feat(security): onboard security-suite (secret + CodeQL) scanning.

gmanal:
## Description

Onboard NVIDIA security scanning to cuda-python using the centrally maintained NVIDIA/security-workflows surfaces. This uses the **security suite** — a single pinned reusable workflow that fans out to the individual scanners — rather than one caller per scan. Two complementary controls:

* **Local (advisory):** `secret-scan-trufflehog` pre-commit hook — catches credentials before commit.
* **CI (enforcement):** the security suite — server-side Pulse secret scan (blocks on verified secrets) and CodeQL SAST, both explicitly enabled.

Every scan in the suite is opt-in, so this repository runs exactly the two scans named here; scanners added upstream later will not switch on by themselves.

## Changes

* Add `.github/workflows/security-suite.yml` calling the centralized security suite, on `main`, `ctk-next`, and copy-pr-bot `pull-request/*` branches; the secret scan runs on `nv-cpu-general` (onboarded for this repo in the enterprise runner config) via the `linux-amd64-cpu4` label.
* Enable both scans explicitly — `enable-secret-scan: true` and `enable-sast-scan: true` — so what runs here is visible in this file rather than inherited from upstream defaults.
* Remove `.github/workflows/codeql.yml`, superseded by the suite's SAST scan. Both publish code scanning results under the category `/language:python`, so running them together would put two analyses on every commit that overwrite each other's alerts. The suite performs the same analysis: `python`, `build-mode: none`, `security-extended` queries, on `ubuntu-latest`.
* Add the `secret-scan-trufflehog` pre-commit hook (pinned to `security-workflows` `v0.3.0`), skipped on hosted pre-commit.ci which lacks the trufflehog binary.
* Set `secret-failure-policy: unverified` explicitly (fail on verified/live secrets, warn on unverified); fails closed on scanner/infra errors.
* Grant the union of the permissions the reusable workflow declares — including scans this repo does not enable — because GitHub validates the caller's grant before evaluating each scan's condition.
* Document local setup in CONTRIBUTING.md.

## Checklist

* New or existing tests cover these changes.
* The documentation is up to date with these changes.


gmanal:
@mdboom @rparolin @kkraus14 @leofang @jrhemstad — this replaces #2405, which I'm closing. Re-requesting your eyes here since I can't move the review requests across.

**What is the same as #2405** (already approved by @mdboom and @rparolin): the `secret-scan-trufflehog` pre-commit hook, and the server-side Pulse secret scan with `failure_policy: unverified`. The hook `rev` is a frozen commit SHA per @mdboom's review comment on #2405.

**What is new here:**

1. **One pinned reference instead of one caller per scan.** This calls the `security-workflows` *security suite*, which fans out to the scans this repo enables. Every scan is opt-in, so scanners added upstream later will not switch themselves on — adding one is a deliberate edit to this file.
2. **CodeQL SAST is enabled** (`sast-languages: '["python"]'`).
3. **`.github/workflows/codeql.yml` is removed.** This is the one change that is not purely additive, so it deserves the most scrutiny. Both it and the suite's SAST publish code scanning results under the category `/language:python`; running both would put two analyses on every commit that overwrite each other's alerts. The suite performs the same analysis the retired workflow did — `python`, `build-mode: none`, `security-extended`, on `ubuntu-latest` — so coverage is unchanged, it is just centrally maintained now. C/C++ coverage for `cuda_bindings` is **not** included here; that needs a build mode and should be a separate decision.
4. **Pinned to `security-workflows` v0.3.0** for both surfaces.

**@kkraus14 — your Git Bash question on #2405 is resolved.** The hook moved to `language: python` and installs the pinned, SHA-256-verified binary into pre-commit's own environment (NVIDIA/security-workflows#14, shipped in v0.3.0). No Git Bash, no curl, no extra developer dependencies; verified on Linux, macOS, and Windows across x86_64 and arm64. The Git Bash line is gone from CONTRIBUTING.md in this PR.

v0.3.0 also fixes a false positive worth knowing about: TruffleHog's Lob detector treats `test_` plus 35 characters as an API key, so it flagged 49 of this repo's own pytest function names as *verified* secrets. That detector is now excluded upstream. I ran the pinned hook across the full tree locally and it passes clean.

.github/workflows/security-suite.yml
```diff
@@ -0,0 +1,47 @@
...
+      enable-secret-scan: true
+      enable-sast-scan: true
+      secret-runs-on: linux-amd64-cpu4
+      # Set failure_policy explicitly so enforcement can't drift with upstream defaults.
+      #   unverified — fail on verified/live secrets (183); warn on unverified (185)  [default]
+      #   strict     — fail on any finding (verified or unverified)
+      #   all        — warn only; never fail the job on findings
+      secret-failure-policy: unverified
+      # Same analysis the retired codeql.yml performed: python, build-mode none, security-extended.
+      sast-languages: '["python"]'
```
mdboom, line 47: We have a small amount of C++ code as well.  Is there a way to add it to the scan?

mdboom (approved):

.github/workflows/security-suite.yml
gmanal: Let's get the Python scan in for now. C/C++ likely needs a build step — the hand-written
C++ here is compiled through the Cython `build_hooks` backend rather than cmake/make — and
I want to understand that properly before wiring it up. It publishes under `/language:c-cpp`, so it can be added later without touching the `python` analysis here.

PR #2590, opened 2026-08-10
fix(core): use the NVVM numba debug option spelling

### LINKED ISSUE #2570: [BUG]: ProgramOptions(numba_debug=True) always fails on the NVVM backend

aryanputta:
Fixes #2570.

libNVVM accepts the single-dash `-numba-debug` option, while cuda-core emitted `--numba-debug` for the NVVM backend. The latter is rejected as `NVVM_ERROR_INVALID_OPTION`.

This change:
- emits `-numba-debug` for NVVM only;
- updates the capability probe so the regression test no longer skips because it probes the invalid spelling;
- updates the option serialization and compile-path tests;
- adds a release note.

Validation:
- Ruff check and format passed.
- Cython lint passed.
- Focused pytest collection was blocked on macOS because the CUDA test helper requires Linux `libc.so.6`.
- Lychee was blocked by sandbox network access to existing external links.

### DIFF
--- a/cuda_core/cuda/core/_program.pyx
+++ b/cuda_core/cuda/core/_program.pyx
@@ -1216,7 +1216,7 @@ cdef inline list _prepare_nvrtc_options_impl(object opts):
             f"--instantiate-templates-in-pch={_handle_boolean_option(opts.instantiate_templates_in_pch)}"
         )
     if opts.numba_debug:
-        options.append("--numba-debug")
+        options.append("-numba-debug")
     return [o.encode() for o in options]
 
 
--- a/cuda_core/docs/source/release/1.2.0-notes.rst
+++ b/cuda_core/docs/source/release/1.2.0-notes.rst
@@ -9,6 +9,9 @@
 Fixes and enhancements
 ----------------------
 
+- Fixed ``ProgramOptions(numba_debug=True)`` for the NVVM backend by emitting
+  the single-dash ``-numba-debug`` option accepted by libNVVM.
+
 - Graph node resources are now retained independently across graph clones,
   executable graphs, updates, node deletion, and in-flight launches. Previously,
   modifying a graph definition could release resources still used by an
--- a/cuda_core/tests/test_program.py
+++ b/cuda_core/tests/test_program.py
@@ -93,12 +93,12 @@ def _check_nvvm_arch(arch: str) -> bool:
 
 
 def _check_nvvm_supports_numba_debug() -> bool:
-    """Check if the installed libNVVM recognizes --numba-debug (CTK 13.2+)."""
+    """Check if the installed libNVVM recognizes -numba-debug."""
     if not _has_check_nvvm_compiler_options():
         return False
     from cuda.bindings.utils import check_nvvm_compiler_options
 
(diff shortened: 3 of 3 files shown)


lijinf2:
Thanks for the contribution!

This looks like a duplicate of [#2571](https://github.com/NVIDIA/cuda-python/pull/2571), which was opened by the issue author to address the same bug in [#2570](https://github.com/NVIDIA/cuda-python/issues/2570). Both PRs fix the NVVM backend to emit -numba-debug instead of --numba-debug.

lijinf2:
@aryanputta Heads up for future contributions: please check the issue’s linked PRs / Development section first, in case a fix is already in progress.

Thanks again for taking the time to put this together — your help is appreciated!

PR #2569, opened 2026-08-09
cuda.core: reject register() on a non-IPC memory resource

fedonman:
## Summary

Closes #2568.

`DeviceMemoryResource.register()` and `PinnedMemoryResource.register()` terminated the process with a segmentation fault when the memory resource did not have IPC enabled, including the default resource from `Device().memory_resource`. Both route through `MP_register`, which read `self._ipc_data` unconditionally although it is `None` unless IPC is enabled.

## Changes

- `_ipc.pyx`: check `is_ipc_enabled` in `MP_register` and raise `RuntimeError("Memory resource is not IPC-enabled")`, the wording the `allocation_handle` property already uses. The check runs before the registry insertion, so a rejected registration no longer leaves an entry behind.
- `tests/memory_ipc/test_errors.py`: regression test covering the raise and the registry.
- `docs/source/release/1.2.0-notes.rst`: release note. The defect is present in released 1.0.1, 1.1.0 and 1.1.1, so this is user-visible rather than confined to the current cycle.

## Verification

Without the production change the new test does not merely fail, it takes the interpreter down (`exit=139`, SIGSEGV). With the change:

```console
$ python -m pytest tests/memory_ipc/test_errors.py -k test_register_rejects_non_ipc -v
tests/memory_ipc/test_errors.py::test_register_rejects_non_ipc_memory_resource PASSED [100%]
======================= 1 passed, 14 deselected in 0.28s =======================
```

No regressions in the surrounding suites:

```console
$ python -m pytest tests/test_memory.py tests/memory_ipc/ -q
192 passed, 133 skipped in 4.48s
```

PR #2591, opened 2026-08-10
fix(core): reject IPC registration without IPC state

### LINKED ISSUE #2568: [BUG]: DeviceMemoryResource.register() segfaults when IPC is not enabled
### Is this a duplicate?

No.

### Type of Bug

Runtime Error

### Component

cuda.core

### Describe the bug

`DeviceMemoryResource.register()` and `PinnedMemoryResource.register()` crash the interpreter with SIGSEGV when the memory resource does not have IPC enabled, including the default resource returned by `Device().memory_resource`. No exception is raised, so the call cannot be guarded with `try`.

The cause is an unconditional read of `self._ipc_data` in `MP_register` (`cuda_core/cuda/core/_memory/_ipc.pyx`), which is `None` unless IPC is enabled.

Present in released `cuda-core` 1.0.1, 1.1.0 and 1.1.1, and on `main`.

### How to Reproduce

```python
import uuid
from cuda.core import Device, DeviceMemoryResource

d = Device()
d.set_current()
DeviceMemoryResource(d).register(uuid.uuid4())
```

Run from a parent process, the child dies with no output at all:

```
stdout: ''
stderr: ''
returncode: -11 -> SIGSEGV
```

### Expected behavior

`register()` should raise `RuntimeError("Memory resource is not IPC-enabled")`, matching the `allocation_handle` property on the same class.

### Operating System

Ubuntu 24.04 on WSL2

### Output from running `nvidia-smi` on the command line:

```
Sun Aug  9 11:22:15 2026       
| NVIDIA-SMI 610.43.02              KMD Version: 610.62        CUDA UMD Version: 13.3     |
|   0  NVIDIA GeForce RTX 4060 Ti     On  |   00000000:01:00.0  On |                  N/A |
|  0%   41C    P8              5W /  160W |    2399MiB /   8188MiB |      8%      Default |
```

aryanputta:
Fixes #2568.

`DeviceMemoryResource.register()` and `PinnedMemoryResource.register()` can reach `MP_register` without IPC enabled. The current implementation dereferences `self._ipc_data`, causing a process-level segmentation fault instead of a Python exception.

This change:
- raises `RuntimeError("Memory resource is not IPC-enabled")` before the dereference;
- preserves the existing-resource fast path;
- adds coverage for both device and pinned memory resources;
- adds a release note.

Validation:
- Ruff check and format passed.
- Cython lint passed.
- Focused pytest collection was blocked on macOS because the CUDA test helper requires Linux `libc.so.6`.
- Lychee was blocked by sandbox network access to existing external links.

### DIFF
--- a/cuda_core/cuda/core/_memory/_ipc.pyx
+++ b/cuda_core/cuda/core/_memory/_ipc.pyx
@@ -268,6 +268,8 @@ cdef _MemPool MP_register(_MemPool self, uuid):
     existing = registry.get(uuid)
     if existing is not None:
         return existing
+    if self._ipc_data is None:
+        raise RuntimeError("Memory resource is not IPC-enabled")
     assert self.uuid is None or self.uuid == uuid
     registry[uuid] = self
     self._ipc_data._alloc_handle._uuid = uuid
--- a/cuda_core/docs/source/release/1.2.0-notes.rst
+++ b/cuda_core/docs/source/release/1.2.0-notes.rst
@@ -9,6 +9,10 @@
 Fixes and enhancements
 ----------------------
 
+- ``DeviceMemoryResource.register()`` and ``PinnedMemoryResource.register()``
+  now raise ``RuntimeError`` when called on a resource without IPC enabled,
+  instead of dereferencing missing IPC state.
+
 - Graph node resources are now retained independently across graph clones,
   executable graphs, updates, node deletion, and in-flight launches. Previously,
   modifying a graph definition could release resources still used by an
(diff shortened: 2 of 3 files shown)


lijinf2:
Thanks for the contribution!

This looks like a duplicate of [#2569](https://github.com/NVIDIA/cuda-python/pull/2569), which was opened by the issue author to address the same bug in [#2568](https://github.com/NVIDIA/cuda-python/issues/2568). Both PRs raise RuntimeError("Memory resource is not IPC-enabled") when register() is called on a non-IPC memory resource.

PR #2593, opened 2026-08-10
cuda.core: Add copy_batch to cuda.core.utils

### LINKED ISSUE #1333: Support batched memory movement
P0:
- [cudaMemcpyBatchAsync](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1g6126baf5d881835091c59e48890d6854)

P1:
- [cudaMemDiscardBatchAsync](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1g5acb7cea41bb9115f10568cc8176f51f) 
- [cudaMemPrefetchBatchAsync](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1ge4fa23c9a26c6e5e702cbe35d001d589)
- [cudaMemDiscardAndPrefetchBatchAsync](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1g0f6d2e27d8f00ee78c5d814f45500605)

https://docs.nvidia.com/cuda/cuda-programming-guide/03-advanced/advanced-host-programming.html#batched-memory-transfers

juenglin:
## Summary
Adds `copy_batch` to `cuda.core.utils`, wrapping `cuMemcpyBatchAsync` so a set of
transfers is issued in a single driver call. This is the P0 item of #1333; the P1
items (prefetch / discard / discard-prefetch) landed in #1775. Closes #1333.
Batched-only, mirroring the `*_batch` functions from #1775: a lone `Buffer` is
rejected in favour of `Buffer.copy_to` / `Buffer.copy_from`.
## Public API
```python
from cuda.core import Device, Host
from cuda.core.utils import CopyOptions, MemcpySrcAccessOrder, copy_batch
# One option applies to every copy...
copy_batch(stream, srcs, dsts, options=CopyOptions(src_access_order=MemcpySrcAccessOrder.STREAM))
# ...or one per copy, paired by index.
copy_batch(stream, srcs, dsts, options=[opt_a, opt_b, opt_c])
# Location hints, for managed-memory operands.
copy_batch(stream, srcs, dsts, options=CopyOptions(
    src_access_order=MemcpySrcAccessOrder.STREAM,
    src_location_hint=Device(0),
    dst_location_hint=Host(),
))
```
New exports from `cuda.core.utils`: `copy_batch`, `CopyOptions`,
`MemcpySrcAccessOrder`, `MemcpyOverlapMode`. Nothing is added to top-level
`cuda.core`.
`CopyOptions` is a frozen dataclass mapping onto `CUmemcpyAttributes`:
`src_access_order`, `src_location_hint`, `dst_location_hint`, `overlap_mode`.
`options=None` selects stream-ordered defaults.
## Behaviour
- Source and destination sizes must match, that is `src.size` must equal `dst.size`.
  Partial-buffer copies are out of scope.
- A scalar `CopyOptions` broadcasts to every copy; a sequence pairs by index and
  must match `len(srcs)`, otherwise `ValueError`.
- Options are run-length encoded into the driver's `attrs` / `attrsIdxs` pair, so
  a broadcast reaches the driver as `numAttrs == 1` instead of being repeated per
  copy.
- On a CUDA 12 build of `cuda.bindings` or when using an older driver, `copy_batch` is emulated by a sequence of `cuMemcpyAsync` calls, consistent with `prefetch_batch`.
- Not capturable into a graph. Use `GraphNode.memcpy` for graph copies.
- `overlap_mode="prefer_overlap_with_compute"` is not supported by all devices.
  
## Refactor to existing files
`_coerce_batch_buffers` was duplicated between the new `_copy_ops.pyx` and `_managed_memory_ops.pyx`, so it now lives once as Buffer_coerce_batch in `_buffer.pyx` / `_buffer.pxd`, parameterized by the per-buffer API to suggest. That accounts for the changes to `_buffer.*` and `_managed_memory_ops.pyx` (no behaviour changes for the #1775 functions).

## Relation to the #1775 batched-API contract
Follows [the contract](https://github.com/NVIDIA/cuda-python/pull/1775#issuecomment-4355502334) on pairing, scalar broadcast, `ValueError` on length mismatch, and options as a frozen dataclass. Three deliberate deviations, which match what #1775 *shipped*
rather than what that comment described:
- No unified `copy()` accepting one-or-many. #1775 shipped batch-only free
  functions with instance methods for N=1; this mirrors that split.
- Parallel `srcs` / `dsts` sequences rather than a sequence of pairs, matching
  `prefetch_batch(stream, buffers, locations)`.
- A single `stream` argument: `cuMemcpyBatchAsync` takes one `CUstream`, so there
  is no per-copy stream array to mirror.

## Tests
`tests/memory/test_copy_batch.py` covers data movement — H2D, D2H, D2D, mixed
sizes, agreement with sequential `Buffer.copy_to`, stream ordering, and the
graph-capture rejection. `tests/memory/test_copy_batch_options.py` covers the
options surface, the attribute run-length encoding, and every validation path.
Shared fixtures live in `tests/memory/conftest.py`. The new
`examples/batched_memcpy.py` is exercised by the existing example-test glob.


Andy-Jost:
1. Reject `LEGACY_DEFAULT_STREAM` in `copy_batch` before dispatch, so both paths behave alike. The native call fails with an unhelpful driver error, and the CUDA 12 fallback accepts the stream and succeeds.

2. State the ordering contract in the docstring, and say that the fallback's ordering is an artifact. The driver may run batch items concurrently and in any order, so a batch must not contain copies that read and write the same bytes. Detecting this at runtime is impractical; documentation plus a release-note warning is enough.

3. Correct graph handling in three places. Allow `GraphBuilder`, because CUDA 13.1 and later support capture of the batch call. Reject `DURING_API_CALL` during capture, because the driver does not allow it. The current `GraphBuilder` rejection is also incomplete, because a caller can pass the builder's active stream and reach the same path.

4. Remove or version-gate the discrete-GPU overlap warning. CUDA 13.1 added non-Tegra support for `PREFER_OVERLAP_WITH_COMPUTE`. The check should also test the stream device instead of the current device.

5. State the source-access contract for each value, not only its performance effect. `STREAM` keeps source reads in stream order. `DURING_API_CALL` lets the driver read the source out of stream order, but all reads finish before the call returns; no earlier stream work may touch the source. `ANY` lets the driver read the source after the call returns, so the caller must keep the source unchanged until the copy completes in stream order.

6. Document the limits of location hints. State which memory types honor them, and state that they neither prefetch memory nor set persistent advice.

7. Make the NUMA tests capability-aware. Do not assume that NUMA node 0 is valid, and cover `Host.numa_current()`.

8. Complete the eager enum validation in `CopyOptions.__post_init__`. Reject values that are neither the applicable enum nor a string, as newer option classes do.

9. Make the public annotations resolvable at runtime. Import `Device` and `Host` outside `TYPE_CHECKING` if no import cycle results, then test `typing.get_type_hints(CopyOptions)`.

10. Correct the specific claims in `cuda_core/docs/source/api.rst` and `cuda_core/docs/source/release/1.2.0-notes.rst`. Remove the statement that graph capture is unavailable, and replace "semantically equivalent" for the CUDA 12 fallback with the ordering and stream limits from items 1 and 2.

Items 6, 8, and 9 are minor and can follow in a later change.

Andy-Jost (changes requested):
Looks good in general. Requested changes are detailed in a separate comment.

cuda_core/cuda/core/_memory/_copy_ops.pyx
```diff
@@ -209,6 +213,15 @@ def copy_batch(
             "pass an explicit stream"
         )
 
+    cdef cydriver.CUstreamCaptureStatus _cap_status
+    HANDLE_RETURN(cydriver.cuStreamGetCaptureInfo(as_cu(s._h_stream), &_cap_status,
```
juenglin, line 217: Extra driver call here.

The alternative is to attempt the batched memcpy call and raise the `TypeError` when `CUDA_ERROR_STREAM_CAPTURE_UNSUPPORTED` is returned.

juenglin:
> 3. Correct graph handling in three places. Allow GraphBuilder, because CUDA 13.1 and later support capture of the batch call. Reject DURING_API_CALL during capture, because the driver does not allow it. 

I could not confirm this. Perhaps this was a misunderstanding of another API change in CUDA 13.1?

cuda_core/cuda/core/_memory/_copy_ops.pyx
Andy-Jost: I think it's preferable to check ahead of time, like you do here.

Andy-Jost:
> > 3. Correct graph handling in three places. Allow GraphBuilder, because CUDA 13.1 and later support capture of the batch call. Reject DURING_API_CALL during capture, because the driver does not allow it.
> 
> I could not confirm this. Perhaps this was a misunderstanding of another API change in CUDA 13.1?

You're right. There is no graph support for batch copies yet.

cuda_core/AGENTS.md
```diff
@@ -151,6 +151,13 @@ a `StrEnum` is accepted as an argument, a `str` should also be acceptable.  An
 invalid value should raise an exception.  When a function returns a `str` drawn
 from a small number of values, return a `StrEnum` subclass instead.
 
+For `__post_init__` validation in frozen dataclasses, use the
+`not isinstance(value, EnumType) → try EnumType(value) except (ValueError,
+TypeError)` pattern (modelled on `_normalize_enum` in
```
Andy-Jost, line 156: "modelled" is a British spelling FYI

Andy-Jost (approved):

PR #2595, opened 2026-08-10
fix(core): respect DLPack byte offsets

### LINKED ISSUE #2592: [BUG]: `StridedMemoryView.from_dlpack` ignores `DLTensor.byte_offset`

aryanputta:
## Summary

Fixes #2592 by including `DLTensor.byte_offset` when initializing `StridedMemoryView.ptr` from a DLPack capsule.

## Root cause

The capsule import path used `DLTensor.data` directly, while the DLPack byte offset was ignored. The internal capsule helper already accounts for this offset.

## Test

Added a regression test that constructs a versioned CPU DLPack capsule with a nonzero byte offset and verifies both the pointer and NumPy round trip.

## Validation

- Python syntax compilation passed.
- `git diff --check` passed.
- Runtime pytest is environment-blocked locally because the repository test helper hardcodes Linux `libc.so.6`; the test remains included for CI validation.


aryanputta:
Closing this as a duplicate of #2594, which already contains the same production fix plus versioned and unversioned regression coverage.

PR #2602, opened 2026-08-11
cuda.core: allow updating Buffer deallocation streams

### LINKED ISSUE #2600: cuda.core: allow Buffer deallocation stream to be updated before teardown
## Summary

Add a public method for changing the stream that will order a `Buffer`'s eventual deallocation without closing the buffer:

`Buffer.set_deallocation_stream(stream: Stream | GraphBuilder) -> None`

This is a follow-up to #2526.

## Problem

A `Buffer` records a deallocation recipe when it acquires an allocation. `Buffer.close(stream=...)` can replace that recipe, but it immediately closes the `Buffer`.

Applications may transfer responsibility for a live allocation from work on stream A to work on stream B while retaining the `Buffer` or relying on automatic/deferred destruction. After establishing the necessary stream dependency, there is currently no public way to make stream B order the eventual deallocation without also closing the buffer.

This matters when the final release occurs through garbage collection or through another retained `DevicePtrHandle`, such as a graph attachment.

## Proposed API

```python
buffer.set_deallocation_stream(stream)
```

The method should:

- Accept the same `Stream | GraphBuilder` values as `Buffer.close(stream=...)`.
- Replace the stored deallocation recipe without freeing or otherwise invalidating the buffer.
- Use the context-binding and PTDS handling introduced by #2526.
- Reject `None`.
- Raise if the buffer is already closed.
- Return `None`.

`Buffer.close(stream=stream)` should remain equivalent to updating the deallocation stream and then closing the buffer.

## Ordering semantics

This method does not enqueue synchronization, transfer ownership to a CUDA stream, or establish dependencies between streams. The caller must ensure that the new deallocation stream is ordered after allocation and all uses of the memory, using events or other CUDA synchronization mechanisms as appropriate.

Concurrent mutation and destruction of the same buffer remain the caller's responsibility.

## Reflection

Do not add a public getter in this initial change.

The stored `DeallocationStream` is a teardown recipe rather than an ordinary `cuda.core.Stream`: default-stream recipes include a bound context, and PTDS recipes also retain host-thread metadata. Returning only a `Stream` would not faithfully represent or round-trip the complete recipe.

A separate immutable diagnostic representation can be considered later if a concrete introspection use case arises.

## Acceptance criteria

- A live buffer can have its deallocation stream changed without being closed.
- Subsequent `close()` or automatic destruction uses the replacement recipe.
- The buffer remains usable after `set_deallocation_stream()`.
- Context-bound default-stream and PTDS behavior matches #2526.
- Existing `close(stream=...)` behavior remains unchanged.

## Related

- #2497
- #2526
- #2001
- #1989

Andy-Jost:
## Summary

Closes #2600.

Add `Buffer.set_deallocation_stream` so callers can replace the stream that orders eventual deallocation without closing the buffer. This supports transferring an allocation's lifetime to another stream while preserving the existing `Buffer` object.

## Changes

- Add the public `Buffer.set_deallocation_stream` API and generated type stub.
- Route initialization, explicit close overrides, and the new setter through the same validation and native `DeallocationStream` capture path.
- Document the caller's responsibility to order prior allocation and access work before deallocation on the replacement stream.
- Add focused tests covering initialization, setter, and close configuration followed by explicit close or garbage collection.

## Related Work

- Depends on #2526 and must be rebased onto `main` after that PR merges.
- [Focused diff relative to #2526's branch](https://github.com/Andy-Jost/cuda-python/compare/ajost/deallocation-stream...ajost/buffer-set-deallocation-stream)

### DIFF
--- a/cuda_core/cuda/core/_memory/_buffer.pyx
+++ b/cuda_core/cuda/core/_memory/_buffer.pyx
@@ -745,10 +745,9 @@ cdef tuple Buffer_coerce_batch(object buffers, str what, str single_hint):
 
 cdef inline void Buffer_set_deallocation_stream(Buffer self, object stream):
     """Validate and replace a live buffer's deallocation recipe."""
-    cdef Stream s
     if not self._h_ptr:
         raise RuntimeError("Cannot set the deallocation stream on a closed Buffer")
-    s = Stream_accept(stream)
+    cdef Stream s = Stream_accept(stream)
     _apply_deallocation_stream(self._h_ptr, s._h_stream)
 
 
--- a/cuda_core/cuda/core/_memory/_buffer.pyx
+++ b/cuda_core/cuda/core/_memory/_buffer.pyx
@@ -252,7 +252,7 @@ cdef class Buffer:
         # The parent process's stream is not portable across processes, so the
         # pickle path cannot thread an explicit stream through. Seed the
         # imported buffer's deallocation with the current context's default
-        # stream; the receiver can override via buffer.close(stream).
+        # stream; the receiver can override it before or during close.
         return Buffer.from_ipc_descriptor(mr, ipc_descriptor, stream=default_stream())
 
     def __reduce__(self) -> tuple[object, ...]:
@@ -347,8 +347,44 @@ cdef class Buffer:
         stream : :obj:`~_stream.Stream` | :obj:`~graph.GraphBuilder`, optional
             The stream object to use for asynchronous deallocation. If None,
             the deallocation stream stored in the handle is used.
+
+        See Also
+        --------
+        set_deallocation_stream
+            Change the deallocation stream without closing the buffer.
         """
         Buffer_close(self, stream)
+
+    def set_deallocation_stream(self, stream: Stream | GraphBuilder) -> None:
+        """Change the stream that orders this buffer's eventual deallocation.
+
+        The buffer remains open and usable. A later :meth:`close` without a
+        stream, garbage collection, or release of the final retained device
+        pointer handle uses the replacement stream.
+
+        This method does not synchronize streams or establish dependencies.
+        The caller must ensure that allocation and all accesses are ordered
+        before the deallocation on ``stream``.
+
+        Parameters
+        ----------
+        stream : :obj:`~_stream.Stream` | :obj:`~graph.GraphBuilder`
+            The stream to use for eventual asynchronous deallocation.
+
+        Raises
+        ------
+        RuntimeError
+            If the buffer is already closed, or if a default-stream token
+            cannot be bound because no CUDA context is current.
+        TypeError
+            If ``stream`` is ``None`` or is not an accepted stream object.
+
+        Notes
+        -----
+        Synchronizing concurrent mutation and destruction of the same buffer
+        is the caller's responsibility.
+        """
+        Buffer_set_deallocation_stream(self, stream)
 
     def __enter__(self):
         return self
@@ -707,15 +743,22 @@ cdef tuple Buffer_coerce_batch(object buffers, str what, str single_hint):
     return tuple(out)
 
 
+cdef inline void Buffer_set_deallocation_stream(Buffer self, object stream):
+    """Validate and replace a live buffer's deallocation recipe."""
+    cdef Stream s
+    if not self._h_ptr:
+        raise RuntimeError("Cannot set the deallocation stream on a closed Buffer")
+    s = Stream_accept(stream)
+    _apply_deallocation_stream(self._h_ptr, s._h_stream)
+
+
 cdef inline void Buffer_close(Buffer self, object stream):
     """Close a buffer, freeing its memory."""
-    cdef Stream s
     if not self._h_ptr:
         return
     # Update deallocation stream if provided
     if stream is not None:
-        s = Stream_accept(stream)
-        _apply_deallocation_stream(self._h_ptr, s._h_stream)
+        Buffer_set_deallocation_stream(self, stream)
     # Reset handle - RAII deleter will free the memory (and release owner ref in C++)
     self._h_ptr.reset()
     self._size = 0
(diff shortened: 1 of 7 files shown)


cuda_core/cuda/core/_memory/_buffer.pyx
```diff
@@ -619,15 +705,23 @@ cdef Buffer Buffer_from_deviceptr_handle(
     return buf
 
 
+cdef inline void Buffer_set_deallocation_stream(Buffer self, object stream):
+    """Validate and replace a live buffer's deallocation recipe."""
+    cdef Stream s
+    if not self._h_ptr:
+        raise RuntimeError("Cannot set the deallocation stream on a closed Buffer")
+    s = Stream_accept(stream)
```
juenglin:
```suggestion
    cdef Stream s = Stream_accept(stream)
```

cuda_core/docs/source/api.rst
```diff
@@ -77,6 +77,12 @@ Memory management
    ManagedMemoryResourceOptions
    VirtualMemoryResourceOptions
 
+A :class:`Buffer` records the stream that will order its eventual deallocation.
+Use :meth:`Buffer.set_deallocation_stream` to replace that stream without
+closing the buffer. Changing the recorded stream does not synchronize streams;
+the caller must order allocation and every access before the deallocation,
```
juenglin, line 83: Could you add a code example for cuda.core users that demonstrates how to safely switch deallocation streams?

cuda_core/cuda/core/_memory/_buffer.pyx
```diff
@@ -49,23 +52,21 @@ cdef void _mr_dealloc_callback(
...
-    the unique exception to the "no implicit default-stream fallback"
-    policy because the teardown has no other source of truth.
-    """
+    """Called by the C++ deleter to deallocate via MemoryResource.deallocate."""
     cdef Stream stream
     try:
-        stream = Stream._from_handle(Stream, h_stream) if h_stream else default_stream()
+        if not h_stream:
+            print(
+                "Warning: no deallocation stream was recorded; falling back to "
```
juenglin, line 60: Should it say "no deallocation stream was recorded or the deallocation stream was closed"?

cuda_core/cuda/core/_memory/_buffer.pyx
```diff
@@ -619,15 +705,23 @@ cdef Buffer Buffer_from_deviceptr_handle(
     return buf
 
 
+cdef inline void Buffer_set_deallocation_stream(Buffer self, object stream):
+    """Validate and replace a live buffer's deallocation recipe."""
+    cdef Stream s
+    if not self._h_ptr:
+        raise RuntimeError("Cannot set the deallocation stream on a closed Buffer")
```
juenglin, line 749: Should there be a similar runtime error for a closed stream?

cuda_core/cuda/core/_memory/_buffer.pyx
Andy-Jost: It's saying no stream is available for ordering the destruction. Not necessarily that a stream was provided and was closed.

cuda_core/cuda/core/_memory/_buffer.pyx
Andy-Jost: Yes. It appears that closed Stream detection is lacking throughout. We should audit the other objects, too (e.g., `Event`s, etc.).

Andy-Jost: Follow-up: issue #2627

cuda_core/examples/buffer_deallocation_stream.py
```diff
@@ -0,0 +1,76 @@
...
+    try:
+        output = pinned_mr.allocate(size)
+        buffer, ready = produce_data(device, producer_stream, size, value)
+        consume_data(buffer, ready, output, consumer_stream)
+
+        # No stream argument is needed. The buffer now records consumer_stream.
+        # The free operation runs after the copy on that stream.
+        buffer.close()
+        buffer = None
+        consumer_stream.sync()
```
juenglin, line 59: This looks wrong: you close a buffer then sync a stream that has pending copy for the buffer. This sync could fail with a RuntimeError, couldn't it?

cuda_core/cuda/core/_memory/_buffer.pyx
juenglin: Yes, but wouldn't this warning also appear when the deallocation stream was closed?

cuda_core/examples/buffer_deallocation_stream.py
Andy-Jost: The buffer in question has stream-ordered deallocation semantics, so closing it places a call to `cuMemFreeAsync` onto `consumer_stream`. The idea is that transferring the buffer to the stream "currently" using it will order its free after the work using it.

cuda_core/cuda/core/_memory/_buffer.pyx
Andy-Jost: Two points: This warning does not occur if a valid stream is passed to `set_deallocation_stream()` and that stream is later closed, because the `DevicePtrHandle` holds an independent stream handle. It can occur if an already closed stream is passed.

The follow-up in #2635 should address the second case: closed streams will be rejected throughout the API, including by `set_deallocation_stream()`. If the proper checks are in place throughout the codebase, this condition should never occur, which is why I flagged it as an internal error.

juenglin: > ... because the DevicePtrHandle holds an independent stream handle.

What does invoking `Stream.close` mean then?

cuda_core/examples/buffer_deallocation_stream.py
juenglin: Ah. For the other methods for stream-ordered operations `stream` is a mandatory parameter. That no stream is passed to `close` here caused me to think it is not an async operation.

juenglin (approved):
LGTM!

cuda_core/cuda/core/_memory/_buffer.pyx
Andy-Jost: A `StreamHandle` is a `std::shared_ptr<CUstream>`. A Python `Stream` class holds one of those, and `Stream.close` resets the shared pointer, dropping that reference. Inside a `DevicePtrHandle` is an independent `StreamHandle` referring to the deallocation stream. Whatever the user-facing Python `Stream` object does, the embedded reference cannot be invalidated.

PR #2606, opened 2026-08-12
fix(cuda.core): declare cdef attrs for system event wrappers and decode packed gpu_id

rluo8:
## Description

Two issues were found when adding coverage tests (internal bug 6595639):
1. **Missing `cdef` attributes** on `SystemEvent`, `SystemEvents`, `EventData`,  and `DeviceAttributes`. These are Cython `cdef class` types (no `__dict__`),  so undeclared `self._event_data` / `self._attributes` assignments raise `AttributeError`. `RegisteredSystemEvents.wait()` therefore crashes as soon as a real bind/unbind event arrives.

2. **`SystemEvent.device` type mismatch.** NVML's packed int `gpu_id`  (`domain[31:16] | bus[15:8] | device[7:0]`), which requires an NVML-style PCI bus ID string (`NVML_DEVICE_PCI_BUS_ID_FMT`). Decode at the call site  via `_pci_bus_id_from_gpu_id`.

Adds regression tests for construction, packing decode, and end-to-end `SystemEvent.device` resolution against live `pci_info`.

**Verification:**
- B10: new regression tests 8 passed; live unbind of `0000:c1:00.0`
  yields `SystemEvents` with `gpu_id=0xC100` → `00000000:C1:00.0`, and
  post-rebind `SystemEvent.device` resolves correctly
- A100 MIG: `DeviceAttributes` cdef fix verified
- System Event register/wait remains Linux-only (Windows/WSL/WOA skip)

### DIFF
--- a/cuda_core/cuda/core/system/_device_attributes.pxi
+++ b/cuda_core/cuda/core/system/_device_attributes.pxi
@@ -7,6 +7,8 @@ cdef class DeviceAttributes:
     """
     Various device attributes.
     """
+    cdef object _attributes
+
     def __init__(self, attributes: nvml.DeviceAttributes):
         self._attributes = attributes
 
--- a/cuda_core/cuda/core/system/_event.pxi
+++ b/cuda_core/cuda/core/system/_event.pxi
@@ -29,6 +29,8 @@ cdef class EventData:
     """
     Data about a single event.
     """
+    cdef object _event_data
+
     def __init__(self, event_data: nvml.EventData):
         self._event_data = event_data
 
--- a/cuda_core/cuda/core/system/_system_events.pyi
+++ b/cuda_core/cuda/core/system/_system_events.pyi
@@ -84,10 +84,9 @@ class RegisteredSystemEvents:
 
 def _pci_bus_id_from_gpu_id(gpu_id: int) -> str:
     """
-    Decode a packed NVML ``gpu_id`` (``domain[31:16] | bus[15:8] | device[7:0]``)
-    into an NVML-style PCI bus ID (``NVML_DEVICE_PCI_BUS_ID_FMT``).
+    Decode an NVML System Event packed ``gpu_id`` into an NVML-style PCI bus ID
+    string.
     """
-
 def register_events(events: SystemEventType | str | list[SystemEventType | str]) -> RegisteredSystemEvents:
     """
     Starts recording of events on test system.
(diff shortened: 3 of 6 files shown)


rluo8:
Hi @mdboom , this issue is about system event. Could you please help review it?
Thanks!

mdboom (commented):
What is the root issue this is trying to solve?

This is making these objects mutable and they definitely shouldn't be -- their intention is to return values from the NVML API, not to set things on them.

The gpu_id / pci_id bridge in `_system_events.pyx` added here makes sense, though.

rluo8:
Hi @mdboom , thanks for the review. I understand your concern. The root issue is that constructing SystemEvent / SystemEvents / EventData / DeviceAttributes throws exception: these are Cython cdef class types with no __dict__, so undeclared self._event_data / self._attributes in __init__ raise AttributeError. 

This is the reproduce script:
```
from cuda.core.system._system_events import SystemEvents

ed = nvml.SystemEventData_v1(1)
ed.event_type = nvml.SystemEventType.GPU_DRIVER_UNBIND
ed.gpu_id = 0xC100
SystemEvents(ed) 
```
Output:
```
   SystemEvents(ed)  # main → AttributeError: no '_event_data' / no __dict__
    ~~~~~~~~~~~~^^^^
  File "cuda/core/system/_system_events.pyx", line 60, in cuda.core.system._system_events.SystemEvents.__init__
AttributeError: 'cuda.core.system._system_events.SystemEvents' object has no attribute '_event_data' and no __dict__ for setting new attributes
```

After the fix, it doesn't make the public API mutable, the cdef field is not visible from Python.
Test script:
```
from cuda.core.system._system_events import SystemEvent

ed = nvml.SystemEventData_v1(1)
ed.event_type = nvml.SystemEventType.GPU_DRIVER_BIND
ed.gpu_id = 0xC100
event = SystemEvent(ed)
print("constructed OK:", event.event_type, hex(event.gpu_id))

try:
    event.event_type = "BIND"
except Exception as err:
    print(repr(err))

try:
    print(event._event_data)
except Exception as err:
    print(repr(err))
```
Output:
```
constructed OK: bind 0xc100
AttributeError("attribute 'event_type' of '...SystemEvent' objects is not writable")
AttributeError("'...SystemEvent' object has no attribute '_event_data'")
```
This looks similar to Cython’s guidance about cdef class AttributeErrors : https://docs.cython.org/en/stable/src/userguide/troubleshooting.html
AttributeErrors can also happen when writing into a cdef class, commonly in __init__:
```
cdef class Company:
    def __init__(self, staff):
        self.staff = staff  # AttributeError!
```
Unlike a regular class, cdef class has a fixed list of attributes that you can write to and you need to declare them explicitly. For example:
```
cdef class Company:
   cdef list staff
   # ...
```
Thanks!

mdboom:
Got it.  Thanks for the explanation.

mdboom (approved):

PR #2608, opened 2026-08-12
PERF: Inline return code checks in cuda_core

mdboom:
This takes ~20ns off of every CUDA API call made in cuda-core using a trick that has existed in cuda_bindings for a few months.

The idea is to inline the check (the comparison to zero) at every call site, and only make a function call to handle raising an exception if you need to.

To avoid making any changes, this is literally just a move of these functions to inline (in the `.pxd` file) and a few declarations to make them work in that context.


Andy-Jost (approved):
Inlining for the win!

PR #2634, opened 2026-08-14
cuda.core: accept dict for XyzOptions that are extension types

### LINKED ISSUE #2248: [DOC]: cuda.core docs/type hints understate option params since dict is accepted at runtime but not documented

juenglin:
Restores undocumented dict acceptance for `options` in several `cuda.core` constructors/methods, which was inadvertently removed by #2619. Covers the three memory-resource constructors and the stream-creation path (`Stream._init`, `Context.create_stream`).

## Why this is necessary

`check_or_create_options` (`cuda_core/cuda/core/_utils/cuda_utils.pyx`) normalizes a plain `dict` into the options dataclass via `cls(**options)`, so passing a dict for `options` has always worked at runtime. The options classes (`DeviceMemoryResourceOptions`, `StreamOptions`, etc.) are `@dataclass cdef class` — Cython extension types.

Cython enforces PEP-484 parameter annotations at the call boundary of `def` methods on `cdef class`es: when a parameter is annotated with an extension type, Cython emits a runtime `isinstance` check that rejects other types (including `dict`) with `TypeError` before the method body runs. See the ["Extension types and None" section of the Extension Types guide](https://docs.cython.org/en/latest/src/userguide/extension_types.html#extension-types-and-none).

#2619 removed `| dict[str, object]` from these annotations so the public stubs would show only the typed dataclass. That was correct in spirit, but it turned the call-boundary check into a regression: dicts that previously flowed through to `check_or_create_options` now get rejected at the boundary, silently breaking any code relying on the dict shorthand.

## What this PR does

Applies Cython's `@cython.annotation_typing(False)` compiler directive to each affected `def` method. The directive tells Cython to ignore the PEP-484 annotations for its own type analysis, so the parameter is treated as a generic `object` at the call boundary and dicts pass through to `check_or_create_options`. The annotations remain in the source and are still picked up by `stubgen-pyx`, so the generated `.pyi` stubs naturally show the clean `XxxOptions | None` signature — no stub postprocessor needed. See the ["Disabling annotations" section of the Pure Python Mode guide](https://docs.cython.org/en/latest/src/tutorial/pure.html#disabling-annotations).

## Changes

- `cuda_core/cuda/core/_memory/_device_memory_resource.pyx`, `_managed_memory_resource.pyx`, `_pinned_memory_resource.pyx`: `@cython.annotation_typing(False)` on each `__init__`; `options` annotation restored to `XxxOptions | None = None`.
- `cuda_core/cuda/core/_stream.pyx`: `@cython.annotation_typing(False)` on `Stream._init`.
- `cuda_core/cuda/core/_context.pyx`: `@cython.annotation_typing(False)` on `Context.create_stream`.
- `tests/memory/test_backward_compatibility.py`: tests asserting each memory-resource constructor accepts a plain dict for `options`.
- `tests/test_stream.py`: test asserting `Device.create_stream` accepts a plain dict for `options`.

Generated `.pyi` stubs are unchanged by re-running `stubgen-pyx` (idempotent): the directive is invisible to the stub generator and the source annotations are preserved verbatim.

closes #2248

### DIFF
--- a/cuda_core/cuda/core/_context.pyx
+++ b/cuda_core/cuda/core/_context.pyx
@@ -8,6 +8,8 @@ from collections.abc import Sequence
 from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
+import cython
+
 from cuda.bindings cimport cydriver
 from cuda.core._device_resources cimport DeviceResources, SMResource, WorkqueueResource
 from cuda.core._device_resources import SMResource, WorkqueueResource
@@ -99,6 +101,7 @@ cdef class Context:
         Context_check_open(self)
         return DeviceResources._init_from_ctx(self._h_context, self._device_id)
 
+    @cython.annotation_typing(False)
     def create_stream(self, options: StreamOptions | None = None) -> Stream:
         """Create a new stream bound to this green context.
 
--- a/cuda_core/cuda/core/_memory/_device_memory_resource.pyx
+++ b/cuda_core/cuda/core/_memory/_device_memory_resource.pyx
@@ -149,7 +149,7 @@ cdef class DeviceMemoryResource(_MemPool):
     def __init__(
         self,
         device_id: Device | int,
-        options: DeviceMemoryResourceOptions | None = None
+        options=None
     ) -> None:
         _DMR_init(self, device_id, options)
 
--- a/cuda_core/cuda/core/_memory/_device_memory_resource.pyx
+++ b/cuda_core/cuda/core/_memory/_device_memory_resource.pyx
@@ -20,6 +20,8 @@ from cuda.core._utils.cuda_utils cimport (
     check_or_create_options,
     HANDLE_RETURN,
 )
+
+import cython
 from dataclasses import dataclass
 import multiprocessing
 import platform  # no-cython-lint
@@ -146,10 +148,11 @@ cdef class DeviceMemoryResource(_MemPool):
     def __cinit__(self, *args, **kwargs) -> None:
         self._dev_id = cydriver.CU_DEVICE_INVALID
 
+    @cython.annotation_typing(False)
     def __init__(
         self,
         device_id: Device | int,
-        options=None
+        options: DeviceMemoryResourceOptions | None = None
     ) -> None:
         _DMR_init(self, device_id, options)
 
--- a/cuda_core/cuda/core/_memory/_managed_memory_resource.pyx
+++ b/cuda_core/cuda/core/_memory/_managed_memory_resource.pyx
@@ -13,6 +13,7 @@ from cuda.core._utils.cuda_utils cimport HANDLE_RETURN
 from cuda.core._utils.cuda_utils cimport check_or_create_options  # no-cython-lint
 from cuda.core._utils.cuda_utils import CUDAError  # no-cython-lint
 
+import cython
 from dataclasses import dataclass
 import threading
 from typing import TYPE_CHECKING
@@ -97,7 +98,8 @@ cdef class ManagedMemoryResource(_MemPool):
     memory pools.
     """
 
-    def __init__(self, options=None) -> None:
+    @cython.annotation_typing(False)
+    def __init__(self, options: ManagedMemoryResourceOptions | None = None) -> None:
         _MMR_init(self, options)
 
     def allocate(self, size_t size, *, stream: Stream | GraphBuilder) -> ManagedBuffer:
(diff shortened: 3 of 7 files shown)


toolshed/run_stubgen_pyx.py
```diff
@@ -36,17 +39,67 @@ def _normalize_stub_headers(root: pathlib.Path) -> None:
         stub.write_bytes(first_line.replace(b"\\", b"/") + data[newline:])
 
 
+def _collect_stub_overrides(root: pathlib.Path) -> dict[pathlib.Path, list[tuple[str, str]]]:
```
mdboom: Significant changes or workarounds to stubgen-pyx should be submitted to that project upstream.  (The existing workarounds in this script have already been submitted and merged, we are just waiting for a new release without regressions to update to, so the plan is always to live without this workaround script).

But the larger question is -- why do we need to have the annotation differ between the `.pyx` and the `.pyi`.  I believe the annotation is ignored in the `.pyx` anyway.  Unless in pure-Python mode, Cython only cares about the type before the argument name, not the type after the name and colon.

mdboom (changes requested):
I think the stubgen-pyx workaround is ultimately unnecessary.  Cython doesn't care about the annotation, so we can just set them directly to what we want them to be in the `.pyi`.

We can declare or not declare whether we accept a dict in the type annotation (that's sort of a larger question we are trying to answer offline -- whether we want to /commit/ to that).

But the tests on their own if useful if we decide we want to accept dicts going forward.

I should add Cython is a weird tool and Python type annotations being informational-only are pretty unique among languages.  When you put them together (Cython pre-dates Python type annotations by at least a decade), it's a very confusing mishmash.

juenglin:
> Cython doesn't care about the annotation, so we can just set them directly to what we want them to be in the .pyi.

That's not what I found. If you run the tests added by this PR against a main build, you will find that they are failing.

I dislike this change also and would be happy to close this PR. But that requires that we accept the behavior change introduced by PR #2619.

mdboom:
> > Cython doesn't care about the annotation, so we can just set them directly to what we want them to be in the .pyi.
> 
> That's not what I found. If you run the tests added by this PR against a main build, you will find that they are failing.

Ah, you are right.  I was thinking of `cdef` functions/methods, not `def` ones.

> I dislike this change also and would be happy to close this PR. But that requires that we accept the behavior change introduced by PR #2619.

Yeah, this is kind of a mess with the way Cython uses type annotations for runtime semantic behavior (which is not how Python works at all).  We should probably discuss the best way forward at the next meeting.

mdboom (commented):
This is 100% a more manageable solution than the previous one, and I'd be fine merging this on a technical/maintenance level.

Not approving just yet because:

1) I think we should have one more quick round of discussion to make sure this is directionally where we want to go

2) If this is, let's add some AGENTS.md comments recommending this annotation on `cdef class` methods that accept option classes (or whatever the exact "rule" we need to apply is).

leofang (approved):
Approving based on the exchange of the Slack thread (and Mike's existing gating -- as Mike noted above, let's discuss this in the next team sync before merging).

mdboom (approved):
LGTM

PR #2635, opened 2026-08-14
cuda.core: reject operations on closed resource objects

### LINKED ISSUE #2627: cuda.core: reject operations on closed resource objects
## Summary

Add a consistent closed-state contract to `cuda.core` resources. Closed objects must fail before a NULL native handle reaches CUDA.

## Problem

Several `cuda.core` objects reset a shared native handle in `close()`, but their methods do not check that state. This can cause late CUDA errors, stale metadata, or invalid native access. Streams are the highest risk because CUDA can interpret a NULL `CUstream` as the default stream. Work that targets a closed stream can therefore run on the wrong stream instead of failing.

The same risk exists in event operations, buffer operations, memory pools, graph resources, compiler resources, texture resources, and cross-object conversion paths.

## Required behavior

- Add `__bool__` where an object has a clear open or closed state.
- Test the shared handle state, not the numeric CUDA handle. An open `Stream.from_handle(0)` and an open zero-size `Buffer` must remain true.
- Raise `RuntimeError("<Type> has been closed")` before an active operation uses a closed resource.
- Keep repeated `close()`, handle inspection, and `repr()` safe.
- Keep current equality and hash behavior. A separate change can address mutable hash or closed-object equality rules.
- Make `Stream.close()` a no-op for CUDA default-stream tokens. These tokens use process-wide singleton wrappers and own no CUDA resource.
- Keep memory-pool `deallocate()` available after pool close so that outstanding buffers can complete cleanup.

## Scope

Apply the contract to these resources and their acceptance paths:

- `Stream`, `Event`, `Buffer`, and `ManagedBuffer`.
- `Context`, `Program`, and `Linker`.
- `DeviceMemoryResource`, `PinnedMemoryResource`, and `ManagedMemoryResource`.
- `GraphBuilder`, executable `Graph`, destroyed `GraphNode` objects, and invalid child `GraphDefinition` views.
- `OpaqueArray`, `MipmappedArray`, `TextureObject`, `SurfaceObject`, and `GraphicsResource`.
- `IPCAllocationHandle`.

Update central conversion and acceptance paths. This includes `Stream_accept`, stream protocol conversion, event use in streams and graphs, buffer use in copy and managed-memory operations, DLPack, IPC, memory views, kernel arguments, virtual memory, and texture descriptors.

A valid virtual graph entry node and a destroyed graph node both have a NULL `CUgraphNode`. Add explicit state that distinguishes them. Reject destroyed nodes in graph construction and adjacency changes. Reject invalid child graph views in graph queries, embedding, instantiation, and updates.

`GraphMemoryResource.close()` and `ProgramCacheResource.close()` are out of scope because they do not create a closed native-resource state.

## Buffer deallocation stream

`Buffer.set_deallocation_stream()` and `Buffer.close(stream=...)` must reject a closed `Stream` or closed `GraphBuilder`. A failed update must leave the current deallocation stream unchanged. A closed `Buffer` must remain rejected as specified in #2600 and #2602.

## Acceptance criteria

- Each covered resource reports the correct Boolean state before and after close or invalidation.
- No covered active operation passes a released handle to CUDA.
- `Stream_accept` rejects every closed stream source.
- Closed events are rejected by stream and graph event operations.
- Closed buffers are rejected by direct methods and indirect consumers.
- Closed memory pools reject new work but still permit required deallocation.
- Destroyed graph nodes and invalid child graphs fail with a clear runtime error.
- CUDA default-stream singleton wrappers remain usable after `close()`.
- Tests cover Boolean state, safe object protocols, idempotent cleanup, active-method guards, conversion paths, and deallocation-stream failure atomicity.

## Related

- #2600
- #2602

Andy-Jost:
## Description

closes #2627

Add a consistent closed-state contract across `cuda.core` resource objects so released native handles are rejected before reaching CUDA. Closeable resources now expose `is_closed`, while `GraphDefinition` and `GraphNode` expose `is_valid` for graph-lifetime invalidation. This change also adds tests ensuring `bool(obj)` returns true for closed and invalid objects, to retain backwards compatibility.

Active methods validate their own state and accepted resource arguments. `Stream_accept()` now rejects closed streams and graph builders, which also makes `Buffer.set_deallocation_stream()` and `Buffer.close(stream=...)` reject a closed stream without replacing the buffer's saved deallocation recipe.

The same validation covers events, buffers, memory pools, IPC handles, contexts, compiler resources, graphs, arrays, textures, surfaces, and graphics resources. Tests cover named lifecycle state, backward-compatible truthiness, idempotent cleanup, safe inspection, cross-object validation, graph invalidation, and deallocation-stream failure atomicity.

## Checklist


cuda_core/tests/memory_ipc/test_send_buffers.py
```diff
@@ -131,10 +131,11 @@ def process_b_main(self, buffer, q_bc, event_b):
 
         # Forward the buffer to C.
         q_bc.put(buffer)
-        buffer.close()
 
-        # Wait for C to receive before exiting.
+        # Queue serialization runs in a feeder thread. Keep the buffer open
+        # until C has received it and the parent releases this process.
         event_b.wait(timeout=CHILD_TIMEOUT_SEC)
+        buffer.close()
```
Andy-Jost, line 138: This fixes a latent bug. `multiprocessing.Queue.put` requires its argument to remain valid until received.

cuda_core/cuda/core/_memory/_buffer.pyx
```diff
@@ -739,14 +759,20 @@ cdef tuple Buffer_coerce_batch(object buffers, str what, str single_hint):
     for item in buffers:
         if not isinstance(item, Buffer):
             raise TypeError(f"{what}: expected Buffer, got {type(item).__name__}")
+        Buffer_check_open(<Buffer>item)
         out.append(item)
     return tuple(out)
 
 
+cdef int Buffer_check_open(Buffer self) except -1:
```
mdboom: Since it's called on basically every path, inlining should help.  Since this is used from other modules, move the implementation to the `.pxd` and add the `inline` keyword.

cuda_core/cuda/core/_memory/_memory_pool.pxd
```diff
@@ -35,6 +35,7 @@ cdef int MP_init_current_pool(
 ) except? -1
 
 cdef int MP_raise_release_threshold(_MemPool self) except? -1
+cdef int MP_check_open(_MemPool self) except -1
```
mdboom: Move the implementation here and inline?

cuda_core/cuda/core/_memory/_managed_buffer.py
```diff
@@ -46,6 +46,8 @@
 
 
 def _get_int_attr(buf: Buffer, attribute: Any) -> int:
+    if buf.is_closed:
+        raise RuntimeError("Buffer has been closed")
```
mdboom, line 50: `Buffer_check_open(buf)`?

cuda_core/cuda/core/_memory/_memory_pool.pyx
```diff
@@ -354,6 +363,12 @@ cdef inline void _MP_deallocate(
         HANDLE_RETURN(cydriver.cuMemFreeAsync(devptr, s))
 
 
+cdef int MP_check_open(_MemPool self) except -1:
+    if not self._h_pool:
+        raise RuntimeError(f"{self.__class__.__name__} has been closed")
+    return 0
```
mdboom: Move to `.pxd` and inline.

mdboom (changes requested):
General comment -- move the helper functions to inline implementations in the `.pxd`.  Then I see a mix of calling these helper functions and doing an `if closed: raise(...)`.  Is there a reason for that difference?  If not, maybe consistently use the helper functions?

Andy-Jost:
> General comment -- move the helper functions to inline implementations in the `.pxd`. Then I see a mix of calling these helper functions and doing an `if closed: raise(...)`. Is there a reason for that difference? If not, maybe consistently use the helper functions?

Generally agree. The latest change consolidates open-state checks in inline functions. Shared Cython checks live in `.pxd` files, and module-local checks live in `.pyx` files. The remaining exceptions involve pure Python modules or circular Cython dependencies.

Details:
- `_managed_buffer.py` is pure Python, so it cannot call the `cdef` checker from `_buffer.pxd`. It now uses a local Python `_check_open()` helper.
- `_virtual_memory_resource.py` is also pure Python. It has only one check, so it uses `buf.is_closed` directly.
- `_stream.pyx` cannot directly `cimport` `GB_check_open` without creating a circular dependency.` Stream_accept()` instead accesses `arg.stream`, whose property calls the local inline checker.

cuda_core/cuda/core/_memory/_managed_buffer.py
Andy-Jost: I don't think the Cython function can be called from a pure Python module.

juenglin:
Python's precedent is `closed`.
```
>>> open("README.md", "r").closed
False
```
We don't need to follow it, of course.

cuda_core/cuda/core/_memory/_managed_buffer.py
mdboom: Ah.  I missed that this was a `.py` file.  Makes sense.

cuda_core/cuda/core/_stream.pyx
```diff
@@ -201,10 +201,18 @@ cdef class Stream:
         underlying CUDA stream. For borrowed streams, this releases the
         reference and allows the Python owner to be GC'd.
         """
+        if self._h_stream and Stream_is_default_token(self):
+            return
```
mdboom: I think this one is legit, but I don't fully understand how these special Stream singletons work.

From Claude:

`Stream.close()` early-returns for any stream whose raw handle equals `CU_STREAM_LEGACY`/`CU_STREAM_PER_THREAD`, because `Stream_is_default_token()` compares raw handle *values*, not identity against the two default-stream singletons. `Stream.from_handle(1)` / `Stream.from_handle(2)` — legitimate user-created borrowed wrappers — become permanently un-closeable: the owner reference is never released and `is_closed` stays `False` forever, contradicting the method's own docstring. Fix: guard on `self is LEGACY_DEFAULT_STREAM or self is PER_THREAD_DEFAULT_STREAM`, not on handle value. The new test (`test_raw_null_stream_is_live_until_closed`) uses `from_handle(0)`, which happens to sidestep this exact case.

cuda_core/cuda/core/_memory/_memory_pool.pyx
mdboom, line 192: I think whether Claude is right about this depends on whether double-calling `cuMemFreeAsync` is an error.

From Claude:

`_MemPool.deallocate()` (backing `DeviceMemoryResource`/`PinnedMemoryResource`/`ManagedMemoryResource`) never got an `MP_check_open(self)` call, unlike every sibling entry point in the same file (`allocate`, `attributes`, `MP_raise_release_threshold`, `__reduce__`, `peer_accessible_by`). After `mr.close()`, calling `mr.deallocate(ptr, size, stream=...)` sails straight through to `cuMemFreeAsync` against a destroyed pool — exactly the failure class this PR is meant to close everywhere else.

cuda_core/cuda/core/_memory/_buffer.pyx
```diff
@@ -335,6 +336,7 @@ cdef class Buffer:
         # Unpickling performs a live CUDA IPC import from descriptor bytes in the
         # pickle stream. Only deserialize Buffers from a trusted principal.
         # Must not serialize the parent's stream!
+        Buffer_check_open(self)
```
mdboom, line 339: This one may or may not be legit.  From Claude:

Adding `Buffer_check_open(self)` to `__reduce__` breaks the `queue.put(buffer); buffer.close()` pattern: `multiprocessing.Queue.put()` serializes on a background feeder thread, so a race lets `close()` win and `__reduce__` raises inside the feeder thread, where multiprocessing logs-and-discards it — the consumer just hangs with no error at the `put()` call site. The PR's own test (`test_send_buffers.py:134`) had to be reordered to work around this. This needs to be called out in the release notes (currently only mention rejecting closed resources, not this pickling/queue-handoff hazard).

cuda_core/cuda/core/graph/_adjacency_set_proxy.pyx
```diff
@@ -140,6 +148,20 @@ cdef class _AdjacencySetCore:
             c_from[0] = as_cu(other._h_node)
             c_to[0] = as_cu(self._h_node)
 
+    cdef inline void check_owner_mutable(self) except *:
+        if as_cu(self._h_graph) == NULL:
+            raise RuntimeError("GraphDefinition is no longer valid")
+        if as_cu(self._h_node) == NULL:
```
mdboom, line 152: From Claude:

`check_owner_mutable()` treats any NULL `_h_node` as "destroyed," but `GraphDefinition._entry` (the virtual entry node) legitimately has `_h_node == NULL` by design — `GN_check_valid` correctly exempts it via `_is_entry`, but `_AdjacencySetCore` never captures that flag for the *owner* node. So `graph_def._entry.succ.add(node)` incorrectly raises `"GraphNode has been destroyed"` for a node that's actually valid. Exposure is limited (private `_entry`, not routed through this path by the public API today), but it's a real inconsistency in the exact contract this PR establishes, and untested.

cuda_core/cuda/core/_memory/_ipc.pyx
```diff
@@ -129,6 +135,11 @@ cdef class IPCAllocationHandle:
         """Close the handle."""
         self._h_fd.reset()
 
+    @property
+    def is_closed(self) -> bool:
+        """Whether this allocation handle has been closed."""
+        return self._h_fd.get() == NULL or as_intptr(self._h_fd) < 0
+
     def __int__(self) -> int:
         if not self._h_fd or as_intptr(self._h_fd) < 0:
```
mdboom: Minor nit:

```suggestion
        if not self.is_closed:
```

cuda_core/cuda/core/_memory/_ipc.pyx
```diff
@@ -129,6 +135,11 @@ cdef class IPCAllocationHandle:
         """Close the handle."""
         self._h_fd.reset()
 
+    @property
+    def is_closed(self) -> bool:
+        """Whether this allocation handle has been closed."""
+        return self._h_fd.get() == NULL or as_intptr(self._h_fd) < 0
```
mdboom: Minor DRY nit:

```suggestion
        return IPCAllocationHandle_check_open(self)
```

cuda_core/cuda/core/graph/_adjacency_set_proxy.pyx
```diff
@@ -52,13 +52,16 @@ class AdjacencySetProxy(MutableSet[GraphNode]):
...
+        (<_AdjacencySetCore>self._core).check_mutation(value)
         if value in self:
             return
         (<_AdjacencySetCore>self._core).add_edge(<GraphNode>value)
 
     def discard(self, value: GraphNode) -> None:
+        (<_AdjacencySetCore>self._core).check_owner_mutable()
         if not isinstance(value, GraphNode):
             return
+        (<_AdjacencySetCore>self._core).check_mutation(value)
```
mdboom: Claude points out that this causes a diversion from the Python `MutableSet.discard` convention -- that an invalid `value` should just return and never raise.  Therefore this check maybe belongs right before the `remove_edge` call.

mdboom (changes requested):
Looks much better from the human-review standpoint.

Claude flagged a few things that seem like legitimate issues, but I don't have the full context to evaluate them.

Andy-Jost:
> Python's precedent is `closed`.
> 
> ```
> >>> open("README.md", "r").closed
> False
> ```
> 
> We don't need to follow it, of course.

cuda.core consistently names Boolean properties as `is_*`.

cuda_core/cuda/core/_memory/_ipc.pyx
Andy-Jost: I noticed the negative-value check was superfluous, so I simply inlined the null-check in the few places where it was needed.

cuda_core/cuda/core/_memory/_ipc.pyx
Andy-Jost: Yeah, it's annoying to have two levels of inline helper for the "is it open" check and "raise if not open" check. I don't think this suggestion can be applied because it would raise. Fortunately, I was able to simplify it by removing the negative-value check.

cuda_core/cuda/core/graph/_adjacency_set_proxy.pyx
Andy-Jost: fixed

cuda_core/cuda/core/_memory/_buffer.pyx
Andy-Jost: The pattern `queue.put(buffer); buffer.close()` was a latent race in the test that I discovered incidentally. That may explain why the test was marked flaky.

Because `mp.Queue` sends objects asynchronously, subsequent mutations (including close) must be ordered after the object is sent. I think the right thing in this case it to fix the test without publishing a release note.

cuda_core/cuda/core/_memory/_memory_pool.pyx
Andy-Jost: `_MemPool.close()` releases the Python object's memory pool handle but does not necessarily destroy the pool. Device pointer handles structurally embed an independent reference that ensures the pool remains live. I think the best option in this case is to leave `deallocate` unguarded and interpret `close` to mean future allocations are disallowed. Frees performed against a closed pool would remain valid. Otherwise, closing a pool would mean leaking any outstanding allocations that used `Buffer.from_handle(..., mr=pool)`.

I think a case could be made that pools ought not be closable. In a language like Python, explicit `close` usually doesn't make sense. Exceptions are made when resources need to be relinquished accurately, though. I'm not sure that applies to CUDA memory pools.

Andy-Jost: I updated the `_MemPool.close` docstring to clarify this.

cuda_core/cuda/core/graph/_adjacency_set_proxy.pyx
Andy-Jost: I believe this is unreachable and should be ignored. The virtual entry node is private and requesting its successors (or predecessors) is meaningless. Since there's no valid path to something like `graph_def._entry.succ` we should assume it is not used and avoid an unnecessary check.

cuda_core/cuda/core/_stream.pyx
Andy-Jost: In thinking about this, I realized it’s probably better to remove the exception that prevents closing the default stream singletons. `LEGACY_DEFAULT_STREAM` and `PER_THREAD_DEFAULT_STREAM` are process-wide, so closing them may break the program, but they are still ordinary `Stream` objects. There is no justification for adding special semantics just because they happen to be global.

This would be like preventing users from closing `sys.stdout`. Going down that path is what led to this subtle problem in the first place. It’s better to let the global remain an ordinary object, even if it can be misused.

To clarify, `LEGACY_DEFAULT_STREAM.close()` only detaches the Python object from its handle. It does not instruct CUDA to destroy the default stream. This exception only protected the Python layer.

mdboom (approved):
LGTM.

PR #2636, opened 2026-08-14
cuda.core: introduce copy options for Buffer.copy_{to/from}

### LINKED ISSUE #2365: [FEA] cuda.core: support copies with attributes (CUDA 13.2)
## Summary

CUDA 13.2 adds attribute-carrying variants of single copies:

- `cuMemcpyWithAttributesAsync(CUdeviceptr dst, CUdeviceptr src, size_t size, CUmemcpyAttributes* attr, CUstream)`
  — flat 1D copy taking the same `CUmemcpyAttributes` (access-order and location hints)
  introduced with the 12.8 batched-copy machinery;
- `cuMemcpy3DWithAttributesAsync(CUDA_MEMCPY3D_BATCH_OP* op, unsigned long long flags, CUstream)`
  — executes a single 3D batch-op descriptor.

`cuda.core`'s copy APIs (`Buffer.copy_to` / `copy_from`) accept no attributes today. #1333
tracks the *batched* entry points (`cuMemcpyBatchAsync` — its P1 discard/prefetch half is
already implemented in
[`_memory/_managed_memory_ops.pyx`](https://github.com/NVIDIA/cuda-python/blob/c000331de6c37aa4565af74b001271ffcf6d5c99/cuda_core/cuda/core/_memory/_managed_memory_ops.pyx));
the attribute plumbing is shared, so this issue should be implemented in coordination with
#1333.

## Underlying C APIs to cover

`cuMemcpyWithAttributesAsync`, `cuMemcpy3DWithAttributesAsync` (13.2); reuses `CUmemcpyAttributes`
/ `CUDA_MEMCPY3D_BATCH_OP` (12.8).

## Design sketch (draft — needs design-meeting review)

> [!IMPORTANT]
> Starting point only, **not** a settled design — review in the cuda.core design meeting,
> together with #1333.

```python
attrs = CopyAttributes(src_access_order="stream", ...)   # names TBD; mirrors CUmemcpyAttributes
buf.copy_to(dst, stream=s, attributes=attrs)
```

The same `CopyAttributes` object would feed #1333's batched API.

Open questions for the meeting:

1. Dataclass shape mirroring `CUmemcpyAttributes` (StrEnum for access order; `Device`/`Host`
   objects for location hints).
2. Whether/how to expose the 3D variant (StridedMemoryView-based copies?).
3. Separate PR or folded into #1333's implementation.
4. Version gating (13.2+) with actionable errors.

## References

- Driver docs: https://docs.nvidia.com/cuda/cuda-driver-api/
- Found during the CUDA 12.8 → 13.3 bindings vs. cuda.core gap sweep (2026-07-14)

-- Leo's bot

juenglin:
## Summary

Adds an optional `options` keyword argument to `Buffer.copy_to` and `Buffer.copy_from`, exposing the same `CopyOptions` dataclass introduced by `copy_batch` (#2593) on the per-buffer path. Also fixes two issues in `copy_batch` itself, uncovered while aligning the two APIs: it now accepts `PER_THREAD_DEFAULT_STREAM` (previously rejected, even though the driver accepts it), and it now raises `RuntimeError` for `src_access_order=DURING_API_CALL` instead of taking the pre-CUDA-13 fallback path because the fallback cannot emulate the synchronization behavior that `cudaMemcpyBatchedAsync` guarantees. Closes #2365.

## Public API

```python
from cuda.core.utils import CopyOptions, MemcpySrcAccessOrder, MemcpyOverlapMode
from cuda.core import Device, Host

# Source access ordering
buf.copy_to(dst, stream=stream, options=CopyOptions(src_access_order=MemcpySrcAccessOrder.ANY))

# Overlap hint
buf.copy_to(dst, stream=stream, options=CopyOptions(overlap_mode=MemcpyOverlapMode.PREFER_OVERLAP_WITH_COMPUTE))

# Location hints (honored for managed memory; silently ignored otherwise)
buf.copy_to(dst, stream=stream, options=CopyOptions(
    src_access_order=MemcpySrcAccessOrder.STREAM,
    src_location_hint=Device(0),
    dst_location_hint=Host(),
))
```

`options=None` (the default) selects the existing `cuMemcpyAsync` path with no change in behaviour.

## Behaviour

- When `options` is set, `cuMemcpyWithAttributesAsync` is used only when `cuda.bindings` and the driver are both CUDA 13.2 or newer, the stream is not currently capturing, and the stream is not `LEGACY_DEFAULT_STREAM`.
- Passing `options` together with `LEGACY_DEFAULT_STREAM` or a stream currently in graph capture mode raises `TypeError`, matching `copy_batch`. These checks only apply when `options` is given; `options=None` never triggers them, so existing code using either a default-stream token or graph capture is unaffected.
- On an older `cuda.bindings`/driver install, `options.src_access_order` values of `STREAM` and `ANY` silently fall back to plain `cuMemcpyAsync` (that fallback already provides stream-ordered access, which is what either value promises). `DURING_API_CALL` promises all source reads complete *before the call returns*; a stream-ordered fallback cannot honor that, so it raises `RuntimeError`. `copy_batch`'s own pre-CUDA-13 fallback loop had the same hazard and is fixed the same way in this PR (see "Relation to `copy_batch`" below).
- `options=None` never raises for stream/capture reasons and always uses `cuMemcpyAsync`, so existing code is unaffected.
- `dst=None` auto-allocation remains valid with options.

## Relation to `copy_batch`

Uses the same `CopyOptions` dataclass, the shared `_reject_unsupported_during_api_call` helper (`_copy_enums.py`) that guards the one hazardous fallback case (`DURING_API_CALL`), and the `_to_cu_memcpy_attributes` conversion helper (in the shared `_copy_attributes` module extracted in the prerequisite refactor PR).

While aligning the two APIs, this PR also fixes `copy_batch` itself:
- It previously rejected `PER_THREAD_DEFAULT_STREAM` unconditionally, even though `cuMemcpyBatchAsync` accepts it just like `Buffer.copy_to`/`copy_from` now do; it now accepts it.
- Its pre-CUDA-13 fallback loop silently ignored all `CopyOptions`, including `DURING_API_CALL`, despite a comment claiming otherwise. It now raises `RuntimeError` for `DURING_API_CALL` on that path, via the same shared `_reject_unsupported_during_api_call` helper.

The single-copy path with options now matches `copy_batch`'s rejection and fallback behavior in every respect but one:

| Condition | `copy_batch` | `Buffer.copy_to`/`copy_from` |
|---|---|---|
| Old driver/`cuda.bindings` | Silent fallback, options ignored (except `DURING_API_CALL` → `RuntimeError`) | Silent fallback, options ignored (except `DURING_API_CALL` → `RuntimeError`) |
| Graph capture | `TypeError` | `TypeError` (only when `options` is given) |
| `LEGACY_DEFAULT_STREAM` | `TypeError` | `TypeError` (only when `options` is given) |
| `PER_THREAD_DEFAULT_STREAM` | Accepted, honored | Accepted, honored |

The remaining difference is intentional: `copy_batch` is a brand-new function with no backward-compatibility constraint, so it rejects `LEGACY_DEFAULT_STREAM`/graph capture unconditionally. `Buffer.copy_to`/`copy_from` are existing APIs; `options=None` must not change behavior for existing callers, so it preserves the pre-#2365 behavior of accepting any default-stream token and remaining capturable into a graph.

## Implementation notes

`_buffer.pyx` cimports the new `_with_attributes_available` CUDA 13.2 gate from that `_copy_attributes`; `_copy_ops.pyx` uses its own separate `_batch_entry_point_available` gate (CUDA 13.0), since `copy_batch` and the per-buffer path have different minimum-version requirements.

`cydriver.cuMemcpyWithAttributesAsync` is absent from `cuda.bindings` older than 13.2, so `_buffer.pyx` cannot cimport it directly without breaking builds against older `cuda.bindings`. It is instead routed through a small C++ function-pointer shim (`memcpy_with_attributes_async` in `_cpp/resource_handles.{cpp,hpp}`), resolved at runtime, mirroring the existing `sm_resource_split` (13.1+) shim pattern.

## Tests

`tests/memory/test_copy_single_options.py` covers:
- Data correctness for `copy_to` and `copy_from` with `options=None` and each of the three `MemcpySrcAccessOrder` values.
- `overlap_mode=PREFER_OVERLAP_WITH_COMPUTE` does not corrupt data.
- `DURING_API_CALL` is honored on the native path, or raises `RuntimeError` on the fallback path, depending on the installed driver/`cuda.bindings`.
- `LEGACY_DEFAULT_STREAM` with options raises `TypeError`; `PER_THREAD_DEFAULT_STREAM` with options is accepted and honored.
- `options=None` continues to work under graph capture and with `LEGACY_DEFAULT_STREAM`, unaffected by the new checks.
- `options` with a capturing stream raises `TypeError` (both methods).
- Rejects non-`CopyOptions` values passed as `options`.
- Device + host location hints with managed memory (`src_location_hint`, `dst_location_hint`).
- NUMA location hints (`Host(numa_id=...)`, `Host.numa_current()`), skipped if the system reports no NUMA topology.
- `dst=None` auto-allocation with options.

`tests/memory/test_copy_batch.py` and `test_copy_batch_options.py` cover the `copy_batch` fixes:
- `PER_THREAD_DEFAULT_STREAM` is accepted and options are honored on it; `LEGACY_DEFAULT_STREAM` is still rejected.
- `DURING_API_CALL` raises `RuntimeError` on the fallback path, matching the per-buffer behavior.
- Direct unit tests of the shared `_reject_unsupported_during_api_call` helper.

`tests/test_memory.py` adds coverage for `Buffer.copy_to`/`copy_from` rejecting mismatched source/destination buffer sizes (a pre-existing check that had no test coverage).

### DIFF
--- a/cuda_core/cuda/core/_cpp/resource_handles.hpp
+++ b/cuda_core/cuda/core/_cpp/resource_handles.hpp
@@ -146,6 +146,15 @@ extern decltype(&cuDevSmResourceSplit) p_cuDevSmResourceSplit;
 extern void* p_cuDevSmResourceSplit;
 #endif
 
+// cuMemcpyWithAttributesAsync (13.2+ — may be null on older drivers/bindings)
+#if CUDA_VERSION >= 13020
+extern decltype(&cuMemcpyWithAttributesAsync) p_cuMemcpyWithAttributesAsync;
+#else
+// cuMemcpyWithAttributesAsync doesn't exist in CUDA < 13.2 headers, so use a
+// void* placeholder. The pointer is always null when built against older CUDA.
+extern void* p_cuMemcpyWithAttributesAsync;
+#endif
+
 // ============================================================================
 // NVRTC function pointers
 //
@@ -1109,5 +1118,22 @@ CUresult sm_resource_split(CUdevResource* result, unsigned int nbGroups,
 
 // Returns true if the cuDevSmResourceSplit function pointer is available.
 bool has_sm_resource_split() noexcept;
+
+// ============================================================================
+// cuMemcpyWithAttributesAsync wrapper (13.2+)
+//
+// Calls through p_cuMemcpyWithAttributesAsync if available, otherwise returns
+// CUDA_ERROR_NOT_SUPPORTED. This avoids a direct Cython cimport of the
+// cydriver cdef function, which would fail at module init on cuda-bindings
+// < 13.2 (see https://github.com/NVIDIA/cuda-python/issues/2063).
+// ============================================================================
+
+// attr is void* so the Cython declaration doesn't reference CUmemcpyAttributes
+// (absent from cuda-bindings built against CUDA < 12.8). The C++ side casts it.
+CUresult memcpy_with_attributes_async(CUdeviceptr dst, CUdeviceptr src, size_t size,
+                                       void* attr, CUstream hStream);
+
+// Returns true if the cuMemcpyWithAttributesAsync function pointer is available.
+bool has_memcpy_with_attributes_async() noexcept;
 
 }  // namespace cuda_core
--- a/cuda_core/cuda/core/_memory/_buffer.pyx
+++ b/cuda_core/cuda/core/_memory/_buffer.pyx
@@ -27,13 +27,17 @@ from cuda.core._resource_handles cimport (
 )
 from cuda.core.typing import DevicePointerType
 
-from cuda.core._stream cimport Stream, Stream_accept, default_stream
+from cuda.core._memory._copy_attributes cimport _with_attributes_available
+from cuda.core._memory._copy_attributes cimport _to_cu_memcpy_attributes  # no-cython-lint
+from cuda.core._stream cimport Stream, Stream_accept, Stream_is_default_token, default_stream
 from cuda.core._utils.cuda_utils cimport HANDLE_RETURN, _parse_fill_value
 
 import sys
+import warnings
 from collections.abc import Sequence
 from typing import TYPE_CHECKING
 
+from cuda.core._memory._copy_enums import CopyOptions
 from cuda.core._utils.pycompat import BufferProtocol
 from cuda.core._dlpack import classify_dl_device, make_py_capsule
 from cuda.core._device import Device
@@ -157,6 +161,29 @@ cdef inline void _init_memory_attrs(Buffer self):
     if not self._mem_attrs_inited.load(memory_order_acquire):
         _query_memory_attrs(self._mem_attrs, as_cu(self._h_ptr))
         self._mem_attrs_inited.store(True, memory_order_release)
+
+
+cdef bint _stream_is_capturing(Stream s):
+    cdef cydriver.CUstreamCaptureStatus cap_status
+    IF CUDA_CORE_BUILD_MAJOR >= 13:
+        HANDLE_RETURN(cydriver.cuStreamGetCaptureInfo(as_cu(s._h_stream), &cap_status,
+                                                     NULL, NULL, NULL, NULL, NULL))
+    ELSE:
+        HANDLE_RETURN(cydriver.cuStreamGetCaptureInfo(as_cu(s._h_stream), &cap_status,
+                                                     NULL, NULL, NULL, NULL))
+    return cap_status == cydriver.CU_STREAM_CAPTURE_STATUS_ACTIVE
+
+
+cdef void _do_copy_with_attributes(
+    cydriver.CUdeviceptr dst, cydriver.CUdeviceptr src, size_t nbytes,
+    object options, cydriver.CUstream hstream,
+):
+    IF CUDA_CORE_BUILD_MAJOR >= 13:
+        cdef cydriver.CUmemcpyAttributes cu_attr = _to_cu_memcpy_attributes(options)
+        with nogil:
+            HANDLE_RETURN(cydriver.cuMemcpyWithAttributesAsync(dst, src, nbytes, &cu_attr, hstream))
+    ELSE:
+        pass  # unreachable: _with_attributes_available() is always False on CUDA 12
 
 
 cdef class Buffer:
@@ -393,7 +420,8 @@ cdef class Buffer:
         self.close()
         return False
 
-    def copy_to(self, dst: Buffer | None = None, *, stream: Stream | GraphBuilder) -> Buffer:
+    def copy_to(self, dst: Buffer | None = None, *, stream: Stream | GraphBuilder,
+                options: CopyOptions | None = None) -> Buffer:
         """Copy from this buffer to the dst buffer asynchronously on the given stream.
 
         Copies the data from this buffer to the provided dst buffer.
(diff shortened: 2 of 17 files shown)


cuda_core/docs/source/release/1.2.0-notes.rst
```diff
@@ -9,29 +9,24 @@
...
+  overlap mode) for buffer-to-buffer copies. The new
+  :func:`utils.copy_batch` accepts it and submits many copies in a single
+  ``cuMemcpyBatchAsync`` call, requiring ``cuda.core`` built against CUDA 13
+  plus ``cuda.bindings``/driver 13.0 or newer. :meth:`Buffer.copy_to` and
+  :meth:`Buffer.copy_from` also accept it now, as a new ``options`` keyword,
+  submitting a single copy via ``cuMemcpyWithAttributesAsync`` and requiring
+  ``cuda.bindings``/driver 13.2 or newer. Both reject
+  ``LEGACY_DEFAULT_STREAM`` with ``TypeError`` (``PER_THREAD_DEFAULT_STREAM``
+  is accepted); ``copy_batch`` always rejects graph capture, while
+  ``Buffer.copy_to``/``copy_from`` reject it only when ``options`` is given.
```
Andy-Jost, line 22: If a future version of CUDA adds graph support, this will need to be relaxed. Is the layered check for graph capture only duplicating what the driver already checks, or does it have a separate benefit? If we're trading extra logic and correctness risk only for a nicer error message it might not be worth it.

cuda_core/cuda/core/_memory/_buffer.pyx
```diff
@@ -159,6 +166,73 @@ cdef inline void _init_memory_attrs(Buffer self):
...
+            "(matches copy_batch); cuMemcpyWithAttributesAsync rejects it outright, "
+            "unlike PER_THREAD_DEFAULT_STREAM, which is a real stream to the driver "
+            "and is accepted. Pass an explicit stream, PER_THREAD_DEFAULT_STREAM, "
+            "or options=None."
+        )
+    if _stream_is_capturing(s):
+        raise TypeError(
+            f"{method_name} does not support graph capture with options "
+            "(matches copy_batch); use GraphNode.memcpy to build attributed copies "
+            "into a graph, or pass options=None."
```
Andy-Jost: I don't think `GraphNode.memcpy` supports attributed copies?

cuda_core/cuda/core/_cpp/resource_handles.hpp
```diff
@@ -146,6 +146,15 @@ extern decltype(&cuDevSmResourceSplit) p_cuDevSmResourceSplit;
...
 #endif
 
+// cuMemcpyWithAttributesAsync (13.2+ — may be null on older drivers/bindings)
+#if CUDA_VERSION >= 13020
+extern decltype(&cuMemcpyWithAttributesAsync) p_cuMemcpyWithAttributesAsync;
+#else
+// cuMemcpyWithAttributesAsync doesn't exist in CUDA < 13.2 headers, so use a
+// void* placeholder. The pointer is always null when built against older CUDA.
+extern void* p_cuMemcpyWithAttributesAsync;
+#endif
```
Andy-Jost, line 156: I have concerns about (1) minor-version-specific version gating, and (2) scope-creep inherent in using `resource_handles` for things besides resource lifetime management. I see this merely extends an existing pattern, so neither of these should block merge; but I'd like to follow up.

Andy-Jost (approved):

cuda_core/docs/source/release/1.2.0-notes.rst
juenglin: Yes, if support for this becomes available the check for capturing streams would be dropped.

And you are right, the explicit check is to reject a capturing stream with a `TypeError` instead of a `CUDAError`, consistent with `copy_batch`.

cuda_core/cuda/core/_memory/_buffer.pyx
juenglin: You're right, thanks.

juenglin: Fixed in commit 9e7b50a.

PR #2588, opened 2026-08-10
Merge main into 13.4.x

rwgk:
## Description

Merge the current `main` history into `13.4.x` to establish a shared baseline for the next release-preparation update.

The corresponding release changes will follow in a separate PR.

PR #2623, opened 2026-08-14
CI: Add native Windows ARM64 testing

rwgk:
## Description

Closes #2622.

Adds native Windows ARM64 GPU testing for the public 13.4.x workflow. The new
job uses one representative Python 3.13 / CUDA 13.4 configuration on the
`windows-arm64-1xgb10` runner and consumes the existing public wheel artifacts.

The workflow now:

- publishes the compiled bindings/core test artifacts needed by the
  single-CUDA-major Windows ARM64 build;
- verifies ARM64 runner hardware, GPU access, native Python, and native compiled
  extension imports;
- runs pathfinder, bindings with both legacy and per-thread default streams,
  bindings Cython tests, and core tests; and
- includes the WoA build and test jobs in the final CI aggregation without
  weakening the existing event, owner, permission, or driver safeguards.

The WoA path remains prerelease-gated, matching the existing Windows ARM64
wheel-build policy. This draft is based on the continuation of #2588 and
therefore also carries the latest main-to-13.4.x merge; it should remain draft
while the first public WoA runner execution validates the experimental job.

## Testing

- Changed-file pre-commit suite, including YAML validation and actionlint
- `git diff --check`
- Bash syntax validation for `ci/tools/run-tests`
- PowerShell parser validation for the Windows ARM64 bootstrap
- Independent workflow, artifact-contract, security, and private-prior-art reviews

Native WoA execution is intentionally left to the draft PR CI run.

## Checklist

- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #2603, opened 2026-08-11
Catch up to current generator main

mdboom:
This catches up the generated files to the current main.

PR #2641, opened 2026-08-15
Continue merging main into 13.4.x

rwgk:
EDIT (for my own reference mostly): I'm freezing this PR after the internal **530** was merged.
___

## Description

Follow-up to #2588.

Merge the newer public `main` history into the 13.4.x release line on top of the baseline established by #2588.

This continuation also repairs merge accidents exposed by the initial CI run:

- temporarily skips the pixi CUDA-version check because CUDA 13.4 Conda packages are not yet available;
- regenerates the CUDA 13.4.0 bindings with the current cybind generator fixes while preserving the 13.4 API surface;
- avoids aggregate `CUmemLocation` initialization now that CUDA 13.4 adds the `localized` member; and
- regenerates the merged cuda.core stubs.

The `13.4.x-merge-main-cont` branch is intentionally also pushed to the upstream repository so it can serve as the base for another main-merge continuation stacked on top of this PR.

## Testing

- CUDA 13.4 bindings/core build runbook
- CUDA 13.4 pathfinder, bindings default-stream and PTDS, bindings Cython, cuda.core, and cuda.core Cython test runbooks
- generated-file seals and cuda.core stub generation
- pre-commit suite, including Ruff, mypy, actionlint, Cython lint, YAML, and secret scanning

## Checklist


rwgk:
codex gpt-5.6-sol ultra:

## Noteworthy merge-main nuances

I want to record the provenance of d31a58c and 5830ba7, because the two fixes appeared one after the other during testing but address independent problems. This also gives us concrete expectations for the next ctk-next merge/transfer-preview cross-check.

### Why the cuDLA and `CUmemLocation` failures appeared sequentially

The merge commit a6ca4f brought the later public `main` into the branch from PR #2588. Public `main` included the current-generator update from #2603 (ffb776c) as well as the change that treats Cython warnings as errors (8457bb0). The generated bindings conflicted because the two sides represented different toolkit surfaces: the release branch had CUDA 13.4 output, whereas public `main` still had CUDA 13.3 output. Keeping the 13.4 files was necessary, but keeping those files wholesale also accidentally discarded the toolkit-independent generator fixes from #2603.

The first visible result was the Linux build failure seen while testing #2623. The stale 13.4 cuDLA output still put `except *` on the Python-object-returning `device_get_attribute` and `module_get_attributes` wrappers. Current Cython reports that exception clause as invalid/ignored, and the project's warnings-as-errors setting makes the diagnostic fatal. Every affected job therefore stopped while building `cuda.bindings`.

Commit d31a58c repairs that merge accident in the release-appropriate way: it regenerates from the CUDA 13.4 inputs using the current cybind generator. It does not copy CUDA 13.3 generated files from `main`. The result retains the CUDA 13.4 API surface while recovering the generic generator fixes from #2603, including removal of the invalid cuDLA exception clauses.

Only after that repair could the build move on to `cuda.core`. The wheel workflow builds and installs `cuda.bindings` before it Cythonizes `cuda.core`; the earlier cuDLA error prevented the latter phase from running at all. The next local build therefore became the first build in this sequence to reach the `CUmemLocation` sites, where Cython reported:

```text
Not all members given for struct 'CUmemLocation'
```

This is what I consider a *latent public 13.4.x problem*. The incompatible declaration and initialization syntax were already together in the PR #2588 base (9d4142), before either d31a58c or 5830ba7. PR #2588 did not exercise this full compilation path; its available checks were metadata/label checks, while the relevant build coverage was not run. Regenerating the bindings did not add the `localized` member and did not cause the `cuda.core` problem. It merely removed the earlier blocker and allowed the sequential build to reveal it.

### Where 5830ba7 comes from

Public `main`'s CUDA 13.3-generated declaration of `CUmemLocation` had two visible members, `type` and `id`. In that layout, the aggregate initializer introduced by #2434 was complete:

```cython
cydriver.CUmemLocation(type=..., id=...)
```

CUDA 13.4 adds a `localized` arm to the anonymous location union. The generated Cython declaration consequently exposes `type`, `id`, and `localized`. Cython 3.2 requires an aggregate initializer to name every member in its declaration, so all of the two-member aggregate initializers fail against the 13.4 declaration.

The correct source-compatible form is to declare the struct and assign only the active fields directly:

```cython
cdef cydriver.CUmemLocation location
location.type = ...
location.id = ...
```

All location kinds currently constructed by `cuda.core` use the `id` arm (`DEVICE`, `HOST`, `HOST_NUMA`, or `HOST_NUMA_CURRENT`). The `localized` arm belongs to `DEVICE_LOCALITY_DOMAIN`, which these paths do not construct. Direct field assignment is therefore both semantically correct for CUDA 13.4 and compatible with the older two-member generated declaration.

This was first repaired on the ctk-next side by ctk-next commit cf9491a54ed ("Avoid aggregate CUmemLocation initialization"). Public commit #2593 (29acb74) subsequently refactored the shared conversion into the new `_memory/_location.pxd`. Commit 5830ba7 is the same ctk-next fix adapted to that newer public tree: it preserves the original five behavioral repairs and also updates the new shared helper. That is why 5830ba7 touches six files while the original ctk-next change touched five.

### Why ctk-next being in sync did not already put this fix in public

The direction of the two companion ctk-next PRs matters:

- ctk-next PR 518 was the authoritative merge view. It brought the public history represented by PR #2588 into ctk-next while retaining ctk-next-only work, including ctk-next commit cf9491a54ed.
- ctk-next PR 519 was the companion public-shaped transfer preview, based exactly on PR #2588 at 9d414267977. Its non-generated transfer commit was 7ca3fb5a855; five of its ten changed files carried the original `CUmemLocation` repair.

PR 519 was a review artifact and was closed-as-planned rather than merged into public. Thus ctk-next contained the public PR #2588 history, but the ctk-next-only correction had only been *previewed* in the reverse direction. It had not reached the public 13.4.x branch. PR #2641 now carries the adapted public-tree version in 5830ba7.

### Cross-check for the next ctk-next merge and preview

After merging the exact PR #2641 head into ctk-next and regenerating the CUDA 13.4.1 outputs, I plan to create both the authoritative merge PR and its public-shaped transfer preview. The old PR 519 transfer commit 7ca3fb5a855 gives us a useful independent reference.

For the new non-generated transfer commit, the expected comparison is:

- The five old `CUmemLocation` hunks, and the newer `_memory/_location.pxd` hunk, should **not** appear. They are now already present in the public preview base through 5830ba7.
- The other five substantive ctk-next-only changes from 7ca3fb5a855 should remain unless public has independently acquired them: `cuda_bindings/tests/cython/build_tests.bat`, `cuda_bindings/tests/cython/build_tests.py`, `cuda_core/tests/example_tests/test_basic_examples.py`, `cuda_core/tests/system/test_system_system.py`, and `cuda_core/tests/test_green_context.py`.
- Generated bindings and generated stubs should be compared separately. The public branch uses CUDA 13.4.0 output, whereas the new ctk-next merge and preview will regenerate based on a newer CUDA 13.4. The strong check there is byte-for-byte agreement between the new merge view and the new preview view, not identity with either old PR 519 or public d31a58c.
- The comparison should be semantic/path-by-path rather than expecting the entire new transfer commit to have the same patch ID as 7ca3fb5a855, because public #2593 changed the source topology and later public work changes surrounding context.

If a `CUmemLocation` change reappears in the new transfer commit, that would indicate either a stale preview base or a merge resolution that lost 5830ba7. If one of the unrelated five ctk-next-only changes disappears without an equivalent public change, that would indicate that the merge lost ctk-next work. Unexplained new non-generated files would similarly point to a base, conflict-resolution, or transfer-attribution problem.

With d31a58c, 5830ba7, and the regenerated stubs applied, the full local CUDA 13.4 build and test runbooks pass: pathfinder, bindings in both default-stream modes, bindings Cython tests, cuda.core, and cuda.core Cython tests. This confirms that the two commits remove successive blockers without replacing or narrowing the CUDA 13.4 API surface.

rwgk:
I added two follow-up commits based on the failures in CI run 31865723802.

### Restore compatibility coverage for clock event reasons (df5356dddd2)

All 11 failing Linux AArch64 test jobs failed at the same enum-coverage case. `cuda.core` exposes `ClocksEventReasons.BOARD_LIMIT` and `ClocksEventReasons.RELIABILITY`, but the older `cuda.bindings` versions used by the CUDA 12.9 and 13.3 compatibility rows do not yet expose the corresponding NVML enum names.

The production mapping already handles this intentionally: when those two names are absent, it uses their stable numeric NVML values. The generic coverage test had lost the matching compatibility allowance, so it incorrectly required an older binding to contain names that only a newer binding provides.

This commit restores the narrow, dynamic test guard from ctk-next commit 91aa75d891a:

- each wrapper member is exempted only when its corresponding binding member is absent;
- with CUDA 13.4 bindings, both names are present, the exemption set is empty, and the normal full-coverage assertion still applies;
- no runtime behavior and no other enum coverage is changed.

This is therefore a compatibility-test correction rather than a blanket skip. It lets the intended older-bindings matrix continue checking all members that those bindings can actually provide.

### Harden prior-branch artifact lookup (1ebebc533f8)

Two Linux wheel-build jobs failed after their current CUDA 13.4 builds had succeeded, while assembling the CUDA 12 half of the combined `cuda.core` wheel. Each matrix job independently asked `gh run list` for one successful run on `12.9.x` and assumed that result was the newest usable run.

That assumption did not hold in this CI run: jobs started within roughly the same minute resolved several different historical run IDs. The Python 3.14t job selected a run whose matching artifact had expired, while the Python 3.15 job selected a run from before that artifact existed. The logs do not establish which GitHub/CLI layer produced the stale ordering, so this change does not depend on attributing the behavior to a specific external bug.

The lookup now:

- fetches a bounded set of completed branch runs, filters successful runs explicitly, and sorts them by creation time;
- scans newest to oldest and uses the paginated Actions artifact API to require a matching, unexpired bindings wheel before selecting a run;
- also requires `cuda-python-wheel` when a test job will download it, so all requested artifacts must come from the same run;
- distinguishes the bindings wheel artifact from its `-tests` sibling, preventing a tests-only match from qualifying a run;
- preserves branch lookup without artifact requirements, including the nightly `run_id` + `head_sha` output, and leaves tag lookup unchanged.

All three prior-branch consumers now use the validated lookup: the wheel builder and the Linux and Windows wheel-test workflows.

The new standalone tests use a fake `gh` executable to cover unordered and failed runs, a workflow selected by filename rather than display name, fallback past expired/missing artifacts, multiple artifacts required from one run, a live `-tests` artifact with an expired wheel, and artifact-API failure propagation.

### Validation

- `cuda_core/tests/test_enum_coverage.py`: 30 passed with `--noconftest`
- `ci/tools/tests`: 60 passed with `--noconftest`
- `bash -n` and standalone `shellcheck` for `ci/tools/lookup-run-id`
- actionlint, YAML, Ruff, SPDX, and `git diff --check`
- full `pre-commit run --all-files` passed with `check-pixi-cuda-version` intentionally skipped because CUDA 13.4 Conda packages are not available yet

The CI rerun remains the authoritative cross-version check: the local environment has CUDA 13.4 bindings, whereas the previously failing rows exercise the conditional enum guard with CUDA 12.9/13.3 bindings and the real prior-branch artifact inventory.

rwgk:
The CI failures are unrelated to this PR. Ignoring for now.

[CI / Docs / Build docs (push)](https://github.com/NVIDIA/cuda-python/actions/runs/31870063964/job/94978076797?pr=2641)
[CI / Test linux-64 / Python 3.15, CUDA 13.3.0 (wheels), GPU l4 (push)](https://github.com/NVIDIA/cuda-python/actions/runs/31870063964/job/94978092729?pr=2641)
[CI / Test linux-64 / Python 3.15t, CUDA 13.3.0 (wheels), GPU l4 (push)](https://github.com/NVIDIA/cuda-python/actions/runs/31870063964/job/94978092705?pr=2641)

rwgk:
Closing after PR #2788 was merged.

PR #2643, opened 2026-08-15
CI: Harden prior-branch artifact lookup

### LINKED ISSUE #2642: [BUG]: CI: Harden prior-branch artifact lookup against unstable run ordering and unavailable artifacts
## Problem

CUDA Python CI reuses `cuda.bindings` wheel artifacts from the configured prior-major branch when it builds and tests the combined `cuda.core` wheel. For example, current `main` and `13.4.x` CI use artifacts produced by successful `12.9.x` runs.

Until now, branch-mode lookup in `ci/tools/lookup-run-id` effectively does this:

```bash
gh run list --branch 12.9.x --limit 1 --workflow CI --status success
```

It then attempts to download the Python- and platform-specific wheel artifact from the one returned run. This relies on two independent assumptions:

1. The first run returned by the branch-filtered GitHub query is the newest successful run we intended to use.
2. That run produced every artifact needed by the current matrix row, and those artifacts are still available rather than missing or expired.

A successful workflow conclusion does not establish either property. The lookup should select a run that is both successful and usable for the requesting consumer.

## How we found this

The problem surfaced while testing the `13.4.x` merge-main repair in [PR #2641](https://github.com/NVIDIA/cuda-python/pull/2641), in [CI run 31865723802](https://github.com/NVIDIA/cuda-python/actions/runs/31865723802).

The current CUDA 13.4 builds succeeded. The failures occurred later, when the Linux wheel jobs tried to obtain the CUDA 12 bindings wheel needed for the prior-major half of the combined `cuda.core` wheel.

Eight Linux matrix jobs made the same prior-branch lookup within about 68 seconds, but received four different successful `12.9.x` run IDs. Most happened to remain usable, so those jobs passed this step. Two did not:

- Python 3.14t selected [run 22911991598](https://github.com/NVIDIA/cuda-python/actions/runs/22911991598), whose matching wheel artifact had expired.
- Python 3.15 selected [run 23885913601](https://github.com/NVIDIA/cuda-python/actions/runs/23885913601), which predated Python 3.15 artifact production and therefore never contained the requested wheel.

By contrast, a nearby successful [`main` run 31848112319](https://github.com/NVIDIA/cuda-python/actions/runs/31848112319) exercised the same old helper in all eight Linux rows, but every lookup happened to return the current usable `12.9.x` run. This explains why the weakness had generally remained latent on `main`: both assumptions usually held, and even some stale runs still had usable artifacts.

This is generic cross-release CI infrastructure, not a CUDA 13.4-specific problem. The same helper and prior-branch consumers exist on `main`, which also currently obtains previous-major bindings from `12.9.x`.

## Scope

Harden branch-mode lookup so that it selects the newest qualifying run from a bounded candidate set instead of trusting one unvalidated result.

The desired behavior is:

- Retrieve a bounded set of completed runs for the requested branch and workflow.
- Filter for the exact branch and a successful conclusion, then sort explicitly by creation time with a deterministic run-ID tie-breaker.
- Let callers provide one or more required artifact-name patterns.
- For each candidate, verify that every required artifact exists in that same run and is not expired.
- Skip candidates with missing or expired requirements and continue to the next candidate.
- Fail with a useful diagnostic when no returned successful run has a complete usable artifact set; also fail rather than silently falling back if run or artifact metadata cannot be queried.
- Update every prior-branch wheel consumer to request the artifact set it will actually download, including `cuda-python-wheel` when that metapackage is tested.
- Preserve branch lookup without artifact requirements, `--head-sha` output, and tag-mode lookup.

This makes the selected run auditable and makes the artifact contract explicit at the lookup boundary.

## Current non-goals

This issue does **not** attempt to guarantee that GitHub's branch-filtered response contains the globally newest matching run.

The first implementation still uses `gh run list --branch ...` and examines at most 100 returned candidates. Public reports show that the branch filter can be backed by a temporarily stale or incomplete index. Local sorting can make the returned set deterministic, but it cannot recover a run that GitHub omitted from that set, nor does the bounded search exhaust older history.

Fully eliminating that remaining source of luck would require a separate design, such as retrieving suitably paginated workflow runs without the server-side branch filter and applying branch selection locally. That has different cost, pagination, and bounding questions and is intentionally deferred.

Also out of scope:

- Redesigning CUDA Python to avoid cross-run artifact reuse altogether.
- Changing artifact retention policy or recreating artifacts that have already expired.
- Centralizing one prior-run resolution across the entire matrix.
- Changing tag-mode release-run lookup.
- Validating artifact size, archive contents, wheel ABI, or download success during lookup; eligibility here means that the name matches and GitHub reports `expired == false`.
- Retrying artifact-metadata API failures. Those errors should remain visible and fatal rather than being mistaken for an ineligible run.

## Public background

[`NVIDIA/cuda-python#329`](https://github.com/NVIDIA/cuda-python/issues/329), implemented by [PR #357](https://github.com/NVIDIA/cuda-python/pull/357), is the origin of reusing build artifacts from the backport branch. It establishes why this lookup exists, but does not address run ordering or artifact expiration.

For run selection:

- The [`gh run list` manual](https://cli.github.com/manual/gh_run_list) and GitHub's [workflow-runs REST documentation](https://docs.github.com/en/rest/actions/workflow-runs#list-workflow-runs-for-a-workflow) provide filters and pagination, but no caller-selected sort direction or documented newest-first contract.
- [`cli/cli#5440`](https://github.com/cli/cli/issues/5440) reports `gh run list --branch ...` returning a seven-day-old run while an unfiltered query showed matching runs from one and five hours earlier. The [CLI maintainer response](https://github.com/cli/cli/issues/5440#issuecomment-1095245255) identifies a sometimes-stale Elasticsearch index as the likely mechanism and recommends fetching without filters and filtering in the client.
- [`cli/cli#6678`](https://github.com/cli/cli/issues/6678) demonstrates consecutive identical `gh run list` calls changing order, including completed runs. A [maintainer response](https://github.com/cli/cli/issues/6678#issuecomment-1334097577) advises consumers not to rely on stable result order.

These reports establish public precedent for treating server-returned order and filtered-result freshness as fallible. They do not prove the precise GitHub-side mechanism behind run 31865723802.

For artifact eligibility:

- [`NVIDIA/cuda-python#299`](https://github.com/NVIDIA/cuda-python/issues/299) uses the same fail-safe principle for selective-CI baseline hydration. Its linked [PR #2467](https://github.com/NVIDIA/cuda-python/pull/2467) treats missing, expired, or incomplete baseline artifacts as unusable and falls back to the full pipeline.
- [`actions/download-artifact#459`](https://github.com/actions/download-artifact/issues/459) requests an explicit artifact-existence preflight, and [`cli/cli#11610`](https://github.com/cli/cli/issues/11610) documents a workaround that filters repository artifacts for `.expired == false`.
- [`dawidd6/action-download-artifact#240`](https://github.com/dawidd6/action-download-artifact/issues/240) is a close combined analogue: an unexpectedly old successful run was selected despite newer runs, then download failed because the selected artifact had expired.

[`NVIDIA/cuda-python#1759`](https://github.com/NVIDIA/cuda-python/issues/1759) proposes a broader release-specific redesign that removes some cross-run handoffs by building and publishing in one workflow. It does not remove the current need for prior-major artifacts in ordinary wheel CI and is not a substitute for this targeted hardening.

No existing CUDA Python issue tracks the specific prior-branch failure chain addressed here: select a successful candidate, require the consumer's complete unexpired artifact set, and fall back to the next usable candidate.

## Acceptance criteria

- [ ] Branch mode filters for the exact branch and successful conclusion, then sorts returned candidates by creation time and run ID instead of trusting API order.
- [ ] Callers can specify repeatable required artifact patterns, all of which must match unexpired artifacts in the same run.
- [ ] A candidate with a missing or expired requirement is skipped in favor of the next returned candidate.
- [ ] The wheel build and Linux/Windows wheel-test consumers require the exact prior-branch artifacts they subsequently download.
- [ ] Failure diagnostics identify skipped candidates and missing requirements.
- [ ] Existing tag mode, branch mode without artifact requirements, and `--head-sha` output remain compatible.
- [ ] Artifact-list API failures propagate as errors rather than being treated as an ordinary candidate miss.
- [ ] Automated tests cover unsorted/failed/wrong-branch candidates, missing or expired artifacts, multiple simultaneous artifact requirements, fallback, no qualifying result, API failure, and the existing no-artifact branch-mode contract.

rwgk:
## Description

Fixes #2642.

CUDA Python CI reuses `cuda.bindings` wheel artifacts from the configured prior-major branch when it builds and tests the combined `cuda.core` wheel. The existing lookup trusted the first successful run returned by `gh run list` without verifying that the artifacts required by the requesting matrix row were present and unexpired.

This change makes branch-mode lookup choose the newest usable run from a bounded candidate set:

- retrieve up to 100 completed runs, filter for the exact branch and successful conclusion, and sort explicitly by creation time with a run-ID tie-breaker;
- accept repeatable `--artifact` patterns and require every pattern to match an unexpired artifact in the same run;
- skip candidates with missing or expired requirements and continue to the next candidate;
- update the wheel build and Linux/Windows wheel-test consumers to declare the artifacts they subsequently download; and
- preserve tag mode, branch mode without artifact requirements, and `--head-sha` output.

This intentionally does not eliminate the possibility that GitHub's server-side branch filter returns a stale or incomplete bounded candidate set; that broader lookup redesign remains outside #2642.

## Testing

- [x] `TestVenv/bin/python -m pytest -v --noconftest ci/tools/tests/test_lookup_run_id.py` (4 passed)
- [x] `bash -n ci/tools/lookup-run-id`
- [x] `shellcheck ci/tools/lookup-run-id`
- [x] `pre-commit run --all-files`
- [x] PR #2641 [CI run 31870063964](https://github.com/NVIDIA/cuda-python/actions/runs/31870063964?pr=2641) exercised the extracted change across prior-branch wheel-build and Linux wheel-test consumers; every exercised lookup selected a usable run and passed.

## Checklist


ci/tools/lookup-run-id
```diff
@@ -86,28 +100,90 @@ done
...
-        -R "${REPOSITORY}" \
-        --json databaseId \
-        | jq -r '.[0].databaseId // empty')
+    RUN_DATA=$(gh run list \
+        --repo "${REPOSITORY}" \
+        --branch "${REF}" \
+        --workflow "${WORKFLOW_NAME}" \
+        --status completed \
+        --json databaseId,workflowName,status,conclusion,headSha,headBranch,createdAt,url \
+        --limit 100)
```
mdboom, line 109: From Claude:

Branch-mode lookup now fetches --status completed --limit 100 runs (any conclusion: success, failure, cancelled, …) and filters for conclusion == "success" client-side in the jq step, replacing the old query which filtered -s success (i.e. conclusion=success) directly at the GitHub API level with -L 1. I confirmed via gh run list --help that --status/-s accepts both status values (completed, in_progress, …) and conclusion values (success, failure, …) through the same flag — so the old code was already using the more precise "only success" filter, unbounded. The new code is bounded to the 100 most recent completed runs regardless of outcome. If a backport/maintenance branch (exactly the kind of branch this feature targets — see 12.9.x in the new tests) accumulates 100+ failing/cancelled scheduled runs since its last success (plausible for a stagnant branch that isn't actively fixed), the real successful run falls outside the fetched window and the script now reports "No successful run found" even though a usable run exists. This is compounded by the new artifact-existence filtering, which discards additional candidates from that already-truncated window before giving up. This directly undermines the PR's stated goal of "choosing the newest usable run." Fix: use --status success (still bounded by --limit 100, but only counting real successes, matching prior semantics) instead of --status completed.

mdboom (changes requested):

ci/tools/lookup-run-id
rwgk: Thanks, good catch. You're right that `--status completed --limit 100` bounded the search before our client-side success filter, so enough newer failed or cancelled runs could hide an older usable success.

I changed branch-mode lookup to use `--status success`, while retaining the client-side `.conclusion == "success"` check as defense in depth. The 100-result candidate budget now counts successful runs; artifact validation can fall back across those candidates without failed/cancelled runs crowding them out.

I also updated the fake `gh` implementation to model server-side status filtering before the limit and added a regression case with 100 newer failed runs followed by one older successful run. That case fails with `--status completed` and passes with `--status success`.

Fixed in 619e507f879. The focused tests pass (5 tests), as do Bash syntax checking, ShellCheck, and the full pre-commit suite.

mdboom (approved):

PR #2644, opened 2026-08-16
Remove timing-sensitive NVML total-energy test

rwgk:
## Description

Addresses cuda-python-private issue 509.

Removes the timing-sensitive `test_device_get_total_energy_consumption` bootstrap test. NVML does not guarantee a refresh cadence, so requiring the cumulative energy counter to advance within 150 ms can fail even when the binding is working correctly.

## Checklist

- [x] New or existing tests cover these changes. (Not applicable: this removes a timing-sensitive bootstrap test without changing product code.)
- [x] The documentation is up to date with these changes. (No documentation changes are required.)

### DIFF
--- a/cuda_bindings/tests/nvml/test_pynvml.py
+++ b/cuda_bindings/tests/nvml/test_pynvml.py
@@ -4,7 +4,6 @@
 # A set of tests ported from https://github.com/gpuopenanalytics/pynvml/blob/11.5.3/pynvml/tests/test_nvml.py
 
 import os
-import time
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


rwgk:
I also looked into whether removing this test should leave behind a comment such as `# [Skipping] pynvml.nvmlDeviceGetTotalEnergyConsumption`.

I found that the existing `[Skipping]` comments in this file were copied verbatim from the upstream pynvml 11.5.3 test when the file was introduced; they are not maintained as a ledger of tests later removed. As direct precedent, #2483 removed `test_device_get_nvlink_capability` from this same file without adding a corresponding `[Skipping]` comment. I therefore left the removal as-is.

mdboom (approved):

PR #2645, opened 2026-08-16
[no-ci] Clarify policy for explicitly requested upstream pushes

rwgk:
## Description

Follow-up to #2378.

The current policy makes the canonical upstream repository unconditionally read-only. That remains the right default for normal pull-request work, but the absolute wording creates significant friction for a small set of valid, user-directed workflows. In practice, agents have repeatedly asked me to perform a push manually even when I explicitly requested it and upstream was needed to test `.github/workflows`, trigger CI from a designated ref, or support similar infrastructure work.

This change keeps approved forks as the normal destination and permits an upstream push only when the user explicitly requests it and provides a rationale. It also requires verification of the complete repository and exact refspec, limits authorization to the requested ref update, and requires separate explicit authorization for higher-risk operations.

This is not intended to give agents more autonomy. The agent does not infer when an upstream push is appropriate: the user supplies the request and rationale. The instructions only allow the agent to execute that narrowly scoped request instead of requiring a manual handoff.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -14,16 +14,30 @@ guide for package-specific conventions and workflows.
 
 # Pull requests
 
-**Never push branches or commits to the canonical upstream repository. Treat
-it as read-only.** Branch creation and pushes for pull-request work must go to
-an approved fork associated with the contributor. The fork may be owned by the
-contributor's personal account or by an organization.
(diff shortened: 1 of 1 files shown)


isVoid (approved):

isVoid:
Thanks. I also ran into CI update frictions when working with agent recently. This should help.

PR #2647, opened 2026-08-16
pathfinder: make LocatedHeaderDir immutable

### LINKED ISSUE #2646: [BUG]: LocatedHeaderDir is mutable, so callers can poison the cached header-directory lookup
### Component

cuda.pathfinder

### What happened?

`locate_nvidia_header_directory()` is `@functools.cache`-backed and returns `LocatedHeaderDir`, which is a plain (mutable) dataclass. The cache hands every caller the *same* object, so a write to `abs_path` on a returned instance changes what every later lookup of that `libname` returns for the rest of the process, including through the `find_nvidia_header_directory()` wrapper.

`LocatedHeaderDir` is also the only one of the three public `Located*` return types that behaves this way: `LocatedStaticLib` (`_static_libs/find_static_lib.py`) and `LocatedBitcodeLib` (`_static_libs/find_bitcode_lib.py`) are both `@dataclass(frozen=True)`. As a side effect of not being frozen, `LocatedHeaderDir` is also unhashable, so it cannot be put in a set or used as a dict key the way its two siblings can.

### Reproduction

```python
from cuda.pathfinder import locate_nvidia_header_directory, find_nvidia_header_directory

first = locate_nvidia_header_directory("cudart")
print(first.abs_path)                                # e.g. /usr/local/cuda/include

first.abs_path = "/somewhere/else"                   # caller "normalizes" the result in place

print(locate_nvidia_header_directory("cudart").abs_path)  # /somewhere/else
print(find_nvidia_header_directory("cudart"))             # /somewhere/else
```

```python
>>> LocatedHeaderDir.__dataclass_params__.frozen, LocatedHeaderDir.__hash__ is not None
(False, False)
>>> LocatedStaticLib.__dataclass_params__.frozen, LocatedStaticLib.__hash__ is not None
(True, True)
>>> LocatedBitcodeLib.__dataclass_params__.frozen, LocatedBitcodeLib.__hash__ is not None
(True, True)
```

### Suggested fix

Make `LocatedHeaderDir` `@dataclass(frozen=True)` so it matches the other two public `Located*` types. The `__post_init__` path normalization then needs `object.__setattr__`. Nothing inside `cuda_pathfinder` or its tests mutates a `LocatedHeaderDir`, so this is contained to the public type's contract.

aryanputta:
## Description

closes #2646

`locate_nvidia_header_directory()` is `@functools.cache`-backed, so every caller gets the same `LocatedHeaderDir` instance. The dataclass was not frozen, so a write to `abs_path` on a returned object changed what every later lookup of that `libname` returned for the rest of the process, including through `find_nvidia_header_directory()`.

This makes `LocatedHeaderDir` `@dataclass(frozen=True)`, matching `LocatedStaticLib` and `LocatedBitcodeLib`, the other two public `Located*` return types. The `__post_init__` path normalization moves to `object.__setattr__`, which is the standard frozen-dataclass form and keeps the normalization behavior identical. Freezing also makes the type hashable, which its two siblings already were.

This is an intentional behavior change on a public type: code that previously mutated a returned `LocatedHeaderDir` now raises `dataclasses.FrozenInstanceError` instead of silently corrupting the shared cache entry. Nothing in `cuda_pathfinder` or its test suite mutates one, and the two sibling types have always rejected it.

Two tests are added: one drives the real cached lookup, asserts the write is rejected, and asserts a second lookup still returns the discovered directory; one pins hashability and value equality.

## Checklist

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_headers/find_nvidia_headers.py
+++ b/cuda_pathfinder/cuda/pathfinder/_headers/find_nvidia_headers.py
@@ -31,13 +31,13 @@ if TYPE_CHECKING:
 # ---------------------------------------------------------------------------
 
 
-@dataclass
+@dataclass(frozen=True)
 class LocatedHeaderDir:
     abs_path: str | None
     found_via: str
 
     def __post_init__(self) -> None:
-        self.abs_path = _abs_norm(self.abs_path)
+        object.__setattr__(self, "abs_path", _abs_norm(self.abs_path))
 
 
 #: Type alias for a header find step callable.
--- a/cuda_pathfinder/docs/source/release/1.6.1-notes.rst
+++ b/cuda_pathfinder/docs/source/release/1.6.1-notes.rst
@@ -35,6 +35,12 @@ Highlights
   architecture-specific executable targets using the native Windows machine
   architecture.
 
+* Make ``LocatedHeaderDir`` immutable, matching ``LocatedStaticLib`` and
+  ``LocatedBitcodeLib``. ``locate_nvidia_header_directory`` is
+  ``functools.cache``-backed, so writing to a returned object previously
+  changed what every later lookup of that ``libname`` returned. The type is
+  now hashable as well.
+
 Internal maintenance
 --------------------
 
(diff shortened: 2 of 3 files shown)


mdboom (approved):

aryanputta:
The remaining substantive failure is infrastructure-only: the Windows Python 3.14 job could not find a successful `CI` baseline run on `12.9.x`, while the other build and test jobs passed. I cannot rerun the upstream workflow as an external contributor. Could a maintainer rerun the failed jobs if this PR is still queued for merge?

PR #2406, opened 2026-07-22
test(cuda.core): isolate pending-call saturation cleanup

Andy-Jost:
## Summary

Isolate the CPython pending-call saturation probe so its process-global queue cannot contaminate other tests running concurrently under free-threaded Python. Explicitly drain the synthetic no-op callbacks before exercising cuda.core's existing retry path.

## Changes

- Run `test_pending_call_queue_saturation_preserves_cleanup` in a subprocess using the current Python executable.
- Call `Py_MakePendingCalls` after saturating the queue and before creating the second graph attachment that retries cleanup scheduling.
- Keep the cleanup-completion and main-thread-finalization assertions inside the isolated process.

## Test Coverage

- `python -m py_compile cuda_core/tests/graph/test_graph_definition_lifetime.py`
- `pre-commit run --all-files`
- Full CI passed all ten Python 3.14t rows across Linux x86-64, Linux arm64, and Windows.
- Targeted Python 3.14t reruns passed on Linux x86-64 (T4), Linux arm64 (L4), and Windows (A100).

## Related Work

- Fixes #2403.
- Follow-up to #2371.

PR #2239, opened 2026-06-21
fix(pathfinder): remove dead/misleading error handling in platform loaders

aryanputta:
## Summary

Removes two dead error-handling paths in the dynamic-lib platform loaders. Both are unreachable / no-ops today; the Linux one additionally carries a factually wrong message that would mislead anyone debugging a load failure. This aligns with the `cuda_pathfinder` policy of not masking why discovery/loading failed.

## Details

**`load_dl_linux.load_with_system_search`** guarded the resolved path with:

```python
abs_path = abs_path_for_dynamic_library(desc.name, handle)
if abs_path is None:
    raise RuntimeError(f"No expected symbol for libname={desc.name!r}")
```

`abs_path_for_dynamic_library` never returns `None` - it returns a resolved path (`os.path.join(l_origin, basename)`) or raises `OSError` from `dlinfo`. So the branch is unreachable, and its message is wrong: it talks about a missing symbol when the only real failure mode is dlinfo path resolution. Dropped the dead branch so the descriptive `OSError` surfaces directly, and updated the docstring `Raises:` accordingly.

**`load_dl_windows.add_dll_directory`** had:

```python
result = kernel32.AddDllDirectory(dirpath)
if not result:
    # Fallback: just update PATH if AddDllDirectory fails
    pass
```

The `if not result: pass` is a no-op - the PATH update below already runs unconditionally. Removed the dead branch and clarified why PATH is updated regardless.

## Behavior

No behavior change: both removed branches were unreachable or no-ops.

## Testing

`ruff check` and `ruff format --check` (v0.15.9, repo-pinned) pass on both changed files.

PR #2648, opened 2026-08-16
test: share one helper for isolated-subprocess test snippets

### LINKED ISSUE #2412: Add test helper for running a specific test in an isolated subprocess
There are 4 places in the code base that run an individual test in an isolated subprocess (to avoid the side effects that test would cause to other tests).

We should refactor that out into a helper.  Probably after #2384 which organizes the test helpers.

Here is an agent-generated rough start to such a helper:

```
import os
import subprocess
import sys
import pytest

# Define an environment variable key to prevent infinite subprocess loops
SUBPROCESS_ENV_KEY = "PYTEST_RUNNING_IN_SUBPROCESS"

@pytest.fixture
def run_in_subprocess(request):
    """
    Reruns the current test inside an isolated python subprocess.
    Skips original in-process execution if triggered.
    """
    # If this token is set, we are already inside the subprocess. Proceed with test logic.
    if os.environ.get(SUBPROCESS_ENV_KEY) == "1":
        yield
        return

    # Build the exact command to re-run only this specific test node
    cmd = [
        sys.executable,  # Current Python interpreter path
        "-m", "pytest",
        request.node.nodeid,  # Specific ID of the active test case
        "-s"  # Allow output to display
    ]

    # Inject the loop prevention variable into the environment
    env = os.environ.copy()
    env[SUBPROCESS_ENV_KEY] = "1"

    # Spawn the subprocess and block until completion
    result = subprocess.run(cmd, env=env)

    # Propagate the subprocess exit status back to the main pytest suite
    if result.returncode != 0:
        pytest.fail(f"Test failed in subprocess with exit code {result.returncode}")

    # Skip standard in-process execution since it has completed externally
    pytest.skip("Test executed successfully inside subprocess.")
```

aryanputta:
## Description

closes #2412

Five tests run a self-contained snippet in a fresh interpreter with `subprocess.run([sys.executable, "-c", code], ...)`, each to keep a process-global side effect from reaching the rest of the suite: the CPython pending-call queue, the import-time rlcompleter patch, and a deliberate glibc abort. They had drifted apart on which of the surrounding hazards they handled.

The sharpest one is the working directory. `python -c` puts the parent's cwd at the head of the child's `sys.path`, so a run started from `cuda_core/` makes the child import the `cuda/core/` source tree instead of the installed wheel. Two sites work around it and carry a comment explaining it; the trap is rediscovered per site.

`cuda_python_test_helpers.subprocess_runner.run_python_snippet` is now the single place these are handled:

```python
run_python_snippet(code, *, cwd, timeout=None, extra_env=None, unset_env=(), check=True)
```

`cwd` is required rather than defaulted, so the shadowing hazard cannot be forgotten. Output is always captured as text so a failure message can quote both streams. `stdin` is always closed, because `PYTHONINSPECT` keeps the interpreter alive after `-c` and an inherited stdin leaves the implicit REPL waiting.

All five call sites move to it:

- `cuda_core/tests/test_rlcompleter_patch.py` (`_run_probe` and the opt-out probe)
- `cuda_core/tests/graph/test_graph_definition_lifetime.py` (two sites)
- `cuda_bindings/tests/test_cuda.py`

This is a pure refactor. `unset_env` is opt-in per call site, so the two graph tests and the bindings test keep inheriting `PYTHONPATH` exactly as they do today, and only `test_rlcompleter_patch.py` drops it as it already did. `_run_probe` keeps `check=False` because both of its callers assert on the exit code themselves. `unset_env` is applied before `extra_env`, which preserves `_run_probe`'s clear-then-set handling of `CUDA_CORE_DONT_FIX_TAB_COMPLETION`; a test pins that ordering.

Net across the five sites is 21 added against 59 removed.

The issue counted four sites. The fifth, `test_graph_definition_lifetime.py:790`, arrived with #2406 on the same day the issue was filed.

The description sketched a `run_in_subprocess` fixture that re-runs the current test node with `pytest <nodeid>`. None of the five sites have that shape, and re-running the node would inherit pytest's rootdir, which is the exact cwd hazard two of them work around. Rationale is in #2412.

## Checklist


aryanputta:
Pushed `872ddc2` on top, a self-review pass against @rwgk's guidance on #2196 and #2239 rather than anything new in the behavior.

- Cut the module docstring down to one line. The version I opened with narrated the refactor and named the specific side effects in other test files. That is PR-description material, and it drifts.
- Removed the comments that restated the code below them or pointed at `_run_probe` by name from a test in a different file.
- Kept one general statement of the `cwd` contract, on the argument itself, since that is what makes the argument required.
- Added `assert cwd` instead of letting an empty string reach `subprocess.run`, in the shape suggested for `_resolve_in_trusted_dirs` on #2196, plus a test that the guard fires.

Diff is now 27 added against 50 removed on those files. No behavior change from the first commit.

cuda_core/tests/test_subprocess_runner.py
```diff
@@ -0,0 +1,107 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
mdboom: As you pointed out, this more correctly belongs in `cuda_python_test_helpers`, but we don't currently have tests there.

It feels like over-testing to me to have these tests -- we are already exercising the required behavior in the other tests, so I say just delete this file.

cuda_python_test_helpers/cuda_python_test_helpers/subprocess_runner.py
```diff
@@ -0,0 +1,67 @@
...
+import sys
+from collections.abc import Mapping
+
+__all__ = ["run_python_snippet"]
+
+
+def run_python_snippet(
+    code: str,
+    *,
+    cwd: str | os.PathLike[str],
```
mdboom: I think we could make this optional.  When not provided, use an auto-cleaned-up tempdir for `cwd`.  That should simplify most (but not all) of the call sites.

mdboom (changes requested):

aryanputta:
@mdboom The requested changes are addressed in `3286d4396b`: removed `cuda_core/tests/test_subprocess_runner.py`, made `cwd` optional with an auto-cleaned temporary directory, and simplified the call sites. The old review threads are now outdated. Could you please re-review this head when convenient?

mdboom (approved):
LGTM.  Just needs the merge conflicts resolved and another round of tests and I think this is good-to-go.

mdboom:
@aryanputta: Tests are failing now (maybe to a logical merge conflict?)  Can you address that?

Andy-Jost:
Hi @aryanputta any update?

aryanputta:
> Hi @aryanputta any update?

Will check this out tonight, wifi on campus is very weak.

PR #2658, opened 2026-08-18
fix(cuda.core): warn instead of silently dropping numba_debug on link paths

### LINKED ISSUE #2640: [BUG]: LinkerOptions(numba_debug=True) is silently dropped
### Type of Bug

Silent incorrect behavior (no error raised)

### Component

cuda.core

### Describe the bug

`LinkerOptions.numba_debug` (`cuda_core/cuda/core/_linker.pyx:309`) is declared as a public field but is never emitted by either backend. Neither `_prepare_nvjitlink_options` nor `_prepare_driver_options` references it, so setting it has no effect and no error.

This is not merely a dead field. `_translate_program_options` (`cuda_core/cuda/core/_program.pyx:745`) actively forwards `numba_debug` from `ProgramOptions` into `LinkerOptions`:

```python
return LinkerOptions(
    ...
    no_cache=options.no_cache,
    numba_debug = options.numba_debug
)
```

That translation runs on the `code_type == "ptx"` path of `Program`, so `Program(ptx, "ptx", ProgramOptions(numba_debug=True))` also drops the option. It is a silent drop on a live code path, which conflicts with the repository's "no silent failures" guidance: an option the user explicitly set should either take effect or raise, not vanish.

Introduced in #2158 (`792ab3c9fea`), the same change that added `numba_debug` to `ProgramOptions`. The NVVM half of that feature was separately broken and is fixed in #2639; this linker half is untouched by that PR.

### How to Reproduce

```python
from cuda.core import LinkerOptions

o = LinkerOptions(arch="sm_80", debug=True, numba_debug=True)
print(o.numba_debug)                              # True
print(o._prepare_nvjitlink_options(as_bytes=True))
```

```
True
[b'-arch=sm_80', b'-g']
```

The field is `True`, but no corresponding option is emitted. `grep -n numba cuda_core/cuda/core/_linker.pyx` returns only the field declaration.

### Expected behavior

One of:

1. Emit the option on the paths that support it, matching the `ProgramOptions` NVVM fix in #2639, or
2. Raise if `numba_debug` is set on a backend that cannot honor it, consistent with how `_prepare_nvvm_options_impl` raises `CUDAError` for options the NVVM backend does not support, or
3. Remove the field if it was never meant to be plumbed through the linker, and stop `_translate_program_options` from forwarding it.

Whichever is chosen, a user-set option should not disappear without a diagnostic.

### Operating System

Linux, CUDA 13.3 (reproduced against `main`)

rparolin:
## Description

closes #2640

`LinkerOptions` carries a public `numba_debug` field that no linking backend reads. nvJitLink rejects the option under every spelling and the driver's cuLink API has no corresponding `CUjit_option`, so setting it does nothing and reports nothing.

It is not merely dead. `_translate_program_options` forwarded `numba_debug` from `ProgramOptions` into `LinkerOptions` on the `code_type="ptx"` path, so `Program(ptx, "ptx", ProgramOptions(numba_debug=True))` silently discarded an option the user explicitly set.

This makes the drop audible without breaking any caller.

| Surface | Behavior |
|---|---|
| `LinkerOptions(numba_debug=...)` | constructs as before; `DeprecationWarning`; value ignored |
| `_prepare_nvjitlink_options` / `_prepare_driver_options` | untouched — the linker needs no knowledge of `numba_debug` to ignore it |
| `Program(ptx, "ptx", ProgramOptions(numba_debug=True))` | compiles as before; `UserWarning`; no longer forwarded |
| program-cache key | `numba_debug` excluded — it cannot change PTX-path output, so it must not perturb the key |
| NVVM / NVRTC paths | unchanged; option still emitted |

Nothing raises and nothing is removed. This follows existing precedent: `_prepare_driver_options` already handles six options the cuLink backend cannot honor (`ftz`, `prec_div`, `prec_sqrt`, `fma`, `kernels_used`, `variables_used`) by warning rather than raising.

### Two deliberate details

**The warning categories differ.** `LinkerOptions.numba_debug` gets a `DeprecationWarning` — the field is going away. The PTX path gets a `UserWarning`, because `ProgramOptions.numba_debug` is *not* deprecated: it is fully supported on NVVM/NVRTC and merely inapplicable to a linking backend. Marking it `DeprecationWarning` would tell users a supported option is being removed.

**The gate polarities differ.** The linker field gates on `is not None` — the field itself is doomed, so any explicit value earns the notice, including `False`. The PTX path gates on truthiness, matching `_prepare_nvvm_options_impl`, because `numba_debug=False` asks for nothing and there is no dropped intent to report.

### Removal

`LinkerOptions.numba_debug` is deprecated here and removed in 2.0.0, per the support policy: breaking API changes are confined to major-version boundaries, with a deprecation notice at least one minor release ahead.

Nothing in this repo tracked a scheduled removal — the only prior public-attribute deprecation, `Device.max_links`, says "a future release" with no version, issue, or test behind it. This adds the tracking to the release checklist (`.github/ISSUE_TEMPLATE/release_checklist.yml`, with matching guidance in `RELEASE-core.md`): before cutting a release, sweep for deprecations whose stated removal version has arrived and remove any that are due.

The sweep greps `deprecated::` rather than `\.\. deprecated::`, so it also catches the two `Device.max_links` sites, which spell the directive `.. version-deprecated::`. Whether that spelling is a real Sphinx directive is worth a separate look — if it is not, those two blocks are not rendering as deprecation notices at all.

### Verification

Full `cuda_core` suite on CUDA 13: 3844 passed, 238 skipped, 4 xfailed, 127 subtests passed, verified against a rebuilt extension. `pre-commit` clean, including `.pyi` stub regeneration and `mypy-cuda-core`.

`numba-cuda` is unaffected — it drives libNVVM directly via its own `cudadrv/nvvm.py` and never routes `numba_debug` through a cuda.core PTX path.

## Checklist

### DIFF
--- a/cuda_core/cuda/core/_linker.pyi
+++ b/cuda_core/cuda/core/_linker.pyi
@@ -186,8 +186,7 @@ class LinkerOptions:
         .. deprecated:: 1.2.0
             Setting this option emits a :class:`DeprecationWarning`. It has never
             had an effect on any linking backend and will be removed in
-            ``cuda.core`` 2.0.0, the next major-version boundary permitted by the
-            :doc:`support policy <support>`. Use
+            ``cuda.core`` 2.0.0. Use
             :attr:`ProgramOptions.numba_debug` on an NVVM or NVRTC compilation
             path instead.
     """
--- a/cuda_core/cuda/core/_linker.pyi
+++ b/cuda_core/cuda/core/_linker.pyi
@@ -177,6 +177,19 @@ class LinkerOptions:
     no_cache : bool, optional
         Do not cache the intermediate steps of nvJitLink.
         Default: False.
+    numba_debug : bool, optional
+        Non-functional. ``numba_debug`` is an NVVM/NVRTC *compiler* option;
+        neither nvJitLink nor the driver's cuLink API recognizes it, so no
+        linking backend can honor it and the value is ignored.
+        Default: None.
+
+        .. deprecated:: 1.2.0
+            Setting this option emits a :class:`DeprecationWarning`. It has never
+            had an effect on any linking backend and will be removed in
+            ``cuda.core`` 2.0.0, the next major-version boundary permitted by the
+            :doc:`support policy <support>`. Use
+            :attr:`ProgramOptions.numba_debug` on an NVVM or NVRTC compilation
+            path instead.
     """
     name: str | None = '<default linker>'
     arch: str | None = None
--- a/cuda_core/cuda/core/_linker.pyx
+++ b/cuda_core/cuda/core/_linker.pyx
@@ -283,6 +283,19 @@ class LinkerOptions:
     no_cache : bool, optional
         Do not cache the intermediate steps of nvJitLink.
         Default: False.
+    numba_debug : bool, optional
+        Non-functional. ``numba_debug`` is an NVVM/NVRTC *compiler* option;
+        neither nvJitLink nor the driver's cuLink API recognizes it, so no
+        linking backend can honor it and the value is ignored.
+        Default: None.
+
+        .. deprecated:: 1.2.0
+            Setting this option emits a :class:`DeprecationWarning`. It has never
+            had an effect on any linking backend and will be removed in
+            ``cuda.core`` 2.0.0, the next major-version boundary permitted by the
+            :doc:`support policy <support>`. Use
+            :attr:`ProgramOptions.numba_debug` on an NVVM or NVRTC compilation
+            path instead.
     """
 
     name: str | None = "<default linker>"
@@ -311,6 +324,21 @@ class LinkerOptions:
     def __post_init__(self) -> None:
         _lazy_init()
         self._name = self.name.encode()
+        # No linking backend reads ``numba_debug``, so warn where the value is
+        # supplied rather than in the option builders -- the user learns once,
+        # at the call site that set it, instead of once per link. The gate is
+        # ``is not None`` (unlike the ignore-warning on the PTX compile path):
+        # it is the *field* that is going away, so any explicit value earns the
+        # notice, including ``False``.
+        if self.numba_debug is not None:
+            warn(
+                "numba_debug is not supported by any linking backend and is ignored. "
+                "LinkerOptions.numba_debug is deprecated and will be removed in "
+                "cuda.core 2.0.0; use ProgramOptions.numba_debug on an NVVM or NVRTC "
+                "compilation path instead.",
+                DeprecationWarning,
+                stacklevel=3,
+            )
 
     def _prepare_nvjitlink_options(self, as_bytes: bool = False) -> list[bytes] | list[str]:
         options = []
--- a/cuda_core/cuda/core/_linker.pyx
+++ b/cuda_core/cuda/core/_linker.pyx
@@ -292,8 +292,7 @@ class LinkerOptions:
         .. deprecated:: 1.2.0
             Setting this option emits a :class:`DeprecationWarning`. It has never
             had an effect on any linking backend and will be removed in
-            ``cuda.core`` 2.0.0, the next major-version boundary permitted by the
-            :doc:`support policy <support>`. Use
+            ``cuda.core`` 2.0.0. Use
             :attr:`ProgramOptions.numba_debug` on an NVVM or NVRTC compilation
             path instead.
     """
--- a/cuda_core/tests/test_linker.py
+++ b/cuda_core/tests/test_linker.py
@@ -7,7 +7,6 @@ import warnings
 
 import pytest
 
-import cuda.core
 from cuda.core import Device, Linker, LinkerOptions, Program, ProgramOptions, _linker
 from cuda.core._module import ObjectCode
 from cuda.core._program import _can_load_generated_ptx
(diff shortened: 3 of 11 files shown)


rparolin:
**Resolved — the PR has been reshaped; the diff no longer contains a breaking change.**

(I'd posted two comments earlier working through this; they seem to have been removed, so this consolidates and supersedes them.)

The open question was whether to remove `LinkerOptions.numba_debug` now, given that the option does nothing and never has, or hold it for a major release. Going by `cuda_core/docs/source/support.rst` strictly, two rules apply:

> **Major**: [...] Breaking API changes only happen at major-version boundaries.

> Deprecation notices will be issued at least for one (1) minor release, before the actual removal happens.

Note this rules out **1.3.0** as well — it's a minor release, so the only legal removal target is 2.0.0. We went with strict compliance:

- **1.2.0** — `LinkerOptions.numba_debug` is deprecated (`DeprecationWarning`, value still ignored, field retained, no signature change). The PTX path emits a `UserWarning` saying the option is ignored instead of dropping it silently. Nothing raises, nothing is removed.
- **2.0.0** — remove the field.

That still fixes the actual bug in #2640 — the *silent* part of the silent drop — without breaking a single caller.

Two things worth your eye:

1. **Removal tracking.** Nothing in this repo tracks a scheduled removal. The only prior public-attribute deprecation, `Device.max_links`, says "a future release" with no version, issue, or test behind it. So this PR adds a version-gated test that fails the build once `cuda.core.__version__` crosses 2.0 and names exactly what to delete. That's a new pattern here — say the word and I'll swap it for a 2.0.0 milestone issue instead.

2. **`.. version-deprecated::` is not a real directive.** It's used at `cuda_core/cuda/core/system/_nvlink.pxi:32` and in the generated stub, but `docs/source/conf.py` registers no extension providing it; the Sphinx built-in is `.. deprecated::`. Those two occurrences are probably rendering as "Unknown directive type". Not touched in this PR — flagging for a follow-up.

cuda_core/cuda/core/_linker.pyi
```diff
@@ -177,6 +177,19 @@ class LinkerOptions:
...
+        linking backend can honor it and the value is ignored.
+        Default: None.
+
+        .. deprecated:: 1.2.0
+            Setting this option emits a :class:`DeprecationWarning`. It has never
+            had an effect on any linking backend and will be removed in
+            ``cuda.core`` 2.0.0, the next major-version boundary permitted by the
+            :doc:`support policy <support>`. Use
+            :attr:`ProgramOptions.numba_debug` on an NVVM or NVRTC compilation
+            path instead.
```
mdboom, line 191:
```suggestion
            Setting this option emits a :class:`DeprecationWarning`. It has never
            had an effect on any linking backend and will be removed in
            ``cuda.core`` 2.0.0. Use
            :attr:`ProgramOptions.numba_debug` on an NVVM or NVRTC compilation
            path instead.
```

cuda_core/cuda/core/_linker.pyx
```diff
@@ -283,6 +283,19 @@ class LinkerOptions:
...
+        linking backend can honor it and the value is ignored.
+        Default: None.
+
+        .. deprecated:: 1.2.0
+            Setting this option emits a :class:`DeprecationWarning`. It has never
+            had an effect on any linking backend and will be removed in
+            ``cuda.core`` 2.0.0, the next major-version boundary permitted by the
+            :doc:`support policy <support>`. Use
+            :attr:`ProgramOptions.numba_debug` on an NVVM or NVRTC compilation
+            path instead.
```
mdboom, line 297:
```suggestion
            Setting this option emits a :class:`DeprecationWarning`. It has never
            had an effect on any linking backend and will be removed in
            ``cuda.core`` 2.0.0. Use
            :attr:`ProgramOptions.numba_debug` on an NVVM or NVRTC compilation
            path instead.
```

cuda_core/tests/test_linker.py
```diff
@@ -436,6 +438,60 @@ def test_prepare_driver_options_unsupported_raises(driver_binding, kwargs, match
...
+    release" -- so this test fails the build the moment the version crosses 2.0,
+    rather than letting a dead field survive into the 2.x line unnoticed.
+    """
+    major = int(cuda.core.__version__.split(".")[0])
+    if major >= 2:
+        pytest.fail(
+            "cuda.core is now 2.x: remove the deprecated LinkerOptions.numba_debug field, "
+            "its __post_init__ DeprecationWarning, and this test."
+        )
+    assert hasattr(LinkerOptions(arch="sm_80"), "numba_debug")
```
mdboom: This is a bomb that will go off when we tag `cuda-core-v2.0.0`, and because we can't pre-test tags it will be a major pain.

I think instead we just rely on the `.. deprecated::` directives in the code base and we remind ourselves to do a sweep of deprecated tags /before/ release.  /During/ release isn't really possible given our workflows.

mdboom (changes requested):

cuda_core/tests/test_linker.py
rparolin: Yes, this was intention b/c we currently don't have a forcing function to ensure that everything we've deprecated gets culled at the new major revision.   I'm open to other alternatives.

rparolin: Agreed, and it's worse than the tag timing — the version bump lands on an ordinary release-prep PR during code freeze, so the test would go red at exactly the wrong moment and look like a real regression to anyone who didn't write it. Dropped it in 344d7bd, along with the now-unused `import cuda.core`.

For the "remind ourselves" half, I put it where release managers already look rather than leaving it to memory: a checklist item in `.github/ISSUE_TEMPLATE/release_checklist.yml` with matching guidance in `RELEASE-core.md`.

```
- label: Sweep deprecations whose stated removal version has arrived
         (`grep -rn 'deprecated::' cuda_core/cuda`) and remove any that are due
```

One detail worth flagging: the pattern is `deprecated::`, not `\.\. deprecated::`. That's deliberate — the two `Device.max_links` sites (`system/_nvlink.pxi:31`, `system/_device.pyi:801`) spell it `.. version-deprecated:: 1.1.0`, which I can't find in Sphinx's directive set. If it isn't one, those blocks aren't rendering as deprecation notices at all. I left them alone as out of scope here, but happy to check it in a docs build and fix separately if you want.

mdboom (approved):

PR #2670, opened 2026-08-18
cuda.core: fix some type signatures that are too generic

juenglin:
Fix `options` parameter type annotations that were typed as `object` instead of their proper options types (`StreamOptions`, `TensorMapDescriptorOptions`).


cuda_core/tests/graph/test_graph_builder.py
juenglin, line 1: Using `StreamOptions` has become necessary with this change because `StreamOptions` is a "cdef dataclass".

See related discussion in PR #2634.

mdboom (approved):
LGTM.

PR #2679, opened 2026-08-20
fix(cuda.core): write NVRTC source to a temp file when debug or lineinfo is enabled

### LINKED ISSUE #2422: cuda.core: materialize NVRTC source to a real path when debug info is enabled
## Summary

When a user compiles a kernel via `cuda.core.Program` with `debug=True` or `lineinfo=True`, `cuda-gdb` cannot show source lines because the program `name=` defaults to the literal string `"default_program"` and no file with that name exists on disk. The debugger reports `default_program:10 — No such file or directory` and users lose source-level debugging.

This is tracked in NVBugs [6420444](https://nvbugspro.nvidia.com/bug/6420444) (originally reported by Linaro). The customer was unblocked by a workaround: pass a real file path as `name=` and write the source there manually. We want this to work by default.

## Proposed fix

When `debug=True` or `lineinfo=True` is set on `Program(...)`:

1. Derive a meaningful filename from the caller's Python frame + a sanitized kernel identifier (e.g. `train_py__matmul.cu`).
2. Materialize the NVRTC source to a temp file at that path.
3. Pass that real path as NVRTC's `name=`.
4. Clean up the temp file on program destruction.

When neither debug flag is set, behavior is unchanged (no disk I/O, same label as today).

## Impact on existing users

- No API or signature changes.
- Users who already pass `name=` explicitly: zero change (their value wins).
- Users who don't request debug info: zero change (no temp file written).
- Users who *do* request debug info: the label in `cuda-gdb` / traces / logs changes from `default_program` to a real path. Cosmetic; easily searchable.

## Edge cases to handle

- Read-only `/tmp` or sandboxed environments: fall back to today's label-only behavior on write failure, do not crash.
- Concurrent compiles of the same-named kernel: use unique suffix (`tempfile.mkstemp` or PID+hash).
- Respect `TMPDIR`; user-only permissions on the temp file.
- Cleanup on both normal destruction and process exit.

## Acceptance criteria

- `cuda-gdb` shows source lines when stepping into a kernel compiled with `debug=True`, without any user-side workaround.
- Existing behavior preserved when debug flags are off.
- Graceful fallback when the filesystem isn't writable.
- Tests covering: debug-on path, debug-off path, read-only-tmp fallback, concurrent compile.

---
Related: NVBugs 6420444

lijinf2:
## Description

closes https://github.com/NVIDIA/cuda-python/issues/2422

When `Program` compiles C++ with `debug=True` or `lineinfo=True` and the caller
does not pass `name=`, NVRTC recorded the label `default_program`. That path
does not exist on disk, so cuda-gdb reported `No such file or directory`.

This change writes the in-memory source to a unique temp file
(`caller_py__kernel_XXXXXXXX.cu`, `mkstemp`, mode 0600) and passes that real
path to `nvrtcCreateProgram`. `ProgramOptions.name` is left unchanged so the
program cache key is unaffected. An explicit `name=` still wins. Write
failures fall back to the old label-only behavior. The temp file is removed
on `close()` / deallocation.

No public API or signature changes.

## Checklist


rparolin (approved):

lijinf2:
Thank you @rparolin!

PR #921, opened 2025-08-28
Reverse tabulated names to achieve new → old search order.

rwgk:
## Description

This addresses issue #839. This simple reversal of the search order is not a completely general solution, but expected to have the desired behavior more often than the current old → new search order. Please see the extensive discussions under #839 for background.

Bump cuda-pathfinder version to `1.2.0a0`

PR #952, opened 2025-09-08
Add newly available `nvidia-nccl-cu13` to `nvidia_wheels_cu13`

rwgk:
I noticed this by chance (internal email announcement).

https://pypi.org/project/nvidia-nccl-cu13/#files

PR #945, opened 2025-09-04
Add `nccl` in supported_nvidia_libs.py

rwgk:
## Description

Closes #944

Bump cuda-pathfinder version to `1.2.2a0`

PR #2680, opened 2026-08-21
Add cuDNN discovery and NCCL header support

DEKHTIARJonathan:
## Summary

This PR adds `cuda-pathfinder` coverage for cuDNN 9 and NCCL. It adds cuDNN dynamic-library and header discovery, and adds the missing NCCL header discovery alongside NCCL's existing dynamic-library and device-bitcode support. The explicitly declared layouts cover released wheels, Conda test environments, selected conventional standalone and system locations, and the verified Windows ARM64 cuDNN archive layout.

Completing that goal required more than adding filenames to the catalogs. cuDNN has required and conditional runtime dependencies, Windows standalone installations need product-specific roots and dependent-DLL directory registration, and cuDNN's component DLL names make unconstrained wildcard matching unsafe. The implementation therefore adds reusable internal metadata for optional dependencies, product installation roots, architecture-specific Windows layouts, and descriptor-authorized DLL fallback globs.

Reviewing those paths also exposed several pre-existing inconsistencies: CUPTI's Windows names were stored in the opposite order from the catalog convention, Windows already-loaded detection did not use the same newest-first order as native loading, and the directory of an already-loaded DLL whose descriptor requires dependent-directory registration was registered only when a potentially unrelated early wheel/Conda search had succeeded. Those issues are fixed here. Finally, the PR contains a hook-local Cython 3.2.9 pin for an unrelated `cython-lint` incompatibility that appeared when Cython 3.3.0 was released while this work was in CI.

No new public Python API is introduced. Existing `cuda.pathfinder` APIs recognize additional products and installation layouts; the new fields and helpers are internal discovery metadata.

## Additional changes at a glance

- Models cuDNN's required cuBLASLt dependency and optional NVRTC dependency without hiding real load or contract errors.
- Adds descriptor-driven `CUDNN_PATH` and `NCCL_HOME` discovery for dynamic libraries and headers, plus standard Linux include layouts and versioned Windows Program Files cuDNN installs.
- Adds narrowly scoped Windows ARM64 discovery for NVIDIA's verified standalone cuDNN archive layout.
- Replaces unsafe generic Windows DLL prefix matching with exact known names followed by descriptor-authorized fallback globs; currently only CUPTI opts into `cupti64_*.dll`.
- Makes Windows DLL selection consistently newest-first for both native loading and already-loaded detection, and repairs the CUPTI catalog order.
- Tightens cuDNN test dependency and distribution matching, expands real test environments, and adds focused coverage for the new layouts and loader semantics.
- Pins only the isolated `cython-lint` hook to Cython 3.2.9 until `cython-lint` supports Cython 3.3.
- Adds release notes in preparation for the `cuda-pathfinder` v1.7.0 release and restores the missing 1.6.1 documentation-version switcher entry.

## Primary goal: cuDNN 9 discovery and NCCL header coverage

The original request was to add cuDNN discovery and complete the NCCL header surface in `cuda-pathfinder`.

| Product surface | Linux | Windows x64 | Windows ARM64 |
| --- | --- | --- | --- |
| cuDNN 9 dynamic library | Added: `libcudnn.so.9` | Added: `cudnn64_9.dll` | Added for the verified standalone archive under `CUDNN_PATH` |
| cuDNN headers | Added: `cudnn.h` | Added: `cudnn.h` | Added through the archive's shared `include` layout |
| NCCL dynamic library | Existing `libnccl.so.2` support; this PR adds `NCCL_HOME` layouts | Out of scope on Windows | Out of scope on Windows |
| NCCL headers | Added: `nccl.h` | Out of scope on Windows | Out of scope on Windows |

The manually created Pathfinder test environments are kept aligned with that scope: the Conda helpers provision cuDNN on Linux and Windows and NCCL on Linux.

### cuDNN dynamic-library discovery

The cuDNN descriptor recognizes the cuDNN 9 ABI names and the released wheel layouts:

- Linux: `libcudnn.so.9` under `nvidia/cudnn/lib`.
- Windows x64: `cudnn64_9.dll` under `nvidia/cudnn/bin`.
- Standalone and system layouts described in the product-root and Windows sections below.

The descriptor is deliberately major-version-specific. The test dependency groups therefore use `nvidia-cudnn-cu12>=9,<10` and `nvidia-cudnn-cu13>=9,<10`; an installed cuDNN 8 or a future cuDNN 10 package must not be treated as evidence that this cuDNN 9 descriptor is valid.

### cuDNN and NCCL header discovery

The existing header APIs now locate:

- cuDNN headers in wheels, Conda environments, `CUDNN_PATH/include`, standard Linux include directories, Linux's running multiarch include directory, and versioned Windows Program Files installs.
- NCCL headers in wheels, Conda environments, `NCCL_HOME/include`, `NCCL_HOME/build/include`, `/usr/include`, and `/usr/local/include`.

`NCCL_HOME/build/include` covers a source or build tree, while `NCCL_HOME/include` covers an installed or extracted prefix. The Windows cuDNN system search uses versioned `v9.*` roots and numeric-aware ordering, so `v9.10` correctly sorts newer than `v9.9`.

## cuDNN runtime dependencies

cuDNN does not fit the existing all-or-nothing dependency model:

- cuDNN requires cuBLASLt in the supported configurations, so it declares `dependencies=("cublasLt",)` and fails if cuBLASLt cannot be loaded.
- cuDNN conditionally uses NVRTC, so it declares `optional_dependencies=("nvrtc",)`.

`optional_dependencies` has intentionally narrow semantics. A plain `DynamicLibNotFoundError` means that an optional runtime component is absent and may be ignored. An unknown descriptor, a platform-availability error, or a dependency that was found but failed to load remains an error. In other words, optional means that absence is acceptable; it does not mean that a broken installation is acceptable.

Required dependencies are loaded first, followed by optional dependencies. This ordering also ensures that cuBLASLt is available before Pathfinder attaches cuDNN.

## Product installation roots and native layouts

cuDNN and NCCL are separately packaged products, so treating every layout as if it were rooted at `CUDA_PATH` is insufficient. This PR adds descriptor-driven installation-root metadata and a shared search step for conventional product variables.

| Product root | Dynamic-library layouts | Header layouts |
| --- | --- | --- |
| `CUDNN_PATH` on Linux | `lib`, `lib64` | `include` |
| `CUDNN_PATH` on Windows x64 | `bin/x64`, `bin` | `include` |
| `CUDNN_PATH` on Windows ARM64 | `bin/arm64` | `include` |
| `NCCL_HOME` on Linux | `lib`, `lib64`, `build/lib` | `include`, `build/include` |
| Windows Program Files | `NVIDIA/CUDNN/v9.*` roots with the x64 cuDNN bin layouts | `NVIDIA/CUDNN/v9.*/include` |

The metadata separates installation roots from the relative layouts beneath those roots. It is also platform- and Windows-architecture-specific. Product variables and Program Files reuse the same root-traversal machinery while retaining source-specific layout metadata. This prevents a verified ARM64 archive layout from leaking into unverified Conda, `CUDA_PATH`, or Program Files dynamic-library searches.

Dynamic descriptors express those responsibilities separately through `install_root_env_vars_linux/windows`, `install_root_env_rel_dirs_linux/windows`, and `program_files_root_globs_windows`. Header descriptors similarly declare `product_root_env_vars`, platform-specific system directories, and the Linux multiarch include policy. Search results retain the winning environment variable in `found_via`, so callers can distinguish `CUDNN_PATH` and `NCCL_HOME` results from wheel, Conda, CUDA, or system results.

Root traversal tries every configured relative directory until it finds the requested primary library. Merely finding an earlier directory is not success; for example, an existing `bin/x64` without `cudnn64_9.dll` does not prevent checking the next supported layout. Version-bearing roots and fallback filenames use shared numeric-aware ordering rather than lexical ordering.

The NCCL ordering is intentional:

- `lib` is NCCL's canonical installed, archive, and build-directory layout.
- `lib64` supports legitimate GNUInstallDirs/CMake and RPM-style installed prefixes.
- `build/lib` supports an NCCL source checkout.

There is no speculative `build/lib64`: NCCL's in-tree build emits `build/lib`. Ordinary distro multiarch installs remain the native loader's responsibility rather than introducing a broad architecture glob here.

### Search precedence and relationship to #1054

These product variables are conventional-layout fallbacks, not developer overrides.

- Dynamic-library loading retains existing load precedence: already loaded, wheel/Conda absolute candidate, native system loading, product environment roots, `CUDA_PATH`/`CUDA_HOME`, Program Files, then the CTK-root canary.
- Header discovery searches wheels and Conda before product roots; product roots then precede generic CUDA and system fallbacks.
- A product variable names an installation root searched through declared layouts. It is not an arbitrary direct-file override and an invalid or non-matching root falls through normally.

This is useful groundwork adjacent to #1054, but it does **not** implement that issue's highest-priority, fail-loud developer override. A custom NVSHMEM build still cannot override an installed NVIDIA wheel, so #1054 remains open for that separately designed capability.

## Windows ARM64 cuDNN support

NVIDIA now publishes a CUDA 13 cuDNN 9.25 standalone Windows ARM64 archive with `bin/arm64/cudnn64_9.dll` and architecture-independent headers under `include`. This PR supports that verified topology through `CUDNN_PATH`.

The scope is intentionally narrow:

- Windows ARM64 site-packages discovery remains disabled because there is no released stable `win_arm64` cuDNN wheel.
- Generic Conda, `CUDA_PATH`, and Program Files dynamic-library layouts remain x64-only because equivalent ARM64 layouts were not verified.
- Only the product-root metadata exposes `CUDNN_PATH/bin/arm64`, so adding ARM64 support does not silently claim unsupported installation forms.

This should be described as end-to-end Pathfinder discovery and loading support for the official standalone archive layout, not as exhaustive cuDNN product support on every Windows ARM64 packaging channel.

## Already-loaded Windows DLL directory registration

Windows descriptors such as NVRTC, cuFFT, and cuDNN use `requires_add_dll_directory` because the primary DLL may later resolve sibling or component DLLs from its own directory.

The old already-loaded path gated this side effect on `have_abs_path`. That name was misleading: the boolean did not say whether Pathfinder knew the resident module's absolute path. Windows already resolved that authoritative path with `GetModuleFileNameW`; the boolean only said whether the earlier wheel/Conda filesystem search had found some candidate.

As a result, a recognized cuDNN DLL that was already resident from a late or external location such as `CUDNN_PATH` or Program Files could be returned before those roots were searched and without registering the directory containing its lazy component DLLs.

This PR removes the phase-derived boolean. If a recognized resident DLL's descriptor sets `requires_add_dll_directory`, Pathfinder now registers the directory of the module that is actually loaded before loading dependencies, regardless of how that module entered the process. This changes runtime behavior only for the affected Windows already-loaded branch; newly loaded absolute paths already registered their directory.

## Safe Windows DLL matching and forward compatibility

Directory-based Windows discovery previously relied on a generic `<descriptor-name>*.dll` pattern. That is unsafe for cuDNN because `cudnn*.dll` also matches component DLLs; loading a sidecar instead of `cudnn64_9.dll` is incorrect. The same generic policy has sibling-name collisions such as `cublas`/`cublasLt`, `cufft`/`cufftw`, `cusolver`/`cusolverMg`, `cusparse`/`cusparseLt`, and `cutensor`/`cutensorMg`.

Removing wildcard matching globally would create a different regression: CUPTI embeds a full release version in its Windows basename, so a future `cupti64_*.dll` would become undiscoverable until the catalog was updated.

The final policy makes that tradeoff explicit:

1. Try the descriptor's known exact DLL basenames first, newest known to oldest known.
2. Only if no exact candidate succeeds, try that descriptor's explicit `windows_dll_fallback_globs`.
3. Sort fallback matches in reverse numeric-aware order.

Only CUPTI currently opts in, with `windows_dll_fallback_globs=("cupti64_*.dll",)`. cuDNN remains exact-only. Known CUPTI names deliberately win over an unlisted fallback match.

The descriptor-aware filesystem helper also applies PE-architecture validation for descriptors that request it, including site-packages candidates.

The wildcard is limited to filesystem-directory discovery. Native Windows name search and already-loaded detection require concrete catalogued basenames, so the exact table remains necessary and an unknown wildcard-only DLL that is already resident is outside this fallback's guarantees.

## Consistent newest-first Windows DLL selection

The established Windows catalog convention stores versioned DLL names from oldest to newest. Runtime lookup reverses that order so the newest supported name wins.

Two pre-existing inconsistencies became visible during this work:

- CUPTI DLL names had been added incrementally in newest-to-oldest order, opposite the catalog convention. Reversing that tuple during lookup therefore preferred older CUPTI names when several were present.
- PR #921 changed native Windows loading to search newest-first but left `check_if_already_loaded_from_elsewhere()` scanning the catalog forward. If multiple known generations were already resident, that path could return the oldest one even though normal loading preferred the newest.

This PR stores CUPTI names in canonical oldest-to-newest order and centralizes Windows runtime traversal in `_candidate_dll_names()`. Both native system loading and already-loaded detection now consume the same newest-to-oldest sequence. Normal single-version installations are unchanged; the fix makes multi-version behavior deterministic and consistent.

## Catalog, packaging, and test-environment correctness

The expanded product coverage exposed several smaller assumptions that needed to be made explicit:

- Test distribution detection now matches only the backend packages `nvidia-cudnn-cu12` and `nvidia-cudnn-cu13`. Broad `nvidia-cudnn-*` matching incorrectly treated packages such as `nvidia-cudnn-frontend` and `nvidia-cudnn-jit-*` as proof that `nvidia/cudnn/include/cudnn.h` exists.
- The Windows add-DLL-directory consistency test now uses the complete all-descriptor Windows DLL projection. Its historical CTK-only projection could not represent separately packaged cuDNN.
- Cross-platform header tests use host-appropriate path construction and real glob expansion. This fixes test-only separator assumptions exposed by the initial Windows matrix without changing runtime search behavior.

## Coincidental `cython-lint` infrastructure fix

Cython 3.3.0 was released while a clean Windows pre-commit environment for this PR was being created. The repository freezes `cython-lint` at v0.19.0, but that package declares an unbounded `Cython>=0.29.32` dependency. The fresh hook therefore resolved Cython 3.3.0.

Cython 3.3 changed its dictionary-comprehension AST, while released `cython-lint` versions still accessed the removed `DictComprehensionAppendNode.value_expr` attribute. The hook crashed on an unchanged, valid dictionary comprehension in `cuda_core`; it did not report a lint violation and the crash was unrelated to this PR's Pathfinder changes.

The fix adds `Cython==3.2.9` only to the isolated `cython-lint` hook's `additional_dependencies`, matching the repository's existing stub-generation hook pin. It does not constrain CUDA Python builds, runtime dependencies, wheels, or users. The pin can be removed after a released `cython-lint` supports Cython 3.3 and the hook passes from a clean cache.

## Validation

Validation combines focused structural tests with real, non-mocked artifact loading:

- Focused tests cover descriptor completeness, required and optional dependency semantics, every declared product-root layout, numeric-aware ordering, Program Files discovery, architecture isolation, exact-versus-fallback DLL matching, sibling-name collision prevention, already-loaded directory registration, and newest-first known-name selection.
- Public PR CI provides strict real wheel discovery and native loading for cuDNN on Windows x64, Linux x86_64, and Linux AArch64 across CUDA 12 and CUDA 13. It also exercises real NCCL wheels, libraries, and headers on both Linux architectures. See the [public CI log analysis](https://github.com/NVIDIA/cuda-python/pull/2680#issuecomment-5375625698).
- ctk-next PR 537 adds CUDA 13.4 coverage across Linux x86_64, Linux AArch64, Windows x64, and native Windows ARM64, from both source and isolated installed wheels.
- The native Windows ARM64 job provisions the official cuDNN 9.25 archive, real cuBLASLt, and real NVRTC. It verifies exact `CUDNN_PATH`/`CUDA_PATH` resolution, ARM64 PE identity, the complete expected cuDNN component set, `cudnnGetVersion`, Graph/Ops/CNN/Adv compatibility checks, and successful `cudnnCreate`/`cudnnDestroy` on a real GB10 GPU.
- Alternate Program Files, system-header, and product-root layouts have focused real-filesystem unit coverage even where CI does not provision a complete external product installation.

This is comprehensive coverage of Pathfinder's discovery and loading contract for the verified layouts. It is not an exhaustive cuDNN or cuBLASLt numerical workload suite, nor does it claim every packaging channel or version/architecture combination.

The real-artifact runs establish the platform and artifact coverage described above. The later review-follow-up loader policies are additionally guarded by focused regression tests.


rwgk:
@DEKHTIARJonathan I pushed four commits (one of them adding WoA support) and triggered the CI here. I'll do the WoA testing on the internal repo.

rwgk:
I forgot to add: I didn't fully review myself yet.

cuda_pathfinder/pyproject.toml
```diff
@@ -24,6 +24,7 @@ cu12 = [
     "cuquantum-cu12; sys_platform != 'win32'",
     "cutensor-cu12",
     "nvidia-cublasmp-cu12; sys_platform != 'win32'",
+    "nvidia-cudnn-cu12>=9,<10",
```
kkraus14, line 27: Should we add nccl as well?

cuda_pathfinder/pyproject.toml
```diff
@@ -39,6 +40,7 @@ cu13 = [
     "cutensor-cu13",
     "nvidia-cublasmp-cu13; sys_platform != 'win32'",
     "nvidia-cudla; platform_system == 'Linux' and platform_machine == 'aarch64'",
+    "nvidia-cudnn-cu13>=9,<10",
```
kkraus14, line 43: +1 add nccl?

cuda_pathfinder/pyproject.toml
rwgk: We have it already (currently line 33 below):

https://github.com/NVIDIA/cuda-python/pull/945/changes

cuda_pathfinder/pyproject.toml
rwgk: This one came a little later (currently line 49 below):

https://github.com/NVIDIA/cuda-python/pull/952/changes

rwgk:
codex gpt-5.6-sol ultra
___

## PR 2680: first successful public CI log analysis

Analyzed against public CI run [32522149396](https://github.com/NVIDIA/cuda-python/actions/runs/32522149396), commit 39f25a00fd0e76e2810e2a3dea47c8b6d1533f41 (`Fix cross-platform header discovery tests`).

## Executive conclusion

The public run is fully green: 100 jobs succeeded, two intentionally gated CUDA Core API checks were skipped, no job failed or was cancelled, and the final aggregate status job succeeded.

The run provides strong, genuinely unmocked integration evidence for the main Pathfinder contract introduced or affected by PR 2680:

- Real NVIDIA cuDNN wheels were installed, their real headers were discovered, and their real shared libraries were loaded through the operating-system loader on Windows x64, Linux x86_64, and Linux AArch64.
- Real NCCL wheel headers were discovered on Linux x86_64 and Linux AArch64. The pre-existing NCCL dynamic-library support also loaded the real NCCL shared library, protecting it from regression.
- Both CUDA 12 and CUDA 13 dependency families were exercised. The resolved cuDNN build was `9.24.0.43`; the resolved NCCL build was `2.31.2`.
- cuDNN's required `cublasLt` dependency and present optional `nvrtc` dependency were both real, loadable artifacts in every strict job.
- All 65 GPU test jobs ran Pathfinder once before installing the non-CTK test wheels and again after installation with `all_must_work`. cuDNN on all three platforms, and NCCL on the 44 Linux jobs where it is supported, are not exempt in strict mode; those successes cannot be explained by permissive skips.

This is sufficient real integration coverage for the pip-wheel discovery and native-load claims on the three public platforms above. It is not sufficient real evidence for Windows on ARM, standalone `CUDNN_PATH` archives, Program Files installations, Linux system packages, conda layouts, or actual cuDNN/NCCL API execution. The alternative installation paths have synthetic-filesystem, mocked, or static coverage; product API execution has no coverage in this public run.

My merge-readiness judgment is therefore:

- **Yes** for the core Pathfinder find/load contract and current CUDA 12/13 wheel layouts on Windows x64 and both Linux architectures.
- **No, not yet as a production-validation claim** for Windows ARM64 or non-wheel installation layouts. The code can still merge with those paths described accurately as structurally tested, but real WoA/archive validation remains the highest-value follow-up.

## Method

I treated a test as "real" only when it used an installed NVIDIA artifact and the production discovery or native-loading code without replacing that operation with a mock. I used three evidence levels:

| Level | Meaning | Examples in this run |
|---|---|---|
| Real integration | Installed NVIDIA artifact plus production discovery/native loader | cuDNN/NCCL wheel headers; `libcudnn.so.9`, `cudnn64_9.dll`, and `libnccl.so.2` loads |
| Real filesystem, synthetic artifact | Production path/glob logic over real temporary directories, but a touched placeholder file | `CUDNN_PATH`, Program Files, version ordering, product-root header searches |
| Mocked/static | Loader, glob, platform, or dependency callback replaced; or metadata checked without loading | Windows ARM64 orchestration, Linux multiarch/system layouts, optional-dependency error semantics, descriptor invariants |

The audit used the completed GitHub Actions metadata and all 65 Linux/Windows GPU test logs, then cross-checked the relevant workflow and test implementation at the run SHA. Representative job links are included below.

## Run overview

- Run: [CI 32522149396](https://github.com/NVIDIA/cuda-python/actions/runs/32522149396)
- Event/ref: `push` to `pull-request/2680`
- Head SHA: 39f25a00fd0e76e2810e2a3dea47c8b6d1533f41
- Duration: approximately 49 minutes 49 seconds (`20:09:44Z` to `20:59:33Z`)
- Jobs: 102 total; 100 success; two skipped; zero failures
- Successful supporting gates: Windows pre-commit, Linux and Windows sdists, documentation, 24 wheel-build jobs, 65 GPU test jobs, and final aggregate status
- Skipped gates: `API check (cuda_core vs. merge base)` and `API check (cuda_core vs. latest release)`, both change-gated and unrelated to the Pathfinder scope

The 24 wheel-build jobs comprise eight each for Linux x86_64, Linux AArch64, and Windows x64. They cover Python 3.10 through 3.15, including selected 3.14t and 3.15t free-threaded builds.

### GPU test matrix

Every one of these 65 jobs successfully completed all three relevant steps:

1. `Run cuda.pathfinder tests with see_what_works`
2. `Install cuda.pathfinder extra wheels for testing`
3. `Run cuda.pathfinder tests with all_must_work`

| Platform | CUDA 12.9.1 local/wheels | CUDA 13.0.2 local/wheels | CUDA 13.3.0 local/wheels | Total |
|---|---:|---:|---:|---:|
| Linux x86_64 | 3 / 4 | 5 / 2 | 7 / 5 | 26 |
| Linux AArch64 | 3 / 3 | 3 / 3 | 4 / 2 | 18 |
| Windows x64 | 3 / 3 | 3 / 3 | 4 / 5 | 21 |
| **Total** | **9 / 10** | **11 / 8** | **15 / 12** | **65** |

The test jobs span Python 3.10, 3.11, 3.12, 3.13, 3.14, 3.14t, 3.15, and 3.15t where selected by the platform matrix. Windows additionally spans MCDM, TCC, and WDDM driver modes.

An important qualification: `local` versus `wheels` describes the CTK/core-bindings setup. Before the strict Pathfinder phase, **every** job installs `cuda_pathfinder`'s `test-cu12` or `test-cu13` dependency group. Consequently, the real cuDNN and NCCL evidence in the strict phase is consistently wheel-layout evidence, not standalone/local-CTK evidence.

## Why the dynamic-library test is genuinely real

The relevant test is data-driven over the descriptor-derived supported-library tables in `cuda_pathfinder/tests/test_load_nvidia_dynamic_lib.py:132`. Adding cuDNN to the descriptor therefore places it in the same real test as the established libraries.

For each library, `test_load_nvidia_dynamic_lib` launches a fresh Python subprocess (`cuda_pathfinder/tests/test_load_nvidia_dynamic_lib.py:136`). The child starts from a neutral temporary directory so it resolves the installed `cuda-pathfinder` wheel rather than accidentally importing from the checkout (`cuda_pathfinder/tests/child_load_nvidia_dynamic_lib_helper.py:18`). It then:

1. Calls the public `load_nvidia_dynamic_lib()` API.
2. Requires the first result not to report "already loaded elsewhere."
3. Requires an absolute path to a real file.
4. Calls the public API again and verifies cache identity.
5. Calls the uncached implementation and verifies the OS now reports the same file as already loaded.

Those assertions are in `cuda_pathfinder/cuda/pathfinder/_dynamic_libs/dynamic_lib_subprocess.py:51`. The production loader reaches `ctypes.CDLL`/`dlopen` on Linux (`cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_linux.py:160`) and `LoadLibraryExW` on Windows (`cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_windows.py:180`). This is not a mock, importability check, or path-exists-only test.

Strict mode is also meaningful. The expected-failure helper at `cuda_pathfinder/tests/test_load_nvidia_dynamic_lib.py:123` exempts only `mathdx` under a missing distribution and non-AArch64 `nvpl_fftw`; cuDLA has a separate platform-specific skip. cuDNN and NCCL use neither exemption. A missing cuDNN/NCCL library in `all_must_work` fails. A library found through an explicit Pathfinder path but rejected by the native loader fails in either phase; a failed OS system-search attempt may be reported as not found and tolerated only by `see_what_works`.

## Real Windows x64 evidence

All 21 Windows GPU test logs were checked. Every strict run reported:

- a real cuDNN header directory under `site-packages\nvidia\cudnn\include`;
- a real loaded `site-packages\nvidia\cudnn\bin\cudnn64_9.dll`;
- real loaded cublasLt and NVRTC DLL paths from the installed CUDA-major wheel family.

Representative CUDA 12 evidence is [job 96899599802](https://github.com/NVIDIA/cuda-python/actions/runs/32522149396/job/96899599802), Python 3.10 / CUDA 12.9 / RTX 2080 WDDM. Its first permissive phase reports cuDNN, cublasLt, and NVRTC as not found. The dependency step then installs `nvidia-cudnn-cu12==9.24.0.43` and `nvidia-cublas-cu12==12.9.2.10`. The strict phase reports:

```text
INFO test_locate_non_ctk_headers[cudnn]: hdr_dir='...\site-packages\nvidia\cudnn\include'
INFO test_load_nvidia_dynamic_lib[cudnn]: abs_path=...\site-packages\nvidia\cudnn\bin\cudnn64_9.dll
INFO test_load_nvidia_dynamic_lib[nvrtc]: abs_path=...\site-packages\nvidia\cuda_nvrtc\bin\nvrtc64_120_0.dll
INFO test_load_nvidia_dynamic_lib[cublasLt]: abs_path=...\site-packages\nvidia\cublas\bin\cublasLt64_12.dll
================= 1447 passed, 8 skipped, 2 warnings in 9.24s =================
```

Representative CUDA 13 evidence is [job 96899599699](https://github.com/NVIDIA/cuda-python/actions/runs/32522149396/job/96899599699), Python 3.14t / CUDA 13.3 / A100 MCDM. It installs `nvidia-cudnn-cu13==9.24.0.43` and `nvidia-cublas==13.6.0.2`, then reports real `cudnn64_9.dll`, `cublasLt64_13.dll`, and `nvrtc64_130_0.dll` paths and finishes the strict suite with `1447 passed, 8 skipped`.

Normal Windows jobs report `1447 passed, 8 skipped`; Python 3.15 reports `1445 passed, 10 skipped` because two CUTLASS cases are version-gated. The cuDNN cases pass rather than skip.

## Real Linux evidence

All 44 Linux GPU test logs were checked: 26 x86_64 and 18 AArch64. Every log showed the same useful before/after transition:

- Before non-CTK wheels were installed, the real `see_what_works` phase reported the cuDNN load as not found and the cuDNN/NCCL header results as `None`.
- After wheel installation, the strict phase found the real headers and loaded the real libraries from `site-packages`.

This transition is particularly good evidence: it rules out a pre-existing system cuDNN or NCCL installation masking an error in wheel discovery.

Across those 44 jobs:

- 13 installed `nvidia-cudnn-cu12==9.24.0.43` and `nvidia-nccl-cu12==2.31.2`.
- 31 installed the corresponding CUDA 13 packages at the same cuDNN/NCCL versions.
- Every strict log reports `libcudnn.so.9`, `libnccl.so.2`, `libcublasLt.so.12` or `.13`, and `libnvrtc.so.12` or `.13` as real absolute files.
- Every strict log reports real `nvidia/cudnn/include` and `nvidia/nccl/include` directories.

Representative jobs:

| Architecture/family | Job | Key evidence |
|---|---|---|
| x86_64, CUDA 12 | [96897991397](https://github.com/NVIDIA/cuda-python/actions/runs/32522149396/job/96897991397) | Real cuDNN/NCCL includes and `libcudnn.so.9`/`libnccl.so.2`; `1480 passed, 5 skipped` |
| x86_64, CUDA 13.0 | [96897991402](https://github.com/NVIDIA/cuda-python/actions/runs/32522149396/job/96897991402) | Real CUDA 13 cuDNN/NCCL headers and loads |
| x86_64, CUDA 13.3 | [96897991411](https://github.com/NVIDIA/cuda-python/actions/runs/32522149396/job/96897991411) | Real CUDA 13 cuDNN/NCCL plus cublasLt/NVRTC loads; `1480 passed, 5 skipped` |
| AArch64, CUDA 12 | [96898120424](https://github.com/NVIDIA/cuda-python/actions/runs/32522149396/job/96898120424) | Real ARM64 Linux cuDNN/NCCL headers and shared libraries |
| AArch64, CUDA 13.0 | [96898120335](https://github.com/NVIDIA/cuda-python/actions/runs/32522149396/job/96898120335) | Real ARM64 Linux CUDA 13 headers and loads; `1481 passed, 4 skipped` |
| AArch64, CUDA 13.3 | [96898120338](https://github.com/NVIDIA/cuda-python/actions/runs/32522149396/job/96898120338) | Real ARM64 Linux CUDA 13 headers and loads; `1481 passed, 4 skipped` |

The 44 strict Linux repetitions total 65,128 passed test instances and 212 expected skips. Those are matrix repetitions, not 65,128 unique tests; the value is breadth across environments rather than independent test-case count.

## Real header evidence

The relevant data-driven test is `test_locate_non_ctk_headers` at `cuda_pathfinder/tests/test_find_nvidia_headers.py:300`. It has no mocks. For each supported non-CTK product it:

- calls the public find API and the richer locate API;
- verifies they agree;
- verifies the returned directory exists;
- verifies the expected primary header file exists (`cudnn.h` or `nccl.h`);
- when the matching distribution is installed, requires a non-`None` result under `site-packages`.

The cuDNN distribution matcher is deliberately limited to backend packages (`^nvidia-cudnn-cu(?:12|13)$`), excluding `nvidia-cudnn-frontend` and JIT packages. Thus a passing installed-distribution assertion cannot be satisfied by an unrelated cuDNN-branded wheel.

This is strong evidence for the actual header-wheel layouts on Windows x64, Linux x86_64, and Linux AArch64. It does not compile a translation unit or validate transitive includes.

## Dependency-loading evidence

The cuDNN descriptor declares required `cublasLt` and optional `nvrtc` dependencies at `cuda_pathfinder/cuda/pathfinder/_dynamic_libs/descriptor_catalog.py:426` and `cuda_pathfinder/cuda/pathfinder/_dynamic_libs/descriptor_catalog.py:427`. Production loading calls `load_dependencies()` before attaching cuDNN (`cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_nvidia_dynamic_lib.py:180`). Required-dependency errors propagate; an optional dependency suppresses only an exact not-found condition (`cuda_pathfinder/cuda/pathfinder/_dynamic_libs/load_dl_common.py:34`).

The strict environments contain real cublasLt and NVRTC wheels, and every platform/family log shows both can be loaded as real native libraries. Therefore the run proves:

- the required cublasLt preload path succeeds with the current CUDA 12 and 13 wheel families;
- the optional-NVRTC-present path does not break cuDNN loading;
- both dependency libraries are independently loadable in the same environment.

The per-cuDNN child payload does not emit a nested trace for each dependency. The source call graph guarantees the attempt, while the separate real load records establish artifact availability. The optional-NVRTC-absent case remains callback-based unit coverage rather than a real-binary integration scenario.

## Coverage of the CI-failure fix

The previous public run failed on two Windows-host path-string assumptions. In representative Windows job [96899599699](https://github.com/NVIDIA/cuda-python/actions/runs/32522149396/job/96899599699), both corrected tests pass in both Pathfinder phases:

- `test_find_in_system_install_dirs_expands_program_files_and_prefers_newest_cudnn`
- `test_find_in_system_install_dirs_uses_running_linux_multiarch`

The first now exercises the host's real `glob` implementation over actual temporary `v9.9` and `v9.10` directories, so it is a useful cross-platform filesystem regression test. The `cudnn.h` files are synthetic, however. The second intentionally patches Linux platform/sysconfig/glob behavior and also uses a synthetic header; it is a unit portability test, not evidence of a real Linux package installation.

The two recurring Pathfinder warnings in representative jobs are pytest configuration/marker warnings for `thread_unsafe_fixtures`/`thread_unsafe`; they do not skip or mask the cuDNN/NCCL tests.

## Coverage map for PR 2680 behavior

| Behavior | Best public-CI evidence | Judgment |
|---|---|---|
| cuDNN wheel discovery and top-level native load, Windows x64 | Real artifact, real `LoadLibraryExW`, 21/21 strict jobs | Sufficient |
| cuDNN wheel discovery and top-level native load, Linux x86_64 | Real artifact, real `dlopen`, 26/26 strict jobs | Sufficient |
| cuDNN wheel discovery and top-level native load, Linux AArch64 | Real artifact, real `dlopen`, 18/18 strict jobs | Sufficient |
| cuDNN wheel header discovery on those platforms | Real wheel, real `cudnn.h`, strict assertion | Sufficient |
| NCCL wheel header discovery on both Linux architectures | Real wheel, real `nccl.h`, strict assertion | Sufficient |
| Required cublasLt and present optional NVRTC | Real libraries plus production dependency call graph | Sufficient for present-dependency path |
| Optional NVRTC absent | Callback-based unit tests only | Semantics covered; no real integration proof |
| `CUDNN_PATH` x64/ARM64 archive layouts | Production search over temp directories and placeholder DLLs | Structurally covered only |
| Program Files cuDNN discovery/version preference | Real glob/temp tree, placeholder headers/DLLs | Structurally covered only |
| `CUDNN_PATH` and `NCCL_HOME` header roots | Real temp filesystem, placeholder headers | Structurally covered only |
| Linux `/usr/include`, `/usr/local/include`, multiarch headers | Mocked glob/sysconfig plus placeholder headers | Unit covered only |
| Windows ARM64 cuDNN loading | ARM64 context, zero-byte DLL, mocked dependency/native loader | Unit covered only; no real WoA evidence |
| cuDNN/NCCL API or workload execution | No test invokes an exported product API | Not covered |
| Header compile/transitive includes | Primary file existence only | Not covered |
| Conda layouts | Mocked/unit tests | No real integration proof |

## Residual risks and recommended follow-ups

### 1. Native Windows ARM64 remains the largest gap

Public CI has no Windows ARM64 job. The WoA orchestration test at `cuda_pathfinder/tests/test_load_nvidia_dynamic_lib_using_mocker.py:55` creates a zero-byte `cudnn64_9.dll`, substitutes an ARM64 search context, mocks dependency loading, and mocks the native loader. Together with the search-step tests, it usefully proves `CUDNN_PATH/bin/arm64` routing, dependency order, and x64/ARM64 directory isolation. It does not inspect the placeholder's PE architecture and proves nothing about a real ARM64 PE, native Python, `AddDllDirectory`, or `LoadLibraryExW`.

Minimum high-value follow-up: on a native WoA runner, provision a pinned/checksummed official ARM64 cuDNN archive plus compatible cublasLt, set `CUDNN_PATH`, run the strict `test_load_nvidia_dynamic_lib[cudnn]` case (or call the public API and require success), and record the returned real path. Add NVRTC when validating the optional-present path. If WoA support is to be called production-validated, this should be required rather than optional.

### 2. Standalone and system installation roots are not real-tested

No public job uses a real standalone cuDNN archive, a real Program Files installation, Linux cuDNN DEB/RPM headers, a real `NCCL_HOME`, or conda. In fact, all 44 Linux pre-install phases report those products absent, confirming that the strict successes come from wheels.

Recommended follow-ups, in priority order:

1. A Windows x64 standalone-archive smoke with `CUDNN_PATH`.
2. A native WoA standalone-archive smoke with `CUDNN_PATH/bin/arm64`.
3. A small Linux container job installing the distro header packages and checking system/multiarch lookup without the wheel present.
4. A real NCCL source/build-root or tarball test for `NCCL_HOME/include` and `NCCL_HOME/build/include`.

### 3. Native attachment is not product functionality

The real dynamic test proves that the top-level cuDNN/NCCL library can be attached and that Pathfinder reports the right file. It does not resolve a product symbol, query a version, create a cuDNN handle, or run a convolution/graph operation. In particular, it may not force every component that cuDNN loads lazily.

A lightweight `cudnnGetVersion` smoke would add useful assurance with little runtime. A representative engine/API operation would be stronger for validating late-loaded components and runtime dependencies, but is beyond Pathfinder's minimal find/load contract and need not block this PR if the claim remains limited to discovery and loading.

### 4. Version breadth is environmental, not contractual

The requirements correctly constrain cuDNN to major 9, but the resolver selected only `9.24.0.43` in this run. Similarly, NCCL resolved only `2.31.2`. The broad matrix repeats these artifacts across Python, toolkit, architecture, GPU, and driver-mode combinations; it does not demonstrate every supported minor wheel version.

## Final assessment

This is a high-quality green run for a discovery/loading library. The most important paths are not merely mocked: the CI installs the built Pathfinder wheel and current NVIDIA wheels, observes the products absent before installation, discovers real headers afterward, and attaches real libraries through fresh-process native loader calls under strict failure semantics. Coverage is broad across 65 GPU environments and both CUDA-major package families.

Accordingly, the public CI is sufficient to support merging the Windows x64 and Linux wheel behavior in PR 2680 and to demonstrate that the two Windows portability regressions are fixed.

The result should not be summarized as "all supported layouts and architectures are real-tested." Windows ARM64, standalone/system roots, conda, optional-NVRTC absence, header compilation, and post-load product behavior remain explicitly outside the real integration evidence. A native WoA archive smoke is the one follow-up most likely to change confidence materially.

kkraus14 (approved):

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_platform.py
```diff
@@ -63,8 +63,13 @@ def _find_so_in_rel_dirs(
...
 
 
-def _find_dll_under_dir(dirpath: str, file_wild: str, target_arch: str | None = None) -> str | None:
-    for path in sorted(glob.glob(os.path.join(dirpath, file_wild))):
+def _find_descriptor_dll_under_dir(
+    dirpath: str,
+    desc: LibDescriptor,
+    target_arch: str | None = None,
+) -> str | None:
+    for dll_basename in reversed(cast(tuple[str, ...], desc.windows_dlls)):
```
kkraus14: Non-blocking: making every Windows filesystem search exact fixes the cuDNN sidecar ambiguity, but it also removes the previous forward-compatible wildcard lookup for versioned DLL names. For example, the catalog currently ends at `cupti64_2026.3.0.dll`, so a newer CUPTI DLL in a wheel, conda env, or toolkit root will be ignored until Pathfinder is updated. Could exact-name selection be an opt-in descriptor policy (enabled for cuDNN), while retaining wildcard discovery for descriptors such as CUPTI?

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/descriptor_catalog.py
```diff
@@ -411,6 +415,27 @@ class DescriptorSpec:
...
+        # The ARM64 layout is verified only for the standalone archive rooted
+        # at CUDNN_PATH, not for conda, CUDA_PATH, or Program Files installs.
+        anchor_rel_dirs_windows=WindowsSearchDirs.x64_only("bin/x64", "bin"),
+        install_root_env_vars_windows=("CUDNN_PATH",),
+        install_root_env_rel_dirs_windows=WindowsSearchDirs(
+            x64=("bin/x64", "bin"),
+            arm64=("bin/arm64",),
+        ),
+        program_files_root_globs_windows=WindowsSearchDirs.x64_only("NVIDIA/CUDNN/v9.*"),
+        requires_add_dll_directory=True,
```
kkraus14, line 443: Non-blocking edge case: this side effect is skipped when `cudnn64_9.dll` was already loaded from a late root such as `CUDNN_PATH` or Program Files. `_load_lib_no_cache()` passes `find is not None` to the already-loaded check, but `find` contains only the early wheel/conda result; the function then returns before late-root discovery. That can leave cuDNN's lazily loaded component DLLs undiscoverable. Could the already-loaded path register its resolved module directory whenever `requires_add_dll_directory` is set, independent of the early-find boolean, with a regression test for a preloaded standalone install?

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_platform.py
```diff
@@ -135,9 +150,19 @@ def conda_anchor_point(self, conda_prefix: str) -> str:
     def anchor_rel_dirs(self, desc: LibDescriptor) -> tuple[str, ...]:
         return cast(tuple[str, ...], desc.anchor_rel_dirs_linux)
 
+    def install_root_env_vars(self, _desc: LibDescriptor) -> tuple[str, ...]:
+        return ()
```
kkraus14: Non-blocking: these no-op Linux hooks make the new product-root dynamic search Windows-only. Header discovery accepts `CUDNN_PATH` and `NCCL_HOME`, but dynamic discovery will not inspect locations such as `$CUDNN_PATH/lib{,64}` or `$NCCL_HOME/{lib,build/lib}` unless users also configure the OS loader path. Could we add Linux product-root metadata/relative directories and focused tests so header and library discovery stay aligned?

kkraus14 (commented):
Non-blocking follow-ups from a local review of the current head. CI and the local suite are green; the inline comments cover forward-compatible Windows DLL matching, DLL-directory registration for an already-loaded standalone cuDNN, and Linux product-root dynamic discovery. I omitted the native Windows ARM64 wheel-dependency point because public CUDA 13.4/WoA packaging is still pending.

rwgk:
Thanks @kkraus14!

I still haven't fully reviewed myself. I'm working on making the internal CI more complete.

I have a fully successful internal CI run, but (codex analysis):

> One important exception must remain explicit: the internal WoA workflow does **not** provision a real ARM64 cuDNN archive or ARM64 cublasLt. It therefore does not load a real `CUDNN_PATH/bin/arm64/cudnn64_9.dll`. The new WoA-specific cuDNN route receives structural and mocked coverage on a real ARM64 host, while native cuDNN attachment on WoA remains the highest-value follow-up.

I believe it's fixable; and it's a good value, long-term.

rwgk:
## PR 2680: final ctk-next CI analysis with real WoA cuDNN coverage

Analyzed against the final green CI for ctk-next PR 537, which mirrors the product changes in public PR 2680 and adds ctk-next-only provisioning and verification for native Windows ARM64.

## Executive conclusion

The intended ctk-next matrix is fully green. Four platform workflows ran on Linux x86_64, Linux AArch64, Windows x64, and native Windows ARM64. Each completed its source/runbook job, wheel-build job, and isolated installed-wheel test job: 12/12 active platform jobs succeeded. Bandit also succeeded. The complete check rollup contains 13 successes and 17 checks intentionally skipped by branch or policy routing; there were no failures or cancellations.

This result complements the already-posted [public CI analysis for PR 2680](https://github.com/NVIDIA/cuda-python/pull/2680#issuecomment-5375625698):

- Public CI supplies broad strict real-artifact coverage for released packages. It discovers real cuDNN headers and loads real cuDNN libraries on Windows x64, Linux x86_64, and Linux AArch64; it does the same for NCCL on both Linux architectures, across CUDA 12 and CUDA 13.
- ctk-next PR 537 supplies CUDA 13.4 and native-platform coverage. It builds and tests the equivalent product patch on all four ctk-next platforms, exercises editable source and isolated installed wheels, and now provisions and tests the official standalone ARM64 cuDNN archive plus real ARM64 cuBLASLt and NVRTC on a native Windows ARM64 GB10 system.

## Method

I used the same evidence standard as the public report:

| Level | Meaning | ctk-next examples |
|---|---|---|
| Real integration | A real native platform and real artifact reach production code without replacing the relevant operation with a mock | ARM64 cuDNN/cuBLASLt/NVRTC discovery and loading; real cuDNN exports and handle creation; wheel imports on native ARM64 |
| Real filesystem, synthetic artifact | Production path logic operates on temporary directories or placeholder files | Unit tests for alternate roots, architecture isolation, and malformed layouts |
| Mocked/static | The loader, dependency callback, search context, or artifact is replaced | Focused orchestration unit tests that validate control flow independently of native loading |

For the new WoA conclusion, only the first level is used as proof. The dedicated strict tests and semantic smoke use the production Pathfinder public APIs, real ARM64 PE files, the native Windows loader, a real driver, and a real GB10 GPU. Their relevant discovery and loading operations are not mocked.

I inspected the final check inventory, all four completed platform workflows, both WoA test-bearing job logs, and the exact workflow and verification source at the tested ctk-next head.

## Run inventory

All four active platform workflows ran at the same final ctk-next PR 537 head:

| Platform | Source/runbook | Wheel build | Isolated installed-wheel test |
|---|---:|---:|---:|
| Linux x86_64 | Passed | Passed | Passed |
| Linux AArch64 | Passed | Passed | Passed |
| Windows x64 | Passed | Passed | Passed |
| Windows ARM64 | Passed | Passed | Passed |

Bandit also passed. Across the final PR check rollup, 13 checks succeeded and 17 were skipped. The skipped generic CI, security, labeling, and PR-policy checks are expected for this ctk-next ref-routing arrangement and should not be counted as test coverage. The four three-job platform workflows are the active matrix.

## Matrix and Pathfinder results

All platform jobs use CUDA 13.4.1, kitpick 011, and Python 3.13. Linux test-bearing jobs run in Ubuntu 24.04 containers with driver 615.62. Windows reports Python 3.13.14 with the native architecture for each runner.

The full Pathfinder suite ran twice on every platform: once from the editable/source runbook installation and once from the built wheel in the isolated wheel-test environment.

| Platform | Native identity | Editable/source result | Installed-wheel result |
|---|---|---:|---:|
| Linux x86_64 | `linux-64`, CUDA 13.4.1 | 1,481 passed, 4 skipped | 1,481 passed, 4 skipped |
| Linux AArch64 | `linux-aarch64`, CUDA 13.4.1 | 1,481 passed, 4 skipped | 1,481 passed, 4 skipped |
| Windows x64 | AMD64 CPython, `win-amd64` | 1,447 passed, 8 skipped | 1,447 passed, 8 skipped |
| Windows ARM64 | ARM64 CPython, `win-arm64` | 1,448 passed, 8 skipped | 1,448 passed, 8 skipped |

The eight full-suite executions total 11,714 passed test instances and 48 skips. These totals reflect repeated source/wheel and platform coverage, not 11,714 unique tests.

In addition, each WoA test-bearing job ran four dedicated strict real-artifact Pathfinder checks. Those eight additional test instances all passed. Each job then ran the semantic native cuDNN verification script successfully.

The wheel jobs create a fresh virtual environment, install the locally built Pathfinder, bindings, and core wheels, move the checkout package trees aside, and fail if `cuda.pathfinder`, `cuda.bindings`, or `cuda.core` imports from the workspace. The installed-wheel WoA result therefore comes from the actual built artifacts rather than an accidental checkout import. `pip check` also reports no broken requirements.

## What ctk-next CI proves with real artifacts

### CUDA 13.4 and production Pathfinder loading

The logs contain broad production-loader evidence against the real CUDA 13.4 kitpick:

- Linux x86_64 and AArch64 load real CUDA 13.4 libraries, including cuBLASLt, NVRTC, CUDA Runtime, cuBLAS, CUPTI, NVVM, nvJitLink, nvFatBin, cuFFT, and cuSOLVER families.
- Windows x64 loads real native Windows CUDA 13.4 DLLs from the x64 toolkit layout.
- Windows ARM64 loads real driver and CUDA DLLs through production Pathfinder/native-loader paths.
- The runs discover real toolkit headers, libdevice, and cudadevrt artifacts and exercise bindings/core against real drivers and GPUs.

This directly exercises the shared install-root/search refactoring under CUDA 13.4 and shows that the new cuDNN metadata and dependency handling do not regress established CUDA library discovery on either Windows architecture.

### Native Windows ARM64

The WoA workflow is genuinely native, not x64 emulation with a patched platform string:

- The runner gate verifies `PROCESSOR_ARCHITECTURE=ARM64`.
- CPython reports `64 bit (ARM64)` and `sysconfig` reports `win-arm64`.
- MSVC uses the native `HostARM64\ARM64\cl.exe` toolchain.
- The wheel build produces and validates `cp313-cp313-win_arm64` bindings and core wheels.
- The installed-wheel job imports all three packages from the ARM64 wheel-test environment after excluding checkout sources.
- The runner exposes an NVIDIA GB10 GPU; the logs report driver 620.16, and bindings/core GPU suites execute successfully.

The full Pathfinder suite passes twice on this native host. Architecture detection, descriptor invariants, Windows path handling, source packaging, and the installed wheel therefore all execute successfully under real ARM64 Python and Windows filesystem semantics.

## Real WoA cuDNN, cuBLASLt, and NVRTC coverage

### Artifact provisioning

The successful source/runbook job had a cuDNN cache miss. It therefore exercised the complete cold path:

1. Download the pinned official cuDNN 9.25.0.15 CUDA 13 Windows ARM64 archive.
2. Verify the complete archive against its pinned SHA-256 digest.
3. Selectively extract the runtime DLLs, headers, license, and import library.
4. Require the expected archive root and required files.
5. Require the exact set of ten cuDNN component DLLs.
6. Set `CUDNN_PATH` to the archive root while deliberately keeping `bin\arm64` off `PATH`.
7. Save the validated result under the checksum-specific cache key.

This matters because it proves discovery through the new descriptor-driven `CUDNN_PATH/bin/arm64` route. A pre-existing PATH entry cannot accidentally satisfy the test.

The CUDA 13.4 mini-toolkit contains real ARM64 `cublas64_13.dll` and `cublasLt64_13.dll`, and its configuration step hard-validates both files. It also contains real `nvrtc64_130_0.dll`. The final source job restored this validated CUDA cache; the artifacts were then resolved and loaded natively by Pathfinder.

The installed-wheel job restored the checksum-keyed cuDNN cache produced by the source job and re-ran the required-file and exact-DLL-set validation. The two jobs therefore cover both the cold download/extraction path and the subsequent cache-restore path.

### Dedicated strict Pathfinder checks

Both the source/runbook environment and the isolated installed-wheel environment set:

```text
CUDA_PATHFINDER_TEST_FIND_NVIDIA_HEADERS_STRICTNESS=all_must_work
CUDA_PATHFINDER_TEST_LOAD_NVIDIA_DYNAMIC_LIB_STRICTNESS=all_must_work
```

Each then runs four targeted cases, all of which pass:

1. Locate real cuDNN headers.
2. Load real ARM64 cuBLASLt.
3. Load real ARM64 NVRTC.
4. Load real ARM64 cuDNN.

The dynamic-library cases run production Pathfinder in fresh subprocesses and use the native Windows loader. They do not mock the search context, dependency callback, loader, or DLL.

The logged absolute paths are the intended artifacts:

```text
CUDNN_PATH\include
CUDNN_PATH\bin\arm64\cudnn64_9.dll
CUDA_PATH\bin\arm64\cublasLt64_13.dll
CUDA_PATH\bin\nvrtc64_130_0.dll
```

The cuDNN result cannot succeed merely by finding the shim. Its descriptor first loads required cuBLASLt and attempts optional NVRTC, registers the cuDNN directory for native dependency resolution, and then attaches the real ARM64 cuDNN DLL.

### Semantic native API smoke

After the strict pytest cases, both environments run a stronger direct verification through the production Pathfinder APIs. It proves all of the following:

- The machine and Python process are natively ARM64.
- The headers resolve exactly to `CUDNN_PATH\include`, with `found_via == "CUDNN_PATH"`.
- The headers declare cuDNN 9.25.0.
- After clearing Pathfinder's cache, one cuDNN request produces exactly three cache misses and three entries: cuDNN, required cuBLASLt, and optional NVRTC.
- cuDNN resolves exactly to `CUDNN_PATH\bin\arm64\cudnn64_9.dll`.
- cuBLASLt resolves exactly to `CUDA_PATH\bin\arm64\cublasLt64_13.dll`, with `found_via == "CUDA_PATH"`.
- NVRTC resolves exactly to `CUDA_PATH\bin\nvrtc64_130_0.dll`.
- All ten cuDNN component DLLs, cuBLASLt, and NVRTC are valid ARM64 PE files.
- `cudnnGetVersion()` returns `92500`, matching the 9.25.0 headers.
- `cudnnGraphVersionCheck`, `cudnnOpsVersionCheck`, `cudnnCnnVersionCheck`, and `cudnnAdvVersionCheck` all return success, exercising cuDNN's Graph/Ops/CNN/Adv component-compatibility checks.
- `cudnnCreate` returns success and a non-null real handle on the GB10 system, and `cudnnDestroy` succeeds.

This goes beyond a top-level `LoadLibraryExW` smoke. It validates symbol resolution, component-version compatibility, dependency orchestration, and real cuDNN initialization against the driver/GPU.

### Source and installed-wheel repetition

The entire strict and semantic sequence succeeds twice:

| Environment | Strict targeted tests | Semantic API smoke | Full Pathfinder suite |
|---|---:|---:|---:|
| Editable/source runbook | 4 passed | Passed | 1,448 passed, 8 skipped |
| Isolated installed wheel | 4 passed | Passed | 1,448 passed, 8 skipped |

The installed-wheel job additionally runs the broader native package suites successfully. Its aggregate result, including the four targeted Pathfinder checks, is 6,265 passed, 414 skipped, two expected xfails, and zero failures, plus 29 successful subtests. This is useful corroboration that the real cuDNN path coexists with the complete native wheel stack rather than succeeding only in a narrowly constructed loader process.

## Combined public and ctk-next coverage

The two runs now cover complementary axes, including the intersection that was missing from the first ctk-next report:

| Behavior or dimension | Public PR 2680 CI | ctk-next PR 537 CI | Combined assessment |
|---|---|---|---|
| Real cuDNN wheel headers and native load | Strict on Windows x64, Linux x86_64, and Linux AArch64; CUDA 12/13 | Not the focus | Strong for released wheel platforms |
| Real NCCL wheel headers and native load | Strict on Linux x86_64 and AArch64; CUDA 12/13 | Not the focus | Strong for supported Linux wheel platforms |
| Real Windows ARM64 cuDNN archive discovery | Not available in public matrix | Real official archive via `CUDNN_PATH` | Covered end to end |
| Real Windows ARM64 cuDNN native attachment | Not available in public matrix | Real top-level DLL; exact ten-component set present and ARM64; Graph/Ops/CNN/Adv checks succeed | Covered end to end |
| Required cuBLASLt and present optional NVRTC in the WoA cuDNN chain | Not available in public matrix | Real CUDA 13.4 ARM64 DLLs and exact dependency-cache proof | Covered end to end for Pathfinder loading |
| Real WoA cuDNN API initialization | Not available in public matrix | Version checks and `cudnnCreate`/`cudnnDestroy` on GB10 | Strong minimal semantic smoke |
| ctk-next/CUDA 13.4 compatibility | Not covered | Real CUDA 13.4.1 on four platforms | Strong |
| Installed-artifact isolation | Built public artifacts in public matrix | Explicit source exclusion and isolated locally built wheels on all four platforms | Strong |
| Native Windows ARM64 platform and wheel pipeline | Not present | Real ARM64 OS, Python, MSVC, wheel tags, CUDA DLLs, driver, and GB10 GPU | Strong |
| Standalone, Program Files, Linux system-package, and conda product layouts | Synthetic or mocked depending on route | Real `CUDNN_PATH` archive only | Mixed; keep other layout claims scoped |
| cuDNN/cuBLASLt numerical workloads | Not covered | Not covered | Outside current Pathfinder integration scope |

The most accurate single-sentence combined summary is:

> Public CI proves strict real cuDNN/NCCL wheel discovery and loading across the released x64/Linux platforms, while ctk-next PR 537 proves the same Pathfinder changes under CUDA 13.4 on all four ctk-next platforms and adds native WoA end-to-end coverage of the official ARM64 cuDNN archive, required cuBLASLt, optional NVRTC, and real cuDNN initialization from both source and an isolated installed wheel.

rwgk:
- Public PR 2680: 23ba0f16393375bc6bbdffe933145b526104d25a
    - 105 successful checks, 3 expected skips.
    - Fresh Windows pre-commit passed, validating the Cython pin.

- Private PR 537: `e096ce986950f8c547df2f11e8dbff04983f86d0`
    - Linux x64/ARM64, Windows x64/ARM64, Bandit, source tests, wheel builds, and installed-wheel tests all passed.
    - Native WoA real-cuDNN verification passed in both source and installed-wheel environments.

rwgk:
For completeness, companion comment for commit 955b5e7404d41bcba04c0008ceec4d81eeec527e:

The first version mixed installation roots with library directories: `program_files_rel_dirs_windows` repeated `bin/x64` and `bin`, even though `anchor_rel_dirs_windows` already owned that layout.

This follow-up gives each descriptor field one responsibility:

- `optional_dependencies` stays unchanged because it accurately models conditional runtime use such as cuDNN's NVRTC path.
- `install_root_env_vars_windows` names product-specific variables whose values are installation roots, such as `CUDNN_PATH`.
- `program_files_root_globs_windows` names versioned installation roots under Program Files and remains architecture-indexed; `anchor_rel_dirs_windows` supplies `bin/x64` and `bin` once for every root source.

Root-based searches now share one root -> anchor -> DLL path, try every declared anchor before failing, and match the descriptor's exact DLL candidates so a cuDNN component DLL cannot be mistaken for `cudnn64_9.dll`. This removes duplicated layout metadata while preserving the public API, top-level search order, and `found_via` labels. Tests cover `CUDNN_PATH`, Program Files version ordering across both layouts, stale component-only anchor directories, exact DLL matching, and x64 gating.

rwgk:
For completeness, companion comment for commit ddaa78fc01138188526a1099f53e4c7e8fc33380:

NVIDIA now publishes a CUDA 13 [cuDNN 9.25 standalone archive for Windows ARM64](https://developer.download.nvidia.com/compute/cudnn/redist/cudnn/windows-arm64/cudnn-windows-arm64-9.25.0.15_cuda13-archive.zip) with `bin/arm64/cudnn64_9.dll`. This follow-up adds support for that verified layout without generalizing beyond the available evidence.

- cuDNN is available to native ARM64 Python, and `CUDNN_PATH/bin/arm64` is searched.
- Pip remains x64-only, and ARM64 conda, `CUDA_PATH`, and Program Files layouts remain unsupported. A new `install_root_env_rel_dirs_windows` field keeps the product-root layout separate from the general anchors shared by those other sources.
- Simulated tests cover x64 and ARM64 archive discovery, cross-architecture isolation, all unsupported-root boundaries, and the loader path through required cuBLASLt plus optional NVRTC handling.

I intentionally did not add a `platform_machine == 'ARM64'` marker to the shared test dependency. PEP 508 defines that value through [`platform.machine()`](https://packaging.python.org/en/latest/specifications/dependency-specifiers/#environment-markers), while current CPython on Windows prefers WMI's native CPU architecture ([implementation](https://github.com/python/cpython/blob/main/Lib/platform.py)). Such a marker can therefore also suppress the compatible `win_amd64` wheel for x64 Python emulated on WoA. Native WoA CI should instead install the base test dependencies and provision the standalone cuDNN/CUDA artifacts explicitly.

The ctk-next PR 537 WoA job subsequently completed this validation using the real archive on native Windows ARM64. It verified strict Pathfinder discovery and loading, real cuBLASLt and NVRTC dependency resolution, ARM64 PE identity, cuDNN component-compatibility checks, and successful cudnnCreate/cudnnDestroy from both source and installed-wheel environments. This validates the loader and DLL-directory behavior beyond the simulated tests; it is not a numerical cuDNN workload.

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/descriptor_catalog.py
rwgk: Done: commit 6e700565b511a655cc212e3efcb9438ea6dfab45

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_platform.py
rwgk: Done: commit 6dbbe8d2eea80b73e63daaffc4c33f442ce0d129

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/search_platform.py
rwgk: Addressed in two commits:
  
c1cd5850b7ee — Restore forward-compatible Windows DLL discovery

It restored wildcard discovery while allowing cuDNN to require exact names.

The implementation was subsequently refined by:

e5c91aa97c93 — Constrain Windows DLL fallback matching

That second commit established the final policy: exact known names first, with explicit descriptor-specific fallback globs—currently only `cupti64_*.dll`.

PR #2683, opened 2026-08-21
Isolate CUmemLocation construction in a helper

juenglin:
## Summary

Cython's aggregate struct-literal syntax (`CUmemLocation(type=..., id=...)`) warns "Not all members given for struct" whenever a call site doesn't name every member Cython knows about for that struct — and `cuda.core` promotes that warning to a build error. CUDA 13.4 adds a `localized` union member to `CUmemLocation_st` alongside the existing `id`, so once `cydriver.pxd` is regenerated from 13.4 headers, every existing `CUmemLocation(type=..., id=...)` call site in `cuda.core` fails to build, even though `id` is a complete initializer for the union arm actually being used.

This PR consolidates all `CUmemLocation` construction in `cuda_core/cuda/core/_memory/_location.pxd` into two small helpers that use field assignment internally (so the warning never fires there), while call sites keep compact constructor-style syntax:

- `to_cumemlocation(kind: str, loc_id: int)` — for callers with a semantic `kind` (`"device"` / `"host"` / `"host_numa"` / `"host_numa_current"`), e.g. from `_LocSpec.kind`.
- `cumemlocation_from_type(loc_type: CUmemLocationType, loc_id: int)` — for callers that already hold a `CUmemLocationType` enum value directly (pool configuration, peer-access, graph alloc nodes).

Every `CUmemLocation`-constructing call site across `cuda_core` now routes through one of these two helpers instead of duplicating field-assignment or aggregate-literal logic inline.

## Relationship to #2673

Both PRs fix the same build failure. #2673 replaces every aggregate constructor with bare field assignment at each call site directly — minimal diff, already passed full CI, but it also removes the partial-initialization warning for this struct everywhere it touches. This PR keeps constructor-style call sites and centralizes the field-assignment logic in two helpers instead of duplicating it across call sites. The safety benefit is modest: Cython's aggregate-literal check isn't union-aware and can't verify "the correct arm was set" either way, so the real gain here is having one tested place for the mapping logic rather than six.

## Test plan

- [x] Rebuild `cuda_core` against local CUDA 13.3 headers — no `-Werror` failures
- [x] `cython-lint` passes on all changed files
- [x] Memory, peer-access, managed-ops, pool, and graph-memory tests — 389 passed, 29 skipped

### DIFF
--- a/cuda_core/build_hooks.py
+++ b/cuda_core/build_hooks.py
@@ -121,6 +121,31 @@ def _determine_cuda_major_version() -> str:
     )
 
 
+@functools.cache
+def _cuda_core_has_localized_location() -> bool:
+    """Whether CUmemLocation exposes the ``localized`` union arm (CUDA 13.4+).
+
+    CUDA 13.4 adds ``CUmemLocation.localized`` and
+    ``CU_MEM_LOCATION_TYPE_DEVICE_LOCALITY_DOMAIN``. This flag is independent of
+    ``CUDA_CORE_BUILD_MAJOR`` so 13.3 and 13.4 can share a major version while
+    still compiling different ``to_cumemlocation`` signatures.
+    """
+    override = os.environ.get("CUDA_CORE_HAS_LOCALIZED_LOCATION")
+    if override is not None:
+        return bool(int(override))
+    cuda_path = _get_cuda_path()
+    cuda_h = os.path.join(cuda_path, "include", "cuda.h")
+    try:
+        with open(cuda_h, encoding="utf-8") as f:
+            for line in f:
+                m = re.match(r"^#\s*define\s+CUDA_VERSION\s+(\d+)\s*$", line)
+                if m:
+                    return int(m.group(1)) >= 13040
+    except OSError:
+        pass
+    return False
+
+
 # used later by setup()
 _extensions = None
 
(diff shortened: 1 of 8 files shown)


rwgk:
It's looking very good! I'll help with the 13.4 testing (I'll share links).

```
• Findings

  1. Medium: CUDA 13.4 compilation remains unverified. The PR’s 114 checks completed successfully, but the public matrix and the stated
     local build use CUDA 13.3 or older. Since the change specifically targets the 13.4 CUmemLocation declaration, I would require one build
     against the generated 13.4 bindings before merging.

  2. Low: the typed helper’s contract is broader than its implementation. cuda_core/cuda/core/_memory/_location.pxd:24 accepts any
     CUmemLocationType but always initializes the id union arm. A future caller could pass CU_MEM_LOCATION_TYPE_DEVICE_LOCALITY_DOMAIN,
     which requires localized. Renaming it to cumemlocation_from_id would make the intended restriction clear; explicitly rejecting non-ID-
     backed types would be stronger.

  3. Low: the CUDA 12 comment and structure are stale. cuda_core/cuda/core/_memory/_location.pxd:7 says locations are populated only for
     CUDA 13, but the new helper is used on CUDA 12 paths too. Because cumemlocation_from_type has identical implementations in both
     branches, it could be defined once outside the IF.
```

rwgk:
Internal CI (ctk-next 538) passed!

rwgk (approved):
This is a really nice solution. I'm approving, but could you please apply the two Low findings under https://github.com/NVIDIA/cuda-python/pull/2683#issuecomment-5375452050 before you merge?

juenglin:
/ok to test https://github.com/NVIDIA/cuda-python/commit/6589041d1e717d16fb2fbc0f37248d1a8d6b8adf

PR #2688, opened 2026-08-23
Avoid aggregate graph access descriptor initialization

rwgk:
## Description

Follow-up to #2683.

PR #2683 centralizes `CUmemLocation` construction, but the graph allocation path still embeds the returned location in a `CUmemAccessDesc` aggregate constructor. With the CUDA 13.4 generated declaration, that nested aggregate causes Cython to emit an invalid conversion for the added `localized` union arm.

Declare the access descriptor separately and assign its `location` and `flags` fields before adding it to the descriptor vector. This keeps the helper introduced by #2683 and preserves the existing runtime behavior.

Please see ctk-next PR 539 for background on how this escaped attention while validating #2683.

## Validation

- The equivalent change is validated against the CUDA 13.4 generated bindings in ctk-next PR 539.

## Checklist

### DIFF
--- a/cuda_core/cuda/core/graph/_graph_node.pyx
+++ b/cuda_core/cuda/core/graph/_graph_node.pyx
@@ -828,6 +828,7 @@ cdef inline AllocNode GN_alloc(GraphNode self, size_t size, object device,
         num_deps = 1
 
     cdef vector[cydriver.CUmemAccessDesc] access_descs
+    cdef cydriver.CUmemAccessDesc access_desc
     cdef int peer_id
     cdef list peer_ids = []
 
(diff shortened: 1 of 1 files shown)


juenglin (approved):

rwgk:
## Additional rationale for merging this PR

The need for this change appears to arise from a Cython code-generation issue exposed by cybind's representation of CUDA's anonymous nested struct:

- The aggregate initializer causes Cython to emit an unused struct-to-Python conversion helper.
- That helper attempts to convert the actual anonymous C++ type of `localized` to cybind's synthetic named struct type, which the generated C++ compiler correctly rejects as incompatible.
- Field-by-field initialization compiles because it avoids generating that conversion helper.
- C struct aggregate initialization is supported Cython syntax, so the original call site was not misusing Cython. See the [Cython language basics documentation](https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#structs-unions-enums).

The field-by-field initialization in this PR is therefore a narrowly scoped workaround that preserves the intended descriptor contents and behavior.

Issue #2631, opened 2026-08-14
[FEA]: Support CU_LAUNCH_ATTRIBUTE_PRIORITY in LaunchConfig


lijinf2:
### Is this a duplicate?

- [x] I confirmed there appear to be no [duplicate issues](https://github.com/NVIDIA/cuda-python/issues) for this request and that I agree to the [Code of Conduct](CODE_OF_CONDUCT.md)

### Area

cuda.core

### Is your feature request related to a problem? Please describe.

`LaunchConfig` does not expose [`CU_LAUNCH_ATTRIBUTE_PRIORITY`](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1g6f6565b334be6bb3134868e10bbdd331), which sets the launch priority for a kernel.

Parent tracking: #496.

### Describe the solution you'd like

- [ ] Users can set `CU_LAUNCH_ATTRIBUTE_PRIORITY` via `LaunchConfig`.
- [ ] Mapping to the native launch attribute is tested.

### Describe alternatives you've considered

Use `cuda.bindings.driver` (`CUlaunchAttribute` / `CU_LAUNCH_ATTRIBUTE_PRIORITY`) and `cuLaunchKernelEx` directly.

Downside: leaves `cuda.core` users without a first-class `LaunchConfig` API; they must drop to low-level bindings, manage the attribute array themselves, and lose consistency with other `LaunchConfig` launch attributes (`is_cooperative`, `programmatic_stream_serialization`, etc.).

### Additional context

Incremental `LaunchConfig` attribute coverage under #496 (same approach as #1334 for `programmatic_stream_serialization`).

No example code in the design notes for this attribute.

ichensun:
The goal is to expose CUDA's per-kernel launch priority through `cuda.core.LaunchConfig`, avoiding the need for users to construct `CUlaunchAttribute` manually with `cuda.bindings`.

We can add an optional `priority` field to `LaunchConfig`, declaring its Cython storage in `_launch_config.pyd`, and maping supplied values to `CU_LAUNCH_ATTRIBUTE_PRIORITY` and `CUlaunchAttributeValue.priority` in both native-conversion paths in `_launch_config.pyx`. Preserve the distinction between omitted priority and explicit `priority=0`.

lijinf2:
[...]

PR #2706, opened 2026-08-26
feat(cuda.core): support CU_LAUNCH_ATTRIBUTE_PRIORITY in LaunchConfig

ichensun:
## Description

issues #2631

Adds a `priority` attribute to `LaunchConfig` that maps to [`CU_LAUNCH_ATTRIBUTE_PRIORITY`](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1g6f6565b334be6bb3134868e10bbdd331), following the same pattern used for `programmatic_stream_serialization` (#1334).

- `priority: int | None = None` (default `None`) — when omitted, the launch uses the stream's priority, matching existing `LaunchConfig` attribute conventions.
- The value is validated in `__init__` against the device's stream priority range: a nonzero `priority` outside `[greatestPriority, leastPriority]`, as returned by `cuCtxGetStreamPriorityRange`, raises `ValueError`.
- The attribute is stored as a C `int` with `0` meaning "unset", so `config.priority` reads back `0` (not `None`) when omitted, and both native-config conversion paths (`LaunchConfig._to_native_launch_config` and the module-level `_to_native_launch_config`) use a truthiness check, emitting no launch attribute for `0`. This makes an explicit `priority=0` indistinguishable from an unset priority, which is behaviorally equivalent in practice: on devices without multiple stream priorities both bounds are `0`, and elsewhere `0` is `leastPriority`, so omitting the attribute and inheriting the stream's priority yields the same scheduling.
- Tests cover the getter/setter (including that direct assignment bypasses therange check), the native attribute mapping for unset/`0`/nonzero priorities, and an updated `LaunchConfig` repr pattern in `test_object_protocols.py`.

### DIFF
--- a/cuda_core/cuda/core/_launch_config.pxd
+++ b/cuda_core/cuda/core/_launch_config.pxd
@@ -16,7 +16,7 @@ cdef class LaunchConfig:
         public int shmem_size
         public bint is_cooperative
         public bint programmatic_stream_serialization
-        public object priority
+        public int priority
 
         vector[cydriver.CUlaunchAttribute] _attrs
         object __weakref__
--- a/cuda_core/cuda/core/_launch_config.pxd
+++ b/cuda_core/cuda/core/_launch_config.pxd
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # SPDX-License-Identifier: Apache-2.0
 
@@ -16,6 +16,7 @@ cdef class LaunchConfig:
         public int shmem_size
         public bint is_cooperative
         public bint programmatic_stream_serialization
+        public object priority
 
         vector[cydriver.CUlaunchAttribute] _attrs
         object __weakref__
--- a/cuda_core/cuda/core/_launch_config.pyi
+++ b/cuda_core/cuda/core/_launch_config.pyi
@@ -2,7 +2,7 @@
 
 from typing import Any
 
-_LAUNCH_CONFIG_ATTRS = ('grid', 'cluster', 'block', 'shmem_size', 'is_cooperative', 'programmatic_stream_serialization')
+_LAUNCH_CONFIG_ATTRS = ('grid', 'cluster', 'block', 'shmem_size', 'is_cooperative', 'programmatic_stream_serialization', 'priority')
 __all__ = ['LaunchConfig']
 
 class LaunchConfig:
@@ -39,6 +39,9 @@ class LaunchConfig:
         Whether to allow programmatic stream serialization (PDL). When True,
         the kernel may overlap with a previous kernel in the same stream that
         signals completion via programmatic means.
+    priority : int, optional
+        Execution priority of the kernel. Lower numbers represent higher
+        priorities. When omitted, the launch uses the stream's priority.
     """
     grid: tuple[Any, ...]
     cluster: tuple[Any, ...]
@@ -46,8 +49,9 @@ class LaunchConfig:
     shmem_size: int
     is_cooperative: bool
     programmatic_stream_serialization: bool
+    priority: object
 
-    def __init__(self, grid: int | tuple[int, ...] | None=None, cluster: int | tuple[int, ...] | None=None, block: int | tuple[int, ...] | None=None, shmem_size: int | None=None, is_cooperative: bool=False, programmatic_stream_serialization: bool=False) -> None:
+    def __init__(self, grid: int | tuple[int, ...] | None=None, cluster: int | tuple[int, ...] | None=None, block: int | tuple[int, ...] | None=None, shmem_size: int | None=None, is_cooperative: bool=False, programmatic_stream_serialization: bool=False, priority: int | None=None) -> None:
         """Initialize LaunchConfig with validation.
 
         Parameters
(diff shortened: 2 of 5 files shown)


cuda_core/cuda/core/_launch_config.pyi
```diff
@@ -39,15 +39,19 @@ class LaunchConfig:
         Whether to allow programmatic stream serialization (PDL). When True,
         the kernel may overlap with a previous kernel in the same stream that
         signals completion via programmatic means.
+    priority : int, optional
```
lijinf2, line 42: Do we have a range for the value we can set? It cannot be any number. 
CUDA doc should provide guidance on what value to set. Let's include the guidance in docstring.

cuda_core/cuda/core/_launch_config.pyi
```diff
@@ -39,15 +39,19 @@ class LaunchConfig:
...
+        Execution priority of the kernel. Lower numbers represent higher
+        priorities. When omitted, the launch uses the stream's priority.
     """
     grid: tuple[Any, ...]
     cluster: tuple[Any, ...]
     block: tuple[Any, ...]
     shmem_size: int
     is_cooperative: bool
     programmatic_stream_serialization: bool
+    priority: object
```
lijinf2: Let's define type as "int" here to be consistent with line 42.

cuda_core/cuda/core/_launch_config.pyx
```diff
@@ -169,6 +178,11 @@ cdef class LaunchConfig:
             attr.value.programmaticStreamSerializationAllowed = 1
             self._attrs.push_back(attr)
 
+        if self.priority is not None:
```
lijinf2: Let's follow cuda.core.Stream(priority) value checking and default behavior [here](https://github.com/NVIDIA/cuda-python/blob/9ca387e98ae6e9f8519caedf2f8f109a7be4a84c/cuda_core/cuda/core/_stream.pyx#L164) to stay consistent.

cuda_core/cuda/core/_launch_config.pyi
ichensun: The range depends on the device, so there's no single fixed number to document. It's queried at runtime via `cuCtxGetStreamPriorityRange()`, which reports the valid bounds for the currently active context. On a device that doesn't support multiple stream priorities, both bounds come back as 0.

cuda_core/cuda/core/_launch_config.pxd
```diff
@@ -16,6 +16,7 @@ cdef class LaunchConfig:
         public int shmem_size
         public bint is_cooperative
         public bint programmatic_stream_serialization
+        public object priority
```
lijinf2: Let's use int instead of obejct

sylvesterkaczmarek (commented):
There is still a semantic gap in the current head around `priority=0`.

`StreamOptions` uses `None` as the sentinel and treats `0` as a real priority. This PR now stores `LaunchConfig.priority` as an `int`, but both construction and native conversion use truthiness (`if priority:` / `if self.priority:`). That makes an explicit `priority=0` indistinguishable from not specifying the launch attribute at all.

That matters when the launch runs on a nonzero-priority stream. For example, if the stream priority is `-1` and the caller passes `LaunchConfig(priority=0)`, the expected behavior is to emit `CU_LAUNCH_ATTRIBUTE_PRIORITY=0` and override the stream priority. The current code emits no priority attribute, so the launch inherits `-1` instead. This also conflicts with the PR description's original distinction that `0` is meaningful rather than an unset sentinel.

Could we keep the public/int storage requested in the earlier review but track whether the option was explicitly supplied separately (for example with a `_priority_is_set` flag), and add a regression that distinguishes `None` from `0`?

sylvesterkaczmarek (commented):
Rechecked current head `070353d8`. The `priority=0` ambiguity I raised is now resolved by making the contract explicit: both the API docs and implementation treat `0` as the unset/inherit-stream-priority value, and the native-conversion tests cover that behavior. This is internally consistent now; no remaining concern from my earlier review.

cuda_core/cuda/core/_launch_config.pyx
```diff
@@ -117,6 +138,26 @@ cdef class LaunchConfig:
...
+            if res_code != cydriver.CUresult.CUDA_SUCCESS:
+                if res_code == cydriver.CUresult.CUDA_ERROR_INVALID_CONTEXT:
+                    raise RuntimeError(
+                        "No current CUDA context. Call dev.set_current() before creating a LaunchConfig with a priority."
+                    )
+                HANDLE_RETURN(res_code)
+            prio = priority
+            if not (low <= prio <= high):
+                raise ValueError(f"{priority=} is out of range {[low, high]}")
+            self.priority = prio
```
lijinf2, line 159: The above code is identical to that in the stream priority setter. Please merge the common code.

lijinf2 (approved):

cuda_core/cuda/core/_launch_config.pyx
lijinf2: Can be done in a follow up.

lijinf2: Tracked in https://github.com/NVIDIA/cuda-python/issues/2793. Please let me know if you have bandwidth to address it. @ichensun

lijinf2:
> Rechecked current head `070353d8`. The `priority=0` ambiguity I raised is now resolved by making the contract explicit: both the API docs and implementation treat `0` as the unset/inherit-stream-priority value, and the native-conversion tests cover that behavior. This is internally consistent now; no remaining concern from my earlier review.

Got it. Yes, priority can be set through either stream or launch config. Stream setting works for all launches. Thank you for helping check the edge case.

PR #2708, opened 2026-08-27
refactor(pathfinder): author candidates in search order

rwgk:
## Description

Follow-up to #2689 and the ordering discussion in
https://github.com/NVIDIA/cuda-python/pull/2689#discussion_r3867646316.

Pathfinder historically authored dynamic-library filenames in ascending order,
then reversed them at each runtime lookup site. That distinction no longer
serves the descriptor catalog and makes preference mistakes easier.

This PR:

- authors Linux SONAMEs and Windows DLL names directly in runtime preference
  order, most preferred first;
- makes filesystem discovery, already-loaded detection, and native loader
  search consume the descriptor tuples as-is;
- preserves the historical order of the private legacy `SUPPORTED_*`
  projections at their compatibility boundary; and
- documents that ordered candidates for dynamic libraries, headers, binary
  utilities, static libraries, and bitcode libraries are consumed first to
  last.

The tuple rewrite is mechanically behavior-neutral: every multi-name dynamic
library tuple is the exact reverse of its previous catalog representation, and
the corresponding runtime reversal is removed.

## Validation

- [x] Full `cuda-pathfinder` suite: 1,608 passed, 5 skipped on CUDA 13.3.
- [x] Repository-wide pre-commit suite (excluding the network-only link
  checker).

### DIFF
--- a/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/supported_nvidia_libs.py
+++ b/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/supported_nvidia_libs.py
@@ -24,6 +24,12 @@ _OTHER_DESCRIPTORS = tuple(desc for desc in DESCRIPTOR_CATALOG if desc.packaged_
 _DRIVER_DESCRIPTORS = tuple(desc for desc in DESCRIPTOR_CATALOG if desc.packaged_with == "driver")
 _NON_CTK_DESCRIPTORS = _OTHER_DESCRIPTORS + _DRIVER_DESCRIPTORS
 
+
+def _legacy_least_preferred_first(names: tuple[str, ...]) -> tuple[str, ...]:
+    """Preserve the historical ordering of legacy filename projections."""
+    return tuple(reversed(names))
+
+
 SUPPORTED_LIBNAMES_COMMON = tuple(desc.name for desc in _CTK_DESCRIPTORS if desc.linux_sonames and desc.windows_dlls)
 SUPPORTED_LIBNAMES_LINUX_ONLY = tuple(
     desc.name for desc in _CTK_DESCRIPTORS if desc.linux_sonames and not desc.windows_dlls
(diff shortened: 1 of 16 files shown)


cuda_pathfinder/cuda/pathfinder/_dynamic_libs/supported_nvidia_libs.py
```diff
@@ -60,14 +66,26 @@
 DIRECT_DEPENDENCIES_CTK = {desc.name: desc.dependencies for desc in _CTK_DESCRIPTORS if desc.dependencies}
 DIRECT_DEPENDENCIES = {desc.name: desc.dependencies for desc in DESCRIPTOR_CATALOG if desc.dependencies}
 
-SUPPORTED_LINUX_SONAMES_CTK = {desc.name: desc.linux_sonames for desc in _CTK_DESCRIPTORS if desc.linux_sonames}
-SUPPORTED_LINUX_SONAMES_OTHER = {desc.name: desc.linux_sonames for desc in _OTHER_DESCRIPTORS if desc.linux_sonames}
-SUPPORTED_LINUX_SONAMES_DRIVER = {desc.name: desc.linux_sonames for desc in _DRIVER_DESCRIPTORS if desc.linux_sonames}
+SUPPORTED_LINUX_SONAMES_CTK = {
+    desc.name: _legacy_least_preferred_first(desc.linux_sonames) for desc in _CTK_DESCRIPTORS if desc.linux_sonames
+}
```
isVoid, line 71: I think the order of libnames are actually non-essential here, is that correct? Not a change request, just confirming my thoughts. I see that the only place uses the global variable is [here](https://github.com/NVIDIA/cuda-python/blob/809407c1fc3b9f458b0db0080738a57075d6db7a/cuda_pathfinder/cuda/pathfinder/_dynamic_libs/supported_nvidia_libs.py#L66) and [this test](https://github.com/NVIDIA/cuda-python/blob/809407c1fc3b9f458b0db0080738a57075d6db7a/cuda_pathfinder/tests/test_load_nvidia_dynamic_lib.py#L35). And neither depend on the order, actually, they use set-like operation at either places.

isVoid (approved):
Just confirming one thing below, else lgtm.

cuda_pathfinder/cuda/pathfinder/_dynamic_libs/supported_nvidia_libs.py
rwgk: Yes, that is correct. There are a few additional in-tree uses of these mappings, but they are all order-insensitive: they use dictionary keys, membership, parametrization, or uniqueness checks. Runtime discovery and loading consume the descriptor tuples directly.

The reversal here is deliberately conservative. supported_nvidia_libs.py is retained for legacy compatibility, and although these constants are private implementation details, client code could have imported and iterated, indexed, or compared their tuple values. Preserving their historical order costs only this small conversion and keeps the refactor mechanically strictly behavior-neutral.

PR #2720, opened 2026-08-28
Avoid absolute generated source paths in cuda.core builds

rwgk:
## Description

Follow-up to #2472.

PR #2472 correctly placed Cython-generated sources in a CUDA-major-keyed directory anchored to the checkout. Cython then returns those generated C++ source paths as absolute paths. When setuptools maps an absolute source into `build/temp`, it reproduces the checkout path beneath that directory; sufficiently deep Windows worktrees can therefore exceed the linker path limit.

This change converts absolute extension source paths to paths relative to the project root after `cythonize()` returns. It preserves the anchored, CUDA-major-keyed generated-source cache from #2472, while preventing setuptools from embedding the full checkout path in its temporary object and linker output paths. Existing relative C++ source paths remain unchanged.

## Verification

- Fixes ctk-next CI failures observed under PR 552

## Checklist

### DIFF
--- a/cuda_core/build_hooks.py
+++ b/cuda_core/build_hooks.py
@@ -174,6 +174,15 @@ def record_build_major() -> None:
     _BUILD_MAJOR_STAMP.write_text(_determine_cuda_major_version() + "\n", encoding="utf-8")
 
 
+def _relativize_extension_sources(extensions) -> None:
+    """Keep absolute source paths out of setuptools' temporary build tree."""
+    for extension in extensions:
+        extension.sources = [
+            os.path.relpath(source, start=Path.cwd()) if os.path.isabs(source) else source
+            for source in extension.sources
+        ]
+
+
 def _build_cuda_core(debug=False):
     # Customizing the build hooks is needed because we must defer cythonization until cuda-bindings,
     # now a required build-time dependency that's dynamically installed via the other hook below,
@@ -292,6 +301,10 @@ def _build_cuda_core(debug=False):
         compile_time_env=compile_time_env,
         **extra_cythonize_kwargs,
     )
+    # Cython returns generated sources under the absolute build_dir above.
+    # setuptools mirrors absolute source paths into build/temp, which can push
+    # MSVC linker output paths past MAX_PATH in deeper Windows checkouts.
+    _relativize_extension_sources(_extensions)
 
     return
 
--- a/cuda_core/tests/test_build_hooks.py
+++ b/cuda_core/tests/test_build_hooks.py
@@ -267,6 +267,19 @@ class TestGeneratedSourceDirIsKeyed:
         assert build_dir.parent.parent == build_hooks._BUILD_MAJOR_STAMP.parent
 
 
+class TestSetuptoolsSourcePaths:
+    @pytest.mark.agent_authored(model="gpt-5.6-sol")
+    def test_absolute_sources_are_made_relative(self, tmp_path, monkeypatch):
+        monkeypatch.chdir(tmp_path)
+        generated = tmp_path / "build" / "cython" / "cu13" / "cuda" / "core" / "_device.cpp"
+        relative = "cuda/core/_cpp/helper.cpp"
+        extension = build_hooks.Extension("cuda.core._device", [str(generated), relative])
+
+        build_hooks._relativize_extension_sources([extension])
+
+        assert extension.sources == [os.path.relpath(generated, start=tmp_path), relative]
+
+
 def _load_setup_py(monkeypatch):
     """Import setup.py for its command classes.
 


leofang:
Eye-balled the build time. Linux seems fine. Not sure if on Windows the average 10 mins is normal these days (I haven't followed closely since I took PTO).

rwgk:
## Cache and Windows long-path considerations

**TL;DR:**

**This change has one expected one-time cost:** the first `cuda.core` Linux builds using the new relative source paths will not reuse `sccache` entries created with the old absolute source paths. We could avoid that transition by applying the path relativization only on Windows. However, we prefer the simpler long-term design: give setuptools the same checkout-independent source paths on every platform, accept one cache-key migration, and avoid maintaining a Windows-only branch in the build logic.

___

Optional details:

The expected steady state is not a lower cache hit rate. The compiler input changes from a deterministic absolute path such as:

```text
/project/cuda_core/build/cython/cu13/...cpp
```

to the equivalent deterministic relative path:

```text
build/cython/cu13/...cpp
```

No timestamps, random temporary directories, or other per-run values are introduced. Once the relative-path entries are populated on `main`, identical compilations should again resolve to identical cache keys. Relative paths may also be more reusable when checkout roots differ.

### Which caches are involved

The repository's [GitHub Actions cache inventory](https://github.com/NVIDIA/cuda-python/actions/caches) includes several mechanisms that are easy to conflate:

- **Mini-CTK archives (`mini-ctk-*`)** cache downloaded CUDA components. Their keys depend on CUDA version, host platform, and the component-list hash. This PR does not affect them.
- **Compiler objects (`sccache/*`)** let Linux wheel builds reuse C/C++ compilation results. `sccache` considers preprocessed input, compiler arguments, and source-path information when computing cache identity, so changing the source path from absolute to relative necessarily creates a new generation of entries. See the [`sccache` caching documentation](https://github.com/mozilla/sccache/blob/main/docs/Caching.md).
- **Cython's generated-source cache**, established by #2472 under `cuda_core/build/cython/cu<major>`, is local build state rather than a GitHub Actions cache entry. This PR preserves both that directory and its CUDA-major keying. It only changes how the generated source filenames are subsequently presented to setuptools.
- **Workflow artifacts**, including built wheels, are separate from all of the above.

GitHub Actions caches are also ref-scoped: a workflow first searches its current ref's caches and can fall back to the default branch, but caches populated on the PR branch do not pre-warm `main`. See GitHub's [dependency caching reference](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching). Therefore, after this PR is merged, the first relevant `main` build is expected to populate the new relative-path `cuda.core` entries. Existing absolute-path entries will simply age out.

### What the CI evidence shows

The cache transition is visible in one representative Linux AMD64 / Python 3.12 job:

| Build | `cuda.core` compiler-cache hits |
|---|---:|
| [Exact base commit](https://github.com/NVIDIA/cuda-python/actions/runs/33128534328/job/98712386477) | 48/49 (97.96%) |
| [Initial PR build](https://github.com/NVIDIA/cuda-python/actions/runs/33141839942/job/98754192459) | 2/49 (4.08%) |
| [PR rerun](https://github.com/NVIDIA/cuda-python/actions/runs/33141839942/job/99560453768) | 2/49 (4.08%) |

The corresponding `cuda.bindings` compilation, whose source-path handling is unchanged, remained at 30/30 hits. This isolates the cold entries to the intended `cuda.core` path-key transition. The PR rerun does not by itself demonstrate a fully warmed steady state: these caches are ref-scoped, the matrix writes them concurrently, and the PR branch's cache population will not transfer to `main`.

Importantly, there is no observed wall-clock regression. Averaged across all eight wheel-build jobs per platform:

| Platform | Exact base | Initial PR run | PR rerun |
|---|---:|---:|---:|
| Windows x86-64 | 10m31s | 10m36s | 10m37s |
| Linux x86-64 | 6m08s | 5m53s | 5m20s |
| Linux ARM64 | 6m34s | 6m38s | 5m33s |

Windows wheel builds do not currently use `sccache`, so the roughly ten-minute Windows duration is normal and is unrelated to this change. The Windows mini-CTK cache continued to work.

### Relationship to Windows long-path support

Enabling and validating Windows long-path support on the internal runners is a worthwhile separate infrastructure goal. It improves those controlled environments and reduces the chance that unrelated tooling encounters legacy path limits.

It is not, however, a complete substitute for this change. Windows long-path behavior depends on both system policy and application/tool support, and we do not control every environment in which CUDA Python may be built or tested. That includes public CI runners, downstream automation, SWQA/DVS-SC systems, developer worktrees, and user build environments.

This PR removes avoidable path amplification: setuptools no longer reproduces the full checkout path beneath `build/temp` merely because Cython returned an absolute generated-source filename. Keeping those temporary paths short makes the build more resilient when long-path support is absent, incomplete, or outside our control. We should pursue runner enablement independently, while retaining this small defensive improvement in the package build itself.

In summary, the informed tradeoff is to accept a bounded, one-time Linux `sccache` repopulation in exchange for simpler cross-platform build logic and greater resilience across Windows environments. No ongoing cache-efficiency regression is expected.

leofang (approved):
Since the build time is not impacted (after retry), I'll approve. Though I still think that making sure the long path support is turned-on on Windows is a more elegant solution than messing with build systems.

rwgk:
Thanks @leofang!

> Though I still think that making sure the long path support is turned-on on Windows is a more elegant solution than messing with build systems.

I agree, I'll work on that. The solution here is great to sidestep the long path issues when it's not possible to reconfigure a machine quickly; it might help us down the road on other machines that we don't have direct control over.

I'll also work the long path smoke test into the internal CI.

PR #2727, opened 2026-08-28
fix(cuda.core): decode signed long long field values

rluo8:
## Description
This is to fix internal bug 6683380 found during adding tests.

Fix `FieldValue.value` decoding for `SIGNED_LONG_LONG` values by reading the `nvmlValue_t.sllVal` member exposed as `sll_val`, instead of the nonexistent `ll_val` member.

Add a regression test covering signed long long field value decoding.


mdboom (approved):
Good catch, thanks!

PR #2728, opened 2026-08-28
coverage: add cuda.core tests for graph, IPC, launcher, program, and DLPack

rluo8:
## Description

Adds focused unit tests across `cuda.core`, targeting input-validation branches, error paths, and object-lifecycle behavior that were previously uncovered. All tests are real-path: host-only tests exercise pure helpers / argument validation directly, and GPU-required tests use the existing fixtures and skips. 

Measured effect on `cuda/core/` line coverage: about +1.65%, +208 lines. Largest gains in `_program.pyx`, `graph/_subclasses.pyx`, `_memory/_ipc.pyx`, `_memoryview.pyx`, `_kernel_arg_handler.pyx`, and `graph/_graph_builder.pyx`.
**Note** that the coverage numbers are for `cuda/core/` only, combined across local systems with Linux, Windows, H800, B10, and B100.

| File | Tests |
|---|---|
| `graph/test_graph_builder.py` | `join` type check; `__cuda_stream__` after close; capture guards (`end_building` / `debug_dot_print` / `callback` / `create_condition` / `embed` / `if_then`); factory-only `GraphBuilder`/`Graph`; `close()` ends an active capture |
| `graph/test_graph_definition.py` | HOST `AllocNode` reconstructs `memory_type` from the driver via a pred walk (`CUDA_ERROR_NOT_SUPPORTED` skip) |
| `graph/test_graph_definition_lifetime.py` | pred-chain `_create_from_driver` for memcpy / event record / event wait / free |
| `graph/test_graph_node_update.py` | memset/memcpy owner + no-op + 2D geometry updates; factory-only `ExecutableGraphNode`; executable-node `repr`; clustered/cooperative `update` reject; ctypes host-callback `repr` |
| `memory_ipc/test_errors.py` | factory-only IPC types; raw-fd `from_allocation_handle`; `ForkingPickler` fd dup; registry dedup of repeated imports (Linux) |
| `test_launcher.py` | numpy scalar-subclass `prepare_numpy_arg` fallback (kernel readback); `CUgraphConditionalHandle` packed as uint64 |
| `test_memory.py` | `Buffer()` forbidden; context-manager close; abstract `MemoryResource` stubs; `copy_to()` without a memory resource |
| `test_optional_dependency_imports.py` | NVVM import rejects cuda-bindings &lt; 12.9 and caches the failure |
| `test_program.py` | nested `define_macro` error; each NVVM-unsupported option; NVRTC `as_bytes` sequence flags; debug temp-write fallback; `use_libdevice` resolves `__nv_sin` |
| `test_utils.py` | CAI-only proxy `.view()`; `view(dtype=)` itemsize mismatch |
| `test_utils_dlpack.py` | CPU array-interface export; Buffer re-view owner reuse; CUDA `stream=None`; versioned readonly; malformed dtype; C-exchange API NULL/scalar/device-accessible cases |


mdboom:
Looks mostly fine to me, but I think @Andy-Jost may want to review the graph stuff.

cuda_core/tests/test_utils_dlpack.py
```diff
@@ -76,6 +80,31 @@ def test_dlpack_export_roundtrip_special_shapes(shape):
     _assert_dlpack_export_roundtrip(np.zeros(shape, dtype=np.complex128))
 
 
+@pytest.mark.agent_authored(model="gpt-5.6-sol")
+def test_dlpack_export_array_interface_reports_cpu(init_cuda):
+    """An array-interface view without a DLPack tensor exports as CPU memory."""
+    src = np.arange(6, dtype=np.int32)
+    view = StridedMemoryView.from_array_interface(src)
+    assert view.is_device_accessible is False
+    assert view.device_id == init_cuda.device_id
```
Andy-Jost: This asserts the current-device value, but the `StridedMemoryView` docstring says `device_id` "is -1 for CPU tensors", and the DLPack import path already returns -1 for `kDLCPU`. `_smv_from_array_interface` is the outlier: it sets `buf.device_id = handle_return(driver.cuCtxGetDevice())`.

Recommend fixing the source in this PR with a one-liner (`buf.device_id = -1` in `_smv_from_array_interface`) and changing this assertion to `assert view.device_id == -1`. That way the new test pins the documented behavior rather than the inconsistency.

cuda_core/tests/test_program.py
```diff
@@ -1243,3 +1235,161 @@ def test_nvrtc_compile_with_logs_capture(init_cuda):
...
+        {"split_compile": 2},
+        {"fdevice_syntax_only": True},
+        {"minimal": True},
+    ],
+    ids=lambda kw: next(iter(kw)),
+)
+@pytest.mark.agent_authored(model="gpt-5.6-sol")
+def test_nvvm_options_reject_each_unsupported_flag(kwargs):
+    """Every NVVM-unsupported option is rejected, named, and reported alone."""
+    # This table mirrors _prepare_nvvm_options_impl's rejection list one-for-one.
```
Andy-Jost, line 1289: `use_bundled_headers` landed in `_prepare_nvvm_options_impl`'s rejection list yesterday (#2753), so the table is one short of one-for-one. Recommend adding `{"use_bundled_headers": True}` to the parametrize list (after `include_path`, matching the source order) so the comment stays true.

Andy-Jost (changes requested):
Looks good overall. I just found two issues that need to be corrected.

cuda_core/tests/test_utils_dlpack.py
rluo8: Thanks @Andy-Jost . I have updated the codes.

cuda_core/tests/test_program.py
rluo8: Thanks @Andy-Jost . Added use_bundled_headers to the list.

Andy-Jost (approved):
LGTM!

PR #2729, opened 2026-08-28
Fix the lychee precommit hook

mdboom:
The lychee precommit hook (which is independently released from lychee itself) recently changed behavior where it needs an actual lychee version, but it converts the hash to a rev using /our/ tags rather than it's own.

Fortunately, we can override this behavior with an environment variable, so the workaround is not to bad and does not reduce the security benefits of hash-pinning.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -97,6 +97,7 @@ repos:
     hooks:
       - id: lychee
         args:
+          - LYCHEE_VERSION=0.24.2  # Must match the pinned rev above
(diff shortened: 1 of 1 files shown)


mdboom:
The reason we didn't catch this in CI is because the pre-commit CI caches its environment based on our `.pre-commit-config.yml` content, so lychee was never getting reinstalled.  You have to create a new environment locally to reproduce the issue.

seberg (approved):
Weird need, but clearly OK to do.

Issue #2722, opened 2026-08-28
EPIC: Fix the nightly CI failures


rparolin:
The nightly CI has failed on **every scheduled run in the retained window — 25/25 from Aug 3 to Aug 27**. The `CI: Coverage` workflow has been red since Aug 15, and the nightly pixi GPU job has never passed since it was added in June.

This epic collects the work to get all of them green.

### The five failures

| # | Issue | Failure | Shape of the problem |
|---|-------|---------|----------------------|
| 1 | #2723 | `coverage html` aborts | Cython's coverage plugin cannot report on `.pxd`/`.pxi` files; the `.coveragerc` ignore list names 3 files by hand and two new ones landed |
| 2 | #2724 | `nvidia-smi: command not found` | Container never gets GPU access — `NVIDIA_VISIBLE_DEVICES` is not set |
| 3a | NVIDIA/numba-cuda#907 | numba-cuda suite dies at import | numpy 2.5 removed `np.row_stack`; numba-cuda registers an overload for it at import time |
| 3b | #2726 + NVIDIA/numba-cuda-mlir#286 | 2 numba-cuda-mlir tests fail | A version gate turned off 12 test-deselects when only 11 should have gone |
| 3c | #2725 | released cuda-core dies at collection | `cuda.bindings._test_helpers` was deleted without a shim; frozen released code still imports it |
| — | #2204 | *(no notifications)* | Why all of the above stayed invisible; pre-existing, folded into this epic |

mdboom:
Also I would suggest #2730 so this kind of thing would become less likely in the future.

mdboom:
**IMPORTANT**: Now we have an over-arching problem with nightly CI not covered here: Since jobs are limited to an hour, nightly runs are now ending as "canceled" rather than "passing".  We will need to separate work out to multiple jobs or some other strategy to reduce the overall runtime.

PR #2296, opened 2026-07-02
CI: retry arm64 GH200 nightly-standard row + add GB300

leofang:
Re-enable the `gh200` `nightly-standard` row in `ci/test-matrix.yml` (previously disabled in 4c70cfac05 because the runner hung on `cudaMallocAsync`). Runner team indicated the pool-side issue has been fixed, so let's give it another CI-visible run.

Also added access to gb300, which exposes a minor issue in the way we generate object/library files for testing.

Also fixed a few nightly CI issues due to the recent numba-cuda-mlir v0.4.1 release

PR #2730, opened 2026-08-28
Make it easy to run nightly tests from a PR

mdboom:
There are scenarios in which we expect there is a high risk that the nightly tests will break:

1) Modifying the nightly testing CI itself
2) Deep changes to CI (such as the recent refactoring of the test_helpers)
3) Changes we no affect some of the optional dependents, like pytorch

In these cases it would be way better to test the PR fully before merging.

This adds a feature so if the user sets the `test-nightly` label on a PR it will go through full nightly testing in addition to the regular CI testing.  It should be used sparingly due to CI cost, but I think this is very preferable to finding out about failures post-merging and then having to do a bunch of back-filling (see #2722).

### DIFF
--- a/.github/workflows/ci-nightly.yml
+++ b/.github/workflows/ci-nightly.yml
@@ -7,13 +7,33 @@
 # test suite on runners reserved for nightly-only use (e.g. arm64 l4×2), and
 # exercises release-time CI helper scripts so they do not silently rot between releases.
 #
-# This workflow does NOT build wheels — it downloads them from the latest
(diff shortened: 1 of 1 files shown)


leofang:
FYI, when I set up the nightly runs and needed to test changes in the PR, I just pushed [a commit like this](https://github.com/NVIDIA/cuda-python/pull/2296/commits/8d51cf761dda86fc29eeb3e68aba8860e366ef19) and got it reverted before merging.

PR #2731, opened 2026-08-28
Fix #2725: Add temporary shims for old cuda_bindings._test_helpers

### LINKED ISSUE #2725: [BUG]: cuda.bindings._test_helpers removed without a shim — breaks released-cuda-core nightly since Aug 4
Part of #2722.

The nightly `Nightly cuda-core (released)` jobs have failed since **2026-08-04** with a pytest **collection** error — zero tests run.

### Symptom

[Run 33038082047](https://github.com/NVIDIA/cuda-python/actions/runs/33038082047), jobs 98405282548 (linux-64) and 98405280561 (win-64):

```
________________________ ERROR collecting tests/system _________________________
tests/system/conftest.py:14: in <module>
    from cuda.bindings._test_helpers.arch_check import hardware_supports_nvml
E   ModuleNotFoundError: No module named 'cuda.bindings._test_helpers'
##[error]Process completed with exit code 2.
```

Confirmed identical on the Aug 20 nightly (job 96295460281), so this has been constant for 3+ weeks.

### Root cause

This job is **deliberately mismatched**: it runs the *released* cuda-core test suite against a cuda-bindings wheel built from *main*, specifically to catch main-branch changes that break already-released cuda-core.

- test tree → **released** cuda-core v1.1.1, checked out at `cuda-core-released/cuda_core/tests/`
- library under test → cuda-bindings from **main**: `cuda_bindings-13.3.2.dev244+g0477f61f6`

PR #2384 (`19e66499c20`, 2026-08-04, "Fix #2377: Reorganize the test helpers (where appropriate) to `cuda_python_test_helpers`") deleted the module from the cuda-bindings package:

```
 cuda_bindings/cuda/bindings/_test_helpers/__init__.py           |   6 --
 cuda_bindings/cuda/bindings/_test_helpers/arch_check.py         |  70 -------
 cuda_python_test_helpers/cuda_python_test_helpers/arch_check.py |  93 +++++++
 cuda_core/tests/system/conftest.py                              |  28 -----
```

That PR correctly updated `cuda_core/tests/system/conftest.py` **on main**. But released cuda-core is frozen and still carries the old import. New bindings + old import + deleted module → collection error.

Verified: `git ls-files cuda_bindings/cuda/bindings/_test_helpers/` returns nothing at HEAD.

### Two things this is *not*

- **Not a Python 3.14 issue.** The linux-64 row runs **Python 3.12.14** and fails with the identical message.
- **Not a CUDA 13.3.0 API change.** Nothing in the traceback reaches a CUDA call, and the released-cuda-core matrix contains only 13.3.0 rows — there is no 12.9.1 contrast to draw a conclusion from. Both rows fail; the job is 100% dead, not conditionally dead.

### Fix

**Preferred:** restore `cuda/bindings/_test_helpers/` as a thin deprecation shim re-exporting from `cuda_python_test_helpers`, for one release cycle:

```python
# cuda_bindings/cuda/bindings/_test_helpers/arch_check.py
"""Deprecated shim. Moved to cuda_python_test_helpers in #2384 (2026-08-04).
Kept for one release so released cuda-core test trees keep importing.
Remove after cuda-core >= 1.2 is the oldest supported release."""
try:
    from cuda_python_test_helpers.arch_check import *  # noqa: F401,F403
except ImportError:  # test-helper package not installed in this environment
    def hardware_supports_nvml(*args, **kwargs):
        return True   # or vendor the real implementation here
```

**Guard the re-export.** A bare `from cuda_python_test_helpers...` makes the shim depend at import time on a test-only package; if that package is not installed in this job, the shim just swaps one `ModuleNotFoundError` for another. Either guard as above or vendor the handful of symbols the released conftest actually imports — today that is only `hardware_supports_nvml`. Worth confirming which package the job installs before choosing.

This addresses the general defect rather than the symptom: **an underscore-private module that a released sibling package imports is, in practice, part of the contract with that sibling.** Removing it needs the same deprecation cycle a public symbol would get.

**Alternatives, both worse:** pinning the job to released cuda-bindings defeats its purpose; patching the released conftest in CI hides the breakage from anyone running the released suite outside CI.

### Prevention

Any PR deleting or moving a module under `cuda/` should trigger the released-compat job at PR time rather than waiting for the nightly. Given a fixed CTK release train, gates have to be PR-time to matter.

### Severity

No user-facing defect — `_test_helpers` is private and no shipped code path imports it. But **this job's single purpose is to catch main-vs-released drift, and it has been sitting on a real instance of exactly that for three weeks without anyone noticing**, because the nightly was already red. Anyone running released cuda-core's own suite against a nightly cuda-bindings wheel hits this today.

mdboom:
🎶 **IMPORTANT NOTE:** This whole issue will naturally disappear at the next `cuda-core` release.  We should confirm it's worth merging this and removing it later vs. just waiting.

Since the bug appears only in nightly CI runs, this was specially [tested here](https://github.com/NVIDIA/cuda-python/actions/runs/33186394099), and it fixes the two affected jobs:

- [win64](https://github.com/NVIDIA/cuda-python/actions/runs/33186394099/job/98900483880)
- [linux64](https://github.com/NVIDIA/cuda-python/actions/runs/33186394099/job/98900480999)

This is complex -- we need to actually import two different versions of `cuda_python_test_helpers` simultaneously -- the one expected by `cuda-core-1.1.1` and the one expected by `cuda-bindings` main.  This reveals perhaps a fundamental problem with the design of `cuda_python_test_helpers` that should be addressed in a follow-up but this papers over the nightly tests for now so that we get actual compatibility testing between `cuda-core [latest release]` and `cuda-bindings [main]`.

(It does not fix the other subissues of #2722).

### DIFF
--- a/.github/RELEASE-core.md
+++ b/.github/RELEASE-core.md
@@ -85,6 +85,13 @@ This must happen *before* the release tag is cut. Removals are breaking
 changes, so they are only permitted at a major-version boundary per the
 [support policy](https://nvidia.github.io/cuda-python/cuda-core/latest/support.html).
 
+There may also be deprecations in `cuda-bindings` tied to a particular `cuda-core` release.
+Find those this way:
+
+```console
+$ grep -rn 'deprecated:: cuda-core' cuda_bindings/cuda
+```
+
 ---
 
 ## Finalize the doc update, including release notes
--- a/cuda_bindings/cuda/bindings/_test_helpers/__init__.py
+++ b/cuda_bindings/cuda/bindings/_test_helpers/__init__.py
@@ -1,0 +1,84 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Deprecated shim package.
+
+These helpers moved to ``cuda_python_test_helpers`` in #2384 (2026-08-04).
+This package is kept for one cuda-core release cycle so that already-released
+cuda-core test trees (<=1.1.x), which still import from here, keep working
+against newer cuda-bindings wheels. See #2725.
+
+The idea here is to import the current cuda_python_test_helpers that cuda_bindings
+expects, even if the real cuda_python_test_helpers name is already claimed by an older,
+incompatible copy.
+
+.. deprecated:: cuda-core-1.2
+    Remove this entire subpackage once cuda-core >= 1.2 is the oldest supported
+    release.
+"""
+
+import importlib
+import importlib.util
+import pathlib
+import sys
+
+# Private alias under which a modern (post-#2384) cuda_python_test_helpers
+# source checkout is loaded, when the real `cuda_python_test_helpers` name is
+# already claimed by an older, incompatible copy. See _import_current below.
+_CURRENT_ALIAS = "_cuda_bindings_test_helpers_current"
+
+
+def _find_current_test_helpers_package():
+    """Locate and load cuda_python_test_helpers source checkout under a private
+    alias name, bypassing sys.modules entirely.
+    """
+    if _CURRENT_ALIAS in sys.modules:
+        return sys.modules[_CURRENT_ALIAS]
+
+    candidates = [*pathlib.Path(__file__).parents, pathlib.Path.cwd(), *pathlib.Path.cwd().parents]
+    seen: set[pathlib.Path] = set()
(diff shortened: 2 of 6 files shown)


mdboom:
With the release of cuda-core 1.2.0 yesterday, the nightly tests have resolved themselves on their own, so no need to merge this and add extra complexity.

https://github.com/NVIDIA/cuda-python/actions/runs/33711185743/job/100510881962

Closing as fixed.

PR #2733, opened 2026-08-29
fix(ci): use numba-cuda-mlir wheel-only skips

### LINKED ISSUE #2726: [BUG]: numba-cuda-mlir cu13 nightly — version gate drops the cuobjdump deselect along with the fixed #135 ones
Part of #2722. Upstream skip-guard bug now filed: NVIDIA/numba-cuda-mlir#286.

`Nightly numba-cuda-mlir (linux-64) / Python 3.12, CUDA 13.3.0` fails with 2 test failures. The CUDA 12.9.1 row passes.

### Symptom

[Run 33038082047](https://github.com/NVIDIA/cuda-python/actions/runs/33038082047), job 98405280699 — `2 failed, 4615 passed, 62 skipped, 301 xfailed`. Both failures are the same test, two parametrizations:

```
_ TestLinkerDumpAssembly.test_nvjitlink_jit_with_linkable_code_lto_dump_assembly_warn

E   FileNotFoundError: [Errno 2] No such file or directory: 'cuobjdump'
E   RuntimeError: cuobjdump has not been found. You may need to install the CUDA toolkit
E                 and ensure that it is available on your PATH.
E   Failed: DID NOT WARN. No warnings of type (<class 'UserWarning'>,) were emitted.
```

### Root cause — a stale version gate, not a CUDA 13 regression

`test-wheel-linux.yml:495` guards the **entire** deselect list on one version comparison:

```bash
if python -c "from packaging.version import Version; import sys; \
    sys.exit(0 if Version('${NUMBA_CUDA_MLIR_VER}') <= Version('0.4.1') else 1)"; then
  DESELECTS+=( ... 12 entries ... )
fi
```

The cu13 row now resolves **`NUMBA_CUDA_MLIR_VER: 0.5.0`**, so the guard is false and **all twelve deselects are dropped**. The log confirms the invocation carried none:

```
+ pytest -rxXs -v --ignore=tests/benchmarks --ignore=tests/doc_examples tests/
```

The gate conflates two unrelated problems that the workflow's own comment carefully distinguishes:

- **NVIDIA/numba-cuda-mlir#135** — serial-pytest state contamination. Genuinely fixed upstream; retiring those 11 deselects at 0.5.0 is **correct**.
- **The cuobjdump deselect** — annotated in-repo as *"not on PATH in the base ubuntu:24.04 container … No upstream fix yet."* This is an **environment** condition with no version fix, and the same gate retired it anyway.

**Why cu12.9.1 passes:** that row still resolves numba-cuda-mlir ≤ 0.4.1, so the guard holds and the workaround stays applied. CUDA 13.3.0 is simply which row upgraded first — nothing in the traceback touches a CUDA API.

### Fix

**Preferred — fix the environment, not the symptom.** Put `cuobjdump` on `PATH` in the container (install `cuda-cuobjdump` for the matrix CUDA major, or export the local CTK `bin/`). This removes the failure *and* re-enables the 4 currently-skipped `test_dwarf_cubin.py` tests — strictly more coverage:

```
SKIPPED [4] tests/infra/test_dwarf_cubin.py:41: cuobjdump not found (set CUDA_HOME or ensure cuobjdump is on PATH).
```

**Minimum — split the gate** so the environment workaround stops riding on an unrelated version check:

```bash
# Environment-conditional: cuobjdump is absent from the base ubuntu:24.04
# container. Version-independent; do NOT fold into the #135 gate below.
# Upstream skip-guard: NVIDIA/numba-cuda-mlir#286
DESELECTS+=(
  --deselect 'tests/numba_cuda_tests/cudadrv/test_nvjitlink.py::TestLinkerDumpAssembly::test_nvjitlink_jit_with_linkable_code_lto_dump_assembly_warn'
)

# NVIDIA/numba-cuda-mlir#135 — fixed in 0.5.0.
if [ ... <= 0.4.1 ]; then
  DESELECTS+=( ...the 11 contamination deselects... )
fi
```

### Wider lesson

Version-gated test workarounds should be gated **individually**. A single shared gate will eventually retire a workaround whose cause was never fixed — which is exactly what happened here. Captured in #2722.

### Severity

Low as a defect — 2 failures out of 4615, purely environmental, no product bug and no CUDA 13 incompatibility. Non-trivial as a cost, because it red-lights the whole nightly. Cheapest of the five in #2722 to fix.

rwgk:
## Description

closes #2726

## Explanation of the failure reported in #2726

The numba-cuda-mlir nightly jobs install the latest compatible release and then check out that release's matching test suite. The failing nightly resolved numba-cuda-mlir v0.5.0.

All ten numba-cuda-mlir `--deselect` entries were guarded by `NUMBA_CUDA_MLIR_VER <= 0.4.1`. Because v0.5.0 did not satisfy that condition, none of the deselects were passed to pytest. That was correct for nine of the entries, which worked around an old-version bug, but the remaining entry addressed a wheel-only environment constraint instead of a version-specific failure.

`test_nvjitlink_jit_with_linkable_code_lto_dump_assembly_warn` exercises some linkable inputs that require `cuobjdump`. The wheel-only Linux runner does not provide that executable. CUDA Python also did not set `NUMBA_CUDA_MLIR_TEST_WHEEL_ONLY`, so the test's existing wheel-only skip guard was not activated. The affected inputs therefore attempted to invoke the missing executable and failed.

## Why none of the `--deselect` entries are needed anymore

The ten entries fall into two classes:

### Nine version-gated deselects

Nine array-slicing and CUDA Array Interface tests were deselected to work around NVIDIA/numba-cuda-mlir#135 in numba-cuda-mlir v0.4.1 and earlier. The fix shipped in v0.4.2.

Current CI installs the latest compatible numba-cuda-mlir release, currently v0.5.0, and no configured nightly row resolves v0.4.1 or earlier. These nine deselects therefore no longer protect any supported CI configuration.

### One `cuobjdump`-related deselect

The remaining entry deselected the complete nvJitLink test to avoid requiring `cuobjdump`. The test already calls `self.skipTest()` for only the linkable inputs in `require_cuobjdump` when `NUMBA_CUDA_MLIR_TEST_WHEEL_ONLY` is set.

This PR sets `NUMBA_CUDA_MLIR_TEST_WHEEL_ONLY=1` for wheel-only (`LOCAL_CTK != 1`) numba-cuda-mlir jobs on Linux and Windows, matching numba-cuda-mlir's own wheel-test contract. The `fatbin` and object-file inputs that require `cuobjdump` are skipped, while the archive, cubin, and PTX inputs continue to run. The whole-test deselect is therefore unnecessary.

The existing benchmark and documentation-example collection ignores remain unchanged because they address unrelated imports of the `numba` package.

## Testing

## Checklist

- [x] Existing upstream tests cover these changes.
- [x] Documentation is not affected by this CI-only change.

### DIFF
--- a/.github/workflows/test-wheel-linux.yml
+++ b/.github/workflows/test-wheel-linux.yml
@@ -478,7 +478,13 @@ jobs:
 
       - name: Run numba-cuda-mlir tests
         if: ${{ inputs.test-mode == 'nightly-numba-cuda-mlir' && env.NUMBA_CUDA_MLIR_VER != '' }}
+        env:
+          LOCAL_CTK: ${{ matrix.LOCAL_CTK }}
         run: |
+          if [[ "${LOCAL_CTK}" != 1 ]]; then
+            # Let numba-cuda-mlir skip inputs that require toolkit executables.
+            export NUMBA_CUDA_MLIR_TEST_WHEEL_ONLY=1
+          fi
           pushd numba-cuda-mlir-released
           # Install this tag's test deps (pytest + plugins + ml-dtypes + ...).
           pip install --upgrade "pip>=25.1"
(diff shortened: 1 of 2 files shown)


rwgk:
**The targeted validation passed: PR #2733 fixes issue #2726.** The workflow remains red only because of unrelated failures.

| Linux CUDA 13.3 MLIR | Original run | Validation run |
|---|---:|---:|
| Passed | 4615 | 4615 |
| Skipped | 62 | 64 |
| Failed | 2 | 0 |
| Xfailed | 301 | 301 |

In the [original job](https://github.com/NVIDIA/cuda-python/actions/runs/33038082047/job/98405280699), two subtests failed because `cuobjdump` was missing. In the [new job](https://github.com/NVIDIA/cuda-python/actions/runs/33295034906/job/99213146895):

- `NUMBA_CUDA_MLIR_TEST_WHEEL_ONLY=1` was exported.
- Pytest ran without any deselects.
- The exact two failures became two skips: `wheel-only environments do not have cuobjdump`.
- Every other test count remained identical.

**That is unusually clean confirmation of the fix.** The new run also used PR workflow SHA `35071a635...` and artifacts from successful CI run [33292118985](https://github.com/NVIDIA/cuda-python/actions/runs/33292118985), built from that same SHA.

**Corrections To Issue #2726**

The issue’s central diagnosis is right, but two details are inaccurate:

- The old workflow had ten deselects—nine for numba-cuda-mlir #135 and one for nvJitLink—not twelve.
- The old CUDA 12.9 row also installed numba-cuda-mlir 0.5.0. It passed because the affected tests skipped earlier for missing NVCC test binaries, not because it resolved version 0.4.1 or earlier.

Therefore, the evidence actually strengthens the PR’s rationale: every tested row now uses 0.5.0, and none needs the old version-gated deselects.

**Remaining Red Jobs**

Both runs had 42 jobs: 31 succeeded and 11 failed. The original #2726 failure disappeared but was replaced by one unrelated Windows MLIR flake:

- Six recurring numba-cuda failures involving NumPy’s removed `row_stack`.
- Two recurring cuda-core failures involving missing `cuda.bindings._test_helpers`.
- One recurring Windows PyTorch access-violation flake.
- One new [Windows CUDA 12.9 MLIR failure](https://github.com/NVIDIA/cuda-python/actions/runs/33295034906/job/99213146108): a device-record assertion comparing identical-looking records containing `NaN`.

That last test uses an uninitialized `np.recarray` in [test_cuda_devicerecord.py](/wrk/forked/numba-cuda-mlir/tests/numba_cuda_tests/cudadrv/test_cuda_devicerecord.py:69), and it passed in the old run under a different random seed. The new environment variable is referenced only by nvJitLink tests in [test_nvjitlink.py](/wrk/forked/numba-cuda-mlir/tests/numba_cuda_tests/cudadrv/test_nvjitlink.py:113), so it cannot cause the device-record failure.

**My conclusion: the run is sufficient targeted evidence that #2733 fixes #2726 and is ready for review.**

brandon-b-miller (approved):
LGTM. Should be able to remove the workaround once https://github.com/NVIDIA/numba-cuda-mlir/pull/287 is released and then we follow up with releasing https://github.com/NVIDIA/cuda-python/pull/2739 and depending on that in numba-cuda-mlir instead of its existing mechanisms

